PR #1, opened 2023-02-12
allow controlling sparkline direction

orhun:
> Upstream: [#218](https://github.com/fdehau/tui-rs/pull/218)

I am rendering cpu and memory events, and would like them to stream in from the right to the left. However, if I populate the dataset by appending new events to the end, the data doesn't always fill the screen, since the data size does not necessarily match the final width of the rendered box. This allows rendering the elements from starting from the right, with left to right remaining the default.


joshka:
Implemented suggested changes and rebased on current main over in #134

PR #2, opened 2023-02-12
feat(sparkline): add show baseline option

orhun:
> Upstream: [#220](https://github.com/fdehau/tui-rs/pull/220)

null

### DIFF
--- a/src/widgets/sparkline.rs
+++ b/src/widgets/sparkline.rs
@@ -31,6 +31,8 @@ pub struct Sparkline<'a> {
     /// The maximum value to take to compute the maximum bar height (if nothing is specified, the
     /// widget uses the max of the dataset)
     max: Option<u64>,
+    /// If true, draws a baseline of `bar::ONE_EIGHTH` spanning the bottom of the sparkline graph
+    show_baseline: bool,
     /// A set of bar symbols used to represent the give data
     bar_set: symbols::bar::Set,
     // The direction to render the sparkine, either from left to right, or from right to left
@@ -50,6 +52,7 @@ impl<'a> Default for Sparkline<'a> {
             style: Default::default(),
             data: &[],
             max: None,
+            show_baseline: false,
             bar_set: symbols::bar::NINE_LEVELS,
             direction: RenderDirection::LeftToRight,
         }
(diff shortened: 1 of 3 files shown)


src/widgets/sparkline.rs
```diff
@@ -153,57 +175,127 @@ impl<'a> Widget for Sparkline<'a> {
...
+        assert_buffer_eq!(buffer, Buffer::with_lines(vec!["xxx█▇▆▅▄▃▂▁ "]));
+    }
+
+    #[test]
+    fn it_renders_baseline() {
+        let widget = Sparkline::default()
+            .data(&[0, 1, 2, 3, 4, 5, 6, 7, 8])
+            .show_baseline(true);
+        let buffer = render(widget, 12, 1);
+        assert_buffer_eq!(buffer, Buffer::with_lines(vec!["▁▁▂▃▄▅▆▇█▁▁▁"]));
```
joshka, line 289: I'm unsure whether a sparkline with a baseline should render empty values as empty or with a baseline.
`"▁▁▂▃▄▅▆▇█▁▁▁"` Baseline even when empty or:
 `"▁▁▂▃▄▅▆▇█   "` No baseline when there is no value.
 
The second option makes more sense to me, or perhaps make it configurable (use an enum rather than bool for the config setting).

src/widgets/sparkline.rs
```diff
@@ -153,57 +175,127 @@ impl<'a> Widget for Sparkline<'a> {
...
+            // the parts above the line
+            // "      ▂▅█xxxxxx",
+            // "   ▁▄▇███xxxxxx",
+            // " ▃▆██████▁▁▁▁▁▁",
+            Buffer::with_lines(vec![
+                "      ▂▅█      ",
+                "   ▁▄▇███      ",
+                " ▃▆██████▁▁▁▁▁▁",
+            ])
+        );
```
joshka, line 262: Added a failing test for this - I'm curious what @cjbassi might suggest is the correct implementation of this?

joshka:
Triage changes:
- rebased on #161 (to make tests easier to write with assert_buffer_eq!)
- updated commit message to conventional commit
- updated to use ratatui rather than tui
- combined with updated sparkline features (bar_sets and direction)
- added a few more tests including one which is broken, and a couple which I'm not 100% certain are correct (and perhaps indicate a direction for how this should work using an enum rather than a bool)

As this is not quite ready, I'm moving it to draft.

joshka:

PR #4, opened 2023-02-12
feat(table): allow page up/down in table

orhun:
NOTE FOR IMPLEMENTER: Please read the following before picking this up for implementation:
- https://github.com/tui-rs-revival/ratatui/issues/174

> Upstream: [#246](https://github.com/fdehau/tui-rs/pull/246)

Hi!

I started playing with the table example and though that for bigger tables it would be nice to be able to scroll the data a page at a time, so I altered the `Table` widget's `TableState` to store a page size for the table and its `render` function to update this at the relevant point. I have also updated the example for the `Table` widget to use this.

Let me know if you think this is a good addition or any other feedback. 😁

### DIFF
--- a/examples/table.rs
+++ b/examples/table.rs
@@ -12,38 +12,28 @@ use ratatui::{
 };
 use std::{error::Error, io};
 
-struct App<'a> {
+struct App {
     state: TableState,
-    items: Vec<Vec<&'a str>>,
+    items: Vec<Vec<String>>,
 }
 
-impl<'a> App<'a> {
-    fn new() -> App<'a> {
+impl App {
+    fn new() -> App {
+        let mut items = vec![];
+        let max_row = 100;
+        items.resize(max_row, vec![]);
+        for row in 0..100 {
+            for column in 0..3 {
+                items[row].push(format!("{}.{}", row, column))
+            }
+        }
+
         App {
             state: TableState::default(),
-            items: vec![
-                vec!["Row11", "Row12", "Row13"],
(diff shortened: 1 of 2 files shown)


joshka:
Rebased on main and this runs. This works the same as the implementation in tui for this change did, but the behavior seems a little odd.

[asciicast](https://asciinema.org/a/luyYvLcKKUw5aNweSYE1Sc22n)

Hitting page down when row 0 is selected with rows 0-13 showing on the screen makes item 1-14 shown on the screen with row 14 selected (and similarly page up when row 14 is selected goes to 0.

I think the more obvious behavior might to display rows 14-27 when paging down and have 14 be the selected row.

Additionally, the wrap around behavior seems odd. Going page up from row 0 shows rows 73 to 86, numerically correct, but would be odd from a user's perspective as it makes rows 87-99 effectively invisible.

Or put another way:
- Page down from 0 selected showing 0-13 should show 14-27 with 14 selected at the top, not 1-14 with 14 selected at the bottom.
- Page up from 0 should show 86-99, with 86 selected at the top.

I might be wrong on this though, so please feel free to correct me, but it took a bit of effort to understand what was going on with this and build a mental model just from using it where the selected item would end up.

I think this is a reasonable feature if the behavior is fixed to properly show all rows without skipping displaying any, and by scrolling by a page rather than an index.

Pinging @Eosis to see if they want to pick this up and complete the feature (and marking it as up for grabs / draft if not)

joshka:

PR #6, opened 2023-02-12
Added margins to tabs widget

orhun:
> Upstream: [#319](https://github.com/fdehau/tui-rs/pull/319)

Closes #299 

Margins can now be added to the tabs widget by using the `Margin` struct, which was previously used for margins between blocks I believe. The [`tabs.rs` example was updated to reflect this](https://github.com/thorlucas/tui-rs/blob/af2f0d47b1e4d4a7ca1a177e1c10d4c9aa5fd53f/examples/tabs.rs#L56).

The `vertical` field sets the margins/padding on the top and bottom of the widget, while the `horizontal` on the left and right. The tabs widget will be trimmed if these margins would push the contents off the screen. For example, setting vertical margins to 1 on a 2-line-tall tabs widget would display nothing, as the two lines would be used as margins.

The margins can be set with a `.margin(...)` builder method on the tabs widget.

Builder methods were also added for the `Margin` struct to facilitate creation. The `Margin::default()` sets both to zero, as is the visual default for other widgets, while the `Margin::default().horizontal(x)` and `.vertical(x)` builder methods set the horizontal and vertical components. `.both(xy)` sets both vertical and horizontal components to the same number.

Several new tests were added to the `widgets_tabs.rs` to ensure that the widget respects horizontal and vertical margins in a variety of cases.

Note that margins will still need to be added to other widgets in able to behave consistently with the new style of adding margins (`.margin(...)` with a `Margin` struct). These will be submitted as separate PRs.

### DIFF
--- a/examples/tabs.rs
+++ b/examples/tabs.rs
@@ -55,7 +55,7 @@ fn main() -> Result<(), Box<dyn Error>> {
                 .highlight_style(Style::default().fg(Color::Yellow))
                 .margin(Margin {
                     vertical: 0,
-                    horizontal: 2,
+                    horizontal: 1,
                 });
             f.render_widget(tabs, chunks[0]);
             let inner = match app.tabs.index {
--- a/examples/tabs.rs
+++ b/examples/tabs.rs
@@ -53,10 +53,7 @@ fn main() -> Result<(), Box<dyn Error>> {
                 .select(app.tabs.index)
                 .style(Style::default().fg(Color::Cyan))
                 .highlight_style(Style::default().fg(Color::Yellow))
-                .margin(Margin {
-                    vertical: 0,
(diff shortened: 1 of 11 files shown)


joshka:
I attempted to rebase this on main. There are significant enough changes since it was submitted that this requires a rewrite. I'm going to close this PR and add a reference to it in an issue for someone (perhaps @thorlucas?) to solve at some point when they have the inclination. Feel free to reopen this PR if you want to continue on this PR rather than cutting a new one.

Also note we recently implemented a `Padding` type that can be used rather than adding a new  `Margins` type.

joshka:
Actually reading this a bit more, I suspect that adding the Padding to block might have actually solved this particular need

PR #9, opened 2023-02-12
feat(reflow): make reflow module pub

orhun:
> Upstream: [#467](https://github.com/fdehau/tui-rs/pull/467)

I'm writing a new widget for `tui-rs` ecosystem that I eventually plan to open-source.
But the new widget needs access to either the `widgets::reflow` library, or copy the code.
Reuse seems like the better option to me.

### DIFF
--- a/src/widgets/mod.rs
+++ b/src/widgets/mod.rs
@@ -23,7 +23,7 @@ mod clear;
 mod gauge;
 mod list;
 mod paragraph;
-mod reflow;
+pub mod reflow;
 mod sparkline;
 mod table;
 mod tabs;


joshka:
@jjpe this issue was imported from tui-rs into ratatui. If you're still interested in getting it over the line, please feel free to continue the discussion on this PR.

My thoughts in avoiding the issue of public vs not is to find some middle ground. We know that the API is subject to change, but it is what it is for now, and subject to change could never happen. So it's probably more important to make your use case possible while still marking it as potentially future breaking.

I know that some libraries have used an approach that uses feature flags to mark unstable APIs. E.g. https://docs.rs/tokio/latest/tokio/index.html#unstable-features has an approach that requires developers to specifically pass `--cfg tokio_unstable` to the rustflags when building. I wonder if this would be the right middle ground?

Marking this as a draft PR as it's not yet ready to ship and probably requires the discussion to be rexolved.

orhun:

joshka:

PR #11, opened 2023-02-12
feat(border): add border! macro for easy bitflag manipulation

orhun:
> Upstream: [#484](https://github.com/fdehau/tui-rs/pull/484)

## Description
I have added the border! macro. It can be called with any combination of TOP, BOTTOM, LEFT, RIGHT, NONE, or ALL to return a  widgets::Borders object with the corrosponding flags set. This is helpful because the current methods of creating custom Borders is ugly and long. Further things such as Borders.add() or Borders.remove() do not allowing chaining which makes constructing the border you want either a multiline operation or requires oring Borders objects. This macro is cleaner, shorter, and much more readable. 
Compare:
This is a use of Borders from the widgets::Block documentation
`Block::default()
    .title("Block")
    .borders(Borders::LEFT | Borders::RIGHT)
    .border_style(Style::default().fg(Color::White))
    .border_type(BorderType::Rounded)
    .style(Style::default().bg(Color::Black));`

This is the same code with the border! macro
`Block::default()
    .title("Block")
    .borders(border!(LEFT, RIGHT))
    .border_style(Style::default().fg(Color::White))
    .border_type(BorderType::Rounded)
    .style(Style::default().bg(Color::Black));`

As you can see the second is much easier to read and understand. This is a fairly trivial example but if you need a border with three sides the differences becomes much clearer. `border!(TOP, LEFT, RIGHT)` instead of `Borders::TOP | Borders::LEFT | Borders::RIGHT`. 

The macro also provides a syntactic sugar for Borders::NONE. Instead of calling border!(NONE) or using Borders::NONE directly you can just call border!() which return Borders::NONE. 

Interally the macro functions by calling Borders::empty() and then using .insert() to insert each specified flag. If no flags are specified the macro directly returns Borders::NONE and doesn't waste effort creating an empty Borders and inserting.

I have done my best to add inline documentation to the addition but I'm unsure if I did it correctly/formatted it correctly. it currently passes cargo test but a quick glance over it would be appreciated.

## Testing guidelines
I have added tests for the macro to [tests](../tests/) in border_macro.rs. These include tests for ALL, specific sides, and the syntactic sugar for NONE: border!(). All tests pass on debian linux when running cargo test.

## Checklist


joshka (approved):
LGTM

sayanarijit:
I think we discussed earlier to avoid proc macros. Similar to macros for borders, it can be also used for style, constraints and entire widgets. But the more macros we add the worse it'll get. They are slow to compile and difficult to debug. cc @orhun @mindoodoo

joshka:
> I think we discussed earlier to avoid proc macros. Similar to macros for borders, it can be also used for style, constraints and entire widgets. But the more macros we add the worse it'll get. They are slow to compile and difficult to debug. cc @orhun @mindoodoo

That's a fair call. Perhaps we suggest this ends up either feature flagged or in a separate lib?

sayanarijit:
Feature flag sounds good.

orhun:
Added a feature flag called `macros` and added this functionality behind it.

sayanarijit (approved):

joshka (approved):

joshka:
Merged with description:

Adds a `border!` macro that takes TOP, BOTTOM, LEFT, RIGHT, and ALL and
returns a Borders object. An empty `border!()` call returns
Borders::NONE

This is gated behind a `macros` feature flag to ensure short build
times. To enable, add the following to your `Cargo.toml`:

```toml
ratatui = { version = 0.21.0, features = ["macros"] }
```

PR #12, opened 2023-02-12
feat(widgets): add offset() and offset_mut() for table and list state

orhun:
> Upstream: [#497](https://github.com/fdehau/tui-rs/pull/497)

## Description

Adds offset access and manipulation. Fixes #495. 

## Testing guidelines

I have a few structs that keep the table state. I use this `trait ScrollableTable` for each of these to attach actions like scroll up/down, etc. I've added this `fn recenter` to test offset manipulation. `ListState` should be pretty much analogously. 

Note that I haven't added tests per se, because I think that these kinds of accessor/mutator functions are quite obvious. 

```rust
use tui::layout::Rect;
use tui::widgets::TableState;

pub trait ScrollableTable {
    fn size(&self) -> Rect;
    fn state(&self) -> &TableState;
    fn state_mut(&mut self) -> &mut TableState;

    fn recenter(&mut self) {
        if let Some(selected) = self.state().selected() {
            let height = self.size().height as usize;

            let offset = self.state_mut().offset_mut();
            *offset = selected.saturating_sub(height / 2);
        }
    }
}
```

## Checklist


sophacles (approved):

joshka (approved):
I'd suggest adding the with_offset / with_selected methods to TableState to make building state in a single statement easy, but this is fine without it.

orhun:
Updated the implementation so that Table and List state implementations are identical. Makes one think if we should have one struct such as e.g. `WidgetState` for both and deprecate the separate usage. 🤔

sayanarijit (approved):

joshka:
That's kinda what the scroll RFC talks about.

PR #14, opened 2023-02-12
Allow specifying margins for Paragraph widget

orhun:
> Upstream: [#506](https://github.com/fdehau/tui-rs/pull/506)

## Description
Margins can now be specified around the text in a `Paragraph` widget via the `.margin(...)` builder method which accepts a `Margin` struct.
There was some discussion last year in #299 about adding margins to all widgets, though it seems like there hasn't been much progress in that respect.
This PR is essentially a sibling to #319, which does the same for the `Tabs` widget instead.

I've only added a single test for now, though if you'd like more to test if the margins play well with other properties such as alignment and text wrapping, I'd be happy to do so!

## Checklist

### DIFF
--- a/src/layout.rs
+++ b/src/layout.rs
@@ -51,6 +51,33 @@ pub struct Margin {
     pub horizontal: u16,
 }
 
+impl Margin {
+    pub fn both(mut self, margin: u16) -> Self {
+        self.vertical = margin;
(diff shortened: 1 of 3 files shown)


joshka:
I believe this is made unnecessary by PR #20 which adds padding to blocks (which most / all widgets support)

PR #16, opened 2023-02-12
feat(widget): add render_ref method to Widget trait

orhun:
> Upstream: [#528](https://github.com/fdehau/tui-rs/pull/528)

## Description
This PR adds `Terminal::render_widget_ref()` and `(Stateful)Widget::render_ref()` functions that take Widgets by reference instead of taking ownership. This allows to call render on a Trait object with unknown size with is useful where the exact object implementing Widget is unknown or depends on some kind of condition. I decided not to change the `render_widget() `and `render()` signatures as to keep backwards compatibility but sadly all custom widgets are still broken because they are now required to implement `render_ref()` instead. This is just a proof of concept PR and I would appreciate any help on getting it up to the standards since this is my first time contributing to a big rust project. I made this to fix issue #526 I posted yesterday and I'm already using this PR in a project of mine. 

## Checklist

### DIFF
--- a/examples/custom_widget.rs
+++ b/examples/custom_widget.rs
@@ -19,7 +19,7 @@ impl<'a> Default for Label<'a> {
 }
 
 impl<'a> Widget for Label<'a> {
-    fn render(self, area: Rect, buf: &mut Buffer) {
+    fn render_ref(&mut self, area: Rect, buf: &mut Buffer) {
         buf.set_string(area.left(), area.top(), self.text, Style::default());
     }
 }
--- a/src/terminal.rs
+++ b/src/terminal.rs
@@ -103,6 +103,13 @@ where
         widget.render(area, self.terminal.current_buffer_mut());
     }
 
+    pub fn render_widget_ref<W>(&mut self, widget: &mut W, area: Rect)
+    where
+        W: Widget + ?Sized,
+    {
+        widget.render_ref(area, self.terminal.current_buffer_mut());
+    }
+
     /// Render a [`StatefulWidget`] to the current buffer using [`StatefulWidget::render`].
     ///
     /// The last argument should be an instance of the [`StatefulWidget::State`] associated to the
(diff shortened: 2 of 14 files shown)


joshka:
I ran into a similar issue with the same root cause (rendering consumes the widget, which makes this call sometimes annoying - particularly as a rust newb). I'd be wary of this implementation however as it breaks all custom widgets, unless I'm mistaken?

sophacles:
Widget is a trait. Why not have a RefWidget trait also? This would allow backwards compatability with the existing widgets, and allow for rendering by ref.

I think though, a bigger thing that's a bit confusing at first about the way ratatui is designed is that widgets aren't supposed to be a place to store state - (I'm not sure of the exact right wording here but heres a try) They are basically the builder pattern for all the arguments to `render`. Instead of having `draw_paragraph(text: T, block: Option<Block>, style: Option<Style>, wrap: Option<Wrap>, scroll: Option<(u16,u16)>, alignment: Option<Alignment>)`,  there's the paragraph widget that effectively does the same thing (that is the widget is created on the stack just like function args would be, and by having the `render` method consume the struct, the compiler/linker can do some nice optimization.

A similar thing is true for several of the small structs and enums that the the widgets consume for arguments (e.g. `Block` or `Style` or `Wrap`. When they are hardcoded (e.g `Paragraph.wrap(Wrap{...})`, the compiler can optimize that away.

When writing the code it can *feel* expensive...  "why am i calling this all the time?" and similar questions come up. One of the places this gets a bit hairy is that `Text`, `Spans`, and some of the widgets (like `List`) all alloc `Vec`, and then consume the vec - this can be expensive, particularly if the text or list changes often. I wonder if it would make sense to look at reducing those allocs instead somehow - e.g. `List` can be generic on `IntoIterator<Item=&ListItem>`, and `Text` heavy widgets could take `Rc<Text>` or even`AsRef<Text>`, and that way they don't have to consume the allocated Vecs and Strings each time.

joshka:
I'm going to suggest that this one be deferred until we resolve the answers to:
- #164 
- #163

orhun:

joshka:

PR #19, opened 2023-02-12
feat(list)!: allow multiple selections

orhun:
> Upstream: [#540](https://github.com/fdehau/tui-rs/pull/540)

## Description
This allows the List to select many items, and it also allows the list to either make visible an initial selected item, or a specified target item. Addresses #463.

This is something of a breaking change, and does change the semantics of ListSelect - in particular, the visible window no longer depends on the selection, but rather, a user provided offset - this allows scrolling over a list independently of any selection you've made, which I think is a far more obvious approach. This is also necessary, since allowing many selections means that any of the target selections might be the one to highlight.

Given a pariticular offset, the List will first try to read forwards until the window is full, then backwards until the window is full - so if an offset is at the end, the window will still be filled.

## Checklist


joshka:
In order to complete this PR, it requires:
- [ ] Unit tests
- [ ] Docs
- [ ] Documenting how this will / won't break existing code (take note of https://github.com/tui-rs-revival/ratatui/discussions/163 for more info)

orhun:

joshka:

PR #20, opened 2023-02-12
feat(widget): support adding padding to Block

orhun:
> Upstream: [#551](https://github.com/fdehau/tui-rs/pull/551)

## Description
If we want to render a widget inside a block with a certain distance from its borders, we need to create another `Layout` element based on the outer `Block`, add a margin and render the Widget into it. Adding a padding property on the block element skips the creation of this second Layout.

This property works especially when rendering texts, as we can just create a block with padding and use it as the text wrapper:

```rust
let block = Block::default().borders(Borders::ALL).padding(4);
let paragraph = Paragraph::new("example paragraph").block(block);
f.render_widget(paragraph, area);
```

Rendering another widget should be easy too, using the `.inner` method:

```rust
let block = Block::default().borders(Borders::ALL).padding(4);
let inner_block = Block::default().borders(Borders::ALL);
let inner_area = block.inner(area);

f.render_widget(block, area);
f.render_widget(inner_block, inner_area);
```

## Testing guidelines
The `block.rs` example contains two cases of use: one with an inner text and another with an inner block:
image

I've also changed the paragraph tests to include a padding of value 1 to the rendering tests.

## Checklist

* [X] I have documented all new additions.


appsubscribe:
Hi, do you know when we can merge this PR? it would be great if this gets merged. Thanks!

orhun:
I resolved the conflict, please review @mindoodoo @sayanarijit @sophacles @joshka

src/widgets/block.rs
```diff
@@ -143,8 +161,33 @@ impl<'a> Block<'a> {
...
+        inner.x = inner.x.saturating_add(self.padding.horizontal);
+        inner.y = inner.y.saturating_add(self.padding.vertical);
+
+        inner.width = inner.width.saturating_sub(self.padding.horizontal * 2);
+        inner.height = inner.height.saturating_sub(self.padding.vertical * 2);
+
         inner
     }
+
+    pub fn padding(mut self, padding: u16) -> Block<'a> {
```
sophacles, line 174: What if we condensed these 3 functions to one function:

```
fn padding(mut self, padding: Padding)
```

And added some constructors for padding:
```
impl Padding {
   pub fn new(x:u16, y: u16) -> Self {...}
   pub fn horizontal(val: u16) -> Self {...}
   pub fn vertical(val: u16) -> Self {...}
   pub fn uniform(val: u16) -> Self {...}
}
```
This would simplify the api for block, and also allow for easier future expansion of the concept (e.g. left, right, top, bottom being independent values).

sophacles (commented):
Overall looks good, and is functionality I was looking for not long ago. Curious about what you and the other reviewers think of my suggestion above

src/widgets/block.rs
orhun: Good point 👍🏼

orhun: Implemented in 981f1e4bf40e59e685249d226fc18b550621d973

joshka:
I'd suggest going left,right,top,bottom on the padding values instead of just horizontal/vertical. With vertical and horizontal being pub fields, that's a breaking change if done later.

Padding will be useful to insert between other things (e.g. `ListItem`s). There at least I'd expect that the top padding to be able to be set differently to the bottom padding (and there are probably use cases for left/right there too). Even the block seems like it would maybe be useful to see different values on each edge.

sayanarijit:
Since we're going with border style attributes, can we use the similar bitwise API as borders?

joshka:
LGTM (I think the only thing I can see is missing is a test for the left/right/top/bottom)

sophacles (approved):

src/widgets/block.rs
sophacles: @orhun - i think you need to resolve this conversation, I don't have buttons for it :D

joshka (approved):

PR #21, opened 2023-02-12
docs: Expand "Apps" and "Third-party" sections

orhun:
> Upstream: [#570](https://github.com/fdehau/tui-rs/pull/570)

null


sayanarijit (changes requested):
I think we need to undo the capitalization of the link titles.

sayanarijit (approved):

PR #22, opened 2023-02-12
Significantly improve `Layout` performance

orhun:
> Upstream: [#585](https://github.com/fdehau/tui-rs/pull/585)

## Main changes
I found that the current implementation of `Layout` was very heavy on allocations and clones, especially considering you're expected to use it plus the `split()` function every frame. So, I made some changes to allow `Layout` to work more on the stack.

This is what I changed `Layout` to
```rust
#[derive(Debug, Clone, PartialEq, Eq, Hash)]
pub struct Layout<'a> {
    direction: Direction,
    margin: Margin,
    constraints: &'a [Constraint],
    /// Whether the last chunk of the computed layout should be expanded to fill the available
    /// space.
    expand_to_fill: bool,
}
```
This replaces the `constraints` field with a reference to a slice, instead of a `Vec<Constraint>`. Considering that most implementations using `Layout` use static arrays anyway, this seemed like a no brainer, until I tried to use the `LAYOUT_CAHCE`. It is impossible to store this new `Layout` inside of the `LAYOUT_CACHE` because that would require that you borrow the given `Layout` for `'static`. My solution to the problem was only storing the hash of the `Layout` into the cache as opposed to moving the entire `Layout`.
```rust
macro_rules! hash_layout {
    ($self:expr, $area:expr) => {{
        let mut to_hash = ahash::AHasher::default();
        $area.hash(&mut to_hash);
        $self.margin.hash(&mut to_hash);
        $self.expand_to_fill.hash(&mut to_hash);
        $self.direction.hash(&mut to_hash);
        $self.constraints.iter().copied().for_each(|f| match f {
            Constraint::Max(max) => to_hash.write_u16(max),
            Constraint::Min(min) => to_hash.write_u16(min),
            Constraint::Ratio(left, right) => {
                to_hash.write_u32(left);
                to_hash.write_u32(right);
            }
            Constraint::Length(length) => to_hash.write_u16(length),
            Constraint::Percentage(percentage) => to_hash.write_u16(percentage),
        });
        to_hash.finish()
    }};
}

#[derive(Clone, Copy)]
#[repr(transparent)]
struct CustomHash(u64);

impl Default for CustomHash {
    #[inline]
    fn default() -> Self {
        Self(0)
    }
}

impl std::hash::Hasher for CustomHash {
    #[inline]
    fn finish(&self) -> u64 {
        self.0
    }

    #[inline]
    fn write(&mut self, _: &[u8]) {
        panic!("unsupported operation");
    }

    #[inline]
    fn write_u64(&mut self, i: u64) {
        self.0 = i;
    }
}

thread_local! {
    static LAYOUT_CACHE: RefCell<HashMap<u64, Vec<Rect>, BuildHasherDefault<CustomHash>>> = RefCell::new(HashMap::default());
}

pub fn split(&self, area: Rect) -> Vec<Rect> {
    // TODO: Maybe use a fixed size cache ?
    LAYOUT_CACHE.with(|c| {
        c.borrow_mut()
            .entry(hash_layout!(self, area))
            .or_insert_with(|| split(area, self))
            .clone()
    })
}
```
This is the code I used for storing the layout. Here is a breakdown

- First I create a hasher using `ahash`, It's not critical that `ahash` be the hashing algorithm (the default hasher works as well). However, considering how often the `split()` function will be run, I think it's worth the sacrifice of one extra dependency.
- After creating the hasher, I feed it the `area` parameter, and then each field of the `Layout` one by one. `cargo +nightly bench` insisted that this was faster than simply running `self.hash(&mut to_hash);`
- After computing the hash of `area` + `self`, I use the final value as the final key into the `LAYOUT_CAHCE`.
- Because `LAYOUT_CACHE` uses `CustomHash` as it's `RandomState`, the hash computed by `ahash` is simply passed through and no further hashing is done. There's no point in hashing a hash.
- Finally, the resulting `Vec<Rect>` is either grabbed from the `LAYOUT_CACHE` or computed, and then cloned to bypass the reference created by the `HashMap`

## Benchmarks
I did some benchmarks to see just how much my changes effected performance. This was my benchmark
```rust
#[bench]
fn bench_vertical_split_by_height(b: &mut test::Bencher) {
    let target = Rect {
        x: 2,
        y: 2,
        width: 10,
        height: 10,
    };

        .constraints(
            [
                Constraint::Percentage(10),
                Constraint::Max(5),
                Constraint::Min(1),
            ]
            .as_ref(),
        );

    b.iter(|| layout.split(target));
}
```
When I benchmarked the original implementation and mine, these were my results. (top one is my implementation).
```
test layout::tests::bench_vertical_split_by_height ... bench:          89 ns/iter (+/- 5)
test layout::tests::bench_vertical_split_by_height ... bench:         243 ns/iter (+/- 2)
```
My implementation seems to be ~3x faster than the original implementation.

Swapping out `ahash` with the default hasher (`std::collections::hash_map::DefaultHasher::default()`) gave me this result
```
test layout::tests::bench_vertical_split_by_height ... bench:         148 ns/iter (+/- 10)
```

## Smaller changes
I noticed that most of the methods inside of this module (`layout`) were very simple and small, so I made them `const` and `#[inline]`ed them.

## Pre-requisites
All of my changes passed the unit tests and ran the examples.

Thanks for considering my PR,
      - StratusFearMe21

### DIFF
--- a/src/layout.rs
+++ b/src/layout.rs
@@ -143,6 +143,13 @@ thread_local! {
 impl<'a> Default for Layout<'a> {
     #[inline]
     fn default() -> Layout<'a> {
+        Layout::new()
+    }
+}
+
+impl<'a> Layout<'a> {
+    #[inline]
+    pub const fn new() -> Layout<'a> {
         Layout {
             direction: Direction::Vertical,
             margin: Margin {
@@ -153,9 +160,7 @@ impl<'a> Default for Layout<'a> {
             expand_to_fill: true,
         }
     }
-}
 
-impl<'a> Layout<'a> {
     #[inline]
     pub const fn constraints(mut self, constraints: &'a [Constraint]) -> Layout<'a> {
         self.constraints = constraints;
@@ -256,17 +261,20 @@ impl<'a> Layout<'a> {
     /// );
     /// ```
 
-    pub fn split(&self, area: Rect) -> Vec<Rect> {
-        LAYOUT_CACHE.with(|c| {
+    pub fn split(&self, area: Rect) -> &'static [Rect] {
+        let vec = LAYOUT_CACHE.with(|c| {
             // SAFETY:
             // "c" is stored in a static variable
             // and can therefore have a static lifetime
+            let mut b = c.borrow_mut();
 
-            c.borrow_mut()
+            let vec = b
                 .entry(hash_layout!(self, area))
-                .or_insert_with(|| split(area, self))
-                .clone()
-        })
+                .or_insert_with(|| split(area, self));
+
+            (vec.as_ptr(), vec.len())
+        });
+        unsafe { core::slice::from_raw_parts(vec.0, vec.1) }
     }
 }
 
--- a/src/layout.rs
+++ b/src/layout.rs
@@ -612,6 +612,12 @@ impl Rect {
 mod tests {
     use super::*;
 
+    const CHUNKS: Layout = Layout::new().direction(Direction::Vertical).constraints(&[
+        Constraint::Percentage(10),
+        Constraint::Max(5),
+        Constraint::Min(1),
+    ]);
+
     #[test]
     fn test_vertical_split_by_height() {
         let target = Rect {
@@ -621,17 +627,7 @@ mod tests {
             height: 10,
         };
 
-        let chunks = Layout::default()
-            .direction(Direction::Vertical)
-            .constraints(
-                [
-                    Constraint::Percentage(10),
-                    Constraint::Max(5),
-                    Constraint::Min(1),
-                ]
-                .as_ref(),
-            )
-            .split(target);
+        let chunks = CHUNKS.split(target);
 
         assert_eq!(target.height, chunks.iter().map(|r| r.height).sum::<u16>());
         chunks.windows(2).for_each(|w| assert!(w[0].y <= w[1].y));
--- a/src/layout.rs
+++ b/src/layout.rs
@@ -598,6 +598,31 @@ mod tests {
     }
 
     #[test]
+    fn test_vertical_split_by_height_ref() {
+        let target = Rect {
+            x: 2,
+            y: 2,
+            width: 10,
+            height: 10,
+        };
+
+        let chunks = Layout::default()
+            .direction(Direction::Vertical)
+            .constraints(
+                [
+                    Constraint::Percentage(10),
+                    Constraint::Max(5),
+                    Constraint::Min(1),
+                ]
+                .as_ref(),
+            )
+            .split_ref(target);
+
+        assert_eq!(target.height, chunks.iter().map(|r| r.height).sum::<u16>());
+        chunks.windows(2).for_each(|w| assert!(w[0].y <= w[1].y));
+    }
+
+    #[test]
     fn test_rect_size_truncation() {
         for width in 256u16..300u16 {
             for height in 256u16..300u16 {
--- a/src/layout.rs
+++ b/src/layout.rs
@@ -217,7 +217,7 @@ impl<'a> Layout<'a> {
     ///     });
     /// assert_eq!(
     ///     chunks,
-    ///     vec![
+    ///     &[
     ///         Rect {
     ///             x: 2,
     ///             y: 2,
@@ -244,7 +244,7 @@ impl<'a> Layout<'a> {
     ///     });
     /// assert_eq!(
     ///     chunks,
-    ///     vec![
+    ///     &[
     ///         Rect {
     ///             x: 0,
     ///             y: 0,
--- a/src/layout.rs
+++ b/src/layout.rs
@@ -212,7 +212,7 @@ impl<'a> Layout<'a> {
     ///     });
     /// assert_eq!(
     ///     chunks,
-    ///     &vec![
+    ///     vec![
     ///         Rect {
     ///             x: 2,
     ///             y: 2,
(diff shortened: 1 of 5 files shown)


src/layout.rs
```diff
@@ -183,14 +260,65 @@ impl Layout {
...
+        //
+        //
+        // It is for the reasons that I have stated above that
+        // I believe that the use of the core::slice::from_raw_parts()
+        // function in this very
+        // specific way will not lead to undefined behaviour or
+        // safety concerns.
+
+        unsafe { core::slice::from_raw_parts(vec.0, vec.1) }
     }
```
conradludgate, line 322:
```suggestion
    pub fn split(&self, area: Rect) -> &'static [Rect] {
        LAYOUT_CACHE.with(|c| {
            let mut b = c.borrow_mut();

            let vec = b
                .entry(hash_layout!(self, area))
                .or_insert_with(|| split(area, self));

            Vec::leak(vec)
        })
    }
```

conradludgate: Ah, it's stored in the hashmap, so we don't have ownership.. I would be much happier with something closer to `Rc<[Rect]>` here than using unsafe

sayanarijit:
It's safety+readability vs speed tradeoff. Needs benchmarking, requirement evaluation and discussion.

conradludgate:
> It's safety+readability vs speed tradeoff. Needs benchmarking, requirement evaluation and discussion.

@sayanarijit No need for unsafe, as I said in my comment and shown in #62

src/layout.rs
```diff
@@ -183,14 +260,65 @@ impl Layout {
...
+
+            let vec = b
+                .entry(hash_layout!(self, area))
+                .or_insert_with(|| split(area, self));
+
+            (vec.as_ptr(), vec.len())
+        });
+
+        // SAFETY: We know 3 things about the vec variable
+        // we are deriving this slice from
```
parasyte, line 276: These are not enough.

`&'static [Rect]` is `Copy`, meaning it can escape the `thread_local`:

```rust
    #[test]
    fn test_layout_split_soundness() {
        let handle = std::thread::spawn(|| {
            Layout::default()
                .constraints([Constraint::Length(5), Constraint::Min(0)].as_ref())
                .split(Rect::new(0, 0, 10, 10))
        });

        let rect = handle.join().unwrap();

        println!("Whoopsie! UAF: {rect:?}");
    }
```

Run this with miri, `cargo miri test -- test_layout_split_soundness`:

```
running 1 test
test layout::tests::test_layout_split_soundness ... error: Undefined Behavior: constructing invalid value at .value.<enum-variant(Some)>.0.<enum-variant(Ok)>.0: encountered a dangling reference (use-after-free)
   --> /Users/parasyte/.rustup/toolchains/nightly-x86_64-apple-darwin/lib/rustlib/src/rust/library/core/src/mem/mod.rs:910:22
    |
910 |         let result = ptr::read(dest);
    |                      ^^^^^^^^^^^^^^^ constructing invalid value at .value.<enum-variant(Some)>.0.<enum-variant(Ok)>.0: encountered a dangling reference (use-after-free)
    |
    = help: this indicates a bug in the program: it performed an invalid operation, and caused Undefined Behavior
    = help: see https://doc.rust-lang.org/nightly/reference/behavior-considered-undefined.html for further information
    = note: BACKTRACE:
    = note: inside `std::mem::replace::<std::option::Option<std::result::Result<&[layout::Rect], std::boxed::Box<dyn std::any::Any + std::marker::Send>>>>` at /Users/parasyte/.rustup/toolchains/nightly-x86_64-apple-darwin/lib/rustlib/src/rust/library/core/src/mem/mod.rs:910:22: 910:37
    = note: inside `std::option::Option::<std::result::Result<&[layout::Rect], std::boxed::Box<dyn std::any::Any + std::marker::Send>>>::take` at /Users/parasyte/.rustup/toolchains/nightly-x86_64-apple-darwin/lib/rustlib/src/rust/library/core/src/option.rs:1623:9: 1623:33
    = note: inside `std::thread::JoinInner::<'_, &[layout::Rect]>::join` at /Users/parasyte/.rustup/toolchains/nightly-x86_64-apple-darwin/lib/rustlib/src/rust/library/std/src/thread/mod.rs:1458:9: 1458:72
    = note: inside `std::thread::JoinHandle::<&[layout::Rect]>::join` at /Users/parasyte/.rustup/toolchains/nightly-x86_64-apple-darwin/lib/rustlib/src/rust/library/std/src/thread/mod.rs:1590:9: 1590:22
note: inside `layout::tests::test_layout_split_soundness`
   --> src/layout.rs:632:20
    |
632 |         let rect = handle.join().unwrap();
    |                    ^^^^^^^^^^^^^
note: inside closure
   --> src/layout.rs:624:38
    |
623 |     #[test]
    |     ------- in this procedural macro expansion
624 |     fn test_layout_split_soundness() {
    |                                      ^
    = note: this error originates in the attribute macro `test` (in Nightly builds, run with -Z macro-backtrace for more info)
```

Yes, `unsafe` is Big and Bad.

parasyte (changes requested):
Please remove the `unsafe` or move `LAYOUT_CACHE` to a real static allocation. (Not a thread local.)

sayanarijit:
Suggested changes without `unsafe` has been merged.
Closing this with respect to https://github.com/tui-rs-revival/tui-rs-revival/pull/62

PR #28, opened 2023-02-12
wip(textinput): Add interactive TextInput widget, Interaction* types for interactive widgets

orhun:
> Upstream: [#639](https://github.com/fdehau/tui-rs/pull/639)

## Description

- Add widget `TextInput` - a single-line stateful widget
- Add trait `InteractionHandler` when `crossterm` is used
- Add trait `InteractiveWidget` for indicating that the widget is interactive

`InteractiveWidget` is added so `Frame` can be passed in by mut reference to the widget's render function. This is used for drawing the cursor for widgets which would like to do such a thing. A new trait is needed, or else backwards compatibility would be broken if the existing `Widget` or `StatefulWidget` traits are changed. 

The `InteractionHandler` trait depends on crossterm types, and is thus gated behind the crossterm feature. Widget state objects that want to process user input events take an event (state objects indicate if they're are focused, analogous to Dear imgui's `WantCaptureKeyboard`), and will return if they consumed the event or not (e.g. should the input be forwarded to the rest of the application).

The `TextInput` widget and demo is a concise example of how the above Interaction* types are used. It implements a single-line text input field, which responds to a subset of `readline` (think bash) keyboard shortcuts.

Shortcuts implemented:
- Go to end of line
- Go to start of line
- Delete to start of word under cursor
- Jump to next word
- Jump to previous word

My hope is that `TextInput` can provide the user a simpler and more feature-rich option than the example provided in `user_input.rs`. While the `user_input` example is very cool (and indeed what inspired this!), the lack of keyboard controls is intuitive, and it requires the user to implement state tracking. 

Consider this a work in progress, as I'd like to get your input on if this would be a good contribution to the project. If so, I'd be happy to spend the time to go add comprehensive docs + doc tests.

## Image / Recording

[asciicast](https://asciinema.org/a/nNGse9zeRaBDSGZM0x3d6wQeK)

## Testing guidelines

- Tested that examples still build when using `termion`
- `cargo test` 
- `cargo run --example text_input` - touches most features
- Unit tests added for interactive keyboard control

## Checklist

### DIFF
--- a/src/widgets/text_input/crossterm_interactive.rs
+++ b/src/widgets/text_input/crossterm_interactive.rs
@@ -1,0 +1,315 @@
+use crossterm::event::{Event, KeyCode, KeyEvent, KeyModifiers};
+
+use crate::widgets::{InteractiveWidgetState, InteractionOutcome, TextInputState};
+
+impl InteractiveWidgetState for TextInputState {
+    fn handle_event(&mut self, event: Event) -> InteractionOutcome {
+        if !self.is_focused() {
+            return InteractionOutcome::Bubble;
+        }
+
+        match event {
+            Event::Key(key) => self.handle_key(key),
+            _ => InteractionOutcome::Bubble,
+        }
+    }
+
+    fn is_focused(&self) -> bool {
+        self.is_focused()
+    }
+
+    fn focus(&mut self) {
+        self.focus()
+    }
+
+    fn unfocus(&mut self) {
+        self.unfocus()
+    }
+}
+
+impl TextInputState {
+    // used in tests
+    #[allow(dead_code)]
+    fn up_to_cursor(&self) -> &str {
+        &self.value[0..self.cursor_pos as usize]
+    }
+
+    fn handle_key(&mut self, key: KeyEvent) -> InteractionOutcome {
+        if key.modifiers == KeyModifiers::ALT || key.modifiers == KeyModifiers::CONTROL {
+            self.handle_modifiers(key.modifiers, key.code)
+        } else {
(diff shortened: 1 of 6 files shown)


src/widgets/text_input/crossterm_interactive.rs
```diff
@@ -0,0 +1,315 @@
...
+            (KeyModifiers::CONTROL, KeyCode::Char('a')) => {
+                self.cursor_pos = 0;
+            }
+            // jump back a word
+            (KeyModifiers::ALT, KeyCode::Char('b')) => {
+                self.cursor_pos = self.word_boundary_idx_under_cursor(true);
+            }
+            // jump forward a word
+            (KeyModifiers::ALT, KeyCode::Char('f')) => {
+                self.cursor_pos = self.word_boundary_idx_under_cursor(false);
```
VelvetThePanda, line 123: Are these going to be changed to match their non-modifier counterparts (and to be consistent with most terminals?)

sayanarijit:
Note: since this is a new feature, we'll skip it in the next release, and add it to a milestone. We have to decide how to add new/extra features. See meeting notes: https://github.com/tui-rs-revival/ratatui/discussions/63

joshka:
Given that there are currently at least two pretty decent crates with this functionality already (https://docs.rs/tui-textarea/latest/tui_textarea/index.html and https://docs.rs/tui-input/latest/tui_input/index.html), I'd suggest doing a gap analysis between this and the other two rather than splitting the effort 3 ways.

@rhysd / @sayanarijit / @dymk what do you think?

joshka:

PR #33, opened 2023-02-12
feat(paragraph): allow word-break in wrapped paragraphs

orhun:
> Upstream: [#653](https://github.com/fdehau/tui-rs/pull/653)

## Description
This PR will allow word-break in wrapped paragraphs.

## Testing guidelines

Create a `Paragraph`  in a `Block`, with the `break_word` value in its `Wrap` set to `true`.

Such as :

```rust
const SAMPLE_STRING: &str = "The library is based on the principle of immediate rendering with \
     intermediate buffers. This means that at each new frame you should build all widgets that are \
     supposed to be part of the UI. While providing a great flexibility for rich and \
     interactive UI, this may introduce overhead for highly dynamic content.";

fn ui<B: Backend>(frame: &mut Frame<B>) {
    let size = Rect::new(0, 0, 20, 10);
    let text = vec![Spans::from(SAMPLE_STRING)];
    let paragraph = Paragraph::new(text)
        .block(Block::default().borders(Borders::ALL))
        .alignment(Alignment::Left)
        .wrap(Wrap {
            trim: true,
            break_words: true,
        });
    frame.render_widget(paragraph, size);
}
```

If set `break_words` as `true`, the result will be like:

```shell
┌──────────────────┐
│The library is bas│
│ed on the principl│
│e of immediate ren│
│dering with interm│
│ediate buffers. Th│
│is means that at e│
│ach new frame you │
│should build all w│
└──────────────────┘
```

If set `break_words` as `false`, the result will be like:

```shell
┌──────────────────┐
│The library is    │
│based on the      │
│principle of      │
│immediate         │
│rendering with    │
│intermediate      │
│buffers. This     │
│means that at each│
└──────────────────┘
```

## Checklist

### DIFF
--- a/examples/demo/ui.rs
+++ b/examples/demo/ui.rs
@@ -293,7 +293,10 @@ where
             .fg(Color::Magenta)
             .add_modifier(Modifier::BOLD),
     ));
-    let paragraph = Paragraph::new(text).block(block).wrap(Wrap { trim: true });
+    let paragraph = Paragraph::new(text).block(block).wrap(Wrap {
+        trim: true,
+        break_words: false,
+    });
     f.render_widget(paragraph, area);
 }
 
--- a/examples/paragraph.rs
+++ b/examples/paragraph.rs
@@ -153,19 +153,28 @@ fn ui<B: Backend>(f: &mut Frame<B>, app: &App) {
         .style(Style::default().bg(Color::White).fg(Color::Black))
         .block(create_block("Left, wrap"))
         .alignment(Alignment::Left)
-        .wrap(Wrap { trim: true });
+        .wrap(Wrap {
+            trim: true,
+            break_words: false,
+        });
     f.render_widget(paragraph, chunks[1]);
     let paragraph = Paragraph::new(text.clone())
         .style(Style::default().bg(Color::White).fg(Color::Black))
         .block(create_block("Center, wrap"))
         .alignment(Alignment::Center)
-        .wrap(Wrap { trim: true })
+        .wrap(Wrap {
+            trim: true,
+            break_words: false,
+        })
         .scroll((app.scroll, 0));
     f.render_widget(paragraph, chunks[2]);
     let paragraph = Paragraph::new(text)
         .style(Style::default().bg(Color::White).fg(Color::Black))
         .block(create_block("Right, wrap"))
         .alignment(Alignment::Right)
-        .wrap(Wrap { trim: true });
+        .wrap(Wrap {
+            trim: true,
+            break_words: false,
+        });
     f.render_widget(paragraph, chunks[3]);
 }
--- a/examples/paragraph.rs
+++ b/examples/paragraph.rs
@@ -170,11 +170,11 @@ fn ui<B: Backend>(f: &mut Frame<B>, app: &App) {
     f.render_widget(paragraph, chunks[2]);
     let paragraph = Paragraph::new(text)
         .style(Style::default().bg(Color::White).fg(Color::Black))
-        .block(create_block("Right, wrap"))
+        .block(create_block("Right, wrap (break words)"))
         .alignment(Alignment::Right)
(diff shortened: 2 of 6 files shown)


pythops:
@orhun this is really needed. Any possibility to merge soon ?

sayanarijit (approved):

mindoodoo:
Barely looked at the first failing test, seems like the trim isn't working anymore, I'll try to come back to this another time as I'm not that familiar with the reflow implementation. Feel free to fix the tests if someone comes to it before I do.

joshka:
This looks like it breaks every current use of Wrap, but doesn't put anything in place to make that better (e.g. Wrap::default() with builders so that things have reasonable defaults. As a first superficial only observation, it might be nicer to use enums to drive the behavior of the wrapping here rather than bools as this allows future extension.

I'd recommend that the approach to getting this in a state ready for review would be to first spec out and agree what wrap should look like after change rather than just iterating on it without direction (I'm fine with breaking this because the feature looks fine, but it's worth noting how users of this should fix their code.

joshka:
Tasks needed to complete this:
- [ ] rebase on main (likely after #149 is merged) - this may actually be a rewrite though due to the size of #149
- [ ] design / agree what constructing a `Wrap` looks like. e.g.`Wrap::default().trim(someEnumValue).break(...)`. My guess is that this will probably want to consider using enums rather than bools for the values to make it clear how wrapping works.
- [ ] write / rewrite tests to fit in the same approach as #149
- [ ] Add proper docs to show how the feature works

TimerErTim:
In case you @orhun need help eventually after the quite big changes from #149, do let me know. I am glad to assist.

Eyesonjune18:
Sort of unrelated, but why is it that `Wrap` includes `trim` field? I feel like it has nothing to do with wrapping. Maybe make `Wrap` an enum and make `Trim` a field of `Paragraph`?

joshka:
> Sort of unrelated, but why is it that `Wrap` includes `trim` field? I feel like it has nothing to do with wrapping. Maybe make `Wrap` an enum and make `Trim` a field of `Paragraph`?

Like most things, it evolved that way from where it was before that.
The commit in question is https://github.com/tui-rs-revival/ratatui/commit/112d2a65f67e0305c43573ca0ae5cafbf882835b which links to a PR in tui-rs: https://github.com/fdehau/tui-rs/pull/327. The PR contains more detail on the design rationale.

Specifically on this one, I think this change probably will be breaking, and will require some of the questions I asked in #163 to be considered as part of rolling this out.

Aside:
I often get on my soapbox and rant that developers should add more info to their commits than they do. Here there's some interesting choices that you could only really understand by reading through the PR history. It's a little annoying that this is only available on github however, and isn't copied as part of a git clone etc. My preference would be to reduce those conversations to some bullet points in the commit that help answer questions like this directly from the commit. You can go further and invest in maintaining architecture design records, but that's probably overkill for this library.

examples/demo/ui.rs
```diff
@@ -293,7 +293,10 @@ where
             .fg(Color::Magenta)
             .add_modifier(Modifier::BOLD),
     ));
-    let paragraph = Paragraph::new(text).block(block).wrap(Wrap { trim: true });
+    let paragraph = Paragraph::new(text).block(block).wrap(Wrap {
```
igoruranskij373-collab, line 296: ### Новый продукт [название продукта] в продаже!

Привет! У нас новый продукт — [название продукта]. **Основные преимущества:** [кратко о плюсах и возможностях].

**Для кого подходит:** [кто может использовать и зачем].

**Особенности:** [что делает продукт особенным].

**О компании:** [краткая информация о нас].

**Контакты:** [имя, должность, телефон, email].

Приходите на презентацию [дата, время, место]. Вопросы? Пишите или звоните.

С уважением, 
[название компании]

PR #38, opened 2023-02-12
docs: add some lines of documentation to clear doubts

orhun:
> Upstream: [#677](https://github.com/fdehau/tui-rs/pull/677)

## Description
It is a PR just to resolve issue #352

Add a few lines to explain how to define the coordinates 

## Testing guidelines
Just test it using VS Code

## Checklist

* [X] I have documented all new additions.

### DIFF
--- a/src/layout.rs
+++ b/src/layout.rs
@@ -374,6 +374,12 @@ pub struct Rect {
 impl Rect {
     /// Creates a new rect, with width and height limited to keep the area under max u16.
     /// If clipped, aspect ratio will be preserved.
+    /// 
+    /// The X Axis represents the count of columns of the Rect.
+    /// 
+    /// The Y Axis represents the count of rows of the Rect.
+    /// 
+    /// The Rect is always drawn from the top left to the bottom and to the right of the container.
     pub fn new(x: u16, y: u16, width: u16, height: u16) -> Rect {
         let max_area = u16::max_value();
         let (clipped_width, clipped_height) =
--- a/src/widgets/canvas/line.rs
+++ b/src/widgets/canvas/line.rs
@@ -4,6 +4,14 @@ use crate::{
 };
 
 /// Shape to draw a line from (x1, y1) to (x2, y2) with the given color
+/// 
+/// The X Axis represents the count of columns of the Line.
+/// 
+/// The Y Axis represents the count of rows of the Line.
+/// 
+/// The Line is always drawn from the top left to the bottom and to the right of the container.
+/// 
+/// And when it is drawn, it will calculate the distance between the two points
 #[derive(Debug, Clone)]
 pub struct Line {
     pub x1: f64,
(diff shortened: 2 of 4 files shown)


Nickiel12 (approved):
Although it feels a little bit wordy, it will be a lot clearer for those not familiar with the ui design.

mindoodoo:
The way this PR defines X and Y is very unclear imo.

Also, looking at the original issue, it would seem like the main difference between OPs perception of the x and y coordinate system and the actual tui implementation is only that the origin is in the top left instead of bottom left.

Nickiel12:
Maybe just a note somewhere prominent that "The origin is in the upper left corner for all widgets"?

mindoodoo:
I honestly feel like it's pretty common sense that the origin is in the top left corner in tui applications. It also feels weird to specify it in this small case where as I'm sure there's dozens if not hundreds of places we could specify this in this crate.

src/layout.rs
```diff
@@ -374,6 +374,12 @@ pub struct Rect {
 impl Rect {
     /// Creates a new rect, with width and height limited to keep the area under max u16.
     /// If clipped, aspect ratio will be preserved.
+    /// 
+    /// The X Axis represents the count of columns of the Rect.
+    /// 
+    /// The Y Axis represents the count of rows of the Rect.
+    /// 
+    /// The Rect is always drawn from the top left to the bottom and to the right of the container.
```
sayanarijit, line 382: I don't think these docs are needed since the fields are/will be already documented.

src/widgets/canvas/line.rs
```diff
@@ -4,6 +4,14 @@ use crate::{
 };
 
 /// Shape to draw a line from (x1, y1) to (x2, y2) with the given color
+/// 
+/// The X Axis represents the count of columns of the Line.
+/// 
+/// The Y Axis represents the count of rows of the Line.
+///
```
sayanarijit, line 11: These should be above the respective fields.

src/widgets/canvas/line.rs
```diff
@@ -4,6 +4,14 @@ use crate::{
...
 
 /// Shape to draw a line from (x1, y1) to (x2, y2) with the given color
+/// 
+/// The X Axis represents the count of columns of the Line.
+/// 
+/// The Y Axis represents the count of rows of the Line.
+/// 
+/// The Line is always drawn from the top left to the bottom and to the right of the container.
+/// 
+/// And when it is drawn, it will calculate the distance between the two points
```
sayanarijit, line 14: These docs aren't necessary, since there's nothing unexpected here.

sayanarijit (changes requested):
Added personal opinions.

mindoodoo:
If I can get a couple upvotes on this message, I'll close this PR I think. We can take another look at it if someone requests it again.

mindoodoo:
Alright, closing this then, if someone stumbles on this and disagrees / would like to discuss it further, feel free to open a discussion.

PR #42, opened 2023-02-12
fix(widget)!: List should not ignore empty string items

orhun:
> Upstream: [#688](https://github.com/fdehau/tui-rs/pull/688)

## Description
Fixes issue [#680](https://github.com/fdehau/tui-rs/issues/680#issue-1462518032). Handles the case where a list item is created with an empty string, which is not split by the `lines` iterator.

## Testing guidelines
New test can be ran with `cargo test`

## Checklist

### DIFF
--- a/src/text.rs
+++ b/src/text.rs
@@ -293,17 +293,13 @@ impl<'a> Text<'a> {
         T: Into<Cow<'a, str>>,
     {
         let lines: Vec<_> = match content.into() {
+            Cow::Borrowed("") => vec![Spans::from("")],
             Cow::Borrowed(s) => s.lines().map(Spans::from).collect(),
+            Cow::Owned(s) if s.is_empty() => vec![Spans::from("")],
             Cow::Owned(s) => s.lines().map(|l| Spans::from(l.to_owned())).collect(),
         };
 
-        Text {
-            lines: if lines.is_empty() {
-                vec![Spans::from("")]
-            } else {
-                lines
-            },
-        }
+        Text { lines }
     }
 
     /// Create some text (potentially multiple lines) with a style.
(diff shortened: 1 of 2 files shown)


src/text.rs
```diff
@@ -292,10 +292,16 @@ impl<'a> Text<'a> {
...
+        let lines: Vec<_> = match content.into() {
+            Cow::Borrowed(s) => s.lines().map(Spans::from).collect(),
+            Cow::Owned(s) => s.lines().map(|l| Spans::from(l.to_owned())).collect(),
+        };
+
         Text {
-            lines: match content.into() {
-                Cow::Borrowed(s) => s.lines().map(Spans::from).collect(),
-                Cow::Owned(s) => s.lines().map(|l| Spans::from(l.to_owned())).collect(),
+            lines: if lines.len() == 0 {
```
sayanarijit, line 301:
```suggestion
            lines: if lines.is_empty() {
```

sayanarijit (changes requested):

mindoodoo (approved):

rhysd:
This branch needs to be rebased on the latest main branch so that newer version of Rust compiler will be used and it will fix the issue on compiling `cargo-make` crate.

orhun:
Rebased.

rhysd (approved):

sayanarijit (changes requested):
Please hold this, I think I missed to see this, it can be improved. is_empty() should be done in the input string.

sayanarijit:
Updated.

sayanarijit:
But this looks like a breaking change. Some users might be using this bug/feature as a condition, So, let's wait for a major release?

sayanarijit:
Ah, good point.

sayanarijit (approved):

mindoodoo (approved):

PR #44, opened 2023-02-12
Update example: Read input only when key was pressed

orhun:
> Upstream: [#690](https://github.com/fdehau/tui-rs/pull/690)

In the `user_input` example, it read input when key was released which is not expected.

### DIFF
--- a/examples/user_input.rs
+++ b/examples/user_input.rs
@@ -10,7 +10,7 @@
 ///   * Pressing Enter pushes the current input in the history of previous
 ///   messages
 use crossterm::{
-    event::{self, DisableMouseCapture, EnableMouseCapture, Event, KeyCode},
+    event::{self, DisableMouseCapture, EnableMouseCapture, Event, KeyCode, KeyEventKind},
     execute,
     terminal::{disable_raw_mode, enable_raw_mode, EnterAlternateScreen, LeaveAlternateScreen},
 };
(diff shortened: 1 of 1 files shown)


mindoodoo:
@a1393323447 I'm having a hard time seing how this change is needed given that it doesn't (to what I've tested) seem to change the behaviour of the example.

a1393323447:
@mindoodoo Sorry about that, maybe I should make a better description next time.
I run this example on windows. Before I change the code, it read input when I release a key. So when I click `h`, the message would be `hh`:

before

After I change the code:

after

mindoodoo:
I'm unable to replicate this on linux unfortunately. What operating system are you using ?

Nickiel12:
The original issue has been replicated on windows powershell, wezterm on windows, and the windows terminal app. All key inputs double pressing - once on the down, and once on the up.

Nickiel12:
Also reproduced on windows 11, windows 11 wezterm with powershell backend, and windows 11 powershell emulator. The issue is not present on WSL interestingly (or not, I suppose).

adamkali (commented):
as @jw013 pointed out on discord windows will set `KeyEvent::Repeat` once someone can confirm that this change will remedy this i believe we can start getting this merged in.

mindoodoo:
[Crossterm documentation](https://docs.rs/crossterm/latest/crossterm/event/struct.KeyEvent.html#structfield.kind) talking about this behaviour.

Nickiel12:
So the proposed fix will only work on windows, and will actually break it on linux without that flag being set?

adamkali:
I believe that windows is being defaulted to both so specifying won't do anything other than specifying that it will only register keypresses not keypress **AND** release

Nickiel12:
@adamkali, we have confirmation that the PR works on Linux/alacrity as well as WSL from @tranzystorek-io

adamkali (approved):

mindoodoo (approved):

PR #62, opened 2023-02-18
better safe shared layout cache

conradludgate:
## Description

https://github.com/tui-rs-revival/tui-rs-revival/pull/22#issuecomment-1427143924 introduced the idea to return `&'static [Rect]` on `Layout::split` using unsafe, but we can use a safe abstraction here instead with `Rc`.

## Testing guidelines

Using the benchmark introduced in #22, I get these results (master/this respectively)

```
test layout::tests::bench_vertical_split_by_height ... bench: 129 ns/iter (+/- 6)
test layout::tests::bench_vertical_split_by_height ... bench: 100 ns/iter (+/- 3)
```

## Checklist

### DIFF
--- a/src/layout.rs
+++ b/src/layout.rs
@@ -69,8 +69,9 @@ pub struct Layout {
     expand_to_fill: bool,
 }
 
+type Cache = HashMap<(Rect, Layout), Rc<[Rect]>>;
 thread_local! {
-    static LAYOUT_CACHE: RefCell<HashMap<(Rect, Layout), Rc<[Rect]>>> = RefCell::new(HashMap::new());
+    static LAYOUT_CACHE: RefCell<Cache> = RefCell::new(HashMap::new());
 }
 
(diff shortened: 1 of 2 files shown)


sayanarijit (approved):

sayanarijit (changes requested):
Need to fix test failure.

sayanarijit (approved):

adamkali (approved):
I think these changes are also better in the long run, using explicit `Rc<[Rect]>` could end up being more maintainable in other parts of the code. We should look into other parts of the codebase where using a safe pointer instead of a vector, to if there will be benefits there.

orhun (approved):

mindoodoo (approved):

StratusFearMe21:
Hey! This is pretty cool 😄👍! I'm glad someone was able to put my work to use.

Issue #98, opened 2023-03-18
Move "apps using ratatui" to dedicated file


mindoodoo:
Considering the quantity of apps using ratatui that are referenced in our readme, maybe we could move them to a table in a separate markdown file ?

This would make the main README leaner, while at the same time enabling us to maybe give the option for adding more details since it would now be a dedicated file.

PR #99, opened 2023-03-18
docs(apps): move 'apps using ratatui' to dedicated file (#98)

orhun:
Addresses #98


mindoodoo (commented):
I was thinking maybe we could do a table view now that it's in a dedicated folder. This would allow to have a name, description, author column. What are your thoughts on that ?

Also, we could brag about the amount of apps using ratatui, such as :

Check out the list of [close to 40]() apps using ratatui !

orhun:
> I was thinking maybe we could do a table view now that it's in a dedicated folder. This would allow to have a name, description, author column. What are your thoughts on that ?
> 
> Also, we could brag about the amount of apps using ratatui, such as :
> 
> "Check out the list of close to 40 apps using ratatui !"

PR is updated!

orhun:
We should also update the documentation/codebase about `tui` -> `ratatui` references. Can you create an issue for that? (I have to run out now.)

sayanarijit (approved):

PR #114, opened 2023-03-23
feat(terminal)!: add inline viewport

conradludgate:
originally taken from https://github.com/fdehau/tui-rs/pull/552.

Updates `Viewport` type, making it an enum with `Fullscreen`, `Inline` and `Fixed` views.
`Inline(_)` includes the height of the inline screen.

Adds an `insert_before` method that can write permanent lines above the inline tui. Intended usecase is like `yarn install`, where there would be multiple updating elements at the bottom of the screen, and then 'xxx installed' would be permanent logs above the updating elements.

I've tried to minimise the breaking changes from the original PR, but there's still a breaking change in the `Backend` trait. In theory I can add some default methods to make it not a breaking change 

(CC @pdecat, https://github.com/ellie/atuin/pull/648)

### DIFF
--- a/tests/backend_termion.rs
+++ b/tests/backend_termion.rs
@@ -15,7 +15,7 @@ fn backend_termion_should_only_write_diffs() -> Result<(), Box<dyn std::error::E
         let mut terminal = Terminal::with_options(
             backend,
             TerminalOptions {
-                viewport: Viewport::fixed(area),
+                viewport: Viewport::Fixed(area),
             },
         )?;
         terminal.draw(|f| {


mindoodoo:
I just checked out the example, seems either bugged / unfinished or I'm just not understanding what @fdehau was trying to do. On my machine the example either wrote on top of itself on my occasions, or failed to "coexist" with previous outputs in my terminal. I'll try and take a look at this PR when I have the time, but I'll be pretty busy these upcoming days.

conradludgate:
> seems either bugged / unfinished

my fault, push incoming

src/backend/mod.rs
```diff
@@ -16,15 +16,25 @@ pub use self::crossterm::CrosstermBackend;
...
+    fn append_lines(&mut self, n: u16) -> io::Result<()>;
     fn hide_cursor(&mut self) -> Result<(), io::Error>;
     fn show_cursor(&mut self) -> Result<(), io::Error>;
     fn get_cursor(&mut self) -> Result<(u16, u16), io::Error>;
     fn set_cursor(&mut self, x: u16, y: u16) -> Result<(), io::Error>;
-    fn clear(&mut self) -> Result<(), io::Error>;
+    fn clear(&mut self, clear_type: ClearType) -> Result<(), io::Error>;
     fn size(&self) -> Result<Rect, io::Error>;
     fn flush(&mut self) -> Result<(), io::Error>;
 }
```
conradludgate, line 40: Assuming we don't want to break this trait, we could do
```diff
     fn draw<'a, I>(&mut self, content: I) -> Result<(), io::Error>
     where
         I: Iterator<Item = (u16, u16, &'a Cell)>;
-    fn append_lines(&mut self, n: u16) -> io::Result<()>;
+    fn append_lines(&mut self, _n: u16) -> io::Result<()> {
+        Ok(())
+    }
     fn hide_cursor(&mut self) -> Result<(), io::Error>;
     fn show_cursor(&mut self) -> Result<(), io::Error>;
     fn get_cursor(&mut self) -> Result<(u16, u16), io::Error>;
     fn set_cursor(&mut self, x: u16, y: u16) -> Result<(), io::Error>;
-    fn clear(&mut self, clear_type: ClearType) -> Result<(), io::Error>;
+    fn clear(&mut self) -> Result<(), io::Error>;
+    fn clear_to(&mut self, clear_type: ClearType) -> Result<(), io::Error> {
+        match clear_type {
+            ClearType::All => self.clear(),
+            ClearType::AfterCursor
+            | ClearType::BeforeCursor
+            | ClearType::CurrentLine
+            | ClearType::UntilNewLine => Err(io::Error::new(
+                io::ErrorKind::Other,
+                format!("clear_type [{clear_type:?}] not supported with this backend"),
+            )),
+        }
+    }
     fn size(&self) -> Result<Rect, io::Error>;
     fn flush(&mut self) -> Result<(), io::Error>;
 }
```

But I'm not sure if that would add confusion

mindoodoo: Yeah I definitely don't think we want to break the trait.
But I don't think ```clear_to``` is the best name for the new function although to be fair I can't come up with anything better.
You could either leave it as is or if you find a better name change it. Maybe add a doc line that explains that ```clear``` will ClearType::All where as  ```clear_to``` can be specified.

Maybe ```clear_specified``` ?

ellie: `clear_to` with documentation would work, though I also think `clear_area` or `clear_region` might also make sense?

gibbz00: `clear_region` implemented in #119! :)

conradludgate: clear_region works for me too! I'll update my PR tomorrow. I'll also add some documentation. Won't hurt

ellie:
Hey! I don't want to pressure or anything, but the next [Atuin](https://github.com/ellie/atuin) release is blocked on this one getting in. Is there anything I can do to help

sayanarijit (approved):
LGTM. But this looks like a breaking change.

orhun (approved):

src/terminal.rs
```diff
@@ -1,32 +1,21 @@
...
 impl Viewport {
-    /// UNSTABLE
     pub fn fixed(area: Rect) -> Viewport {
-        Viewport {
-            area,
-            resize_behavior: ResizeBehavior::Fixed,
-        }
+        Self::Fixed(area)
     }
 }
```
sayanarijit, line 20: Let's also implement pub fn inline(u16) and fullscreen() constructors...

sayanarijit: @conradludgate

conradludgate: Actually I'd rather delete this one. It's commented as "unstable" so it shouldn't be in use anyway, and it would be a small change of
```diff
- Viewport::fixed(n)
+ Viewport::Fixed(n)
```

sayanarijit: Sounds good, as long as it's consistent.

orhun (approved):

mindoodoo (approved):

orhun:
Thanks @hermannm! I'm currently having a chat with the other maintainers about the next possible release. We certainly don't want to keep people waiting 🐻

PR #119, opened 2023-03-26
POC: Viewport scrolling

gibbz00:
This can be viewed as an extension to paragraph scroll, but where it is now possible for multiple widgets to be scrolled in unison.

[asciicast](https://asciinema.org/a/570332)

*Text fuckery is caused by the asciinema player (looks right locally).

### Implementation: 

Creates a clear separation of the corresponding sizes between `Buffer` and `Viewport`. The viewport size is always synced with the backend size, but the terminal buffer size can be resized to something larger. The buffer has no offset, but the viewport does. `render_widget()` calls are rendered to the terminal buffer, and `flush` only sends over parts of the buffer that are covered by the viewport to the backend.

There are probably dozens of optimizations and improvements to be done, and some things have been left unfinished, extensive tests, documentation etc. This PR is POC after all, so roast me into oblivion. 

Development is messy and so are my commits, but I would start looking at their messages for further explanations on what has changed and sometimes why. (All commits do on the other hand pass all test.) I've sent in plenty PRs with varying atomicity and degree of breaking changes, and rarely have I seen it play a role in the likelihood of them getting merged. The most telling factor has from my anecdotal experience instead been maintainer attitudes. So let's just first have a mature discussion about the changes I've made. What I'm trying to say then; is that I won't bother to create a pristine commit history with split up PRs before a sincere, elaborated and/or justified request is made for me to do so. 

All the best,


mindoodoo:
Closing this for the time being as the feature will be reworked by @gibbz00 in a future PR.

PR #121, opened 2023-03-28
Split scrollable viewports

gibbz00:
Basically pt. 2 of #119.

[asciicast](https://asciinema.org/a/571532)

Most of the same things apply as described in the first PR. 

### Implementation and usage

Viewport can now be split by using the `Terminal::new_split(constraints)` constructor where the constraints are those from `layout.rs`. Each constraint results in one viewport, meaning that two imply one split. The viewports can then be scrolled by calling `terminal.split_viewport_scroll(|index| -> (x_step, y_step))`. This takes a closure that returns the scroll for a given split viewport index. The index is set by the index of the respective constraint used when calling the terminal constructor.

```rust
let mut terminal = Terminal::new_split(
    backend,
    vec![Constraint::Length(3), Constraint::Min(0)],
    Direction::Vertical,
)?;

/// ...draw UI and what not.

terminal.split_viewport_scroll(|index| match index {
    1 => (0, 1),
    _ => (0, 0),
})?
.unwrap_or(())

```
`unwrap_or(())` represents the action of skipping viewport over scroll attempts. This includes both split viewport overlapping (disallowed), and when attempting to scroll beyond the buffer. `viewport_scroll()` explains this in more detail.

Also, calling `split_viewport()` will only update the region with the updated scroll. All other regions are left untouched.


mindoodoo:
Yooooo, sorry for the delay in getting back to you.

We had a chat with the maintainers a couple days ago where we discussed some open PRs, including your two open PR.
First of all, the consensus was that your feature was pretty sick, although we couldn't come up with many use cases off the top of our heads, it does seem quite promising and we're sure that if the scrollable viewport made it into the crate people would find very cool uses for it.

However, we can't really review the PRs under their current form. There are way more files modified than needed for this feature, and a lot of the modifications are also irrelevant to the feature you're adding. With that in mind, could you please  create a new PR only containing code that is relevant to the feature ?

gibbz00:
Hii 😁

Apologies on my end, I was the one that was supposed to reach out when time allowed.

Thanks nonetheless for looking into it. The reason I developed it was because I was writing a markdown reader. The idea was that there would be a fixed tab line at the top, and where the markdown view used many of the built-in widgets. (Tables, paragraphs etc...) Scrolling up and down required for all widgets to move in unison. I've now made the repo public if you're interested in watching how Ratatui works in the context of the two PRs.  https://github.com/gibbz00/grow. It is by no means a finished state, but it does achieve this concept if I remember correctly.

I completely understand the difficulty in reviewing the drafts in their current state. Making them reviewable that can realistically only be achieved by creating a series of split up PRs each containing a small subset of cherry-picked/reworked commits. 

The reason as to why I haven't heard back, is that I've lately been fully occupied with work. I can make an attempt at beginning to clean up the drafts on Sunday, but use otherwise these drafts as sample references for things that could be changed in the framework.

All the best,

mindoodoo:
Yeah no rush, whenever you get around to it we'll be ready to merge it !

You mentioned splitting it up in multiple PRs again, feel free to chat us on discord in order to potentially discuss what to split in what PRs. The goal being to avoid ending up in a situation where we have to ask you for more major edits.

I'll close both draft PRs for the time being, if you need more input on this feature I think you should potentially create an issue where other users might also be able to give you their input on the feature.

We appreciate the work you poured into this and are looking forward to seing the revamped PR(s) :smile:

PR #122, opened 2023-03-31
feat(widgets)!: make widgets optionally reusable

bolshoytoster:
In some situations, ratatui's immediate-mode rendering can be inefficient, i.e. for a list, you have to clone either the vec or it's items again each frame to make the new list object.

This commit makes it so the `{,Stateful}Widget::render` functions only take `self` by immutable reference, meaning they're not consumed and can be reused.

The `Frame::render_{,stateful_}widget` functions take `widget` by value though, and changing it would break all code using it, so I've added separate functions: `render_{,stateful_}widget_reusable` functions which take the `widget` by reference, allowing it to be reused.

This could also allow `StatefulWidget`s to store their state themselves, rather than relying on the user to store it.

I also had to remove any `take`s in the `render` functions since they would invalidate the widget.

I also ran `cargo fmt` with the default config to remove formatting accidentally done under my config. I can undo the formatting if needed.

### DIFF
--- a/examples/custom_widget.rs
+++ b/examples/custom_widget.rs
@@ -19,7 +19,7 @@ struct Label<'a> {
 }
 
 impl<'a> Widget for Label<'a> {
-    fn render(self, area: Rect, buf: &mut Buffer) {
+    fn render(&self, area: Rect, buf: &mut Buffer) {
         buf.set_string(area.left(), area.top(), self.text, Style::default());
     }
 }
--- a/src/terminal.rs
+++ b/src/terminal.rs
@@ -2,7 +2,7 @@ use crate::{
     backend::{Backend, ClearType},
     buffer::Buffer,
     layout::Rect,
-    widgets::{StatefulWidget, Widget},
+    widgets::{AsStatefulWidgetRef, AsWidgetRef, StatefulWidget, Widget},
 };
 use std::io;
 
@@ -83,11 +83,14 @@ where
     /// let mut frame = terminal.get_frame();
     /// frame.render_widget(block, area);
     /// ```
-    pub fn render_widget<W>(&mut self, widget: W, area: Rect)
+    pub fn render_widget<W, R>(&mut self, widget: R, area: Rect)
     where
         W: Widget,
+        R: AsWidgetRef<W>,
     {
-        widget.render(area, self.terminal.current_buffer_mut());
+        widget
+            .as_widget_ref()
+            .render(area, self.terminal.current_buffer_mut());
     }
 
     /// Render a [`StatefulWidget`] to the current buffer using [`StatefulWidget::render`].
@@ -115,11 +118,14 @@ where
     /// let mut frame = terminal.get_frame();
     /// frame.render_stateful_widget(list, area, &mut state);
     /// ```
-    pub fn render_stateful_widget<W>(&mut self, widget: W, area: Rect, state: &mut W::State)
+    pub fn render_stateful_widget<W, R>(&mut self, widget: R, area: Rect, state: &mut W::State)
     where
         W: StatefulWidget,
+        R: AsStatefulWidgetRef<W>,
     {
-        widget.render(area, self.terminal.current_buffer_mut(), state);
+        widget
+            .as_widget_ref()
+            .render(area, self.terminal.current_buffer_mut(), state);
     }
 
     /// After drawing this frame, make the cursor visible and put it at the specified (x, y)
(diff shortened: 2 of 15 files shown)


sophacles:
The `Vec` arguments require lots of allocs every time they are constructed then consumed, and that can get expensive. The other stuff, not so much (see also my comment on #16, which is a similar PR, tl;dr the widgets are just a way to build function argumetns). I wonder if it would make sense to instead have (e.g.) List be generic on `IntoIterator<Item=&ListItem>` or similar.

orhun:
Hey, can you revert the formatting changes?

bolshoytoster:
@orhun bam.

joshka:
I like this change - it enables some things that were a pain to work around the widget being consumed on render calls or when you can get a reference but it's more difficult to get something consumable (e.g. https://github.com/joshka/ratatui/commit/e56d8eab4c10c5e24e4596af1e7189ed630a2747 where I can easily call render on a boxed dyn widget which means the List Widget can be extended to display any widget and not just specific ones).

One possible other approach to implementing a new method would to actually break callers by making render_widget and render_stateful_widget actually borrow the widget (and probably area too) instead of having two methods. We're in a spot where it could be reasonable to make a choice to break consumers here if the outcome enables many more scenarios and makes the library more ergonomic - this is probably controversial and I'd defer to those who have more experience than I do on this.

If we keep the two methods naming, do you think `render_widget_as_ref` or `render_as_borrowed()` might be more idiomatic than `_reusable`?

bolshoytoster:
@joshka
> One possible other approach to implementing a new method would to actually break callers by making `render_widget` and `render_stateful_widget` actually borrow the widget (and probably area too) instead of having two methods.

Would it be possible to have them accept either a `Widget` or a `&Widget` (something like `AsRef<Widget>`), that would probably be the best option, but I'm not sure if that would work.

> If we keep the two methods naming, do you think `render_widget_as_ref` or `render_as_borrowed()` might be more idiomatic than `_reusable`?

You're absolutely right, `_reusable` was mainly because I couldn't think of a better name at the time.

joshka:
I couldn’t quite get that to work - if you can that would be definitely be the best way to make this work.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=f2fb5c4961dbd7201ad60ea912e44ff0

bolshoytoster:
@joshka I think you got in the playground is because `Widget` isn't local, you could try adding it to `src/widgets/mod.rs` and see if that works.

aschey:
Unfortunately, you can't make a blanket implementation for a foreign trait like that. What you can do is make your own trait that mimics `AsRef` and make a blanket implementation for that: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=9389c135b72d2ebeb1a002aa70fd1d03

joshka:
> Unfortunately, you can't make a blanket implementation for a foreign trait like that. What you can do is make your own trait that mimics `AsRef` and make a blanket implementation for that: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ff64715562c78e2d9891bb508530ce3f

Nice! That sounds like a realistic approach!

joshka:
Thanks for the fix @aschey. That approach works pretty nicely (I updated code in this PR to use that approach).

I think it's worth playing with this more to understand some of the pros and cons. We probably want to also make the area Rect param pass by ref also (that's particularly useful when rendering inner blocks).

joshka:
Updated Widget list PoC based on this: https://github.com/joshka/ratatui/commit/e911bee068236b3add700374daa8bf427e69eac9

joshka:
Updated Widget list PoC based on this: https://github.com/joshka/ratatui/commit/e911bee068236b3add700374daa8bf427e69eac9

[asciicast](https://asciinema.org/a/F0Jq1N34hG3ZShJyb3avWPREI)

joshka:
> I wonder if it would make sense to instead have (e.g.) List be generic on `IntoIterator<Item=&ListItem>` or similar.

Playing with this in the Widget List PoC and noticing what I needed to change in the tests to make them work nicely, I definitely agree with that @sophacles

sayanarijit:
This PR looks good. The scope is limited. We may need to accompany this change with many other related PRs, but as far as this PR goes, LGTM.

deifactor:
In terms of the `Widget` trait, if we want to make this change then it's better to just bite the bullet and change the type signature of `render`. It'll result in a breaking change for calls to `Frame::render_widget`, but it means that removing support for render-by-move won't result in having to do another change to rename `render_as_ref` to `render` down the line.

orhun (approved):
+1 from me. It is a change that user will definitely benefit in the long run.

joshka:
Closing with deference to https://github.com/ratatui-org/ratatui/pull/833 which achieves the same result by implementing `Widget` for `&Block` etc.

PR #129, opened 2023-04-07
fix(examples): update input in examples to only use press events

lesleyrs:
Hi I saw the `user_input` example was already updated to work on windows but I just did the rest of them that use input other than quit.

Also added a hack in panic.rs because on windows/powershell a release event is sent from when you originally press enter to `cargo run`. No more double presses now either.

Fix https://github.com/tui-rs-revival/ratatui/issues/120

Edit: I just looked at the conventional commit thing and it's just a fix, you can merge this now

### DIFF
--- a/examples/panic.rs
+++ b/examples/panic.rs
@@ -102,11 +102,6 @@ fn run_tui<B: Backend>(terminal: &mut Terminal<B>, app: &mut App) -> io::Result<
                     app.chain_hook();
                 }
 
-                // hack to make it work on powershell
-                KeyCode::Enter => {
-                    ();
-                }
-
                 _ => {
                     return Ok(());
                 }
--- a/examples/panic.rs
+++ b/examples/panic.rs
@@ -87,6 +87,7 @@ fn reset_terminal() -> Result<()> {
 }
 
 /// Runs the TUI loop.
+#[allow(clippy::no_effect)]
 fn run_tui<B: Backend>(terminal: &mut Terminal<B>, app: &mut App) -> io::Result<()> {
     loop {
         terminal.draw(|f| ui(f, app))?;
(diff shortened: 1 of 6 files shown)


lesleyrs:
Closing this because it's being reverted https://github.com/crossterm-rs/crossterm/pull/778

mindoodoo:
Any reason for the reopen ?

lesleyrs:
This can actually be merged, https://github.com/crossterm-rs/crossterm/pull/778#issuecomment-1500910673.

It's not 100% needed if crossterm puts release events behind a feature, but this will stop it from breaking whenever the feature is enabled on both windows and using kitty keyboard protocol on other platforms.

At this moment all the modified examples aren't compatible with release events, and without this PR they don't work as intended because of "double" events.

examples/panic.rs
```diff
@@ -87,6 +87,7 @@ fn reset_terminal() -> Result<()> {
 }
 
 /// Runs the TUI loop.
+#[allow(clippy::no_effect)]
```
mindoodoo, line 90: This seems unnecessary. Other than that, disregard my previous comment, until crossterm merges the opt out default, it makes sense to have this explicitly handled.

mindoodoo (changes requested):

examples/panic.rs
lesleyrs: Ah yes that was a result of not remembering to undo it after removing the powershell hack, which is no longer needed assuming the crossterm PR will be merged.

The press events check doesn't necessarily need to be removed later, as making use of the release events could have some uses.

mindoodoo (approved):
Yeah we don't know when they'll merge the PR putting this behind a feature in crossterm so makes sense to merge this for the time being !

orhun (approved):

PR #133, opened 2023-04-12
fix(canvas)!: use full block for Marker::Block

joshka:
Fixes https://github.com/tui-rs-revival/ratatui/issues/82

When drawing a vertical line on a canvas, using `Marker::Block` would
draw a gap in between each line. This is because the default block
character was a half bar. This changes the default block character to a
full block, and adds a new marker type `Marker::Bar` to replace the
current `Marker::Block`.

BREAKING CHANGE: `Marker::Block` now uses a full block character, and
`Marker::Bar` is the new half block character.

- Adds tests for each `Marker` in the canvas widget.
- Adds marker to the canvas example to show the difference between
  the marker types.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,6 +35,7 @@ serde = { version = "1", optional = true, features = ["derive"]}
 [dev-dependencies]
 rand = "0.8"
 argh = "0.1"
+indoc = "2.0"
 
 [[example]]
 name = "barchart"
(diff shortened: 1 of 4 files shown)


joshka:
Changed `.split('\n')` to `.lines()` in tests to attempt to make ubuntu happy.

joshka:
*sigh* in improving the tests, I was too hasty and introduced a typo bug (which is now fixed, and should succeed).

mindoodoo (approved):

sayanarijit (approved):

orhun (approved):

Issue #142, opened 2023-04-20
ratatui prints random bytes on exit


alimkoca:
## Description
When I execute one of the example code from your documentation, it prints random bytes on exit.

## To Reproduce
```rs
use std::{io, thread, time::Duration};
use ratatui::{
    backend::CrosstermBackend,
    widgets::{Widget, Block, Borders},
    layout::{Layout, Constraint, Direction},
    Terminal
};
use crossterm::{
    event::{self, DisableMouseCapture, EnableMouseCapture, Event, KeyCode},
    execute,
    terminal::{disable_raw_mode, enable_raw_mode, EnterAlternateScreen, LeaveAlternateScreen},
};

fn main() -> Result<(), io::Error> {
    // setup terminal
    enable_raw_mode()?;
    let mut stdout = io::stdout();
    execute!(stdout, EnterAlternateScreen, EnableMouseCapture)?;
    let backend = CrosstermBackend::new(stdout);
    let mut terminal = Terminal::new(backend)?;

    terminal.draw(|f| {
        let size = f.size();
        let block = Block::default()
            .title("Block")
            .borders(Borders::ALL);
        f.render_widget(block, size);
    })?;

    thread::sleep(Duration::from_millis(5000));

    // restore terminal
    disable_raw_mode()?;
    execute!(
        terminal.backend_mut(),
        LeaveAlternateScreen,
        DisableMouseCapture
    )?;
    terminal.show_cursor()?;

    Ok(())
}
```

## Expected behavior
To not print any byte to command line

## Screenshots
![Bildschirmfoto vom 2023-04-20 20-42-03](https://user-images.githubusercontent.com/66552942/233445884-4be02fc6-bf73-427d-a314-b3103d20b86d.png)

## Environment

- OS: Fedora 36 
- Terminal Emulator: GNOME Terminal
- Font: Monospace
- Crate version: 0.20.1
- Backend: crossterm

## Additional context
None

sophacles:
What version of crossterm did you specify in your Cargo.toml?

alimkoca:
crossterm version is 0.26 @sophacles

alimkoca:
Uh, probably main problem is related to `EnableMouseCapture`.

sophacles:
Oh I see. I can repro this when I move the mouse.

I think what's happening here is this:

[...]

PR #145, opened 2023-04-21
fix(142): cleanup doc example

sophacles:
Include a thread to handle events so that the pty's buffer doesn't fill up with events from during the program runtime and then get read by the shell and echoed on the prompt.


sophacles:
This is the doc fix for #142

orhun (approved):

PR #148, opened 2023-04-23
Spans.parse_style()

Eyesonjune18:
Added a method to `Spans` that parses the style of each `Span` in the `Spans` similarly to `Text.parse_style()`.

### DIFF
--- a/src/text.rs
+++ b/src/text.rs
@@ -46,7 +46,7 @@
 //!     Span::raw(" title"),
 //! ]);
 //! ```
-use crate::style::{Style, Color, Modifier};
+use crate::style::Style;
 use std::borrow::Cow;
 use unicode_segmentation::UnicodeSegmentation;
 use unicode_width::UnicodeWidthStr;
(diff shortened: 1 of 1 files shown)


Eyesonjune18:
I'm not sure how all the checks failed because it's literally a drop-in method that does not remove or change anything public

sophacles (commented):
I think `patch_style` could be helpful to widget authors. I think I'd like to see a similar `Span.patch_style()` also.

Eyesonjune18:
> I think `patch_style` could be helpful to widget authors. I think I'd like to see a similar `Span.patch_style()` also.

Sorry, just a typo. The method is called `patch_style()`.

sophacles:
> Sorry, just a typo. The method is called `patch_style()`.

I know - the current PR only changes `struct Spans`, It also like to see `patch_style()` for the `struct Span`. (Spans is a collectio of Span)

sophacles:
Reply to myself:

Span has a public `.style`, so one could do `.style.patch` so maybe it's not necessary.

Eyesonjune18:
Do you guys want `Text.reset_style()` and `Span.reset_style()` as well?

Eyesonjune18:
I tried to make the commit messages follow convention but it duplicated the commits :/

orhun (approved):

src/text.rs
```diff
@@ -214,6 +234,57 @@ impl<'a> Spans<'a> {
...
+    ///     Span::styled(" text", Style::default().add_modifier(Modifier::BOLD)),
+    /// ]);
+    ///
+    /// spans.reset_style();
+    /// assert_eq!(Style::reset(), spans.0[0].style);
+    /// assert_eq!(Style::reset(), spans.0[1].style);
+    /// ```
+    pub fn reset_style(&mut self) {
+        for span in &mut self.0 {
+            span.patch_style(Style::reset());
```
joshka, line 285: is it worth adding `Span.reset_style()` for api consistency?

src/text.rs
```diff
@@ -179,6 +179,26 @@ impl<'a> Span<'a> {
             })
             .filter(|s| s.symbol != "\n")
     }
+
+    /// Applies a new style to an existing Span.
```
joshka, line 183: It's probably worth noting that this is additive rather than replacing the style.

```suggestion
    /// Patch the style of the Span to be the combined style of the existing style plus the new style.
```

src/text.rs
```diff
@@ -214,6 +234,57 @@ impl<'a> Spans<'a> {
     pub fn width(&self) -> usize {
         self.0.iter().map(Span::width).sum()
     }
+
+    /// Applies a new style to each Span in an existing Spans.
```
joshka, line 238:
```suggestion
    /// Patch the style of each Span to be the combined style of the existing style plus the new style.
```

joshka (changes requested):
Some minor changes, can you squash this into a single commit with a conventional commit like `feat(widget): add style methods to Span and Spans` (or something similar). Looks like an easy merge once that's done.

Eyesonjune18:
@joshka The requested changes have been made 👍

joshka (approved):
LGTM. Thanks for your contribution.

PR #150, opened 2023-04-26
list padding and truncation

joshka:
This PR introduces two semi-related changes to the list, padding between elements and being able to set the truncation of the last item in the list.

- feat: add list padding
Adds padding between items. Items that fit at the end of the list except for the padding are still shown. Note. I wasn't sure whether this really needs top/bottom/left/right padding or if a single number is actually sufficient. Left and right padding seem like it would be the same as adding padding to a block, and top and bottom padding don't seem to have an obvious difference. It's possible that padding isn't the correct name for this perhaps?

- feat: add list truncation behavior
Rather than not displaying the last item in a list, this displays the partial item, which is nicer.

One concern over the truncation behavior is whether this implementation might get in the way of future functionality that makes list selection better. E.g. it would be nice to be able to select between truncating the first item and truncating the last item, and even make this dependent on the direction of scroll (e.g. when scrolling down, the top of the list should be truncated, while the bottom should show a full item).

I'm also not 100% sure what to do with items that have a height greater than the list_area.

Lastly, I don't think the naming of the feature `truncate_last_item` is quite correct and would appreciate any better ones if there's something obvious in describing this. Perhaps `display_partial_items` or something similar?

Demo (and the reason for the two features):
[asciicast](https://asciinema.org/a/580726)


src/widgets/list.rs
```diff
@@ -211,26 +255,36 @@ impl<'a> StatefulWidget for List<'a> {
         {
             let (x, y) = match self.start_corner {
                 Corner::BottomLeft => {
-                    current_height += item.height() as u16;
+                    current_height += item.height() as u16 + self.padding as u16;
                     (list_area.left(), list_area.bottom() - current_height)
```
karthago1, line 259: :lady_beetle:  :lady_beetle:  :lady_beetle: 
`thread 'main' panicked at 'attempt to substruct with overflow', src/widgets/list.rs:259:40`

joshka: Thanks for catching this error. I resolved the problem in th latest push, and added some tests that help it out. The code is kinda ugly still, so I might refactor it as another commit on top of this. I'm still not 100% sure what the right name for this feature is (and whether it makes sense to have an enum or a boolean for it).

src/widgets/list.rs
```diff
@@ -142,31 +165,52 @@ impl<'a> List<'a> {
...
+                height = height.saturating_sub(self.items[start].height() + self.padding);
                 start += 1;
             }
         }
+
+        // here we do the same thing as above, but in the opposite direction
+        // to ensure that the start and end bounds are correct. We remove one
+        // item from the start, and add 0 or more items to the end to ensure
+        // that the height is less than or equal to the max height.
         while selected < start {
```
karthago1, line 199: this `while` can be replaced by a simple if at the beginning: here the complete code of `get_items_bounds`.

```
fn get_items_bounds(
    &self,
    selected: Option<usize>,
    offset: usize,
    max_height: usize,
) -> (usize, usize) {
    let mut start = offset.min(self.items.len().saturating_sub(1));
    if let Some(selected) = selected {
        start = start.min(selected);
    }
    let mut end = start;
    let mut height = 0;
    // calculate which items will be showing starting from offset and
    // filling only the items that fit in the available space
    for item in self.items.iter().skip(start) {
        // don't include the padding in the max height calculation
        // as we don't need to render the last item's padding
        if height + item.height() > max_height {
            break;
        }
        height += item.height() + self.padding;
        end += 1;
    }

    // if the selected item is after the range of items that will be
    // showing, then we need to adjust the start and end bounds to
    // reflect this. We add one item at the end, and remove 0 or more
    // items from the start to ensure that the height is less than or
    // equal to the max height.
    let selected = selected.unwrap_or(0).min(self.items.len() - 1);
    while selected >= end {
        height = height.saturating_add(self.items[end].height() + self.padding);
        end += 1;
        while height > max_height {
            height = height.saturating_sub(self.items[start].height() + self.padding);
            start += 1;
        }
    }

    if !self.truncate_last_item && height < max_height {
        // if the height is less than the max height, then we need to
        // adjust the end bound to include the last item.
        end += 1;
    }

    (start, end)
}
```

joshka:
Ha - just found https://github.com/tui-rs-revival/ratatui/pull/29/files which also adds padding. I'll take a look at that change tonight.

joshka:
Marked as Draft. Consider the following as an alternative implementation to this:
- https://github.com/tui-rs-revival/ratatui/pull/29

PR #156, opened 2023-04-29
test(widget): add unit tests for Paragraph

joshka:
Putting this in place to help ensure #149 doesn't break anything.

I wasn't 100% certain how the `zero_width_char_at_end_of_line` test was supposed to actually fail / not-fail as there's no assertion. Rewrote it to conform to the pattern of the other tests (but perhaps @orhun and @rhysd can add a bit more context if this is incorrect?

### DIFF
--- a/src/widgets/paragraph.rs
+++ b/src/widgets/paragraph.rs
@@ -203,12 +203,218 @@ impl<'a> Widget for Paragraph<'a> {
 #[cfg(test)]
 mod test {
     use super::*;
+    use crate::{
+        style::Color,
+        text::{Span, Spans},
+        widgets::Borders,
+    };
 
     #[test]
     fn zero_width_char_at_end_of_line() {
         let line = "foo\0";
-        let paragraph = Paragraph::new(line);
-        let mut buf = Buffer::with_lines(vec![line]);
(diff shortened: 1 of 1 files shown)


joshka:
Ping @TimerErTim for a review

joshka:
Force pushed formatting fixes

TimerErTim (commented):
The test cases generally look good, I will integrate them into my branch. I will extend them there when I can think of new/expanded ones.

src/widgets/paragraph.rs
```diff
@@ -202,13 +202,220 @@ impl<'a> Widget for Paragraph<'a> {
 
 #[cfg(test)]
 mod test {
+    use crate::{
+        style::Color,
+        text::{Span, Spans},
+        widgets::Borders,
+    };
+
```
orhun, line 210: nit: can we import `use super::*;` before this line to make the formatting consistent with other modules?

joshka: Done

joshka:
Rebased on #161 to use `assert_buffer_eq!` instead of `assert_eq`

orhun (approved):

PR #160, opened 2023-05-01
feat(test): expose test buffer

a-kenji:
Allow a way to expose the buffer of the `TestBackend`, to easier support different testing methodologies.

### DIFF
--- a/src/backend/test.rs
+++ b/src/backend/test.rs
@@ -3,7 +3,10 @@ use crate::{
     buffer::{Buffer, Cell},
     layout::Rect,
 };
-use std::{fmt::Write, io};
+use std::{
+    fmt::{Display, Write},
+    io,
+};
 use unicode_width::UnicodeWidthStr;
 
 /// A backend used for the integration tests.
@@ -102,6 +105,12 @@ impl TestBackend {
             .join("\n");
         debug_info.push_str(&nice_diff);
         panic!("{}", debug_info);
+    }
+}
+
+impl Display for TestBackend {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", buffer_view(&self.buffer))
     }
 }
 


a-kenji:
Superceded by #161,
assuming `to_test_string()` stays in the public interface.

src/backend/test.rs
```diff
@@ -66,6 +66,10 @@ impl TestBackend {
         self.height = height;
     }
 
+    pub fn print_buffer(&self) -> String {
```
joshka, line 69: `print` doesn't seem like the right verb for this (as it would imply sending to stdout) - perhaps `to_string()`? Do you have an idea on what the idiomatic rust name for something like this might be?

I'm concerned however that this function might be modified at some stage in the future (like perhaps better display of the multi-widget overwrites, or removal of the quotation marks / last newline. That could cause anything relying on this (which would generally be in tests) to break. Do you have a use case in mind for this?

I'm assuming that you probably need this for integration tests in your app (if you're using this for unit tests, perhaps take a quick look at #161 and see if that suits your use case at all). Guessing that you could be using this for documentation too?

a-kenji: I agree, I wasn't sure on the naming. 

I thought about adding a `Display` implementation for `Buffer`, but I intentionally wanted to have it exclusively in the `test` interface and didn't want to shadow `to_string()` itself. I like your solution in #161, but I will want to rely on `to_test_string()` in that interface.

> I'm concerned however that this function might be modified at some stage in the future (like perhaps better display of the multi-widget overwrites, or removal of the quotation marks / last newline. 

I am OK with that, I don't mind going over the tests in that case.

a-kenji:
Reopening, since It seems that this approach is more suited to my immediate needs, after looking closer at #161.

src/backend/test.rs
a-kenji: I opted implementing `Display` for `TestBuffer` in the end, and I am quite happy with that. That way the Buffer itself doesn't get a `to_string()` method, but because it stays in the `TestBuffer`, it also stays in the test context.

src/backend/test.rs
```diff
@@ -105,6 +108,12 @@ impl TestBackend {
     }
 }
 
+impl Display for TestBackend {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", buffer_view(&self.buffer))
+    }
+}
+
```
joshka, line 116: I think you're right. `impl Display` is the right choice.

Do you think we should move buffer_view within this method? It seems unnecessary other than for display purposes. Happy to approve either way.

joshka (approved):
Consider moving the code from `buffer_view()` into `fmt()`. But approved regardless as this seems useful.

src/backend/test.rs
a-kenji: > Do you think we should move buffer_view within this method? It seems unnecessary other than for display purposes. Happy to approve either way.

Currently, it is still used by `assert_buffer()`, but once #161 lands , I think the assertion macro can be used in its stead. 
And then moving the `buffer_view()` will make sense IMO.

joshka: Ah yes, you’re right. The use in the assert method precludes moving it. So nothing to do there.

sophacles (approved):
+1

orhun (approved):

PR #167, opened 2023-05-06
test(list): add characterization tests for list

joshka:
- also adds builder methods on list state to make it easy to construct
      a list state with selected and offset as a one-liner. Uses `with_` as
      the prefix for these methods as the selected method currently acts as
      a getter rather than a builder.
- cargo tarpaulin suggests only two lines are not covered (the two
      match patterns of the self.start_corner match 223 and 227).
      the body of these lines is covered, so this is probably 100% 

Contextually, this PR is about putting in place some tests for the list widget so that we can easily refactor some of the complexity of the methods to be easy to read and maintain.

Note: this sits on #161 which is not yet merged, so ignore the changes to buffer / sparkline for now.

### DIFF
--- a/src/widgets/list.rs
+++ b/src/widgets/list.rs
@@ -14,6 +14,16 @@ pub struct ListState {
 }
 
 impl ListState {
+    pub fn with_selected(mut self, selected: Option<usize>) -> Self {
+        self.selected = selected;
+        self
+    }
+
+    pub fn with_offset(mut self, offset: usize) -> Self {
+        self.offset = offset;
+        self
+    }
+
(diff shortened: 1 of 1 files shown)


sophacles:
Would you mind rebasing against main now that #161 is merged? Nothing in here is objectionable, just not sure what's different from main now.

joshka:
> Would you mind rebasing against main now that #161 is merged? Nothing in here is objectionable, just not sure what's different from main now.

Done

sophacles (approved):
Yay tests!

orhun (approved):
LGTM!

PR #32, opened 2023-02-12
feat(text): allow text to be masked when rendering

orhun:
> Upstream: [#648](https://github.com/fdehau/tui-rs/pull/648)

## Description
This PR will enable to have a text object which hides the actual content. This will be specifically useful to show sensative content like password in tui but being still able to access the value, but without rendering it

## Testing guidelines
Create a `Span` through `Span::masked` method, and such widget should be rendered as masked character as specified

## Checklist

* [X] I have read the [contributing guidelines](../CONTRIBUTING.md).
* [X] I have added relevant tests.
* [X] I have documented all new additions.

PR #168, opened 2023-05-07
feat(text)!: add `Masked` to display secure data

joshka:
Adds a new type Masked that can mask data with a mask character, and can be used anywhere we expect Cow<'a, str> or Text<'a>. E.g. Paragraph, ListItem, Table Cells etc.

BREAKING CHANGE:
Because Masked implements From for Text<'a>, code that binds Into<Text<'a>> without type annotations may no longer compile (e.g. `Paragraph::new("".as_ref())`)
To fix this, annotate or call to_string() / to_owned() / as_str()

This is an alternative to https://github.com/tui-rs-revival/ratatui/pull/32

[asciicast](https://asciinema.org/a/ftwC1T9gyvW6VA0WVRokF11I9)

### DIFF
--- a/examples/paragraph.rs
+++ b/examples/paragraph.rs
@@ -7,7 +7,7 @@ use ratatui::{
     backend::{Backend, CrosstermBackend},
     layout::{Alignment, Constraint, Direction, Layout},
     style::{Color, Modifier, Style},
-    text::{Span, Spans},
+    text::{Masked, Span, Spans},
     widgets::{Block, Borders, Paragraph, Wrap},
     Frame, Terminal,
 };
@@ -133,6 +133,13 @@ fn ui<B: Backend>(f: &mut Frame<B>, app: &App) {
                 .fg(Color::Green)
                 .add_modifier(Modifier::ITALIC),
         )),
+        Spans::from(vec![
+            Span::raw("Masked text: "),
+            Span::styled(
+                Masked::new("password", '*'),
+                Style::default().fg(Color::Red),
+            ),
+        ]),
     ];
 
     let create_block = |title| {
--- a/examples/user_input.rs
+++ b/examples/user_input.rs
@@ -155,7 +155,7 @@ fn ui<B: Backend>(f: &mut Frame<B>, app: &App) {
     let help_message = Paragraph::new(text);
     f.render_widget(help_message, chunks[0]);
 
-    let input = Paragraph::new(app.input.as_ref())
+    let input = Paragraph::new(app.input.as_str())
         .style(match app.input_mode {
             InputMode::Normal => Style::default(),
             InputMode::Editing => Style::default().fg(Color::Yellow),
--- a/src/text.rs
+++ b/src/text.rs
@@ -48,6 +48,7 @@
 //! ```
 use crate::style::Style;
 use std::borrow::Cow;
+use std::fmt::{self, Debug, Display};
 use unicode_segmentation::UnicodeSegmentation;
 use unicode_width::UnicodeWidthStr;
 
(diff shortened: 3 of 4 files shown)


mindoodoo (approved):
Awesome stuff, this looks great. Breaking change is relatively small and isn't an issue IMHO. I'd also add I much prefer this as opposed to #32, although it has the one of advantage of not having a breaking change.

joshka:
> Awesome stuff, this looks great. Breaking change is relatively small and isn't an issue IMHO. I'd also add I much prefer this as opposed to #32, although it has the one of advantage of not having a breaking change.

Yeah, I'm not sure I'd have noticed the breaking change if it wasn't for the two places where we have as_ref() calls. The impact is only really felt when there is some large string that's being copied (and the largest string that can be reasonably rendered on a screen isn't likely to ever be a perf issue, so meh). For this change I wouldn't bother with a major version bump ever. I only called it out as it's useful to have notes on what to fix when something breaks (or when assumptions about impact are wrong).

mindoodoo:
> For this change I wouldn't bother with a major version bump ever. I only called it out as it's useful to have notes on what to fix when something breaks (or when assumptions about impact are wrong).

I think the next release will most likely have breaking changes anyways so might as well slip it in with it.

mindoodoo:
@joshka Regarding the information about the breaking change, did you find a way to be able to keep the commit footer and body on github ?

joshka:
> @joshka Regarding the information about the breaking change, did you find a way to be able to keep the commit footer and body on github ?

I prefer to merge / push changes via the cli rather than github. Rebase on main, test, push. Done.

Ideally all changes would be rebased on the HEAD of main (or close to it) prior to getting approval anyway. My general argument (rant? ;)) is that merging / rebasing is actually writing software and it's often a source of introduced weird bugs when developers choose to treat it as not an important step of the process.

sophacles (approved):
I like it. Nice feature.

Idea for expansion of it in the future:
Have an option to mask only a subset of the text, for example:

````
Choose how you'd like to receive your authrization code:
 Email: a*****@gmail.com
 Phone: ***-***-1234
 ```
 Etc.

Maybe in the form of 
```rust
// constructor or setter on Masked
// args to masker: input txt, mask_char
fn masker(self, masker: Fn(&str, &str) -> &str) -> Self {...}
```

joshka:
> Idea for expansion of it in the future:
> Have an option to mask only a subset of the text, for example:

The main benefit of having this as part of the library is that it is useful in its simple form, but I think once you go beyond a certain complexity on this, it's probably more the user's responsibility to be the one that handles their own input / output, and this isn't all that difficult. An example of keeping that idea super simple:

```
struct MaskedCreditCard<'a> {
    card_number: &'a str,
}

impl<'a> MaskedCreditCard<'a> {
    pub fn new(card_number: &'a str) -> Self {
        // in a real app, should never have anything more than the last 4 digit available
        // but for the sake of an example
        MaskedCreditCard { card_number }
    }
}

impl Default for MaskedCreditCard<'_> {
    fn default() -> Self {
        MaskedCreditCard::new("0000-0000-0000-0000")
    }
}

impl<'a> From<MaskedCreditCard<'a>> for Cow<'a, str> {
    fn from(value: MaskedCreditCard) -> Self {
        ("****-****-****-".to_string() + value.card_number.rsplit_once('-').unwrap().1).into()
    }
}
```

joshka:
Rebased on main

orhun (approved):

sophacles:
Sure, not looking to make it complex, just was thinking something like this:

```
fn masked_credit_card(ccn: &str) -> Masked<'a> {
    Masked::new(ccn).masker(|txt, char| {
        let (to_mask, plain) = start.rsplit_once('-').unwrap();
        let mut s = to_mask.replace(['0', '1', '2', '3', '4', '5', '6', '7', '8', '9'], "*");
        s.push_str("-");
        s.push_str(plain);
    })
}
```

PR #169, opened 2023-05-07
chore(git): ignore tarpaulin (code coverage) report

joshka:


mindoodoo (approved):

orhun:
Not sure if we need this. `cargo-tarpaulin` is not integrated into this repository so what's the point of adding it? Some people might use [`cargo-llvm-cov`](https://github.com/taiki-e/cargo-llvm-cov) for code coverage and merging this might mean that we want to cover similar cases too.

Now that I looked at `.gitignore`, not sure why we have `.gdb_history` and `.idea` there as well. Those are very specific to an user's development environment and I'm not a proponent of adding anything to `.gitignore` which is not directly related to the project. I would remove them too.

Requesting comment from @joshka @mindoodoo @sophacles and anyone else watching the repo 🐻

joshka:
Mostly I was running it locally to check some coverage stats on tests, and was annoyed that the report doesn't splat somewhere that's ignored. I can add it to my local gitignore instead - this isn't really a big deal.

orhun:
I think you can simply use `.git/info/exclude` for that purpose 🐻

joshka:
> I think you can simply use `.git/info/exclude` for that purpose 🐻

Sounds good - I'll do that  :)

PR #171, opened 2023-05-09
build: bump MSRV to 1.65.0

joshka:
The latest version of the time crate requires Rust 1.65.0

```
cargo +1.64.0-x86_64-apple-darwin test --no-default-features \
  --features serde,crossterm,all-widgets --lib --tests --examples
error: package `time v0.3.21` cannot be built because it requires rustc
1.65.0 or newer, while the currently active rustc version is 1.64.0
```

The calendar widget added a dependency on time v0.3.11 (though a quick check suggests that 0.3.9 compiles, I'm unsure if the later bug fixes are notable for ratatui. Regardless, the dependency requirement is satisfied by any version later than 0.3.11. Later versions require rust 1.65.0, so our MSRV is 1.65.0.

I don't think there's an obvious way to only specify a MSRV for a specific set of features (i.e. 1.59 if you're not using termwiz/calendar), though perhaps we leave this as 1.59, and create a specific 1.59 test in the ci that doesn't test calendar and termwiz. I'm not certain I know why we'd do that (ping @rhysd for input on what the reason for supporting 1.59 was in https://github.com/tui-rs-revival/ratatui/pull/85

I'm a rust noob, so my experience with the impact of choosing an MSRV is limited. I've seen a few libraries that state a policy of supporting stable minus 2 version, which seems reasonable on the face of it. If we took that approach, we'd go with 1.67 which was released Jan (26 2023). But we only "need" to go to 1.65 for now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ you may rely on the previously cited libraries to achieve such features.
 
 ## Rust version requirements
 
-Since version 0.17.0, `ratatui` requires **rustc version 1.59.0 or greater**.
+Since version 0.21.0, `ratatui` requires **rustc version 1.65.0 or greater**.
 
 # Documentation
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        rust: ["1.59.0", "stable"]
+        rust: ["1.65.0", "stable"]
         include:
           - os: ubuntu-latest
             triple: x86_64-unknown-linux-musl
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ license = "MIT"
 exclude = ["assets/*", ".github", "Makefile.toml", "CONTRIBUTING.md", "*.log", "tags"]
 autoexamples = true
 edition = "2021"
-rust-version = "1.59.0"
+rust-version = "1.65.0"
 
 [badges]
 
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -16,7 +16,7 @@ pub use self::crossterm::CrosstermBackend;
 mod test;
 pub use self::test::TestBackend;
 
-#[derive(Debug, Clone, Copy, PartialEq)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum ClearType {
     All,
     AfterCursor,
--- a/src/terminal.rs
+++ b/src/terminal.rs
@@ -6,14 +6,14 @@ use crate::{
 };
 use std::io;
 
-#[derive(Debug, Clone, PartialEq)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub enum Viewport {
     Fullscreen,
     Inline(u16),
     Fixed(Rect),
 }
 
-#[derive(Debug, Clone, PartialEq)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 /// Options to pass to [`Terminal::with_options`]
 pub struct TerminalOptions {
     /// Viewport used to draw to the terminal
--- a/src/widgets/block.rs
+++ b/src/widgets/block.rs
@@ -313,17 +313,8 @@ impl<'a> Widget for Block<'a> {
 
         // Title
         if let Some(title) = self.title {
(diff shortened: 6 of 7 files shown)


sophacles:
Re: failures - I didn't know numeric types had `From<bool>` now!

Re msrv and dependencies and how they interact - that's still open and ongoing discussion: https://github.com/rust-lang/cargo/issues/9930

I think bumping this is fine, probably a good idea to make clippy happy at the same time.

joshka:
Pushed clippy fixes (and added the following to the commit):
--
Also fixes several clippy warnings added in 1.63/1.65. Although these
    have been since moved to nursery / pedantic, it doesn't hurt to fix
    these issues as part of this change:

    - https://rust-lang.github.io/rust-clippy/master/index.html#derive_partial_eq_without_eq (nursery)
    - https://rust-lang.github.io/rust-clippy/master/index.html#bool_to_int_with_if (pedantic)

sophacles (approved):

README.md
```diff
@@ -51,7 +51,7 @@ you may rely on the previously cited libraries to achieve such features.
 
 ## Rust version requirements
 
-Since version 0.17.0, `ratatui` requires **rustc version 1.59.0 or greater**.
+Since version 0.21.0, `ratatui` requires **rustc version 1.65.0 or greater**.
```
orhun, line 54: Are we sure that the next version will be `0.21.0`?

orhun: Regardless, we can fix it afterwards if we settle on something else. It was a just question to get your input 🐻

orhun:
Interestingly enough, when I run [`cargo-msrv`](https://github.com/foresterre/cargo-msrv) it says the MSRV is `1.60.0`.

However if I run `cargo msrv -- cargo check --features all-widgets` then the MSRV is `1.63.0`

cap

For some reason, I get different results from this PR. Is there any other way to validate?

README.md
sophacles: Here's what the cargo book has to say about it: https://doc.rust-lang.org/cargo/reference/semver.html#possibly-breaking-changing-the-minimum-version-of-rust-required

(tl;dr - minor version or major version bump is suggested, not just patch)

joshka: I'd highly advocate for this to go to 0.21 regardless as there are a couple of small breaking changes in flight (e.g. the PR for MaskedText, which breaks anything where someone calls methods that take `Into<Text>' by with a `something.as_ref()` instead of a `something.as_str()`)

joshka:
>For some reason, I get different results from this PR. Is there any other way to validate?

Yep, run `cargo update` - the time crate is at 0.3.21 which [requires rust 1.65.0](https://github.com/time-rs/time/blob/main/time/Cargo.toml#L6).

```
❯ cargo msrv -- cargo check --all-features
Fetching index
Determining the Minimum Supported Rust Version (MSRV) for toolchain x86_64-apple-darwin
Using check command cargo check --all-features

Check for toolchain '1.63.0-x86_64-apple-darwin' failed with:
┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│ error: package `time-core v0.1.1` cannot be built because it requires rustc 1.65.0 or newer, while the currently active rustc version is    │
│ 1.63.0                                                                                                                                      │
└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
Check for toolchain '1.66.1-x86_64-apple-darwin' succeeded

Check for toolchain '1.64.0-x86_64-apple-darwin' failed with:
┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│ error: package `time v0.3.21` cannot be built because it requires rustc 1.65.0 or newer, while the currently active rustc version is 1.64.0 │
└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
Check for toolchain '1.65.0-x86_64-apple-darwin' succeeded
   Finished The MSRV is: 1.65.0   ████████████████████████████████████████████████████████████████████████████████████████████████████ 00:00:06
```

orhun:
> Yep, run `cargo update` - the time crate is at 0.3.21 which [requires rust 1.65.0](https://github.com/time-rs/time/blob/main/time/Cargo.toml?rgh-link-date=2023-05-11T22%3A39%3A51Z#L6).

Good point! I got `1.65.0` as MSRV now.

orhun (approved):

README.md
orhun: Sounds reasonable.

PR #172, opened 2023-05-09
Renamed `text::Spans` to `text::Line` for clarity

Eyesonjune18:


src/text.rs
```diff
@@ -179,6 +179,43 @@ impl<'a> Span<'a> {
             })
             .filter(|s| s.symbol != "\n")
     }
+
+    /// Patches the style an existing Span, adding modifiers from the given style.
```
joshka, line 183: The patch style changes to patch the style seem like they're unrelated to this change. Is this just based on a previous change that's not yet merged, or included by mistake?

src/text.rs
```diff
@@ -275,7 +363,7 @@ impl<'a> From<Spans<'a>> for String {
 /// ```
 #[derive(Debug, Clone, PartialEq, Default, Eq)]
 pub struct Text<'a> {
-    pub lines: Vec<Spans<'a>>,
+    pub lines: Vec<Line<'a>>,
```
joshka, line 366: This line is the main line that makes this PR obviously the right thing to do.
We already refer to a Vec of Spans as `lines` :)

src/text.rs
```diff
@@ -195,53 +232,104 @@ impl<'a> From<&'a str> for Span<'a> {
 
 /// A string composed of clusters of graphemes, each with their own style.
 #[derive(Debug, Clone, PartialEq, Default, Eq)]
-pub struct Spans<'a>(pub Vec<Span<'a>>);
+pub struct Line<'a>(pub Vec<Span<'a>>);
```
joshka, line 236: Do you think it's worth leaving a type alias around so that consumers are not broken. E.g.:
```suggestion

#[deprecated(since = "0.21.0", note = "`Spans` is renamed to `Lines` and will be removed in some future release"]
pub type Spans = Lines;

```

joshka (changes requested):
With the two suggestions to maintain backwards compatibility, this becomes non-breaking and hence probably a good idea. I think the name `Line` generally makes more sense than Spans also. That's my personal view, but I'd probably want a few more people to weigh in on the need for the change.

I'm going to add the notes I mentioned in #163 as a conversation rather than a review.

joshka:
From #163 (discussion on breaking changes):

1. Do we have a good reason for the breaking change?
  `Line` is a better description of what `Spans` purpose is
3. Have we considered non-breaking changes to fix the issue?
  Some suggestions in my review for approaching this (a `#[deprecated]` type alias and method.
5. Have we documented what the change is in detail before making it?
  This could be a little better in the commit message body
7. Are there other changes around the same area that would reduce the fatigue of responding to updates?
  There are a few PRs currently outstanding that might be worth shipping at the same time.
9. Have we documented how to fix the change in downstream code?
  This should just be a fairly easy rename:
  ```
  find . -name '*.rs' -exec sed -i '' 's/Spans/Line/g' {} \;
  ```
  and then check for any places that the rename conflicts (e.g. when using Canvas
11. Do we know how many dependents / migrations use the code being changed?
  Ballpark 3.5K files: https://github.com/search?q=%22use+tui%22+%22Spans%22+language%3Arust&type=code
12. Do we have a plan for communicating the change (and a timeline)?
  No
13. Do we need user feedback on this change?
  Probably
14. Should we wait a little for other things to settle before making this change?
  Maybe
15. What else?
  I don't know - what do you think?

joshka:
P.S. can you please squash your changes into either 1 or two commits and force push an update?

src/text.rs
Eyesonjune18: Well, it looks like I accidentally pushed stuff from a PR that has already been merged. Is there any way for me to get rid of these commits? I'm not super familiar with using forked repos.

src/text.rs
Eyesonjune18: No, I think we should have them use `Line` both for consistency in examples/codebases and also readability of their own code. It's not a particularly difficult refactor. Though I understand where you're coming from.

src/text.rs
joshka: in your local, assuming that you have your repo setup as origin and this one as upstream, you can do something like:
```text
git switch main
git pull upstream main
git switch text-style
git rebase main
git push origin text-style --force
```
but handling the rebase with 13 commits is probably going to be more painful than doing it with just 1 or 2.
I'd suggest squashing the commits before doing all of that
```
git rebase --interactive main
```
(choose squash for everything except your first commit)

You may end up having to either skip commits that are already merged (if you don't have the exact same version of that commit) `git rebase --skip` or update your commit to address any incompatible changes. Depending on the number of conflicts, it could be easier to re-do the renames, so it's worth splitting the changes that add new patch style into a separate commit to make that easy.

src/text.rs
joshka: This is for current consumers and future migration, not for internal use.
By breaking this here, you make it impossible for a consumer of this library to migrate from tui or to update from 0.20.1 to 0.21.0. There are around 3500 files from a quick github search where that causes a problem. It's not difficult to make each change, but the impact is large. Why foist all that work onto others when it can be avoided or deferred? By adding the type alias, you have the best of both worlds - communication that something will break, instruction on how to fix it, backwards compatible change,  and it's all delivered via the compiler.

src/text.rs
Eyesonjune18: Would you be fine with just migrating your comments to a different PR and closing this one?

src/text.rs
Eyesonjune18: I didn't read this carefully. I think that marking it as deprecated is a good solution.

src/text.rs
Eyesonjune18: Actually I'll just try what you said first, and see if it works.

src/text.rs
sophacles: Honestly, `Line` is a better name, until you render a paragraph with wrapping, then a `Line` can be several lines long.... (just an amusing observation that naming things is hard)

Eyesonjune18:
I am remaking this PR: https://github.com/tui-rs-revival/ratatui/pull/175

src/text.rs
joshka: I know this is probably a super pedantic take, but I'd still call that a line. It's a wrapped line, but still a line.

PR #175, opened 2023-05-11
refactor(text): rename `Spans` to `Line`

Eyesonjune18:
Because `Spans` conflicts with the plural `Span`s, and because the type represents a line of text, I've renamed it to `Line`. I will also be adding a deprecated type alias to avoid breaking existing codebases (haven't committed it yet).

### DIFF
--- a/examples/demo/ui.rs
+++ b/examples/demo/ui.rs
@@ -4,7 +4,7 @@ use ratatui::{
     layout::{Constraint, Direction, Layout, Rect},
     style::{Color, Modifier, Style},
     symbols,
-    text::{Span, Line},
+    text::{Line, Span},
     widgets::canvas::{Canvas, Circle, Line as ShapeLine, Map, MapResolution, Rectangle},
     widgets::{
         Axis, BarChart, Block, Borders, Cell, Chart, Dataset, Gauge, LineGauge, List, ListItem,
--- a/examples/demo/ui.rs
+++ b/examples/demo/ui.rs
@@ -4,8 +4,8 @@ use ratatui::{
     layout::{Constraint, Direction, Layout, Rect},
     style::{Color, Modifier, Style},
     symbols,
-    text::{Span, Spans},
-    widgets::canvas::{Canvas, Circle, Line, Map, MapResolution, Rectangle},
+    text::{Span, Line},
+    widgets::canvas::{Canvas, Circle, Line as ShapeLine, Map, MapResolution, Rectangle},
     widgets::{
         Axis, BarChart, Block, Borders, Cell, Chart, Dataset, Gauge, LineGauge, List, ListItem,
         Paragraph, Row, Sparkline, Table, Tabs, Wrap,
@@ -21,7 +21,7 @@ pub fn draw<B: Backend>(f: &mut Frame<B>, app: &mut App) {
         .tabs
         .titles
         .iter()
-        .map(|t| Spans::from(Span::styled(*t, Style::default().fg(Color::Green))))
+        .map(|t| Line::from(Span::styled(*t, Style::default().fg(Color::Green))))
         .collect();
     let tabs = Tabs::new(titles)
         .block(Block::default().borders(Borders::ALL).title(app.title))
@@ -137,7 +137,7 @@ where
                 .tasks
                 .items
                 .iter()
-                .map(|i| ListItem::new(vec![Spans::from(Span::raw(*i))]))
+                .map(|i| ListItem::new(vec![Line::from(Span::raw(*i))]))
                 .collect();
             let tasks = List::new(tasks)
                 .block(Block::default().borders(Borders::ALL).title("List"))
@@ -161,7 +161,7 @@ where
                         "WARNING" => warning_style,
                         _ => info_style,
                     };
-                    let content = vec![Spans::from(vec![
+                    let content = vec![Line::from(vec![
                         Span::styled(format!("{:<9}", level), s),
                         Span::raw(evt),
                     ])];
@@ -261,9 +261,9 @@ where
     B: Backend,
 {
     let text = vec![
-        Spans::from("This is a paragraph with several lines. You can change style your text the way you want"),
-        Spans::from(""),
-        Spans::from(vec![
+        Line::from("This is a paragraph with several lines. You can change style your text the way you want"),
+        Line::from(""),
+        Line::from(vec![
             Span::from("For example: "),
             Span::styled("under", Style::default().fg(Color::Red)),
             Span::raw(" "),
@@ -272,7 +272,7 @@ where
             Span::styled("rainbow", Style::default().fg(Color::Blue)),
             Span::raw("."),
         ]),
-        Spans::from(vec![
+        Line::from(vec![
             Span::raw("Oh and if you didn't "),
             Span::styled("notice", Style::default().add_modifier(Modifier::ITALIC)),
             Span::raw(" you can "),
@@ -283,7 +283,7 @@ where
             Span::styled("text", Style::default().add_modifier(Modifier::UNDERLINED)),
             Span::raw(".")
         ]),
-        Spans::from(
+        Line::from(
             "One more thing is that it should display unicode characters: 10€"
         ),
     ];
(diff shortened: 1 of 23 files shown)


joshka:
This PR is a remake of #172. Some relevant comments from there (note it looks like the type alias / deprecated consideration for backward compatibility was made part of this change and the extra patch methods were removed), and that this is probably not a breaking change anymore:

From #163 (discussion on breaking changes):

1. Do we have a good reason for the breaking change?
  `Line` is a better description of what `Spans` purpose is
3. Have we considered non-breaking changes to fix the issue?
  Some suggestions in my review for approaching this (a `#[deprecated]` type alias and method.
5. Have we documented what the change is in detail before making it?
  This could be a little better in the commit message body
7. Are there other changes around the same area that would reduce the fatigue of responding to updates?
  There are a few PRs currently outstanding that might be worth shipping at the same time.
9. Have we documented how to fix the change in downstream code?
  This should just be a fairly easy rename: `find . -name '*.rs' -exec sed -i '' 's/Spans/Line/g' {} \;` and then check for any places that the rename conflicts (e.g. when using Canvas
11. Do we know how many dependents / migrations use the code being changed?
  Ballpark 3.5K files: https://github.com/search?q=%22use+tui%22+%22Spans%22+language%3Arust&type=code
12. Do we have a plan for communicating the change (and a timeline)?
  No
13. Do we need user feedback on this change?
  Probably
14. Should we wait a little for other things to settle before making this change?
  Maybe
15. What else?
  I don't know - what do you think?

joshka:
Can you please rerun `cargo fmt` so the checks succeed?

I found that it's worth configuring my editor (VSCode) to automatically run rustfmt on save to avoid this problem, YMMV.

Eyesonjune18:
> Can you please rerun `cargo fmt` so the checks succeed?
> 
> I found that it's worth configuring my editor (VSCode) to automatically run rustfmt on save to avoid this problem, YMMV.

Fixed.

joshka (approved):
LGTM - I'm generally for this going ahead as a change as it's non-breaking.

Eyesonjune18:
Will there be any way to non-breakingly deprecate `Line.0`? I don't know if you can do that with fields.

joshka:
> Will there be any way to non-breakingly deprecate `Line.0`? I don't know if you can do that with fields. It would be replaced with `Line.spans`.

Probably not if we went direct to this implementation. We could possibly leave `Spans` in the project, and instead modify every place that currently takes a `Spans` to instead take an `Into<Line>` and implement `From<Spans>` on `Line`. It could be worth attempting that to see if it works as an alternative to this change. It seems like that would have similar properties of letting us deprecate functionality in a way that notifies users that things will break soon with a compiler message.

I like that you're considering other changes that might be worth making at the same time on this. :)

Eyesonjune18:
> I like that you're considering other changes that might be worth making at the same time on this. :)

Do you think this should be done in the same PR?

sayanarijit:
Can we make Line.0 private and implement Into and From for Span list? Or maybe a public method spans()?

And keep Spans, not as a deprecated struct, but undocumented, maybe in a separate module spans.rs?

This way, as a lazy user, if I don't want to move to Line, I won't feel pressured. And Line will have a nicer, more flexible api.

Eyesonjune18:
> Can we make Line.0 private and implement Into and From for Span list? Or maybe a public method spans()?
> 
> And keep Spans, not as a deprecated struct, but undocumented, maybe in a separate module spans.rs?
> 
> This way, as a lazy user, if I don't want to move to Line, I won't feel pressured. And Line will have a nicer, more flexible api.

If you can't be bothered to find-and-replace `.0` with `.spans` in your codebase, just don't update the crate. I'm fine with temporarily leaving `Spans` in the crate as a deprecated feature but I think we should then remove it on the next minor version release.

Eyesonjune18:
> Let's actually discuss this whole compatibility thing in a meeting. That I believe is the best to go about this.

When are the meetings held?

sayanarijit:
There was some discussion on having meetings on a scheduled time, but I was inactive for some time, so I need to check with others. Hope we're not in a hurry.

Eyesonjune18:
> There was some discussion on having meetings on a scheduled time, but I was inactive for some time, so I need to check with others. Hope we're not in a hurry.

The main issue for me is that I'm needing new functionality to be developed for my own project (a shell) because as it stands it's going to be basically impossible to continue with the lack of features. I have a group of people working on it and I really don't want them to feel like the project has been abandoned.

joshka:
> The main issue for me is that I'm needing new functionality to be developed for my own project (a shell) because as it stands it's going to be basically impossible to continue with the lack of features. I have a group of people working on it and I really don't want them to feel like the project has been abandoned.

Having specific needs to implement a feature is a really good counterbalance to the idea that breaking shouldn't happen. I'd definitely encourage you to share the blockers (because others might have them too, and we probably will mostly agree that breaking things to fix this is a good approach)

Eyesonjune18:
> > The main issue for me is that I'm needing new functionality to be developed for my own project (a shell) because as it stands it's going to be basically impossible to continue with the lack of features. I have a group of people working on it and I really don't want them to feel like the project has been abandoned.
> 
> Having specific needs to implement a feature is a really good counterbalance to the idea that breaking shouldn't happen. I'd definitely encourage you to share the blockers (because others might have them too, and we probably will mostly agree that breaking things to fix this is a good approach)

Some particular pain points for me have been text wrapping (requires breaking), lack of documentation (does not require breaking), widget reusability/odd architecture requirement for generating frames (requires breaking), and cursor movement tracking (probably does not require breaking).

sayanarijit:
> because others might have them too

Agree, and we actually need to assess whether this is truly the case. Otherwise I'm against breaking >300 crates to accomodate the need of one single crate.

joshka:
I'd say that as this currently stands it can safely be merged, with the caveat that delaying merging this creates an opportunity to decide on whether to change to fields rather than a tuple.

As I don't think this is a fix that specifically blocks anyone from using the feature, we're not on a time crunch for shipping this, so let's experiment with whether that idea works and go from there.

TL;DR I've approved this PR in the current form, but my preference is to not merge it yet.

joshka:
Actually, scratch that. I forgot that https://github.com/tui-rs-revival/ratatui/pull/149 makes the change to Spans already in order to add single line alignment. Perhaps @TimerErTim could weigh in on whether that change could be made backwards compatible (using the approach of leaving the current `Spans` as is and changing places that accept spans to accept e.g. `Into<Line>` (https://github.com/tui-rs-revival/ratatui/pull/175#issuecomment-1544915625)

TimerErTim:
At first glance I see no reason why such approach would not be possible. 

On the contrary, we could even make #149 non-breaking by keeping the current `Spans` tuple struct as is and returning the new `Line` struct on `Spans::alignment`.

Eyesonjune18:
> Agree, and we actually need to assess whether this is truly the case. Otherwise I'm against breaking >300 crates to accomodate the need of one single crate.

This is 300 crates that *could* use our crate, not that *do*, to my knowledge. Those aren't even remotely similar metrics in my opinion.

joshka:
> This is 300 crates that _could_ use our crate, not that _do_, to my knowledge. Those aren't even remotely similar metrics in my opinion. People generally expect to need to do some modifications to their existing codebase when migrating dependencies.

Maybe let's keep that concern over on #163

Eyesonjune18:
Can I get a second reviewer on this?

TimerErTim:
Should we take the approach of leaving `Spans` as is and refactoring for ´impl Into<Line>`?

If yes, that change would enable a smooth transition between working and breaking by marking `Spans` as deprecated. I would then (if there is no blocker) merge the PR quickly so other PRs can build upon the change/addition and make use of the new `Line` struct.

mindoodoo:
Yeah either that or a `.spans()` method as @sayanarijit mentionned earlier, but one of the two would be good I think.

Other than that, lgtm @Eyesonjune18

Eyesonjune18:
> Yeah either that or a `.spans()` method as @sayanarijit mentionned earlier, but one of the two would be good I think.
> 
> Other than that, lgtm @Eyesonjune18

I'm confused as to what you mean by those two suggestions. Can you give me an example?

joshka:
Closing as #178 is merged - thank you for the robust conversation and initial implementation on this one @Eyesonjune18

PR #177, opened 2023-05-16
feat(widget): add `offset()` to `ListState`

LeoRiether:
This PR implements a getter method `offset` for `ListState`, similarly to `TableState` (#10)


LeoRiether:
Oh wait, this is a duplicate of #12... Forgot to check that, sorry

PR #178, opened 2023-05-17
refactor(text): replace Spans with Line

joshka:
`Line` is a significantly better name over `Spans` as the plural causes
confusion and the type really is a representation of a line of text made
up of spans.

This is a backwards compatible version of the approach from
https://github.com/tui-rs-revival/ratatui/pull/175

There is a significant amount of code that uses the Spans type and
methods, so instead of just renaming it, we add a new type and replace
parameters that accepts a `Spans` with a parameter that accepts
`Into<Line>`.

Note that the examples have been intentionally left using `Spans` in
this commit to demonstrate the compiler warnings that will be emitted in
existing code.

Implementation notes:
- moves the Spans code to text::spans and publicly reexports on the text
module. This makes the test in that module only relevant to the Spans
type.
- adds a line module with a copy of the code and tests from Spans with a
single addition: `impl<'a> From<Spans<'a>> for Line<'a>`
- adds tests for `Spans` (created and checked before refactoring)
- adds the same tests for `Line`
- updates all widget methods that accept and store Spans to instead
store `Line` and accept `Into<Line>`

---
There is a second related commit on this PR to move text::Masked -> text::masked::Masked, with no changes.

---
Tagging @Eyesonjune18 and @TimerErTim for their input
This will likely require changes to be made on the changes in #149 (hopefully, just in `Text` and `Line`?)

I also updated the examples in a third commit. I wanted to make sure that backwards compatibility was maintained in a way that made sense first though so I intentionally kept the two commits separate. E.g.:

```
 1  warning: use of deprecated struct `ratatui::text::Spans`: Use `ratatui::text::Line` instead
   --> examples/panic.rs:29:20
    |
 29 | use ratatui::text::Spans;
    |                    ^^^^^
    |
    = note: `#[warn(deprecated)]` on by default
```
I think the only breaking change on this is any code that specifically calls `Text::into_iter()` and then stores that in a variable or field of type IntoIter<Spans<'a>> rather than just letting type inference do its thing. This seems unlikely enough that it's probably not even worth noting.


Eyesonjune18 (approved):
Thanks for the PR rework, LGTM.

sayanarijit (approved):

sayanarijit:
Lookg good 👍

TimerErTim (approved):
LGTM
Great work and great idea to begin with!

orhun (approved):
LGTM 🦀

Issue #143, opened 2023-04-20
Support conversion from `String` to `Color`


orhun:
## Problem

In my projects I sometimes need a conversion like this:

```
"black" => TuiColor::Black,
"red" => TuiColor::Red,
"green" => TuiColor::Green,
```

## Solution

Implement [`FromStr`](https://doc.rust-lang.org/std/str/trait.FromStr.html) for [`ratatui::style::Color`](https://docs.rs/ratatui/latest/ratatui/style/enum.Color.html).

An example can be found [here](https://github.com/orhun/systeroid/blob/7ad12c455f93988c769330a8caab019a88a061b9/systeroid-tui/src/style.rs#L65-L94).

In the example, I used [`rgb`](https://github.com/kornelski/rust-rgb) crate for converting HEX values to `Color` as well. Since it introduces a new dependency, I'm not sure if we should do that as well.

## Alternatives
None.

## Additional context
I also did something like this in the past: https://github.com/orhun/kmon/blob/66f27006fed614c048469b2511b600d07c8912d6/src/style.rs#L110-L152

sophacles:
This might be a "good first issue" too (unless you're planning on knocking it out yourself)

orhun:
Oh yeah, I was waiting for some comments, but sure. Anyone should be free to pick it up 🐻

pythops:
I would like to give it a try :)

orhun:
Go for it!

pythops:
@orhun 
> In the example, I used [rgb](https://github.com/kornelski/rust-rgb) crate for converting HEX values to Color as well. Since it introduces a new dependency, I'm not sure if we should do that as well.

what would be the ideal solution for you ?

orhun:
New dep ain't cool. I'm not sure how `TrueColor` conversion would work without it though. I haven't thought about it in depth.

But if you ask me I would blindly go with that solution. I simply created this issue to get more comments about it.

sayanarijit:
I wouldn't use strings if I have enums. Enums are more maintainable. If someone wants easy string to enum conversation, that'd be a separate feature flag, with someone dedicated to maintain it.

orhun:
> I wouldn't use strings if I have enums. Enums are more maintainable.

[...]

PR #180, opened 2023-05-18
feat(color): add `FromStr` implementation for `Color`

Mehrbod2002:
fixes: #143

### DIFF
--- a/src/style.rs
+++ b/src/style.rs
@@ -247,73 +247,6 @@ impl Style {
     }
 }
 
-#[cfg(test)]
-mod tests {
-    use super::*;
-
-    fn styles() -> Vec<Style> {
-        vec![
-            Style::default(),
-            Style::default().fg(Color::Yellow),
-            Style::default().bg(Color::Yellow),
-            Style::default().add_modifier(Modifier::BOLD),
-            Style::default().remove_modifier(Modifier::BOLD),
-            Style::default().add_modifier(Modifier::ITALIC),
-            Style::default().remove_modifier(Modifier::ITALIC),
-            Style::default().add_modifier(Modifier::ITALIC | Modifier::BOLD),
-            Style::default().remove_modifier(Modifier::ITALIC | Modifier::BOLD),
-        ]
-    }
-
-    #[test]
-    fn combined_patch_gives_same_result_as_individual_patch() {
-        let styles = styles();
-        for &a in &styles {
-            for &b in &styles {
-                for &c in &styles {
-                    for &d in &styles {
-                        let combined = a.patch(b.patch(c.patch(d)));
-
-                        assert_eq!(
-                            Style::default().patch(a).patch(b).patch(c).patch(d),
-                            Style::default().patch(combined)
-                        );
-                    }
-                }
-            }
-        }
-    }
-
-    #[test]
-    fn combine_individual_modifiers() {
-        use crate::{buffer::Buffer, layout::Rect};
-
-        let mods = vec![
-            Modifier::BOLD,
-            Modifier::DIM,
-            Modifier::ITALIC,
-            Modifier::UNDERLINED,
-            Modifier::SLOW_BLINK,
-            Modifier::RAPID_BLINK,
-            Modifier::REVERSED,
-            Modifier::HIDDEN,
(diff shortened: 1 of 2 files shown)


joshka (changes requested):
There's a note in the issue about whether we should / shouldn't bring in a new dependency for this (it seems like a fairly small amount of code to avoid doing so). I could go either way, but lean towards not bringing it in. Do we gain anything by putting the implementation off onto colorsys?

src/style.rs
```diff
@@ -312,3 +314,49 @@ mod tests {
...
+            "lightred" | "light red" => Self::LightRed,
+            "lightgreen" | "light green" => Self::LightGreen,
+            "lightyellow" | "light yellow" => Self::LightYellow,
+            "lightblue" | "light blue" => Self::LightBlue,
+            "lightmagenta" | "light magenta" => Self::LightMagenta,
+            "lightcyan" | "light cyan" => Self::LightCyan,
+            "white" => Self::White,
+            _ => {
+                if let Ok(rgb) = Rgb::from_hex_str(&format!("#{s}")) {
+                    Self::Rgb(rgb.red() as u8, rgb.green() as u8, rgb.blue() as u8)
```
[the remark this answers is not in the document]
Mehrbod2002: but I should think without #

sophacles (changes requested):
Code looks good to me.

Would like to see documentation - its non-obvious how to make Indexed and Rgb from a string without reading the code.

orhun:
I think we should also add unit tests for this.

Mehrbod2002:
> I think we should also add unit tests for this.

The tests added

joshka (approved):
Nice work. This looks good to me.

sophacles (dismissed):

src/style.rs
```diff
@@ -311,4 +386,28 @@ mod tests {
...
+    fn test_custom_color() {
+        let color: Color = Color::from_str("lightblue").unwrap();
+        assert_eq!(color, Color::LightBlue);
+    }
+
+    #[test]
+    fn test_invalid_color() {
+        let color: Result<Color, _> = Color::from_str("invalid_color");
+        assert!(color.is_err());
+    }
```
sophacles, line 412: Modifying this test to: 
```rust
    #[test]
    fn test_invalid_colors() {
        let bad_colors = [
            "invalid_color", // not a color string
            "abcdef0",       // 7 chars is not a color
            " bcdefa",       // doesn't start with a '#'
            "blue ",         // has space at end
            " blue",         // has space at start
            "#abcdef00",     // too many chars
        ];

        for bad_color in bad_colors {
            assert!(
                Color::from_str(bad_color).is_err(),
                "bad color: '{}'",
                bad_color
            );
        }
    }
```
Fails

src/style.rs
```diff
@@ -246,6 +247,80 @@ impl Style {
...
+            "lightgreen" | "light green" => Self::LightGreen,
+            "lightyellow" | "light yellow" => Self::LightYellow,
+            "lightblue" | "light blue" => Self::LightBlue,
+            "lightmagenta" | "light magenta" => Self::LightMagenta,
+            "lightcyan" | "light cyan" => Self::LightCyan,
+            "white" => Self::White,
+            _ => {
+                if let Ok(index) = s.parse::<u8>() {
+                    Self::Indexed(index)
+                } else if let (Ok(r), Ok(g), Ok(b)) = (
```
sophacles, line 310: If we do something like this (my code here is ugly, but gets the job done):
```rust
                } else if let (Ok(r), Ok(g), Ok(b)) = {
                    if !s.starts_with('#') || s.len() != 7 {
                        return Err(ParseColorError);
                    }
                    (
                        u8::from_str_radix(&s[1..3], 16),
                        u8::from_str_radix(&s[3..5], 16),
                        u8::from_str_radix(&s[5..7], 16),
                    )
                } {
                    Self::Rgb(r, g, b)
```
The test cases below pass.

sophacles (changes requested):
There's some edge cases that we probably want to deal with.

sophacles (approved):

orhun (approved):
Neat.

Issue #173, opened 2023-05-10
Table: Add scrollbar to table widget


lcmgh:
## Problem
I noticed the table widget is already scrollable via keyboard navigation: https://github.com/tui-rs-revival/ratatui/blob/main/examples/table.rs

But it is lacking of a scrollbar.

Relates to https://github.com/tui-rs-revival/ratatui/discussions/146

## Solution
Add a config `scrollbar: bool` option or something similar that integrates a scrollbar to visualize the current position.

joshka:
Perhaps @gibbz00 has some input on this? They mentioned that they were experimenting with Viewport scrolling a little while ago in https://github.com/tui-rs-revival/ratatui/discussions/96#discussioncomment-5426219. It might be related a little.

I wonder whether the entire concept of scrolling shouldn't be pulled out of widgets themselves and into something a little higher level (in a similar fashion to how block acts?). I'm not sure how that would look though.

gibbz00:
Close this now that #228 has landed?

joshka:
I think I'd still like to consider adding scrollbars to the widget as part of the config - e.g. what if table looks something like:
```
Table::new(rows).scrollbar(Scrollbar::new()...))
```
Instead of having to manually lay this out?

joshka:
I'm closing this out for now to help cut down the open issues to actionable things. This will be addressed as part of #174 / https://github.com/ratatui/ratatui/discussions/1924

PR #181, opened 2023-05-19
fix: MouseEvent for table for scrolling

Mehrbod2002:
fixes: #173   

and also able to set style in right side of table widget and BG color and use it up and down with it on ScrollEvent , but not sure how to grab it up and down . supposed grab an item done before in this crate and hope developers guide me if the attention of issue was this .

### DIFF
--- a/src/widgets/table.rs
+++ b/src/widgets/table.rs
@@ -5,6 +5,8 @@ use crate::{
     text::Text,
     widgets::{Block, StatefulWidget, Widget},
 };
+#[cfg(feature = "crossterm")]
+use crossterm::event::{self, Event, MouseEventKind};
 use unicode_width::UnicodeWidthStr;
 
 /// A [`Cell`] contains the [`Text`] to be displayed in a [`Row`] of a [`Table`].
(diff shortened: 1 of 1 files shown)


joshka:
Hiya, thanks for raising this PR. We appreciate all contributions and I hope that I can encourage you to continue doing so even though I'm pushing back against this specific PR.

This PR adds crossterm event reading (if the crossterm feature is enabled) into the render method of the table widget. This is something that no other widgets in the library do so far, so it's a bit of a new idea. I think there's probably a bunch of problems that come from it. The first one I can see is that the read() call is blocking, which would mean that if no mouse events are coming, then this blocks all future renders.

Another issue is that this would mean that there is only a single scroll target for an entire application.

One big thing that should be considered is that applications written with ratatui are currently able to handle their own events etc. with whatever method is appropriate for the app (e.g. I have an app that does this with non blocking event streams, joined with other event sources, and which bubbles events to various widgets based on which is currently visible).

So, while I welcome this idea, I need to suggest that talking through the design of this needs to happen before a PR.

For some notes and discussion about scrolling approaches generally, see Take a look at https://github.com/tui-rs-revival/ratatui/issues/174. It's a work in progress, and would be great to hear some feedback on.

Mehrbod2002:
@joshka So it should be a State for scroll which is upper state and scrollbar to drag .
even I can't understand how it will possible on this idea as core code events all defined on loop and not in core .

joshka:
> @joshka So it should be a State for scroll which is upper state and scrollbar to drag . even I can't understand how it will possible on this idea as core code events all defined on loop and not in core .

I'm having some difficulty understanding what you're saying. If English is your second language, can you write a response in your own language and ask ChatGPT to translate it for you?

The app loop is code that users of this library write themselves. We don't provide the loop (except in a sample app). Each app can write that loop differently. E.g. some might render every 250ms, some others might only ever call render when there is a new event that changes the application's state.

The render function of a widget should not block waiting for a crossterm event. Blocking will prevent other events causing a render. To see this problem, run `cargo run --example table`. You will not see any output until you move the mouse. This is a problem.

The second problem is what to do when there are two tables. The application should make a choice about how to handle that. We can't make that choice in this library.

Therefore the mouse event handling needs to be outside of the widget, not part of the widget.

orhun:
Closing due to inactivity.

PR #188, opened 2023-05-21
docs(apps): add tts-tui

lesleyrs:


sayanarijit:
Nice... We're moving things from APPS.md to [this wiki](https://github.com/tui-rs-revival/ratatui/wiki/Apps-using-Ratatui). Can you add it there?

PR #191, opened 2023-05-22
chore: lint and doc cleanup

nyurik:
A few more minor cleanups, mostly in documentation

### DIFF
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -37,7 +37,6 @@ pub trait Backend {
 
     /// Insert `n` line breaks to the terminal screen
     fn append_lines(&mut self, _n: u16) -> io::Result<()> {
-        // to get around the unused warning
         Ok(())
     }
 
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -36,9 +36,8 @@ pub trait Backend {
         I: Iterator<Item = (u16, u16, &'a Cell)>;
 
     /// Insert `n` line breaks to the terminal screen
-    fn append_lines(&mut self, n: u16) -> io::Result<()> {
+    fn append_lines(&mut self, _n: u16) -> io::Result<()> {
         // to get around the unused warning
-        let _n = n;
         Ok(())
     }
 
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -503,7 +503,11 @@ macro_rules! assert_buffer_eq {
                         .enumerate()
                         .map(|(i, (x, y, cell))| {
                             let expected_cell = expected.get(*x, *y);
-                            format!("{i}: at ({x}, {y})\n  expected: {expected_cell:?}\n  actual:   {cell:?}")
+                            indoc::formatdoc! {"
+                                {i}: at ({x}, {y})
+                                  expected: {expected_cell:?}
+                                  actual:   {cell:?}
+                            "}
                         })
                         .collect::<Vec<String>>()
                         .join("\n");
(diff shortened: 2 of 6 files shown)


joshka:
I notice a bunch of checks are failing on this.

It looks like this change effectively ignores a bunch of places that could be better written. I wonder if it would be worth leaving the lints for those on rather than ignoring them, with a goal to fix the issues rather than ignore them.

nyurik:
> I notice a bunch of checks are failing on this.
> 
> It looks like this change effectively ignores a bunch of places that could be better written. I wonder if it would be worth leaving the lints for those on rather than ignoring them, with a goal to fix the issues rather than ignore them.

ok, makes sense, i removed the clippy changes for now

src/buffer.rs
```diff
@@ -531,11 +531,9 @@ impl Debug for Buffer {
...
     ///
     /// The format is like a pretty printed struct, with the following fields:
-    /// area: displayed as Rect { x: 1, y: 2, width: 3, height: 4 }
-    /// content: displayed as a list of strings representing the content of the
-    ///     buffer
-    /// styles: displayed as a list of
-    ///     { x: 1, y: 2, fg: Color::Red, bg: Color::Blue, modifier: Modifier::BOLD }
+    /// * `area`: displayed as `Rect { x: 1, y: 2, width: 3, height: 4 }`
+    /// * `content`: displayed as a list of strings representing the content of the buffer
+    /// * `styles`: displayed as a list of `{ x: 1, y: 2, fg: Color::Red, bg: Color::Blue, modifier: Modifier::BOLD }`
```
joshka, line 536: I'd suggest leaving this as a 100 char line limit (that limit is used quite a few places in the rust ecosystem). Objectively it helps to avoid horizontal scrolling or wrapping being necessary to read code:

E.g.:
image

vs current:

image

src/backend/mod.rs
```diff
@@ -38,6 +38,7 @@ pub trait Backend {
     /// Insert `n` line breaks to the terminal screen
     fn append_lines(&mut self, n: u16) -> io::Result<()> {
         // to get around the unused warning
+        #[allow(clippy::no_effect_underscore_binding)]
         let _n = n;
         Ok(())
     }
```
mindoodoo, line 44:
```suggestion
    /// Insert `n` line breaks to the terminal screen
    #[allow(unused)]
    fn append_lines(&mut self, n: u16) -> io::Result<()> {
        Ok(())
    }
```

I don't understand why the `let _n = n` was used in the first place as this removes that warning, maybe wasn't an available feature before, or am I missing something ?

mindoodoo (changes requested):

mindoodoo:
Wasn't exactly sure for the conventional commit (PR title) as this is both linting and a couple doc fixes so marked it as unscoped chore, maintainers feel free to change it to something more adequate if you have any ideas.

src/backend/mod.rs
joshka: My guess is that there is a concern that `n` in the doc comment and `_n` in the parameter list are noth the same, or that `_n` doesn't match the trait perhaps? Either way, I think probably just using _n in the parameter list is good enough.

joshka:
>Wasn't exactly sure for the conventional commit (PR title) as this is both linting and a couple doc fixes so marked it as unscoped chore, maintainers feel free to change it to something more adequate if you have any ideas.

chore is fine. I made the changes that we both suggested (just used _n instead of a lint exception)

joshka (approved):

orhun (approved):

nyurik:
@joshka were the Cargo.toml changes intentional?

nyurik:
Ah, I see - you moved the indoc into dependencies because of the format usage, plus sorted them.  Slightly outside of the initial scope, but sure, not a biggie :)

I removed the no-longer-needed comment about the underscore, and rebased on main

joshka (approved):

PR #205, opened 2023-05-31
build(deps)!: upgrade bitflags to 2.3

joshka:
build(deps)!: upgrade bitflags to 2.3

BREAKING CHANGE: The serde representation of bitflags has changed. Any
existing serialized types that have Borders or Modifiers will need to be
re-serialized. This is documented in the bitflags changelog.
https://github.com/bitflags/bitflags/blob/main/CHANGELOG.md#200-rc2

### DIFF
--- a/src/widgets/mod.rs
+++ b/src/widgets/mod.rs
@@ -31,6 +31,8 @@ mod sparkline;
 mod table;
 mod tabs;
 
+use std::fmt::{self, Debug};
+
 pub use self::barchart::BarChart;
 pub use self::block::{Block, BorderType, Padding};
 pub use self::chart::{Axis, Chart, Dataset, GraphType};
@@ -47,6 +49,7 @@ use bitflags::bitflags;
 
 bitflags! {
     /// Bitflags that can be composed to set the visible borders essentially on the block widget.
+    #[derive(Clone, Copy, Default, PartialEq, Eq)]
     pub struct Borders: u8 {
         /// Show no border (default)
         const NONE   = 0b0000;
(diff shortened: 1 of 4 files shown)


joshka:
Fixes the dependencies out of date badge on the new README in https://github.com/tui-rs-revival/ratatui/pull/204

src/widgets/mod.rs
```diff
@@ -59,7 +62,37 @@ bitflags! {
...
+            if border == Borders::NONE {
+                continue;
+            }
+            if first {
+                write!(f, "{}", name)?;
+                first = false;
+            } else {
+                write!(f, " | {}", name)?;
+            }
+        }
```
orhun, line 94:
```suggestion
        for (i, (name, border)) in self.iter_names().enumerate() {
            if border == Borders::NONE {
                continue;
            }
            if i == 0 {
                write!(f, "{}", name)?;
            } else {
                write!(f, " | {}", name)?;
            }
        }
```

sayanarijit (approved):

src/widgets/mod.rs
joshka: I actually wrote that approach and then switched to the version in this commit because there doesn't seem like an obvious guarantee that Borders::NONE will always be the first element. If bitflags changes to alpha order, or highest to lowest bit order then this fails. Maybe I'm too pessimistic on that though and we'd just catch that in tests and so it's not really a problem.

orhun: Oh I see, it makes sense then.

orhun (approved):

PR #207, opened 2023-05-31
feat(paragraph): allow setting vertical alignment

Jules-Bertholet:
Add an option to the `Paragraph` widget to allow aligning text to the bottom of the widget (overflowing on top).

### DIFF
--- a/src/layout.rs
+++ b/src/layout.rs
@@ -59,6 +59,13 @@ pub enum Alignment {
     Right,
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum VerticalAlignment {
+    Top,
+    Bottom,
+}
+
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub struct Layout {
     direction: Direction,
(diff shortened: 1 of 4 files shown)


orhun:
Thanks for the PR! Haven't tested this but I'm curious about your use-case for this feature since it seems like an edge case to place the text at the bottom/top of the widget like you mentioned.

Also, updating the examples would be nice if we decide to go forward with this feature. Needs more comments from other maintainers.

mindoodoo:
+1 for the use case + examples

Also maybe add a doc comment for the `VerticalAlignment` struct.

sayanarijit:
LGTM. Need to resolve conflict.

joshka:
We need to hold off on merging this until #193 has landed because both this and that have some pretty large changes to the same code. I'd expect there to be a lot of rework of this necessary after 193 gets over the line.

Note that the conflict between two people working in a similar area highlights the need for better code factorization here (to follow OCP) and prevent similar issues.

TimerErTim:
One use case (at least for me) is to extend a wall of text at the bottom of the paragraph instead of the top. Currently, this is impossible to do since the only system that allows this is scrolling and that is feasible with wrapping because it is impossible to predict the wrapped line count.

orhun:
Hey @Jules-Bertholet, we would like to get this over the finish line. Are you still interested in working on this? 🐻

orhun:
Closing due to inactivity. Would be nice to come back to this at some point.

PR #217, opened 2023-06-03
feat(list): allow widgets as list items

joshka:
This is a WIP alternative to #155, built on top of #122.

Instead of creating a new widget list type, this PR extends the existing list type to allow it to render widgets as well as text (and does so without impacting existing code that uses lists).

The majority of this PR is from the the changes in #122, which make it possible to render widgets by ref rather than value.

This doesn't implement the widget sizing in #155 however (this is currently manual and TODO).

Ping @karthago1 for some thoughts.

### DIFF
--- a/src/widgets/list.rs
+++ b/src/widgets/list.rs
@@ -1,3 +1,5 @@
+use std::fmt::Debug;
+
 use crate::{
     buffer::Buffer,
     layout::{Corner, Rect},
@@ -6,6 +8,8 @@ use crate::{
     widgets::{Block, StatefulWidget, Widget},
 };
 use unicode_width::UnicodeWidthStr;
+
+use super::Paragraph;
 
 #[derive(Debug, Clone, Default)]
 pub struct ListState {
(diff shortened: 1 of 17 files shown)


src/widgets/list.rs
```diff
@@ -73,6 +77,64 @@ impl<'a> ListItem<'a> {
...
+pub struct WidgetWrapper<'a> {
+    widget: Box<dyn Widget + 'a>,
+    height: usize,
+    width: usize,
+}
+
+impl<'a> WidgetWrapper<'a> {
+    pub fn new(widget: impl Widget + 'a, height: usize, width: usize) -> Self {
+        Self {
+            widget: Box::new(widget),
```
karthago1, line 95: it is really a nice alternative solution.

1. In my implementation I tried to avoid the heap allocations at all cost. but maybe it is fine to use the heap instead of generic types / templates. it makes the code a little bit cleaner. When I have time, maybe I will try to do some benchmarks with a long list.
2. I like that the render function takes a reference instead of consuming self. This will make some problems easier to solve.

joshka:
Closing out as a stale PoC

PR #224, opened 2023-06-05
Add tui-journal to APPS.md 

Owletti:
tui-journal - https://github.com/AmmarAbouZor/tui-journal
Still on tui instead of ratatui, but using the library nonetheless


orhun:
Hey, thanks for your PR! Is there any plans to migrate to `ratatui`? Let us know if we can help with anything!

mindoodoo:
Aren't we putting the apps in the wiki now ? or are we doing both ?

orhun:
I think we can remove `APPS.md` and only use the Wiki.

orhun:
APPS.md no longer exists, please use the Wiki.

PR #225, opened 2023-06-07
docs(readme): add contributors graph image

joshka:


mindoodoo (commented):
I kind of like the idea of potentially keeping the acknowledgement section in addition to the contributors section. What do you think ?

The idea being we can keep special acknowledgements in the acknowledgement section, such as the logo creator, the og maintainer, etc...

orhun:
I agree with @mindoodoo, let's keep the acknowledgements section.

joshka:
I moved it back in to an acknowledgment section, reformated to 100 columns, added a markdownlint config, changed the link to the APPs wiki page, removed APPS.md, and added a link to the CHANGELOG (it's always useful being able to find that from the crates.io page).

mindoodoo (approved):
LGTM, thanks !

orhun (approved):

PR #228, opened 2023-06-08
feat: add scrollbar widget

kdheepak:
I've added a simple scrollbar widget.

```
 <--▮------->
 ^  ^   ^   ^
 │  │   │   └ end arrow
 │  │   └──── track symbol
 │  └──────── thumb symbol
 └─────────── begin arrow
```

image

It can be used to add a scrollbar to any `Rect`. 
It can also be customized with different styles and symbols for, the bar, the "scrollwheel" and arrows.

This addresses https://github.com/tui-rs-revival/ratatui/issues/173

A larger discussion is to be had about scrollable widgets in general: https://github.com/tui-rs-revival/ratatui/issues/174

I'm open to feedback on improving this. I've needed a scrollbar for a couple of projects now and figured I'd clean up the code and share it as a `Widget` here. 

_Docstrings were generated using ChatGPT_.

### DIFF
--- a/examples/scrollbar.rs
+++ b/examples/scrollbar.rs
@@ -1,0 +1,273 @@
+use std::{
+    error::Error,
+    io,
+    time::{Duration, Instant},
+};
+
+use crossterm::{
+    event::{self, DisableMouseCapture, EnableMouseCapture, Event, KeyCode},
+    execute,
+    terminal::{disable_raw_mode, enable_raw_mode, EnterAlternateScreen, LeaveAlternateScreen},
+};
+use ratatui::{
+    backend::{Backend, CrosstermBackend},
+    layout::{Alignment, Constraint, Direction, Layout, Margin},
+    style::{Color, Modifier, Style},
+    text::{Line, Masked, Span},
+    widgets::{
+        scrollbar, Block, Borders, Paragraph, Scrollbar, ScrollbarOrientation, ScrollbarState,
+    },
+    Frame, Terminal,
+};
+
+#[derive(Default)]
+struct App {
+    pub vertical_scroll_state: ScrollbarState,
+    pub horizontal_scroll_state: ScrollbarState,
+    pub vertical_scroll: usize,
+    pub horizontal_scroll: usize,
+}
+
+fn main() -> Result<(), Box<dyn Error>> {
+    // setup terminal
+    enable_raw_mode()?;
+    let mut stdout = io::stdout();
+    execute!(stdout, EnterAlternateScreen, EnableMouseCapture)?;
+    let backend = CrosstermBackend::new(stdout);
+    let mut terminal = Terminal::new(backend)?;
+
+    // create app and run it
+    let tick_rate = Duration::from_millis(250);
+    let app = App::default();
+    let res = run_app(&mut terminal, app, tick_rate);
+
+    // restore terminal
+    disable_raw_mode()?;
+    execute!(
+        terminal.backend_mut(),
+        LeaveAlternateScreen,
+        DisableMouseCapture
+    )?;
+    terminal.show_cursor()?;
+
+    if let Err(err) = res {
+        println!("{err:?}");
+    }
+
+    Ok(())
+}
+
+fn run_app<B: Backend>(
+    terminal: &mut Terminal<B>,
+    mut app: App,
+    tick_rate: Duration,
+) -> io::Result<()> {
+    let mut last_tick = Instant::now();
+    loop {
+        terminal.draw(|f| ui(f, &mut app))?;
+
+        let timeout = tick_rate
+            .checked_sub(last_tick.elapsed())
+            .unwrap_or_else(|| Duration::from_secs(0));
+        if crossterm::event::poll(timeout)? {
+            if let Event::Key(key) = event::read()? {
+                match key.code {
+                    KeyCode::Char('q') => return Ok(()),
+                    KeyCode::Char('j') => {
+                        app.vertical_scroll = app.vertical_scroll.saturating_add(1);
+                        app.vertical_scroll_state = app
+                            .vertical_scroll_state
+                            .position(app.vertical_scroll as u16);
+                    }
+                    KeyCode::Char('k') => {
+                        app.vertical_scroll = app.vertical_scroll.saturating_sub(1);
+                        app.vertical_scroll_state = app
+                            .vertical_scroll_state
+                            .position(app.vertical_scroll as u16);
+                    }
+                    KeyCode::Char('h') => {
+                        app.horizontal_scroll = app.horizontal_scroll.saturating_sub(1);
+                        app.horizontal_scroll_state = app
+                            .horizontal_scroll_state
+                            .position(app.horizontal_scroll as u16);
+                    }
+                    KeyCode::Char('l') => {
+                        app.horizontal_scroll = app.horizontal_scroll.saturating_add(1);
+                        app.horizontal_scroll_state = app
+                            .horizontal_scroll_state
+                            .position(app.horizontal_scroll as u16);
+                    }
+                    _ => {}
+                }
+            }
+        }
+        if last_tick.elapsed() >= tick_rate {
+            last_tick = Instant::now();
+        }
+    }
+}
+
+fn ui<B: Backend>(f: &mut Frame<B>, app: &mut App) {
+    let size = f.size();
+
+    // Words made "loooong" to demonstrate line breaking.
+    let s = "Veeeeeeeeeeeeeeeery    loooooooooooooooooong   striiiiiiiiiiiiiiiiiiiiiiiiiing.   ";
+    let mut long_line = s.repeat(usize::from(size.width) / s.len() + 4);
+    long_line.push('\n');
+
+    let block = Block::default().style(Style::default().fg(Color::Black));
+    f.render_widget(block, size);
+
+    let chunks = Layout::default()
+        .direction(Direction::Vertical)
+        .margin(2)
+        .constraints(
+            [
+                Constraint::Min(1),
+                Constraint::Percentage(25),
+                Constraint::Percentage(25),
+                Constraint::Percentage(25),
+                Constraint::Percentage(25),
+            ]
+            .as_ref(),
+        )
+        .split(size);
+
+    let text = vec![
+        Line::from("This is a line "),
+        Line::from(Span::styled(
+            "This is a line   ",
+            Style::default().fg(Color::Red),
+        )),
+        Line::from(Span::styled(
+            "This is a line",
+            Style::default().bg(Color::DarkGray),
+        )),
+        Line::from(Span::styled(
+            "This is a longer line",
+            Style::default().add_modifier(Modifier::CROSSED_OUT),
+        )),
+        Line::from(Span::styled(&long_line, Style::default())),
+        Line::from(Span::styled("This is a line", Style::default())),
+        Line::from(vec![
+            Span::raw("Masked text: "),
+            Span::styled(
+                Masked::new("password", '*'),
+                Style::default().fg(Color::Red),
+            ),
+        ]),
+        Line::from("This is a line "),
+        Line::from(Span::styled(
+            "This is a line   ",
+            Style::default().fg(Color::Red),
+        )),
+        Line::from(Span::styled(
+            "This is a line",
+            Style::default().bg(Color::DarkGray),
+        )),
+        Line::from(Span::styled(
+            "This is a longer line",
+            Style::default().add_modifier(Modifier::CROSSED_OUT),
+        )),
+        Line::from(Span::styled(&long_line, Style::default())),
+        Line::from(Span::styled("This is a line", Style::default())),
+        Line::from(vec![
+            Span::raw("Masked text: "),
+            Span::styled(
+                Masked::new("password", '*'),
+                Style::default().fg(Color::Red),
+            ),
+        ]),
+    ];
+    app.vertical_scroll_state = app.vertical_scroll_state.content_length(text.len() as u16);
+    app.horizontal_scroll_state = app
+        .horizontal_scroll_state
+        .content_length(long_line.len() as u16);
+
+    let create_block = |title| {
+        Block::default()
+            .borders(Borders::ALL)
+            .style(Style::default().fg(Color::Gray))
+            .title(Span::styled(
+                title,
+                Style::default().add_modifier(Modifier::BOLD),
+            ))
+    };
+
+    let title = Block::default()
+        .title("Use h j k l to scroll ◄ ▲ ▼ ►")
+        .title_alignment(Alignment::Center);
+    f.render_widget(title, chunks[0]);
+
+    let paragraph = Paragraph::new(text.clone())
+        .style(Style::default().fg(Color::Gray))
+        .block(create_block("Vertical scrollbar with arrows"))
+        .scroll((app.vertical_scroll as u16, 0));
+    f.render_widget(paragraph, chunks[1]);
+    f.render_stateful_widget(
+        Scrollbar::default()
+            .orientation(ScrollbarOrientation::VerticalRight)
+            .begin_symbol(Some("↑"))
+            .end_symbol(Some("↓")),
+        chunks[1],
+        &mut app.vertical_scroll_state,
+    );
+
+    let paragraph = Paragraph::new(text.clone())
+        .style(Style::default().fg(Color::Gray))
+        .block(create_block(
+            "Vertical scrollbar without arrows and mirrored",
+        ))
+        .scroll((app.vertical_scroll as u16, 0));
+    f.render_widget(paragraph, chunks[2]);
+    f.render_stateful_widget(
+        Scrollbar::default()
+            .orientation(ScrollbarOrientation::VerticalLeft)
+            .symbols(scrollbar::VERTICAL)
+            .begin_symbol(None)
+            .end_symbol(None),
+        chunks[2].inner(&Margin {
+            vertical: 1,
+            horizontal: 0,
+        }),
+        &mut app.vertical_scroll_state,
+    );
+
+    let paragraph = Paragraph::new(text.clone())
+        .style(Style::default().fg(Color::Gray))
+        .block(create_block(
+            "Horizontal scrollbar with only begin arrow & custom thumb symbol",
+        ))
+        .scroll((0, app.horizontal_scroll as u16));
+    f.render_widget(paragraph, chunks[3]);
+    f.render_stateful_widget(
+        Scrollbar::default()
+            .orientation(ScrollbarOrientation::HorizontalBottom)
+            .thumb_symbol("🬋")
+            .end_symbol(None),
+        chunks[3].inner(&Margin {
+            vertical: 0,
+            horizontal: 1,
+        }),
+        &mut app.horizontal_scroll_state,
+    );
+
+    let paragraph = Paragraph::new(text.clone())
+        .style(Style::default().fg(Color::Gray))
+        .block(create_block(
+            "Horizontal scrollbar without arrows & custom thumb and track symbol",
+        ))
+        .scroll((0, app.horizontal_scroll as u16));
+    f.render_widget(paragraph, chunks[4]);
+    f.render_stateful_widget(
+        Scrollbar::default()
+            .orientation(ScrollbarOrientation::HorizontalBottom)
+            .thumb_symbol("░")
+            .track_symbol("─"),
+        chunks[4].inner(&Margin {
+            vertical: 0,
+            horizontal: 1,
+        }),
+        &mut app.horizontal_scroll_state,
+    );
+}
--- a/src/widgets/scrollbar.rs
+++ b/src/widgets/scrollbar.rs
@@ -1,0 +1,871 @@
+use super::StatefulWidget;
+use crate::{
+    buffer::Buffer,
+    layout::Rect,
+    style::Style,
+    symbols::{block::FULL, line},
+};
+
+/// Scrollbar Set
+/// <--▮------->
+/// ^  ^   ^   ^
+/// │  │   │   └ end
+/// │  │   └──── track
+/// │  └──────── thumb
+/// └─────────── begin
+#[derive(Debug, Clone)]
+pub struct Set {
+    pub track: &'static str,
+    pub thumb: &'static str,
+    pub begin: &'static str,
+    pub end: &'static str,
+}
+
+pub const DOUBLE_VERTICAL: Set = Set {
+    track: line::DOUBLE_VERTICAL,
+    thumb: FULL,
+    begin: "▲",
+    end: "▼",
+};
+
+pub const DOUBLE_HORIZONTAL: Set = Set {
+    track: line::DOUBLE_HORIZONTAL,
+    thumb: FULL,
+    begin: "◄",
+    end: "►",
+};
+
+pub const VERTICAL: Set = Set {
+    track: line::VERTICAL,
+    thumb: FULL,
+    begin: "↑",
+    end: "↓",
+};
+
+pub const HORIZONTAL: Set = Set {
+    track: line::HORIZONTAL,
+    thumb: FULL,
+    begin: "←",
+    end: "→",
+};
+
+/// An enum representing the direction of scrolling in a Scrollbar widget.
+#[derive(Clone, Copy, Debug, Default, Eq, PartialEq)]
+pub enum ScrollDirection {
+    /// Forward scroll direction, usually corresponds to scrolling downwards or rightwards.
+    #[default]
+    Forward,
+    /// Backward scroll direction, usually corresponds to scrolling upwards or leftwards.
+    Backward,
+}
+
+/// A struct representing the state of a Scrollbar widget.
+///
+/// For example, in the following list, assume there are 4 bullet points:
+///
+/// - the `position` is 0
+/// - the `content_length` is 4
+/// - the `viewport_content_length` is 2
+///
+/// ┌───────────────┐
+/// │1. this is a   █
+/// │   single item █
+/// │2. this is a   ║
+/// │   second item ║
+/// └───────────────┘
+///
+/// If you don't have multi-line content, you can leave the `viewport_content_length` set to the
+/// default of 0 and it'll use the track size as a `viewport_content_length`.
+#[derive(Clone, Copy, Debug, Default)]
+pub struct ScrollbarState {
+    // The current position within the scrollable content.
+    position: u16,
+    // The total length of the scrollable content.
+    content_length: u16,
+    // The length of content in current viewport.
+    viewport_content_length: u16,
+}
+
+impl ScrollbarState {
+    /// Sets the scroll position of the scrollbar and returns the modified ScrollbarState.
+    pub fn position(mut self, position: u16) -> Self {
+        self.position = position;
+        self
+    }
+
+    /// Sets the length of the scrollable content and returns the modified ScrollbarState.
+    pub fn content_length(mut self, content_length: u16) -> Self {
+        self.content_length = content_length;
+        self
+    }
+
+    /// Sets the length of the viewport content and returns the modified ScrollbarState.
+    pub fn viewport_content_length(mut self, viewport_content_length: u16) -> Self {
+        self.viewport_content_length = viewport_content_length;
+        self
+    }
+
+    /// Decrements the scroll position by one, ensuring it doesn't go below zero.
+    pub fn prev(&mut self) {
+        self.position = self.position.saturating_sub(1);
+    }
+
+    /// Increments the scroll position by one, ensuring it doesn't exceed the length of the content.
+    pub fn next(&mut self) {
+        self.position = self
+            .position
+            .saturating_add(1)
+            .clamp(0, self.content_length.saturating_sub(1))
+    }
+
+    /// Sets the scroll position to the start of the scrollable content.
+    pub fn first(&mut self) {
+        self.position = 0;
+    }
+
+    /// Sets the scroll position to the end of the scrollable content.
+    pub fn last(&mut self) {
+        self.position = self.content_length.saturating_sub(1)
+    }
+
+    /// Changes the scroll position based on the provided ScrollDirection.
+    pub fn scroll(&mut self, direction: ScrollDirection) {
+        match direction {
+            ScrollDirection::Forward => {
+                self.next();
+            }
+            ScrollDirection::Backward => {
+                self.prev();
+            }
+        }
+    }
+}
+
+/// Scrollbar Orientation
+#[derive(Default, Debug, Clone)]
+pub enum ScrollbarOrientation {
+    #[default]
+    VerticalRight,
+    VerticalLeft,
+    HorizontalBottom,
+    HorizontalTop,
+}
+
+/// Scrollbar widget for tui-rs library.
+///
+/// This widget can be used to display a scrollbar in a terminal user interface.
+/// The following components of the scrollbar are customizable in symbol and style.
+///
+/// <--▮------->
+/// ^  ^   ^   ^
+/// │  │   │   └ end
+/// │  │   └──── track
+/// │  └──────── thumb
+/// └─────────── begin
+#[derive(Debug, Clone)]
+pub struct Scrollbar<'a> {
+    orientation: ScrollbarOrientation,
+    thumb_style: Style,
+    thumb_symbol: &'a str,
+    track_style: Style,
+    track_symbol: &'a str,
+    begin_symbol: Option<&'a str>,
+    begin_style: Style,
+    end_symbol: Option<&'a str>,
+    end_style: Style,
+}
+
+impl<'a> Default for Scrollbar<'a> {
+    fn default() -> Self {
+        Self {
+            orientation: ScrollbarOrientation::default(),
+            thumb_symbol: DOUBLE_VERTICAL.thumb,
+            thumb_style: Style::default(),
+            track_symbol: DOUBLE_VERTICAL.track,
+            track_style: Style::default(),
+            begin_symbol: Some(DOUBLE_VERTICAL.begin),
+            begin_style: Style::default(),
+            end_symbol: Some(DOUBLE_VERTICAL.end),
+            end_style: Style::default(),
+        }
+    }
+}
+
+impl<'a> Scrollbar<'a> {
+    pub fn new(orientation: ScrollbarOrientation) -> Self {
+        Self::default().orientation(orientation)
+    }
+
+    /// Sets the orientation of the scrollbar.
+    /// Resets the symbols to [`DOUBLE_VERTICAL`] or [`DOUBLE_HORIZONTAL`] based on orientation
+    pub fn orientation(mut self, orientation: ScrollbarOrientation) -> Self {
+        self.orientation = orientation;
+        let set = if self.is_vertical() {
+            DOUBLE_VERTICAL
+        } else {
+            DOUBLE_HORIZONTAL
+        };
+        self.symbols(set)
+    }
+
+    /// Sets the orientation and symbols for the scrollbar from a [`Set`].
+    pub fn orientation_and_symbol(mut self, orientation: ScrollbarOrientation, set: Set) -> Self {
+        self.orientation = orientation;
+        self.symbols(set)
+    }
+
+    /// Sets the symbol that represents the thumb of the scrollbar.
+    pub fn thumb_symbol(mut self, thumb_symbol: &'a str) -> Self {
+        self.thumb_symbol = thumb_symbol;
+        self
+    }
+
+    /// Sets the style that represents the thumb of the scrollbar.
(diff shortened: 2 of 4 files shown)


src/widgets/scrollbar.rs
```diff
@@ -0,0 +1,208 @@
...
+        if scroll < 0.0 {
+            self.scroll = 0.0;
+        } else {
+            self.scroll = scroll;
+        }
+        self
+    }
+
+    /// Sets the symbol that represents the scroller.
+    pub fn symbol(mut self, symbol: &'a str) -> Self {
```
joshka, line 44: I wonder if we should just use https://crates.io/crates/derive_setters for new widgets.

src/widgets/scrollbar.rs
```diff
@@ -0,0 +1,208 @@
...
+        // Adjust the scroll position to fit within the range.
+        let scroll = if self.scroll >= 1.0 {
+            // Conversions rounds down, so make this slightly less than 1.0
+            0.9999
+        } else {
+            self.scroll
+        };
+        // Calculate the position of the scroller based on the scroll position and the height of the
+        // scrollbar.
+        let pos = height as f64 * scroll;
```
joshka, line 103: can you use `clamp()` here? e.g. something like:
 `let pos = (height * self.scroll).clamp(0, height) as u16;`?

src/widgets/scrollbar.rs
```diff
@@ -0,0 +1,208 @@
...
+    }
+}
+
+impl<'a> Widget for Scrollbar<'a> {
+    /// Renders the scrollbar.
+    ///
+    /// The scrollbar is drawn with a fixed width on the right edge of the given area.
+    fn render(self, area: Rect, buf: &mut Buffer) {
+        // If the area is too small, we don't draw the scrollbar.
+        if area.height <= 2 {
```
joshka, line 74: nit: use a constant to avoid writing a comment (e.g. `MIN_HEIGHT`)

src/widgets/scrollbar.rs
```diff
@@ -0,0 +1,208 @@
...
+            return;
+        }
+
+        // Calculate the rightmost column for the scrollbar.
+        let right = area.right().saturating_sub(1);
+
+        // If the rightmost column is out of bounds, we don't draw the scrollbar.
+        if right < area.left() {
+            return;
+        };
```
joshka, line 84: nit: handle this in the same if statement above (as `MIN_WIDTH`)

src/widgets/scrollbar.rs
```diff
@@ -0,0 +1,208 @@
...
+
+        // Calculate the top row and the height for the scrollbar.
+        let (top, height) = { (area.top() + 1, area.height.saturating_sub(2)) };
+
+        // Draw the bar of the scrollbar.
+        for y in top..(top + height) {
+            buf.set_string(right, y, self.bar_symbol.clone(), self.bar_style);
```
joshka: what's the clone() issue that github flags here?

joshka:
Thanks for working on this btw - looking forward to seeing it be part of ratatui.

kdheepak:
Thanks for the detailed feedback!!! 

Let me address some of the comments and get back to you when it is ready for another round for review. 

I'm also planning to update this scrollbar widget so that it works horizontally as well.

src/widgets/scrollbar.rs
kdheepak: I previously had `self.bar_symbol` as a `String` in my personal project but changed it to `&str` before submitting the PR, but I didn't remove the `.clone()`. I think with `&str` the `.clone()` is not required. With the `-D` flag, we get this error:

```
error: using `clone` on a double-reference; this will copy the reference of type `&str` instead of cloning the inner type
Error:   --> src/widgets/scrollbar.rs:91:38
   |
91 |             buf.set_string(right, y, self.bar_symbol.clone(), self.bar_style);
   |                                      ^^^^^^^^^^^^^^^^^^^^^^^
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#clone_double_ref
   = note: `#[deny(clippy::clone_double_ref)]` on by default
help: try dereferencing it
   |
91 |             buf.set_string(right, y, &(*self.bar_symbol).clone(), self.bar_style);
   |                                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~
help: or try being explicit if you are sure, that you want to clone a reference
   |
91 |             buf.set_string(right, y, <&str>::clone(self.bar_symbol), self.bar_style);
   |                                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

src/widgets/scrollbar.rs
kdheepak: Do you mean writing it like this?

```
        const MIN_HEIGHT: u16 = 0;

        if area.height == MIN_HEIGHT {
            return;
        }
```

src/widgets/scrollbar.rs
```diff
@@ -0,0 +1,208 @@
...
+    symbols::{block::FULL, line::DOUBLE_VERTICAL},
+    widgets::Widget,
+};
+
+/// Scrollbar widget for tui-rs library.
+///
+/// This widget can be used to display a scrollbar in a terminal user interface.
+#[derive(Default, Debug, Clone)]
+pub struct Scrollbar<'a> {
+    scroll: f64,
```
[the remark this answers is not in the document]
kdheepak: I've updated the code to make it show the size relative to the "number of lines" that the scroll bar can scroll over.
Previously, `scroll` was a ratio of `position / num_elements`. Now, the widget requires a user to pass in both `position` and `num_elements` explicitly.

src/widgets/scrollbar.rs
kdheepak: This is not required with the new implementation using `position` and `num_elements`.

src/widgets/scrollbar.rs
```diff
@@ -0,0 +1,208 @@
...
+        // If the rightmost column is out of bounds, we don't draw the scrollbar.
+        if right < area.left() {
+            return;
+        };
+
+        // Calculate the top row and the height for the scrollbar.
+        let (top, height) = { (area.top() + 1, area.height.saturating_sub(2)) };
+
+        // Draw the bar of the scrollbar.
+        for y in top..(top + height) {
```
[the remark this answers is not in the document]
kdheepak: I've modified it to render to the entire width or height of the rectangle now.

src/widgets/scrollbar.rs
```diff
@@ -0,0 +1,208 @@
...
+
+        terminal
+            .draw(|f| {
+                let size = f.size();
+                f.render_widget(paragraph.clone(), size);
+                f.render_widget(scrollbar.clone(), size);
+            })
+            .unwrap();
+
+        terminal.backend().assert_buffer(&expected);
```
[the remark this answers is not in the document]
kdheepak: Thanks for the suggestion on how to do tests. I've updated all the tests here to use your suggestion.

src/widgets/scrollbar.rs
joshka: More try to explain why the min height is 2 in a constant - here's it's because we're subtracting the top and the bottom row and not rendering them. Something like
```rust
// subtract space for top and bottom
let Some(bar_height) = area.height.checked_sub(2) or else { return; };
```

I think this has disappeared in the current iteration though (which checks for 0 - which is obvious and fine).

src/widgets/scrollbar.rs
```diff
@@ -0,0 +1,208 @@
...
+
+impl<'a> Scrollbar<'a> {
+    /// Creates a new scrollbar with default styles and symbols.
+    pub fn new(scroll: f64) -> Self {
+        Self {
+            scroll,
+            style: Style::default(),
+            symbol: FULL,
+            bar_style: Style::default(),
+            bar_symbol: DOUBLE_VERTICAL,
```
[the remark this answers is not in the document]
joshka: The context on this one is consistency with the rest of the crate (e.g. the bar_sets idea).

src/widgets/scrollbar.rs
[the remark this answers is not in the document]
joshka: I think it could be useful to have multiple approaches to this (like the gauge does - it allows ratio, percent, etc.)

I'm not sure that I understand what num_elements represents just from the name:
1. The number of lines of the bar part of the scrollbar 
2. The total number of elements that are being scrolled (e.g. 1000 list items)
3. The total number of lines in the elements that are being scrolled (e.g. 1000 list items * 3 lines each)
4. The number of elements that are currently being shown (10 list items)
5. The number of lines that are currently being shown (30 lines)

src/widgets/scrollbar.rs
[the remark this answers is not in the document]
kdheepak: I've taken an initial stab at an interface with a `Set` in https://github.com/tui-rs-revival/ratatui/pull/228/commits/ab154507fc0d0dbc9ebdc63d10fa2fb08608136f. Let me know what you think.

src/widgets/scrollbar.rs
```diff
@@ -4,11 +4,54 @@ use crate::{
...
     widgets::Widget,
 };
 
+/// Scrollbar Set
+/// <--▮------->
+/// ^  ^   ^   ^
+/// │  │   │   └ end
+/// │  │   └──── track
+/// │  └──────── thumb
+/// └─────────── begin
```
joshka, line 18: This is a neat way to show this.

src/widgets/scrollbar.rs
```diff
@@ -67,35 +120,79 @@ impl<'a> Scrollbar<'a> {
...
+        self.thumb_style = style.clone();
+        self.begin_style = style.clone();
+        self.end_style = style.clone();
+        self
+    }
+
+    pub fn arrows(mut self, arrows: bool) -> Self {
+        self.arrows = arrows;
         self
     }
```
joshka: Do you think we could omit this and make start and end `Optional<&str>` where there's nothing rendered if they're set to none?

src/widgets/scrollbar.rs
```diff
@@ -0,0 +1,466 @@
...
+                vec![" ║", " █", " █", " ║"]
+            } else {
+                vec![" ║", " ║", " █", " █"]
+            };
+            assert_buffer_eq!(buffer, Buffer::with_lines(expected.clone()));
+        }
+    }
+
+    #[test]
+    fn render_scrollbar_two_size_horizontal() {
```
joshka, line 423: neat!

src/widgets/scrollbar.rs
kdheepak: Do we want to support arrows just at the beginning or just at the end, if one is set to `Some` and the other is set to `None`? The rendering logic gets more complicated in that case.

joshka (commented):
Nice overall

I'd still prefer to have the position part out of the scroll bar - based on the stuff in the other PR / design Issue, but this is really good. I'll take a full review sometime soon.

src/widgets/scrollbar.rs
joshka: Perhaps we could have an `arrows(start: &str, end: &str)` method?

kdheepak: I've added a version in https://github.com/tui-rs-revival/ratatui/pull/228/commits/aec39611820f4cbbea501fbcbad498444d2b3790 and https://github.com/tui-rs-revival/ratatui/pull/228/commits/e6a68954245ccf2eee1c9f7354dc195b77e0a49f that uses `Option<&str>` for the begin and end symbols.

src/widgets/scrollbar.rs
```diff
@@ -0,0 +1,208 @@
...
+            scroll,
+            style: Style::default(),
+            symbol: FULL,
+            bar_style: Style::default(),
+            bar_symbol: DOUBLE_VERTICAL,
+        }
+    }
+
+    /// Sets the scroll position of the scrollbar.
+    pub fn scroll(mut self, scroll: f64) -> Self {
```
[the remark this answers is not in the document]
kdheepak: I've not really worked with stateful widgets all that much, so I'm still trying to wrap my head around it. I'll have to review the code in your PR to get a sense of what is required here.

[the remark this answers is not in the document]
kdheepak: I took a crack at it in https://github.com/tui-rs-revival/ratatui/pull/228/commits/739693c1ca879f09d01bd3885d02e2821dc92e88 and https://github.com/tui-rs-revival/ratatui/pull/228/commits/7f410a6dd890435d0581ebd022859a623e24fb73.

Is a StatefulWidget mainly useful just to "separate concerns"? Are there other practical reasons one would want to choose a StatefulWidget over a stateless widget?

src/widgets/scrollbar.rs
```diff
@@ -0,0 +1,606 @@
...
+        Self {
+            orientation: ScrollbarOrientation::default(),
+            thumb_symbol: DOUBLE_VERTICAL.thumb,
+            thumb_style: Style::default(),
+            track_symbol: DOUBLE_VERTICAL.track,
+            track_style: Style::default(),
+            begin_symbol: None, // Some(DOUBLE_VERTICAL.begin),
+            begin_style: Style::default(),
+            end_symbol: None, // Some(DOUBLE_VERTICAL.end),
+            end_style: Style::default(),
```
kdheepak, line 158: Should arrows be enabled by default?

src/widgets/scrollbar.rs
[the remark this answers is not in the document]
joshka: I think mostly separating the state means that the widget can modify the state during rendering. That’s necessary if you want to track what’s visible.

imagine we just want to render a scroll bar that responds to arrow keys. We can manually track the scroll offset etc, and handle edge cases or just call scroll_state.up()/down() and just pass that to the scroll bar. All the logic about how many items there are etc is part of the state. The scroll bar then simply handles the rendering.

src/widgets/scrollbar.rs
kdheepak: I'm leaning toward enabling arrows by default. 

One reason is when using a scrollbar with the same size of the rect, the arrows take up the corners:

image

when is a better out of the box experience than this:

image

Without arrows, most users will want to write this:

```
        Scrollbar::default()
            .orientation(ScrollbarOrientation::Vertical)
            .arrows(false),
        chunks[0].inner(&Margin {
            vertical: 1,
            horizontal: 0,
        })
```

i.e. decrease the size of the scrollbar by 1 so that it looks nicer:

image

That is a less nice first experience for a user.

src/widgets/scrollbar.rs
kdheepak: I believe this comment has been addressed now.

src/widgets/scrollbar.rs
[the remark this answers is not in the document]
kdheepak: I believe this has been addressed now. I followed the same pattern as in bar::Set.

src/widgets/scrollbar.rs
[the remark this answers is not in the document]
kdheepak: I've updated to use a State. In this case, we only need one direction because currently we are not rendering a 2D scroll bar.

kdheepak:
I think this is ready for a final review now!

src/widgets/scrollbar.rs
joshka: +1 to arrows by default.

I'd be inclined to say that for the scrollbar to render as part of a paragraph that has borders, it should be inside the borders not on the borders (but also that this should be configurable - I like the idea of being able to use the border as a scroll bar like what you have here).

kdheepak: The scrollbar widget here doesn’t know anything about other widgets. 

Are you imaging that another widget like a  `Paragraph` would store an instance to a scrollbar and manipulate where it is being rendered?

kdheepak: Btw, I like the scrollbar in the border because it reuses space that isn’t useful (inside it may overwrap on text or eat into the width or height of a text area), and because it can be turned off and on without any reflow on content in the widget.

joshka: yup - I'm thinking about how existing widgets will use the scrollbar.
The would create one in their render() method.

kdheepak: That makes sense! Happy to add any other functions you think will improve the interface for other widgets or users.

sayanarijit:
Quick glance looks good.

src/widgets/scrollbar.rs
kdheepak: I'm assuming this can be done as a separate PR for all widgets, so I'll resolve this conversation for now.

joshka: yep

sayanarijit (approved):

src/widgets/scrollbar.rs
```diff
@@ -0,0 +1,698 @@
...
+        Scrollbar::default()
+            .arrows(false)
+            .orientation(ScrollbarOrientation::Horizontal)
+            .render(buffer.area, &mut buffer, &mut state);
+        let expected = vec!["    ", "══█═"];
+        assert_buffer_eq!(buffer, Buffer::with_lines(expected.clone()));
+    }
+
+    #[test]
+    fn render_scrollbar_two_size_double_vertical() {
```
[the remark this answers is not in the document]
kdheepak: I'm looking at all the test function names in this PR and they are not good at conveying the intent of the test.

I came across this article a few days ago:

https://bitfieldconsulting.com/golang/test-names

This recommends having test function names be full sentences. For example, in this case the test function name would be:

```rust
fn when_height_is_twice_content_length_then_thumb_size_is_length_two() {
```

Thoughts?

src/widgets/scrollbar.rs
```diff
@@ -0,0 +1,698 @@
...
+
+    /// Sets the orientation of the scrollbar.
+    /// Resets the symbols to [`DOUBLE_VERTICAL`] or [`DOUBLE_HORIZONTAL`] based on orientation
+    pub fn orientation(mut self, orientation: ScrollbarOrientation) -> Self {
+        self.orientation = orientation.clone();
+        self.symbol(match orientation {
+            ScrollbarOrientation::Vertical => DOUBLE_VERTICAL,
+            ScrollbarOrientation::Horizontal => DOUBLE_HORIZONTAL,
+        })
+    }
```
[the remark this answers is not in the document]
kdheepak: > `ScrollBar::default().orientation(Orientation::Horizontal) != ScrollBar::new(Orientation::Horizontal)`

Why do you say that this is the case? Aren't they identical?

I would say that majority of the users are going to want to just choose between a Vertical or Horizontal scrollbar, so we want to make that "path" easy to use. Which is why I think setting orientation should change the symbols used.

src/widgets/scrollbar.rs
```diff
@@ -0,0 +1,698 @@
...
+    pub orientation: ScrollbarOrientation,
+    pub thumb_style: Style,
+    pub thumb_symbol: &'a str,
+    pub track_style: Style,
+    pub track_symbol: &'a str,
+    pub begin_symbol: Option<&'a str>,
+    pub begin_style: Style,
+    pub end_symbol: Option<&'a str>,
+    pub end_style: Style,
+    pub mirror: bool,
```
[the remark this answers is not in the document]
kdheepak: You are right, they don't. 

Resolved in https://github.com/tui-rs-revival/ratatui/pull/228/commits/c26cd1fe14dac16858f99a4af76ed6b10b992362.

src/widgets/scrollbar.rs
```diff
@@ -0,0 +1,698 @@
...
+
+    pub fn symbol(mut self, symbol: Set) -> Self {
+        self.track_symbol = symbol.track;
+        self.thumb_symbol = symbol.thumb;
+        if self.begin_symbol.is_some() {
+            self.begin_symbol = Some(symbol.begin);
+        }
+        if self.end_symbol.is_some() {
+            self.end_symbol = Some(symbol.end);
+        }
```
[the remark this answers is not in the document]
kdheepak: I've added a few lines to the docstring. 

```
    /// Only sets begin_symbol and end_symbol if they are already contain a value.
    /// If begin_symbol and/or end_symbol were set to `None` explicitly, this function will respect
    /// that choice.
```

Hopefully that helps clarify it?

src/widgets/scrollbar.rs
```diff
@@ -0,0 +1,698 @@
...
+    pub fn arrows(mut self, arrows: bool) -> Self {
+        if arrows {
+            self = self.begin_symbol(Some(DOUBLE_VERTICAL.begin));
+            self = self.end_symbol(Some(DOUBLE_VERTICAL.end));
+        } else {
+            self = self.begin_symbol(None);
+            self = self.end_symbol(None);
+        }
+        self
+    }
```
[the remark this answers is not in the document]
kdheepak: I decided to remove the `arrows` method. 

So now, this will draw arrows:

```rust
            Scrollbar::default()
```

And this will not draw arrows:

```rust
            Scrollbar::default()
                .begin_symbol(None)
                .end_symbol(None)
```

src/widgets/scrollbar.rs
```diff
@@ -0,0 +1,698 @@
...
+
+    pub fn mirror(mut self, mirror: bool) -> Self {
+        self.mirror = mirror;
+        self
+    }
```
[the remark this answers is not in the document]
kdheepak: I've removed the `mirror` method in favor of the orientation enum capturing the location of the scrollbar.

src/widgets/scrollbar.rs
```diff
@@ -0,0 +1,698 @@
...
+}
+
+impl<'a> StatefulWidget for Scrollbar<'a> {
+    type State = ScrollbarState;
+
+    fn render(self, area: Rect, buf: &mut Buffer, state: &mut Self::State) {
```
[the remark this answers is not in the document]
kdheepak: Good idea. I've broken it down in this commit:

https://github.com/tui-rs-revival/ratatui/pull/228/commits/cd579d703c8eb38928b9ad7a0cd6ba316144f98e.

src/widgets/scrollbar.rs
```diff
@@ -0,0 +1,698 @@
...
+                Rect::new(area.x, area.y, area.width.saturating_sub(1), area.height)
+            }
+        } else {
+            area
+        };
+
+        // info!("{:?}", area);
+
+        // Define main size, cross position, and main position based on scrollbar orientation
+        let (main_size, main_position, cross_position, other_cross_position) =
```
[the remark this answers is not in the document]
kdheepak: I've added more documentation to the code to explain this.

```
        // `main` is the axis along which the scrollbar is being rendered.
        // `cross` is the other axis.
        //
        // For ScrollbarOrientation::VerticalRight
        //
        //        ┌───────────────────────── other_cross_position
        //        │               ┌───────── cross_position
        //        v               v
        //        ┌───────────────┐  ┌────── main_position
        //        │               ║<─┘ \
        //        │               █    }
        //  main  │               █    } ─── main_size
        //        │               ║    }
        //        │               ║    /
        //        └───────────────┘
        //               cross
        //
        // For ScrollbarOrientation::HorizontalBottom
        //
        //              main
        //        ┌───────────────┐<──────── other_cross_position
        //        │               │
        //  cross │               │
        //        │               │
        //        └═══███═════════┘<──────── cross_position
        //         ^
        //         │
        //         └──────────────────────── main_position
        //         \~~~~~~~~~~~~~/
        //                 └──────────────── main_size

```

src/widgets/scrollbar.rs
```diff
@@ -0,0 +1,698 @@
...
+            if is_vertical && self.mirror {
```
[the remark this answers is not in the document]
kdheepak: I moved it to a separate method and added a comment. Along with the new enum that captures the orientation, hopefully it'll be clearer.

src/widgets/scrollbar.rs
```diff
@@ -0,0 +1,698 @@
...
+            return;
+        }
+
+        // Bail on rendering if the cross dimension (horizontal or vertical)
+        // doesn't have room to draw.
+        if (self.mirror && cross_position >= other_cross_position)
+            || (!(self.mirror) && cross_position <= other_cross_position)
+        {
+            return;
+        }
```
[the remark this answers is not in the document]
kdheepak: I previously chose to not render the scrollbar vertically if the width == 1 but I think that is not needed. I removed it now and it will render a vertical scrollbar even if the width of the rect is 1.

src/widgets/scrollbar.rs
```diff
@@ -0,0 +1,698 @@
...
+
+        // Calculate the scroll ratio based on the current position over the length of
+        // scrollable content.
+        let scroll_position_ratio = (state.position as f64 / state.content_length as f64).min(1.0);
+
+        // Calculate the scrollbar size based on the length of scrollable content, using
+        // a proportion of the total area main size.
+        let scrollbar_size = (((main_size as f64 / state.content_length as f64) * main_size as f64)
+            .round() as u16)
+            .max(1);
```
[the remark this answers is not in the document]
kdheepak: I've refactored this into a new method.

src/widgets/scrollbar.rs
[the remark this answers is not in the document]
kdheepak: I've refactored all the test names so hopefully the tests will be more clearer now.

src/widgets/scrollbar.rs
```diff
@@ -0,0 +1,698 @@
...
+                vec!["        ", "◄═══██═►"]
+            } else {
+                vec!["        ", "◄════██►"]
+            };
+            assert_buffer_eq!(buffer, Buffer::with_lines(expected.clone()));
+        }
+    }
+
+    #[test]
+    fn render_scrollbar_position_horizontal_arrows_mirror() {
```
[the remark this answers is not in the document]
kdheepak: I made the ScrollbarOrientation like so:

```
/// Scrollbar Orientation
#[derive(Default, Debug, Clone)]
pub enum ScrollbarOrientation {
    #[default]
    VerticalRight,
    VerticalLeft,
    HorizontalBottom,
    HorizontalTop,
}
```

This removes the need for mirror as method and option.

src/widgets/scrollbar.rs
```diff
@@ -0,0 +1,698 @@
...
+                    area.bottom().saturating_sub(1),
+                    area.top(),
+                )
+            };
+
+        // Bail on rendering if the main dimension (vertical or horizontal)
+        // doesn't have room to draw, or if length of the contents to be scrolled is zero.
+        if main_size == 0 || state.content_length == 0 {
+            return;
+        }
```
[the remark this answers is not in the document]
kdheepak: I'm not sure I follow what you mean by this comment unfortunately. Can you clarify?

src/widgets/scrollbar.rs
```diff
@@ -0,0 +1,698 @@
...
+use super::StatefulWidget;
+
+/// Scrollbar Set
+/// <--▮------->
+/// ^  ^   ^   ^
+/// │  │   │   └ end
+/// │  │   └──── track
+/// │  └──────── thumb
+/// └─────────── begin
+#[derive(Debug, Clone)]
```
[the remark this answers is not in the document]
kdheepak: The current render function allows just a begin arrow or just a end arrow, and that is done by checking whether or not there's an `Some` or `None` for those symbols. But the `Set` stores all the values, so we'd have to store additional variables to check for those conditions. The current implementation seems cleaner to me.

kdheepak:
@joshka I've addressed a number of your comments and asked for some clarification on a new questions.

src/widgets/scrollbar.rs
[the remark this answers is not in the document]
joshka: Yep makes sense - I just misread the way the interaction between new and default worked.
I'm used to seeing new actually returning Self { ... } and assumed wrongly that it would be setting the orientation directly on the new struct, but it's actually creating it in default() and then setting the orientation with the builder method. This is fine I think.

joshka (changes requested):
Picking back up the conversation about the cross. The code doesn't make it clear how the cross (perhaps you mean corner?) is relevant to the the algorithm for rendering the scroll_bar.

If I'm thinking through this from basics, 200 lines to render a scrollbar seems excessive.
It's possible that there are some assumptions or non obvious edge cases that you have factored in here, but they are difficult to see. I'm not sure it would be easy to fix a problem in this code, or to work out how to add a new feature.

I think perhaps this works too much in absolute coordinates and worries about the vertical / horizontal distinction throughout rather than using a much simpler relative approach.

An algorithm that treats the rendering as rendering a line from 0 to bar length would look like:
```rust
// this is psuedo rust, please forgive any errors
let length = length_of_bar_without_arrows(); // e.g. either area.width or area.height -1 or -2 
let (start_track, end_track) = relative_positions_of_track_without_arrows(); // e.g. (0, length - 1) or (1, length -2);
let (start_thumb, end_thumb) = relative_positions_of_thumb(); // e.g. (3, 6)

// this can be a method... absolute_positions()
let (left, top, right, bottom) =
  (area.x, area.y, area.x + area.width - 1, area.y + area.height - 1);
let (start_x, start_y, end_x, end_y) = match position {
  Position::HorizontalTop => (left, top, right, top),
  Position::HorizontalBottom => (left, bottom, right, bottom),
  Position::VerticalLeft => (left, top, left, bottom),
  Position::VerticalRight => right, top, right, bottom),
};

index = 0;
for x in (start_x..end_x) { // either this line
  for y in (start_y..end_y) { // or this line will just be a single iteration
    if index < start_track { draw_start_arrow(x, y); }
    else if index < start_thumb { draw_track(x, y); }
    else if index < end_thumb { draw_thumb(x, y); }
    else if index < end_track { draw_track(x, y); }
    else { draw_end_arrow(x, y); }
    index++;
  }
}
```

If I have indeed missed some complexity that is non obvious here, I apologize.

src/widgets/scrollbar.rs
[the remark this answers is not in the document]
joshka: I've used a number of different conventions in various programming languages over the years. In the rust code I've read so far, there doesn't seem to be a wide consensus on naming (or in many cases doing the normal ActArrangeAssert style or GivenWhenThen style tests). I'd go with the idea that the name of the test should give a developer that sees it enough information to understand the intent of what is being tested. Test names help as a way of memorializing the design decisions, or by characterizing the behavior of edge cases.

If you take that approach, then a name like `thumb_size_is_proportional_to_content_length()` might be the descriptive part that captures the requirement / intent. Saying the parts about "twice" and "length two" are implementation details of that broader requirement. You can get that from reading the test, but you can't easily see why you care about height being twice content length without a test name, assertion message, or doc comment (or all three).

kdheepak:
I guess the cross is confusing naming. 

I made two changes based on your most recent comment:

1. I changed all uses for `size` to `start` and `end`. Previously I was using the `main_position` and `main_size` to render, now it uses `start` and `end` to render. Maybe that'll make it a little clearer?
2. I renamed all the variables to more accurately match the exact part of the scrollbar they are used for, i.e. `track`, `thumb`, etc. (What was previously `cross` is now `track_axis`, and `main_start` is now `track_start`. And there's a corresponding `track_end`. Similarly there's also a `thumb_start` and `thumb_end`.)

```
        //
        // For ScrollbarOrientation::VerticalRight
        //
        //                   ┌───────── track_axis  (x)
        //                   v
        //   ┌───────────────┐
        //   │               ║<──────── track_start (y1)
        //   │               █
        //   │               █
        //   │               ║
        //   │               ║<──────── track_end   (y2)
        //   └───────────────┘
        //
        // For ScrollbarOrientation::HorizontalBottom
        //
        //   ┌───────────────┐
        //   │               │
        //   │               │
        //   │               │
        //   └═══███═════════┘<──────── track_axis  (y)
        //    ^             ^
        //    │             └────────── track_end   (x2)
        //    │
        //    └──────────────────────── track_start (x1)
        //

```

examples/scrollbar.rs
mindoodoo, line 1: Imo it would be interesting to have the keybindings for the scrolling be displayed at the top of the example.

src/widgets/scrollbar.rs
```diff
@@ -0,0 +1,853 @@
...
+    VerticalRight,
+    VerticalLeft,
+    HorizontalBottom,
+    HorizontalTop,
+}
+
+/// Scrollbar widget for tui-rs library.
+///
+/// This widget can be used to display a scrollbar in a terminal user interface.
+/// The following components of the scrollbar are customization in symbol and style.
```
mindoodoo, line 157:
```suggestion
/// The following components of the scrollbar are customizable in symbol and style.
```

src/widgets/scrollbar.rs
```diff
@@ -0,0 +1,853 @@
...
+    /// ^  ^   ^   ^
+    /// │  │   │   └ end
+    /// │  │   └──── track
+    /// │  └──────── thumb
+    /// └─────────── begin
+    ///
+    /// Only sets begin_symbol and end_symbol if they are already contain a value.
+    /// If begin_symbol and/or end_symbol were set to `None` explicitly, this function will respect
+    /// that choice.
+    pub fn symbol(mut self, symbol: Set) -> Self {
```
mindoodoo, line 277:
```suggestion
    pub fn symbols(mut self, symbol: Set) -> Self {
```

src/widgets/scrollbar.rs
```diff
@@ -0,0 +1,853 @@
...
+    /// Sets the symbols used for the various parts of the scrollbar from a [`Set`].
+    ///
+    /// <--▮------->
+    /// ^  ^   ^   ^
+    /// │  │   │   └ end
+    /// │  │   └──── track
+    /// │  └──────── thumb
+    /// └─────────── begin
+    ///
+    /// Only sets begin_symbol and end_symbol if they are already contain a value.
```
mindoodoo, line 274:
```suggestion
    /// Only sets begin_symbol and end_symbol if they already contain a value.
```

src/widgets/scrollbar.rs
```diff
@@ -0,0 +1,853 @@
...
+        viewport_content_length: u16,
+        track_start: u16,
+        track_end: u16,
+    ) -> (u16, u16) {
+        let scroll_position_ratio = (state_position as f64 / state_content_length as f64).min(1.0);
+
+        let thumb_size = (((viewport_content_length as f64 / state_content_length as f64)
+            * (track_end - track_start) as f64)
+            .round() as u16)
+            .max(1);
```
mindoodoo, line 381:
```suggestion
        fn get_thumb_start_end(
        &self,
        state: &ScrollbarState,
        viewport_content_length: u16,
        track_start: u16,
        track_end: u16,
    ) -> (u16, u16) {
        let scroll_position_ratio = (state.position as f64 / state.content_length as f64).min(1.0);

        let thumb_size = (((viewport_content_length as f64 / state.content_length as f64)
            * (track_end - track_start) as f64)
            .round() as u16)
            .max(1);
```

Againt nit pick, but I'd simplify the arg list a tad bit

src/widgets/scrollbar.rs
```diff
@@ -0,0 +1,853 @@
...
+        let area = self.get_track_area(area);
+        let (track_start, track_end, track_axis) = self.get_track_start_end(area);
+
+        if self.should_not_render(track_start, track_end, state.content_length) {
+            return;
+        }
+
+        let (thumb_start, thumb_end) = self.get_thumb_start_end(
+            state.content_length,
+            state.position,
```
mindoodoo, line 433:
```suggestion
        let (thumb_start, thumb_end) = self.get_thumb_start_end(
            &state
```

(if the previous suggestion is applied)

mindoodoo (approved):
Awesome stuff guys, looks great !!

Only a couple nit picks, not blocking per say. Looking forward to seeing this in the next release ! :rocket:

src/widgets/scrollbar.rs
kdheepak: Resolved in https://github.com/tui-rs-revival/ratatui/pull/228/commits/61da1cc3ee57951f786cd67af8ef34ddf2273513.

src/widgets/scrollbar.rs
kdheepak: Resolved in https://github.com/tui-rs-revival/ratatui/pull/228/commits/2284b202bd71c2e1c0549aaa37a89276aeee2c42.

src/widgets/scrollbar.rs
kdheepak: Resolved in https://github.com/tui-rs-revival/ratatui/pull/228/commits/2284b202bd71c2e1c0549aaa37a89276aeee2c42.

examples/scrollbar.rs
kdheepak: I added the keybindings at the top of the example:

image

src/widgets/scrollbar.rs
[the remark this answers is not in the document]
kdheepak: I took another stab at better test names in https://github.com/tui-rs-revival/ratatui/pull/228/commits/0dc6770505dbdaff3f2b81d586207cbb970e10d2 and https://github.com/tui-rs-revival/ratatui/pull/228/commits/c436716621e969b7e299b00b0f23797e8277e60e.

joshka (approved):
LGTM

joshka:
To make this ready to merge, I'm going to push a rebase squash and cleanup the commit message

PR #231, opened 2023-06-09
feat(buffer): use 'smartstring' for the buffer cell for better cache coherence

Philipp-M:
This should optimize the buffer a little bit.
Currently all symbol lookups when iterating through the buffer are to the heap , it's better to keep the data close to each other for cache coherence. I have not benched this though, this just caught my eye while going through the source.
With this change, there will likely only be one heap lookup, when iterating through the buffer, only extremely exotic symbols will be on the heap I guess.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,7 @@ cargo-args = ["-Zunstable-options", "-Zrustdoc-scrape-examples"]
 
 [dependencies]
 bitflags = "2.3"
+smartstring = "1.0"
 cassowary = "0.3"
 crossterm = { version = "0.26", optional = true }
 indoc = "2.0"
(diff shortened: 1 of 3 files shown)


joshka:
+1 to benchmark before perf changes. They’re necessary to quantify the problem and show that the solution addresses that meaningfully.

also, can you add a bit more description on what this change brings - sort of a 101 for those that aren’t familiar with the problem smartstring solves.

Philipp-M:
`smartstring` should be inlined for very small strings (which should be the case for `Cell` with <= 23 bytes (https://fasterthanli.me/articles/small-strings-in-rust#the-smartstring-crate)) thus I thought, this may be a cheap optimization, but... I've just benched it (with the benchmarks of `tui-textarea`), and indeed all those benchmarks regressed by ~20-40%. Quite surprising, how well optimized the std types are.

Thanks for the links, I've overlooked that PR, the `TeenyString` implementation of wezterm looks quite interesting, might be better suited for that use case...

PR #239, opened 2023-06-11
chore(ci): bump cargo-make version

orhun:
See <https://github.com/sagiegurari/cargo-make/releases>


sayanarijit (approved):

joshka:
Perhaps replace with https://github.com/davidB/rust-cargo-make and just use latest.

Alternatively I think there is a way to reference the latest url on a github release.

joshka (approved):

PR #268, opened 2023-06-17
docs(lib): add `tui-term` a pseudoterminal library

a-kenji:
Add [`tui-term`](https://github.com/a-kenji/tui-term/tree/main) to the list of third party libraries. 

Is the capitalization of ratatui correct here?


orhun (approved):
Neat.

PR #269, opened 2023-06-17
chore(github): add pull request template

orhun:
Saw a similar thing in another project, I think having this is better than nothing.


mindoodoo (approved):
Makes sense

PR #270, opened 2023-06-18
refactor(text)!: remove the use of borrowed data and replace with owned data in all text types

lthoerner:
This PR greatly simplifies the handling of `Text`, `Line`, `Span`, and `StyledGrapheme` by refactoring the types to use owned data rather than borrowed data. This makes adding new functionality much more straightforward due to the tendency of lifetime parameters to cause difficult-to-track errors. It also rearranges the `text` module into a more consistent structure.

This has been regression tested and although it shows a clear tendency of decreased performance, @joshka has let me know that the performance impacts would be mostly insignificant in real-world scenarios.

In addition, this integrates changes from #193, which was cherry-picked into https://github.com/tui-rs-revival/ratatui/tree/feat-wrapping. I believe that these changes are related enough to warrant being merged together. #259 is a bigger undertaking so I think they should be kept separate.


endepointe:
What scenarios would this change impact performance? Is there a security aspect to consider?

Eyesonjune18:
> What scenarios would this change impact performance?

This impacts the performance of instantiating `Paragraph` widgets, along with the text types mentioned.

> Is there a security aspect to consider?

No.

haydenflinner:
Could we instead swap to Rc rather than directly owned, so that we remove lifetimes, but the underlying bytes of the string aren't copied around, but instead a small pointer to it? Seems like best of both worlds, and mirrors how higher-level languages generally treat "String passed/stored by value". I'm curious on the exact numbers performance-wise we're seeing as difference, but I guess I'm not that surprised that trying to copy around 26kb of data between frames may start to lag.

Eyesonjune18:
> Could we instead swap to Rc rather than directly owned, so that we remove lifetimes, but the underlying bytes of the string aren't copied around, but instead a small pointer to it? Seems like best of both worlds, and mirrors how higher-level languages generally treat "String passed/stored by value". I'm curious on the exact numbers performance-wise we're seeing as difference, but I guess I'm not that surprised that trying to copy around 26kb of data between frames may start to lag.

@joshka you may want to take a look at this, because using memory containers is not within my expertise :)

joshka:
I caught that you were working on this, but never quite saw the problem that you were solving. Can you break down the logic of how you got to needing this better?

"This makes adding new functionality much more straightforward due to the tendency of lifetime parameters to cause difficult-to-track errors. It also rearranges the text module into a more consistent structure."

Eyesonjune18:
> I caught that you were working on this, but never quite saw the problem that you were solving. Can you break down the logic of how you got to needing this better?

I was running into issues where methods that took `&self` (non-mut methods) were causing `self` to be moved (I believe this was with some unit tests I was writing for `Text`). I was unable to track down the lifetime mistake that caused this.

joshka:
P.S. there's a good article on the rationale for using `Cow<'a, str>` and `Into<Cow<'a, str>>` at https://jwilm.io/blog/from-str-to-cow/

mindoodoo:
Hey @Eyesonjune18, could you sum up the benchmark results for this change for everyone to see here ?

lthoerner:
I am going to close this for now - many of the lifetime issues have been resolved in my other code.

PR #271, opened 2023-06-18
chore(github): simplify the CODEOWNERS file

orhun:
Auto review-requests were not working so let's try this way.


joshka:
Ha  yep:
https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#codeowners-syntax
>If you want to match two or more code owners with the same pattern, all the code owners must be on the same line. If the code owners are not on the same line, the pattern matches only the last mentioned code owner.

joshka (approved):

PR #275, opened 2023-06-24
refactor(const): Mark low level fns const

a-kenji:
Mark functions of `structs` that suit themselves to be composed with multiple other widgets as const, namely:

- `Block`
- `Layout`
- `Rect`

Continuation of: #115

This allows with some limitations to share blocks, layouts, through many widgets and with some luck also to remove `lazy_static!` from some of those use cases.

### DIFF
--- a/src/layout.rs
+++ b/src/layout.rs
@@ -86,6 +86,12 @@ thread_local! {
 
 impl Default for Layout {
     fn default() -> Layout {
+        Layout::new()
+    }
+}
+
+impl Layout {
+    pub const fn new() -> Layout {
         Layout {
             direction: Direction::Vertical,
             margin: Margin {
@@ -96,9 +102,7 @@ impl Default for Layout {
             expand_to_fill: true,
         }
     }
-}
 
-impl Layout {
     pub fn constraints<C>(mut self, constraints: C) -> Layout
     where
         C: Into<Vec<Constraint>>,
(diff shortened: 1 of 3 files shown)


orhun (approved):

joshka (changes requested):
Can you please add unit tests? These help document the intent of this change and prevent the accidental removal of the const keyword later. These can be simple one liners that check that we can use these in a constant context. Something like:
```rust
fn layout_can_be_const() {
  const layout: Layout = Layout::default()
    .margin(1)
    .horizontal_margin(1)
    ...
// or 
  const layout: Layout = Layout::default().margin(1);
  const layout: Layout = Layout::default().horizontal_margin(1);
  ...
}
```

For a commit message that targets the changelog, this is not a refactor as it really enables some extra functionality (the ability to use the layout and block in a constant context. So perhaps:
```
feat(misc): allow layout and block to be used in constant context

Block can now be used like:
`const BORDER_BLOCK: Block = Block::default().borders(...`
Layout can now be used like:
`const MAIN_LAYOUT: Layout = Layout::default().margin(...`
```

a-kenji:
@joshka, thank you for the review!

> For a commit message that targets the changelog, this is not a refactor as it really enables some extra functionality (the ability to use the layout and block in a constant context.

Agreed! I went for consistency here initially, but I changed the commit message now.

I added unit tests in the vein you were suggesting.
I also added the same unit tests for the `const Style` changes, should I keep them in this commit or split them into it's own commit?

joshka (approved):

joshka:
I rebased on main and rewrote the commit message a little.
We've enabled the setting to verify commits using GPG, which blocks merging directly if you don't have that setup.
See https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification for details.

PR #276, opened 2023-06-25
feat(backend): add window size api

benjajaja:
Depends on https://github.com/crossterm-rs/crossterm/pull/790, and should update crossterm in an independent PR. Related to #215

For image (sixel, iTerm2, Kitty...) support that handles graphics in
terms of `Rect` so that the image area can be included in layouts.

For example: an image is loaded with a known pixel-size, and drawn, but
the image protocol has no mechanism of knowing the actual cell/character
area that been drawn on. It is then impossible to skip overdrawing the
area.

Returning the window size in pixel-width / pixel-height, together with
colums / rows, it can be possible to account the pixel size of each cell
/ character, and then known the `Rect` of a given image, and also resize
the image so that it fits exactly in a `Rect` and avoid artifacts.

image

### DIFF
--- a/src/backend/crossterm.rs
+++ b/src/backend/crossterm.rs
@@ -18,9 +18,10 @@ use crossterm::{
 };
 
 use crate::{
-    backend::{Backend, ClearType},
+    backend::{Backend, ClearType, WindowSize},
     buffer::Cell,
-    layout::Rect,
+    layout::Size,
+    prelude::Rect,
     style::{Color, Modifier},
 };
 
@@ -169,10 +170,24 @@ where
     }
 
     fn size(&self) -> io::Result<Rect> {
-        let (width, height) =
-            terminal::size().map_err(|e| io::Error::new(io::ErrorKind::Other, e.to_string()))?;
-
+        let (width, height) = terminal::size()?;
         Ok(Rect::new(0, 0, width, height))
+    }
+
+    fn window_size(&mut self) -> Result<WindowSize, io::Error> {
+        let crossterm::terminal::WindowSize {
+            columns,
+            rows,
+            width,
+            height,
+        } = terminal::window_size()?;
+        Ok(WindowSize {
+            columns_rows: Size {
+                width: columns,
+                height: rows,
+            },
+            pixels: Size { width, height },
+        })
     }
 
     fn flush(&mut self) -> io::Result<()> {
--- a/src/backend/mod.rs
+++ b/src/backend/mod.rs
@@ -29,7 +29,7 @@ use std::io;
 
 use strum::{Display, EnumString};
 
-use crate::{buffer::Cell, layout::Rect};
+use crate::{buffer::Cell, layout::Size, prelude::Rect};
 
 #[cfg(feature = "termion")]
 mod termion;
@@ -58,6 +58,18 @@ pub enum ClearType {
     BeforeCursor,
     CurrentLine,
     UntilNewLine,
+}
+
+/// The window sizes in columns,rows and optionally pixel width,height.
+pub struct WindowSize {
+    /// Size in character/cell columents,rows.
+    pub columns_rows: Size,
+    /// Size in pixel width,height.
+    ///
+    /// The `pixels` fields may not be implemented by all terminals and return `0,0`.
+    /// See https://man7.org/linux/man-pages/man4/tty_ioctl.4.html under section
+    /// "Get and set window size" / TIOCGWINSZ where the fields are commented as "unused".
+    pub pixels: Size,
 }
 
 /// The `Backend` trait provides an abstraction over different terminal libraries.
@@ -111,8 +123,15 @@ pub trait Backend {
         }
     }
 
-    /// Get the size of the terminal screen as a [`Rect`].
+    /// Get the size of the terminal screen in columns/rows as a [`Rect`].
     fn size(&self) -> Result<Rect, io::Error>;
+
+    /// Get the size of the terminal screen in columns/rows and pixels as [`WindowSize`].
+    ///
+    /// The reason for this not returning only the pixel size, given the redundancy with the
+    /// `size()` method, is that the underlying backends most likely get both values with one
+    /// syscall, and the user is also most likely to need columns,rows together with pixel size.
+    fn window_size(&mut self) -> Result<WindowSize, io::Error>;
 
     /// Flush any buffered content to the terminal screen.
     fn flush(&mut self) -> Result<(), io::Error>;
--- a/src/backend/termwiz.rs
+++ b/src/backend/termwiz.rs
@@ -11,13 +11,14 @@ use termwiz::{
     cell::{AttributeChange, Blink, Intensity, Underline},
     color::{AnsiColor, ColorAttribute, SrgbaTuple},
     surface::{Change, CursorVisibility, Position},
-    terminal::{buffered::BufferedTerminal, SystemTerminal, Terminal},
+    terminal::{buffered::BufferedTerminal, ScreenSize, SystemTerminal, Terminal},
 };
 
 use crate::{
-    backend::Backend,
+    backend::{Backend, WindowSize},
     buffer::Cell,
-    layout::Rect,
+    layout::Size,
+    prelude::Rect,
     style::{Color, Modifier},
 };
 
@@ -169,22 +170,31 @@ impl Backend for TermwizBackend {
     }
 
     fn size(&self) -> Result<Rect, io::Error> {
-        let (term_width, term_height) = self.buffered_terminal.dimensions();
-        let max = u16::max_value();
-        Ok(Rect::new(
-            0,
-            0,
-            if term_width > usize::from(max) {
-                max
-            } else {
-                term_width as u16
+        let (cols, rows) = self.buffered_terminal.dimensions();
+        Ok(Rect::new(0, 0, u16_max(cols), u16_max(rows)))
+    }
+
+    fn window_size(&mut self) -> Result<WindowSize, io::Error> {
+        let ScreenSize {
+            cols,
+            rows,
+            xpixel,
+            ypixel,
+        } = self
+            .buffered_terminal
+            .terminal()
+            .get_screen_size()
+            .map_err(|e| io::Error::new(io::ErrorKind::Other, e))?;
+        Ok(WindowSize {
+            columns_rows: Size {
+                width: u16_max(cols),
+                height: u16_max(rows),
             },
-            if term_height > usize::from(max) {
-                max
-            } else {
-                term_height as u16
+            pixels: Size {
+                width: u16_max(xpixel),
+                height: u16_max(ypixel),
             },
-        ))
+        })
     }
 
     fn flush(&mut self) -> Result<(), io::Error> {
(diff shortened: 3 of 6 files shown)


Cargo.toml
```diff
@@ -37,7 +37,8 @@ cargo-args = ["-Zunstable-options", "-Zrustdoc-scrape-examples"]
 [dependencies]
 bitflags = "2.3"
 cassowary = "0.3"
-crossterm = { version = "0.26", optional = true }
+# crossterm = { version = "0.26", optional = true }
+crossterm = { git = "https://github.com/benjajaja/crossterm", rev = "ad84a3829d55eac27cfdb7791b2e34c1732040ad", optional = true }
```
joshka, line 41: I think I'd prefer to wait until this is merged to merge this into ratatui. This is fine as a draft for now however.

You can also set this in a patch section and leave the main dependency in place. See https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#the-patch-section

```toml
[patch.crates-io]
crossterm = { git = "https://github.com/benjajaja/crossterm", rev = "ad84a3829d55eac27cfdb7791b2e34c1732040ad", optional = true }
```

examples/window_size.rs
```diff
@@ -0,0 +1,80 @@
+use std::{error::Error, io};
```
joshka, line 1: It could be worthwhile adding some doc comments to help a reader understand what this example shows.

src/backend/crossterm.rs
```diff
@@ -86,7 +86,7 @@ where
         for (x, y, cell) in content {
             // Move the cursor if the previous location was not (x - 1, y)
             if !matches!(last_pos, Some(p) if x == p.0 + 1 && y == p.1) {
-                map_error(queue!(self.buffer, MoveTo(x, y)))?;
+                queue!(self.buffer, MoveTo(x, y))?;
```
joshka, line 89: I see this is removed a bunch of times. Can you please find out or explain why this was necessary in the first place?

src/backend/mod.rs
```diff
@@ -112,6 +112,10 @@ pub trait Backend {
     /// Get the size of the terminal screen as a [`Rect`].
     fn size(&self) -> Result<Rect, io::Error>;
 
+    /// Get the size of the terminal screen as well as windows sixe in pixels as a [`Rect`] and a
+    /// `(u16, u16)` tuple.
+    fn window_size(&mut self) -> Result<(Rect, (u16, u16)), io::Error>;
```
joshka, line 117: Do you think it makes sense to split this to two methods with more explicit names instead of a single returning a tuple? Is the Rect ever not positioned at x=0, y=0?

src/backend/termwiz.rs
```diff
@@ -221,3 +220,18 @@ impl From<Color> for ColorAttribute {
...
+}
+
+fn u16_max(i: usize) -> u16 {
+    let max = u16::max_value();
+    if i > usize::from(max) {
+        max
+    } else {
+        i as u16
+    }
+}
```
joshka, line 237: Replace with `u16::try_from(i).unwrap_or(u16::MAX);` (probably obvious enough to just inline into the methods that use it)

src/backend/test.rs
```diff
@@ -177,6 +177,10 @@ impl Backend for TestBackend {
         Ok(Rect::new(0, 0, self.width, self.height))
     }
 
+    fn window_size(&mut self) -> Result<(Rect, (u16, u16)), io::Error> {
+        Ok((Rect::new(0, 0, self.width, self.height), (7, 12)))
```
joshka, line 181: What's the significance of 7 and 12 here? Can you document it?

joshka (changes requested):
Rewrite the commit message title in the convention commit style using imperative tone (e.g. `feat(backend): add window_size_in_pixels() method`).

This is pretty close to being mergable (or will be once the crossterm PR is merged), so can you please rebase and squash your commits into a single commit and setup commit signing? See https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits for info.

joshka:
Backend is woefully under-tested. I wonder if there's some way to fix that.

Cargo.toml
benjajaja: Oooh TIL, that's so cool, thank you!

examples/window_size.rs
benjajaja: I removed this in the end, as it's not a valuable demo/example.

src/backend/crossterm.rs
benjajaja: It seems like the `queue` macro, or `QueueableCommand::queue`, now return `io::Result` making the `map_error` unnecessary.

src/backend/mod.rs
benjajaja: I say no, because 2 out of 3 implementations return both things at once in one call, due to the underlying API.

src/backend/termwiz.rs
benjajaja: Much better, thank you!

src/backend/test.rs
benjajaja: It was from when it was `font_size`, not `window_size` - my bad.
Updated to a better value and added a comment.

src/backend/mod.rs
benjajaja: Oh and the `Rect` is just like `size`, always x=0,y=0.

benjajaja:
Regarding tests, I don't think it's really feasible to cover this. `window_size()` really depends on being run as a graphical application, i.e. xorg or wayland.

joshka (changes requested):
I think there was a mention of this not yet working with windows on crossterm if I read the linked issue right. Can you make sure to document this in the crossterm docs for the window_size method?

src/backend/mod.rs
```diff
@@ -112,6 +112,9 @@ pub trait Backend {
     /// Get the size of the terminal screen as a [`Rect`].
     fn size(&self) -> Result<Rect, io::Error>;
 
+    /// Get the font size of the terminal screen as [`(width, height)`].
+    fn window_size(&mut self) -> Result<(Rect, (u16, u16)), io::Error>;
```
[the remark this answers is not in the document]
joshka: I've been thinking about this a little - I think there's a few places within various widgets where we care more about the particular offset that a size. A good example of that `Buffer::empty()` which takes a Rect, but really only needs that to be a Size. We use this a lot for tests. So we might treat a Rect as a Point(x,y) and a Size(width, height).

It's probably out of scope for this PR to do all that however. PixelSize is still a bit ambiguous (I could see users getting confused over whether it is the size of a cell in pixels or the size of a window?)

What about:
```rust
struct WindowSize {
  size_in_rows_and_cols: Size,
  size_in_pixels: Size,
}
```

> Changing the return type of size() would also mean that this is a breaking change. Perhaps it would be better to have window_size() just return (u16, u16) as you mentioned? What if that was named `size_in_pixels()`?

joshka:
BTW, another PR upgraded crossterm 0.27 and brought in the changes to the error bits. Can you rebase this on main please?

src/terminal.rs
```diff
@@ -420,7 +420,7 @@ where
         let height = height.min(self.last_known_size.height);
         self.backend.append_lines(height)?;
         let missing_lines =
-            height.saturating_sub(self.last_known_size.bottom() - self.viewport_area.top());
+            height.saturating_sub(self.last_known_size.height - self.viewport_area.top());
```
benjajaja, line 423: `last_known_size` was a `Rect` that was only set from `with_options()` and `resize()`. Both did only have non-zero x,y for `Viewport::Fixed`. This method only runs for `Viewport::Inline(_)` and it is the only method that uses the field.

benjajaja:
I tried adding some test coverage for `insert_before`, but that did require to implement `clear_region()` and `append_line()` for `TestBackend` for which I don't know exactly how to implement the different "clear types", e.g. `AfterCursor` etc. In general, it seemed a bit pointless to test `insert_before` against `TestBackend`.

src/backend/mod.rs
```diff
@@ -109,8 +114,16 @@ pub trait Backend {
         }
     }
 
-    /// Get the size of the terminal screen as a [`Rect`].
-    fn size(&self) -> Result<Rect, io::Error>;
+    /// Get the size of the terminal screen in columns/rows as a [`Rect`].
+    fn size(&self) -> Result<Size, io::Error>;
```
joshka, line 118: How about leaving this as Rect and just updating the window_size method?

otherwise nit: doc comment => Size instead of Rect

src/layout.rs
```diff
@@ -587,6 +587,41 @@ fn try_split(area: Rect, layout: &Layout) -> Result<Rc<[Rect]>, AddConstraintErr
...
+/// A simple size struct
+#[derive(Debug, Default, Clone, Copy, Eq, PartialEq, Hash)]
+pub struct Size {
+    pub width: u16,
+    pub height: u16,
+}
+
+impl Size {
+    /// Creates a new size, with width and height limited to keep the area under max u16.
+    /// If clipped, aspect ratio will be preserved.
```
joshka, line 599: Do we actually need this truncation for anything?

joshka (changes requested):
I think perhaps make this non-breaking with the existing size() and resize() continuing to take Rects? We can come back to that decision at a later date however.

Aiming for a 0.23.0 release after the weekend, so this should be in that if you can make the changes in time.

benjajaja:
@joshka I toned this down to only adding the new `window_resize()` method, still introducing the new `Size` struct but not using it in src/terminal.rs so no breaking changes. Also dropped the truncation in `Size` copied from `Rect`, as its main purpose is to keep `area` (and maybe indirectly `Buffer::content`) in bounds, and there are no more conversions between `Size` and `Rect` now.

It's cool if we can get this into the release, but it's not critical like the cell skipping, as the window pixel size can be queried outside of `ratatui`.

--------------
Some thoughts on `Size` vs `Rect` in `Terminal`:

In `Terminal`, there is a `last_known_size` that may have non-zero x,y if the `viewport` is `Viewport::Fixed(Rect)`. However it seems like the field is used exclusively by `Viewport::Inline(u16)`.

In general, I think it could be healthy to cut down a few public methods in `Terminal`: `autoresize`, `resize`, `size` (is just a wrapper to `backend.size()`), and `area` of `CompletedFrame` returned by `draw`.

src/backend/mod.rs
```diff
@@ -60,6 +60,11 @@ pub enum ClearType {
     UntilNewLine,
 }
 
+pub struct WindowSize {
+    pub columns_rows: Size,
+    pub window_pixels: Size,
```
joshka, line 65: nit: what do you think about naming this just `pixels` given that the struct is named **`Window`**`Size`?

joshka (approved):
This looks good now. Thanks again. This seems like something we could merge before release. I'll flip it to the other maintainers to take a look.

Comments are non-blocking

src/backend/mod.rs
kdheepak: I think there should be a doc string here too, right? Maybe describing when someone might want to use this.

mindoodoo (approved):
Looks good ! Thanks for this !

I'll have to take a look at the image crate soon ! Exciting stuff.

@joshka what do you think about the scope of the conventional commit PR title ? let me know if anything else would be more suited. Thanks for all the review on this !

joshka:
> @joshka what do you think about the scope of the conventional commit PR title ? let me know if anything else would be more suited. Thanks for all the review on this !

I think it's probably ok as is. Merging this now. Thanks again @benjajaja. This is going to lead to some fun apps.

PR #277, opened 2023-06-25
docs: improve CONTRIBUTING.md

joshka:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,69 +1,149 @@
-# Fork Status
+# Contribution guidelines
 
-## Pull Requests
+First off, thank you for considering contributing to Ratatui.
 
-**All** pull requests opened on the original repository have been imported. We'll be going through any open PRs in a timely manner, starting with the **smallest bug fixes and README updates**. If you have an open PR make sure to let us know about it on our [discord](https://discord.gg/pMCEU9hNEj) as it helps to know you are still active.
+If your contribution is not straightforward, please first discuss the change you wish to make by
+creating a new issue before making the change, or starting a discussion on
+[discord](https://discord.gg/pMCEU9hNEj).
 
-## Issues
+## Reporting issues
 
-We have been unsuccessful in importing all issues opened on the previous repository.
-For that reason, anyone wanting to **work on or discuss** an issue will have to follow the following workflow :
+Before reporting an issue on the [issue tracker](https://github.com/tui-rs-revival/ratatui/issues),
+please check that it has not already been reported by searching for some related keywords. Please
+also check [`tui-rs` issues](https://github.com/fdehau/tui-rs/issues/) and link any related issues
+found.
 
-- Recreate the issue
-- Start by referencing the **original issue**: ```Referencing issue #[<issue number>](<original issue link>)```
-- Then, paste the original issues **opening** text
+## Pull requests
 
-You can then resume the conversation by replying to this new issue you have created.
+All contributions are obviously welcome. Please include as many details as possible in your PR
+description to help the reviewer (follow the provided template). Make sure to highlight changes
+which may need additional attention or you are uncertain about. Any idea with a large scale impact
+on the crate or its users should ideally be discussed in a "Feature Request" issue beforehand.
 
-### Closing Issues
+### Keep PRs small, intentional and focused
 
-If you close an issue that you have "imported" to this fork, please make sure that you add the issue to the **CLOSED_ISSUES.md**. This will enable us to keep track of which issues have been closed from the original repo, in case we are able to have the original repository transferred.
+Try to do one pull request per change. The time taken to review a PR grows exponential with the size
+of the change. Small focused PRs will generally be much more faster to review. PRs that include both
+refactoring (or reformatting) with actual changes are more difficult to review as every line of the
+change becomes a place where a bug may have been introduced. Consider splitting refactoring /
+reformatting changes into a separate PR from those that make a behavioral change, as the tests help
+guarantee that the behavior is unchanged.
 
-# Contributing
+### Search `tui-rs` for similar work
+
+The original fork of Ratatui, [`tui-rs`](https://github.com/fdehau/tui-rs/), has a large amount of
+history of the project. Please search, read, link, and summarize any relevant
+[issues](https://github.com/fdehau/tui-rs/issues/),
+[discussions](https://github.com/fdehau/tui-rs/discussions/) and [pull
+requests](https://github.com/fdehau/tui-rs/pulls).
+
+### Use conventional commits
+
+We use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) and check for them as
+a lint build step. To help adhere to the format, we recommend to install
+[Commitizen](https://commitizen-tools.github.io/commitizen/). By using this tool you automatically
+follow the configuration defined in [.cz.toml](.cz.toml). Your commit messages should have enough
+information to help someone reading the [CHANGELOG](./CHANGELOG.md) understand what is new just from
+the title. The summary helps expand on that to provide information that helps provide more context,
+describes the nature of the problem that the commit is solving and any unintuitive effects of the
+change. It's rare that code changes can easily communicate intent, so make sure this is clearly
+documented.
+
+### Clean up your commits
+
+The final version of your PR that will be committed to the repository should be rebased and tested
+against main. Every commit will end up as a line in the changelog, so please squash commits that are
+only formatting or incremental fixes to things brought up as part of the PR review. Aim for a single
+commit (unless there is a strong reason to stack the commits). See [Git Best Practices - On Sausage
+Making](https://sethrobertson.github.io/GitBestPractices/#sausage) for more on this.
+
+### Run CI tests before pushing a PR
+
+We're using [cargo-husky](https://github.com/rhysd/cargo-husky) to automatically run git hooks,
+which will run `cargo make ci` before each push. To initialize the hook  run `cargo test`. If
+`cargo-make` is not installed, it will provide instructions to install it for you. This will ensure
+that your code is formatted, compiles and passes all tests before you push. If you need to skip this
+check, you can use `git push --no-verify`.
+
+### Sign your commits
+
+We use commit signature verification, which will block commits from being merged via the UI unless
+they are signed. To set up your machine to sign commits, see [managing commit signature
+verification](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification)
+in GitHub docs.
 
 ## Implementation Guidelines
 
+### Setup
+
+Clone the repo and build it using [cargo-make](https://sagiegurari.github.io/cargo-make/)
+
+Ratatui is an ordinary Rust project where common tasks are managed with
+[cargo-make](https://github.com/sagiegurari/cargo-make/). It wraps common `cargo` commands with sane
+defaults depending on your platform of choice. Building the project should be as easy as running
+`cargo make build`.
+
+```shell
+git clone https://github.com/tui-rs-revival/ratatui.git
+cd ratatui
+cargo make build
+```
+
+### Tests
+
+The [test coverage](https://app.codecov.io/gh/tui-rs-revival/ratatui) of the crate is reasonably
+good, but this can always be improved. Focus on keeping the tests simple and obvious and write unit
+tests for all new or modified code. Beside the usual doc and unit tests, one of the most valuable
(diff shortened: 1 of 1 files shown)


joshka:
Note - there is intentional repetition in this doc. It's intended as people often only read the section that they care about.

orhun (approved):

CONTRIBUTING.md
```diff
@@ -1,69 +1,130 @@
...
 
-### Use of unsafe for optimization purposes
+### Setup
 
-**Do not** use unsafe to achieve better performances. This is subject to change, [see.](https://github.com/tui-rs-revival/tui-rs-revival/discussions/66)
-The only exception to this rule is if it's to fix **reproducible slowness.**
+Clone the repo and build it using [cargo-make](https://sagiegurari.github.io/cargo-make/)
 
-## Building
+`ratatui` is an ordinary Rust project where common tasks are managed with [cargo-make]. It wraps
```
orhun, line 64: Oops, can you add a link to the `cargo-make` here?

joshka: Done

joshka:
Fixed link to cargo-make and typos

CONTRIBUTING.md
```diff
@@ -1,69 +1,138 @@
...
+which may need additional attention or you are uncertain about. Any idea with a large scale impact
+on the crate or its users should ideally be discussed in a "Feature Request" issue beforehand.
+
+### Keep PRs small, intentional and focused
+
+Try to do one pull request per change. The time taken to review a PR grows exponential with the size
+of the change. Small focused PRs will generally be much more faster to approved and merge. Don't
+include large refactoring or formatting changes in PRs
+
+### Search tui-rs for similar work
```
mindoodoo, line 28: nit: we could link to the og repo here

CONTRIBUTING.md
```diff
@@ -1,69 +1,138 @@
-# Fork Status
+# Contribution guidelines
 
-## Pull Requests
+First off, thank you for considering contributing to ratatui.
```
Eyesonjune18, line 3: Capitalize Ratatui or surround it in single backticks, whichever is most consistent with the rest of the documentation/other written materials.

CONTRIBUTING.md
```diff
@@ -1,69 +1,138 @@
...
 
-**All** pull requests opened on the original repository have been imported. We'll be going through any open PRs in a timely manner, starting with the **smallest bug fixes and README updates**. If you have an open PR make sure to let us know about it on our [discord](https://discord.gg/pMCEU9hNEj) as it helps to know you are still active.
+If your contribution is not straightforward, please first discuss the change you wish to make by
+creating a new issue before making the change.
```
Eyesonjune18, line 6: A discussion may be more appropriate in some cases.

CONTRIBUTING.md
```diff
@@ -1,69 +1,138 @@
...
+All contributions are obviously welcome. Please include as many details as possible in your PR
+description to help the reviewer (follow the provided template). Make sure to highlight changes
+which may need additional attention or you are uncertain about. Any idea with a large scale impact
+on the crate or its users should ideally be discussed in a "Feature Request" issue beforehand.
+
+### Keep PRs small, intentional and focused
+
+Try to do one pull request per change. The time taken to review a PR grows exponential with the size
+of the change. Small focused PRs will generally be much more faster to approved and merge. Don't
+include large refactoring or formatting changes in PRs
```
Eyesonjune18, line 26: I'm not entirely sure what this means. There are, of course, many times where large refactors may be necessary. Formatting maybe not so much.

CONTRIBUTING.md
```diff
@@ -1,69 +1,138 @@
...
+
+### Search tui-rs for similar work
 
-### Closing Issues
+Tui-rs has a large amount of history in issues, discussions and PRs that is relevant to this fork.
```
Eyesonjune18, line 30: Should we make this Tui-Rs, Tui-RS, TUI-RS?

CONTRIBUTING.md
```diff
@@ -1,69 +1,138 @@
...
+Please make sure to search and link any previous issues / discussions related to your PR.
```
Eyesonjune18, line 31: This is somewhat redundant because of issue searching having been mentioned in relation to creating new issues. I would consolidate them into one sentence.

CONTRIBUTING.md
```diff
@@ -1,69 +1,138 @@
...
+follow the configuration defined in [.cz.toml](.cz.toml). Your commit messages should have enough
+information to help someone reading the [CHANGELOG](./CHANGELOG.md) understand what is new just from
+the title. The summary helps expand on that to provide information that helps provide more context,
+describes the nature of the problem that the commit is solving and any unintuitive effects of the
+change. It's rare that code changes can easily communicate intent, so make sure this is clearly
+documented here.
+
+When the PR is close to approval, please ensure that it is rebased against main, squashed into an
+appropriate amount of commits (usually 1). We don't need to have a commit for every small change
+that was made during the PR process. Use `git push --force` to update the branch associated with
```
Eyesonjune18, line 47: This needs context. What can you do with force push? Probably
- Amend commits incrementally if they do not need to be separate (`--amend --no-edit`)
- Squash commits, or otherwise edit them (reword etc.)

CONTRIBUTING.md
```diff
@@ -1,69 +1,138 @@
...
+We're using [cargo-husky](https://github.com/rhysd/cargo-husky) to automatically run git hooks,
+which will run `cargo make ci` before each push. To initialize the hook  run `cargo test`. If
+`cargo-make` is not installed, it will provide instructions to install it for you. This will ensure
+that your code is formatted, compiles and passes all tests before you push. If you need to skip this
+check, you can use `git push --no-verify`.
+
+### Sign your commits
+
+We use commit signature verification, which blocks your commits from being merged via the UI unless
+you have setup your machine to sign commits. See
```
Eyesonjune18, line 61: set up

CONTRIBUTING.md
```diff
@@ -1,69 +1,138 @@
...
 
-### Use of unsafe for optimization purposes
+### Setup
 
-**Do not** use unsafe to achieve better performances. This is subject to change, [see.](https://github.com/tui-rs-revival/tui-rs-revival/discussions/66)
-The only exception to this rule is if it's to fix **reproducible slowness.**
+Clone the repo and build it using [cargo-make](https://sagiegurari.github.io/cargo-make/)
 
-## Building
+`ratatui` is an ordinary Rust project where common tasks are managed with
```
Eyesonjune18, line 71: Needs consistent formatting for library name.

CONTRIBUTING.md
```diff
@@ -1,69 +1,138 @@
...
+tests for all new or modified code. Beside the usual doc and unit tests, one of the most valuable
+test you can write for `ratatui` is a test against the `TestBackend`. It allows you to assert the
+content of the output buffer that would have been flushed to the terminal after a given draw call.
+See `widgets_block_renders` in [tests/widgets_block.rs](./tests/widget_block.rs) for an example.
 
-All contributions are obviously welcome.
-Please include as many details as possible in your PR description to help the reviewer (follow the provided template).
-Make sure to highlight changes which may need additional attention or you are uncertain about.
-Any idea with a large scale impact on the crate or its users should ideally be discussed in a "Feature Request" issue beforehand.
+Prefer unit tests (and doc tests) in the code file over integration tests.
```
Eyesonjune18, line 91: Not a syntactically valid sentence :P

CONTRIBUTING.md
```diff
@@ -1,69 +1,138 @@
...
 
-## Committing
+If an area that you're making a change in is not tested, write tests to characterize the existing
+behavior before changing it. This helps ensure that we don't introduce bugs to existing software
+using ratatui (and helps make it easy to migrate apps still using tui-rs).
```
Eyesonjune18, line 95: Needs consistent case convention for lib mention.

CONTRIBUTING.md
```diff
@@ -1,69 +1,138 @@
...
 
-To avoid any issues that may arrise with the CI/CD by not following the [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/) syntax, we recommend to install [Commitizen](https://commitizen-tools.github.io/commitizen/).\
-By using this tool you automatically follow the configuration defined in [.cz.toml](.cz.toml).
+### Use of unsafe for optimization purposes
 
-Additionally, we're using [cargo-husky](https://github.com/rhysd/cargo-husky) to automatically load pre-push hook, which will run `cargo make ci` before each push. It will load the hook automatically when you run `cargo test`. If `cargo-make` is not installed, it will install it for you.\
-This will ensure that your code is formatted, compiles and passes all tests before you push. If you want to skip this check, you can use `git push --no-verify`.
+**Do not** use unsafe to achieve better performances. This is [subject to
+change](https://github.com/tui-rs-revival/tui-rs-revival/discussions/66) The only exception to this
+rule is if it's to fix **reproducible slowness.**
```
Eyesonjune18, line 101: Need more info on what reproducible slowness means.

Eyesonjune18 (changes requested):
Overall a great writeup.

Eyesonjune18 (commented):
Overall a great writeup.

CONTRIBUTING.md
joshka: There's a button to convert an issue to a discussion, but not the other way around (you can create a new issue that references a discussion however). There's also a stronger linkage in the PR list to issues. I think maybe I'll rewrite this to be issue / discussion / discord chat.

CONTRIBUTING.md
joshka: When large refactoring (changes that don't alter the external behavior of some code) and actual changes that alter behavior, it can be difficult to understand which changes are relevant to the PR vs just cleanups. The chances of detecting an error are lower.

Evaluating a refactor is:
- is it easier to read / maintain
- does it do the same thing (i.e. behavior was fully covered by tests and tests continue to pass)

Evaluating a change is:
- does it do the things it says were changed
- does it have tests that have also changed / been added to cover the new / changed feature

re: formatting, something which happens often is that people will make some changes, but reformat everything around the area that they've read to make the change. Again, this makes it difficult to actually see the diff of what's changed and evaluate it. Instead this means that the reviewer has to read every line of code to understand it in full again.

I'll reword it a bit to make some of that more clear.

CONTRIBUTING.md
joshka: probably `tui-rs` to match the crate / repo casing, but I generally capitalize sentences like this. I'm not too concerned either way on this.

CONTRIBUTING.md
joshka: I probably need to link to some other doc here. But yep, context would make sense here (iow, why am I asking for this)

CONTRIBUTING.md
joshka: It actually is syntactically correct (imperative mood form of "prefer" with an implied but missing -"to create"), but I'll reword this to something like "When writing tests, generally prefer to unit and doc tests within the file that is being tested rather than integration tests in the `tests/` folder"

CONTRIBUTING.md
joshka: My guess is that we can just drop this and say something about discussing the issue instead.

CONTRIBUTING.md
mindoodoo: I don't think we should be going this in depth on how to do any of those git operations, seems out of scope for a contributing guide. Additionally, we haven't ever done anything else than squash and merge so even if we are faced with a large number of commits they always get squashed, so imho I think we should drop everything after the first sentence. It would also help to not overcomplicate the contributing guide.

joshka: The problem I'm trying to solve here is a communication one. We use commit messages as our change log, which means that a commit message should be written with the library user in mind. It's the best place to capture the intent of the change and a way of communicating the change to someone so that they don't have to read the entire source code to work out how something works or find the changes docs.

When we accept PRs that have a bunch of fixes to little things found as part of the review it's just as important to communicate the end state as it was when we started. But things like "fixed comments", "cargo fmt", etc. are noise that distracts from that.

We don't necessarily need to document the steps in the guide, but perhaps instead something like try to aim for a single commit that has a commit message that documents your new feature (and then linking to some other docs).

It's my experience that some of the weirdest annoying bugs or legacy "can't change that because we don't know why it's there" things are strongly associated with messy commits where care was not taken to remove these sorts of things.

Anyway, I'll reword this to try to make some of that clear and find some good external docs to link to so our doc is smaller.

joshka:
All mentioned feedback incorporated.

orhun (approved):

Issue #261, opened 2023-06-15
Update the terminal type to be better at hiding the backend choice


joshka:
## Problem
To write apps that use crossterm / termion / termwiz, each crate has different methods to clear the screen, enable raw mode, etc. This means that most code that uses a terminal calls methods directly on the underlying library during setup / restore.

e.g.:
https://github.com/tui-rs-revival/ratatui/blob/0bf6af17e7d72b959b2776ca66a9c9986388d2df/examples/demo/crossterm.rs#L20-L25
https://github.com/tui-rs-revival/ratatui/blob/0bf6af17e7d72b959b2776ca66a9c9986388d2df/examples/demo/termion.rs#L17-L25
https://github.com/tui-rs-revival/ratatui/blob/0bf6af17e7d72b959b2776ca66a9c9986388d2df/examples/demo/termwiz.rs#L13-L15

It would be nice to be able to be able to switch implementations in end user apps using only a feature flag (and effectively hide these differences).

## Solution

Update the terminal type in ratatui so that apps built using ratatui are agnostic to the choice of terminal crate. One approach to is already done by @veeso in tui-realm at https://github.com/veeso/tui-realm/tree/main/src/adapter which adds the various crossterm and termion (but not termwiz) methods that are missing in the terminal type as adapters / a bridge pattern.

I think we should do something inspired by the pattern used there, but we can improve on it by pulling the terminal methods into the backend interface.

Each Backend however is still a little dfferent to construct. It would be really nice to be able to write:
```rust
let terminal = Terminal::on(io::std_err())?;
terminal.enable_mouse_capture()?;
terminal.enter_alternate_screen()?;
terminal.enable_raw_mode()?;
terminal.clear()?;
```

Additionally we should handle restoring both manually, and automatically by implementing the `Drop` trait e.g.
```rust
terminal.disable_mouse_capture()?;
terminal.leave_alternate_screen()?;
terminal.disable_raw_mode()?;

// or just drop the terminal variable (either explicitly or automatically on going out of scope)
// the terminal should cleans up the alternate screen and raw mode back to how they started.
drop(terminal);
```

Something that tui-realm also does is makes the Terminal and Frame types as newtypes to hide the Backend:
https://github.com/veeso/tui-realm/blob/547728dd3541571df6d741adc96d2617c0adc47a/src/adapter/crossterm/mod.rs#L20-L24
```rust
/// Frame represents the Frame where the view will be displayed in
[...]

PR #280, opened 2023-06-26
feat: simplify backend and terminal creation

joshka:
This PR simplifies the creation of backends and terminals by
implementing `Default` for the backends and adding builder style
methods to the terminal.

For example, the following code now creates a terminal that enters the
alternate screen and enables raw mode and cleans up after itself when
dropped.

```rust
use anyhow::{Context, Result};
use crossterm::event::{self, Event, KeyCode};
use ratatui::{prelude::*};

fn main() -> Result<()> {
    run()?;
    println!("This is on the main screen!");
    Ok(())
}

fn run() -> Result() {
    let backend = CrosstermBackend::on_stdout()?; 
    loop { 
        terminal.draw(|frame| {
            frame.render_widget(
                Paragraph::new("this is on the alternate screen"),
                frame.size(),
            );
        })?; 
            if let KeyCode::Char('q') = key.code { 
                return Ok(()); 
            } 
        } 
    } 
}
```

CrosstermBackend and TermionBackend both gain `on_stdout()` and
`on_stderr()` methods which enable raw mode and switch to the alternate
screen before returning, and builder style methods to enable or disable
raw mode and the alternate screen.

This commit implements `Drop` for the CrosstermBackend which will
disable raw mode and switch back to the main screen. This ensures the
terminal is left in a usable state if the application panics. The
underlying Termion and Termwiz code both already implement `Drop` so no
changes are required to them.

This commit adds a Deref implementation to terminal to allow the backend
to be accessed directly.

The README and library, terminal, backends, types, traits and modules
docs are all updated to reflect these changes and to add some missing
documentation and examples.

Fixes https://github.com/ratatui-org/ratatui/issues/261

## PR Review notes

This PR consists of 2 commits, the first makes the changes which should be backwards
compatible with all existing code. The second updates the examples to use the new
methods.

To make it easier to review this, I'd recommend looking at each commit (there is no overlap in the code):

- Code and docs: https://github.com/ratatui-org/ratatui/pull/280/commits/2498f7bba6387055698c65edb97a0f0d7e2654f8
The change is large (1206 insertions, 212 deletions), but a large chunk is comments (468 lines)
- Examples: https://github.com/ratatui-org/ratatui/pull/280/commits/85fe06606a185fff99f73c6ed82125426cdf4039

Note that https://github.com/ratatui-org/ratatui/pull/280#issuecomment-1657115361 explains a few choices that might not be obvious.

I'd recommend generating the docs locally and viewing the main library doc, backend and terminal module docs, Terminal type, Backend trait, and each Backend type.
```
cargo +nightly doc -Zunstable-options -Zrustdoc-scrape-examples --all-features --open
```

I'm seeking feedback on anything that's missing in the docs, anything that's not where you'd expect it to be, anything that's not idiomatic (taking into account Ratatui's general builder pattern as well as Rust norms), and any problems this new code might cause for app / library builders. Thoughts on how to communicate these changes to users would be appreciated as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ project.
 [![Crates.io](https://img.shields.io/crates/v/ratatui?logo=rust&style=flat-square)](https://crates.io/crates/ratatui)
 [![License](https://img.shields.io/crates/l/ratatui?style=flat-square)](./LICENSE) [![GitHub CI
 Status](https://img.shields.io/github/actions/workflow/status/ratatui-org/ratatui/ci.yml?style=flat-square&logo=github)](https://github.com/ratatui-org/ratatui/actions?query=workflow%3ACI+)
-[![Docs.rs](https://img.shields.io/docsrs/ratatui?logo=rust&style=flat-square)](https://docs.rs/crate/ratatui/)  
+[![Docs.rs](https://img.shields.io/docsrs/ratatui?logo=rust&style=flat-square)](https://docs.rs/crate/ratatui/)\
 [![Dependency
 Status](https://deps.rs/repo/github/ratatui-org/ratatui/status.svg?style=flat-square)](https://deps.rs/repo/github/ratatui-org/ratatui)
 [![Codecov](https://img.shields.io/codecov/c/github/ratatui-org/ratatui?logo=codecov&style=flat-square&token=BAQ8SOKEST)](https://app.codecov.io/gh/ratatui-org/ratatui)
@@ -94,30 +94,16 @@ the [Docs](https://docs.rs/ratatui) and [Examples](#examples). There is also a s
 available at [rust-tui-template](https://github.com/ratatui-org/rust-tui-template).
 
 ```rust
-fn main() -> Result<(), Box<dyn Error>> {
-    let mut terminal = setup_terminal()?;
-    run(&mut terminal)?;
-    restore_terminal(&mut terminal)?;
-    Ok(())
-}
+use std::time::Duration;
 
-fn setup_terminal() -> Result<Terminal<CrosstermBackend<Stdout>>, Box<dyn Error>> {
-    let mut stdout = io::stdout();
-    enable_raw_mode()?;
-    execute!(stdout, EnterAlternateScreen)?;
-    Ok(Terminal::new(CrosstermBackend::new(stdout))?)
-}
+use anyhow::Result;
+use crossterm::event::{self, Event, KeyCode};
+use ratatui::{prelude::*, widgets::Paragraph};
 
-fn restore_terminal(
-    terminal: &mut Terminal<CrosstermBackend<Stdout>>,
-) -> Result<(), Box<dyn Error>> {
-    disable_raw_mode()?;
-    execute!(terminal.backend_mut(), LeaveAlternateScreen,)?;
-    Ok(terminal.show_cursor()?)
-}
-
-fn run(terminal: &mut Terminal<CrosstermBackend<Stdout>>) -> Result<(), Box<dyn Error>> {
-    Ok(loop {
+fn main() -> Result<()> {
+    let backend = CrosstermBackend::on_stdout()?;
+    let mut terminal = Terminal::new(backend)?;
+    loop {
         terminal.draw(|frame| {
             let greeting = Paragraph::new("Hello World!");
             frame.render_widget(greeting, frame.size());
@@ -125,11 +111,11 @@ fn run(terminal: &mut Terminal<CrosstermBackend<Stdout>>) -> Result<(), Box<dyn 
         if event::poll(Duration::from_millis(250))? {
             if let Event::Key(key) = event::read()? {
                 if KeyCode::Char('q') == key.code {
-                    break;
+                    return Ok(());
                 }
             }
         }
-    })
+    }
 }
 ```
 
@@ -220,8 +206,8 @@ be installed with `cargo install cargo-make`).
   `tui::text::Text`
 * [color-to-tui](https://github.com/uttarayan21/color-to-tui) — Parse hex colors to
   `tui::style::Color`
-* [rust-tui-template](https://github.com/ratatui-org/rust-tui-template) — A template for bootstrapping a
-  Rust TUI application with Tui-rs & crossterm
+* [rust-tui-template](https://github.com/ratatui-org/rust-tui-template) — A template for
+  bootstrapping a Rust TUI application with Tui-rs & crossterm
 * [simple-tui-rs](https://github.com/pmsanford/simple-tui-rs) — A simple example tui-rs app
 * [tui-builder](https://github.com/jkelleyrtp/tui-builder) — Batteries-included MVC framework for
   Tui-rs + Crossterm apps
--- a/README.md
+++ b/README.md
@@ -101,8 +101,7 @@ use crossterm::event::{self, Event, KeyCode};
 use ratatui::{prelude::*, widgets::Paragraph};
 
 fn main() -> Result<()> {
-    let backend = CrosstermBackend::on_stdout()?;
-    let mut terminal = Terminal::new(backend)?;
+    let mut terminal = TerminalBuilder::crossterm_on_stdout();
     loop {
         terminal.draw(|frame| {
             let greeting = Paragraph::new("Hello World!");
--- a/bacon.toml
+++ b/bacon.toml
@@ -12,44 +12,84 @@ command = ["cargo", "check", "--all-features", "--color", "always"]
 need_stdout = false
 
 [jobs.check-all]
-command = ["cargo", "check", "--all-targets", "--all-features", "--color", "always"]
+command = [
+    "cargo",
+    "check",
+    "--all-targets",
+    "--all-features",
+    "--color",
+    "always",
+]
 need_stdout = false
 
 [jobs.check-crossterm]
-command = ["cargo", "check", "--color", "always", "--all-targets", "--no-default-features", "--features", "crossterm"]
+command = [
+    "cargo",
+    "check",
+    "--color",
+    "always",
+    "--all-targets",
+    "--no-default-features",
+    "--features",
+    "crossterm",
+]
 need_stdout = false
 
 [jobs.check-termion]
-command = ["cargo", "check", "--color", "always", "--all-targets", "--no-default-features", "--features", "termion"]
+command = [
+    "cargo",
+    "check",
+    "--color",
+    "always",
+    "--all-targets",
+    "--no-default-features",
+    "--features",
+    "termion",
+]
 need_stdout = false
 
 [jobs.check-termwiz]
-command = ["cargo", "check", "--color", "always", "--all-targets", "--no-default-features", "--features", "termwiz"]
+command = [
+    "cargo",
+    "check",
+    "--color",
+    "always",
+    "--all-targets",
+    "--no-default-features",
+    "--features",
+    "termwiz",
+]
 need_stdout = false
 
 [jobs.clippy]
-command = [
-    "cargo", "clippy",
-    "--all-targets",
-    "--color", "always",
-]
+command = ["cargo", "clippy", "--all-targets", "--color", "always"]
 need_stdout = false
 
 [jobs.test]
 command = [
-    "cargo", "test",
+    "cargo",
+    "test",
     "--all-features",
-    "--color", "always",
-    "--", "--color", "always", # see https://github.com/Canop/bacon/issues/124
+    "--color",
+    "always",
+    "--",
+    "--skip",
+    "terminal_builder::tests::termion", # competes with bacon for the terminal
+    "--color",
+    "always",                           # see https://github.com/Canop/bacon/issues/124
 ]
 need_stdout = true
 
 [jobs.doc]
 command = [
-    "cargo", "+nightly", "doc",
-    "-Zunstable-options", "-Zrustdoc-scrape-examples",
+    "cargo",
+    "+nightly",
+    "doc",
+    "-Zunstable-options",
+    "-Zrustdoc-scrape-examples",
     "--all-features",
-    "--color", "always",
+    "--color",
+    "always",
     "--no-deps",
 ]
 env.RUSTDOCFLAGS = "--cfg docsrs"
@@ -59,10 +99,14 @@ need_stdout = false
 # to the previous job
 [jobs.doc-open]
 command = [
-    "cargo", "+nightly", "doc",
-    "-Zunstable-options", "-Zrustdoc-scrape-examples",
+    "cargo",
+    "+nightly",
+    "doc",
+    "-Zunstable-options",
+    "-Zrustdoc-scrape-examples",
     "--all-features",
-    "--color", "always",
+    "--color",
+    "always",
     "--no-deps",
     "--open",
 ]
@@ -72,19 +116,27 @@ on_success = "job:doc" # so that we don't open the browser at each change
 
 [jobs.coverage]
 command = [
-    "cargo", "llvm-cov",
-    "--lcov", "--output-path", "target/lcov.info",
-    "--all-features", 
-    "--color", "always",
+    "cargo",
+    "llvm-cov",
+    "--lcov",
+    "--output-path",
+    "target/lcov.info",
+    "--all-features",
+    "--color",
+    "always",
 ]
 
 [jobs.coverage-unit-tests-only]
 command = [
-    "cargo", "llvm-cov",
-    "--lcov", "--output-path", "target/lcov.info",
+    "cargo",
+    "llvm-cov",
+    "--lcov",
+    "--output-path",
+    "target/lcov.info",
     "--lib",
     "--all-features",
-    "--color", "always",
+    "--color",
+    "always",
 ]
 
 # You may define here keybindings that would be specific to
--- a/examples/block.rs
+++ b/examples/block.rs
@@ -14,11 +14,9 @@ use ratatui::{
 // These type aliases are used to make the code more readable by reducing repetition of the generic
 // types. They are not necessary for the functionality of the code.
 type Frame<'a> = ratatui::Frame<'a, CrosstermBackend<Stdout>>;
-type Terminal = ratatui::Terminal<CrosstermBackend<Stdout>>;
 
 fn main() -> Result<()> {
-    let backend = CrosstermBackend::on_stdout()?;
-    let mut terminal = Terminal::new(backend)?;
+    let mut terminal = TerminalBuilder::crossterm_on_stdout().build()?;
 
     loop {
         terminal.draw(ui)?;
--- a/examples/block.rs
+++ b/examples/block.rs
@@ -1,15 +1,7 @@
-use std::{
-    error::Error,
-    io::{stdout, Stdout},
-    ops::ControlFlow,
-    time::Duration,
-};
+use std::{io::Stdout, ops::ControlFlow, time::Duration};
 
-use crossterm::{
-    event::{self, Event, KeyCode},
-    execute,
-    terminal::{disable_raw_mode, enable_raw_mode, EnterAlternateScreen, LeaveAlternateScreen},
-};
+use anyhow::Result;
+use crossterm::event::{self, Event, KeyCode};
 use itertools::Itertools;
 use ratatui::{
     prelude::*,
@@ -23,35 +15,11 @@ use ratatui::{
 // types. They are not necessary for the functionality of the code.
 type Frame<'a> = ratatui::Frame<'a, CrosstermBackend<Stdout>>;
 type Terminal = ratatui::Terminal<CrosstermBackend<Stdout>>;
-type Result<T> = std::result::Result<T, Box<dyn Error>>;
 
 fn main() -> Result<()> {
-    let mut terminal = setup_terminal()?;
-    let result = run(&mut terminal);
-    restore_terminal(terminal)?;
+    let backend = CrosstermBackend::on_stdout()?;
+    let mut terminal = Terminal::new(backend)?;
 
-    if let Err(err) = result {
-        eprintln!("{err:?}");
-    }
-    Ok(())
-}
-
-fn setup_terminal() -> Result<Terminal> {
-    enable_raw_mode()?;
-    let mut stdout = stdout();
-    execute!(stdout, EnterAlternateScreen)?;
-    let backend = CrosstermBackend::new(stdout);
-    let terminal = Terminal::new(backend)?;
-    Ok(terminal)
-}
-
-fn restore_terminal(mut terminal: Terminal) -> Result<()> {
-    disable_raw_mode()?;
-    execute!(terminal.backend_mut(), LeaveAlternateScreen)?;
-    Ok(())
-}
-
-fn run(terminal: &mut Terminal) -> Result<()> {
     loop {
         terminal.draw(ui)?;
         if handle_events()?.is_break() {
--- a/src/backend/crossterm.rs
+++ b/src/backend/crossterm.rs
@@ -31,29 +31,32 @@ use crate::{
 ///
 /// # Examples
 ///
-/// Generally, apps should use [`CrosstermBackend::on_stderr()`] or
-/// [`CrosstermBackend::on_stdout()`] methods, which both enable raw mode and enter the alternate
-/// screen. Choosing `stderr` over `stdout` ensures your app displays on the terminal even if the
-/// standard output stream is redirected and makes it easy to write apps that can be piped to other
-/// programs.
+/// Generally, apps should use [`TerminalBuilder`] to create a `Terminal` with a `CrosstermBackend`
+/// instead of using `CrosstermBackend` directly. This will enable raw mode, enter the alternate
+/// screen, and disable mouse capture. See [`TerminalBuilder`] for more information.
 ///
 /// ```rust,no_run
-/// use ratatui::backend::CrosstermBackend;
-///
-/// let mut backend = CrosstermBackend::on_stdout()?;
-/// // alternatively
-/// let mut backend = CrosstermBackend::on_stderr()?;
+/// use ratatui::TerminalBuilder;
+/// let mut terminal = TerminalBuilder::crossterm_on_stdout().build()?;
 /// # std::io::Result::Ok(())
 /// ```
 ///
-/// For more control over raw mode, the alternate screen, or mouse capture use
-/// [`CrosstermBackend::new()`].
+/// Alternatively, you can use [`CrosstermBackend::on_stdout()`] or
+/// [`CrosstermBackend::on_stderr()`] to create a `CrosstermBackend` with [`std::io::stdout()`] or
+/// [`std::io::stderr()`] as the writer.
+///
 ///
 /// ```rust,no_run
-/// use std::io::stdout;
 /// use ratatui::backend::CrosstermBackend;
 ///
-/// let mut backend = CrosstermBackend::new(stdout())
+/// let mut backend = CrosstermBackend::on_stdout()
+///     .with_raw_mode()?
+///     .with_alternate_screen()?
+///     .with_mouse_capture()?;
+///
+/// // alternatively
+///
+/// let mut backend = CrosstermBackend::on_stderr()
 ///     .with_raw_mode()?
 ///     .with_alternate_screen()?
 ///     .with_mouse_capture()?;
@@ -64,7 +67,7 @@ use crate::{
 #[derive(Debug, Clone, Eq, PartialEq, Hash)]
 pub struct CrosstermBackend<W: Write> {
     /// The buffer being written to.
-    buffer: W,
+    writer: W,
     /// Whether raw mode is enabled. This is used to determine whether to disable raw mode when
     /// dropping the backend.
     raw_mode_enabled: bool,
@@ -90,9 +93,9 @@ where
     /// let mut backend = CrosstermBackend::new(stdout());
     /// # std::io::Result::Ok(())
     /// ```
-    pub fn new(buffer: W) -> CrosstermBackend<W> {
+    pub fn new(writer: W) -> CrosstermBackend<W> {
         CrosstermBackend {
-            buffer,
+            writer,
             raw_mode_enabled: false,
             alternate_screen_entered: false,
             mouse_capture_enabled: false,
@@ -178,7 +181,7 @@ where
     ///
     /// See [`crate::backend#alternate-screen`] for more information.
     pub fn enter_alternate_screen(&mut self) -> io::Result<()> {
-        execute!(self.buffer, EnterAlternateScreen)?;
+        execute!(self.writer, EnterAlternateScreen)?;
         self.alternate_screen_entered = true;
         Ok(())
     }
@@ -189,7 +192,7 @@ where
     ///
     /// See [`crate::backend#alternate-screen`] for more information.
     pub fn leave_alternate_screen(&mut self) -> io::Result<()> {
-        execute!(self.buffer, LeaveAlternateScreen)?;
+        execute!(self.writer, LeaveAlternateScreen)?;
         self.alternate_screen_entered = false;
         Ok(())
     }
@@ -200,7 +203,7 @@ where
     ///
     /// See [`crate::backend#mouse-capture`] for more information.
     pub fn enable_mouse_capture(&mut self) -> io::Result<()> {
-        execute!(self.buffer, EnableMouseCapture)?;
+        execute!(self.writer, EnableMouseCapture)?;
         self.mouse_capture_enabled = true;
         Ok(())
     }
@@ -211,7 +214,7 @@ where
     ///
     /// See [`crate::backend#mouse-capture`] for more information.
     pub fn disable_mouse_capture(&mut self) -> io::Result<()> {
-        execute!(self.buffer, DisableMouseCapture)?;
+        execute!(self.writer, DisableMouseCapture)?;
         self.mouse_capture_enabled = false;
         Ok(())
     }
@@ -220,88 +223,28 @@ where
 impl CrosstermBackend<Stderr> {
     /// Creates a new `CrosstermBackend` using [`std::io::stderr()`] as the buffer.
     ///
-    /// This method also enables [raw mode] and enters the [alternate screen].
-    ///
     /// # Example
     ///
     /// ```rust,no_run
     /// # use ratatui::backend::{Backend, CrosstermBackend};
-    /// let mut backend = CrosstermBackend::on_stderr()?;
-    /// # std::io::Result::Ok(())
+    /// let mut backend = CrosstermBackend::on_stderr();
     /// ```
-    ///
-    /// [raw mode]: crate::backend#raw-mode
-    /// [alternate screen]: crate::backend#alternate-screen
-    pub fn on_stderr() -> Result<CrosstermBackend<Stderr>, io::Error> {
+    pub fn on_stderr() -> CrosstermBackend<Stderr> {
         CrosstermBackend::new(io::stderr())
-            .with_raw_mode()?
-            .with_alternate_screen()
     }
 }
 
 impl CrosstermBackend<Stdout> {
     /// Creates a new `CrosstermBackend` using [`std::io::stdout()`] as the buffer.
     ///
-    /// This method also enables [raw mode] and enters the [alternate screen].
-    ///
     /// # Example
     ///
     /// ```rust,no_run
     /// # use ratatui::backend::{Backend, CrosstermBackend};
-    /// let mut backend = CrosstermBackend::on_stdout()?;
-    /// # std::io::Result::Ok(())
+    /// let mut backend = CrosstermBackend::on_stdout();
     /// ```
-    ///
-    /// [raw mode]: crate::backend#raw-mode
-    /// [alternate screen]: crate::backend#alternate-screen
-    pub fn on_stdout() -> Result<CrosstermBackend<Stdout>, io::Error> {
+    pub fn on_stdout() -> CrosstermBackend<Stdout> {
         CrosstermBackend::new(io::stdout())
-            .with_raw_mode()?
-            .with_alternate_screen()
-    }
-}
-
-impl Default for CrosstermBackend<Stderr> {
-    // impl Default for CrosstermBackend {
-    /// Creates a new `CrosstermBackend` using [`std::io::stderr()`] as the buffer.
-    ///
-    /// It will usually be more convenient to use [`CrosstermBackend::on_stderr()`], which enables
-    /// raw mode and enters the alternate screen.
-    ///
-    /// # Example
-    ///
-    /// ```rust,no_run
-    /// # use std::io::Stderr;
-    /// # use ratatui::backend::CrosstermBackend;
-    /// let mut backend = CrosstermBackend::<Stderr>::default()
-    ///     .with_raw_mode()?
-    ///     .with_alternate_screen()?;
-    /// # std::io::Result::Ok(())
-    /// ```
-    fn default() -> Self {
-        Self::new(io::stderr())
-    }
-}
-
-impl Default for CrosstermBackend<Stdout> {
-    // impl Default for CrosstermBackend {
-    /// Creates a new `CrosstermBackend` using [`std::io::stdout()`] as the buffer.
-    ///
-    /// It will usually be more convenient to use [`CrosstermBackend::on_stdout()`], which enables
-    /// raw mode and enters the alternate screen.
-    ///
-    /// # Example
-    ///
-    /// ```rust,no_run
-    /// # use std::io::Stdout;
-    /// # use ratatui::backend::{Backend, CrosstermBackend};
-    /// let mut backend = CrosstermBackend::<Stdout>::default()
-    ///     .with_raw_mode()?
-    ///     .with_alternate_screen()?;
-    /// # std::io::Result::Ok(())
-    /// ```
-    fn default() -> Self {
-        Self::new(io::stdout())
     }
 }
 
@@ -341,12 +284,12 @@ where
 {
     /// Writes a buffer of bytes to the underlying buffer.
     fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
-        self.buffer.write(buf)
+        self.writer.write(buf)
     }
 
     /// Flushes the underlying buffer.
     fn flush(&mut self) -> io::Result<()> {
-        self.buffer.flush()
+        self.writer.flush()
     }
 }
 
@@ -366,7 +309,7 @@ where
         for (x, y, cell) in content {
             // Move the cursor if the previous location was not (x - 1, y)
             if !matches!(last_pos, Some(p) if x == p.0 + 1 && y == p.1) {
-                queue!(self.buffer, MoveTo(x, y))?;
+                queue!(self.writer, MoveTo(x, y))?;
             }
             last_pos = Some((x, y));
             if cell.modifier != modifier {
@@ -374,30 +317,30 @@ where
                     from: modifier,
                     to: cell.modifier,
                 };
-                diff.queue(&mut self.buffer)?;
+                diff.queue(&mut self.writer)?;
                 modifier = cell.modifier;
             }
             if cell.fg != fg {
                 let color = CColor::from(cell.fg);
-                queue!(self.buffer, SetForegroundColor(color))?;
+                queue!(self.writer, SetForegroundColor(color))?;
                 fg = cell.fg;
             }
             if cell.bg != bg {
                 let color = CColor::from(cell.bg);
-                queue!(self.buffer, SetBackgroundColor(color))?;
+                queue!(self.writer, SetBackgroundColor(color))?;
                 bg = cell.bg;
             }
             if cell.underline_color != underline_color {
                 let color = CColor::from(cell.underline_color);
-                queue!(self.buffer, SetUnderlineColor(color))?;
+                queue!(self.writer, SetUnderlineColor(color))?;
                 underline_color = cell.underline_color;
             }
 
-            queue!(self.buffer, Print(&cell.symbol))?;
+            queue!(self.writer, Print(&cell.symbol))?;
         }
 
         queue!(
-            self.buffer,
+            self.writer,
             SetForegroundColor(CColor::Reset),
             SetBackgroundColor(CColor::Reset),
             SetUnderlineColor(CColor::Reset),
@@ -406,11 +349,11 @@ where
     }
 
     fn hide_cursor(&mut self) -> io::Result<()> {
-        execute!(self.buffer, Hide)
+        execute!(self.writer, Hide)
     }
 
     fn show_cursor(&mut self) -> io::Result<()> {
-        execute!(self.buffer, Show)
+        execute!(self.writer, Show)
     }
 
     fn get_cursor(&mut self) -> io::Result<(u16, u16)> {
@@ -419,7 +362,7 @@ where
     }
 
     fn set_cursor(&mut self, x: u16, y: u16) -> io::Result<()> {
-        execute!(self.buffer, MoveTo(x, y))
+        execute!(self.writer, MoveTo(x, y))
     }
 
     fn clear(&mut self) -> io::Result<()> {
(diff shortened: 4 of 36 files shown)


joshka:
My goal with putting up a WIP here is to get some early feedback on the changes.

a-kenji:
Generally a :+1:  from me. I have been implementing the drop trait in a TerminalWrapper multiple times now.

> Chooses the Stderr to display the UI as this allows the app to use
stdout for logging or output to pipes.

I'm not sure, if that is something we really want. This seems like unexpected behavior to me.

> decide whether to enable mouse support by default

I am against that. At least in it's current state. We don't provide an easy way to handle rich mouse input yet, which means most apps won't implement that. If we provide that by default, then not even selecting/copying text from the screen will work, because that is unhandled. I think generally speaking this will lead to a worse user experience.

mindoodoo:
Yeah +1 from my end regarding the need for this feature however I agree with @a-kenji on both points he brought up.

Additionally, why did you enable all backend features as default ?

joshka:
Having mouse capture on by default wasn't something I thought was a good idea either. Glad that there's some agreement there. I'll remove that.

The change for enabling the features was purely for convenience of having vscode highlight errors and provide fixes etc. It won't be there once cleaned up.

Regarding Stdout vs Stderr. We have to pick a convention as a default (otherwise we can't implement `Default`).
I also queried this in the past, but have come around to the idea that Stderr is the right approach. I can't see any obvious benefit for rendering on Stdout, while there are a couple for Stderr.

As an experiment, change this to Stdout by default, and run `cargo run --example hello_world | grep foo`.
The app doesn't render anything to the screen and there's no indication of what's going wrong. The app loop still runs, and the input is handled, but you don't see anything until the app quits.
Flip the default to Stderr, and the app runs fine.

Stdout:
- every app needs to add code to check if the output is a tty and do something different based on the result
- app can't write a result to the user that can be passed in a pipeline `my-select-some-value-app | grep foo`
- default behavior would require more code to be robust
Stderr:
- no special setup necessary in order to run in a pipeline
- default behavior just works

Regarding the `drop(terminal)` calls, I'm going to change this to the more explicit `terminal.leave_alternate_mode()?` which makes more sense (as it states what needs to be done to print after the tui app is done rather than implicitly relying on drop() to do that).

I forgot to mention the other leftover change - what to do about the panic example. I think most of the example probably disappears if coded so that the terminal goes out of scope on panic.

joshka:
In the [changes pushed earlier](https://github.com/tui-rs-revival/ratatui/compare/5e0622b2afa3d904756a4781dc50cd66b4e9afc5..8ab47be8aa499251c446ad245f97aa28fffa8f4c) (work in progress - I'll split this into separate PRs once finished):
- examples now call `terminal.leave_alternate_screen()` if they need to display an error message on exit.
- changed the demo example to be able to compile with all features enabled (which makes it easier to work on backends)
- Playing a bit with the Default impl as I don't like that it has to panic in order to get a good default (perhaps that's unavoidable?)
- Added a tonne of doc comments
- Added `on_stderr()` and `on_stdout()` to make it more explicit to specify a destination (and to allow for this to return a result rather than panic)
- Made mod terminal private to avoid having two ways to specify `Terminal` (and reexported the remaining necessary type in the mod that wasn't already exported)

(don't stress too much on leaving reviews on this PR yet - it's mostly placeholder to help explain some things in discord conversations)
-

a-kenji:
> Regarding the `drop(terminal)` calls, I'm going to change this to the more explicit `terminal.leave_alternate_mode()?` 

That does make sense, that still leaves open the possibility for consumers to implement drop in an ergonomic way.

> As an experiment, change this to Stdout by default, and `run cargo run --example hello_world | grep foo`.

That's my issue. I wouldn't expect this to work, at all. Unless the app developer communicated to me that there is some sort of functionality.

What I would expect is that per default the following would work:
```
run cargo run --example hello_world  2> error.log
```
Which allows the app to render fine, but also log error messages that are printed to a specific file.

> app can't write a result to the user that can be passed in a pipeline

> every app needs to add code to check if the output is a tty and do something different based on the result

I see this similar to the mouse issue, if apps want users to be able to do it, then the experience is in general better, if they implement it. And the check is in the [stdlib](https://doc.rust-lang.org/stable/std/io/trait.IsTerminal.html#tymethod.is_terminal).

That way the pipeline can be implemented in an expected way, if it is a desired functionality:
```
cargo run --example hello_world | grep foo
```
This will then have correct stdout for the pipelined process, but is still able to provide error messages in an expected way.

I am all for enabling users to provide awesome and ergonomic features. If we provide a default, I think we should provide a generally good experience for users of the apps and the users of the library. I don't think rendering to `stderr` by default does strike that balance.

To make it easier for apps to implement such a feature, I would rather suggest having an example that checks for a tty and if possible some functionality that enables such a use case in a cleaner way.

joshka:
Thanks for presenting your perspective on this. It's helpful to understand it a bit more.

> That's my issue. I wouldn't expect this to work, at all. Unless the app developer communicated to me that there is some sort of functionality.

The problem though is that we'd have to add a bunch more code to handle this properly (I think termwiz has some logic around checking tty etc.). What I wouldn't expect for any command line app ever is for it to just hang without any indication of how to get pass that.

> To make it easier for apps to implement such a feature, I would rather suggest having an example that checks for a tty and if possible some functionality that enables such a use case in a cleaner way.

We can, but then demo code gets copied. We're talking about the default happy path on this one.

> What I would expect is that per default the following would work:
```
cargo run --example hello_world  2> error.log
```
If I modify hello_world to:
```rust
fn main() -> Result<()> {
    eprintln!("BEFORE SETUP");
    let mut terminal = setup_terminal().context("setup failed")?;
    eprintln!("AFTER SETUP");
    run(&mut terminal).context("app loop failed")?;
    eprintln!("AFTER RUN");
    restore_terminal(&mut terminal).context("restore terminal failed")?;
    eprintln!("AFTER RESTORE");
    Ok(())
}
```
I see:
```
❯ cargo run --example hello_world
    Finished dev [unoptimized + debuginfo] target(s) in 0.05s
     Running `target/debug/examples/hello_world`
BEFORE SETUP
AFTER RESTORE
```
Which makes it not very useful as a stream of error information (unless the app drops the alternate screen every it needs to write an error log), or it stores up the errors until the app finishes. Either of those approaches seem bad as a general rule (though OK if done intentionally).

Taking into account that switching to raw mode / alternate screen / mouse capture can fail, I think perhaps just avoiding implementing Default might be preferred and instead having something like `fn with_crossterm() -> Result<Terminal>`, but the same issue comes up there with regard to choosing StdErr over StdOut.

I think this comes down to which of the following is likely to be more common:
```
app | grep ...
```
or 
```
app 2> error.log
```

Do you have any idea what app patterns are out there for things like this. I know that the "returns a value on stdout" pattern is around (e.g. fzf, ijq and similar tools that display a tui for selection / filtering / etc.)

---

BTW, I'm also experimenting a little what a better builder would look like for this. E.g.:
```rust
TerminalBuilder::crossterm().stderr().alternate_screen().raw_mode().mouse_capture().build()?
```
With a convenience method on Terminal that calls that. E.g:
```rust
fn with_crossterm() -> Result<Terminal> {}
```
Or perhaps:
```rust
fn with_crossterm_on_stderr() -> Result<Terminal> {}
fn with_crossterm_on_stdout() -> Result<Terminal> {}
```

a-kenji:
> The problem though is that we'd have to add a bunch more code to handle this properly (I think termwiz has some logic around checking tty etc.).

Why would we need to add a bunch more code to handle this properly?

> Taking into account that switching to raw mode / alternate screen / mouse capture can fail, I think perhaps just avoiding implementing Default might be preferred

Yes, that does sound reasonable.


> but the same issue comes up there with regard to choosing StdErr over StdOut.


I still think the most safe, expected behavior for an average app is to show it's output on stdout. 

This will fail for more normal tui apps:
```
app | grep
htop | grep foo
zellij | grep foo
gh-dash | grep foo
```
This will work for more normal tui apps:
```
gh-dash 2> error.log
```

>  I know that the "returns a value on stdout" pattern is around (e.g. fzf, ijq and similar tools that display a tui for selection / filtering / etc.)

They are crafted specifically for this purpose.
And they are very useful, but they are not the norm. And they behave differently than other apps.

> What I wouldn't expect for any command line app ever is for it to just hang without any indication of how to get pass that.

Most cli's return after a very short time and are made for pipes.
Most tui's don't return for any specified amount of time. Tui's usually have a main loop. If I redirect the output of an app, that doesn't return for any specified amount of time, this is exactly what I would expect. At least unless the app, or it's author/documentation led me to believe something different.

haydenflinner:
I have never seen a TUI which used stderr for logging, even the linked gh-dash [doesn't seem to.](https://github.com/dlvhdr/gh-dash/blob/3caed9238b661f1782afd906b0f9a7d201b36293/cmd/root.go#L52). To that end using stderr for the display seems like a reasonable default with no real downside, though there seems to be no loss in making it an explicit choice for the user.

>Added on_stderr() and on_stdout() to make it more explicit to specify a destination (and to allow for this to return a result rather than panic)
This sounds perfect, always better to avoid panics to whatever extent possible. I can't imagine a reason that switching to raw mode would panic except if the chosen output wasn't a tty, in which case the user might check the other of out/err, or fall back to some other mode. Plus if you already need to give back a result (panicking is IMO not an option, hiding failure conditions behind runtime throws is not what Rust is about), the few extra characters to type if one wants out/err isn't any extra burden, and keeps things explicit.

This is a great initiative by the way, lowering the friction to use a library helps I think more than people realize.

joshka:
TL;DR: Implemented the changes mentioned, a tonne of docs, and simplifications of all the examples this. It's all backwards compatible except for the inline example right now. The result is ~38 lines becomes ~12 in most examples:

Before:
https://github.com/ratatui-org/ratatui/blob/440f62ff5435af9536c55d17707a9bc48dae92cc/examples/layout.rs#L10-L47

After:
https://github.com/ratatui-org/ratatui/blob/4fdf732573e9097e9c25636fc2a78890034aaedc/examples/layout.rs#L5-L16

There's still some TODOs left on this, the inline example is broken, and the panic example needs a look, and some docs and tests.

Reviewing this is going to be a pain as the change is quite large. So I'd definitely appreciate some thoughts on how best to make that easier. I split the examples changes out to a separate commit to show that they still compile and run even before applying the simplification (caveats on inline breaking). I suspect it's probably necessary to explain the rationale for the changes in each module more thoroughly, and perhaps it's worth implementing some backend unit tests before this lands?

<details>
<summary>
Commit details
</summary>
This is two commits - the first does the majority of the lifting:

```
commit 60548930b696ba8bb909c07948f7a1ff38e24fa4
Author: Josh McKinney <joshka@users.noreply.github.com>
Date:   Mon Jun 26 15:36:25 2023 -0700

    feat: simplify backend and terminal creation

    Partially addresses https://github.com/tui-rs-revival/ratatui/issues/261

    - Implement `on_stdout()` and `on_stderr()` methods for
    `CrosstermBackend` and `TermionBackend`. These create a backend that
    enable raw mode and switch to the alternate screen before returning.

    - Implement `Default` for `CrosstermBackend<Stdout>`,
    CrosstermBackend<Stderr>, `TermionBackend<Stdout>` and
    `TermionBackend<Stderr>` to make it possible to create a terminal with
    e.g. `Terminal<CrosstermBackend<Stdout>>::default()`. This will not
    enable raw mode or switch to the alternate screen.

    - Implement `Drop` for CrosstermBackend which will disable raw mode and
    switch back to the main screen. This ensures the terminal is left in a
    usable state if the application panics.

    - The underlying Termion and Termwiz code already implements `Drop` so
    no changes are required to the backend code.

    - Enables the Backends to be created using builder style methods, which
    avoid the need to call into the backend library directly.

    - Adds methods to the backend trait to handle raw mode, alternate screen
    and mouse support.

    - Adds a Deref implementation to terminal to allow the backend to be
    accessed directly.

    - Documents the terminal, backends, backend traits.

    TODO:
    - tests
    - full docs
    - finish builder methods
    - finish backend trait methods

 README.md                |  40 ++++++----------
 src/backend/crossterm.rs | 263 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++--------------
 src/backend/mod.rs       | 256 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++----------
 src/backend/termion.rs   | 198 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++-------------
 src/backend/termwiz.rs   | 126 ++++++++++++++++++++++++++++++++++++--------------
 src/backend/test.rs      |   3 +-
 src/lib.rs               | 185 +++++++++++++++++++++++++++++++++++++------------------------------------
 src/terminal.rs          | 351 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++---------------------------------------------
 src/widgets/block.rs     |   2 +-
 9 files changed, 1063 insertions(+), 361 deletions(-)
```

The second highlights how much code we save with the simplifications:

```
commit 4fdf732573e9097e9c25636fc2a78890034aaedc (HEAD -> wip-backend-defaults, origin/wip-backend-defaults)
Author: Josh McKinney <joshka@users.noreply.github.com>
Date:   Sun Jul 30 00:46:20 2023 -0700

    docs(examples): Simplify examples

 examples/barchart.rs       | 48 +++++++-----------------------------------------
 examples/block.rs          | 42 +++++++-----------------------------------
 examples/calendar.rs       | 23 ++++++-----------------
 examples/canvas.rs         | 45 ++++++---------------------------------------
 examples/chart.rs          | 48 +++++++-----------------------------------------
 examples/custom_widget.rs  | 36 +++---------------------------------
 examples/demo/crossterm.rs | 46 +++++-----------------------------------------
 examples/demo/main.rs      |  6 +++---
 examples/demo/termion.rs   | 38 ++++++++------------------------------
 examples/demo/termwiz.rs   | 71 +++++++++++++++++++++++++++++------------------------------------------
 examples/gauge.rs          | 46 ++++++----------------------------------------
 examples/hello_world.rs    | 41 +++++++----------------------------------
 examples/inline.rs         | 14 +++++---------
 examples/layout.rs         | 39 ++++-----------------------------------
 examples/list.rs           | 47 ++++++-----------------------------------------
 examples/panic.rs          | 53 +++++++++++++++--------------------------------------
 examples/paragraph.rs      | 48 +++++++-----------------------------------------
 examples/popup.rs          | 37 +++++--------------------------------
 examples/scrollbar.rs      | 47 ++++++-----------------------------------------
 examples/sparkline.rs      | 46 ++++++----------------------------------------
 examples/table.rs          | 38 +++++---------------------------------
 examples/tabs.rs           | 38 +++++---------------------------------
 examples/user_input.rs     | 39 +++++----------------------------------
 23 files changed, 163 insertions(+), 773 deletions(-)
```

Most of the examples had code that created a terminal (switching to raw/alternate screen), ran an app, storing the result, restored the terminal then if there was an error then the app would print the error. That's all handled automatically now, so this easily collapses to a single method

</details>


<details>
<summary>
There are some big differences in the way the backends handle raw/alternate/mouse modes. More details:
</summary>

- Crossterm is lax and just lets you do whatever (send commands to change modes). Everything is manual and there is no Drop that restores
- Termion handles the modes by wrapping the writer, the api for raw/alt and mouse are slightly different (stdout.into_raw/into_alternate_screen vs `MouseTerminal::from(writer)`) and the manual mode changes are exposed differently as well. Dropping the wrapper instance restores the mode.
- Termwiz has the ability to handle stderr, but we haven't written that. Mouse mode is enabled automatically with raw mode. Modes are restored on drop, and our new implementation automatically enables raw/alternate screen (we should perhaps fix this).
- Because of that, it's not possible to just have a trait that abstracts the calls, or even easily just use deref for this as the implementation details are different enough between the backend. I settled for putting methods on the backend directly where it makes sense.
</details>

joshka:
I think this is now in a good state for a review now.

- All examples now work 100% (the issue with the inline example was that leaving the alt screen when you're not in the alt screen cleared the output that was suppose to be left - this is fixed)
- I added / updated a bunch of doc comments everywhere. Generate the docs locally and look at the docs for the library, backend module, terminal module, as well as the Terminal type and each Backend type and traits.
- Removed leftover commented stuff

To make it easier to review this, I'd recommend looking at each commit (there is no overlap in the code):

- Code and docs: https://github.com/ratatui-org/ratatui/pull/280/commits/2498f7bba6387055698c65edb97a0f0d7e2654f8
The change is large, but a large chunk is comments 
```text
 README.md                |  40 ++++++-----------
 src/backend/crossterm.rs | 318 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++----------
 src/backend/mod.rs       | 188 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++-----------
 src/backend/termion.rs   | 332 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++--------
 src/backend/termwiz.rs   | 170 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++-----
 src/backend/test.rs      |   3 +-
 src/lib.rs               | 221 +++++++++++++++++++++++++++++++++++++++++++++++++-------------------------------------------
 src/terminal.rs          | 144 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++--
 src/widgets/block.rs     |   2 +-
 9 files changed, 1206 insertions(+), 212 deletions(-)
```
but
```
git show HEAD~ |grep -c  '[+-]//'
468
```
- Examples: https://github.com/ratatui-org/ratatui/pull/280/commits/85fe06606a185fff99f73c6ed82125426cdf4039

Note https://github.com/ratatui-org/ratatui/pull/280#issuecomment-1657115361 as it explains a few choices that might not be obvious.

src/backend/crossterm.rs
```diff
@@ -1,20 +1,19 @@
-//! This module provides the `CrosstermBackend` implementation for the `Backend` trait.
-//! It uses the `crossterm` crate to interact with the terminal.
+//! This module provides the [`CrosstermBackend`] implementation for the [`Backend`] trait. It uses
+//! the [Crossterm] crate to interact with the terminal.
```
kdheepak, line 2: Is there a reason `Crossterm` doesn't have backticks?

```suggestion
//! the [`Crossterm`] crate to interact with the terminal.
```

kdheepak: Never mind, I see the link at bottom now.

src/backend/crossterm.rs
```diff
@@ -24,36 +23,310 @@ use crate::{
...
     W: Write,
 {
     /// Creates a new `CrosstermBackend` with the given buffer.
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// # use std::io::stdout;
+    /// # use ratatui::backend::CrosstermBackend;
+    /// let mut backend = CrosstermBackend::new(stdout());
```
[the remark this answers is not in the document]
joshka: My rationale for including even trivial examples like this is because Ratatui is the sort of crate that first time rust users end up using, so making it as easy as possible at the expense of some verbosity is a reasonable tradeoff.

The secondary thing I like about super simple examples is that they act like a canary for when a method changes signature in an incompatible ways.

src/backend/mod.rs
```diff
@@ -79,24 +111,115 @@ pub trait Backend {
...
+    /// backend.hide_cursor()?;
+    /// // do something with hidden cursor
+    /// backend.show_cursor()?;
+    /// # Ok(())
+    /// # }
+    /// ```
+    ///
+    /// # Errors
+    ///
+    /// This method will return an error if the cursor could not be hidden.
```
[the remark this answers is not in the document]
joshka: I agree with your perspective that this doesn't add much value as written.

Do you think it's worth documenting these further for the 'what should I do about this' part of your comment? If so, perhaps it would be worth leaving these as-is adding an issue to do a dive into the underlying reasons and document further?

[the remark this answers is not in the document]
lthoerner: I don't think it's realistic to cover common failure cases on any large scale throughout the crate. This could be a separate project if someone really wants to do it.

src/backend/crossterm.rs
```diff
@@ -24,36 +23,310 @@ use crate::{
...
+    /// [raw mode]: crate::backend#raw-mode
+    /// [alternate screen]: crate::backend#alternate-screen
+    pub fn on_stdout() -> Result<CrosstermBackend<Stdout>, io::Error> {
+        CrosstermBackend::new(io::stdout())
+            .with_raw_mode()?
+            .with_alternate_screen()
+    }
+}
+
+impl Default for CrosstermBackend<Stderr> {
```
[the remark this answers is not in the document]
deifactor: I would assume that `Terminal::default()` would be a default that I can then use with no further setup for the common use case. Since that doesn't seem to be the case, I think 1 we should go with 1 here.

lthoerner (approved):
This makes the crate a *lot* more usable, great work.

src/backend/mod.rs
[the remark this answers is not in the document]
deifactor: I think if there's some where you already know where they could fail, it'd be good to document them. Otherwise, it's better to just not have the error section there at all.

One counterexample is the clear method; I'd leave that as-is, since "the backend doesn't support the clear type" is a concrete reason, and just having "Fails if the backend doesn't support the clear type" would give me the false impression that it *always* succeeds if the backend does support it.

src/backend/crossterm.rs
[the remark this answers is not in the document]
deifactor: Hmm. Looking at the rust docs, it does look like they have example docs even for 'trivial' methods, so it's possible I'm just wrong on this. Ultimately this isn't something I feel super passionate about either way. :)

src/backend/mod.rs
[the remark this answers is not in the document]
deifactor: We should have a section saying "these should succeed on any modern terminal, so failure is unexpected and indicates something abnormal is occurring" (assuming that's true), but it's not necessary to document the exact reasons why these might fail, just that "panic and exit" is a valid thing to do here.

deifactor:
> This makes the crate a _lot_ more usable, great work.

Agreed; I may be a little nitpicky about docs but I'm very glad I'll never have to write my own `Drop` wrappers or manage panic hooks again :)

src/backend/crossterm.rs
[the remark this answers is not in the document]
joshka: Sounds good - I'll wait on other opinions. Leaning towards keeping it as-is.

src/backend/mod.rs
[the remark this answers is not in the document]
joshka: Enabling raw mode is something that does fail in doc tests as the terminal presented to the backend doesn't support it. This might be worth calling out here specifically.

joshka:
> Agreed; I may be a little nitpicky about docs but I'm very glad I'll never have to write my own `Drop` wrappers or manage panic hooks again :)

Panic hooks will have to be written right now if you want the panic report to be shown on the main screen. The panic hook runs before the drop method. I'm not fully satisfied with this approach and would like to revisit it to indeed make it possible to easily handle panics. Perhaps we should hook this automatically in `Terminal::new()` I'd push this out to another PR though as there's some decisions to be made on how to configure that that I'd prefer not to get into on this PR.

kdheepak:
What happens if the user has a panic hook that exits alternate screen and prints out the stacktrace? The Drop running after that is still fine?

joshka:
> What happens if the user has a panic hook that exits alternate screen and prints out the stacktrace? The Drop running after that is still fine?

Yep - check the panic example for this behavior (`cargo run --example panic`).

joshka:
I just added a couple of review pointers to the main PR comment (updates don't notify those reading the thread already)
https://github.com/ratatui-org/ratatui/pull/280#issue-1775841533

src/backend/crossterm.rs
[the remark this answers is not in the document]
joshka: It sort of does, raw mode and alternate screen are not strictly necessary to make an app work, they're mainly nice to have.

src/backend/crossterm.rs
```diff
@@ -24,36 +23,310 @@ use crate::{
...
+/// The `CrosstermBackend` struct is a wrapper around a type implementing [`std::io::Write`], which
+/// is used to send commands to the terminal. It provides methods for drawing content, manipulating
+/// the cursor, and clearing the terminal screen.
+///
+/// # Examples
 ///
-/// The `CrosstermBackend` struct is a wrapper around a type implementing `Write`, which
-/// is used to send commands to the terminal. It provides methods for drawing content,
-/// manipulating the cursor, and clearing the terminal screen.
+/// Generally apps should [`CrosstermBackend::on_stderr()`] or [`CrosstermBackend::on_stdout()`]
```
kdheepak, line 34:
```suggestion
/// Generally, apps should use [`CrosstermBackend::on_stderr()`] or [`CrosstermBackend::on_stdout()`]
```

src/backend/crossterm.rs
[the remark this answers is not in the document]
kdheepak: In general, I think more examples are better, especially for people new to rust _and_ new to ratatui. In this particular case, I don't feel strongly one way or the other.

src/backend/crossterm.rs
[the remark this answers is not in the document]
kdheepak: I'm a little torn on whether `on_stderr()` should do `raw_mode` and `alternate_screen` as well. 

Maybe it's just a naming thing. I do like having a convenience function that does `stderr + raw_mode + alternate_screen`, but if I were reading code that used this for the first time, I'd be very surprised to learn that a function called `on_stderr()` also activated `raw_mode` and also toggled to the `alternate_screen`.

What are your thoughts on this? Can we brainstorm alternative names for this?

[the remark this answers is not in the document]
kdheepak: Maybe `on_stderr_tui()`?

src/backend/crossterm.rs
```diff
@@ -24,36 +23,310 @@ use crate::{
...
+impl<W> Drop for CrosstermBackend<W>
+where
+    W: Write,
+{
+    /// Disables raw mode, and mouse capture, leaves the alternate screen and shows the cursor.
+    ///
+    /// # Panics
+    ///
+    /// This method panics if it fails to disable raw mode, leave the alternate screen, disable
+    /// mouse capture, or show the cursor.
```
kdheepak, line 316: Can we not panic here instead? Maybe just log a warning? I can anticipate all kinds of weird crashes in CI or edge cases like that where I think it'd be weird if it just panicked on drop.

src/backend/mod.rs
```diff
@@ -1,30 +1,57 @@
...
+//!
+//! Raw mode is a mode where the terminal does not perform any processing on the input and output.
+//! This is useful for applications that want to process the input and output themselves. For
+//! example, in raw mode, the terminal will not perform line buffering on the input, so the
+//! application will receive each key press as it is typed instead of waiting for the user to press
+//! enter.
+//!
+//! Each backend handles raw mode differently, so the behavior may vary depending on the backend
+//! being used.
+//!
```
kdheepak, line 33: Expanding documentation further (aided with ChatGPT)

```suggestion
//! # Raw Mode
//!
//! Raw mode is a mode where the terminal does not perform any processing or handling of the input and output.
//! This means that features such as echoing input characters, line buffering, and special character processing
//! (e.g., CTRL-C for SIGINT) are disabled. This is useful for applications that want to have complete control
//! over the terminal input and output, processing each keystroke themselves.
//!
//! For example, in raw mode, the terminal will not perform line buffering on the input, so the
//! application will receive each key press as it is typed, instead of waiting for the user to press
//! enter. This makes it suitable for real-time applications like text editors, terminal-based games, and more.
//!
//! Each backend handles raw mode differently, so the behavior may vary depending on the backend
//! being used. Be sure to consult the backend's specific documentation for exact details on how it
//! implements raw mode.
```

src/backend/mod.rs
```diff
@@ -1,30 +1,57 @@
...
+//! # Alternate Screen
+//!
+//! The alternate screen is a screen that is hidden when not in use. When an application exits, the
+//! terminal will switch back to the main screen and the contents of the alternate screen will be
+//! cleared. This is useful for applications that want to display their own output without having to
+//! manage the contents of the terminal.
+//!
+//! Each backend handles the alternate screen differently, so the behavior may vary depending on the
+//! backend being used.
```
kdheepak, line 42:
```suggestion
//! # Alternate Screen
//!
//! The alternate screen is a separate buffer that some terminals provide, distinct from the main screen.
//! When activated, the terminal will display the alternate screen, hiding the current content of the
//! main screen. Applications can write to this screen as if it were the regular terminal display, but
//! when the application exits, the terminal will switch back to the main screen, and the contents of
//! the alternate screen will be cleared. This is useful for applications like text editors or terminal games
//! that want to use the full terminal window without disrupting the command line or other terminal content.
//! 
//! This creates a seamless transition between the application and the regular terminal session, as the
//! content displayed before launching the application will reappear after the application exits.
//!
//! Note that not all terminal emulators support the alternate screen, and even those that do may handle it
//! differently. As a result, the behavior may vary depending on the backend being used. Always consult the
//! specific backend's documentation to understand how it implements the alternate screen.
```

src/backend/mod.rs
```diff
@@ -1,30 +1,57 @@
...
+//!
+//! # Mouse Capture
+//!
+//! Mouse capture is a mode where the terminal captures mouse events and sends them to the
+//! application. This is useful for applications that want to process mouse events themselves.
+//!
+//! Each backend handles mouse capture differently, so the behavior may vary depending on the
+//! backend being used.
```
kdheepak, line 50:
```suggestion
//! # Mouse Capture
//!
//! Mouse capture is a mode where the terminal captures mouse events such as clicks, scrolls, and
//! movement, and sends them to the application as special sequences or events. This enables the
//! application to handle and respond to mouse actions, providing a more interactive and graphical
//! user experience within the terminal. It's particularly useful for applications like terminal-based
//! games, text editors, or other programs that require more direct interaction from the user.
//!
//! Each backend handles mouse capture differently, with variations in the types of events that can be
//! captured and how they are represented. As such, the behavior may vary depending on the backend
//! being used, and developers should consult the specific backend's documentation to understand how it
//! implements mouse capture.
```

src/backend/crossterm.rs
joshka: I started with a perspective that you were right - it's odd to panic during drop, but I think I've convinced myself that panicking is better.

The times when this will fail are likely the same as when enabling / entering fail, so it's likely that the app should have failed before the object is created and so this will never be hit in practice. But in a total edge case when this happens (likely on a CI where the app has intentionally ignored a previous error), we likely want the panic to happen because it lets the app's panic hook actually do something with it (like log to a file). Writing anything to the screen is a problem if the app is in raw mode (newlines don't get printed), or it's on the alternate screen (the output disappears altogether)

I had a brief play to look at how a backtrace for this would look like, and whether it would have any useful info that could help diagnose the issue.

```rust
    #[test]
    fn drop_simulated_failure() {
        let mut backend = CrosstermBackend::on_stdout().unwrap();
        let _ = backend.clear();
        eprintln!("foo");
    }
```
```rust
    fn drop(&mut self) {
        // snip actual impl
        Err::<(), &str>("").expect("simulated failure in drop");
    }
```
`RUST_BACKTRACE = 1 cargo test`
```
---- backend::crossterm::tests::drop_simulated_failure stdout ----
foo
thread 'backend::crossterm::tests::drop_simulated_failure' panicked at 'simulated failure in drop: ""', src/backend/crossterm.rs:330:29
stack backtrace:
   0: rust_begin_unwind
             at /rustc/8ede3aae28fe6e4d52b38157d7bfe0d3bceef225/library/std/src/panicking.rs:593:5
   1: core::panicking::panic_fmt
             at /rustc/8ede3aae28fe6e4d52b38157d7bfe0d3bceef225/library/core/src/panicking.rs:67:14
   2: core::result::unwrap_failed
             at /rustc/8ede3aae28fe6e4d52b38157d7bfe0d3bceef225/library/core/src/result.rs:1651:5
   3: core::result::Result<T,E>::expect
             at /rustc/8ede3aae28fe6e4d52b38157d7bfe0d3bceef225/library/core/src/result.rs:1033:23
   4: <ratatui::backend::crossterm::CrosstermBackend<W> as core::ops::drop::Drop>::drop
             at ./src/backend/crossterm.rs:330:9
   5: core::ptr::drop_in_place<ratatui::backend::crossterm::CrosstermBackend<std::io::stdio::Stdout>>
             at /rustc/8ede3aae28fe6e4d52b38157d7bfe0d3bceef225/library/core/src/ptr/mod.rs:497:1
   6: ratatui::backend::crossterm::tests::drop_simulated_failure
             at ./src/backend/crossterm.rs:562:5
   7: ratatui::backend::crossterm::tests::drop_simulated_failure::{{closure}}
             at ./src/backend/crossterm.rs:558:33
   8: core::ops::function::FnOnce::call_once
             at /rustc/8ede3aae28fe6e4d52b38157d7bfe0d3bceef225/library/core/src/ops/function.rs:250:5
   9: core::ops::function::FnOnce::call_once
             at /rustc/8ede3aae28fe6e4d52b38157d7bfe0d3bceef225/library/core/src/ops/function.rs:250:5
```
The info this adds over logging is where the drop occurred and the call pattern that got us there. I think that's probably a reasonable thing to happen.

src/backend/mod.rs
joshka: This is good. We probably need to add something about how handling input is not really part of Ratatui's scope, and so while we provide the means to enable this, we don't provide any code that does anything useful for the mouse capture mode.

src/backend/crossterm.rs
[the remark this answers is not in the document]
joshka: > Maybe `on_stderr_tui()`?

I'm not sure that really communicates anything about defaults better.

Happy to brainstorm some more ideas on this if you'd like. I'd start with what it is that we want to configure and then design something that handles the normal use cases pretty seamlessly:

Options:
- Backend: Crossterm / Termion / Termwiz
- Buffer choice: Stdout / Stderr / (any impl Write) / no choice for Termwiz right now
- Raw mode: yes / no. (Termion requires this to construct, Termwiz always enables this)
- Alternate Screen: yes / no. (Termion is optional, but changes the buffer type, Termwiz always enables this)
- Terminal Viewport type (Inline, Fixed, Fullscreen)

The two most common use cases (due to compatibility and the types of apps people write) are:
- Full screen Crossterm on stdout with raw mode and alternate screen
- Full screen Crossterm on stderr with raw mode and alternate screen

I really wanted to find something that does the most useful thing in one call here. Perhaps a proper builder approach would be better - that way we can defer the failure check to `build()` and the actual `default()` call never fails. E.g.:

```rust
let backend = CrosstermBackendBuilder::<Stdout>::default().build()?
```
I don't like that because the generic makes it unweildy

Perhaps:
```rust
let terminal = TerminalBuilder::with_crossterm().on_stderr().build()?
```

Using this approach the choices that we make in the defaults are part of the builder and can be communicated as such, not part of the backend / terminal types. Not sure if this is better though and it's different from all the other builders in Ratatui (which is the reason I didn't give it much consideration).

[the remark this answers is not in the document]
kdheepak: > I really wanted to find something that does the most useful thing in one call here. 

I really like the functionality. I'm just wondering if there's a better way to communicate what is going on there.

> ```rust
> let terminal = TerminalBuilder::with_crossterm().on_stderr().build()?
> ```

Oddly enough, I actually like it.

I think where my comment is coming from is that `on_$io` doesn't seem to convey enough. `build()` tells me something is being "built" that needs to be "destroyed" which maps with switching to alternate screen and back. 

Just tossing some other ideas out there:

```rust
// Option 1:
let terminal = TerminalBuilder::with_crossterm().build(stderr=true)?
// Option 2:
let terminal = TerminalBuilder::with_crossterm().init(stdout=false)?
// Option 3:
let terminal = TerminalBuilder::with_crossterm().setup()? // defaults to stderr. If people want stdout they can build the old fashion way. 
// Option 4:
let terminal = TerminalBuilder::with_crossterm().tui_on_stderr()?
let terminal = TerminalBuilder::with_crossterm().tui_on_stdout()?
```

src/backend/crossterm.rs
kdheepak: > Writing anything to the screen is a problem if the app is in raw mode (newlines don't get printed), or it's on the alternate screen (the output disappears altogether)

Yeah, which is why I was thinking of using log::warn and then it's on the users to setup logging. 

Btw, is this how the backtrace looks without a panic hook?

src/backend/crossterm.rs
```diff
@@ -24,36 +23,310 @@ use crate::{
...
+    fn drop(&mut self) {
+        if self.raw_mode_enabled {
+            self.disable_raw_mode().expect("failed to disable raw mode");
```
kdheepak, line 319: Do you think the failure case here is ever possible for user to see in their terminal? Or is this something that might only happen in tests or CI?

If the former is possible, it'd be nice if we can add a message that says something to the effect of

> Type `reset` and hit enter to fix your terminal state.

joshka: Good idea. Will do. I'm unsure of this can be triggered in an interactive terminal - likely it is emulator specific?

README.md
```diff
@@ -9,7 +9,7 @@ project.
 [![Crates.io](https://img.shields.io/crates/v/ratatui?logo=rust&style=flat-square)](https://crates.io/crates/ratatui)
 [![License](https://img.shields.io/crates/l/ratatui?style=flat-square)](./LICENSE) [![GitHub CI
 Status](https://img.shields.io/github/actions/workflow/status/ratatui-org/ratatui/ci.yml?style=flat-square&logo=github)](https://github.com/ratatui-org/ratatui/actions?query=workflow%3ACI+)
-[![Docs.rs](https://img.shields.io/docsrs/ratatui?logo=rust&style=flat-square)](https://docs.rs/crate/ratatui/)  
+[![Docs.rs](https://img.shields.io/docsrs/ratatui?logo=rust&style=flat-square)](https://docs.rs/crate/ratatui/)
```
orhun, line 12: nit: trailing space is needed for breaking down the badges into two lines, but maybe there is a better way of doing it

src/backend/crossterm.rs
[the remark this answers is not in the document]
orhun: > let terminal = TerminalBuilder::with_crossterm().on_stderr().build()?

> let terminal = TerminalBuilder::with_crossterm().setup()?

These options look the most sane to me.

src/backend/crossterm.rs
orhun: `Drop` approach for cleaning up the terminal is definitely a good idea but yeah, I'm not sure about panicking here as well.

examples/block.rs
```diff
@@ -23,35 +15,11 @@ use ratatui::{
...
-    let backend = CrosstermBackend::new(stdout);
-    let terminal = Terminal::new(backend)?;
-    Ok(terminal)
-}
-
-fn restore_terminal(mut terminal: Terminal) -> Result<()> {
-    disable_raw_mode()?;
-    execute!(terminal.backend_mut(), LeaveAlternateScreen)?;
-    Ok(())
-}
```
orhun, line 52: Now that a restore/reset function is not needed anymore and everything is handled when the `Terminal` goes out of scope, how would that affect the backwards compatibility? In other words, do we need a major bump for these changes?

I saw that you mentioned this in the PR commments but I couldn't grasp how this is backwards compatible.

src/backend/crossterm.rs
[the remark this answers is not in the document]
joshka: Of those two I prefer

>     let terminal = TerminalBuilder::with_crossterm().on_stderr().build()?

Because setup()? seems weird and like it shouldn't return anything.

README.md
joshka: `\` works
https://github.github.com/gfm/#hard-line-break

examples/block.rs
joshka: I think we have a few breaking changes already, so that part of the question is moot.

In the initial code for this, I had it always leave alternate / disable raw mode, but changed to a flag for those because of the issues mentioned. The previous setup / restore code still works, it's just unnecessary if you create the backend using code that tracks the various things to cleanup. If you instead call crossterm::enable_raw_mode() etc. then things still behave in the same way and you need to do the normal things to cleanup.

To check for backwards compatibility, pull this PR locally and checkout the previous commit (before the changes for the examples). Every example should run fine without any changes (tested on my Mac). The ones that were most concerning were `inline` and `panic` as they're a little different from the rest. With inline I found that leaving the alternate screen cleared the text that was expected to be left. With panic, I mainly needed to find a good way to restore the terminal from in the panic hook as drop runs after the hook. Both things are fixed.

bacon.toml
joshka, line 1: *sigh* - unsure why the autoformatting of this changed. The only change is to line 77.

src/backend/crossterm.rs
```diff
@@ -64,7 +67,7 @@ use crate::{
 #[derive(Debug, Clone, Eq, PartialEq, Hash)]
 pub struct CrosstermBackend<W: Write> {
     /// The buffer being written to.
-    buffer: W,
+    writer: W,
```
joshka, line 70: Changed to avoid confusion with `Buffer`

src/backend/crossterm.rs
```diff
@@ -220,88 +223,28 @@ where
...
-    ///
-    /// [raw mode]: crate::backend#raw-mode
-    /// [alternate screen]: crate::backend#alternate-screen
-    pub fn on_stdout() -> Result<CrosstermBackend<Stdout>, io::Error> {
+    pub fn on_stdout() -> CrosstermBackend<Stdout> {
         CrosstermBackend::new(io::stdout())
-            .with_raw_mode()?
-            .with_alternate_screen()
-    }
-}
```
joshka: No longer does raw and alternate for this per feedback - the "easy mode" is moved to the builder.

src/backend/termion.rs
```diff
@@ -63,7 +63,7 @@ pub struct TermionBackend<W>
 where
     W: Write,
 {
-    stdout: W,
+    writer: W,
```
joshka, line 66: Changed because this can be stderr/stdout or one of the more custom wrappers.

src/terminal_builder.rs
joshka, line 1: This is the main change to review (it needs some docs)

src/terminal_builder.rs
```diff
@@ -0,0 +1,386 @@
...
+}
+
+#[derive(Debug)]
+pub struct TerminalBuilder;
+
+impl TerminalBuilder {
+    #[cfg(feature = "crossterm")]
+    pub fn crossterm_on_stdout() -> crossterm::TerminalBuilder<Stdout> {
+        Self::crossterm(stdout())
+    }
```
joshka, line 37: These methods could also just hang directly off the module (either `terminal_builder` or `terminal` instead of needing an unnecessary struct. I wouldn't mind feedback on that idea. It would make startup something like:
```rust
let terminal = ratatui::terminal::crossterm_on_stdout().build()?;
// or
let terminal = ratatui::terminal_builder::crossterm_on_stdout().build()?;
// or 
let terminal = ratatui::crossterm_on_stdout().build()?;
```

joshka (commented):
I've left a few comments across the new changes to help point in the direction of the new stuff.

joshka:
Rebased on main, and added terminal builder code. Addressed much of the feedback on this.
I'd say this is probably going to miss 0.23 release and be released in an alpha release sometime after than in order to avoid a last minute destabilization. I think it's fully backwards compatible, but I don't want to lock in the API just yet without some discussion.

README.md
joshka: addressed in [63128f](https://github.com/ratatui-org/ratatui/pull/280/files#diff-63128fb186b5e73ba20dfb73d335955be10a51c76f6bf598efc1ada3e173e5b9)

src/backend/crossterm.rs
joshka: Addressed in latest push - I run all the resets and collect the results.

I went with keeping the panic however as this allows the app to take action on the failure, whereas just logging does not.

src/backend/crossterm.rs
joshka: I left the panic in because it allows the app writer to take action as a result of the app failing. They can log the problem, notify, restart, whatever. Panicking allows the user to log, but logging doesn't allow the app to panic, so it's the worse of the two choices objectively.

haydenflinner: We should never panic with the expectation that it can be caught and acted on. This is what Results are for. The pattern I have seen elsewhere is that there is an explicit close method on the object, which can be called if the user is interested in the result. If not, Drop will call the close too, and simply ignore the failure. Logging would probably be fine too. But to panic on this case is IMO very unidiomatic Rust, especially when done just for convenience.

https://doc.rust-lang.org/book/ch09-03-to-panic-or-not-to-panic.html

First comment here:
https://www.reddit.com/r/rust/comments/9q3jqn/how_is_rust_safe_when_panics_can_happen_out_of/

joshka: Thank you for the feedback. In addition to the links provided, I some time reading a bunch of rust lang users forum posts regarding panic during drop, and I'm convinced that those who mentioned that we should not panic are correct. Particularly https://users.rust-lang.org/t/return-error-condition-from-drop-impl/4933 was the best argument for this.
So our options become:
- ignore errors during drop - termion does this right now
- eprintln / log / trace any errors
- write a destroy method that returns `Result` and don't do anything in drop if that's been called
- write methods on terminal that handle reverting things nicely. E.g. a slight modification from https://users.rust-lang.org/t/return-error-condition-from-drop-impl/4933/4?u=joshka:
```rust
impl Terminal {
    pub fn draw(&mut self, ...) { ... }
    fn finalize(self) -> Result { ... }

    pub fn with<F: FnOnce(&mut Self) -> Result>(func: F) -> Result {
        let mut terminal = Terminal { ... };
        try!(func(&mut terminal))
        terminal.finalize()
    }
}
```
I'm going to play a bit more with these ideas

src/backend/crossterm.rs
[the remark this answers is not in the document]
kdheepak: But I think it is common in rust to have builder methods with names like `setup()` that return.

Here's code from `tracing_subscriber`:

```rust
  let file_subscriber = tracing_subscriber::fmt::layer()
    .with_file(true)
    .with_line_number(true)
    .with_writer(log_file)
    .with_target(false)
    .with_ansi(false)
    .with_filter(tracing_subscriber::filter::EnvFilter::from_default_env());
```

The `config` crate has examples like this:

```rust
    let settings = Config::builder()
        // File::with_name(..) is shorthand for File::from(Path::new(..))
        .add_source(File::with_name("examples/glob/conf/00-default.toml"))
        .add_source(File::from(Path::new("examples/glob/conf/05-some.yml")))
        .add_source(File::from(Path::new("examples/glob/conf/99-extra.json")))
        .build()
        .unwrap();
```

`notcurses` has:

```
impl NotcursesBuilder {
    /// Returns a new default `NotcursesBuilder`, without banners.
    pub fn new() -> Self {
        Self::default()
    }

    /// Returns a `Notcurses` instance.
    pub fn build(self) -> Result<Notcurses> {
        Notcurses::lock_notcurses()?;
        let nc = unsafe { Nc::with_options(self.options.build())? };
        Ok(Notcurses {
            nc,
            options: self.options,
        })
    }
}
```

And [all the methods in `NotcursesBuilder`](https://github.com/dankamongmen/notcurses-rs/blob/4caf961306866ddc70f62d52be9083779b48d8f9/src/notcurses/builder.rs#L49) seem to be verbs.

What do you think about prefixing everything with `with`? e.g.:

```rust
let terminal = TerminalBuilder::with_crossterm().with_stderr().build()?
```

kdheepak:
Just wanted to say that after re-implementing `Drop`, `Deref` and `DerefMut` over and over again for a few projects (and in writing tutorial docs), my interest in this PR has grown a lot more.

joshka:
> Just wanted to say that after re-implementing `Drop`, `Deref` and `DerefMut` over and over again for a few projects (and in writing tutorial docs), my interest in this PR has grown a lot more.

Hehe - yeah. I think I need to chop some of the parts of this off again so that the surface area is smaller. I'm definitely convinced that we should not panic during Drop and have been meaning to come back and fix this.

joshka:
Closing out as stale for a minute. Will come back to work on it soon.

Issue #241, opened 2023-06-11
Reduce boilerplate on every `style()` call


joshka:
## Problem
There's a lot of boilerplate for common methods like `style()`:
```rust
Paragraph::new("foo").style(Style::default().fg(Color::Blue).bg(Color::Red).add_modifier(Modifier::BOLD);
```

## Solution
Make it easier to specify styles similar to Crossterm's [Stylize](https://docs.rs/crossterm/latest/crossterm/style/trait.Stylize.html)

```rust
Paragraph::new("foo").blue().on_red().bold();
```

I think this can probably be implemented across every widget that supports style by adding a `Styled` trait and implementing a `Stylize` trait that has all the right methods (in crossterm these are generated using macros rather than the repetition I've written here).

```
pub trait Styled {
    fn style(&self) -> Style;
    fn set_style(self, style: Style) -> Self;
}

pub trait Stylize {
    fn blue(self) -> Self;
    fn on_red(self) -> Self;
    fn bold(self) -> Self;
}

impl<T> Stylize for T
where
    T: Styled,
{
    fn blue(self) -> Self {
        let style = self.style().fg(Color::Blue);
        self.set_style(style)
    }

    fn on_red(self) -> Self {
        let style = self.style().bg(Color::Red);
        self.set_style(style)
    }

    fn bold(self) -> Self {
        let style = self.style().add_modifier(Modifier::BOLD);
        self.set_style(style)
    }
}

impl<'a> Styled for Paragraph<'a> {
    fn style(&self) -> Style {
        self.style
    }
    fn set_style(self, style: Style) -> Paragraph<'a> {
        Paragraph::style(self, style)
    }
}
```

## Alternatives
Perhaps:
```rust
Paragraph::new("foo").stylize().blue().on_red().bold();
```
Not sure about other options that would make sense.

## Questions
1. How should widgets that have multiple style methods (for different parts of the widget) act?

## Additional context
This is a RFC. There are probably better ways of doing this, and I want to consider them before implementing this approach. Perhaps there's some magic that we can do using `AsRef` / `Deref` for this?

joshka:
That seems too close to what we would be moving away from. It just drops the style(style::default() and imports Color::* etc.

kdheepak:
I really like the idea of decreasing the boilerplate:

I think what could be really cool is some kind of "Styled String" macro that expands to the appropriate styled contented.

[...]

PR #283, opened 2023-06-29
feat: stylization shorthands

samyosm:
fix: #241

This PR adds the ability to use style shorthands to `str`, `Span`, and `Paragraph`.

## Examples
```rust
assert_eq!(
    "hello".red().on_blue().bold(),
    styled("hello", Style::default().fg(Color::Red).bg(Color::Blue).add_modifier(Modifier::BOLD))
)
```

### DIFF
--- a/src/text.rs
+++ b/src/text.rs
@@ -51,7 +51,7 @@ use std::{borrow::Cow, fmt::Debug};
 use unicode_segmentation::UnicodeSegmentation;
 use unicode_width::UnicodeWidthStr;
 
-use crate::style::Style;
+use crate::style::{Style, Styled};
 
 mod line;
 mod masked;
@@ -235,6 +235,17 @@ impl<'a> From<String> for Span<'a> {
 impl<'a> From<&'a str> for Span<'a> {
     fn from(s: &'a str) -> Span<'a> {
         Span::raw(s)
+    }
+}
+
+impl<'a> Styled for Span<'a> {
+    type Item = Span<'a>;
+    fn style(&self) -> Style {
+        self.style
+    }
+
+    fn set_style(self, style: Style) -> Self {
+        Self::styled(self.content, style)
     }
 }
 
--- a/src/widgets/paragraph.rs
+++ b/src/widgets/paragraph.rs
@@ -3,7 +3,7 @@ use unicode_width::UnicodeWidthStr;
 use crate::{
     buffer::Buffer,
     layout::{Alignment, Rect},
-    style::Style,
+    style::{Style, Styled},
     text::{StyledGrapheme, Text},
     widgets::{
         reflow::{LineComposer, LineTruncator, WordWrapper},
@@ -194,6 +194,18 @@ impl<'a> Widget for Paragraph<'a> {
                 break;
             }
         }
+    }
+}
+
+impl<'a> Styled for Paragraph<'a> {
+    type Item = Paragraph<'a>;
+
+    fn style(&self) -> Style {
+        self.style
+    }
+
+    fn set_style(self, style: Style) -> Self::Item {
+        self.style(style)
     }
 }
 
--- a/examples/block.rs
+++ b/examples/block.rs
@@ -8,8 +8,7 @@ use crossterm::{
 use ratatui::{
     backend::{Backend, CrosstermBackend},
     layout::{Alignment, Constraint, Direction, Layout},
-    style::{Color, Modifier, Style},
-    text::Span,
+    style::{Color, Style, Stylize},
     widgets::{block::title::Title, Block, BorderType, Borders, Padding, Paragraph},
     Frame, Terminal,
 };
@@ -80,24 +79,13 @@ fn ui<B: Backend>(f: &mut Frame<B>) {
 
     // Top left inner block with green background
     let block = Block::default()
-        .title(vec![
-            Span::styled("With", Style::default().fg(Color::Yellow)),
-            Span::from(" background"),
-        ])
-        .style(Style::default().bg(Color::Green));
+        .title(vec!["With".yellow(), " background".into()])
+        .on_green();
     f.render_widget(block, top_chunks[0]);
 
     // Top right inner block with styled title aligned to the right
-    let block = Block::default().title(
-        Title::from(Span::styled(
-            "Styled title",
-            Style::default()
-                .fg(Color::White)
-                .bg(Color::Red)
-                .add_modifier(Modifier::BOLD),
-        ))
-        .alignment(Alignment::Right),
-    );
+    let block = Block::default()
+        .title(Title::from("Styled title".white().on_red().bold()).alignment(Alignment::Right));
     f.render_widget(block, top_chunks[1]);
 
     // Bottom two inner blocks
--- a/examples/canvas.rs
+++ b/examples/canvas.rs
@@ -12,9 +12,8 @@ use crossterm::{
 use ratatui::{
     backend::{Backend, CrosstermBackend},
     layout::{Constraint, Direction, Layout, Rect},
-    style::{Color, Style},
+    style::{Color, Stylize},
     symbols::Marker,
-    text::Span,
     widgets::{
         canvas::{Canvas, Map, MapResolution, Rectangle},
         Block, Borders,
@@ -177,11 +176,7 @@ fn ui<B: Backend>(f: &mut Frame<B>, app: &App) {
                 color: Color::White,
                 resolution: MapResolution::High,
             });
-            ctx.print(
-                app.x,
-                -app.y,
-                Span::styled("You are here", Style::default().fg(Color::Yellow)),
-            );
+            ctx.print(app.x, -app.y, "You are here".yellow());
         })
         .x_bounds([-180.0, 180.0])
         .y_bounds([-90.0, 90.0]);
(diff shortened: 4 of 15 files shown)


joshka (commented):
I'd like to see how this works for styled things outside this module. What imports are needed to make it work, and what do the calls for real-ish ui code look like? E.g. https://github.com/tui-rs-revival/ratatui/issues/241#issuecomment-1610620895)

Can you please add some integration test for this in `tests/stylize.rs`?

I suspect that it might make sense to move the `impl Styled` code to e.g. `paragraph.rs` etc.

src/text/stylized.rs
```diff
@@ -0,0 +1,478 @@
...
+    }
+    fn on_light_cyan(self) -> T
+    where
+        Self: Sized,
+    {
+        self.on_color(Color::LightCyan)
+    }
+
+    // Styles
+    fn normal(self) -> T;
```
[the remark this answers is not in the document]
samyosm: You're right, `reset` would make more sense.

src/text/stylized.rs
```diff
@@ -0,0 +1,478 @@
...
+pub trait Stylize<'a, T> {
+    // Foreground colors
+    fn color<S: Into<Color>>(self, color: S) -> T;
+
+    fn black(self) -> T
+    where
+        Self: Sized,
+    {
+        self.color(Color::Black)
+    }
```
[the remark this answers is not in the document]
samyosm: I copied _a little_ from [crossterm](https://github.com/crossterm-rs/crossterm/blob/1efdce7ef63cba6992729db5f22262a60936fa8b/src/style/stylize.rs#L5-L56) on this one. Would you like me to mention it in the code?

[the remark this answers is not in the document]
joshka: If it's a direct copy, then it is covered the license of crossterm (and do all the things required by that - attribution mainly)

samyosm:
> Can you please add some integration test for this in tests/stylize.rs?

All the necessary tests are already in `src/text/stylized.rs`. Would it not be better to modify the already existing tests in the `tests` folder to use this new feature instead? Not just to put it to the test, but also to advocate its usage?

joshka:
>All the necessary tests are already in src/text/stylized.rs. Would it not be better to modify the already existing tests in the tests folder to use this new feature instead? Not just to put it to the test, but also to advocate its usage?

My thinking behind this was that usually I'd organize tests which deal with just the things defined in a module within that module (e.g. here it's just the traits, which don't actually need tests of themselves - which probably means that they could just be added in style.rs perhaps?) Whereas the stylized methods are only really part of the {`Text`, `Paragraph`, `str` ... } when the traits are also imported. So we could add them to `text.rs`, `paragraph.rs` etc. but I think it probably makes sense to have them in a single place to make it easy to see how they work.

Does that make sense or am I overthinking it?

src/text/stylized.rs
```diff
@@ -0,0 +1,224 @@
...
+    fn set_style(self, style: Style) -> Self::Item;
+}
+// Workaround for https://github.com/rust-lang/rust/issues/78835
+macro_rules! calculated_docs {
+    ($(#[doc = $doc:expr] $item:item)*) => { $(#[doc = $doc] $item)* };
+}
+
+// https://github.com/crossterm-rs/crossterm/blob/1efdce7ef63cba6992729db5f22262a60936fa8b/src/style/stylize.rs#L5-L56
+macro_rules! stylize_method {
+    ($method_name:ident Modifier::$modifier:ident) => {
```
joshka, line 17: It could be worth using https://github.com/dtolnay/paste#case-conversion and https://github.com/dtolnay/paste#pasting-documentation-strings to simplify this to e.g.:
```rust
color_method!(red);
...
modifier_method(bold);
...
```

(Not a blocker, just small simplification of the code)

src/text/stylized.rs
```diff
@@ -0,0 +1,224 @@
...
+    stylize_method!(reversed Modifier::REVERSED);
+    stylize_method!(hidden Modifier::HIDDEN);
+    stylize_method!(crossed_out Modifier::CROSSED_OUT);
+}
+
+impl<'a, T, U> Stylize<'a, T> for U
+where
+    U: Styled<Item = T>,
+{
+    fn color<S: Into<Color>>(self, color: S) -> T {
```
joshka, line 113: I think this probably still makes sense to be named `fg`, as users will see method this in the type's autocomplete when they import Stylized. Perhaps add some docs on this?

src/text/stylized.rs
```diff
@@ -0,0 +1,224 @@
...
+        let style = self.style().fg(color.into());
+        self.set_style(style)
+    }
+
+    fn modifier(self, modifier: Modifier) -> T {
+        let style = self.style().add_modifier(modifier);
+        self.set_style(style)
+    }
+
+    fn on_color(self, color: Color) -> T {
```
joshka, line 123: And likewise this `bg`

src/text.rs
```diff
@@ -238,6 +239,17 @@ impl<'a> From<&'a str> for Span<'a> {
...
+    type Item = Span<'a>;
+    fn style(&self) -> Style {
+        self.style
+    }
+
+    fn set_style(self, style: Style) -> Self {
+        Self::styled(self.content, style)
+    }
+}
+
```
joshka, line 252: Would making a `style_mut()` method instead work for this? It's a pattern that I've seen a lot, and might simplify the code below to something like:
```rust
self.style_mut().fg(color.into());
self
```

src/text/stylized.rs
```diff
@@ -0,0 +1,224 @@
...
+/// };
+///
+/// assert_eq!(
+///    "hello".red().on_blue().bold(),
+///     Span::styled("hello", Style::default().fg(Color::Red).bg(Color::Blue).add_modifier(Modifier::BOLD))
+/// )
+pub trait Stylize<'a, T>: Sized {
+    // Colors
+    fn color<S: Into<Color>>(self, color: S) -> T;
+    fn on_color(self, color: Color) -> T;
```
joshka, line 73: Same here for fg/bg

src/widgets/paragraph.rs
```diff
@@ -197,6 +197,18 @@ impl<'a> Widget for Paragraph<'a> {
     }
 }
 
+impl<'a> Styled for Paragraph<'a> {
+    type Item = Paragraph<'a>;
+
+    fn style(&self) -> Style {
```
joshka, line 203: Please add some tests

joshka (changes requested):
This is looking pretty good.

I'd like some of the other devs to sanity check this (ping @sayanarijit @orhun), but I think it will be good once the missing tests and docs are added.

When adding the tests, have a think about the user of this and what the imports look like. I  think it might make more sense for a use to find and import this feature as `style::Stylize" instead of `text::Stylize`. If so, let's move the entire stylized file under the style mod and re-export the trait(s).

src/text/stylized.rs
```diff
@@ -0,0 +1,224 @@
+use super::Span;
+use crate::style::{Color, Modifier, Style};
+
+pub trait Styled {
+    type Item;
+
+    fn style(&self) -> Style;
+    fn set_style(self, style: Style) -> Self::Item;
+}
+// Workaround for https://github.com/rust-lang/rust/issues/78835
```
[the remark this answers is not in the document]
samyosm: I don't think so. For some reason, without it, rustfmt doesn't work properly.

SLASHLogin:
If this is the preferred way of stylizing text, I think it would be nice to convert examples to use those changes. This would also show how the changes would look like from the user's perspective.

src/text.rs
samyosm: I'm not sure how to implement that

joshka:
```rust
fn style_mut(&mut self) -> &mut Style {
  &mut self.style;
}
```

samyosm: I was talking about the trait, sorry for not being clear.

joshka: The `style_mut()` change is one that I'm not 100% sure is possible either (it's probably some sort of funky combination of defining `From<...> for Styled` and then `Stylized` taking an `Into<Styled>` instead. Not sure it's a big deal though.

joshka:
Looking at the examples convinces me that this is really useful (and that `stylize` does indeed make sense being in the style module).

What documentation is necessary to help users find and understand this feature in the generated docs? (Happy to make that another PR if that would work better?)

joshka (approved):

src/text.rs
samyosm: I'm not sure if having a reference only to the style works. Isn't it necessary to also have a reference to the widget (`Self::Item`)? (for the `Styled` trait)

samyosm:
Though styling is introduced [here](https://docs.rs/ratatui/latest/ratatui/style/struct.Style.html), I think I will add a short explanation for using style and the shorthands [here](https://docs.rs/ratatui/latest/ratatui/style/index.html) instead, at the module level right such that a user sees it right after clicking the Style module link in the homepage.

joshka (approved):

joshka:
> By the way, should I advocate the use of style variables or not?

It's neither here nor there whether people use inline, variables or the easier methods. Each has their pros and cons.

joshka:
Thanks for your effort in this.

PR #286, opened 2023-06-30
chore(features): enable building with all-features

joshka:
Before this change, it wasn't possible to build all features and all
targets at the same time, which prevents rust-analyzer from working
for the whole project.

Adds a bacon.toml file to the project, which is used by bacon
https://dystroy.org/bacon/

Configures docs.rs to show the feature flags that are necessary to
make modules / types / functions available.


joshka:
E.g.:
image

E.g:
image

joshka:
One of the big benefits of this is that it reduces the need to do specific CI builds for each backend, and instead we can just use https://github.com/taiki-e/cargo-hack to do all features / feature power sets, while just building once with all-features locally for speed.

sophacles (approved):

PR #288, opened 2023-07-01
feat(barchart): enable barchart groups

karthago1:
I modified the barchart to support adding multiple bars from different data set
see the the following image:

barchart

if this feature is needed, then I can continue and write some unit tests


SLASHLogin:
I'm not sure how this compares to current barchart, implementation wise, but maybe it could be solved by adding a method/flag to the current barchart, instead of introducing a new widget?

karthago1:
true I changed the old barchart. it should be backwards compatible

joshka:
I like the idea and direction - perhaps consider documenting the user's perspective a little before continuing into the code. (It's important to tell us what you want to do, so that when it's complete, it's obvious).

I notice that this adds a new "Group" feature, but "Group" only appears in the API once. How would you design the API so that it made this feature more explicitly obvious?

I think it's probably possible to keep the single `data()` function, and expect the user to call it multiple times rather than adding a new `add_data()` function which is confusingly similar.

Can you please make an issue for this to discuss the functionality of the new feature? (Would be nice to get a few more eyes on it).

karthago1:
> I think it's probably possible to keep the single data() function, and expect the user to call it multiple times rather than adding a new add_data() function which is confusingly similar.

true. the problem with `data()`, that it passes always one dataset with the labels. the labels should be the same for all datasets. so if the user calls `data()` twice. the first time I can store the labels. and the second call should I ignore them?

possible Solutions:
1. remove `data()` and rename `add_data()` to `data()`. this is not backwards compatible solution. labels can be passed by the function `labels()` instead of the old `data()`
2. maybe I can modify `data()` to pass all datasets at once and stay backwards compatible too. I will check this possibility
3. keep the same `data()` and add new function `add_data()` which adds only one data set. (current solution)

the following 2 code sections produces the same result:
```
BarChart::default()
        .data(&[("Mar", 9), ("Apr", 12), ("May", 5), ("Jun", 8)])
        .add_data(&[1, 2, 3, 4])
        .group_gap(4);
```

```
BarChart::default()
        .add_data(&[9, 12, 5, 8])
        .add_data(&[1, 2, 3, 4])
        .group_gap(4)
        .labels(&["Mars", "Apr", "May", "Jun"]);
```


> Can you please make an issue for this to discuss the functionality of the new feature? (Would be nice to get a few more eyes on it).

done

tests/widgets_barchart.rs
```diff
@@ -36,7 +37,67 @@ fn widgets_barchart_not_full_below_max_value() {
         "│       █████████████████████│",
         "│       █████████████████████│",
         "│       ██50█████99█████100██│",
-        "│empty  half   almost full   │",
+        "│ empty  half  almost  full  │",
```
karthago1, line 40: @joshka the only thing which is not backwards compatible now is this.

by default the labels are centered. I see this as an improvement.

joshka: I agree that this is better. It's a small enough breaking change that it doesn't sound like anyone would worry about it.

Would you mind splitting out this one change outside of the barchart group PR?
Doing it that way will mean we can note the breaking change separately in the changelog.
E.g. something like "feat(barchart)!: center labels by default"

src/widgets/barchart/mod.rs
joshka, line 1: Can you please add a bit of information to the doc comment (and make sure to take a look at how it renders). Consider how the user might discover how groups work from the documentation.

src/widgets/barchart/mod.rs
```diff
@@ -122,12 +120,39 @@ impl<'a> BarChart<'a> {
...
+    }
+
     pub fn style(mut self, style: Style) -> BarChart<'a> {
         self.style = style;
         self
     }
 }
 
+impl<'a> BarChart<'a> {
+    fn remove_unvisible_groups_and_bars(&mut self, mut width: u16) {
```
joshka, line 135: This method doesn't have an obvious reason to exist (which makes it difficult to judge whether it works correctly and difficult to maintain). Can you please write a doc comment that helps clarify why it exists and what it should do.

src/widgets/barchart/mod.rs
```diff
@@ -122,12 +120,39 @@ impl<'a> BarChart<'a> {
...
+                    self.data.truncate(group_index + 1);
+                }
+                break;
+            }
+        }
+    }
+}
+
 impl<'a> Widget for BarChart<'a> {
     fn render(mut self, area: Rect, buf: &mut Buffer) {
```
joshka, line 157: This method started out large and got almost twice as big, which will make it twice as difficult to change in the future (likely more than twice as the number of paths through the method increases non linearly). If you were to describe the responsibilities of this method or describe the algorithm, would there be easy parts of it to refactor towards something easier to maintain?

It might be helpful to add some unit tests of the the new functionality so that any refactoring doesn't introduce bugs.

I haven't reviewed the code changes here yet, because I'd expect this to change pretty wildly if you refactor.

src/widgets/barchart/group.rs
joshka, line 1: This definitely needs some comments

src/widgets/barchart/mod.rs
```diff
@@ -45,40 +49,34 @@ pub struct BarChart<'a> {
...
 }
 
 impl<'a> BarChart<'a> {
-    pub fn data(mut self, data: &'a [(&'a str, u64)]) -> BarChart<'a> {
-        self.data = data;
-        self.values = Vec::with_capacity(self.data.len());
-        for &(_, v) in self.data {
-            self.values.push(format!("{v}"));
-        }
+    pub fn data(mut self, data: impl Into<Group<'a>>) -> BarChart<'a> {
```
joshka, line 78: Given that this is the core method for providing data, and we've changed it from accepting the (str, data) tuple array, it would be good to help the user understand their options for providing `Into<Group>` values.

joshka (changes requested):
Hey there. I have some concerns around the size of the new render method, and the lack of docs / new unit tests. I'm liking where this is heading though.

mindoodoo (changes requested):

mindoodoo:
(disregard the second review, for some reason my review got posted twice)

src/widgets/barchart/mod.rs
```diff
@@ -122,12 +120,39 @@ impl<'a> BarChart<'a> {
...
 }
 
+impl<'a> BarChart<'a> {
+    fn remove_unvisible_groups_and_bars(&mut self, mut width: u16) {
+        for group_index in 0..self.data.len() {
+            let n_bars = self.data[group_index].bars.len() as u16;
+            let bars_width = n_bars * self.bar_width + n_bars.saturating_sub(1) * self.bar_gap;
+
+            if width > bars_width {
+                width = width.saturating_sub(bars_width + self.group_gap);
```
mindoodoo, line 141: You seem to be underestimating the width required to render the bar chart by one for each group. This results in the bar chart rendering over the border when resizing the terminal.

The following suggestion would fix it, but I haven't looked too much at `render_widget` so this might be something you want to change there.

```suggestion
                width = width.saturating_sub(bars_width + self.group_gap + 1);
```

examples/barchart.rs
```diff
@@ -128,35 +165,90 @@ fn ui<B: Backend>(f: &mut Frame<B>, app: &App) {
...
+        .group_gap(3);
+
+    for group in groups {
+        barchart = barchart.data(group)
+    }
+
+    f.render_widget(barchart, area);
+
+    const LEGEND_HEIGHT: u16 = 6;
+    if area.height >= LEGEND_HEIGHT {
```
mindoodoo, line 215: You need to add a similar check for the width if you want to avoid panicking.

mindoodoo (changes requested):

src/widgets/barchart/mod.rs
karthago1: In the latest commit I moved the code, which handles printing the bars to an extra function `print_bars`.

maybe it improves the readability a little bit. if it is a good idea, then I can squash commit it with the rest of the changes

mindoodoo (commented):
Moving the bar printing into it's own method helped quite a bit with readability. Dropped a couple thoughts on how `render` could be further split up / shortened.

Additionally you could also move the label printing (group and bar) in another (or other) method(s).

Finally you have a little bug in your example where the bar colors are applied to parts of the legend's border.

src/widgets/barchart/mod.rs
```diff
@@ -0,0 +1,705 @@
...
+        // remove unvisible groups and bars, since we don't need to print them
+        self.remove_unvisible_groups_and_bars(chart_area.width);
+
+        {
+            let bars_area = Rect {
+                height: chart_area.height - chart_offset,
+                ..chart_area
+            };
+            self.print_bars(buf, bars_area, max);
+        }
```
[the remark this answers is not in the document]
karthago1: just to limit the lifetime of `bars_area`. so you don't think, it is used later.

if you don't like it, then I can remove the block. but I think, it doesn't represent any issue.

src/widgets/barchart/mod.rs
```diff
@@ -0,0 +1,705 @@
...
+            .max
+            .unwrap_or_else(|| {
+                self.data
+                    .iter()
+                    .map(|t| {
+                        t.bars
+                            .iter()
+                            .max_by_key(|v| v.value)
+                            .map(|v| v.value)
+                            .unwrap_or_default()
```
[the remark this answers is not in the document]
karthago1: done

src/widgets/barchart/mod.rs
```diff
@@ -0,0 +1,705 @@
...
+    pub fn style(mut self, style: Style) -> BarChart<'a> {
+        self.style = style;
+        self
+    }
+}
+
+impl<'a> BarChart<'a> {
+    /// Check the bars, which fits inside the available space and removes
+    /// the bars and the groups, which are outside of the available space.
+    fn remove_unvisible_groups_and_bars(&mut self, mut width: u16) {
```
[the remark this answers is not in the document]
karthago1: done

src/widgets/barchart/mod.rs
```diff
@@ -0,0 +1,705 @@
...
+
+        {
+            let bars_area = Rect {
+                height: chart_area.height - chart_offset,
+                ..chart_area
+            };
+            self.print_bars(buf, bars_area, max);
+        }
+
+        // print labels and values on one go
```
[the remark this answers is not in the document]
karthago1: done

src/widgets/barchart/mod.rs
```diff
@@ -0,0 +1,705 @@
...
+                    let width = value_label.len() as u16;
+                    if width < self.bar_width {
+                        buf.set_string(
+                            bar_x + (self.bar_width.saturating_sub(value_label.len() as u16) >> 1),
+                            chart_area.bottom() - y_bottom_offset,
+                            value_label,
+                            bar.value_style.patch(self.value_style),
+                        );
+                    }
+                }
```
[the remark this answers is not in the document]
karthago1: yes. but the loops are very different. I introduced 2 private methods to the `Bar`:  `render_label` and `render_value`

so the render function of `BarChart` is now shorter and I hope it is enough.

I renamed also the function `print_bars` to `render_bars`.

karthago1:
> Finally you have a little bug in your example where the bar colors are applied to parts of the legend's border.

this is feature not a bug :smile: .  But it is fixed. (there still is another hidden small bug in the legend :smile_cat:, but this MR is about the BarChart )


I renamed `Group` to `BarGroup`, because I think `Group` is too general term

src/widgets/barchart/mod.rs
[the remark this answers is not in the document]
mindoodoo: Ah ok I guess that makes sense. I'm good with either way though I would lean slightly towards removing it for consistency with the rest of the code base. At your discretion :)

Out of curiosity would love to know what @joshka and @orhun think about it. But definitely non blocking.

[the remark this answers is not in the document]
joshka: It does look a little odd in that spot - I can see why @mindoodoo called it out.
This seems like a problem caused by the method being 80 lines rather than something smaller.

How about the following refactoring of this method:
```rust
impl<'a> Widget for BarChart<'a> {
    fn render(mut self, mut area: Rect, buf: &mut Buffer) {

        self.render_block(&mut area, buf);

        if self.data.is_empty() {
            return;
        }

        let label_height = self.label_height();
        if area.height <= label_height {
            return;
        }

        let max = self.maximum_data_value();

        // remove invisible groups and bars, since we don't need to print them
        self.remove_invisible_groups_and_bars(area.width);

        let bars_area = Rect {
            height: area.height - label_height,
            ..area
        };
        self.render_bars(buf, bars_area, max);
        self.render_labels_and_values(area, label_height, buf);
    }
}

impl <'a> Barchart<'a> {
    /// renders the block if there is one and updates the area to the inner area
    fn render_block(&mut self, area: &mut Rect, buf: &mut Buffer) {
        if let Some(block) = self.block.take() {
            let inner_area = block.inner(*area);
            block.render(*area, buf);
            *area = inner_area
        }
    }

    /// Get the number of lines needed for the labels.
    ///
    /// The number of lines depends on whether we need to print the bar labels and/or the group labels.
    /// - If there are no labels, return 0.
    /// - If there are only bar labels, return 1.
    /// - If there are only group labels, return 1.
    /// - If there are both bar and group labels, return 2.
    fn label_height(&self) -> u16 {
        let has_group_labels = self.data.iter().any(|e| e.label.is_some());
        let has_data_labels = self
            .data
            .iter()
            .any(|e| e.bars.iter().any(|e| e.label.is_some()));

        // convert true to 1 and false to 0 and add the two values
        u16::from(has_group_labels) + u16::from(has_data_labels)
    }
```
(You can go further than this, but this helps immensely in communicating how this works to future maintainers.)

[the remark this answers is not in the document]
karthago1: done, thank you for the immense improvements.  I adapted all suggestion.

karthago1:
> The truncate approach seems more difficult to reason about as it changes fields outside of the method to communicate with other methods. (This is like Alice calling Bob on the phone to tell him something then calling Carol and asking her to ask Bob for the information just provided)

Preprocessing data before working on it, is very common approach. I don't think the Alice bob and Carol example is very accurate :smile: .

> I think I can sort of see how the code landed there, but the following general approach (using a remaining area Rect to render the bar / group into) might help make this code clearer:

you are suggesting a third method, which is in my opinion worse than the truncate and filtering methods, which are both very similar. here my comparison:

1. Truncate method

preprocess the data once. afterwards there is no chance that you are working on the wrong data, because they simply doesn't exist. There is in total 4 places where we must filter. (in the future maybe more)

2. Filter method

the original code had 2 loops, we need to filter twice. if someone adds a 3. loop, the chances are high, that he will forget about filtering.

with this patch we need to filter the bars too, which makes this method a little bit harder to implement. now we need to filter 2 nested loops in 2 different functions. I can't see any reason why this method is superior then preprocesing the data first

3. you suggestion (general approach)

every time you draw, you need to check, if there is enough space. I see 2 drawbacks. the first one: the check logic will needs to be duplicated at least twice. when printing the bars and later when printing the labels. 

the second drawback: this check will be executed  in `render_bars` many times (area height * number of groups * number of bars inside a group) (we have 3 nested loops) 

`render_labels_and_values` is similar. It contains 2 nested loops. 

I would rather to filter or preprocess the data, than checking in every iteration, whether there is enough space. and the problem get worse with every additional loop. (code duplication)


till now you improved the code immensely, the length of that method became very short. I think most developers will understand it.  :smile:

joshka:
> Preprocessing data before working on it, is very common approach. I don't think the Alice bob and Carol example is very accurate 😄 .

The concept that's at play here is https://en.wikipedia.org/wiki/Action_at_a_distance_%28computer_programming%29.
This implementation is ok only because the widget is consumed after rendering.

> you are suggesting a third method, which is in my opinion worse than the truncate and filtering methods, which are both very similar

My suggested approach seems to me like it would lead to a simpler and more obvious solution with generally shorter methods that just do one thing. The difference in my mind is doing things in batches (calculate, render bars, render labels) vs doing things one by one in a nested fashion. In the batch approach, you have to understand how each part affects the whole, whereas the individual nested approach the interaction between the pieces is right there.

> every time you draw, you need to check, if there is enough space. I see 2 drawbacks. the first one: the check logic will needs to be duplicated at least twice. when printing the bars and later when printing the labels.

This is a problem only because the rendering of the bars and labels is separate - if bars and their label / value are rendered at the same time then this isn't a problem.

> the second drawback: this check will be executed in render_bars many times (area height * number of groups * number of bars inside a group) (we have 3 nested loops)

This check happens already, it just happens within the delete invisible method.

src/widgets/barchart/mod.rs
joshka: Happy to fix this up as another PR I think

joshka (approved):
Approved Subject to @mindoodoo's review

mindoodoo (approved):
LGTM ! Thanks for the PR !

joshka:
Merged! Thanks for your contribution @karthago1

karthago1:
@joshka and @mindoodoo thank you too for your help and patience :smile: . I can't wait to switch to the crate.io ratatui version at work.

The use case we have is to show the operating hours of multiple temperature sensors in different temperature ranges:
it looks similar to the following:
```
          ████   ████                ████
          ████   ████                ████
     ████ ████   ████ ████           ████ ████
▄▄▄▄ ████ ████   ████ ████ ████      ████ ████ ████
████ ████ ████   ████ ████ ████ ...  ████ ████ ████ 
████ ████ ████   ████ ████ ████      ████ ████ ████
█50█ █60█ █90█   █90█ █60█ █50█      █90█ █60█ █50█
   -40..-31°C       -30..-21°C         150..159°C
```

the only think I wish to do, but I don't know if it is possible, is to write the numbers vertically. (instead of horizontally)
For now I made the bars very wide so I can print the values. but I have a lot bars and I need to scroll horizontally. I don't know if it is possible to write text horizontally with ratatui.

PR #296, opened 2023-07-02
perf(buffer): improve buffer diff perfomance

joshka:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -58,6 +58,10 @@ rand = "0.8"
 name = "paragraph"
 harness = false
 
+[[bench]]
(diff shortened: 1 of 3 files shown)


orhun:
Uhh, what exactly this PR is doing?

sayanarijit:
Need to try it on all the examples visually, when I get some time. Of if any of you have already tried, pls post your results here.

joshka:
> Need to try it on all the examples visually, when I get some time. Of if any of you have already tried, pls post your results here.

On an actual human detectable scale, this won't be something that you can notice (as far as I understand it).

PR #301, opened 2023-07-08
test(barchart): add unit tests

joshka:
Adds unit tests in advance of https://github.com/tui-rs-revival/ratatui/issues/294
These tests should help ensure that existing code that uses `Barchart` will continue to work with the grouped version.

Pinging @karthago1 for review

Has full line coverage on the `Barchart` type.


orhun (approved):

karthago1:
I updated the MR #288. it 100% compatible with these unit tests.
and  [I marked here](https://github.com/tui-rs-revival/ratatui/pull/288#discussion_r1257429718) the one thing which is not backwards compatible and I think it is an improvment.

PR #302, opened 2023-07-09
feat(examples): cursor can be moved left/right and character at cursor position can be deleted/inserted

BoolPurist:
# Changes

Added cursor navigation to the user input example. User can now move cursor left and right. In addition user can also insert/delete characters at the cursor position. 

## Demo

pr_demo

## Note

Could also put this demonstration into a new  example file if you prefer it this way.

### DIFF
--- a/examples/user_input.rs
+++ b/examples/user_input.rs
@@ -6,18 +6,20 @@ use std::{error::Error, io};
 /// started.
 ///
 /// This is a very simple example:
-///   * A input box always focused. Every character you type is registered
-///   here
-///   * Pressing Backspace erases a character
+///   * An input box always focused. Every character you type is registered
+///   here.
+///   * An entered character is inserted at the cursor position.
+///   * Pressing Backspace erases the left character before the cursor position
 ///   * Pressing Enter pushes the current input in the history of previous
-///   messages
+///   messages.
+/// **Note: ** as this is a relatively simple example unicode characters are unsupported and
+/// their use will result in undefined behaviour.
 use crossterm::{
     event::{self, DisableMouseCapture, EnableMouseCapture, Event, KeyCode, KeyEventKind},
     execute,
     terminal::{disable_raw_mode, enable_raw_mode, EnterAlternateScreen, LeaveAlternateScreen},
 };
 use ratatui::prelude::*;
-use unicode_width::UnicodeWidthStr;
 
 enum InputMode {
     Normal,
@@ -28,6 +30,8 @@ enum InputMode {
 struct App {
     /// Current value of the input box
     input: String,
+    /// Position of cursor in the editor area.
+    cursor_position: usize,
     /// Current input mode
     input_mode: InputMode,
     /// History of recorded messages
@@ -40,7 +44,62 @@ impl Default for App {
             input: String::new(),
             input_mode: InputMode::Normal,
             messages: Vec::new(),
+            cursor_position: 0,
         }
+    }
+}
+
+impl App {
+    fn move_cursor_left(&mut self) {
+        let cursor_moved_left = self.cursor_position.saturating_sub(1);
+        self.cursor_position = self.clamp_cursor(cursor_moved_left);
+    }
+
+    fn move_cursor_right(&mut self) {
+        let cursor_moved_right = self.cursor_position.saturating_add(1);
+        self.cursor_position = self.clamp_cursor(cursor_moved_right);
+    }
+
+    fn enter_char(&mut self, new_char: char) {
+        self.input.insert(self.cursor_position, new_char);
+
+        self.move_cursor_right();
+    }
+
+    fn delete_char(&mut self) {
+        let is_not_cursor_leftmost = self.cursor_position != 0;
+        if is_not_cursor_leftmost {
+            // Method "remove" is not used on the saved text for deleting the selected char.
+            // Reason: Using remove on String works on bytes instead of the chars.
+            // Using remove would require special care because of char boundaries.
+
+            let current_index = self.cursor_position;
+            let from_left_to_current_index = current_index - 1;
+
+            // Getting all characters before the selected character.
+            let before_char_to_delete = self.input.chars().take(from_left_to_current_index);
+            // Getting all characters after selected character.
+            let after_char_to_delete = self.input.chars().skip(current_index);
+
+            // Put all characters together except the selected one.
+            // By leaving the selected one out, it is forgotten and therefore deleted.
+            self.input = before_char_to_delete.chain(after_char_to_delete).collect();
+            self.move_cursor_left();
+        }
+    }
+
+    fn clamp_cursor(&self, new_cursor_pos: usize) -> usize {
+        new_cursor_pos.clamp(0, self.input.len())
+    }
+
+    fn reset_cursor(&mut self) {
+        self.cursor_position = 0;
+    }
+
+    fn submit_message(&mut self) {
+        self.messages.push(self.input.clone());
+        self.input.clear();
+        self.reset_cursor();
     }
 }
 
@@ -88,14 +147,18 @@ fn run_app<B: Backend>(terminal: &mut Terminal<B>, mut app: App) -> io::Result<(
                     _ => {}
                 },
                 InputMode::Editing if key.kind == KeyEventKind::Press => match key.code {
-                    KeyCode::Enter => {
-                        app.messages.push(app.input.drain(..).collect());
-                    }
-                    KeyCode::Char(c) => {
-                        app.input.push(c);
+                    KeyCode::Enter => app.submit_message(),
+                    KeyCode::Char(to_insert) => {
+                        app.enter_char(to_insert);
                     }
                     KeyCode::Backspace => {
-                        app.input.pop();
+                        app.delete_char();
+                    }
+                    KeyCode::Left => {
+                        app.move_cursor_left();
+                    }
+                    KeyCode::Right => {
+                        app.move_cursor_right();
                     }
                     KeyCode::Esc => {
                         app.input_mode = InputMode::Normal;
@@ -165,8 +228,9 @@ fn ui<B: Backend>(f: &mut Frame<B>, app: &App) {
             // Make the cursor visible and ask ratatui to put it at the specified coordinates after
             // rendering
             f.set_cursor(
-                // Put cursor past the end of the input text
-                chunks[1].x + app.input.width() as u16 + 1,
+                // Draw the cursor at the current position in the input field.
+                // This position is can be controlled via the left and right arrow key
+                chunks[1].x + app.cursor_position as u16 + 1,
                 // Move one line down, from the border to the input line
                 chunks[1].y + 1,
             )


examples/user_input.rs
```diff
@@ -35,6 +35,10 @@ enum InputMode {
 struct App {
     /// Current value of the input box
     input: String,
+    /// Position of cursor in the editor area.
+    /// It is updated if a character is inserted/deleted or if left or right
+    /// arrow is pressed.
+    cursor_in_input: usize,
```
mindoodoo, line 41:
```suggestion
    /// Position of cursor in the editor area.
    cursor_position: usize,
```

examples/user_input.rs
```diff
@@ -47,10 +51,67 @@ impl Default for App {
...
+            self.cursor_in_input = cursor_moved_left;
+        };
+    }
+
+    fn move_cursor_right(&mut self) {
+        if !self.is_cursor_rightmost() {
+            let cursor_moved_right = self.cursor_in_input + 1;
+            self.cursor_in_input = cursor_moved_right;
+        };
+    }
```
mindoodoo, line 72: Look into using `saturating_sub` and `saturating_add` for these functions. This along with other changes would eliminate the need for `is_cursor_leftmost` and `is_cursor_rightmost`.

examples/user_input.rs
```diff
@@ -47,10 +51,67 @@ impl Default for App {
...
+
+    fn enter_char_at_cursor(&mut self, new_char: char) {
+        if self.is_cursor_leftmost() {
+            self.input.push(new_char);
+        } else {
+            self.input.insert(self.cursor_in_input, new_char);
+        }
```
mindoodoo, line 79:
```suggestion
        self.input.insert(self.cursor_in_input, new_char);
```

examples/user_input.rs
```diff
@@ -47,10 +51,67 @@ impl Default for App {
...
+            let current_index = self.cursor_in_input;
+            let from_left_to_current_index = current_index - 1;
+
+            let before_char_to_delete = self.input.chars().take(from_left_to_current_index);
+            let after_char_to_delete = self.input.chars().skip(current_index);
+
+            self.input = before_char_to_delete.chain(after_char_to_delete).collect();
+
+            self.move_cursor_left();
+        }
```
mindoodoo, line 95:
```suggestion
        if self.cursor_in_input > 0 {
            self.input.remove(self.cursor_in_input - 1);
        }

        self.move_cursor_left();
```

examples/user_input.rs
```diff
@@ -47,10 +51,67 @@ impl Default for App {
...
+    }
+
+    fn is_cursor_leftmost(&self) -> bool {
+        self.cursor_in_input == 0
+    }
+    fn is_cursor_rightmost(&self) -> bool {
+        self.cursor_in_input == self.input.len()
+    }
```
mindoodoo, line 103: These can be removed

examples/user_input.rs
```diff
@@ -47,10 +51,67 @@ impl Default for App {
...
+
+    fn reset_cursor(&mut self) {
+        self.cursor_in_input = 0;
+    }
+
+    fn submit_message(&mut self) {
+        self.messages.push(self.input.drain(..).collect());
```
mindoodoo, line 110: I don't understand why you're pushing the message like this, why not simply push the string ?

examples/user_input.rs
```diff
@@ -47,10 +51,67 @@ impl Default for App {
...
+    }
+
+    fn move_cursor_right(&mut self) {
+        if !self.is_cursor_rightmost() {
+            let cursor_moved_right = self.cursor_in_input + 1;
+            self.cursor_in_input = cursor_moved_right;
+        };
+    }
+
+    fn enter_char_at_cursor(&mut self, new_char: char) {
```
mindoodoo, line 74:
```suggestion
    fn enter_char(&mut self, new_char: char) {
```

examples/user_input.rs
```diff
@@ -47,10 +51,67 @@ impl Default for App {
...
+        if self.is_cursor_leftmost() {
+            self.input.push(new_char);
+        } else {
+            self.input.insert(self.cursor_in_input, new_char);
+        }
+
+        self.move_cursor_right();
+    }
+
+    fn delete_char_at_cursor(&mut self) {
```
mindoodoo, line 84:
```suggestion
    fn delete_char(&mut self) {
```

examples/user_input.rs
```diff
@@ -173,7 +238,7 @@ fn ui<B: Backend>(f: &mut Frame<B>, app: &App) {
             // rendering
             f.set_cursor(
                 // Put cursor past the end of the input text
```
mindoodoo, line 240: Please update this comment

mindoodoo (changes requested):

examples/user_input.rs
joshka: (and clamp)

joshka:
Let's not get too much more complicated with this example. An example that uses the tui-textarea or tui-input libraries might be worth considering.

examples/user_input.rs
BoolPurist: I copied it from the earlier version. I agree that just pushing the string is simpler and better.

examples/user_input.rs
BoolPurist: In the moment of writing saturating_add seems to be nightly only.

https://doc.rust-lang.org/std/intrinsics/fn.saturating_add.html

Would that be okay for an example.

mindoodoo: You should be able to use the `saturating_add` implemented on [usize](https://doc.rust-lang.org/std/primitive.usize.html#method.saturating_add).

Also as @joshka pointed out, you'll also need to `clamp` the value.

examples/user_input.rs
BoolPurist: Using remove on Strings in Rust requires to pay attention to the char boundaries. 
If I try you suggestion here, the deletion of chars is quite janky. Sometimes it removes one or two.

examples/user_input.rs
BoolPurist: Do you mean that I should clamp the cursor position so that if it is to the most left, at 0, then it should go to the rightmost, at the end of the input field ?

mindoodoo: No, I mean clamp it so that it doesn't set the cursor position past the width of the input field.

examples/user_input.rs
mindoodoo: Oh I didn't think of that nice catch.

examples/user_input.rs
BoolPurist: I removed those functions

examples/user_input.rs
BoolPurist: I applied your simpler version of adding a char.

examples/user_input.rs
BoolPurist: Saturating_add  and saturating_sub are used instead of + 1/ -1.
Clamped is also preformed via num::clamp.

examples/user_input.rs
BoolPurist: Updated the comment

examples/user_input.rs
mindoodoo: As this is an example I'd suggest just cloning it. Wouldn't the `.collect()` allocate in this case anyways ?

mindoodoo:
> Let's not get too much more complicated with this example. An example that uses the tui-textarea or tui-input libraries might be worth considering.

This example in it's current state also has a couple bugs with unicode characters width, so might be interesting to look into how those crates handle unicode char width and if their solution might be better.

I'm unsure though about including examples using external crate widgets. Would you add another example alongside this one or would it take it's place ?

examples/user_input.rs
BoolPurist: Since clearning is needed, I would offer this version 

```rs
fn submit_message(&mut self) {
    self.messages.push(self.input.clone());
    self.input.clear();
    self.reset_cursor();
}
```

If you like, I apply this version.

BoolPurist: Applied the offered change.

sayanarijit:
Agree. Examples should be as simple as possible, and not deviate from demonstrating its primary use case.

Related/unrelated, we should probably have unicode handling built-in. I believe this falls under this crate's primary objectives.

BoolPurist:
I realize now that the previously mentioned crates, tui-input and tui-textarea, are meant for that purpose xD

In my application I often end up needing some kind of input field.

In my opinion an input field example with a non-adjustable cursor seems not particular useful to me and the lack of Unicode width support appears like trap which bites me back later.

As a new user of ratatui I would appreciate 2 kind of examples in the area of input boxes.

1. A simple one line input field.
2. A multi-line input field.

Both example should include input with unicode width support and the ability to move the cursor around.

if this is out of the scope for this crate, then I find examples for user input more useful if they at least show that you use these external crates for input boxes.

That said, I would also not mind if my solution is replaced by such an example with the usage of an external crate.

joshka:
I think eventually we should combine the two crates and bring them into ratatui as a core widget (or perhaps as a separate widgets crate?)

joshka:
@mindoodoo wrote: 
> I'm unsure though about including examples using external crate widgets. Would you add another example alongside this one or would it take it's place ?

My intent was to communicate that we don't need to try to re-implement those crates in this example. Perhaps it would be a good idea to add a small comment in the current example pointing readers in the correct direction of the crates.

mindoodoo:
Ah I see what you mean now. I think it might be interesting to provide examples for those external crates in the meantime, but bringing one or a combination of the two into ratatui would be interesting I think.

examples/user_input.rs
joshka: It's worth documenting this in the code as it's not obvious why this is just by reading the implementation. Also does the following work?
```rust
self.input = self.input.take(position).skip(1).collect();
```

examples/user_input.rs
```diff
@@ -47,10 +49,58 @@ impl Default for App {
...
+    fn delete_char(&mut self) {
+        let is_not_cursor_leftmost = self.cursor_position != 0;
+        if is_not_cursor_leftmost {
+            let current_index = self.cursor_position;
+            let from_left_to_current_index = current_index - 1;
+
+            let before_char_to_delete = self.input.chars().take(from_left_to_current_index);
+            let after_char_to_delete = self.input.chars().skip(current_index);
+
+            self.input = before_char_to_delete.chain(after_char_to_delete).collect();
```
joshka, line 83: This is not immediately obvious from the implementation (something which an example needs).

Also does the following work here?
```rust
self.input = self.input.take(position).skip(1).collect();
```

joshka (approved):
Approved subject to the small requested change (document the reason for not using input.remove())
Also, please rebase this (there's some small changes to the imports in the main branch.

examples/user_input.rs
BoolPurist:
```rs
self.input.take(position).skip(1).collect();
```
Does not compile, take is a method of the Itrator trait. String does not implement the Iterator trait.

This would compile:

```rs
self.input = self.input.chars().take(current_index).skip(1).collect();
```

If I understand you correctly then you suggest the following implementation for the function delete_char

```rs
fn delete_char(&mut self) {
    let is_not_cursor_leftmost = self.cursor_position != 0;
    if is_not_cursor_leftmost {
        let current_index = self.cursor_position;
        self.input = self.input.chars().take(current_index).skip(1).collect();
        self.move_cursor_left();
    }
    
}
```

I tested this implementation. It does delete roughly all chars after the selected one. I think that is not intended.

joshka: Yep you're right - my bad. Just the doc is fine.

examples/user_input.rs
BoolPurist: Commented inside the function delete_char. The comments also explain why remove was not used.

joshka (approved):
LGTM - thanks for submitting this PR. :)
I'll leave this to @mindoodoo to merge

mindoodoo (approved):
LGTM, would just like to see a mention that this example does not try to handle unicode characters and then will merge !

examples/user_input.rs
```diff
@@ -6,18 +6,18 @@ use std::{error::Error, io};
...
-///   * A input box always focused. Every character you type is registered
-///   here
-///   * Pressing Backspace erases a character
+///   * An input box always focused. Every character you type is registered
+///   here.
+///   * An entered character is inserted at the cursor position.
+///   * Pressing Backspace erases the left character before the cursor position
 ///   * Pressing Enter pushes the current input in the history of previous
-///   messages
+///   messages.
```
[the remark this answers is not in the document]
joshka: The message seems reasonable - I'm going to commit this change rather than wait on another round trip to the author.

joshka:
force push has the comment update, a rebase on main and a cargo fmt of the new comment

mindoodoo (approved):

PR #303, opened 2023-07-09
docs(examples): Add examples readme with gifs

joshka:
This commit adds a readme to the examples directory with gifs of each
example. This should make it easier to see what each example does
without having to run it.

I modified the examples to fit better in the gifs. Mostly this was just
removing the margins, but for the block example I cleaned up the code a
bit to make it more readable.

NOTE!!!!

There are some problems with rendering over ssh, which are not yet resolved. These are noted in the README file. It's possible that these problems are related to https://github.com/tui-rs-revival/ratatui/issues/298.

There is a github action that can automatically run VHS as part of a CI process - which would be possibly useful to keep things up to date. https://github.com/charmbracelet/vhs-action


orhun:
Hey, what else is needed to be done here? Do you want me to update the workflow to add `vhs-action` maybe?

joshka:
Note the TODOs in the doc. I haven't had a chance to dig into why some of the problems are happening (our side vs vhs side)

examples/README.md
```diff
@@ -0,0 +1,189 @@
...
+
+```shell
+cargo run --example=inline --features=crossterm
+```
+
+![Inline][inline.gif]
+
+## Layout ([layout.rs](./layout.rs))
+
+TODO Layout renders poorly
```
orhun, line 85: FWIW, setting the `FontSize` to `15` or updating the constraints to 20-60-20 fixes this.

examples/README.md
```diff
@@ -0,0 +1,189 @@
...
+
+```shell
+cargo run --example=sparkline --features=crossterm
+```
+
+![Sparkline][sparkline.gif]
+
+## Table ([table.rs](./table.rs))
+
+TODO Table Panics!
```
orhun, line 143: I couldn't reproduce this.

examples/README.md
```diff
@@ -0,0 +1,189 @@
...
+
+```shell
+cargo run --example=custom_widget --features=crossterm
+```
+
+![Custom Widget][custom_widget.gif]
+
+## Gauge ([gauge.rs](./gauge.rs))
+
+TODO Gauges render poorly on a tty
```
orhun, line 59: Seems fine to me, only thing is the last gauge is missing borders but can be fixed via:

```diff
diff --git a/examples/gauge.rs b/examples/gauge.rs
index 5423ca7..26773d6 100644
--- a/examples/gauge.rs
+++ b/examples/gauge.rs
@@ -154,7 +154,7 @@ fn ui<B: Backend>(f: &mut Frame<B>, app: &App) {
 
     let label = format!("{}/100", app.progress4);
     let gauge = Gauge::default()
-        .block(Block::default().title("Gauge4"))
+        .block(Block::default().title("Gauge4").borders(Borders::ALL))
         .gauge_style(
             Style::default()
                 .fg(Color::Cyan)

```

joshka: From my terminal:
image

From the image:
image

examples/README.md
joshka: It's likely a size constraint thing - I haven't dug in to this yet.

examples/README.md
joshka: yep - will do

joshka:
In Block.rs, the entire block background should be green and the top two blocks should render similar widths:
image
vs 
image

Popup has an issue rendering the background:
Terminal:
image
VHS:
image


Perhaps it would be sufficient to just note that there are rendering issues with VHS (or rewrite the examples to avoid the issues). @orhun given that you are not seeing the same issues, It's possible that there's some sort of difference in rendering on a mac vs on a linux machine due to tty differences and installed fonts - so I'll check there to compare.

orhun:
Yeah, it is likely that those problems are mostly about `vhs` itself so nothing that would be a good idea.

joshka:
Fixed block, layout and table by changing the examples a bit and updated the README with the new gifs.
Direct link to [README](https://github.com/tui-rs-revival/ratatui/blob/e2acc79036b8fa58458effcb500c9d989402e153/examples/README.md)

image

image

image

examples/README.md
joshka: Updated the example to be even nicer.

examples/README.md
joshka: Repro was due to the width of the space that it was rendering in, and the interaction with Length and Min Constraints. If the constraints can't be satisfied then we panic.

examples/README.md
joshka: Added the border to this.

joshka:
- Updated the output to point at `target/xxx.gif` to avoid having to ignore the files.
- Updated the gauge example to have a border around the last gauge (it doesn't address the issue, just makes it look a little nicer)
- Fixed the table example
- Updated the README
- Added instructions for updating the examples in a comment in the README

orhun:
@joshka Should I add the `vhs-action` in this PR or a separate one? Which one do you prefer?

joshka:
>@joshka Should I add the vhs-action in this PR or a separate one? Which one do you prefer?

Let's hold off on that for now. It only handles the generate part, not the publishing part, so if we wanted to fully automate the updating of the readme that might be a little difficult.

joshka:
- Braille dots are fixed in VHS [0.6.0](https://github.com/charmbracelet/vhs/releases/tag/v0.6.0). 
  - Removed notes about the dots in the README
  - Regenerated all images
- Updated calendar.tape to wait a little longer as the sleep was catching the last of the cargo build command

examples/user_input.tape
```diff
@@ -0,0 +1,21 @@
...
+Type e
+Sleep 1s
+Type "Hello, world!"
+Enter
+Sleep 2s
+Backspace 13
+Sleep 1s
+Type "Goodbye, world!"
+Enter
+Sleep 5s
```
kdheepak, line 21: Should we add a lint check for checking if a new line exists at the end of a file?

examples/README.md
```diff
@@ -0,0 +1,204 @@
...
+cargo run --example=custom_widget --features=crossterm
+```
+
+This is not a particularly exciting example visually, but it demonstrates how to implement your own widget.
+
+![Custom Widget][custom_widget.gif]
+
+## Gauge ([gauge.rs](./gauge.rs))
+
+TODO: work out why the gauge background is rendering poorly via VHS.
```
kdheepak, line 60: What do you think about making these `TODO`s comments?

```suggestion
<!-- TODO: work out why the gauge background is rendering poorly via VHS.-->
```

examples/user_input.tape
kdheepak: I tend to use `precommit` hooks locally and there's a `end-of-file-fixer`. I can look for something that we can run in CI if that's of interest.

examples/README.md
joshka: I wanted to make this visible so that users viewing the README would understand that this is not the how this should render and it's a problem with VHS not Ratatui. I've raised a bug with VHS at https://github.com/charmbracelet/vhs/issues/344 for this

kdheepak: I see. Maybe we should change it to "the background is rendering poorly when we automate generating these examples, it will render correctly when you run this example." or something like that?

joshka: Updated the messages

examples/user_input.tape
joshka: Yeah, go for it :)

kdheepak (approved):

examples/README.md
```diff
@@ -0,0 +1,209 @@
...
+cargo build --examples --features=crossterm,all-widgets
+for i in examples/*.tape
+do
+    echo -n "[${i:s:examples/:::s:.tape:.gif:}]: "
+    vhs $i --publish --quiet
+    # may need to adjust this depending on if you see rate limiting from VHS
+    sleep 1
+done
+```
+-->
```
kdheepak, line 190: I think it would be nice for us to have a developer documentation somewhere, either in this repo or part of the ratatui book.

joshka: Right now that's https://github.com/ratatui-org/ratatui/blob/main/CONTRIBUTING.md, but we could split some of that out to DEVELOPING.md (or whatever the convention is)

PR #309, opened 2023-07-13
feat(barchart) set custom text value in the bar

karthago1:
Hello again :smile_cat: 

somehow I missed this feature, while integrating the new code. (In my original PR #288, I had a function pointer, which formats the values. `format: fn(u64) -> String`).

This patch suggests adding a new function `text_value` to the `Bar`, which will override the default text value

```
Bar::default()
     .label("Temperature".into())
     .value(10)
     .text_value("10°C".to_string());
```

Commit message:

for now the value is converted to a string and then printed. in many cases the values are too wide or double values. so it make sense to set a custom value text instead of the default behavior.

this patch suggests to add a method `fn text_value(mut self, text_value: String)` to the Bar, which allows to override the value printed in the bar

bar

### DIFF
--- a/examples/barchart.rs
+++ b/examples/barchart.rs
@@ -57,17 +57,17 @@ impl<'a> App<'a> {
             companies: [
                 Company {
                     label: "Comp.A",
-                    revenue: [9, 12, 5, 8],
+                    revenue: [9500, 12500, 5300, 8500],
                     bar_style: Style::default().fg(Color::Green),
                 },
                 Company {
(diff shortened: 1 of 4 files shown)


sayanarijit (approved):

joshka:
@karthago1 to make it easier to merge in the future, can you please enable signing commits?
See https://docs.github.com/articles/about-gpg/ for more info

joshka:
LGTM. Thanks again!

PR #310, opened 2023-07-14
Add underline color style for crossterm backend

### LINKED ISSUE #308: Enable setting the underline color
## Problem
There is a nonstandard ansi escape code that allows setting the underline color.

This is done with the code `58`. it uses the same arguments as the `38` and `48` codes do for settings their colors.
The underline color can be reset with the code `59`.

See [kitty documentation](https://sw.kovidgoyal.net/kitty/underlines/) for more info.

This is implemented in at least kitty, wezterm, alacritty, gnome terminal and konsole.

## Solution
Allow this to be possible
```rust
Style::default().underline(Color::Blue).add_modifier(Modifier::UNDERLINE),
```

I can make a PR if needed.

Nogesma:
Fixes #308.

### DIFF
--- a/src/style.rs
+++ b/src/style.rs
@@ -200,7 +200,9 @@ impl fmt::Debug for Modifier {
 /// # use ratatui::layout::Rect;
 /// let styles = [
 ///     Style::default().fg(Color::Blue).add_modifier(Modifier::BOLD | Modifier::ITALIC),
-///     Style::default().bg(Color::Red),
+///     Style::default().bg(Color::Red).add_modifier(Modifier::UNDERLINED),
+///     #[cfg(feature = "crossterm")]
+///     Style::default().underline_color(Color::Green),
 ///     Style::default().fg(Color::Yellow).remove_modifier(Modifier::ITALIC),
 /// ];
 /// let mut buffer = Buffer::empty(Rect::new(0, 0, 1, 1));
@@ -211,7 +213,9 @@ impl fmt::Debug for Modifier {
 ///     Style {
 ///         fg: Some(Color::Yellow),
 ///         bg: Some(Color::Red),
-///         add_modifier: Modifier::BOLD,
+///         #[cfg(feature = "crossterm")]
+///         underline_color: Some(Color::Green),
+///         add_modifier: Modifier::BOLD | Modifier::UNDERLINED,
 ///         sub_modifier: Modifier::empty(),
 ///     },
 ///     buffer.get(0, 0).style(),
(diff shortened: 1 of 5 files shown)


joshka:
That clippy error seems unrelated - probably due to today's rust 1.71.0 update.

Nogesma:
The clippy check seems to fail on a file I haven't touched, and I can't replicate it on my computer.
I think I can actually remove the `set_underline_color` in `buffer.rs` as it's not used anywhere.

joshka:
Fix for lint is in https://github.com/tui-rs-revival/ratatui/pull/311

src/style.rs
```diff
@@ -308,6 +320,22 @@ impl Style {
...
+    ///
+    /// ## Examples
+    ///
+    /// ```rust
+    /// # use ratatui::style::{Color, Modifier, Style};
+    /// let style = Style::default().underline_color(Color::Blue).add_modifier(Modifier::UNDERLINED);
+    /// let diff = Style::default().underline_color(Color::Red).add_modifier(Modifier::UNDERLINED);
+    /// assert_eq!(style.patch(diff), Style::default().underline_color(Color::Red).add_modifier(Modifier::UNDERLINED));
+    /// ```
+    #[cfg(feature = "crossterm")]
```
joshka, line 333: Can you please link to the spec here (and note that the reason the feature is only available in the crossterm feature is because it's not yet implemented on the other backends).

joshka (changes requested):
LGTM generally. It's a small targeted change. I'll wait for @orhun's feedback on this before merging.

src/style.rs
Nogesma: As it's non-standard I don't think there's an official document for it. I can either link to [Wikipedia](https://en.wikipedia.org/wiki/ANSI_escape_code#SGR_(Select_Graphic_Rendition)_parameters) or [Kitty's documentation](https://sw.kovidgoyal.net/kitty/underlines/)

joshka: Either is fine - just something to help those that come behind.

Nogesma:
Rebased on top of main and added a link to some documentation.

joshka:
The current fail is a code coverage reduction can you add a couple of tests to make that happy?

Nogesma:
I'm not sure I understand why it says it's a reduction in coverage. The coverage went up on all the files I modified.
Also the website doesn't seem to have updated with the latest commit. It might be because of the force-pushes.
I can try do do another commit then squash them, maybe that will fix it?

joshka:
I just too a better look. The missing coverage is on the crossterm backend (which is at 0% coverage), apologies for the wrong direction.

joshka (approved):

orhun (approved):
LGTM

PR #317, opened 2023-07-15
fix(prelude): remove widgets module from prelude

joshka:
This helps to keep the prelude small and less likely to conflict with
other crates.

- remove widgets module from prelude as the entire module can be just as
  easily imported with `use ratatui::widgets::*;`
- move prelude module into its own file
- update examples to import widgets module instead of just prelude
- added several modules to prelude to make it possible to qualify
  imports that collide with other types that have similar names

## Note

In the [discussion of this on discord](https://discord.com/channels/1070692720437383208/1072879985762762812/1129381084010909756), just the canvas was suggested to be removed, however given that it's fairly easy to import widgets::*, it makes sense to drop the entire module from the prelude.

### DIFF
--- a/examples/barchart.rs
+++ b/examples/barchart.rs
@@ -9,7 +9,7 @@ use crossterm::{
     execute,
     terminal::{disable_raw_mode, enable_raw_mode, EnterAlternateScreen, LeaveAlternateScreen},
 };
-use ratatui::prelude::*;
+use ratatui::{prelude::*, widgets::*};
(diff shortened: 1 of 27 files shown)


sayanarijit (approved):

joshka:
Will wait on a second review on this one rather than merging.

mindoodoo (approved):

orhun (approved):

PR #325, opened 2023-07-16
feat(barchart): Add direction attribute. (horizontal bars support)

### LINKED ISSUE #321: BarChart: horizontal bars
## Problem
If we want to show the value in the bars, we need to set the bar width as wide as the value text. Hence we can't use the available space in an efficient manner. 

## Solution
1. Print the value vertically. I made a lot of research. I think it is not possible to print the numbers vertically and 90° rotated. (or is it possible?)
2. Print the Bars horizontally instead of vertically. 
I am still playing with this idea. all we need is a new API function `horizontal_bars(true)`
I need to integrate this in our application and get some feedback first. but the horizontal bars looks weird to me, but equally functional. I don't know what is your opinion.

barchart

karthago1:
Fixes #321

this patch introduces new function `direction` to allow the barchart to render the bars horizontally.

Note: when `direction` is set to `Direction::Horizontal` the function `Bar::label()` is useless, because the bar label is nowhere printed. 

any suggestion how to make the horizontal bar look better?

The bar's `value_text` is rendered using 2 styles (`value_style` and the `bar_style`). Reason: see the following image :smile: 

top


kdheepak:
Thanks for the PR! Horizontal bar charts look great. 

It looks like there's a merge conflict in one of the tests. Would you be able to address that?

ZhaoQian888:
nice pr, hope merge

karthago1:
@kdheepak done

examples/barchart.rs
```diff
@@ -207,11 +216,44 @@ where
 
     const LEGEND_HEIGHT: u16 = 6;
     if area.height >= LEGEND_HEIGHT && area.width >= TOTAL_REVENUE.len() as u16 + 2 {
+        let legent_width = TOTAL_REVENUE.len() as u16 + 2;
```
joshka, line 219: nit: typo (legend_width)

src/widgets/barchart/bar.rs
```diff
@@ -56,6 +56,40 @@ impl<'a> Bar<'a> {
         self
     }
 
+    pub(super) fn render_value_with_different_styles(
```
joshka, line 59: This name is not very descriptive. Different from what?

Something that would help someone unfamiliar with this code maintain it easier would be some doc comments that talk about which exact part of the widget / bar that this renders, and the various decisions that are made based on settings.

src/widgets/barchart/bar.rs
```diff
@@ -56,6 +56,40 @@ impl<'a> Bar<'a> {
...
+        bar_style: Style,
+    ) {
+        let text = if let Some(text) = self.text_value {
+            text
+        } else {
+            self.value.to_string()
+        };
+
+        if !text.is_empty() {
+            let style = self.value_style.patch(default_value_style);
```
joshka, line 74: I don't understand this line - the parameter that's passed to the method is `self.value_style`, so doesn't this just mean `self.value_style.patch(self.value_style)`?

A unit test for this method would help clarify the expected rendered output.

src/widgets/barchart/bar_group.rs
```diff
@@ -35,6 +39,18 @@ impl<'a> BarGroup<'a> {
     pub(super) fn max(&self) -> Option<u64> {
         self.bars.iter().max_by_key(|v| v.value).map(|v| v.value)
     }
+
+    pub(super) fn render_label(self, buf: &mut Buffer, area: Rect, default_label_style: Style) {
+        if let Some(mut label) = self.label {
+            label.patch_style(default_label_style);
+            buf.set_line(
+                area.x + (area.width.saturating_sub(label.width() as u16) >> 1),
```
joshka, line 47: This is an odd calculation. Can you explain it?

src/widgets/barchart/mod.rs
```diff
@@ -53,6 +53,8 @@ pub struct BarChart<'a> {
     /// Value necessary for a bar to reach the maximum height (if no value is specified,
     /// the maximum value in the data is taken as reference)
     max: Option<u64>,
+    /// value to render the bars in the horizontal direction
+    horizontal_bars: bool,
```
joshka, line 57: [major]: Let's prefer to use a direction instead of a bool for this?

src/widgets/barchart/mod.rs
```diff
@@ -196,6 +205,71 @@ impl<'a> BarChart<'a> {
...
+            if self.group_gap > 0 && bar_y >= bars_area.top() + label_offset {
+                let label_rect = Rect {
+                    y: bar_y - label_offset,
+                    ..bars_area
+                };
+                group.render_label(buf, label_rect, self.label_style);
+            }
+
+            for (bar_size, bar) in group_data.into_iter().zip(bars) {
+                let bar_style = bar.style.patch(self.bar_style);
```
joshka, line 239: Does bar_style override the individual bar.style? Why?

src/widgets/barchart/mod.rs
```diff
@@ -196,6 +205,71 @@ impl<'a> BarChart<'a> {
...
+                        buf.get_mut(bars_area.left() + x, bar_y - y)
+                            .set_symbol(self.bar_set.full)
+                            .set_style(bar_style);
+                    }
+                    for x in bar_size..bars_area.width {
+                        buf.get_mut(bars_area.left() + x, bar_y - y)
+                            .set_symbol(self.bar_set.empty)
+                            .set_style(bar_style);
+                    }
+                }
```
joshka, line 252: I suspect that this sort of code calls for a nicer method on buffer to set an entire Rect perhaps?
This makes it a bit more obvious how this works:

```suggestion
                for y in 0..self.bar_width {
                    for x in 0..bars_area.width {
                        let symbol = if x < bar_size { self.bar_set.full } else { self.bar_set.empty };
                        buf.get_mut(bars_area.left() + x, bar_y - y)
                            .set_symbol(symbol)
                            .set_style(bar_style);
                    }
                }
```

src/widgets/barchart/mod.rs
```diff
@@ -756,4 +835,44 @@ mod tests {
         let expected = Buffer::with_lines(vec!["  █", "█ █", " G "]);
         assert_buffer_eq!(buffer, expected);
     }
+
+    #[test]
+    fn test_horizontal_bars() {
```
joshka, line 840: Imagine the next person to implements a new feature in barchart makes this test fail. One of the problems with writing tests that test all the features at once is that it requires any failure to now fully understand the entire code being tested rather than just the part that failed.

One way to help that person (which could even be your future self) would be to forget about the implementation for a minute and write a list of the things that this test tells you about how the implementation works (and why).

For instance what is the significance to the number 5 in the first bar?

Is there a reason that the bars are rendered in reverse order from their definition?

Is there a reason why the group name is the part that is not rendered rather than one of the bars?

How is each edge case handled? What are they?

joshka (changes requested):
This definitely is something that will be useful. This needs some docs about how it works and some more detailed testing. There's a couple of places where I'm not sure what the behavior should be, so I'm not sure if it's a bug or an non-intuitive feature that you've considered.

src/widgets/barchart/bar.rs
karthago1: one `self` has the type `Bar` and he other one has the type `BarChart`. So the styles are different.

but since you mentioned it somewhere else. this line should be:

`let style = default_value_style.patch(self.value_style);`

I will create unit tests for all styles, since patching the style is everywhere wrong / (swapped). :smile:

src/widgets/barchart/bar_group.rs
```diff
@@ -35,6 +39,18 @@ impl<'a> BarGroup<'a> {
     pub(super) fn max(&self) -> Option<u64> {
         self.bars.iter().max_by_key(|v| v.value).map(|v| v.value)
     }
+
+    pub(super) fn render_label(self, buf: &mut Buffer, area: Rect, default_label_style: Style) {
+        if let Some(mut label) = self.label {
+            label.patch_style(default_label_style);
```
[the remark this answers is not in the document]
karthago1: true, now I checked the patching code. It is everywhere wrong. in some cases all I need to do, is to swap the styles. but in this case I feel that `Label` is missing a function similar to `patch_style`, but it take the passed style as base or default style.

what do you think? should I add a new function to `Label`? Or do you have a better idea?

src/widgets/barchart/mod.rs
karthago1: true. I was lazy :smiley_cat: . I will rework this

src/widgets/barchart/bar_group.rs
karthago1: we want to center the label.

`area.x`: where the `BarChart` starts to draw
`(area.width - label.width) / 2` : is the offset where we can draw a centered label

src/widgets/barchart/bar.rs
karthago1: comment added. hope it is fine now

src/widgets/barchart/bar.rs
```diff
@@ -56,6 +56,40 @@ impl<'a> Bar<'a> {
...
+            // Since the value may be longer than the bar itself, we need to use 2 different styles
+            // while rendering. Render the first part with the default value style
+            buf.set_stringn(area.x, area.y, &text, text.len().min(bar_size), style);
+
+            // render the second part with the bar_style
```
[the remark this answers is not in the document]
karthago1: done. hope the documentation is clear

src/widgets/barchart/bar_group.rs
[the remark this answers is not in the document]
karthago1: just patched the label directly

```
// patch label styles
for span in &mut label.spans {
    span.style = default_label_style.patch(span.style);
}
```

src/widgets/barchart/mod.rs
karthago1: > For instance what is the significance to the number 5 in the first bar?

has no significance. all I need a bi bar and a small one, which doesn't fit the text

> Is there a reason that the bars are rendered in reverse order from their definition?

I guess, you mean that the first bar is rendered at the bottom. I think looks nicer. usually graphs, bars, charts start always from the bottom

> Is there a reason why the group name is the part that is not rendered rather than one of the bars?

because there is no test case for group name outside the scope.  On the other hand bars outside the scope is already tested in the old code and still valid  for this MR, because the function responsible for removing bars outside the scope is `remove_invisible_groups_and_bars`. It is a common function for the horizontal and vertical rendering.

> How is each edge case handled? What are they?

I added tests for the styles. I hope it enough :smile_cat:

src/widgets/barchart/bar_group.rs
```diff
@@ -35,6 +39,23 @@ impl<'a> BarGroup<'a> {
...
+            // patch label styles
+            for span in &mut label.spans {
+                span.style = default_label_style.patch(span.style);
+            }
+
+            let x_offset = match label.alignment {
+                Some(Alignment::Center) => area.width.saturating_sub(label.width() as u16) >> 1,
+                Some(Alignment::Right) => area.width.saturating_sub(label.width() as u16),
+                _ => 0,
+            };
```
karthago1, line 54: just to communicate clearly. This is a breaking change. It will allow the user to decide, where to render labels of the group. As a result the labels will be rendered Left by default.

I can move this to another PR if needed

joshka (changes requested):
A couple of small typos, but generally looks good to me. If you can get these fixed up in the next few days this should make the 0.23 release.

I'd still query whether bottom to top rendering is the correct approach, but happy to take that as a future PR if needed.

karthago1:
> I'd still query whether bottom to top rendering is the correct approach, but happy to take that as a future PR if needed.

@joshka I don't care much. I slightly like printing from the bottom. here I made a comparison. It will be nice if you choose, which one looks better? (it will not require much time changing it)

top: 
top

(here we should maybe move the legend to the bottom )

botttom:
bottom

karthago1:
or maybe let's do in future PR, as you mentioned. maybe we make it configurable. Maybe variable name `layout_direction` or something similar

joshka:
> or maybe let's do in future PR, as you mentioned. maybe we make it configurable. Variable name would be `layout_direction` or something similar

Yep that's a good way to handle it

joshka:
Let's go top down for the first release and add another property in sometime soon in a .1 release?

karthago1:
> Let's go top down for the first release and add another property in sometime soon in a .1 release?

ok fine.

joshka (approved):

joshka:
LGTM. Thanks again for the PR and for your patience sticking with it to get this reviewed and merged.

src/widgets/barchart/bar_group.rs
joshka: (this is fine as part of this PR). Resolving.

joshka:
@ratatui-org/core can someone please take a second review on this?

kdheepak (approved):

joshka:
Merging! Thanks @karthago1

PR #329, opened 2023-07-17
docs: improve scrollbar doc comment

a-kenji:
I find the title is not really needed for the doc comment.

Alternative:

Rename `tui-rs` to `ratatui` in the title, but I find that this is not needed.

### DIFF
--- a/src/widgets/scrollbar.rs
+++ b/src/widgets/scrollbar.rs
@@ -155,9 +155,8 @@ pub enum ScrollbarOrientation {
     HorizontalTop,
 }
 
-/// Scrollbar widget for tui-rs library.
+/// A widget to display a scrollbar
 ///
-/// This widget can be used to display a scrollbar in a terminal user interface.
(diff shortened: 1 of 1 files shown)


joshka:
Actually that highlights something for me. That title should show up in the one line description of the struct, but it doesn't because it shows up in the re-exports because the scrollbar module is public. For consistency it shouldn't actually be public, but that required moving the sets. PR for that incoming. There should still be a one line description of the struct though, so just drop the tui-rs part.

a-kenji:
Good catch!

I changed the title to be in line with existing widget titles.
I found this made the first line in the body obsolete.

joshka (approved):

joshka:
Are you able to setup commit signing? https://docs.github.com/articles/about-gpg/

orhun (approved):

joshka:
Thanks :) Merged!

PR #330, opened 2023-07-17
fix(scrollbar)!: move symbols and set to symbols module

joshka:
This makes it consistent with the other symbol sets

BREAKING CHANGE: The symbols are now in the `symbols` module. To update
your code, add an import for `ratatui::symbols::scrollbar::*` (or the
specific symbols you need).

### DIFF
--- a/examples/scrollbar.rs
+++ b/examples/scrollbar.rs
@@ -9,7 +9,7 @@ use crossterm::{
     execute,
     terminal::{disable_raw_mode, enable_raw_mode, EnterAlternateScreen, LeaveAlternateScreen},
 };
-use ratatui::{prelude::*, widgets::*};
+use ratatui::{prelude::*, symbols::scrollbar, widgets::*};
 
 #[derive(Default)]
 struct App {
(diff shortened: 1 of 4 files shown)


examples/scrollbar.rs
```diff
@@ -9,7 +9,7 @@ use crossterm::{
     execute,
     terminal::{disable_raw_mode, enable_raw_mode, EnterAlternateScreen, LeaveAlternateScreen},
 };
-use ratatui::{prelude::*, widgets::*};
+use ratatui::{prelude::*, symbols::scrollbar, widgets::*};
```
kdheepak, line 12: Why doesn't `widgets::scrollbar` conflict with `symbols::scrollbar`?

joshka: This is because widgets::scrollbar is no longer re-exported, so there is no widgets::scrollbar module.

Before:

image

After:

image

orhun:
nit: commit message should be `fix(scrollbar)!:` instead of `fix!(scrollbar)`.

joshka:
>nit: commit message should be fix(scrollbar)!: instead of fix!(scrollbar).

Fixed :)

kdheepak (approved):

orhun (approved):

PR #332, opened 2023-07-17
build: forbid unsafe code

EdJoPaTo:
This indicates good (high level) code and is used by tools like cargo-geiger.

Reading #66 it also sounds like a good idea to have this unless someone has a really good explanation to remove this.

Also see https://github.com/tui-rs-revival/ratatui/pull/157#discussion_r1200281899


sayanarijit (approved):

joshka:
Thanks @EdJoPaTo for the code contribution!

PR #337, opened 2023-07-20
fix(example): Fix typo

joshrotenberg:
the existential feels


joshka (approved):
Perhaps we shouldn't change this in the interest of those who relied on being able to "Press q to exist". :D

joshrotenberg:
> Perhaps we shouldn't change this in the interest of those who relied on being able to "Press q to exist". :D

I say they are on their own now.

Issue #307, opened 2023-07-13
Chore: implement common traits


joshka:
## Description

All types should when possible implement the common traits.
https://rust-lang.github.io/api-guidelines/interoperability.html#c-common-traits

This allows consumers of the library to derive these traits themselves when needed rather than having to make manual implementations or avoid them.

Many of the types in ratatui do not implement these traits that could do so (particularly Hash is absent most of the time).

## To Reproduce

Example: Try to create a type that derives `Hash` and has a `Style` field

## Expected behavior

All common traits that make sense from https://rust-lang.github.io/api-guidelines/interoperability.html#c-common-traits should be implemented
- [Copy](https://doc.rust-lang.org/std/marker/trait.Copy.html) - Not possible for some items
- [Clone](https://doc.rust-lang.org/std/clone/trait.Clone.html)
- [Eq](https://doc.rust-lang.org/std/cmp/trait.Eq.html)
- [PartialEq](https://doc.rust-lang.org/std/cmp/trait.PartialEq.html)
- [Ord](https://doc.rust-lang.org/std/cmp/trait.Ord.html) - Usually not necessary
- [PartialOrd](https://doc.rust-lang.org/std/cmp/trait.PartialOrd.html) - Usually not necessary
- [Hash](https://doc.rust-lang.org/std/hash/trait.Hash.html)
- [Debug](https://doc.rust-lang.org/std/fmt/trait.Debug.html)
- [Display](https://doc.rust-lang.org/std/fmt/trait.Display.html) - Not sure what this should look like TBH
- [Default](https://doc.rust-lang.org/std/default/trait.Default.html) - Sometimes not

TieWay59:
Hi @joshka , I have one question: 

Is is OK to split this task up? I did small search and I have found too many structs that scatter around the code base, if someone raise a PR with trivial changes in different files and logic, it would be hard for others to review I guess. WDUT?

| Struct Name         | Debug | Default | PartialEq | Eq | Clone | Copy | Hash | FromArgs |
|---------------------|-------|---------|-----------|----|-------|------|------|----------|
| SinSignal           |   -   |   -     |     -     |  - |   ✔️   |  -   |  -   |    -     |
| Label               |   -   |   ✔️     |     -     |  - |   -   |  -   |  -   |    -     |
| App                 |   -   |   ✔️     |     -     |  - |   -   |  -   |  -   |    -     |
[...]

PR #339, opened 2023-07-21
chore(widgets): Implement `Debug & Default` common traits

TieWay59:
Implement `Debug & Default` common traits for most structs in src.

Reorder the derive fields to be more consistent:

    Debug, Default, Clone, Copy, Eq, PartialEq, Ord, PartialOrd, Hash

see: https://github.com/ratatui-org/ratatui/issues/307

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -14,7 +14,7 @@ use crate::{
 };
 
 /// A buffer cell
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, Eq, PartialEq)]
 pub struct Cell {
     pub symbol: String,
     pub fg: Color,
@@ -135,7 +135,7 @@ impl Default for Cell {
 /// buf.get_mut(5, 0).set_char('x');
 /// assert_eq!(buf.get(5, 0).symbol, "x");
 /// ```
-#[derive(Clone, PartialEq, Eq, Default)]
+#[derive(Default, Clone, Eq, PartialEq)]
 pub struct Buffer {
     /// The area represented by this buffer
     pub area: Rect,
--- a/src/layout.rs
+++ b/src/layout.rs
@@ -11,27 +11,35 @@ use cassowary::{
     WeightedRelation::{EQ, GE, LE},
 };
 
-#[derive(Debug, Hash, Clone, Copy, PartialEq, Eq)]
+#[derive(Debug, Default, Hash, Clone, Copy, PartialEq, Eq)]
 pub enum Corner {
+    #[default]
     TopLeft,
     TopRight,
     BottomRight,
     BottomLeft,
 }
 
-#[derive(Debug, Hash, Clone, PartialEq, Eq)]
+#[derive(Debug, Default, Hash, Clone, Eq, PartialEq)]
 pub enum Direction {
     Horizontal,
+    #[default]
     Vertical,
 }
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+#[derive(Debug, Clone, Copy, Eq, PartialEq, Hash)]
 pub enum Constraint {
     Percentage(u16),
     Ratio(u32, u32),
     Length(u16),
     Max(u16),
     Min(u16),
+}
+
+impl Default for Constraint {
+    fn default() -> Self {
+        Constraint::Percentage(100)
+    }
 }
 
 impl Constraint {
@@ -56,14 +64,15 @@ impl Constraint {
     }
 }
 
-#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[derive(Debug, Default, Clone, PartialEq, Eq, Hash)]
 pub struct Margin {
     pub vertical: u16,
     pub horizontal: u16,
 }
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq, Hash)]
 pub enum Alignment {
+    #[default]
     Left,
     Center,
     Right,
@@ -359,6 +368,7 @@ fn split(area: Rect, layout: &Layout) -> Rc<[Rect]> {
 }
 
 /// A container used by the solver inside split
+#[derive(Debug)]
 struct Element {
     x: Variable,
     y: Variable,
@@ -395,7 +405,7 @@ impl Element {
 
 /// A simple rectangle used in the computation of the layout and to give widgets a hint about the
 /// area they are supposed to render to.
-#[derive(Debug, Clone, Copy, Hash, PartialEq, Eq, Default)]
+#[derive(Debug, Default, Clone, Copy, Eq, PartialEq, Hash)]
 pub struct Rect {
     pub x: u16,
     pub y: u16,
--- a/src/symbols.rs
+++ b/src/symbols.rs
@@ -21,6 +21,12 @@ pub mod block {
         pub empty: &'static str,
     }
 
+    impl Default for Set {
+        fn default() -> Self {
+            NINE_LEVELS
+        }
+    }
+
     pub const THREE_LEVELS: Set = Set {
         full: FULL,
         seven_eighths: FULL,
@@ -67,6 +73,12 @@ pub mod bar {
         pub one_quarter: &'static str,
         pub one_eighth: &'static str,
         pub empty: &'static str,
+    }
+
+    impl Default for Set {
+        fn default() -> Self {
+            NINE_LEVELS
+        }
     }
 
     pub const THREE_LEVELS: Set = Set {
@@ -158,6 +170,12 @@ pub mod line {
         pub cross: &'static str,
     }
 
+    impl Default for Set {
+        fn default() -> Self {
+            NORMAL
+        }
+    }
+
     pub const NORMAL: Set = Set {
         vertical: VERTICAL,
         horizontal: HORIZONTAL,
@@ -222,9 +240,10 @@ pub mod braille {
 }
 
 /// Marker to use when plotting data points
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Default, Clone, Copy)]
 pub enum Marker {
     /// One point per cell in shape of dot
+    #[default]
     Dot,
     /// One point per cell in shape of a block
     Block,
@@ -246,7 +265,7 @@ pub mod scrollbar {
     /// │  └──────── thumb
     /// └─────────── begin
     /// ```
-    #[derive(Debug, Clone)]
+    #[derive(Debug, Default, Clone)]
     pub struct Set {
         pub track: &'static str,
         pub thumb: &'static str,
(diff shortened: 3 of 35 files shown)


TieWay59:
Hey @joshka, I've been doing this today. And I wish to share my discoveries with you before more next step of review. 

1. I can't split these files into four PR, because many of them are related (for example, if A deps on B, then we need B to have some C-Trait, then, add it to A.) So this is a huge one hard to review, if you have any idea on splitting this up, let me know if possible.
2. I following an intention to give as many common traits as we can, in  order of:
    - [Copy](https://doc.rust-lang.org/std/marker/trait.Copy.html)
    - [Clone](https://doc.rust-lang.org/std/clone/trait.lone.html)
    - [Eq](https://doc.rust-lang.org/std/cmp/trait.Eq.html)
    - [PartialEq](https://doc.rust-lang.org/std/cmp/trait.PartialEq.html)
    - [Ord](https://doc.rust-lang.org/std/cmp/trait.Ord.html)
    - [PartialOrd](https://doc.rust-lang.org/std/cmp/trait.PartialOrd.html)
    - [Hash](https://doc.rust-lang.org/std/hash/trait.Hash.html) 
    - [Debug](https://doc.rust-lang.org/std/fmt/trait.Debug.html)
    - [Display](https://doc.rust-lang.org/std/fmt/trait.Display.html)
    - [Default](https://doc.rust-lang.org/std/default/trait.Default.html) 
3. But.. `Copy` trait might be implicitly memory-costing if have it around too much. 
4. `Display` is more difficult to design for every struct, so I never actively add it.
5. Some `Default impl` can be replaced by derive if every member has a `Default` trait already, I tried some cases, and I wonder if it's necessary to write a test case for it. 
6. `Ord`, and `PartialOrd`, might seem redundant but I add them all the way if it's possible.

src/buffer.rs
```diff
@@ -14,7 +14,7 @@ use crate::{
 };
 
 /// A buffer cell
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug)]
```
joshka, line 17: Nit: can we make the order of the derives the same. This is mostly because I'm a little OCD about things that are the same should look the same, but it does have the benefit of being really easy to see at a glance which things are supported if the traits are always in the same order. Perhaps use the following order: Debug, Default, Clone, Copy, Eq, PartialEq, Ord, PartialOrd, Hash.

Taking a look at Ord/PartialOrd and the things which implement these, perhaps it doesn't make sense for most of the types to implement this. Perhaps let's defer that?

src/layout.rs
```diff
@@ -34,6 +35,12 @@ pub enum Constraint {
     Min(u16),
 }
 
+impl Default for Constraint {
+    fn default() -> Constraint {
+        Constraint::Percentage(100)
+    }
+}
```
joshka, line 42: I wonder if Constraint makes sense to have a Default. Where would you see this being used?

src/layout.rs
```diff
@@ -19,13 +19,14 @@ pub enum Corner {
     BottomLeft,
 }
 
-#[derive(Debug, Hash, Clone, PartialEq, Eq)]
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Hash, Debug, Default)]
 pub enum Direction {
+    #[default]
```
joshka, line 24: We actually use Vertical as the default layout direction in Layout::new()

src/symbols.rs
```diff
@@ -21,6 +21,12 @@ pub mod block {
         pub empty: &'static str,
     }
 
+    impl Default for Set {
+        fn default() -> Self {
+            THREE_LEVELS
```
joshka, line 26: NINE_LEVELS is probably more generally useful for these as it shows more information to the user whereas three levels throws some away.

src/symbols.rs
```diff
@@ -69,6 +75,12 @@ pub mod bar {
...
```
joshka, line 80: Same - nine levels?

src/widgets/gauge.rs
```diff
@@ -375,4 +363,22 @@ mod tests {
                 .remove_modifier(Modifier::DIM)
         )
     }
+
+    #[test]
+    fn line_gauge_default_is_identical_to_old_default() {
```
joshka, line 368: The naming of this implies that there is something interesting about the old defaults, but it's not obvious what that might be.

Try to phrase tests as defining current behavior unless there's something specifically novel about what the old behavior was (and if that's so, add docs on why that is interesting). Here I think the defaults are just ordinary defaults and `fn default()` would be fine with `assert_eq!(LineGauge::default(), LineGauged { ... }`

joshka (changes requested):
LGTM generally - I'll answer the question comments in the conversation

My main suggestion is to remove the Eq/PartialEq where it doesn't seem to make sense (unless you think we need it for some reason).

joshka:
> Hey @joshka, I've been doing this today. And I wish to share my discoveries with you before more next step of review.

Nice

>     1. I can't split these files into four PR, because many of them are related (for example, if A deps on B, then we need B to have some C-Trait, then, add it to A.) So this is a huge one hard to review, if you have any idea on splitting this up, let me know if possible.

Makes sense, I'd even prefer to see a single commit if the split commits don't compile (as it makes it easier to `git bisect` if every commit compiles). No big deal on this as the merge will probably join the commits, but doesn't hurt.

>     2. I following an intention to give as many common traits as we can, in  order of:
>        
>        * [Copy](https://doc.rust-lang.org/std/marker/trait.Copy.html)
>        * [Clone](https://doc.rust-lang.org/std/clone/trait.lone.html)
>        * [Eq](https://doc.rust-lang.org/std/cmp/trait.Eq.html)
>        * [PartialEq](https://doc.rust-lang.org/std/cmp/trait.PartialEq.html)
>        * [Ord](https://doc.rust-lang.org/std/cmp/trait.Ord.html)
>        * [PartialOrd](https://doc.rust-lang.org/std/cmp/trait.PartialOrd.html)
>        * [Hash](https://doc.rust-lang.org/std/hash/trait.Hash.html)
>        * [Debug](https://doc.rust-lang.org/std/fmt/trait.Debug.html)
>        * [Display](https://doc.rust-lang.org/std/fmt/trait.Display.html)
>        * [Default](https://doc.rust-lang.org/std/default/trait.Default.html)

Note my comment about ordering. Of these, I'd probably say that Debug and Default are the two most important, followed by Clone and Copy. Being able to glance at a derives list and see that these are implemented is something I personally often rely on. Small nit but useful I think.

>     3. But.. `Copy` trait might be implicitly memory-costing if have it around too much.

Can you point out the impact of any problems you might see with this?

>     4. `Display` is more difficult to design for every struct, so I never actively add it.

Let's defer Display to another PR?

>     5. Some `Default impl` can be replaced by derive if every member has a `Default` trait already, I tried some cases, and I wonder if it's necessary to write a test case for it.

It's probably worthwhile adding default tests as it helps catch any future changes in the defaults that are accidentally made.

>     6. `Ord`, and `PartialOrd`, might seem redundant but I add them all the way if it's possible.

Perhaps these should not be implemented except for where it's necessary? Is there better guidance on this available? Are there use cases where having these makes things easier to implement that aren't obvious?

src/layout.rs
TieWay59: I just saw this is more widely used, but it's OK to remove it if not so helpful.

src/symbols.rs
TieWay59: OK, learned it today~

src/widgets/gauge.rs
TieWay59: I got your point, but still have no idea what test name is better, what do you like?

src/buffer.rs
TieWay59: I agree~

TieWay59:
@joshka Now my PR is focused on Debug & Default trait, I hope this shift can make you feel better while reviewing. And I also reordered some old patterns to align with the preferred traits order.

src/widgets/gauge.rs
joshka: I'd go with just `fn default()` as the test name.

src/widgets/gauge.rs
```diff
@@ -375,4 +363,25 @@ mod tests {
                 .remove_modifier(Modifier::DIM)
         )
     }
+
+    #[test]
+    fn line_gauge_default_values() {
+        // TODO: replace to `assert_eq!(LineGauge::default(), LineGauge::default())`
+        // when `Eq` or `PartialEq` is implemented for `LineGauge`.
```
joshka, line 370: Is there a blocker to implementing these (or is it just that doing so means a bunch of changes to all the types?)

joshka (approved):
LGTM - just one question that is not blocking.

src/widgets/gauge.rs
TieWay59:
```rust
#[derive(Debug, Default, PartialEq)] // <-- if add `PartialEq`
pub struct LineGauge<'a> {
    block: Option<Block<'a>>,
    ratio: f64,
    label: Option<Line<'a>>,
    line_set: symbols::line::Set, // <-- we'll need this to have PartialEq trait
    style: Style,
    gauge_style: Style,
}
```

So, I suppose deferring it would be nice.

TieWay59:
@joshka About the coverage, Should I write a test for every struct? I found it would be very huge, and since there are still many old codes that don't have coverage before me, I'd like to hear your idea. 

It's OK if you feel this is necessary. But I have to warn that, since we don't have much `PartialEq` trait, I may have to write cases like:

```rust
        assert_eq!(
            format!("{:?}", LineGauge::default()),
            format!(
                "{:?}",
                LineGauge {
                    block: None,
                    ratio: 0.0,
                    label: None,
                    style: Style::default(),
                    line_set: symbols::line::NORMAL,
                    gauge_style: Style::default(),
                }
            ),
            "LineGauge::default() should have correct default values."
        );
```

(Yes, it looks weird. it's my workaround, please guide me if you have a better suggestion :heart: )

TieWay59:
Sorry I pushed again, I found one struct I missed, and I give a shorter name for the default test.

joshka (approved):

joshka:
I'd say that adding coverage on the derived traits can probably be deferred. I'll leave that up for whoever does the second review on this.

kdheepak:
I don't have a blocking review, but I was curious what the impact on library size (or binary size) this has. Is there some metrics you can add to the PR description?

kdheepak (approved):

joshka:
> I don't have a blocking review, but I was curious what the impact on library size (or binary size) this has. Is there some metrics you can add to the PR description?

This is pretty much zero.
I check on the debug build one of the examples shows up in the debug builds as 4KB extra on a 3.8MB binary. The release mode build is the exact same size. I'd assume most use cases will never care about binary size for this sort of thing, but if there's something obvious I'm missing...

joshka:
The scrollbar::Set type moved to symbols::scrollbar::Set. I updated the commit with that change (just needs an approval to merge).

Thanks for submitting this PR. It's something that I have run into a few times (particularly Debug being missing).

joshka:
P.s. to make future PRs easier for us merge, can you please setup commit signing?
See https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

PR #341, opened 2023-07-21
doc(book): mdbook demo

joshrotenberg:
Quick demo to see what consolidating docs into a book looks like. You can see a live demo on my fork: <http://joshrotenberg.com/ratatui/index.html> 

This consolidates the docs in the README and the wiki. If the project is interested I can keep going and clean it up. If not, no big deal, only took a few minutes.


joshka:
Hey, take a look over at https://github.com/ratatui-org/ratatui-book

I think that perhaps there's a more manual approach that would have some value on this.

There's some discussion about the docs in the [Docs channel on discord](https://discord.com/channels/1070692720437383208/1080107705672413264) (and elsewhere).

I have two concerns about putting the mdbook into the main like this:
- the process for writing a book / docs is a bit more relaxed than a library, we decided to split that out for that reason mainly
- keeping the docs up to date with any changes to the README etc. seems like it would be annoying and having two places with similar but not quite the same info can be confusine.

Take a look at https://diataxis.fr/ for some general ideas on what we'd like to see when it comes to docs.

joshrotenberg:
> Hey, take a look over at https://github.com/ratatui-org/ratatui-book

Opps, didn't even see this! (FYI, the link to the published site ([tui-rs-revival.github.io/ratatui-book/](https://tui-rs-revival.github.io/ratatui-book/) needs to be updated for the org name change).

> I think that perhaps there's a more manual approach that would have some value on this.
> 
> There's some discussion about the docs in the [Docs channel on discord](https://discord.com/channels/1070692720437383208/1080107705672413264) (and elsewhere).

Will look at this and maybe join for some discussion.

> I have two concerns about putting the mdbook into the main like this:
> 
> * the process for writing a book / docs is a bit more relaxed than a library, we decided to split that out for that reason mainly

I hear that. I'm back and forth, but, at least from my experience, having the manual live with the code vs a separate repo seems to at least have the psychological effect of it being a little closer to where more changes are taking place and make it more likely to stay up to date. Only a personal experience, though.

> * keeping the docs up to date with any changes to the README etc. seems like it would be annoying and having two places with similar but not quite the same info can be confusine.

You're absolutely right, and my suggestion would be to pare down the README to the basics, and link to the docs where appropriate. 

> Take a look at https://diataxis.fr/ for some general ideas on what we'd like to see when it comes to docs.

This is great, never seen it before. I'll take a closer look this weekend.

joshka:
>I hear that. I'm back and forth, but, at least from my experience, having the manual live with the code vs a separate repo seems to at least have the psychological effect of it being a little closer to where more changes are taking place and make it more likely to stay up to date. Only a personal experience, though.

Yup I have a similar experience, but because the people that are spearheading the docs are a little different from the maintainers of the crate, Conway's law strikes and we have two repos ;)

joshrotenberg:
Gotcha. I can close this, and I'll look for ways to help on the other repo as well.

PR #344, opened 2023-07-24
Update README.md: fix broken link

t-nil:


joshka (approved):

joshka:
Thanks for the fix. I pushed a quick fix to the commit message. (We use conventional commits and check for that prior to merge).

kdheepak (approved):

PR #345, opened 2023-07-24
docs(examples): Add color and modifiers examples

joshka:
The intent of these examples is to show the available colors and
modifiers.

- added impl Display for Color

colors
modifiers

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,7 +51,9 @@ unicode-width = "0.1"
 [dev-dependencies]
 anyhow = "1.0.71"
 argh = "0.1"
-cargo-husky = { version = "1.5.0", default-features = false, features = ["user-hooks"] }
+cargo-husky = { version = "1.5.0", default-features = false, features = [
+  "user-hooks",
+] }
 criterion = { version = "0.5", features = ["html_reports"] }
 fakeit = "1.1"
 itertools = "0.10"
@@ -92,6 +94,12 @@ required-features = ["crossterm"]
 doc-scrape-examples = true
 
 [[example]]
+name = "colors"
+required-features = ["crossterm"]
+# this example is a bit verbose, so we don't want to include it in the docs
+doc-scrape-examples = false
+
+[[example]]
 name = "custom_widget"
 required-features = ["crossterm"]
 doc-scrape-examples = true
@@ -120,6 +128,12 @@ doc-scrape-examples = true
 name = "list"
 required-features = ["crossterm"]
 doc-scrape-examples = true
+
+[[example]]
+name = "modifiers"
+required-features = ["crossterm"]
+# this example is a bit verbose, so we don't want to include it in the docs
+doc-scrape-examples = false
 
 [[example]]
 name = "panic"
--- a/examples/README.md
+++ b/examples/README.md
@@ -46,6 +46,14 @@ cargo run --example=chart --features=crossterm
 
 ![Chart][chart.gif]
 
+## Colors ([colors.rs](./colors.rs))
+
+```shell
+cargo run --example=colors --features=crossterm
+```
+
+![Colors][colors.gif]
+
 ## Custom Widget ([custom_widget.rs](./custom_widget.rs))
 
 ```shell
@@ -102,6 +110,14 @@ cargo run --example=list --features=crossterm
 ```
 
 ![List][list.gif]
+
+## Modifiers ([modifiers.rs](./modifiers.rs))
+
+```shell
+cargo run --example=modifiers --features=crossterm
+```
+
+![Modifiers][modifiers.gif]
 
 ## Panic ([panic.rs](./panic.rs))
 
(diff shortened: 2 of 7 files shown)


examples/style.rs
```diff
@@ -0,0 +1,238 @@
...
+    let backend = CrosstermBackend::new(stdout);
+    let mut terminal = Terminal::new(backend)?;
+    terminal.hide_cursor()?;
+    Ok(terminal)
+}
+
+fn restore_terminal(mut terminal: Terminal<CrosstermBackend<Stdout>>) -> Result<()> {
+    disable_raw_mode()?;
+    execute!(terminal.backend_mut(), LeaveAlternateScreen)?;
+    terminal.clear()?;
```
kdheepak, line 235: nitpick: I like not having `terminal.clear()` in restore, I feel like that is a better UX.

```suggestion
```

examples/style.rs
```diff
@@ -0,0 +1,238 @@
...
+        let paragraph =
+            Paragraph::new(color.to_string()).style(Style::default().bg(*color).fg(Color::Black));
+        frame.render_widget(paragraph, Rect::new(i as u16 * 8, 2, 8, 1));
+    }
+    for (i, color) in bright_colors.iter().enumerate() {
+        let paragraph =
+            Paragraph::new(color.to_string()).style(Style::default().bg(*color).fg(Color::Black));
+        frame.render_widget(paragraph, Rect::new(i as u16 * 8, 3, 8, 1));
+    }
+}
```
kdheepak, line 94: What do you think about adding some kind of title for each row? It's easier to grok what is going on when looking at the screenshot that way. e.g.:

image


```suggestion
fn render_bg_colors<B: Backend>(
    frame: &mut Frame<B>,
    normal_colors: [Color; 8],
    bright_colors: [Color; 8],
) {
    frame.render_widget(
        Paragraph::new("normal colors with white foreground"),
        Rect::new(0, 0, 50, 1),
    );
    for (i, color) in normal_colors.iter().enumerate() {
        let paragraph = Paragraph::new(color.to_string()).style(Style::default().bg(*color));
        frame.render_widget(paragraph, Rect::new(i as u16 * 8, 1, 8, 1));
    }
    frame.render_widget(
        Paragraph::new("bright colors with white foreground"),
        Rect::new(0, 2, 50, 1),
    );
    for (i, color) in bright_colors.iter().enumerate() {
        let paragraph = Paragraph::new(color.to_string()).style(Style::default().bg(*color));
        frame.render_widget(paragraph, Rect::new(i as u16 * 8, 3, 8, 1));
    }
    frame.render_widget(
        Paragraph::new("normal colors with black foreground"),
        Rect::new(0, 4, 50, 1),
    );
    for (i, color) in normal_colors.iter().enumerate() {
        let paragraph =
            Paragraph::new(color.to_string()).style(Style::default().bg(*color).fg(Color::Black));
        frame.render_widget(paragraph, Rect::new(i as u16 * 8, 5, 8, 1));
    }
    frame.render_widget(
        Paragraph::new("bright colors with black foreground"),
        Rect::new(0, 6, 50, 1),
    );
    for (i, color) in bright_colors.iter().enumerate() {
        let paragraph =
            Paragraph::new(color.to_string()).style(Style::default().bg(*color).fg(Color::Black));
        frame.render_widget(paragraph, Rect::new(i as u16 * 8, 7, 8, 1));
    }
}
```

examples/style.rs
```diff
@@ -0,0 +1,238 @@
...
+    render_indexed_colors(frame);
+    render_indexed_grayscale(frame);
+    render_modifiers(frame);
+}
+
+fn render_bg_colors<B: Backend>(
+    frame: &mut Frame<B>,
+    normal_colors: [Color; 8],
+    bright_colors: [Color; 8],
+) {
```
kdheepak, line 75: I think it's already good and I'm not sure how much work you want to put into this to make it more easier for users to read. Some minor thoughts I had were it might be nicer if it was written like this:

```rust
fn render_bg_colors<B: Backend>(
    frame: &mut Frame<B>,
    normal_colors: [Color; 8],
    bright_colors: [Color; 8],
) {
    render_paragraphs(frame, 0, normal_colors, Color::White, "normal bg with white fg");
    render_paragraphs(frame, 1, bright_colors, Color::White, "bright bg with white fg");
    render_paragraphs(frame, 2, normal_colors, Color::Black, "normal bg with black fg");
    render_paragraphs(frame, 3, bright_colors, Color::Black, "bright bg with white fg");
}
```

If I were a complete noob and see the example screenshot and then look at this code, I imagine this might be much more clear as to what is going on?

kdheepak (approved):
I think the `impl Display for Color` part of this PR is great, and the example you've added is good for that. I think most of my comments are about improving the screenshot for making it easier for beginners to grok and I imagine we can lump those changes into a separate discussion / PR where we modify all the examples.

joshka:
Ignore the push just now - I just saw the comments.

examples/style.rs
joshka: Sounds good

examples/style.rs
joshka: Sounds good

examples/style.rs
joshka: I ended up making a slew of changes to take some inspiration from https://gist.github.com/HaleTom/89ffe32783f89f403bba96bd7bcd1263, but that makes a whole tonne of vhs bug workaround that make the code even less readable :/

this is a good idea though.

examples/style.rs
joshka: The problem with doing this is that I'm already height constrained at least on the terminal settings I use on my Macbook 16in which hits 171x51, and width constrained as I want a consistent size for all the examples that fits well into the same width vhs dimensions (which ended up nicely hitting 80 chars).

Perhaps it's worth splitting colors and modifiers into two examples.

examples/style.rs
joshka: I did something similar to this.

examples/style.rs
joshka: I split the examples

examples/style.rs
joshka: Done

joshka:
I split the style example in two (colors and modifiers) and addressed the comments.
Switched to using layout rather than manually calculating the columns (note that this highlights bug https://github.com/ratatui-org/ratatui/issues/367)



I added a color to the example to make this useful to see non-grayscale interactions.

kdheepak:
That visually looks awesome now! 

Is the RGB section missing a section title?

joshka:
> Is the RGB section missing a section title?

Nope - it's all indexed colors 0-255. The gaps are there because there's sort of 3 groups of colors.

joshka:
Bumped this with some light changes:
- Added Color::Reset to the colors to be shown so that terminals that distinguish the default background from black will be more obvious
- Render the titles centered with a top border to make it a little more obvious that they apply to the colors below
- simplify some of the layout calls
- Fixed the GIF height to remove the space at the end
- Chose a theme (OceanicMaterial) that makes the background, foreground, black, white, dark gray, gray, and light vs normal colors distinct.
- Updated the readme with the new GIFs

colors
modifiers

joshka:
Rebased on main to check that the codecov setup should ignore changes to examples.

joshka:
Rebased on main and added tests for `impl Display for Color`.

kdheepak:
I have a minor preference for using reset for the background.

The top part of the screenshot is the PR unchanged:

image

And the bottom is when I use Reset for the indexed part of the demo.

This is with this colorscheme: https://github.com/catppuccin/iterm

joshka:
How does the VHS output look?

kdheepak:
The VHS output should remain unchanged, right? I can try it when I'm back at my personal computer. 

I think the screenshot looks great if people are just viewing it online (probably like 95% of folk). There's probably a handful of people that might try to run it. I don't think we should optimize for that use case, but if it is low effort it would be nice to accommodate that as well.

joshka:
(My concern was around the background bleed issue in VHS - it *should* render the same, but that bug is annoying). Changing to Reset works fine, and I managed to fix up the background problem by using a full block character.

  Made with VHS

<details>
<summary>Catppuccin Mocha version</summary>

  Made with VHS

</details>

kdheepak:
Neat!

kdheepak (approved):

joshka:
Thanks for the review @kdheepak :)

Issue #349, opened 2023-07-27
Block title_style is not rendered


joshka:
## Description
In the [recent PR to support multiple titles](https://github.com/ratatui-org/ratatui/pull/232), we accidentally made title_style no longer effective. It's useful to have this as a fallback style as it makes it easy to apply a style to all titles except one:
e.g. 
```rust
Block::new()
  .title_style(Style::new().bold())
  .title(Title::from("foo").alignment(Alignment::Left))
  .title(Title::from("bar").alignment(Alignment::Center))
  .title(Title::from("foo".not_bold().dim()).alignment(Alignment::Right))
```

## To Reproduce
```rust
    frame.render_widget(
        Block::new()
            .borders(Borders::all())
            .title("Styled Title")
            .title_style(Style::new().yellow()),
        area1,
    );
    frame.render_widget(
        Block::new()
            .borders(Borders::all())
            .title(Title::from(Line::styled(
                "Styled Line",
                Style::new().yellow(),
            ))),
        area 2,
    );
```

## Expected behavior
The Block's title style should be applied to the title.
The order which styles should effect the title (later styles should override earlier):
1. Any style applied to the surrounding area
2. `Block.style`
3. `Block.border_style`
4. `Block.title_style` - this is the missing piece
5. `Title.content.style` (`Line.style`): Out of scope for this issue as `Line` currently doesn't have a style, but should
6. `Title.content.spans` -> `Span.style`

## Screenshots
<img width="399" alt="image" src="https://github.com/ratatui-org/ratatui/assets/381361/1cb9e23e-2634-4e0b-860d-156b145c60a2">

## Environment
- OS: Mac
- Terminal Emulator: iterm2
- Font: FiraCode NF
- Crate version: 0.22
- Backend: crossterm

## Additional context

This is the code that needs to be modified for this:

https://github.com/ratatui-org/ratatui/blob/bf4944683d6afb6f42bec80a1bd308ecdac50cbc/src/widgets/block.rs#L418-L423
https://github.com/ratatui-org/ratatui/blob/bf4944683d6afb6f42bec80a1bd308ecdac50cbc/src/widgets/block.rs#L444-L449
https://github.com/ratatui-org/ratatui/blob/bf4944683d6afb6f42bec80a1bd308ecdac50cbc/src/widgets/block.rs#L465-L470

There's a simple ways to approach this (just create a new line from the old line with the appropriate title_style applied before the line's style).

[...]

PR #351, opened 2023-07-27
docs(examples): Update block example

joshka:
Block example

### DIFF
--- a/examples/block.rs
+++ b/examples/block.rs
@@ -1,123 +1,253 @@
-use std::{error::Error, io, time::Duration};
+use std::{
+    error::Error,
+    io::{stdout, Stdout},
+    ops::ControlFlow,
+    time::Duration,
+};
 
 use crossterm::{
-    event::{self, DisableMouseCapture, EnableMouseCapture, Event, KeyCode},
+    event::{self, Event, KeyCode},
     execute,
     terminal::{disable_raw_mode, enable_raw_mode, EnterAlternateScreen, LeaveAlternateScreen},
 };
-use ratatui::{prelude::*, widgets::*};
+use itertools::Itertools;
+use ratatui::{
+    prelude::*,
+    widgets::{
+        block::{Position, Title},
+        Block, BorderType, Borders, Padding, Paragraph, Wrap,
+    },
+};
 
-fn main() -> Result<(), Box<dyn Error>> {
-    // setup terminal
-    enable_raw_mode()?;
-    let mut stdout = io::stdout();
-    execute!(stdout, EnterAlternateScreen, EnableMouseCapture)?;
-    let backend = CrosstermBackend::new(stdout);
-    let mut terminal = Terminal::new(backend)?;
+// These type aliases are used to make the code more readable by reducing repetition of the generic
+// types. They are not necessary for the functionality of the code.
+type Frame<'a> = ratatui::Frame<'a, CrosstermBackend<Stdout>>;
+type Terminal = ratatui::Terminal<CrosstermBackend<Stdout>>;
+type Result<T> = std::result::Result<T, Box<dyn Error>>;
 
-    // create app and run it
-    let res = run_app(&mut terminal);
+fn main() -> Result<()> {
+    let mut terminal = setup_terminal()?;
+    let result = run(&mut terminal);
+    restore_terminal(terminal)?;
 
-    // restore terminal
-    disable_raw_mode()?;
-    execute!(
-        terminal.backend_mut(),
-        LeaveAlternateScreen,
-        DisableMouseCapture
-    )?;
-    terminal.clear()?;
-    terminal.show_cursor()?;
-
-    if let Err(err) = res {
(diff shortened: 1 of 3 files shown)


examples/block.rs
```diff
@@ -1,123 +1,248 @@
...
-fn main() -> Result<(), Box<dyn Error>> {
-    // setup terminal
-    enable_raw_mode()?;
-    let mut stdout = io::stdout();
-    execute!(stdout, EnterAlternateScreen, EnableMouseCapture)?;
-    let backend = CrosstermBackend::new(stdout);
-    let mut terminal = Terminal::new(backend)?;
+type Frame<'a> = ratatui::Frame<'a, CrosstermBackend<Stdout>>;
+type Terminal = ratatui::Terminal<CrosstermBackend<Stdout>>;
+type Result<T> = std::result::Result<T, Box<dyn Error>>;
```
orhun, line 24: Why exactly these type aliases are needed and simple imports are not sufficient?

(I can imagine a situation where a Rust beginner using these examples and asking the same question.)

joshka: The idea behind this was to avoid repetition of the generic constraints in the render_* methods.
- The `Result` type alias is a fairly common practice to see, both in app code and library code (e.g. std::io::Result. In app code often people use `use anyhow::Result`).
- The `Frame` alias would expand to either `fn render_xxx<B: Backend>(frame: Frame<B>)` or the full `fn render_xxx(frame: Frame<'a, CrosstermBackend<Stdout>>)` across 11 methods that are arranged that way to make it easy to see structurally what's happening. 
- The `Terminal` alias is to my mind the most debatable choice as it's only 4 usages. It's there mainly for consistency with the choice to alias `Frame`.

I see these as simplifications generally, but given your question about why they are here, they sound like they could be changes that add complexity rather than reduce it. Would a comment explaining them work better? Perhaps:
```suggestion
// type aliases to reduce repetition of the generic parameters in method parameters
type Frame<'a> = ratatui::Frame<'a, CrosstermBackend<Stdout>>;
type Terminal = ratatui::Terminal<CrosstermBackend<Stdout>>;
type Result<T> = std::result::Result<T, Box<dyn Error>>;
```

joshka:
I have several goals in re-doing this example:
- to create consistency the the setup / restore code, so we can document that once for all the examples
- to use similar styles, colors, sizing, etc. across widgets so it looks like a harmonious set of examples
- to show off every feature of the widget (a method per block makes it easy to see which code draws which block)
- to easily see catch visual regressions (e.g. this example highlights the [title style bug](https://github.com/ratatui-org/ratatui/issues/349))
- to be able to see visually how any PRs change the display of a widget

My hope is that by starting with a single easy example and then doing a more complex one, we have a couple of high quality examples that we can point at and then make the other widgets consistent with (e.g. add some issues labelled `good first issue`). Keeping the example code super simple and obvious makes it easy to update when we add / change functionality.

joshka:
Added a couple of comments, and removed unnecessary anonymous lifetimes on `Paragraph`s.

https://github.com/ratatui-org/ratatui/blob/52c660aa326c424c9aa554b22b8bd78996a4c3f1/examples/block.rs#L22-L26

https://github.com/ratatui-org/ratatui/blob/52c660aa326c424c9aa554b22b8bd78996a4c3f1/examples/block.rs#L103-L106

kdheepak:
I like having an example like this!

Thoughts on making this interactive? If I could hit `j` and `k` to cycle through the various options you have here and the gif showed all cycles, that would be visually easier to grok.

Also, maybe for styled borders, only change foreground colors? (this is feedback that I think could go in a separate PR about making everything prettier)

joshka:
Those are both good points.

I think it’s worth making interactive examples separate to the examples that demonstrate how things look. There’s a lot of ways of doing interactivity that work well small for demos but don’t scale well to apps that have more a diverse UI.

I experimented a little with having a cli flag to to choose the example to display so we could use VHS to generate smaller images to include directly in the docs, but it’s difficult to keep it simple with that approach. This example would be easy to refactor to add interactivity though, so perhaps let’s revisit the idea after this.

My rationale behind the styled borders using the background color was to make it easy to contrast the difference between the block style and border style. The background makes it obvious how the dark gray styling on the paragraph combines with the background block style. Background color also helps show which things affect blank space not just the text (which the foreground doesn’t). Do you think this makes sense?

Overall my goal is to try to pare this down to the absolute minimal amount of code necessary to show the info we need. There’s some ideas I have about fixing the backend implementations that could make this even smaller.

examples/block.rs
orhun: Yeah, I would say let's add a comment to make their usage clear-er.

joshka: The newer version has the comments

orhun (approved):

joshka:
Merging this to make it easy to see that https://github.com/ratatui-org/ratatui/pull/363 fixes the bug

PR #359, opened 2023-07-31
feat(release): add automated nightly releases

### LINKED ISSUE #147: Nightly releases
## Problem

There might be some time gaps between new releases so the users cannot test/use the latest functionality without using `ratatui` as a `git` dependency in `Cargo.toml` which makes the crate _unpublishable_ on crates.io. (crates.io doesn't accept `git` dependencies)

See what `atuin` did recently as a workaround: https://github.com/ellie/atuin/commit/a515b06bcb556c1be2d0fc3095cd778d413fe40d

## Solution

Create automated nightly releases every week or so. `ratatui` can be used as e.g. `version = "nightly-2023-04-21"` afterwards. 

## Alternatives

Just plan things out and release more frequently.

## Additional context

None.

---

We need comments/thoughts on this.

orhun:
As discussed in the [last meeting](https://github.com/ratatui-org/ratatui/discussions/316), this PR updates the release workflow (`cd.yml`) for creating periodic nightly releases.

The version schema is `<version>-alpha.<num>` and `<num>` is increased each time the workflow runs. It is also possible to trigger a nightly release manually via the GitHub interface (see [`workflow_dispatch`](https://docs.github.com/en/actions/using-workflows/manually-running-a-workflow)).

Tested here: https://github.com/orhun/tui-rs-revival/tags

Closes #147

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -15,8 +15,8 @@ defaults:
     shell: bash
 
 jobs:
-  publish-nightly:
-    name: Create a nightly release
+  publish-alpha:
+    name: Create an alpha release
     runs-on: ubuntu-latest
     if: ${{ !startsWith(github.event.ref, 'refs/tags/v') }}
     steps:
@@ -42,7 +42,7 @@ jobs:
           msg="# crate version"
           sed -E -i "s/^version = .* ${msg}$/version = \"${next_tag#v}\" ${msg}/" Cargo.toml
           echo "NEXT_TAG=${next_tag}" >> $GITHUB_ENV
-          echo "Next nightly release: ${next_tag} 🐭"
+          echo "Next alpha release: ${next_tag} 🐭"
 
       - name: Publish on crates.io
         uses: actions-rs/cargo@v1
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -27,16 +27,20 @@ jobs:
 
       - name: Calculate the next release
         run: |
-          suffix="-alpha"
+          suffix="alpha"
           last_tag="$(git describe --abbrev=0 --tags `git rev-list --tags --max-count=1`)"
-          if [[ "${last_tag}" = *"${suffix}"* ]]; then
+          if [[ "${last_tag}" = *"-${suffix}"* ]]; then
             # increment the alpha version
-            alpha=$(echo "${last_tag}" | grep -oE '([0-9]+)$')
-            next_alpha=$((alpha + 1))
-            next_tag=$(echo "${last_tag}" | sed "s/\.[0-9]\+$/\.${next_alpha}/")
+            # e.g. v0.22.1-alpha.12 -> v0.22.1-alpha.13
+            alpha="${last_tag##*-${suffix}.}"
+            next_alpha="$((alpha + 1))"
+            next_tag="${last_tag/%${alpha}/${next_alpha}}"
           else
-            # start the alpha version from 0
-            next_tag="${last_tag}${suffix}.0"
+            # increment the patch and start the alpha version from 0
+            # e.g. v0.22.0 -> v0.22.1-alpha.0
+            patch="${last_tag##*.}"
+            next_patch="$((patch + 1))"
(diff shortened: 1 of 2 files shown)


mindoodoo (approved):

.github/workflows/cd.yml
```diff
@@ -1,18 +1,79 @@
...
+          suffix="-alpha"
+          last_tag="$(git describe --abbrev=0 --tags `git rev-list --tags --max-count=1`)"
+          if [[ "${last_tag}" = *"${suffix}"* ]]; then
+            # increment the alpha version
+            alpha=$(echo "${last_tag}" | grep -oE '([0-9]+)$')
+            next_alpha=$((alpha + 1))
+            next_tag=$(echo "${last_tag}" | sed "s/\.[0-9]\+$/\.${next_alpha}/")
+          else
+            # start the alpha version from 0
+            next_tag="${last_tag}${suffix}.0"
```
joshka, line 39: I think we probably need to update the patch version here:
```suggestion
            # increment the patch and start the alpha version from 0 e.g. v0.22.0 -> v0.22.1-alpha.0
            patch=${last_tag##*.}
            next_patch=$((patch + 1))
            next_tag="${last_tag/%${patch}/${next_patch}}-${suffix}.0"
```

.github/workflows/cd.yml
```diff
@@ -1,18 +1,79 @@
...
     steps:
       - name: Checkout the repository
         uses: actions/checkout@v3
-      - name: Publish
+        with:
+          fetch-depth: 0
+
+      - name: Calculate the next release
+        run: |
+          suffix="-alpha"
```
joshka, line 30: (nit: to go along with the suggestions below), it might be easier to drop the hyphen here and add it to the bits below
```suggestion
          suffix="alpha"
```

.github/workflows/cd.yml
```diff
@@ -1,18 +1,79 @@
...
+          last_tag="$(git describe --abbrev=0 --tags `git rev-list --tags --max-count=1`)"
+          if [[ "${last_tag}" = *"${suffix}"* ]]; then
+            # increment the alpha version
+            alpha=$(echo "${last_tag}" | grep -oE '([0-9]+)$')
+            next_alpha=$((alpha + 1))
+            next_tag=$(echo "${last_tag}" | sed "s/\.[0-9]\+$/\.${next_alpha}/")
```
joshka, line 36: nit: no need for sed/grep here. [Bash is enough](https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html).

```suggestion
          if [[ "${last_tag}" = *"-${suffix}"* ]]; then
            # increment the alpha version e.g. v0.22.1-alpha.12 -> v0.22.1-alpha.13
            alpha=${last_tag##*-${suffix}.}
            next_alpha=$((alpha + 1))
            next_tag=${last_tag/%${alpha}/${next_alpha}}
```

joshka (changes requested):
Semver suggests that 0.22.0-whatever is before 0.22.0. We should go to 0.22.1-alpha.0 from 0.22.0.

Otherwise, looks good.

.github/workflows/cd.yml
```diff
@@ -1,18 +1,79 @@
...
+  run:
+    shell: bash
+
 jobs:
-  publish:
-    name: Publish on crates.io
+  publish-nightly:
+    name: Create a nightly release
     runs-on: ubuntu-latest
+    if: ${{ !startsWith(github.event.ref, 'refs/tags/v') }}
```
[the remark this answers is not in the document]
orhun: To be honest, I like how we only have 2 workflow files for CI/CD as `ci.yml` and `cd.yml` 🐻 

Also what would we call them if we split it? `publish-alpha.yml` and `publish-stable.yml`?

joshka (approved):
LGTM ααα

.github/workflows/cd.yml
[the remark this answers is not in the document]
joshka: no probs - let's leave it like this.

PR #361, opened 2023-07-31
chore(toolchain)!: bump msrv to 1.67

a-kenji:
This updates the msrv to  `1.67`.

Because an optional dependency of ratatui requires it.

Alternatives: Choose to not bump the versions for optional dependencies, or pin the time version to the latest version that supports `1.65`

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,7 +150,7 @@ you are interested in working on a PR or issue opened in the previous repository
 
 ## Rust version requirements
 
-Since version 0.21.0, The Minimum Supported Rust Version (MSRV) of `ratatui` is 1.65.0.
+Since version 0.21.0, The Minimum Supported Rust Version (MSRV) of `ratatui` is 1.67.0.
 
 ## Documentation
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -93,7 +93,7 @@ jobs:
     strategy:
       matrix:
         os: [ ubuntu-latest, windows-latest, macos-latest ]
-        toolchain: [ "1.65.0", "stable" ]
+        toolchain: [ "1.67.0", "stable" ]
     runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout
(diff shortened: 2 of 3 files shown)


TieWay59:
LGTM. Many CI-checks failed on this.

orhun:
Hey, what is the exact reason for this bump? i.e. which optional dependency is that? I'm asking because I'm a proponent of avoiding MSRV bump as much as possible.

a-kenji:
It is the current reason that the ci for `1.65` fails on some current PR's.

> error: package `time v0.3.24` cannot be built because it requires rustc 1.67.0 or newer, while the currently active rustc version is 1.65.0

I believe it is the optional `calendar` feature that pulls this dependency in.

sophacles (approved):

joshka:
The N-2 policy change for the `time` crate is documented at https://github.com/time-rs/time/discussions/535.. Having read some of that thread (but not all, it's long), the arguments sound reasonable and I think we shouldn't avoid MSRV bumps.

I'd prefer to avoid the pin `time` approach (as that probably requires extra manual effort to keep abreast of any CVEs etc.), and so that the only other approach is to remove the calendar widget from the repo. I don't think that's reasonable either.

I'm approving this, but won't merge as I think it's worth hearing @orhun's perspective on this before merging.

joshka (approved):

orhun (approved):
Sounds reasonable, let's bump it up! 🆙

joshka (approved):

README.md
```diff
@@ -150,7 +150,7 @@ you are interested in working on a PR or issue opened in the previous repository
 
 ## Rust version requirements
 
-Since version 0.21.0, The Minimum Supported Rust Version (MSRV) of `ratatui` is 1.65.0.
```
Valentin271, line 153: I think *"Since version 0.21.0"* should be updated to *0.23.0* as the MSRV will be bumped in the next release.

orhun: Ah, right. Feel free to submit a PR!

Valentin271: Done at #369 :)

PR #363, opened 2023-08-03
fix(block): Fixed title_style not rendered

### LINKED ISSUE #349: Block title_style is not rendered

Valentin271:
This PR fixes #349 the easy way (see issue). I think the _complex fix_ requires another separate PR.

I added a unit test and made sure it didn't pass prior to this modification. I added a lot of test cases because there are three separate places where this bug can occur.

I didn't change the corresponding example to demonstrate the behavior as there is already an [opened PR](https://github.com/ratatui-org/ratatui/pull/351) that will likely be merged.

I didn't want to create a new spans (or lines) because in the future they might contain more than just `content` and `style` and this would lead to some half cloned spans. Thus I preferred cloning and modifying them locally.

### DIFF
--- a/src/widgets/block.rs
+++ b/src/widgets/block.rs
@@ -415,10 +415,16 @@ impl<'a> Block<'a> {
                 let title_x = current_offset;
                 current_offset += title.content.width() as u16 + 1;
 
+                // Clone the title's content, applying block title style then the title style
+                let mut content = title.content.clone();
+                for span in content.spans.iter_mut() {
+                    span.style = self.titles_style.patch(span.style);
+                }
+
                 buf.set_line(
                     title_x + area.left(),
                     self.get_title_y(position, area),
-                    &title.content,
+                    &content,
                     title_area_width,
                 );
             });
@@ -441,10 +447,16 @@ impl<'a> Block<'a> {
             let title_x = current_offset;
             current_offset += title.content.width() as u16 + 1;
 
+            // Clone the title's content, applying block title style then the title style
+            let mut content = title.content.clone();
+            for span in content.spans.iter_mut() {
+                span.style = self.titles_style.patch(span.style);
+            }
+
             buf.set_line(
                 title_x + area.left(),
                 self.get_title_y(position, area),
-                &title.content,
+                &content,
                 title_area_width,
             );
         });
@@ -462,10 +474,16 @@ impl<'a> Block<'a> {
                 current_offset += title.content.width() as u16 + 1;
                 let title_x = current_offset - 1; // First element isn't spaced
 
+                // Clone the title's content, applying block title style then the title style
+                let mut content = title.content.clone();
+                for span in content.spans.iter_mut() {
+                    span.style = self.titles_style.patch(span.style);
+                }
+
                 buf.set_line(
                     area.width.saturating_sub(title_x) + area.left(),
                     self.get_title_y(position, area),
-                    &title.content,
+                    &content,
                     title_area_width,
                 );
             });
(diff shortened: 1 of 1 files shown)


src/widgets/block.rs
```diff
@@ -415,10 +415,16 @@ impl<'a> Block<'a> {
                 let title_x = current_offset;
                 current_offset += title.content.width() as u16 + 1;
 
+                // Clone the title's content, applying block title style then the title style
+                let mut content = title.content.clone();
+                for sp in content.spans.iter_mut() {
+                    sp.style = self.titles_style.patch(sp.style);
```
joshka, line 421: Small nitpick: I'd generally avoid abbreviations when possible.

src/widgets/block.rs
```diff
@@ -917,4 +936,50 @@ mod tests {
             assert_buffer_eq!(buffer, Buffer::with_lines(vec![expected]));
         }
     }
+
+    #[test]
+    fn title_style_overrides_block_title_style() {
```
joshka, line 941: I have a strong preference for keeping tests super simple. I found myself scrolling up to the test data and back down to the test itself to cross reference the data and behavior. We're testing 4 things that matter over 45 lines, which can be expressed as perhaps 4 tests each ~9 lines. The main benefit of splitting things this way is that the tests can help document the code. E.g. what about something like:

```rust
    #[test]
    fn title_content_style_renders_style() {
        for alignment in [Alignment::Left, Alignment::Center, Alignment::Right] {
            let mut buffer = Buffer::empty(Rect::new(0, 0, 4, 1));
            Block::default()
                .title_alignment(alignment)
                .title("test".yellow())
            let mut expected = Buffer::with_lines(vec!["test"]);
            expected.set_style(expected.area, Style::new().yellow());
        }
    }
```
etc.

joshka (approved):
Thanks for the PR and welcome to Ratatui.

Generally LGTM. I had a nitpick and a preference that has to do with keeping tests simple. Can be merged as-is or updated to address the feedback (both are not huge points).

src/widgets/block.rs
Valentin271: Fixed :)

src/widgets/block.rs
Valentin271: I was was testing basically 3 different cases, there are now separated in 3 tests and are easier to understand.

joshka (approved):
LGTM - Thanks for the fix. We appreciate your effort.

joshka:
Merging this even though checks are failing as the failures are due to the MSRV issue.

Issue #137, opened 2023-04-15
Add benchmarking for widgets


orhun:
## Problem

There might be performance issues/concerns about tui widgets and adding benchmarks is important to uncover them.

## Solution

Add benches for widgets. See #22 for an example.

## Alternatives

None.

## Additional context

#22

Sufilevy:
I've looked at the linked issue, and tried to add a bench (for `Block`, just to try).

```rs
    #[bench]
    fn bench_title_rounded_corners(b: &mut test::Bencher) {
        b.iter(|| {
            Block::default()
                .borders(Borders::ALL)
                .title("Main block with round corners")
                .title_alignment(Alignment::Center)
                .border_type(BorderType::Rounded)
        })
    }
```

Does this look okay? I just took one of the example usages of `Block` in the repo, and I put it in a `bench`. Am I missing something? Is this the right direction for writing benches fi the widgets?

sophacles:
It's a good start. Most of the work for widgets is done in rendering, so I think it would be a good idea to set up a buffer outside of the  measured func, and render to that buffer in the measured code.

Sufilevy:
When benching the rendering of widgets, should I use the `TestBackend`?
Or should I create different rendering benches for `crossterm` and `termion`?

Sufilevy:
Got it. Thanks!

Valentin271:
Are you working on this @Sufilevy (or anyone else) ? Otherwise I'm interested in submitting a PR.

Sufilevy:
You're welcome to!

EdJoPaTo:
> Most of the work for widgets is done in rendering, so I think it would be a good idea to set up a buffer outside of the measured func, and render to that buffer in the measured code.

Most widgets are created on every render, so I think they should also be included in the benchmarks. Either they are fast and no problem or an issue shows up.

PR #368, opened 2023-08-05
test(block): add benchmarks

Valentin271:
# Description

I added benchmarks to the `block` widget to uncover eventual performance issues.  
I tried to cover common (creation, render) or expensive (inner area) calculations. I also ran the benchmark with three different common buffer sizes.  
From what I can see it doesn't seem to have any bottleneck here.

What is covered by the benches :

- `Block` creation
- Render with different parameters (empty, titles, borders, padding)
- Computing inner area

# Details

- This PR adds benchmarks only for `block`. I think it's a better idea to split the work on benchmarks to add them incrementally.

- I'm not sure the `black_box` is really necessary for the `block` creation bench. Otherwise the compiler advises to put the `new` function reference as parameter (see below). I'm afraid this is less legible and might not work as expected.
```rust
group.bench_function("new", |b| b.iter(Block::new));
```

- Lastly, I have a small concern about my benches reports not looking the same as the `paragraph` ones : 
image
I based myself off of this bench (the `paragraph` one), so I'm not sure this is the expected behavior (works in the terminal and results are correct once clicked on though).

---
This PR relates to #137.

Thank you for your review and your work.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,6 +61,11 @@ rand = "0.8"
 name = "paragraph"
 harness = false
 
+[[bench]]
+name = "block"
+harness = false
+
+
 [[example]]
 name = "barchart"
 required-features = ["crossterm"]
--- a/benches/block.rs
+++ b/benches/block.rs
@@ -1,0 +1,58 @@
+use criterion::{criterion_group, criterion_main, Bencher, BenchmarkId, Criterion};
+use ratatui::{
+    buffer::Buffer,
+    layout::Rect,
+    prelude::Alignment,
+    widgets::{
+        block::{Position, Title},
+        Block, Borders, Padding, Widget,
+    },
+};
+
+/// Benchmark for rendering a block.
+pub fn block(c: &mut Criterion) {
+    let mut group = c.benchmark_group("block");
+
+    for buffer_size in &[
+        Rect::new(0, 0, 100, 50),  // vertically split screen
+        Rect::new(0, 0, 200, 50),  // 1080p fullscreen with medium font
+        Rect::new(0, 0, 256, 256), // Max sized area
+    ] {
+        let buffer_area = buffer_size.area();
+
+        // Render an empty block
+        group.bench_with_input(
+            BenchmarkId::new("render_empty", buffer_area),
+            &Block::new(),
+            |b, block| render(b, block, buffer_size),
+        );
+
+        // Render with all features
+        group.bench_with_input(
+            BenchmarkId::new("render_all_feature", buffer_area),
+            &Block::new()
+                .borders(Borders::ALL)
+                .title("test title")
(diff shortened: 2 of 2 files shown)


joshka:
I think you probably need to have a numeric value to get the output like the paragraph (could use `Rect::area()` for this value)

joshka (changes requested):
LGTM. I'd be surprised if this really ever has significant perf issues.

To keep the benchmark times down, it's worth consolidating some of the benches. E.g. forget about the bare `new()` bench. This was only necessary in the paragraph benches as we're testing some fairly large strings (~200*65535 chars) to understand the limits of scrolling and wrapping.

I suspect (without running these) that the render times are likely fairly similar (or the absolute value is small enough that the relative difference is irrelevant). In which case, it's probably sufficient to bench an empty block, and one with all the features (titles, borders, styles, padding etc.)

Can you paste some of the output metrics into the PR? It doesn't hurt to have them recorded for posterity later.

Valentin271:
> I think you probably need to have a numeric value to get the output like the paragraph (could use Rect::area() for this value)

Yes, you're right. I used area as suggested. Too bad criterion cannot use text as a table entry :/ 

> LGTM. I'd be surprised if this really ever has significant perf issues.

Same here, just thought I'd benchmark every building block starting with the easiest for me.

> I suspect (without running these) that the render times are likely fairly similar (or the absolute value is small enough that the relative difference is irrelevant). In which case, it's probably sufficient to bench an empty block, and one with all the features (titles, borders, styles, padding etc.)

After modifying the benchmark according to your review and looking more in depth, render times **are** similar : only 10 to 50 µs between empty and with all features for the biggest area (depending on the run, on my machine of course).

> Can you paste some of the output metrics into the PR? It doesn't hurt to have them recorded for posterity later.

Sure, here are some comparisons I found interesting (keep in mind this was recorded on my machine, you might not get the same results) :

_Figure 1: Time to render all features relative to the buffer area_
image

_Figure 2: Time to render an empty block relative to the buffer area_
image

_Figure 3: Time to render an empty block and one with all features on a maxed ou area_
image

_Figure 4: Time to render an empty block and one with all features on a common area (200x50) _
image

The last two graphs are interesting because we can see a difference when rendering _all features_ on _figure 3_ (which is to be expected) but when used in a common context there's actually no visible difference (_figure 4_).

---

- I removed the smallest area (20x10) because it was always about 1µs, I don't those times are really of any use
- I also removed the `inner` computation as it was taking 5 ns constantly

joshka:
Looks good - I think Figure 4 is the one that matters the most here. And it's fast enough.
Thanks for doing this - it's another good base test to add to.

joshka (approved):

orhun (approved):

PR #371, opened 2023-08-05
docs(layout::Constraint): add doc-comments for all variants

### LINKED ISSUE #354: Layout: Documentation for all `Constraint` variants
## Problem
Currently all variants for `Constraint` dont have documentation, making it difficult to understand the difference between `Max` and `Length` and how they are meant to behave

## Solution
Add rust-doc comments explaining how the variants are meant to be used, like `Length is similar to Max but does ... differently`

it would also be great to have a example for all options (example `layout.rs` is currently missing `Constraint::Max` example)

## Additional context
i am coming from `tui-rs` after having learned that it is not maintained anymore, and just found that `Length` is very similar to `Max` and dont quite understand the difference (if it still had been maintained i would also still have opened the issue there for `Constraint` documentation)

hasezoey:
This PR adds at least some documentation on what each variant does, with examples (and so doc-tests)

fixes #354

### DIFF
--- a/src/layout.rs
+++ b/src/layout.rs
@@ -27,12 +27,64 @@ pub enum Direction {
     Vertical,
 }
 
+/// Constraints to apply  
 #[derive(Debug, Clone, Copy, Eq, PartialEq, Hash)]
 pub enum Constraint {
+    /// Apply a percentage to a given amount  
+    /// Converts the given percentage to a f32, and then converts it back, trimming off the decimal
+    /// point (effectively rounding down)
+    /// ```
+    /// # use ratatui::prelude::Constraint;
+    /// assert_eq!(0, Constraint::Percentage(50).apply(0));
+    /// assert_eq!(2, Constraint::Percentage(50).apply(4));
+    /// assert_eq!(5, Constraint::Percentage(50).apply(10));
+    /// assert_eq!(5, Constraint::Percentage(50).apply(11));
+    /// ```
     Percentage(u16),
+    /// Apply a ratio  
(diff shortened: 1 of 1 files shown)


joshka:
See https://github.com/ratatui-org/ratatui/issues/354#issuecomment-1666512893 for some detail on the panic. I was mistaken.

src/layout.rs
```diff
@@ -27,12 +27,67 @@ pub enum Direction {
     Vertical,
 }
 
+/// Constraints to apply  
+/// Currently there is no error or panic when a constraint cannot be fulfilled, and behavior for that is undefined (see <https://github.com/ratatui-org/ratatui/issues/354>)
```
hasezoey, line 31: > See [#354 (comment)](https://github.com/ratatui-org/ratatui/issues/354#issuecomment-1666512893) for some detail on the panic. I was mistaken.

what should i change it to instead?

joshka: It's the other functions that use the constraint that panic not the constraint itself (right?), so perhaps just omit the note about panics here.

joshka (changes requested):
This looks great. There was some discussion about when to use doctests vs unit tests in https://github.com/ratatui-org/ratatui/discussions/324. I think this is a perfect example of when asserts in doc tests are useful as it helps show much more detail about the edge cases. If these were instead unit tests, they would not be as discoverable.

Change requested is just the comments about the panicing / github comment.

joshka (approved):

Issue #372, opened 2023-08-05
Ignore examples in code coverage


joshka:
This is a small task to add a codecov configuration file and exclude the examples folder from coverage so that it doesn't fail in CI.

PR #373, opened 2023-08-05
ci(coverage): exclude examples directory from coverage

a-kenji:
Closes:  #372

Alternative: Use `cargo make` also in `bacon`, that way updates need to be only made in the `Makefile.toml`

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,2 @@
+ignore:
+	- "examples"


joshka:
I thought this would probably be via https://docs.codecov.com/docs/ignoring-paths rather than tweaking the lcov output. Not 100% sure though as I haven't looked deeply. Do you have any guidance on that?

orhun:
Yeah, this should be done via `ignore` directive in `codecov.yml`

a-kenji:
I wasn't sure about the change. 

[This](https://docs.codecov.com/docs/supported-languages):

> Since Codecov works by ingesting your [coverage report](https://docs.codecov.com/docs/codecov-uploader)s from the testing tools, the key component needed for the integration is the ability to upload a coverage report (ideally from your CI provider) to Codecov.

Lead me to believe if we exclude the directory from the report itself, it wouldn't be ingested. But that doesn't seem what is happening.

I'll add a configuration file.

PR #375, opened 2023-08-06
feat(widgets::table): add option to always allocate the "selection" constraint

hasezoey:
This PR adds a option for `Table` to always allocate the "selection column", regardless of if something is selected, this way there is less "moving around" of the layout.

The default is still `false` (only allocated if something is selected), but it can both be enabled or disabled in a builder function.

Note: also renamed some internal variables to better reflect on what they do / track

### DIFF
--- a/src/widgets/table.rs
+++ b/src/widgets/table.rs
@@ -160,23 +160,24 @@ impl<'a> Styled for Row<'a> {
     }
 }
 
-/// This option allows to configure the behavior of the "highlight symbol" column width allocation
+/// This option allows the user to configure the "highlight symbol" column width spacing
 #[derive(Debug, PartialEq, Eq, Clone, Default, Hash)]
 pub enum HighlightSpacing {
-    /// Always add spacing for the row selection symbol
+    /// Always add spacing for the selection symbol column
     ///
     /// With this variant, the column for the selection symbol will always be allocated, and so the
     /// table will never change size, regardless of if a row is selected or not
     Always,
-    /// Only add spacing for the row selection symbol, if a row is selected
+    /// Only add spacing for the selection symbol column if a row is selected
     ///
     /// With this variant, the column for the selection symbol will only be allocated if there is a
     /// selection, causing the table to shift if selected / unselected
     #[default]
     WhenSelected,
-    /// Never add spacing selection symbol spacing, regardless of if something is selected or not
+    /// Never add spacing to the selection symbol column, regardless of whether something is
+    /// selected or not
     ///
-    /// This effectively changes that the highlight symbol will never be drawn
+    /// This means that the highlight symbol will never be drawn
     Never,
 }
 
@@ -323,7 +324,7 @@ impl<'a> Table<'a> {
 
     /// Set when to show the highlight spacing
     ///
-    /// see [HighlightSpacing] about which variant affects spacing in which way
+    /// See [HighlightSpacing] about which variant affects spacing in which way
     pub fn highlight_spacing(mut self, value: HighlightSpacing) -> Self {
         self.highlight_spacing = value;
         self
--- a/src/widgets/table.rs
+++ b/src/widgets/table.rs
@@ -320,7 +320,9 @@ impl<'a> Table<'a> {
         self
     }
 
-    fn get_columns_widths(&self, max_width: u16, add_selection_space: bool) -> Vec<u16> {
+    /// Get all offsets and widths of all user specified columns
+    /// Returns (x, width)
+    fn get_columns_widths(&self, max_width: u16, add_selection_space: bool) -> Vec<(u16, u16)> {
         let mut constraints = Vec::with_capacity(self.widths.len() * 2 + 1);
         if add_selection_space {
             let highlight_symbol_width = self.highlight_symbol.map_or(0, |s| s.width() as u16);
@@ -347,7 +349,7 @@ impl<'a> Table<'a> {
         if add_selection_space {
             chunks = &chunks[1..];
         }
-        chunks.iter().step_by(2).map(|c| c.width).collect()
+        chunks.iter().step_by(2).map(|c| (c.x, c.width)).collect()
     }
 
     fn get_row_bounds(
@@ -462,7 +464,6 @@ impl<'a> StatefulWidget for Table<'a> {
         };
         let columns_widths = self.get_columns_widths(table_area.width, add_selection_space);
         let highlight_symbol = self.highlight_symbol.unwrap_or("");
-        let blank_symbol = " ".repeat(highlight_symbol.width());
         let mut current_height = 0;
         let mut rows_height = table_area.height;
 
@@ -478,22 +479,18 @@ impl<'a> StatefulWidget for Table<'a> {
                 },
                 header.style,
             );
-            let mut col = table_area.left();
-            if add_selection_space {
-                col += (highlight_symbol.width() as u16).min(table_area.width);
-            }
-            for (width, cell) in columns_widths.iter().zip(header.cells.iter()) {
+            let inner_offset = table_area.left();
+            for ((x, width), cell) in columns_widths.iter().zip(header.cells.iter()) {
                 render_cell(
                     buf,
                     cell,
                     Rect {
-                        x: col,
+                        x: inner_offset + x,
                         y: table_area.top(),
                         width: *width,
                         height: max_header_height,
                     },
                 );
-                col += *width + self.column_spacing;
             }
             current_height += max_header_height;
             rows_height = rows_height.saturating_sub(max_header_height);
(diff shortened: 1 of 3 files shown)


hasezoey:
should i do a similar thing for `List`?

tests/widgets_table.rs
joshka, line 1: I'd prefer to see unit tests (e.g. in `src/widgets/table.rs`) that just test the feature over integration tests that test the feature with a bunch of other things. (header / block / etc.). There's two reasons for this - simplicity, and reducing test failure surface.

These tests can each be about ~15 lines of code that are pure linear code (arrange, act, assert):

```rust
fn render_with_always_show_highlight_allocates_space() {
  let buffer = Buffer::empty(Rect(0, 0, 8, 3));
  let table = Table::new(vec![Row::new(vec!["a", "b", "c"])]; 2)
    .highlight_symbol(">>>")
    .highlight_show(Show::Always);
  table.render(buffer, buffer.area);
  assert_buffer_eq(
    buffer,
    Buffer::with_lines(
      "   a b c",
      "        ",
      "   a b c",
      ...
    ));
}
```

Take a look at the src/widgets/list.rs for some move examples of simplified tests.

joshka (changes requested):

joshka:
Also +1 to implementing this idea - this is otherwise pretty jarring.

src/widgets/table.rs
```diff
@@ -229,6 +229,8 @@ pub struct Table<'a> {
     header: Option<Row<'a>>,
     /// Data to display in each row
     rows: Vec<Row<'a>>,
+    /// Always allocate the space for the selection
+    highlight_always_space: bool,
```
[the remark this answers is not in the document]
hasezoey: enum added in 6e3407cfd554ce021c3eb50e869ef6ccc35d6e44, is the naming of the enum and variants OK?

src/widgets/table.rs
```diff
@@ -291,8 +302,8 @@ impl<'a> Table<'a> {
 
     /// If set to `true`, the selection gap will always be allocated, regardless of if something is
     /// selected or not
-    pub fn highlight_set_always_space(mut self, value: bool) -> Self {
-        self.highlight_always_space = value;
+    pub fn highlight_set_selection_space(mut self, value: HighlightSpacing) -> Self {
+        self.highlight_spacing = value;
```
joshka: Enum is named fine - the method should be more builder-like (so `highlight_spacing()`?) Don't forget to update the doc comment here too.

joshka (changes requested):

src/widgets/table.rs
joshka: In the doc comment it would be good to note why this feature exists. Sort of reiterate the commit message but reword it so that it fits well with the tone of the docs - i.e. "this works like ..." rather than "this is changed because ..."

src/widgets/table.rs
```diff
@@ -286,14 +300,20 @@ impl<'a> Table<'a> {
         self
     }
 
+    /// Set which style of selection space allocation to use
+    pub fn highlight_set_selection_space(mut self, value: HighlightSpacing) -> Self {
```
joshka, line 304: You pushed just as my last comment was written :)

In case you missed it:

> In the doc comment it would be good to note why this feature exists. Sort of reiterate the commit message but reword it so that it fits well with the tone of the docs - i.e. "this works like ..." rather than "this is changed because ..."

Putting my user hat on, this comment is difficult to understand. Perhaps something like:

```suggestion
    /// When to show the highlight spacing.
    ///
    /// A highlight symbol can be displayed to the left of the row. If there is no selected row, this setting allows
    /// space to be shown or omitted. If the space is omitted then the table columns will change size when...
    pub fn highlight_spacing(mut self, value: HighlightSpacing) -> Self {
```

src/widgets/table.rs
```diff
@@ -286,14 +289,21 @@ impl<'a> Table<'a> {
...
+        self
+    }
+
     pub fn column_spacing(mut self, spacing: u16) -> Self {
         self.column_spacing = spacing;
         self
     }
 
-    fn get_columns_widths(&self, max_width: u16, has_selection: bool) -> Vec<u16> {
+    fn get_columns_widths(&self, max_width: u16, add_selection_space: bool) -> Vec<u16> {
```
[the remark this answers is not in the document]
hasezoey: done with 21a7869ccd5e3380a0cd82826f1044ecd528e352, but had to do e2a53e1d1e3fcb543f60e8a60b26b48023a96f74 before, otherwise it would not made have proper sense

tests/widgets_table.rs
hasezoey: > that just test the feature over integration tests that test the feature with a bunch of other things. (header / block / etc.). There's two reasons for this - simplicity, and reducing test failure surface.

it is true that there should probably be unit tests instead of just integration tests, but when first looking where to add tests, i didnt find any (output testing) tests in `src/widgets/table.rs`, but in `tests/widgets_table.rs`.

also this change does need to test extra things like `header`, `block` and selection, because the column allocation also affects headers, and blocks need to be tested to ensure that the calculation is not off if there are blocks

hasezoey:
i think i addressed all requests, is the current state OK?

PS: someone with write, could the "CI / coverage (pull_request)" workflow be re-run?

tests/widgets_table.rs
joshka: Sounds fine

joshka (approved):

joshka:
I'll wait on a second review on this one.

a-kenji:
Thanks for implementing this, this has been a pet peeve of mine on other `ratatui` apps.

I am not sure if allocating is a fitting word here, it confused me initially.
I also think that this is a breaking change, since the interface of a public struct is changing.

Otherwise LGTM.

joshka:
> I also think that this is a breaking change, since the interface of a public struct is changing.

All the fields on `Table` are private, and the change just adds a new method, so I think this is probably in the mostly ok bucket (except for people doing "Strange Things"(TM)). What do you think?

hasezoey:
> I am not sure if allocating is a fitting word here, it confused me initially.

any suggestions what to use instead?

joshka:
> any suggestions what to use instead?

How about something like:

`feat(table): add option to reserve space for highlight symbol`

with a commit body like:

```text
Before this option was available, selecting an item when there was no item selected
previously made the table's layout change. This enables the table width to take up the
same amount of space whether there is a selection or not.
```

hasezoey:
is the change in 146d7913ab36093dbe6b89647408c4c5ee34394a enough or should something else be changed?

joshka:
Sorry - I miscommunicated here a little. The suggested text in my previous comment was for the git commit history rather than the doc comments.

I suggest the docs should be written about the current implementation, and they don't need to cover history unless there's a strong likelihood of users needing to transition. Because this is new code, I'd avoid this.

The git commit is where we talk in terms of what the change was "This is how things changed".

joshka (approved):

src/widgets/table.rs
```diff
@@ -160,6 +160,38 @@ impl<'a> Styled for Row<'a> {
     }
 }
 
+/// This option allows to configure the behavior of the "highlight symbol" column width allocation
```
kdheepak, line 163:
```suggestion
/// This option allows the user to configure the "highlight symbol" column width spacing
```

src/widgets/table.rs
```diff
@@ -160,6 +160,38 @@ impl<'a> Styled for Row<'a> {
...
+#[derive(Debug, PartialEq, Eq, Clone, Default)]
+pub enum HighlightSpacing {
+    /// Always add spacing for the row selection symbol
+    ///
+    /// With this variant, the column for the selection symbol will always be allocated, and so the
+    /// table will never change size, regardless of if a row is selected or not
+    Always,
+    /// Only add spacing for the row selection symbol, if a row is selected
```
kdheepak, line 171:
```suggestion
    /// Only add spacing for the row selection symbol if a row is selected
```

src/widgets/table.rs
```diff
@@ -160,6 +160,38 @@ impl<'a> Styled for Row<'a> {
...
+    ///
+    /// With this variant, the column for the selection symbol will only be allocated if there is a
+    /// selection, causing the table to shift if selected / unselected
+    #[default]
+    WhenSelected,
+    /// Never add spacing selection symbol spacing, regardless of if something is selected or not
```
kdheepak, line 177:
```suggestion
    /// Never add spacing selection symbol spacing, regardless of whether something is selected or not
```

src/widgets/table.rs
```diff
@@ -160,6 +160,38 @@ impl<'a> Styled for Row<'a> {
     }
 }
 
+/// This option allows to configure the behavior of the "highlight symbol" column width allocation
+#[derive(Debug, PartialEq, Eq, Clone, Default)]
+pub enum HighlightSpacing {
+    /// Always add spacing for the row selection symbol
```
kdheepak, line 166:
```suggestion
    /// Always add spacing for the selection symbol column
```

src/widgets/table.rs
```diff
@@ -160,6 +160,38 @@ impl<'a> Styled for Row<'a> {
...
+    Always,
+    /// Only add spacing for the row selection symbol, if a row is selected
+    ///
+    /// With this variant, the column for the selection symbol will only be allocated if there is a
+    /// selection, causing the table to shift if selected / unselected
+    #[default]
+    WhenSelected,
+    /// Never add spacing selection symbol spacing, regardless of if something is selected or not
+    ///
+    /// This effectively changes that the highlight symbol will never be drawn
```
kdheepak, line 179:
```suggestion
    /// This effectively ensures that the highlight symbol will never be drawn
```

src/widgets/table.rs
```diff
@@ -286,17 +321,24 @@ impl<'a> Table<'a> {
         self
     }
 
+    /// Set when to show the highlight spacing
+    ///
+    /// see [HighlightSpacing] about which variant affects spacing in which way
```
kdheepak, line 326:
```suggestion
    /// See [HighlightSpacing] about which variant affects spacing in which way
```

kdheepak:
Thanks for the PR! This is a feature that I've wanted for a while as well!

kdheepak (approved):
I only have minor documentation related comments. The PR LGTM otherwise.

hasezoey:
just as a note, i have explicitly mentioned "row selection symbol" instead of just "selection symbol" for things like #376

hasezoey:
force-pushed to rebase onto latest master (which added `Hash` on `Table`) and to fix cargo fmt for the suggestions commit

hasezoey:
i will ask again, just to be sure:

> should i do a similar change for `List`?

orhun (approved):
All good, but let's use a commit message such as `feat(table): ...` before merge 🐻

joshka (approved):

joshka:
> @orhun are the commits correct now? i personally use that style for my projects, to quickly note which subject (file) is meant, because there could be multiple things that have the same name in different paths (example in this project `table` as in `widgets` or as in `example`)

Keeping it broad is fine for this - it's mostly for the [Changelog](https://github.com/ratatui-org/ratatui/blob/main/CHANGELOG.md)

joshka:
>  should i do a similar change for List?

Yes please. Thanks for this change.

PR #377, opened 2023-08-06
test(list): Added benchmarks

Valentin271:
I added benchmarks for the `list` widget : 

- Render
- Render half scrolled

Chose 16284 as the highest number of lines so we can see the three tests on the graph.

I tried benchmarking the `new` operation but I think it is not necessary as it only involves copying/moving.

---
Some benchmarks

_Figure 1: Render time relative to number of lines_
image
_Figure 2: Halfway scrolled render time relative to number of lines_
image
_Figure 3: Render vs halfway scrolled render_
image

The number of lines impact the performances (probably because of memory usage?) but where the list is rendered (the `offset`) doesn't seem to.

---

Relates to #137

### DIFF
--- a/benches/list.rs
+++ b/benches/list.rs
@@ -1,0 +1,73 @@
+use criterion::{criterion_group, criterion_main, BatchSize, Bencher, BenchmarkId, Criterion};
+use ratatui::{
+    buffer::Buffer,
+    layout::Rect,
+    widgets::{List, ListItem, ListState, StatefulWidget, Widget},
+};
+
+/// Benchmark for rendering a list.
+/// It only benchmarks the render with a different amount of items.
+pub fn list(c: &mut Criterion) {
+    let mut group = c.benchmark_group("list");
+
+    for line_count in [64, 2048, 16384] {
+        let lines: Vec<ListItem> = (0..line_count)
+            .map(|_| ListItem::new(fakeit::words::sentence(10)))
+            .collect();
+
+        // Render default list
+        group.bench_with_input(
+            BenchmarkId::new("render", line_count),
+            &List::new(lines.clone()),
+            render,
+        );
+
+        // Render with an offset to the middle of the list and a selected item
+        group.bench_with_input(
+            BenchmarkId::new("render_scroll_half", line_count),
+            &List::new(lines.clone()).highlight_symbol(">>"),
+            |b, list| {
+                render_stateful(
+                    b,
+                    list,
+                    ListState::default()
+                        .with_offset(line_count / 2)
+                        .with_selected(Some(line_count / 2)),
+                )
+            },
+        );
+    }
+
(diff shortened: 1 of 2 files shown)


joshka (approved):

benches/list.rs
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    group.finish();
+}
+
+/// render the list into a common size buffer
+fn render(bencher: &mut Bencher, list: &List) {
+    let mut buffer = Buffer::empty(Rect::new(0, 0, 200, 50));
+    bencher.iter(|| {
+        Widget::render(list.clone(), buffer.area, &mut buffer);
```
joshka, line 48: I wonder if `list.clone()` call could be an expensive part of this?

Valentin271: Yes it was, didn't find a solution for that originally. After more research I found [this post](https://stackoverflow.com/questions/69540182/benchmark-function-with-input-that-doesnt-implement-the-copy-trait) that dealt with this exact scenario.

To have a setup function, you have to use [`iter_batched`](https://docs.rs/criterion/0.5.1/criterion/struct.Bencher.html#method.iter_batched).
See [this doc](https://docs.rs/criterion/0.5.1/criterion/enum.BatchSize.html) for why I chose `LargeInput`. Short answer is
> `SmallInput` [...] is small enough that millions of values can be safely held in memory

and after testing, it uses less memory and runs about 10% faster.

Using `iter_batched` also runs about 20% faster, according to criterion output.

---

Here are the new performance graphs

image
image

---

I tested this on other benches and all were improved, paragraph saw an improvement of 65%!
I'm going to submit a PR next to address that.

joshka: Nice one! I thought of the paragraph issue when I saw this too, but wasn't sure how to solve it.

benches/list.rs
```diff
@@ -0,0 +1,69 @@
...
+
+/// render the list into a common size buffer
+fn render(bencher: &mut Bencher, list: &List) {
+    let mut buffer = Buffer::empty(Rect::new(0, 0, 200, 50));
+    bencher.iter_batched(
+        || list.to_owned(),
+        |bench_list| {
+            Widget::render(bench_list, buffer.area, &mut buffer);
+        },
+        BatchSize::LargeInput,
```
joshka, line 52: Can you please add a comment that helps describe why batched is necessary (here and below). It's sort of non-obvious, and would be definitely helpful to the next person writing benchmarks (or even yourself down the line). Something short like "This is necessary because ..." with a link to this PR would probably do fine.

Valentin271: Done, it's definitely not obvious.

joshka (approved):
LGTM - thanks again

kdheepak (approved):
Thanks for the PR! This is great.

PR #380, opened 2023-08-07
build(deps): upgrade crossterm to 0.27

a-kenji:
[Changelog](https://github.com/crossterm-rs/crossterm/blob/master/CHANGELOG.md#version-027)

Upgrade `crossterm` to `v0.27`.

I don't think this is a breaking change, since the `crossterm` internals aren't exposed, is that correct?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,7 +38,7 @@ rustdoc-args = ["--cfg", "docsrs"]
 [dependencies]
 bitflags = "2.3"
 cassowary = "0.3"
-crossterm = { version = "0.26", optional = true }
+crossterm = { version = "0.27", optional = true }
 indoc = "2.0"
 paste = "1.0.2"
 serde = { version = "1", optional = true, features = ["derive"] }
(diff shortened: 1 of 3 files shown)


sayanarijit:
IMO it is a breaking change because ratatui requires us to install crossterm separately for input handling. Some extensions will break too.

a-kenji:
You should be able to have `ratatui` with `0.27` and a separate `crossterm` for handling input events that is on `0.26`.

And there shouldn't be conflicting dependencies since `ratatui` is not using the `event` feature.

Can you show me an example of an extension that will break as a result of this change? 

I am happy to mark this as a breaking change if that is the consensus.

joshka:
What should we do about doc comments / readme for this? Make the change now or leave it until we release? I wonder how many people use the version number from the github readme.

sayanarijit:
Cool... I have tested with [tui-input](https://github.com/sayanarijit/tui-input). After a `cargo update`, works fine.

sayanarijit (approved):

joshka (approved):

PR #381, opened 2023-08-07
chore: implement `Hash` common traits

TieWay59:
Reorder the derive fields to be more consistent:

    Debug, Default, Clone, Copy, Eq, PartialEq, Ord, PartialOrd, Hash

Hash trait won't be impl in this PR due to rust std design. If we need hash trait for f64 related structs in the future, we should consider wrap f64 into a new type.

see: https://github.com/ratatui-org/ratatui/issues/307

### DIFF
--- a/src/backend/crossterm.rs
+++ b/src/backend/crossterm.rs
@@ -42,7 +42,7 @@ use crate::{
 /// # Ok(())
 /// # }
 /// ```
-#[derive(Debug, Default, Clone, Eq, PartialEq)]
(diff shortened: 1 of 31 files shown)


TieWay59:
> If we need a hash trait for f64-related structs in the future, we should consider wrapping f64 into a new type.

This has been done by some crates like: https://github.com/reem/rust-ordered-float

But, I don't think we need such high precision in terminal UI, I believe if precision is limited, we can always use int to represent most of the data struct values.

joshka (approved):

kdheepak (approved):

PR #382, opened 2023-08-07
chore(docs): add doc comment bump to release documentation

a-kenji:


orhun (approved):
Good catch, we can probably automate all of this with a script.

PR #388, opened 2023-08-08
feat(lazy-list) Implement LazyList widget

### LINKED ISSUE #387: Lazy list widget
## Problem
List widget is unusable with big data sets due to eager evaluation

## Solution
Use some kind of lazy loading/evaluation, like taking in a closure

## Alternatives
Extend API in a way that would make it more convenient to remove early items without changing selection (this way it would be easier to remove early elements in a list without affecting currently visible elements and selection), perhaps some sort of selection offset logic?

## Additional context
https://github.com/fdehau/tui-rs/issues/537
https://github.com/fdehau/tui-rs/issues/626

PolyMeilex:
Allows to initialize list items lazily, this helps a lot with performance in list that contain big set of data. 

closes  #387

### DIFF
--- a/src/widgets/list.rs
+++ b/src/widgets/list.rs
@@ -1,3 +1,5 @@
+use std::rc::Rc;
+
 use unicode_width::UnicodeWidthStr;
 
 use crate::{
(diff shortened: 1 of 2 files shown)


joshka:
No problems - I'll close it out for now, but feel free to re-open if you get to it.

PR #389, opened 2023-08-08
feat: expand serde attributes for `TestBuffer`

a-kenji:
This makes it possible to deserialize/serialize the whole test buffer with the `serde` feature.

### DIFF
--- a/src/backend/test.rs
+++ b/src/backend/test.rs
@@ -29,6 +29,7 @@ use crate::{
 /// # }
 /// ```
 #[derive(Debug, Clone, Eq, PartialEq, Hash)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub struct TestBackend {
     width: u16,
     buffer: Buffer,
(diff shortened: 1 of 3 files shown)


joshka:
I'm curious what you're using this for.

Do you think this needs unit / integ tests? It would help catch any serialization breaking changes if these were tested.

a-kenji:
I am not quite sure how to test the serialization without bringing in another dependency. I will look further into it.


I am in the process of prototyping a snapshot testing helper, specifically for ratatui.
Which should be capable of rendering exactly what ratatui is able to render, to make testing styles and modifiers easier and maybe even color themes.
[serde_ratatui.webm](https://github.com/ratatui-org/ratatui/assets/65275785/f94b2a5e-7b25-4a9e-ac2f-a7a5ed47a7f7)

joshka:
>I am in the process of prototyping a snapshot testing helper, specifically for ratatui.
Which should be capable of rendering exactly what ratatui is able to render, to make testing styles and modifiers easier and maybe even color themes.

Cool stuff!

joshka (approved):
No big deal on the tests for this - unless there's a simple way to do this. None of our other serializable types have tests right now.

joshka:
Needs to be rebased before merging. Can you please take a look?

joshka (approved):

PR #390, opened 2023-08-09
chore: use vhs to create demo.gif

joshka:
The bug that prevented braille rendering is fixed, so switch to VHS for
rendering the demo gif

Demo of Ratatui

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ Status](https://deps.rs/repo/github/ratatui-org/ratatui/status.svg?style=flat-sq
 [![Discord](https://img.shields.io/discord/1070692720437383208?label=discord&logo=discord&style=flat-square)](https://discord.gg/pMCEU9hNEj)
 
 <!-- See RELEASE.md for instructions on creating the demo gif --->
-![Demo of Ratatui](https://github.com/ratatui-org/ratatui/assets/24392180/93ab0e38-93e0-4ae0-a31b-91ae6c393185)
+![Demo of Ratatui](https://vhs.charm.sh/vhs-tF0QbuPbtHgUeG0sTVgFr.gif)
 
 <details>
 <summary>Table of Contents</summary>
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -3,22 +3,16 @@
 [crates.io](https://crates.io/crates/ratatui) releases are automated via [GitHub
 actions](.github/workflows/cd.yml) and triggered by pushing a tag.
 
-1. Record a new demo gif. The preferred tool for this is [ttyrec](http://0xcc.net/ttyrec/) and
-   [ttygif](https://github.com/icholy/ttygif). [Asciinema](https://asciinema.org/) handles block
-   character height poorly, [termanilizer](https://www.terminalizer.com/) takes forever to render,
-   [vhs](https://github.com/charmbracelet/vhs) handles braille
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -16,7 +16,7 @@ Status](https://deps.rs/repo/github/ratatui-org/ratatui/status.svg?style=flat-sq
 [![Discord](https://img.shields.io/discord/1070692720437383208?label=discord&logo=discord&style=flat-square)](https://discord.gg/pMCEU9hNEj)
 
 <!-- See RELEASE.md for instructions on creating the demo gif --->
-![Demo of Ratatui](https://github.com/ratatui-org/ratatui/assets/24392180/93ab0e38-93e0-4ae0-a31b-91ae6c393185)
+![Demo of Ratatui](https://vhs.charm.sh/vhs-tF0QbuPbtHgUeG0sTVgFr.gif)
```
hasezoey, line 19: is this a external link (and not a link inside the repo) for the readme on something like crates.io / docs.rs?

joshka: Yep - see https://github.com/charmbracelet/vhs#publish-tapes

The benefit of this is that it's easy to automate. The downside is that it's someone else's service.
I'd like to avoid having gifs in the repo as they make the repo take longer to clone (and grow without limit over time).

Previously @orhun saw that a vhs link wasn't visible, so we uploaded this to somewhere on github.
The implication of the comment suggests that you might be aware of something that might have caused this?

hasezoey: >  The implication of the comment suggests that you might be aware of something that might have caused this?

no i dont

---

nevermind, i misread on where the other gifs are stored, i assumed that they were in the repo itself, instead also on the vhs server, because i only looked at [this line](https://github.com/ratatui-org/ratatui/pull/390/files#diff-49aaa2819e35a856818ecec8c9fa7e1c79ad028d3f44bd749736353cfb51bac9R19)

joshka: Ah yes - that's so we can update all the image links with a single command:

https://github.com/ratatui-org/ratatui/blob/19470eaf411ed4ca25372769c59ab78d305589f1/examples/README.md?plain=1#L186-L201

README.md
```diff
@@ -16,7 +16,7 @@ Status](https://deps.rs/repo/github/ratatui-org/ratatui/status.svg?style=flat-sq
 [![Discord](https://img.shields.io/discord/1070692720437383208?label=discord&logo=discord&style=flat-square)](https://discord.gg/pMCEU9hNEj)
 
 <!-- See RELEASE.md for instructions on creating the demo gif --->
-![Demo of Ratatui](https://github.com/ratatui-org/ratatui/assets/24392180/93ab0e38-93e0-4ae0-a31b-91ae6c393185)
```
orhun, line 19: I would rely on GitHub for hosting major files like this. I have projects from 4+ years ago that still has no broken links/images since everything is hosted on GitHub. On the other hand, we don't know if `vhs.charm.sh` will go offline tomorrow or not 🐻

kdheepak: It would be nice if we can upload it to a comment on a gist on GitHub programmatically. That way we don’t bloat the git history of this project. Or a separate repo just for demos also works.

kdheepak (approved):

README.md
kdheepak: I made a issue to track this request: https://github.com/ratatui-org/ratatui/issues/401

orhun (approved):

PR #391, opened 2023-08-09
Document all features in one place

hasezoey:
This PR extends the `lib.rs` `Features` section to list all the features that are available (publicly)

also changes `time` to `dep:time` for `widget-calendar` to be consistent with other features (like `serde`)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ rust-version = "1.67.0"
 [features]
 default = ["crossterm"]
 all-widgets = ["widget-calendar"]
-widget-calendar = ["time"]
+widget-calendar = ["dep:time"]
 macros = []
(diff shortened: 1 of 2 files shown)


joshka (commented):
I've gone a bit deeper on the docs as part of the changes to lib.rs in:
https://github.com/ratatui-org/ratatui/pull/280/files#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759

joshka:
I extracted the docs for features from #280 and updated your commit here so we can get this merged before the next release. (Also rebased on main to pick up some other changes)

kdheepak (approved):

joshka:
Thanks for the PR @hasezoey - We merged the contents of relevant part of #280 instead (with some small edits based on your commit).

Issue #367, opened 2023-08-05
Bug: Layout leaves gaps


joshka:
## Description
Using either of the following results in layouts that have gaps between areas for some terminal widths. It doesn't seem possible to allocate the full terminal width to the contents without gaps.

```rust
Layout::default()
        .direction(Direction::Horizontal)
        .constraints(vec![Constraint::Ratio(1, 8); 8])
        .split(frame.size());
```
or
```rust
Layout::default()
        .direction(Direction::Horizontal)
        .constraints(vec![Constraint::Percentage(13); 8])
        .split(frame.size());
```

## To Reproduce
Grab PR #345 for code that hits this and run:
```shell
cargo run --example colors
```

## Expected behavior
Layout split should avoid gaps

## Screenshots
<img width="794" alt="image" src="https://github.com/ratatui-org/ratatui/assets/381361/64a247e3-ab03-4328-8e33-3653ad69c489">

## Environment
Ratatui on main - terminal / font are not relevant

## Additional context
Tried various percentages  and ratios

joshka:
See #303 for the code that triggers this

joshka:
in https://github.com/ratatui-org/ratatui/blob/344efe3ed2a11644a960193fd7f847773bdb3c41/src/layout.rs#L482-L490
I think this can probably be fixed using (just testing for any issues this causes):

```rust
        .map(|element| {
            let x = *changes.get(&element.x).unwrap_or(&0.0);
            let y = *changes.get(&element.y).unwrap_or(&0.0);
            let width = *changes.get(&element.width).unwrap_or(&0.0);
            let height = *changes.get(&element.height).unwrap_or(&0.0);
            Rect {
                x: x.round() as u16,
                y: y.round() as u16,
                width: ((x + width).round() - x.round()) as u16,
                height: ((y + height).round() - y.round()) as u16,
            }
        })
```
Which rounds the left / right / top / bottom to the nearest int rather than the width. A minimal test that shows the problem:
```rust
let layout = Layout::default().constraints(vec![
            Constraint::Percentage(50),
            Constraint::Percentage(50),
            Constraint::Min(0),
        ]);
        assert_eq!(
            layout.split(Rect::new(0, 0, 1, 1))[..],
            [
                Rect::new(0, 0, 1, 0), // should be height = 1
                Rect::new(0, 1, 1, 0),
                Rect::new(0, 1, 1, 0)
            ]
        );
```

PR #393, opened 2023-08-11
docs(examples): show layout constraints

joshka:
Shows the way that layout constraints interact visually

example

# Review Notes

I'm not sure if this is "good" as an example for users to get value out of the code, as it isn't super simple code. But it covers a reasonable combination of all constraints. It doesn't cover layout margins, and situations where constraints cause gaps (e.g. when Percentage results in a size < full cell width).

Inspired by the following issues:
- https://github.com/ratatui-org/ratatui/issues/354
- https://github.com/ratatui-org/ratatui/issues/367
- https://github.com/ratatui-org/ratatui/issues/374
- https://github.com/ratatui-org/ratatui/pull/385

### DIFF
--- a/examples/layout.rs
+++ b/examples/layout.rs
@@ -5,7 +5,8 @@ use crossterm::{
     execute,
     terminal::{disable_raw_mode, enable_raw_mode, EnterAlternateScreen, LeaveAlternateScreen},
 };
-use ratatui::{prelude::*, widgets::*};
+use itertools::Itertools;
+use ratatui::{layout::Constraint::*, prelude::*, widgets::*};
 
 fn main() -> Result<(), Box<dyn Error>> {
     // setup terminal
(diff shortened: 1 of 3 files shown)


kdheepak:
The screen shot looks great! For me personally something that would also be useful is describing in the documentation the math of the linear program for solving the constraints.

examples/layout.rs
```diff
@@ -47,50 +48,174 @@ fn run_app<B: Backend>(terminal: &mut Terminal<B>) -> io::Result<()> {
...
+    let inner = block.inner(area);
+    frame.render_widget(block, area);
+    let layout = Layout::default()
+        .direction(Direction::Vertical)
+        .constraints(vec![Length(1); constraints.len() + 1])
+        .split(inner);
+    for (i, (a, b)) in constraints.iter().enumerate() {
+        render_single_example(frame, layout[i], vec![*a, *b, Min(0)]);
+    }
+    frame.render_widget(Paragraph::new("123456789012"), layout[6]);
```
hasezoey, line 192: maybe note why there is a extra paragraph with 12 characters?

joshka:
> The screen shot looks great! For me personally something that would also be useful is describing in the documentation the math of the linear program for solving the constraints.

I don't know whether that would necessarily be worth your time, especially since we get pretty close to it with taffy for the cost of significantly less code and the benefit of fixing some annoying gaps. I'm a big fan of simple / obvious code over complex code when possible.

examples/layout.rs
joshka: Will do (it's there to be able to visually measure the columns in the example above it)

kdheepak:
Here's a ChatGPT generated solution for the mathematical program:

1. **Definitions**:
   - Let $x_i$, $y_i$, $w_i$, $h_i$ represent the x-coordinate, y-coordinate, width, and height of the $i^{th}$ element, respectively.
   - The destination area is defined by $[area_l, area_t, area_r, area_b]$.
   - The direction of layout is either Horizontal or Vertical.

2. **Basic Constraints**:
   For each element $i$:
   - $w_i \geq 0$ 
   - $h_i \geq 0$ 
   - $x_i \geq area\_l$ 
   - $y_i \geq area\_t$ 
   - $x_i + w_i \leq area\_r$ 
   - $y_i + h_i \leq area\_b$

3. **For the first element**:
   If `direction = Horizontal`:
   - $x_1 = area\_l$
   
   If `direction = Vertical`:
   - $y_1 = area\_t$

4. **If the layout expands to fill**:
   For the last element $n$:
   If `direction = Horizontal`:
   - $x_n + w_n = area\_r$
   
   If `direction = Vertical`:
   - $y_n + h_n = area\_b$

5. **Element adjacency**:
   If `direction = Horizontal`:
   For every consecutive pair $(i, i+1)$:
   - $x_i + w_i = x_{i+1}$
   
   If `direction = Vertical`:
   For every consecutive pair $(i, i+1)$:
   - $y_i + h_i = y_{i+1}$

6. **Special size constraints for each element $i$ based on the layout**:
   If `direction = Horizontal`:
   - $y_i = area\_t$
   - $h_i = area\_b - area\_t$
   
   If `direction = Vertical`:
   - $x_i = area\_l$
   - $w_i = area\_r - dest\_l$
   
   Depending on the type of constraint for the element:
   - For `Length`: $dimension_i = value$ where dimension is width for horizontal and height for vertical.
   - For `Percentage`: $dimension_i = percentage \times dimension$
   - For `Ratio`: $dimension_i = (dimension \times numerator) / denominator$
   - For `Min`: $dimension_i \geq value$
   - For `Max`: $dimension_i \leq value$

When I did this in Julia, I added one constraint. I added a weak constraint saying all widths should be the same for each block. Because it is a weak constraint, it is the last to take precedence. That seemed to give me results that I felt were more intuitive or what a user might naturally expect.

kdheepak:
This talk on YouTube is a great resource for understanding why Cassowary is typically used for this kind of problem, and gives some insight into the math above: https://www.youtube.com/watch?v=gxfyb3ipUFg

kdheepak (approved):

joshka:
Merging this to help make it easier to visually look at various layout changes.

PR #394, opened 2023-08-11
feat(list): add option to always allocate the "selection" column width

hasezoey:
Before this option was available, selecting a item in a list when nothing was selected previously made the row layout change (the same applies to unselecting) by adding the width of the "highlight symbol" in the front of the list, this option allows to configure this behavior.

re #375

TL;DR: the same as #375, but for `widgets::List`

implementing this for list was easier than `Table` because it already had basically everything in-place

### DIFF
--- a/src/widgets/list.rs
+++ b/src/widgets/list.rs
@@ -151,7 +151,7 @@ impl<'a> List<'a> {
 
     /// Set when to show the highlight spacing
     ///
-    /// See [HighlightSpacing] about which variant affects spacing in which way
+    /// See [`HighlightSpacing`] about which variant affects spacing in which way
     pub fn highlight_spacing(mut self, value: HighlightSpacing) -> Self {
         self.highlight_spacing = value;
         self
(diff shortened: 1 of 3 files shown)


src/widgets/list.rs
```diff
@@ -146,6 +149,14 @@ impl<'a> List<'a> {
         self
     }
 
+    /// Set when to show the highlight spacing
+    ///
+    /// See [HighlightSpacing] about which variant affects spacing in which way
```
joshka, line 154: nit:
```suggestion
    /// See [`HighlightSpacing`] about which variant affects spacing in which way
```

joshka (approved):
LGTM

src/widgets/list.rs
hasezoey: https://github.com/ratatui-org/ratatui/blob/f63ac72305f80062727d81996f9bdb523e666099/src/widgets/table.rs#L327

hasezoey: done for both table and list in 20e988ac1ef462b96e40bfc05995cb251cd34c44

kdheepak (approved):

Issue #370, opened 2023-08-05
Table: Add option to make `expand_to_fill` configurable for `Table::widths`


hasezoey:
## Problem
Currently on widget `Table` (`widgets/table.rs`), all `.widths` are thrown into a `Layout` (like the user would do for a block), but with the added (currently non-configurable) `expand_to_fill` set to `false`, making all Constraints other than Percentage use only the minimal amount

## Solution
Add option to have `widths` `expand_to_fill` configurable, maybe something like `widths_expand_to_fill()`

## Alternatives
also a alternative would be to just remove the `expand_to_fill` call

## Additional context
Currently if any other constraint than `Percentage` is used, the columns will always use the minimal amount of width, example:

```rust
// testing backend for consistent size and printing to console
let backend = TestBackend::new(116, 27);

// widths as applied to the table
.widths(&[
    Constraint::Length(4),
    Constraint::Max(30),
    Constraint::Min(10),
    Constraint::Max(4),
]);

// debug print in table.rs get_columns_widths
println!("max_width {}, chunks: {:#?}", max_width, chunks.iter().map(|v| v.width).collect::<Vec<u16>>());
```

would result in:

```
max_width 114, chunks: [
    4,
    1, // spacer
    30,
    1, // spacer
    10,
    1, // spacer
    4,
]
```

whereas changing `expand_to_fill` to `true` would result in:

```
max_width 114, chunks: [
    4,
    1, // spacer
    30,
    1, // spacer
    73,
    1, // spacer
    4,
]
```

hasezoey:
i dont know if this is actually a proper solution (exposing `expand_to_fill`), because if there are multiple elements that could grow, only one gets grown (as per cassowary documentation)

> Due to the nature of the algorithm, "in-between" solutions, although just as valid, are not picked.

i may also just completely misunderstand this though

hasezoey:
this may be (at least partially) fixed by #395

joshka:
I've been thinking a bit about this specific issue also over the last couple of days from the perspective of `Layout` though. I suspect that the rationale for keeping the method private was that there's a few different ways that making things fit right could be achieved, and it's difficult to make that stable.

I think this idea is fine, but it's worth pausing for a little while to let it bubble together with the other things we're playing with around layout. Do you have a specific use case for this, or is it more about removing a limitation?

hasezoey:
> Do you have a specific use case for this, or is it more about removing a limitation?

[...]

PR #395, opened 2023-08-12
Add weak constraints to make rects closer to each other in size 

kdheepak:
This PR introduces a `SegmentSize` `enum` that will allow adding the following constraints optionally:

```rust

        for i in 0..elements.len() {
            for j in (i + 1)..elements.len() {
                solver.add_constraint(elements[j].size() | EQ(WEAK) | (elements[i].size()))?;
            }
        }

```

This makes chunks similar in size to each other if other constraints are satisfied.

This PR also changes the implementation of `Max` and `Min` by adding a `MEDIUM` equality constraint instead of a `WEAK` equality constraint. 

```diff
     Constraint::Max(m) => {
         solver.add_constraints(&[                           
             element.size() | LE(STRONG) | f64::from(m),     
-            element.size() | EQ(WEAK) | f64::from(m),
+            element.size() | EQ(MEDIUM) | f64::from(m),
         ])?;                                                
     }                                                       
     Constraint::Min(m) => {                                 
         solver.add_constraints(&[                           
             element.size() | GE(STRONG) | f64::from(m),     
-            element.size() | EQ(WEAK) | f64::from(m),
+            element.size() | EQ(MEDIUM) | f64::from(m),
         ])?;                                                
     }                                                       
```

It should result in the same layouts for us though, because there were no other `MEDIUM` constraints used.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,6 +40,7 @@ bitflags = "2.3"
 cassowary = "0.3"
 crossterm = { version = "0.27", optional = true }
 indoc = "2.0"
+itertools = "0.11"
 paste = "1.0.2"
 serde = { version = "1", optional = true, features = ["derive"] }
 termion = { version = "2.0", optional = true }
@@ -56,7 +57,6 @@ cargo-husky = { version = "1.5.0", default-features = false, features = [
 ] }
 criterion = { version = "0.5", features = ["html_reports"] }
 fakeit = "1.1"
-itertools = "0.10"
 rand = "0.8"
 pretty_assertions = "1.4.0"
 
--- a/src/layout.rs
+++ b/src/layout.rs
@@ -6,10 +6,11 @@ use std::{
 };
 
 use cassowary::{
-    strength::{REQUIRED, STRONG, WEAK},
+    strength::{MEDIUM, REQUIRED, STRONG, WEAK},
     AddConstraintError, Expression, Solver, Variable,
     WeightedRelation::{EQ, GE, LE},
 };
+use itertools::Itertools;
 
 #[derive(Debug, Default, Clone, Copy, Eq, PartialEq, Hash)]
 pub enum Corner {
@@ -232,6 +233,14 @@ impl Rect {
     }
 }
 
+#[derive(Debug, Default, Clone, Eq, PartialEq, Hash)]
+pub enum SegmentSize {
+    EvenDistribution,
+    #[default]
+    LastTakesRemainder,
+    None,
+}
+
 /// A layout is a set of constraints that can be applied to a given area to split it into smaller
 /// ones.
 ///
@@ -281,9 +290,8 @@ pub struct Layout {
     direction: Direction,
     margin: Margin,
     constraints: Vec<Constraint>,
-    /// Whether the last chunk of the computed layout should be expanded to fill the available
-    /// space.
-    expand_to_fill: bool,
+    /// option for segment size preferences
+    segment_size: SegmentSize,
 }
 
 impl Default for Layout {
@@ -298,7 +306,7 @@ impl Layout {
     /// - direction: [Direction::Vertical]
     /// - margin: 0, 0
     /// - constraints: empty
-    /// - expand_to_fill: true
+    /// - segment_size: SegmentSize::LastTakesRemainder
     pub const fn new() -> Layout {
         Layout {
             direction: Direction::Vertical,
@@ -307,7 +315,7 @@ impl Layout {
                 vertical: 0,
             },
             constraints: Vec::new(),
-            expand_to_fill: true,
+            segment_size: SegmentSize::LastTakesRemainder,
         }
     }
 
@@ -419,10 +427,9 @@ impl Layout {
         self
     }
 
-    /// Builder method to set whether the last chunk of the computed layout should be expanded to
-    /// fill the available space.
-    pub(crate) const fn expand_to_fill(mut self, expand_to_fill: bool) -> Layout {
-        self.expand_to_fill = expand_to_fill;
+    /// Builder method to set whether chunks should be of equal size.
+    pub(crate) const fn segment_size(mut self, segment_size: SegmentSize) -> Layout {
+        self.segment_size = segment_size;
         self
     }
 
@@ -524,7 +531,7 @@ fn try_split(area: Rect, layout: &Layout) -> Result<Rc<[Rect]>, AddConstraintErr
         solver.add_constraint(first.start | EQ(REQUIRED) | area_start)?;
     }
     // ensure the last element touches the right/bottom edge of the area
-    if layout.expand_to_fill {
+    if layout.segment_size != SegmentSize::None {
         if let Some(last) = elements.last() {
             solver.add_constraint(last.end | EQ(REQUIRED) | area_end)?;
         }
@@ -547,15 +554,21 @@ fn try_split(area: Rect, layout: &Layout) -> Result<Rc<[Rect]>, AddConstraintErr
             Constraint::Max(m) => {
                 solver.add_constraints(&[
                     element.size() | LE(STRONG) | f64::from(m),
-                    element.size() | EQ(WEAK) | f64::from(m),
+                    element.size() | EQ(MEDIUM) | f64::from(m),
                 ])?;
             }
             Constraint::Min(m) => {
                 solver.add_constraints(&[
                     element.size() | GE(STRONG) | f64::from(m),
-                    element.size() | EQ(WEAK) | f64::from(m),
+                    element.size() | EQ(MEDIUM) | f64::from(m),
                 ])?;
             }
+        }
+    }
+    // prefer equal chunks if other constraints are all satisfied
+    if layout.segment_size == SegmentSize::EvenDistribution {
+        for (left, right) in elements.iter().tuple_combinations() {
+            solver.add_constraint(left.size() | EQ(WEAK) | right.size())?;
         }
     }
 
@@ -600,7 +613,97 @@ fn try_split(area: Rect, layout: &Layout) -> Result<Rc<[Rect]>, AddConstraintErr
 
 #[cfg(test)]
 mod tests {
-    use super::*;
+    use super::{SegmentSize::*, *};
+    use crate::prelude::Constraint::*;
+
+    fn get_x_width_with_segment_size(
+        segment_size: SegmentSize,
+        constraints: Vec<Constraint>,
+        target: Rect,
+    ) -> Vec<(u16, u16)> {
+        let layout = Layout::default()
+            .direction(Direction::Horizontal)
+            .constraints(constraints)
+            .segment_size(segment_size);
+        let chunks = layout.split(target);
+        chunks.iter().map(|r| (r.x, r.width)).collect()
+    }
+
+    #[test]
+    fn test_split_equally_in_underspecified_case() {
+        let target = Rect::new(100, 200, 10, 10);
+        assert_eq!(
+            get_x_width_with_segment_size(LastTakesRemainder, vec![Min(2), Min(2), Min(0)], target),
+            [(100, 2), (102, 2), (104, 6)]
+        );
+        assert_eq!(
+            get_x_width_with_segment_size(EvenDistribution, vec![Min(2), Min(2), Min(0)], target),
+            [(100, 3), (103, 4), (107, 3)]
+        );
+    }
+
+    #[test]
+    fn test_split_equally_in_overconstrained_case_for_min() {
+        let target = Rect::new(100, 200, 100, 10);
+        assert_eq!(
+            get_x_width_with_segment_size(
+                LastTakesRemainder,
+                vec![Percentage(50), Min(10), Percentage(50)],
+                target
+            ),
+            [(100, 50), (150, 10), (160, 40)]
+        );
+        assert_eq!(
+            get_x_width_with_segment_size(
+                EvenDistribution,
+                vec![Percentage(50), Min(10), Percentage(50)],
+                target
+            ),
+            [(100, 45), (145, 10), (155, 45)]
+        );
+    }
+
+    #[test]
+    fn test_split_equally_in_overconstrained_case_for_max() {
+        let target = Rect::new(100, 200, 100, 10);
+        assert_eq!(
+            get_x_width_with_segment_size(
+                LastTakesRemainder,
+                vec![Percentage(30), Max(10), Percentage(30)],
+                target
+            ),
+            [(100, 30), (130, 10), (140, 60)]
+        );
+        assert_eq!(
+            get_x_width_with_segment_size(
+                EvenDistribution,
+                vec![Percentage(30), Max(10), Percentage(30)],
+                target
+            ),
+            [(100, 45), (145, 10), (155, 45)]
+        );
+    }
+
+    #[test]
+    fn test_split_equally_in_overconstrained_case_for_length() {
+        let target = Rect::new(100, 200, 100, 10);
+        assert_eq!(
+            get_x_width_with_segment_size(
+                LastTakesRemainder,
+                vec![Percentage(50), Length(10), Percentage(50)],
+                target
+            ),
+            [(100, 50), (150, 10), (160, 40)]
+        );
+        assert_eq!(
+            get_x_width_with_segment_size(
+                EvenDistribution,
+                vec![Percentage(50), Length(10), Percentage(50)],
+                target
+            ),
+            [(100, 45), (145, 10), (155, 45)]
+        );
+    }
 
     #[test]
     fn test_rect_size_truncation() {
@@ -705,7 +808,7 @@ mod tests {
         const _DEFAULT_LAYOUT: Layout = Layout::new()
             .direction(Direction::Horizontal)
             .margin(1)
-            .expand_to_fill(false);
+            .segment_size(SegmentSize::LastTakesRemainder);
         const _HORIZONTAL_LAYOUT: Layout = Layout::new().horizontal_margin(1);
         const _VERTICAL_LAYOUT: Layout = Layout::new().vertical_margin(1);
     }
--- a/src/widgets/table.rs
+++ b/src/widgets/table.rs
@@ -2,7 +2,7 @@ use unicode_width::UnicodeWidthStr;
 
 use crate::{
     buffer::Buffer,
-    layout::{Alignment, Constraint, Direction, Layout, Rect},
+    layout::{Alignment, Constraint, Direction, Layout, Rect, SegmentSize},
     style::{Style, Styled},
     text::Text,
     widgets::{Block, StatefulWidget, Widget},
@@ -350,7 +350,7 @@ impl<'a> Table<'a> {
         let chunks = Layout::default()
             .direction(Direction::Horizontal)
             .constraints(constraints)
-            .expand_to_fill(false)
+            .segment_size(SegmentSize::None)
             .split(Rect {
                 x: 0,
                 y: 0,


src/layout.rs
```diff
@@ -559,6 +574,46 @@ impl Rect {
...
+        let chunks = Layout::default()
+            .direction(Direction::Vertical)
+            .constraints([Constraint::Min(2), Constraint::Min(2), Constraint::Min(0)].as_ref())
+            .split(target);
+
+        assert_eq!(target.height, chunks.iter().map(|r| r.height).sum::<u16>());
+        assert_eq!(chunks[0].height, 3);
+        assert_eq!(chunks[1].height, 3);
+        assert_eq!(chunks[2].height, 4);
+    }
```
joshka, line 615: What happens when this fails on a CI server somewhere because of some unrelated change (or due to something non-deterministic), meaning the failure is difficult to repro locally?

The first check failing here will prevent the 2nd to 4th checks executing, which contain the information you'd need to understand how to fix the test  A better test for this would show how those numbers change for various split widths:

```suggestion
    fn test_horizontal_split_equally_in_underspecified_case() {
        use Constraint::*;
        fn split_widths(constraints: &Vec<Constraint>, target_width: u16) -> Vec<u16> {
            let target = Rect::new(2, 2, target_width, 10);
            Layout::default()
                .direction(Direction::Horizontal)
                .constraints(constraints.as_ref())
                .split(target)
                .iter()
                .map(|r| r.width)
                .collect_vec()
        }

        let constraints = vec![Min(2), Min(2), Min(0)];
        assert_eq!(split_widths(&constraints, 0), vec![0, 0, 0]);
        assert_eq!(split_widths(&constraints, 1), vec![1, 0, 0]);
        assert_eq!(split_widths(&constraints, 2), vec![2, 0, 0]);
        assert_eq!(split_widths(&constraints, 3), vec![2, 1, 0]);
        assert_eq!(split_widths(&constraints, 4), vec![2, 2, 0]);
        assert_eq!(split_widths(&constraints, 5), vec![2, 2, 1]);
        assert_eq!(split_widths(&constraints, 6), vec![2, 2, 2]);
        assert_eq!(split_widths(&constraints, 7), vec![2, 2, 3]);
        assert_eq!(split_widths(&constraints, 8), vec![2, 2, 4]);
        assert_eq!(split_widths(&constraints, 9), vec![2, 2, 5]);
        assert_eq!(split_widths(&constraints, 10), vec![2, 2, 6]);
    }
```
Those are the numbers before the change to the split function is applied. Normally I much prefer to have tests be  linear and simpler than this one, but I think a good reason to find an exception is where the test communicates the relationship between the outputs of similar inputs that would be otherwise difficult to understand in isolation.

What this test doesn't communicate / test is where the edge cases on the new code take effect.

Oddly, I can't actually get this to succeed with the new numbers as this now fails non-deterministically. Is the code intended to be deterministic?

E.g. here are the results of `cargo test` run several times:
```
thread 'layout::tests::test_horizontal_split_equally_in_underspecified_case' panicked at 'assertion failed: `(left == right)`
  left: `[1, 1, 1]`,
 right: `[1, 1, 0]`', src/layout.rs:596:9
```
```
thread 'layout::tests::test_horizontal_split_equally_in_underspecified_case' panicked at 'assertion failed: `(left == right)`
  left: `[1, 0, 0]`,
 right: `[1, 1, 0]`', src/layout.rs:596:9
```
```
thread 'layout::tests::test_horizontal_split_equally_in_underspecified_case' panicked at 'assertion failed: `(left == right)`
  left: `[0, 0, 1]`,
 right: `[0, 0, 0]`', src/layout.rs:595:9
```
```
thread 'layout::tests::test_horizontal_split_equally_in_underspecified_case' panicked at 'assertion failed: `(left == right)`
  left: `[2, 2, 0]`,
 right: `[1, 1, 0]`', src/layout.rs:598:9
```

src/layout.rs
```diff
@@ -318,6 +318,21 @@ fn split(area: Rect, layout: &Layout) -> Rc<[Rect]> {
             });
         }
     }
+    for (i, c1) in elements.iter().enumerate() {
```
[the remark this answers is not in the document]
kdheepak: I actually like keeping all the constraints in the same function. What do you think about me adding comments instead?

src/layout.rs
kdheepak: This is a can of worms if it is non-deterministic. Maybe we should only have tests that are deterministic?

joshka: Yes, we should only have deterministic tests, but so too behavior.

Ideally we need to find a way to push the constraints to choose more deterministic values for constraints - perhaps by being more granular with the weighting of the importance of each constraint? I think this probably becomes complex fast though, but the worms are there already I think.

src/layout.rs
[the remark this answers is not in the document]
joshka: My general rationale for adding a function instead of just commented sections is that everything you need to understand in a particular block is limited to the scope of the function. A commented section could be affected by any change above, and any change within can effect any code below, which means you have to understand the full function to maintain it. In any particular part of the lib, that's not a big deal, but when large functions dominate, every maintenance task tends to become larger as the time spent understanding is magnified.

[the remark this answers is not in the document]
kdheepak: Fair. In that case I think I’d like to move all the different constraints into individual functions too then.

src/layout.rs
```diff
@@ -41,7 +41,7 @@ pub enum Constraint {
     /// assert_eq!(5, Constraint::Percentage(50).apply(11));
     /// ```
     Percentage(u16),
-    /// Apply a ratio  
+    /// Apply a ratio
```
hasezoey, line 44: i dont know if this change was automatic, but removing the double space at the end will make the lines as if it were a single line, either re-add the double space, or add a empty line inbetween

joshka: Let's use an empty line rather than invisible characters that affect things. The rust-doc convention is that the first line of a doc comment is a single sentence paragraph, not multiple lines with a line break.

hasezoey (changes requested):
i know this a bit unrelated, but this change causes some side-effects, please test against #404 with this change applied, see https://github.com/ratatui-org/ratatui/issues/370#issuecomment-1678722149 for more context

joshka:
> i know this a bit unrelated, but this change causes some side-effects, please test against #404 with this change applied, see [#370 (comment)](https://github.com/ratatui-org/ratatui/issues/370#issuecomment-1678722149) for more context

The main blocker for this PR right now is that it results in non-deterministic splits more often, which is makes it difficult to reason about how space will be allocated for users.

kdheepak:
I did some testing with this and wrote a integer optimization program (in Julia) to simulate "ideal" behavior to compare. I think we are never going to be able to guarantee that there won't be gaps with cassowary. 

For example, take a constraints `Percent(33), Percent(33), Percent(33)` in a `Horizontal` layout of `area` whose width is 100. The "ideal" result is going to be `33.3333`, `33.3333` and `33.3333` for each `Rect`, and rounded you are going to get `33`, `33`, `33` which will leave a gap of `1`.

I implemented [this using integer variables in Julia](https://gist.github.com/kdheepak/afc1ed987c19d67d738661559c2cc6b5) and I get a solution that we would expect, i.e. `33`, `34` and `33`.

```julia
    area = Rect(0, 0, 100, 1)
    constraints = [Percent(33), Percent(33), Percent(33)]
    r1, r2, r3 = split(Horizontal(constraints), area; kwargs...)
    @test r1.width == 33
    @test r2.width == 34
    @test r3.width == 33
```

image

Unfortunately, this behavior is not guaranteed with cassowary because cassowary uses floating point values.

kdheepak:
Adding the weak constraints to make the widths (or heights) similar did have the intended effect (imo). Take this test for example:

```julia
    area = Rect(0, 0, 100, 1)
    constraints = [Percent(50), HardMin(10), Percent(50)]
    r1, r2, r3 = split(Horizontal(constraints), area; kwargs...)
    @test r1.width == 45
    @test r2.width == 10
    @test r3.width == 45
```

Without the weak constraints, I was getting results like this:

```julia
    area = Rect(0, 0, 100, 1)
    constraints = [Percent(50), HardMin(10), Percent(50)]
    r1, r2, r3 = split(Horizontal(constraints), area; kwargs...)
    @test r1.width == 50
    @test r2.width == 10
    @test r3.width == 40
```

joshka:
>I think we are never going to be able to guarantee that there won't be gaps with cassowary.

The approach that makes this round without gaps is to use the left/right position rather than the widths to work out how to round.
Thus instead of thinking of [33.33, 33.33, 33.33] => [33, 33, 33], think instead of [0.00, 33.33, 66.66, 100.00] => [0, 33, 67, 100]

This works pretty nicely as far as I can tell and is my intended change as soon as the refactor in #405 lands.

kdheepak:
Interesting idea! I think there are some corner cases there as well (I'll have to test to come up with examples), but maybe fewer issues than what we have at the moment, so it should be good!

joshka:
> Interesting idea! I think there are some corner cases there as well (I'll have to test to come up with examples), but maybe fewer issues than what we have at the moment, so it should be good!

Perhaps something that evaluates to (positions): `[0.0, 0.51, 1.49, 2]` => `[0, 1, 1, 2]`. The widths would be `[0.51, 0.98, 0.51]`, but become `[1, 0, 1]` after rounding.

src/layout.rs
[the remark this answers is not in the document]
kdheepak: With the recent refactors, I'm assuming it is okay to have the equations in the same function.

src/layout.rs
kdheepak: I ran it multiple times locally and it doesn't appear to be non-deterministic for me, although I'm not sure why since I thought the constraints were essentially the same as before.

src/layout.rs
kdheepak: This has been resolved in another PR.

hasezoey (commented):
code looks good to me, though it would probably be great to also have some tests for `Max` and `Length` to ensure they are not somehow heigher / lower than what they should be

kdheepak:
I added tests for `Max` and `Length` too. Open to better suggestions for the test for `Max`.

src/layout.rs
```diff
@@ -591,6 +608,200 @@ fn try_split(area: Rect, layout: &Layout) -> Result<Rc<[Rect]>, AddConstraintErr
...
+                .as_ref(),
+            )
+            .prefer_equal_chunks(true)
+            .split(target);
+
+        assert_eq!(target.width, chunks.iter().map(|r| r.width).sum::<u16>());
+        assert_eq!(chunks[0].width, 33);
+        assert_eq!(chunks[1].width, 34);
+        assert_eq!(chunks[2].width, 33);
+    }
```
kdheepak, line 758: I don't understand why this test is behaving this way. This seems like a bug to me.

kdheepak: Fixed in https://github.com/ratatui-org/ratatui/pull/395/commits/0b41dac3944df4c052ab872385d6f2bf926d3dd1

This commit changes the implementation of `Max` and `Min` by adding a `MEDIUM` constraint instead of a `WEAK` constraint. It should result in the same layouts for us though, because there were no other `MEDIUM` constraints used.

joshka: Good fix!

joshka (changes requested):

src/layout.rs
```diff
@@ -308,6 +310,7 @@ impl Layout {
             },
             constraints: Vec::new(),
             expand_to_fill: true,
+            prefer_equal_chunks: false,
```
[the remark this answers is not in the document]
kdheepak: Addressed in https://github.com/ratatui-org/ratatui/pull/395/commits/17790b7fbd2a3f9252c0f8b6b63f4560fd904920. Let me know if you have better suggestions for naming.

src/layout.rs
```diff
@@ -547,17 +556,25 @@ fn try_split(area: Rect, layout: &Layout) -> Result<Rc<[Rect]>, AddConstraintErr
...
                     element.size() | GE(STRONG) | f64::from(m),
-                    element.size() | EQ(WEAK) | f64::from(m),
+                    element.size() | EQ(MEDIUM) | f64::from(m),
                 ])?;
             }
         }
     }
+    // prefer equal chunks if other constraints are all satisfied
+    if layout.prefer_equal_chunks {
+        for i in 0..elements.len() {
```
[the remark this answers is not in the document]
kdheepak: Done in https://github.com/ratatui-org/ratatui/pull/395/commits/17790b7fbd2a3f9252c0f8b6b63f4560fd904920

src/layout.rs
```diff
@@ -591,6 +608,200 @@ fn try_split(area: Rect, layout: &Layout) -> Result<Rc<[Rect]>, AddConstraintErr
 mod tests {
     use super::*;
 
+    #[test]
+    fn test_split_equally_in_underspecified_case() {
```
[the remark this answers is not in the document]
kdheepak: Good points, I meant to do simplifications to the tests myself but didn't get to it in time.

src/layout.rs
```diff
@@ -591,6 +608,200 @@ fn try_split(area: Rect, layout: &Layout) -> Result<Rc<[Rect]>, AddConstraintErr
...
+
+        let chunks = Layout::default()
+            .direction(Direction::Horizontal)
+            .constraints([Constraint::Min(2), Constraint::Min(2), Constraint::Min(0)].as_ref())
+            .split(target);
+
+        assert_eq!(target.width, chunks.iter().map(|r| r.width).sum::<u16>());
+        assert_eq!(chunks[0].width, 2);
+        assert_eq!(chunks[1].width, 2);
+        assert_eq!(chunks[2].width, 6);
```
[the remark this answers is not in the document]
kdheepak: I changed it to simplify the tests. Hopefully this change in https://github.com/ratatui-org/ratatui/pull/395/commits/17790b7fbd2a3f9252c0f8b6b63f4560fd904920 will address this comment.

src/layout.rs
```diff
@@ -591,6 +608,200 @@ fn try_split(area: Rect, layout: &Layout) -> Result<Rc<[Rect]>, AddConstraintErr
...
+
+        let chunks = Layout::default()
+            .direction(Direction::Horizontal)
+            .constraints([Constraint::Min(2), Constraint::Min(2), Constraint::Min(0)].as_ref())
+            .prefer_equal_chunks(true)
+            .split(target);
```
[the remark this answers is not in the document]
kdheepak: I didn't choose the letters approach, which I like, because we are testing a width of `100` here and it was a very long string.

src/layout.rs
```diff
@@ -284,6 +292,8 @@ pub struct Layout {
     /// Whether the last chunk of the computed layout should be expanded to fill the available
     /// space.
     expand_to_fill: bool,
+    /// Whether each chunk should be similar in size to every other chunk
+    chunk_size_preference: ChunkSizePreference,
```
kdheepak, line 296: I would have liked to do this:

```
            chunk_size_preference: Default::default(),
```

But I can't call `non-const` in `const` functions.

joshka: Yeah - it's really annoying that trait methods like `Default::default()` can't be const :/ There's an RFC to fix it though - so someday...

src/layout.rs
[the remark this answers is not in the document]
joshka: What about something like AllocateLeftoverSpace (and have this incorporate the expand to fill option, which is pub(crate) so can be removed without breaking anything)
```rust
enum AllocateLeftoverSpace {
  None,
  Last,
  Equal,
}
```
(I'm never quite sure whether to add an explicit `None` vs using Option though)

src/layout.rs
```diff
@@ -590,6 +617,84 @@ fn try_split(area: Rect, layout: &Layout) -> Result<Rc<[Rect]>, AddConstraintErr
...
+            });
+        let chunks = layout.split(target);
+        chunks.iter().map(|r| (r.x, r.x + r.width)).collect()
+    }
+
+    #[test]
+    fn test_split_equally_in_underspecified_case() {
+        let target = Rect::new(100, 200, 10, 10);
+        assert_eq!(
+            equal_splits(false, vec![Min(2), Min(2), Min(0)], target),
```
joshka, line 643: nit: Can this either pass the ChunkSizePreference directly (add `use ChunkSizePreference::*` to make it concise) or create two methods. Generally bool params mean that people have to scroll up to work out the meaning.

joshka (changes requested):
This looks great now. I think this is the right way to make layout more responsive.

src/layout.rs
```diff
@@ -232,6 +233,13 @@ impl Rect {
     }
 }
 
+#[derive(Debug, Default, Clone, Eq, PartialEq, Hash)]
+pub enum ChunkSizePreference {
```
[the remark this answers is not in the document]
kdheepak: I like the idea of using a single `enum` for capturing the behavior of "dumping excess into the last chunk" _and_ "prefer making chunks equal in size".

I think one issue is that "prefer making chunks equal in size" can be done even when there's no "leftover space", whereas "dumping excess into the last chunk" only happens when there's "leftover space".

1. Should we combine these concepts that are slightly different? I'm leaning toward yes.
2. If so, what should the name of the `enum` be? I think `AllocateLeftoverSpace` underplays what happens when you choose `Equal`.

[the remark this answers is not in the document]
joshka: Asking for ideas in https://discord.com/channels/1070692720437383208/1072879985762762812/1142281786974544065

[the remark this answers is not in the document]
kdheepak: Maybe this?

```rust
#[derive(Debug, Default, Clone, Eq, PartialEq, Hash)]
pub enum AllocationPreference {
    #[default]
    None,
    LastGathersRemainder,
    AllEqual,
}
```

Also, do we want to allow `expand_fill: true` + `AllEqual`? Using the enum doesn't allow for that unless we add a new enum member, and in that case I don't know what to name it.

src/layout.rs
kdheepak: Fixed in https://github.com/ratatui-org/ratatui/pull/395/commits/819ed54821b6c4abe2c5e8d826e1da34cf5e0d52.

Cargo.toml
```diff
@@ -40,6 +40,7 @@ bitflags = "2.3"
 cassowary = "0.3"
 crossterm = { version = "0.27", optional = true }
 indoc = "2.0"
+itertools = "0.10"
```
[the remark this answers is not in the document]
kdheepak: Fixed in https://github.com/ratatui-org/ratatui/pull/395/commits/819ed54821b6c4abe2c5e8d826e1da34cf5e0d52.

src/layout.rs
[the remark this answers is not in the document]
kdheepak: I've settled for this for now:

```rust
#[derive(Debug, Default, Clone, Eq, PartialEq, Hash)]
pub enum SegmentSizePreference {
    EqualDistribution,
    #[default]
    LastTakesRemainder,
    None,
}
```

`AllEqual` also activates `LastGathersRemainder`.

joshka (approved):
Approved, but let's hold off a minute to workout the right naming for this.

We could alternatively keep this `pub(crate)` but expose it as `pub` within an "experimental" feature flag to help make it obvious that this might change.

src/layout.rs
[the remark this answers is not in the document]
hasezoey: > (I'm never quite sure whether to add an explicit None vs using Option though)

personally, if it is a enum similar to this, then i (as a user) would prefer to just have to define `AllocateLeftoverSpace::None` and `AllocateLeftoverSpace::Equal` instead of `(Option::)None` and `Some(AllocateLeftoverSpace::Equal)`

also as for a techincal aspect, this *could* save one WORD(bytes) of space, because of the option identification

kdheepak:
> i know this a bit unrelated, but this change causes some side-effects, please test against #404 with this change applied, see [#370 (comment)](https://github.com/ratatui-org/ratatui/issues/370#issuecomment-1678722149) for more context

I had to fix some tests after #404 was merged and after I rebased on top of it.

src/widgets/table.rs
```diff
@@ -751,25 +751,25 @@ mod tests {
...
-                &[(0, 6), (7, 6)]
+                &[(0, 7), (8, 6)]
             );
 
             // with selection, more than needed width
             assert_eq!(
                 get_test(&[Constraint::Ratio(1, 3), Constraint::Ratio(1, 3)], 20, 3),
-                &[(3, 6), (10, 6)]
+                &[(3, 7), (11, 6)]
             );
```
hasezoey, line 761: something is not right here, they should be the same width (there is more than available space)

kdheepak: I think this test is not deterministic? I stashed all my changes and ran this test on `main`; and I'm always getting a failure.

image

kdheepak: When I run on `main` with these two changes:

```rust
            dbg!(changes.get(&element.start));
            dbg!(changes.get(&element.end));
```

```
$ cargo test -- widgets::table::tests::table_column_widths::ratio_constraint --nocapture
running 1 test
[src/layout.rs:568] changes.get(&element.start) = None
[src/layout.rs:569] changes.get(&element.end) = Some(
    0.0,
)
[src/layout.rs:568] changes.get(&element.start) = Some(
    0.0,
)
[src/layout.rs:569] changes.get(&element.end) = Some(
    6.666666666666666,
)
[src/layout.rs:568] changes.get(&element.start) = Some(
    6.666666666666666,
)
[src/layout.rs:569] changes.get(&element.end) = Some(
    7.666666666666666,
)
[src/layout.rs:568] changes.get(&element.start) = Some(
    7.666666666666666,
)
[src/layout.rs:569] changes.get(&element.end) = Some(
    14.333333333333332,
)
thread 'widgets::table::tests::table_column_widths::ratio_constraint' panicked at 'assertion failed: `(left == right)`
  left: `[(0, 7), (8, 6)]`,
 right: `[(0, 6), (7, 6)]`', src/widgets/table.rs:752:13
test widgets::table::tests::table_column_widths::ratio_constraint ... FAILED

failures:

failures:
    widgets::table::tests::table_column_widths::ratio_constraint
```

I think in #404 there was no rounding, so your tests were passing then. But #408 introduced rounding.

There's something I don't understand though. Here's the git log history:

image

You can see that #408 was before #404. After merging #404 to `main`, the tests were indeed failing: https://github.com/ratatui-org/ratatui/actions/runs/5911487064/job/16034036047.

But my understanding was that GitHub's merge queue would run the tests again on the latest HEAD before merging; which would have caught this issue before merging #404. Does anyone know why that did not happen? Is there a feature that we need to enable on the merge queue feature on GitHub?

hasezoey: > You can see that https://github.com/ratatui-org/ratatui/pull/408 was before https://github.com/ratatui-org/ratatui/pull/404. And after merging https://github.com/ratatui-org/ratatui/pull/404 to main, the tests were indeed failing on CI: https://github.com/ratatui-org/ratatui/actions/runs/5911487064/job/16034036047.

after some quick tests, i came to the same conclusion

> But my understanding was that GitHub's merge queue would run the tests again on the latest HEAD before merging; which would have caught this issue before merging https://github.com/ratatui-org/ratatui/pull/404. Does anyone know why that did not happen? Is there a feature that we need to enable on the merge queue feature on GitHub?

i was also under that assumption, but maybe the branch-protection on `main` does not require all tests to pass? i came to this conclusion because of the following doc snippit

> Once a pull request has passed all required branch protection checks, a user with write access to the repository can add the pull request to the queue. The merge queue will ensure the pull request's changes pass all required status checks when applied to the latest version of the target branch and any pull requests already in the queue.

another possibility is that maybe because [of these lines](https://github.com/ratatui-org/ratatui/blob/b6b2da5eb761ac5894cc7a2ee67f422312b63cfc/.github/workflows/ci.yml#L34-L38) the *pull request head* was checked out *instead of the merged head*, but i am not sure, because i dont know if [this run](https://github.com/ratatui-org/ratatui/actions/runs/5911487064) is the run that tested the merge-ability, or the test after the merge

hasezoey: for now i have created a meta-issue to track this instead of this PR https://github.com/ratatui-org/ratatui/issues/418

kdheepak: Great, I'll resolve this conversation.

kdheepak:
I renamed it to `SegmentSize` and made it `pub(crate)` so that we can finalize the name later with more feedback.

src/layout.rs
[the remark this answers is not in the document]
kdheepak: I renamed it to `SegmentSize` and made it `pub(crate)` so that we can finalize the name later with more feedback.

src/widgets/table.rs
joshka: I fixed the table tests that were broken in https://github.com/ratatui-org/ratatui/pull/419

joshka:
I pushed a rebase on the fix in #419 - happy for this to be approved and merged now @kdheepak. We can work out the public name for the API at some point later. I lean towards using `Allocation` as the name and communicating that this is a preference not a requirement in the enum variant itself, and also using an imperative tone that says what the allocation will do (e.g. `PreferEvenWidths`, `FillLastArea`, etc.)

joshka (approved):

kdheepak:
I did the rebase locally as well, you are faster than me :) I'll discard my changes.

Yes, let's merge and figure out the right names for it later. I'm not tied to any names here.

joshka (approved):

sophacles (approved):

PR #396, opened 2023-08-12
refactor(layout): change "split"'s vars to be of known labeled values

hasezoey:
This PR changes all values for the `layout`'s `split`'s `vars` value to be known and labeled, this changes the code to be more readable and defines the values in a common place, also removing a extra match arm

### DIFF
--- a/src/layout.rs
+++ b/src/layout.rs
@@ -273,27 +273,13 @@ impl Layout {
 
 fn split(area: Rect, layout: &Layout) -> Rc<[Rect]> {
     let mut solver = Solver::new();
-    let mut vars: HashMap<Variable, (usize, usize)> = HashMap::new();
     let elements = layout
         .constraints
         .iter()
         .map(|_| Element::new())
         .collect::<Vec<Element>>();
-    let mut res = layout
-        .constraints
-        .iter()
-        .map(|_| Rect::default())
-        .collect::<Rc<[Rect]>>();
-
-    let results = Rc::get_mut(&mut res).expect("newly created Rc should have no shared refs");
 
     let dest_area = area.inner(&layout.margin);
-    for (i, e) in elements.iter().enumerate() {
-        vars.insert(e.x, (i, 0));
-        vars.insert(e.y, (i, 1));
-        vars.insert(e.width, (i, 2));
-        vars.insert(e.height, (i, 3));
-    }
     let mut ccs: Vec<CassowaryConstraint> =
         Vec::with_capacity(elements.len() * 4 + layout.constraints.len() * 6);
     for elt in &elements {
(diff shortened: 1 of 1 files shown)


kdheepak (approved):
LGTM

src/layout.rs
```diff
@@ -387,19 +396,10 @@ fn split(area: Rect, layout: &Layout) -> Rc<[Rect]> {
...
-                results[index].y = value;
-            }
-            2 => {
-                results[index].width = value;
-            }
-            3 => {
-                results[index].height = value;
-            }
-            _ => {}
+            AttributeType::X => results[index].x = value,
```
joshka, line 399: It's possible to flip this this and delete `vars`, the `AttributeType`, and the previous initialization of res/results.

`solver.fetch_changes()` returns `&[(Variable, f64)]` and we already store all the variables in `elements`. The following passes the tests, and produces the same layout example image:

```rust
let changes: HashMap<Variable, f64> = solver.fetch_changes().iter().copied().collect();
    let mut results = elements
        .iter()
        .map(|element| Rect {
            x: changes.get(&element.x).map(|&v| v as u16).unwrap_or(0),
            y: changes.get(&element.y).map(|&v| v as u16).unwrap_or(0),
            width: changes.get(&element.width).map(|&v| v as u16).unwrap_or(0),
            height: changes.get(&element.height).map(|&v| v as u16).unwrap_or(0),
        })
        .collect::<Vec<Rect>>();

    if layout.expand_to_fill {
        // Fix imprecision by extending the last item a bit if necessary
        if let Some(last) = results.last_mut() {
            match layout.direction {
                Direction::Vertical => {
                    last.height = dest_area.bottom() - last.y;
                }
                Direction::Horizontal => {
                    last.width = dest_area.right() - last.x;
                }
            }
        }
    }
    results.into()
```
image

src/layout.rs
```diff
@@ -271,9 +271,18 @@ impl Layout {
     }
 }
 
+/// Enum that represents the type of the value that has been inserted
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+enum AttributeType {
+    X = 0,
+    Y = 1,
+    Width = 2,
+    Height = 3,
```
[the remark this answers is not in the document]
hasezoey: if you mean the explicit numbers, i just transferred it over from the old way, just to be sure it is the exact same, though yes, they could be removed now

src/layout.rs
hasezoey: should i change this PR to use this code, or close this PR?

src/layout.rs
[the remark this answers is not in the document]
joshka: Yep (I'm not sure if I've actually used the right technical term for the numbers)

src/layout.rs
joshka: Yeah, I think this would be better to change it to this. @kdheepak do you have any perspective. The current code:
1. allocates an array of default `Rect`s
2. creates `Element`s that represent the x/y/width/height vars
3. stores the vars as keys in a `HashMap` onto the attribute type
4. runs the layout and gets the changes
5. for each change, work out which dimension of the Rect to update (using the vars map) and update it

This simplified code:
1. creates elements
2. runs the layout
3. creates the `Rect`s based on the layout changes (using zero for no changes)

These are result equivalent, but I'm not sure if there's some preferential rationale for doing it the first way.

kdheepak: The simplification looks good to me!

I have a question about the code you shared above though

```
        // Fix imprecision by extending the last item a bit if necessary
```

I'm not seeing any "if necessary" part in the code. The way you wrote it out above would always change the `last` width or height, right? Or am I missing something?

hasezoey: > I have a question about the code you shared above though

the `expand_to_fille` `if` is copy-paste from the current master:

https://github.com/ratatui-org/ratatui/blob/8c551588224ca97ee07948b445aa2ac9d05f997d/src/layout.rs#L406-L408

hasezoey: changed with 9a533494c42bcdd39bb35f8ab449709cbf7804fb

kdheepak: Thanks, I missed that. I guess I'm not understanding why it works in all cases in the original code (not relevant for this PR). I'll need to look into that.

joshka:
LGTM too. Can you please squash this into a single commit?

joshka:
Rebased on main, squashed and reworded the commit.

joshka (approved):

PR #397, opened 2023-08-12
refactor(layout): extract common constraints into function

hasezoey:
This PR refactors some duplicated constraints to be in a single function, reducing duplicate code and reducing possible mismatches

(re https://github.com/ratatui-org/ratatui/pull/395#discussion_r1292008364)

marking this PR as Draft, because i dont think the current struct name and field names are going to stay this way
also because i want to know if this is even a change that is wanted instead of having vertical and horizontal constraints separate, even if they are the same
(also some cleanup would need to be done, like removing commented-out fields)

### DIFF
--- a/src/layout.rs
+++ b/src/layout.rs
@@ -323,18 +323,16 @@ fn split(area: Rect, layout: &Layout) -> Rc<[Rect]> {
             for pair in elements.windows(2) {
                 ccs.push((pair[0].x + pair[0].width) | EQ(REQUIRED) | pair[1].x);
             }
-            for (i, size) in layout.constraints.iter().enumerate() {
-                add_constraints_for_constraint(&mut ccs, &elements[i], &dest_area, layout.direction, *size);
-            }
+            
         }
         Direction::Vertical => {
             for pair in elements.windows(2) {
                 ccs.push((pair[0].y + pair[0].height) | EQ(REQUIRED) | pair[1].y);
             }
-            for (i, size) in layout.constraints.iter().enumerate() {
-                add_constraints_for_constraint(&mut ccs, &elements[i], &dest_area, layout.direction, *size);
-            }
         }
+    }
+    for (i, size) in layout.constraints.iter().enumerate() {
+        add_constraints_for_constraint(&mut ccs, &elements[i], &dest_area, layout.direction, *size);
     }
     solver.add_constraints(&ccs).unwrap();
     for &(var, value) in solver.fetch_changes() {
--- a/src/layout.rs
+++ b/src/layout.rs
@@ -324,28 +324,7 @@ fn split(area: Rect, layout: &Layout) -> Rc<[Rect]> {
                 ccs.push((pair[0].x + pair[0].width) | EQ(REQUIRED) | pair[1].x);
             }
             for (i, size) in layout.constraints.iter().enumerate() {
-                ccs.push(elements[i].y | EQ(REQUIRED) | f64::from(dest_area.y));
-                ccs.push(elements[i].height | EQ(REQUIRED) | f64::from(dest_area.height));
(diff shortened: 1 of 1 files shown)


kdheepak:
Thanks for taking the time to tackle this, I like the idea.

When I'm reading code for a linear programming problem like this, the things I'm looking to understand easily are:

1. what are the decision variables?
2. what is the objective function?
3. what are the constraints?

In the code as it is right now, since everything is in one function and all the constraints are added to a vector, that is easy to follow (for me).

If this is to be refactored, I'd like it to be something like the following:

```rust

fn _decision_variables(...) {
    for (i, e) in elements.iter().enumerate() {
        vars.insert(e.x, (i, 0));
        vars.insert(e.y, (i, 1));
        vars.insert(e.width, (i, 2));
        vars.insert(e.height, (i, 3));
    }
}

fn _add_constraints_for_outer_bounds(...) {
    for elt in &elements {
        ccs.push(elt.width | GE(REQUIRED) | 0f64);
        ccs.push(elt.height | GE(REQUIRED) | 0f64);
        ccs.push(elt.left() | GE(REQUIRED) | f64::from(dest_area.left()));
        ccs.push(elt.top() | GE(REQUIRED) | f64::from(dest_area.top()));
        ccs.push(elt.right() | LE(REQUIRED) | f64::from(dest_area.right()));
        ccs.push(elt.bottom() | LE(REQUIRED) | f64::from(dest_area.bottom()));
    }
}

fn _add_constraints_for_first_element() {
    ...
}

fn split(area: Rect, layout: &Layout) -> Rc<[Rect]> {
    let s = Solver(); 
    let mut vars: HashMap<Variable, (usize, usize)> = HashMap::new();
    let elements: Vec<Element> = [...];
    let mut ccs: Vec<CassowaryConstraint> = [...];

    _decision_variables(...);

    _add_constraints_for_outer_bounds(...);
    _add_constraints_for_first_element(...);
    _add_constraints_based_on_user_preferences(...);
    if layout.expand_to_fill {
      _add_constraints_to_make_layout_spread_to_full_length(...);
    }
}
```

This maps more closely to how I think about the problem, and maps to the mathematical representation more closely:

https://github.com/ratatui-org/ratatui/pull/393#issuecomment-1674789843

I believe it'll also be easier to understand and test the code that way.

I personally don't think we should introduce any new `struct`s since that just makes it more separated from the math and adds another layer of abstraction here which can make it confusing to read and understand (imho); but I'm open to hearing what others have to say.

hasezoey:
> I personally don't think we should introduce any new structs since that just makes it more separated from the math and adds another layer of abstraction here which can make it confusing to read and understand (imho); but I'm open to hearing what others have to say.

i added a struct to avoid having multiple match arms (`let varA = match direction { horizontal => ..., vertical => ... };`) and also avoid having massing tuples (`let (a,b,c,d,e,...) = match direction {...};`)

> If this is to be refactored, I'd like it to be something like the following:

i dont understand cassowary constraints well enough yet (and how they are currently defined) to describe the function, this PR only tries to combine the 2 which were equal, only the values themself were reversed

kdheepak:
I'm assuming we don't want to do this anymore, now that the solver code has been refactored to be a lot simpler.

joshka:
Yeah - I think the refactor in #405 worked out easier to understand. Thank you @hasezoey for working on this - it sparked some good discussion and ideas around how this should be written.

I definitely lean towards extracting things to functions for readability. But I took a look at doing this for the current code, and the function parameter lists get formatted to wrap to 4-5 lines per extracted part part of the code. This makes the split function more difficult to read rather than easier (subjectively at least).

kdheepak:
Thanks @hasezoey for taking a stab at this!

PR #404, opened 2023-08-15
test(table): add test for consistent table-column-width

hasezoey:
This PR adds a test for widget `Table` to test that the widths for columns are consistent, added this test because i have found inconsistencies when trying #395, see https://github.com/ratatui-org/ratatui/issues/370#issuecomment-1678722149


src/widgets/table.rs
```diff
@@ -602,6 +602,41 @@ mod tests {
...
+        assert_eq!(widths[1], (5, 30)); // constraint "Max(30)"
+        assert_eq!(widths[2], (36, 10)); // constraint "Min(10)"
+        assert_eq!(widths[3], (47, 4)); // constraint "Max(4)"
+        assert_eq!(widths[4], (52, 10)); // constraint "Min(10)"
+
+        let used_length = widths.iter().fold(0, |acc, v| acc + v.1);
+
+        assert_eq!(used_length, 58); // test combined length
+
+        assert_eq!(62, widths[4].0 + widths[4].1); // test last draw position
```
joshka, line 637: These two assertions can never fail unless 4+30+10+4+10 != 58 and 52+10 != 62

joshka (commented):
The main thing I notice with this test is that it's testing a lot of things at once with more code than seems necessary to just test the behavior it's interested with. Let's try to simplify the test (perhaps by splitting this into multiple tests if that would help).

Observations:
- unsure what "consistent" means in this context. Use a name that describes this better.
- multiple rows, but nothing seems dependent on rows. Remove all but the first and use whatever test data makes sense
- 3 cells in each row, but 5 constraints. Add cells or comment on why this is necessary
- large numbers that don't seem obvious why those particular numbers were chosen. Use smaller numbers that more easily show the properties being tested.
- only testing one particular combination of constraints - unsure what the significance of this combination is. I think this test just tests that if there is enough space to render all the columns then all the column widths will be full (?)
- no point testing sums when the values that are being added are already tested

How about testing this by making a small function that returns the widths of a table with given constraints. E.g.:
```rust
fn get_column_widths() {
  use ratatui::layout::Constraints::*;
  fn test(widths: &[Constraint], width: u16, selection_width: u16) {
    let table = Table::new(vec![Row::new(vec!["test"; constraints.len()])])
        .widths(widths);
    table.get_column_widths(width, selection_width)
  }

  assert_eq!(test(&[Min(0)], 10, 0), &[(0, 10)], "min(0)");
  assert_eq!(test(&[Max(5)], 10, 0), &[0, 5)], "max(5)");
  ...
}
```

hasezoey:
are those tests better? (de96ea0448bce3cb35385b48f2b3e7d7807d0a61)

hasezoey:
> unsure what "consistent" means in this context. Use a name that describes this better.

> could you maybe explain what line you mean / context? i am pretty sure the comments like constraint Length(4) should be obvious to the constraint in the widths

> I meant that the word "consistent" has several meanings, and either:

i am sorry, i misread `unsure what "consistent" means in this context` as `unsure what "constraint" means in this context`

hasezoey:
> For the last test, what I'm looking for is a bit more intent to be communicated from part of the test itself 

i have removed it for now, because it is actually now covered by `min_constraint` (to be more specific, the first 2 `assert_eq` of that)

joshka (approved):
LGTM. Thanks for the PR. Can you squash this into a single commit please?

orhun (approved):

src/widgets/table.rs
```diff
@@ -602,6 +602,176 @@ mod tests {
...
+            // with selection, more than needed width
+            assert_eq!(
+                get_test(&[Constraint::Min(4), Constraint::Min(4)], 20, 3),
+                &[(3, 4), (8, 4)]
+            );
+
+            // without selection, less than needed width
+            assert_eq!(
+                get_test(&[Constraint::Min(4), Constraint::Min(4)], 7, 0),
+                &[(0, 4), (4, 3)]
```
hasezoey, line 694: i just noticed this evaluates to `4+1+3 = 8`, which is more than the given area, thankfully #405 would fix this

joshka: There's a followup I have waiting on #405 - which will fix the rounding of f64->u16, which should clear up the gaps and maybe some places that end up over allocating due to rounding.

joshka: Oh, and characterization tests are fine to be wrong like this - they document the current state, and can be changed when we fix things.

hasezoey: > Oh, and characterization tests are fine to be wrong like this - they document the current state, and can be changed when we fix things.

i know, i just wanted to mention this here, because it currently does not mention this via a comment, should i add a comment?

hasezoey: actually, nevermind this calculation excludes the spacer (spacer is width `0`), so this initial comment is wrong

context:

```txt
[
    Rect { // selection
        x: 0,
        y: 0,
        width: 0,
        height: 1,
    },
    Rect { // constraint 1
        x: 0,
        y: 0,
        width: 4,
        height: 1,
    },
    Rect { // spacer
        x: 4,
        y: 0,
        width: 0,
        height: 1,
    },
    Rect { // constraint 2
        x: 4,
        y: 0,
        width: 3,
        height: 1,
    },
]
```

and #405 actually changes to:

```txt
[
    Rect { // selection
        x: 0,
        y: 0,
        width: 0,
        height: 1,
    },
    Rect { // constraint 1
        x: 0,
        y: 0,
        width: 4,
        height: 1,
    },
    Rect { // spacer
        x: 4,
        y: 0,
        width: 0,
        height: 1,
    },
    Rect { // constraint 2
        x: 3,
        y: 0,
        width: 4,
        height: 1,
    },
]
```
meaning full length would be `8`, which is more than the given `7`

hasezoey:
should i rebase this PR onto main?

hasezoey:
rebased onto latest main and squashed commits

had to change one test line though, [this test](https://github.com/ratatui-org/ratatui/pull/404/files#diff-0cb5556404f3a0b1b3d617cf40ba7bf70756e9e4e7fae5e49e779921df29f5a8R694) had to be changed from

```rust
// without selection, less than needed width
assert_eq!(
    get_test(&[Constraint::Min(4), Constraint::Min(4)], 7, 0),
    &[(0, 4), (4, 3)]
);
```

into 

```rust
// without selection, less than needed width
assert_eq!(
    get_test(&[Constraint::Min(4), Constraint::Min(4)], 7, 0),
    &[(0, 4), (4, 4)]
);
```

joshka:
Can you explain the difference to that test? Would this have been because of the refactoring changes in #405 ? In which case, that was unintentional?

hasezoey:
i am pretty sure it is a side-effect of #405, though i am not sure which exact change caused it.
the difference now is that:
- for the first mentioned one, no spacer is allocated anymore, but both constraints are fullfilled through 0 to 7, resulting in 8 characters drawn, without spacer (even though there should probably be a spacer)*¹
- for the second one, i guess it actually allocated space for the selection width this time and somehow made the following constraints allocate the same space and so overwrite each other (see the same starting `x`(the first in the tuple) coordinate)

*¹ i guess it is up to debate if that is actually the correct choice that the `available` width is +1 after 0

some example from before / after #405
```text
(testing 2 constraints of "Min(4)")

after #405
0 width:
[(0, 0), (0, 0)]
1 width:
[(0, 1), (0, 1)]
2 width:
[(0, 2), (0, 2)]
3 width:
[(0, 3), (0, 3)]
4 width:
[(0, 4), (0, 4)]
5 width:
[(0, 4), (1, 4)]
6 width:
[(0, 4), (2, 4)]

before #405
0 width:
[(0, 0), (0, 0)]
1 width:
[(0, 1), (1, 0)]
2 width:
[(0, 2), (2, 0)]
3 width:
[(0, 3), (3, 0)]
4 width:
[(0, 4), (4, 0)]
5 width:
[(0, 4), (4, 1)]
6 width:
[(0, 4), (4, 2)]
```

joshka:
Thanks - that was really helpful. A minimal repro cutting the table out of the loop and just focusing on the layout is:
```rust
                .constraints(vec![Min(1), Length(0), Min(1)])
                .split(Rect::new(0, 0, 1, 1));
            assert_eq!(
                layout[..],
                [
                    Rect::new(0, 0, 1, 1),
                    Rect::new(1, 0, 0, 1),
                    Rect::new(0, 0, 1, 1), // x should never start before previous x
                ]
            )
```

src/widgets/table.rs
```diff
@@ -602,6 +602,176 @@ mod tests {
...
+
+        fn get_test(
+            widths: &[Constraint],
+            available_width: u16,
+            selection_width: u16,
+        ) -> Vec<(u16, u16)> {
+            let table = Table::new(vec![]).widths(widths);
+
+            table.get_columns_widths(available_width, selection_width)
+        }
```
kdheepak, line 617: Can you add a comment either in the function or as a docstring of what the return values are?

kdheepak: Alternatively, maybe we can rename the function from `get_test` to `get_x_width`?

hasezoey: sure, just to note it here, the return is the same as `get_columns_widths`, which is `x, width`

joshka: I definitely had to scroll up a few times to work out what this represented too.

joshka:
For the issue of the spacer going missing - this happens because the spacers have equal weighting as the columns.
There's two ways we could fix this:
- add optional weights to the constraints to make the spacers higher priority
- add a configurable explicit gap between chunks (though how configurable this should be would be a question to explore).

hasezoey:
> For the issue of the spacer going missing - this happens because the spacers have equal weighting as the columns.

i think this is somewhat out-of-scope for this PR, though yes something like configurable priorities would be great for such a case

joshka (approved):
Thanks for this PR and the patience to continue on it while there were a few other things in flux around it. Much appreciated!

PR #405, opened 2023-08-16
refactor(layout): simplify and doc split()

joshka:
This is mainly a reduction in density of the code with a goal of
improving mainatainability so that the algorithm is clear.

(Testing - the layout example renders the same)

### DIFF
--- a/src/layout.rs
+++ b/src/layout.rs
@@ -629,31 +629,6 @@ mod tests {
     use super::*;
 
     #[test]
-    fn test_vertical_split_by_height() {
-        let target = Rect {
-            x: 2,
-            y: 2,
-            width: 10,
-            height: 10,
-        };
-
-        let chunks = Layout::default()
-            .direction(Direction::Vertical)
-            .constraints(
-                [
-                    Constraint::Percentage(10),
-                    Constraint::Max(5),
-                    Constraint::Min(1),
-                ]
-                .as_ref(),
-            )
-            .split(target);
-
-        assert_eq!(target.height, chunks.iter().map(|r| r.height).sum::<u16>());
-        chunks.windows(2).for_each(|w| assert!(w[0].y <= w[1].y));
-    }
-
-    #[test]
     fn test_rect_size_truncation() {
         for width in 256u16..300u16 {
             for height in 256u16..300u16 {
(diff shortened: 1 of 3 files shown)


src/layout.rs
```diff
@@ -782,4 +764,409 @@ mod tests {
...
+                Paragraph::new(s).render(layout[i], &mut buffer);
+            }
+            assert_eq!(buffer.content, Buffer::with_lines(vec![expected]).content);
+        }
+
+        #[test]
+        fn length() {
+            test(Rect::new(0, 0, 1, 1), &[Length(0)], "a"); // underflow
+            test(Rect::new(0, 0, 1, 1), &[Length(1)], "a"); // exact
+            test(Rect::new(0, 0, 1, 1), &[Length(2)], "a"); // overflow
```
hasezoey, line 807: does `overflow` in this case mean that it allocates more than the given area, or that the individual elements themself should be longer?

src/layout.rs
```diff
@@ -782,4 +764,409 @@ mod tests {
...
+            let layout = Layout::default()
+                .direction(Direction::Horizontal)
+                .constraints(constraints)
+                .split(area);
+            let mut buffer = Buffer::empty(area);
+            for (i, c) in ('a'..='z').take(constraints.len()).enumerate() {
+                let s: String = c.to_string().repeat(area.width as usize);
+                Paragraph::new(s).render(layout[i], &mut buffer);
+            }
+            assert_eq!(buffer.content, Buffer::with_lines(vec![expected]).content);
```
hasezoey, line 800: why use `Paragraph` instead of testing sizes directly?

src/layout.rs
joshka: Mostly that the allocated space is more than the available space. Overflow is not strictly the correct term, but I couldn't think of a better one. I wanted to convey simply that there was a rationale for each test being there.

src/layout.rs
joshka: Visualization and brevity.

It's easy to look at each line in comparison to the line above and notice patterns and things which look right or wrong. With this many tests if the lines were wrapped, it would be more difficult to see those.

These tests should make it easy to see that there is no gaps and no overlaps - both of which are easy bugs to write with layout code - and visually show the impact of fixing the rounding.

src/layout.rs
hasezoey: maybe add terminology along the lines of

```rust
/// underflow = less than given space used
/// exact = space is used up
/// overflow = constraint requires more but got less
```
at least that is my current understanding of the given terminology

joshka:
Thanks for the review comments @hasezoey. I've updated the comments to better explain the testing approach (and removed a couple of comments that were accidentally left in while testing the rounding fix)

src/layout.rs
```diff
@@ -116,6 +116,22 @@ impl Constraint {
...
     }
+
+    pub(crate) fn to_cassowary(self, lhs: Variable, rhs: Variable) -> Vec<cassowary::Constraint> {
+        match self {
+            Constraint::Percentage(p) => vec![lhs | EQ(STRONG) | (rhs * f64::from(p) / 100.00)],
+            Constraint::Ratio(n, d) => vec![lhs | EQ(STRONG) | (rhs * f64::from(n) / f64::from(d))],
+            Constraint::Length(l) => vec![lhs | EQ(STRONG) | f64::from(l)],
+            Constraint::Max(m) => vec![
+                lhs | LE(STRONG) | f64::from(m),
+                lhs | EQ(WEAK) | f64::from(m),
```
kdheepak, line 127: > `lhs | LE(STRONG) | f64::from(m),`

By using `STRONG` it is not guaranteed that this constraint will be met.

> `lhs | EQ(WEAK) | f64::from(m),`

I think by adding this you incentivize the algorithm to be closer to the value in `Max`, i.e. `m`.

I'm not sure if both of these are what users would expect?

I personally would prefer the following:

```
lhs | LE(REQUIRED) | f64::from(m),
```

i.e. just one constraint. 

Maybe we should add another `Constraint::ApproxMax`?

src/layout.rs
```diff
@@ -366,126 +382,139 @@ impl Layout {
...
+            // be satisfied if the constraints are impossible to satisfy.
+            if layout.expand_to_fill {
+                if let Some(last) = elements.last() {
+                    solver.add_constraint(last.bottom() | EQ(STRONG) | area.bottom())?;
                 }
             }
+            // apply the constraints
+            for (constraint, element) in layout.constraints.iter().zip(elements.iter()) {
+                solver
+                    .add_constraints(constraint.to_cassowary(element.height, area.height).iter())?;
```
kdheepak, line 475: I like the refactor. 

I do have a mild preference for the `to_cassowary` method to be inlined here.

src/layout.rs
joshka: This PR is intended to be a pure refactor without changing any behavior. This is the current way that the min and max constraints work. I don't think it is necessarily right, and agree we should discuss how this should work.
https://github.com/ratatui-org/ratatui/blob/ea48af1c9abac7012e3bf79e78c6179f889a6321/src/layout.rs#L419-L428

The reason that this isn't required is to avoid panics when the user supplied values can't be satisfied.
Contrast the required constraints that can always be satisfied (area size, all chunks within the area, all chunks laid out left to right / top to bottom in order, and the left / top touching the start).

Another way of thinking about this is why should `Max` be more special than `Length` or `Min` if we have asked the area to expand to fill, and if it's not, then those should be required.

src/layout.rs
```diff
@@ -366,126 +382,139 @@ impl Layout {
...
-                match *size {
-                    Constraint::Min(v) | Constraint::Max(v) => {
-                        ccs.push(elements[i].width | EQ(WEAK) | f64::from(v));
-                    }
-                    _ => {}
+            // ensure the last element touches the right edge of the area. This constraint may not
+            // be satisfied if the constraints are impossible to satisfy.
+            if layout.expand_to_fill {
+                if let Some(last) = elements.last() {
+                    solver.add_constraint(last.right() | EQ(STRONG) | area.right())?;
```
kdheepak, line 440:
```
solver.add_constraint(last.right() | EQ(STRONG) | area.right())?;
```

Should this be `REQUIRED` instead?

src/layout.rs
joshka: I've looked at this both ways, and I think subjectively readability suffers when this is inlined. I wasn't sure if there was a better name or place to put the conversion however.

The main benefit on readability by putting this in a method is it's easy to see that the horizontal and vertical algorithms are essentially just switching top/bottom/left/right for left/right/top/bottom. This halves the cognitive load of the function.

I thought about some ways to condense the horizontal and vertical into a single function (e.g. by aliasing top/bottom/left/right as start/end/cross_start/cross_end), but can't find a way that keeps it simple, neat and obvious.

src/layout.rs
joshka: Actually yeah, I think you're right (incoming push for fixing this)

src/layout.rs
joshka: Actually, I found a way to do both that works nicely, (by making `Element` direction aware and adding `start()`, `end`, etc. to pick the right variable.

src/layout.rs
kdheepak: My bad, I misread / misremembered what the original was doing, and I didn't realize until your comment this PR didn't change any behavior.

> we should discuss how this should work.

I agree. I'm working on something that might help with this. I'll discuss in a separate thread.

src/layout.rs
```diff
@@ -366,140 +366,144 @@ impl Layout {
...
 /// A container used by the solver inside split
 #[derive(Debug, Clone, Copy, Eq, PartialEq, Hash)]
-struct Element {
+pub(crate) struct Element {
+    direction: Direction,
     x: Variable,
     y: Variable,
     width: Variable,
     height: Variable,
 }
```
kdheepak, line 501: I like this simplification, and I'm perfectly fine with it this way. But I _think_ you can go one step further.

For `Horizontal` layouts the resulting `Rect`s `top` and `bottom` are always equal to the `top` and `bottom` of the target area. Similarly, for `Vertical`, `left` and `right` are always equal to the target area's `left` and `right`.

In the code you have at the moment, you are running it through the algorithm:

```rust
    for element in &elements {
        solver.add_constraints(&[
            element.cross_start() | EQ(REQUIRED) | area.cross_start(),
            element.cross_end() | EQ(REQUIRED) | area.cross_end(),
        ])?;
    }
```

But you can probably just set the values on the rect in the result directly, based on direction.

If you do that, I _think_ `Element` can be simplified further to just:

```rust
struct Element {
  coordinate: Variable,
  length: Variable,
}
```

joshka: Neat - I'll give that a try

src/layout.rs
```diff
@@ -366,161 +366,147 @@ impl Layout {
...
-
-                match *size {
-                    Constraint::Min(v) | Constraint::Max(v) => {
-                        ccs.push(elements[i].width | EQ(WEAK) | f64::from(v));
-                    }
-                    _ => {}
-                }
+            Constraint::Ratio(n, d) => {
+                // avoid division by zero by using 1 when denominator is 0
+                let ratio = f64::from(n) / f64::from(d.max(1));
```
kdheepak, line 426: I think if `d` is zero, `d` should be `n`, i.e. ratio == 1. Right now, if you have ratio `Ratio(10, 0)` a strong constraint will be added that will try to make the rect size 10 times the target area and that constraint is always going to be violated.

joshka: I think I'm fine with leaving this for now as it matches the behavior in `Constraint::apply()`. Current behavior is to panic, which is probably more correct.

src/layout.rs
```diff
@@ -366,161 +366,147 @@ impl Layout {
...
+                Direction::Vertical => {
+                    last.height = inner.bottom() - last.y;
                 }
             }
         }
     }
 
-    results
+    Ok(results)
 }
```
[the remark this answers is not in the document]
kdheepak: I guess the rounding issue could result in this code being used. I'll resolve this conversation.

kdheepak:
Sorry about the staged second round of comments! This looks great now btw!

src/layout.rs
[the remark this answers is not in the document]
joshka: This issue is triggered by the rounding problem. E.g. splitting width 2 into position = [0, 0.99], width = [0.99, 1.01] when taking the floor becomes position [0, 0], width [0, 1], which then gets modified to widths: [0, 2].

My guess is that it will probably go away with that fix.

kdheepak (approved):
LGTM!

kdheepak (approved):

PR #406, opened 2023-08-16
Span tests

hasezoey:
This PR adds some tests for `widget::Span`, also adding a `From<&String>` implementation

what should be done about the doc-tests, should they stay in-place or should they be refactored?

### DIFF
--- a/src/text/span.rs
+++ b/src/text/span.rs
@@ -132,15 +132,12 @@ impl<'a> Span<'a> {
     }
 }
 
-impl<'a> From<String> for Span<'a> {
-    fn from(s: String) -> Span<'a> {
-        Span::raw(s)
-    }
-}
-
-impl<'a> From<&'a str> for Span<'a> {
-    fn from(s: &'a str) -> Span<'a> {
-        Span::raw(s)
+impl<'a, T> From<T> for Span<'a>
+where
+    T: Into<Cow<'a, str>>,
+{
+    fn from(s: T) -> Self {
+        Span::raw(s.into())
     }
 }
 
@@ -180,6 +177,13 @@ mod tests {
         let content_clone = content.clone();
         let span = Span::from(content);
         assert_eq!(span.content, Cow::Owned::<str>(content_clone));
+    }
+
+    #[test]
+    fn from_ref_string_borrowed_cow() {
+        let content = String::from("some string");
+        let span = Span::from(&content);
+        assert_eq!(span.content, Cow::Borrowed(content.as_str()));
     }
 
     #[test]
--- a/src/text/span.rs
+++ b/src/text/span.rs
@@ -155,3 +155,58 @@ impl<'a> Styled for Span<'a> {
         self
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn from_ref_str_borrowed_cow() {
+        let content = "some string";
+        let span = Span::from(content);
+        assert_eq!(span.content, Cow::Borrowed(content));
+    }
+
+    #[test]
+    fn from_string_ref_str_borrowed_cow() {
+        let content = String::from("some string");
+        let span = Span::from(content.as_str());
+        assert_eq!(span.content, Cow::Borrowed(content.as_str()));
+    }
+
+    #[test]
+    fn from_string_owned_cow() {
+        let content = String::from("some string");
+        let content_clone = content.clone();
+        let span = Span::from(content);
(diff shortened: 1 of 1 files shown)


orhun (approved):
It looks good with a quick glance. Let's see what other maintainers think.

src/text/span.rs
```diff
@@ -155,3 +161,73 @@ impl<'a> Styled for Span<'a> {
...
+    }
+
+    #[test]
+    fn from_ref_string_borrowed_cow() {
+        let content = String::from("some string");
+
+        let span = Span::from(&content);
+
+        assert_eq!(span.content, Cow::Borrowed(content.as_str()));
+    }
```
joshka, line 204: Nit: I understand the AAA approach, but for simple tests dropping the spaces make the tests easier to compare without  scrolling. A couple of small rearrangements too:

```suggestion
    #[test]
    fn from_ref_str_borrowed_cow() {
        let content = "some string";
        let span = Span::from(content);
        assert_eq!(span.content, Cow::Borrowed(content));
    }

    #[test]
    fn from_ref_string_borrowed_cow() {
        let content = String::from("some string");
        let span = Span::from(&content);
        assert_eq!(span.content, Cow::Borrowed(content.as_str()));
    }

    #[test]
    fn from_string_ref_str_borrowed_cow() {
        let content = String::from("some string");
        let span = Span::from(content.as_str());
        assert_eq!(span.content, Cow::Borrowed(content.as_str()));
    }

    #[test]
    fn from_string_owned_cow() {
        let content = String::from("some string");
        let span = Span::from(content.clone());
        assert_eq!(span.content, Cow::Owned::<str>(content));
    }
```

src/text/span.rs
```diff
@@ -155,3 +161,73 @@ impl<'a> Styled for Span<'a> {
...
+    #[test]
+    fn reset_should_set_style_reset() {
+        let mut span = Span::styled("test", Style::default().fg(crate::style::Color::Green));
+
+        assert_eq!(span.style, Style::default().fg(crate::style::Color::Green));
+
+        span.reset_style();
+
+        assert_eq!(span.style, Style::reset());
+        assert_ne!(span.style, Style::default());
```
joshka, line 215: How about simplifying this:
```suggestion
        let mut span = "test".green();
        span.reset_style();
        assert_eq!(span.style, Style::reset());
```

src/text/span.rs
```diff
@@ -155,3 +161,73 @@ impl<'a> Styled for Span<'a> {
...
+
+        span.patch_style(Style::default().fg(crate::style::Color::Green));
+
+        assert_eq!(
+            span.style,
+            Style::default()
+                .bg(crate::style::Color::Cyan)
+                .fg(crate::style::Color::Green)
+        );
+    }
```
joshka, line 232:
```suggestion
        let mut span = "test".on_cyan();
        span.patch_style(Style::new().green());
        assert_eq!(span.style, Style::new().green().on_cyan();
    }
```

joshka:
> what should be done about the doc-tests, should they stay in-place or should they be refactored?

The examples in the docs should contain enough information that the effect of the methods is unambiguous to a user. They shouldn't have to read unit tests or implementation detail in order to know how a method works.
They also shouldn't have any details that aren't necessary to help that understanding.
Last they should show idiomatic usage as much as possible (e.g. try to go with the simple stylized methods rather than the builder methods).

That is to say, the docs could do with a small amount of an update. They're not the worst, but also aren't perfect.

For the patch style, We'd want to convey that existing fg/bg/modifiers will be replaced if present in the patched style, and left alone if not. Probably avoid `assert_ne` and instead focus on what the expected output is. So a couple of examples that show going from raw, and replacing some existing colors / modifiers would be good.

src/text/span.rs
hasezoey: i didnt know `.green()` would make it a `Span`, but i think this should be explicit, it is also just testing before the reset that it is in a known proper condition (and that a change actually happened)

joshka: I tend to avoid precondition tests and instead test the preconditions in tests that actually care about that condition.
The concise approach seems sufficient enough. Not a blocker though.

joshka (approved):

joshka:
Can you please squash and rebase this on main?

hasezoey:
rebased onto latest main and squashed into 2 commits, one for the initial tests added, and one converting all the `From<*>` to `From<Into<Cow>>`

PR #408, opened 2023-08-17
Fix layout gap

joshka:
Fixes https://github.com/ratatui-org/ratatui/issues/367

~~Note: this builds on #405, so it has the commits for that here.~~ Look at just the last commit in this PR to understand it. The relevant change is just the addition of `round()`:
https://github.com/ratatui-org/ratatui/blob/2364d9b3098024459c4360900228e2b3e79fb848/src/layout.rs#L453-L454

Note each of the test changes is a subjectively better rendering choice. I'm not 100% sure if there are some tests that would make it easier to see this change's effect rather than those that are already changed.

I updated the colors example to use a ratio instead of a percentage as this now renders correctly without gaps.

Before (using `constraint: vec![Ratio(1,8); 8]` )

image

After:

image

### DIFF
--- a/src/layout.rs
+++ b/src/layout.rs
@@ -568,47 +568,28 @@ fn try_split(area: Rect, layout: &Layout) -> Result<Rc<[Rect]>, AddConstraintErr
     let changes: HashMap<Variable, f64> = solver.fetch_changes().iter().copied().collect();
 
     // convert to Rects
-    let mut results = elements
+    let results = elements
         .iter()
         .map(|element| {
-            let start = *changes.get(&element.start).unwrap_or(&0.0);
-            let end = *changes.get(&element.end).unwrap_or(&0.0);
+            let start = changes.get(&element.start).unwrap_or(&0.0).round() as u16;
+            let end = changes.get(&element.end).unwrap_or(&0.0).round() as u16;
             let size = end - start;
             match layout.direction {
                 Direction::Horizontal => Rect {
-                    x: start as u16,
+                    x: start,
                     y: inner.y,
-                    width: size as u16,
+                    width: size,
                     height: inner.height,
                 },
                 Direction::Vertical => Rect {
                     x: inner.x,
-                    y: start as u16,
+                    y: start,
                     width: inner.width,
-                    height: size as u16,
+                    height: size,
                 },
             }
         })
         .collect::<Rc<[Rect]>>();
-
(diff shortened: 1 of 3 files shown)


src/layout.rs
```diff
@@ -570,8 +570,8 @@ fn try_split(area: Rect, layout: &Layout) -> Result<Rc<[Rect]>, AddConstraintErr
     let mut results = elements
         .iter()
         .map(|element| {
-            let start = *changes.get(&element.start).unwrap_or(&0.0);
-            let end = *changes.get(&element.end).unwrap_or(&0.0);
+            let start = changes.get(&element.start).unwrap_or(&0.0).round();
+            let end = changes.get(&element.end).unwrap_or(&0.0).round();
```
kdheepak, line 574: I'm not sure this works in the general case. Can we add a comment about this? Something like `// TODO: explore corner cases for when .round() results in gaps`

joshka: Round can't leave gaps as end[i-1] always = start[i] (required constraint). If that constraint can not be met, then the split method will panic before this point.

This suggests that perhaps we could make this code explicitly gap free by only using the end values as variables and dropping the start variables.

src/layout.rs
```diff
@@ -976,8 +975,8 @@ mod tests {
             test(Rect::new(0, 0, 2, 1), &[FULL, FULL], "aa");
 
             // should probably be "abb" but this is what the current algorithm produces
```
a-kenji, line 977: This comment seems to be outdated now?

joshka: yep

joshka:
Rebased on main to pickup the bug fix in #408

Removed the no longer necessary comment in tests.

I also removed the code that checked the last element filled the area as the required constraint and corrected rounding take care of that fully now. I can't think of any possible way to trigger this code to mis-allocate the last constraint to less than the full amount that wouldn't also panic earlier due. Prior to this change, this was possible when the code took the floor of the start and floor of the width. E.g.: with area of 10:
- when the last element is (start: 9.4, width: 0.6), the old start and width were (start: 9, width: 0). With this change that becomes (start: 9, width: 1).
- Similarly for (9.6, 0.4) old was (9, 0) and this change becomes (10, 0), with the extra space allocated to the previous element.

kdheepak (approved):

orhun (approved):

PR #410, opened 2023-08-17
fix(layout): ensure left <= right

joshka:
The recent refactor (#405) missed the positive width constraint:
https://github.com/ratatui-org/ratatui/blob/ea48af1c9abac7012e3bf79e78c6179f889a6321/src/layout.rs#L380-L381

Addresses the bug noticed in https://github.com/ratatui-org/ratatui/pull/404#issuecomment-1681891809

ping @hasezoey - I took a quick look and this may change a few tests on #404. It might not yet be perfect and it may change again if https://github.com/ratatui-org/ratatui/pull/408 is merged on top of this.

### DIFF
--- a/src/layout.rs
+++ b/src/layout.rs
@@ -518,6 +518,7 @@ fn try_split(area: Rect, layout: &Layout) -> Result<Rc<[Rect]>, AddConstraintErr
         solver.add_constraints(&[
             element.start | GE(REQUIRED) | area.start,
             element.end | LE(REQUIRED) | area.end,
+            element.start | LE(REQUIRED) | element.end,
         ])?;
     }
     // ensure there are no gaps between the elements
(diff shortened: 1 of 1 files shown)


src/layout.rs
```diff
@@ -1133,6 +1138,35 @@ mod tests {
                     Rect::new(0, 1, 1, 0)
                 ]
             );
+
+            // minimal bug from
+            // https://github.com/ratatui-org/ratatui/pull/404#issuecomment-1681850644
```
hasezoey, line 1143: maybe mention what this test is meant to test (that start is below end and below area length?)

hasezoey:
can confirm that this change gives the behavior of before #405 in #404 again (behavior mentioned in https://github.com/ratatui-org/ratatui/pull/404#issuecomment-1681891809)

kdheepak (approved):

orhun (approved):

src/layout.rs
joshka: I will definitely add a few more tests that hit more edge cases more explicitly in another PR. I'm going to merge this now to fix the bug and come back to this.

PR #411, opened 2023-08-18
Simplify split function

kdheepak:
Further simplification of `try_split` by removing the `area` element and constraints for that, and using the `start` and `end` values directly.

### DIFF
--- a/src/layout.rs
+++ b/src/layout.rs
@@ -493,17 +493,11 @@ fn try_split(area: Rect, layout: &Layout) -> Result<Rc<[Rect]>, AddConstraintErr
     let mut solver = Solver::new();
     let inner = area.inner(&layout.margin);
 
-    let (start, end) = match layout.direction {
-        Direction::Horizontal => (inner.x, inner.right()),
-        Direction::Vertical => (inner.y, inner.bottom()),
+    let (area_start, area_end) = match layout.direction {
+        Direction::Horizontal => (f64::from(inner.x), f64::from(inner.right())),
+        Direction::Vertical => (f64::from(inner.y), f64::from(inner.bottom())),
     };
-
-    // setup the bounds of the area
-    let area = Element::new();
-    solver.add_constraints(&[
-        area.start | EQ(REQUIRED) | f64::from(start),
-        area.end | EQ(REQUIRED) | f64::from(end),
-    ])?;
+    let area_size = area_end - area_start;
 
     // create an element for each constraint that needs to be applied. Each element defines the
     // variables that will be used to compute the layout.
@@ -516,8 +510,8 @@ fn try_split(area: Rect, layout: &Layout) -> Result<Rc<[Rect]>, AddConstraintErr
     // ensure that all the elements are inside the area
     for element in &elements {
         solver.add_constraints(&[
-            element.start | GE(REQUIRED) | area.start,
-            element.end | LE(REQUIRED) | area.end,
+            element.start | GE(REQUIRED) | area_start,
+            element.end | LE(REQUIRED) | area_end,
             element.start | LE(REQUIRED) | element.end,
         ])?;
     }
@@ -527,12 +521,12 @@ fn try_split(area: Rect, layout: &Layout) -> Result<Rc<[Rect]>, AddConstraintErr
     }
     // ensure the first element touches the left/top edge of the area
     if let Some(first) = elements.first() {
-        solver.add_constraint(first.start | EQ(REQUIRED) | area.start)?;
+        solver.add_constraint(first.start | EQ(REQUIRED) | area_start)?;
     }
     // ensure the last element touches the right/bottom edge of the area
     if layout.expand_to_fill {
         if let Some(last) = elements.last() {
-            solver.add_constraint(last.end | EQ(REQUIRED) | area.end)?;
+            solver.add_constraint(last.end | EQ(REQUIRED) | area_end)?;
         }
     }
     // apply the constraints
@@ -540,12 +534,12 @@ fn try_split(area: Rect, layout: &Layout) -> Result<Rc<[Rect]>, AddConstraintErr
         match constraint {
             Constraint::Percentage(p) => {
                 let percent = f64::from(p) / 100.00;
-                solver.add_constraint(element.size() | EQ(STRONG) | (area.size() * percent))?;
+                solver.add_constraint(element.size() | EQ(STRONG) | (area_size * percent))?;
             }
             Constraint::Ratio(n, d) => {
                 // avoid division by zero by using 1 when denominator is 0
                 let ratio = f64::from(n) / f64::from(d.max(1));
-                solver.add_constraint(element.size() | EQ(STRONG) | (area.size() * ratio))?;
+                solver.add_constraint(element.size() | EQ(STRONG) | (area_size * ratio))?;
             }
             Constraint::Length(l) => {
                 solver.add_constraint(element.size() | EQ(STRONG) | f64::from(l))?


src/layout.rs
```diff
@@ -527,25 +520,25 @@ fn try_split(area: Rect, layout: &Layout) -> Result<Rc<[Rect]>, AddConstraintErr
...
+            solver.add_constraint(last.end | EQ(REQUIRED) | end)?;
         }
     }
     // apply the constraints
     for (&constraint, &element) in layout.constraints.iter().zip(elements.iter()) {
         match constraint {
             Constraint::Percentage(p) => {
                 let percent = f64::from(p) / 100.00;
-                solver.add_constraint(element.size() | EQ(STRONG) | (area.size() * percent))?;
+                solver.add_constraint(element.size() | EQ(STRONG) | ((end - start) * percent))?;
```
orhun, line 536: I assuming there are no edge cases where we need `checked_sub` here?

joshka (approved):
General approval of this change with suggestions that are non blocking.

I think if we're going to do this, then what about if we go one step further and remove Element altogether?

The only variables that we care about are the `end` of each chunk (as the `start` is required to be strictly equal to the previous chunk).

I think there's probably even a simple approach where the start and end of the area don't really need to be treated any differently than the start / end of a chunk (using an Vec of the positions and `.window(2)` to setup every constraint).

A quick and dirty check of this suggests that we delete 18-24 lines of code in the diff

joshka:
Note - this doesn't have #408 included yet (just merged that)

src/layout.rs
```diff
@@ -494,17 +494,10 @@ fn try_split(area: Rect, layout: &Layout) -> Result<Rc<[Rect]>, AddConstraintErr
     let inner = area.inner(&layout.margin);
 
     let (start, end) = match layout.direction {
-        Direction::Horizontal => (inner.x, inner.right()),
-        Direction::Vertical => (inner.y, inner.bottom()),
+        Direction::Horizontal => (f64::from(inner.x), f64::from(inner.right())),
+        Direction::Vertical => (f64::from(inner.y), f64::from(inner.bottom())),
```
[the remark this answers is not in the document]
kdheepak: Resolved in https://github.com/ratatui-org/ratatui/pull/411/commits/7693ec29e16c606c07c8d6dd0720e07025b696ec.

src/layout.rs
kdheepak: Resolved in https://github.com/ratatui-org/ratatui/pull/411/commits/7693ec29e16c606c07c8d6dd0720e07025b696ec.

orhun (approved):

joshka:
> I'm not sure if I'm following this suggestion. We'll always need two decision variables for each chunk, e.g. start and end OR start and length, right? Maybe I'm just being slow this morning, can you can elaborate further?

If you have: `area = [0..=10]`, and `constraints = [Length(3), Length(4), Length(5)]`, and setup an array of variables `positions = [area_start, end[0], end[1], end[2]]`, then the constraints can be applied to `positions.window(2)`. There's no need to have variables for `start[0..2]` as those just strictly equal `[area_start, end[0], end[1]]` and the size of each is just `position[n] - position[n-1]`.

```rust
    let (area_start, area_end, area_size) = match layout.direction {
        Direction::Horizontal => (
            f64::from(inner.left()),
            f64::from(inner.right()),
            f64::from(inner.width),
        ),
        Direction::Vertical => (
            f64::from(inner.top()),
            f64::from(inner.bottom()),
            f64::from(inner.height),
        ),
    };
    // the positions of each chunk. The first variable is the start, and each successive variable is
    // the end of that chunk and the start of the next one
    let positions = std::iter::from_fn(|| Some(Variable::new()))
        .take(layout.constraints.len() + 1)
        .collect_vec();

    // ensure the first element is at the start
    solver.add_constraint(positions[0] | EQ(REQUIRED) | area_start)?;

    // ensure the positions are monotonically increasing
    for &position in positions.iter() {
        solver.add_constraints(&[
            position | GE(REQUIRED) | area_start,
            position | LE(REQUIRED) | area_end,
        ])?;
    }

    // ensure the last element touches the right/bottom edge of the area
    if layout.expand_to_fill {
        if let Some(&last) = positions.last() {
            solver.add_constraint(last | EQ(REQUIRED) | area_end)?;
        }
    }

    let sizes = positions
        .iter()
        .tuple_windows()
        .map(|(&start, &end)| end - start)
        .collect_vec();

    // apply the constraints
    for (size, &constraint) in sizes.iter().cloned().zip(layout.constraints.iter()) {
        solver.add_constraint(size.clone() | GE(REQUIRED) | 0.0)?;
        match constraint {
                solver.add_constraint(size | EQ(STRONG) | (area_size * percent))?;
            }
            Constraint::Ratio(n, d) => {
                // avoid division by zero by using 1 when denominator is 0
                let ratio = f64::from(n) / f64::from(d.max(1));
                solver.add_constraint(size | EQ(STRONG) | (area_size * ratio))?;
            }
            Constraint::Length(l) => solver.add_constraint(size | EQ(STRONG) | f64::from(l))?,
            Constraint::Max(m) => {
                solver.add_constraints(&[
                    size.clone() | LE(STRONG) | f64::from(m),
                    size | EQ(WEAK) | f64::from(m),
                ])?;
            }
            Constraint::Min(m) => {
                solver.add_constraints(&[
                    size.clone() | GE(STRONG) | f64::from(m),
                    size | EQ(WEAK) | f64::from(m),
                ])?;
            }
        }
    }
```

> Also for consideration, we might want to add this constraint in the future (equal sizes):

This can be written as:
```rust
for (left, right) in sizes.iter().tuple_window() {
  solver.add_constraint(left | EQ(WEAK) | right)?;
}
```

It's probably not necessary to add all the comparisons as [a=b, b=c] implies [a=c]. I'm not sure what the constraint solver does though for constraints that can't be satisfied.

kdheepak:
Okay, that makes sense. Thanks for explaining and the code examples. I do find it more confusing to read though. 

> It's probably not necessary to add all the comparisons as [a=b, b=c] implies [a=c]. I'm not sure what the constraint solver does though for constraints that can't be satisfied.

I think it is necessary to add all combinations because you can have one of the chunks be smaller because of a higher priority constraint. For example take a case like `[Percentage(50), Length(10), Percentage(50)]`, we want `b` to be close to `10` units but the rest to be spread equally between the `Percentage(50)`s. I haven't tested this case specifically but I can imagine corner cases like this popping up.

joshka:
You're right. I tried that and found that I was wrong. Your explanation of why makes a lot of sense - I couldn't quite get there. I replaced `tuple_window()` with `tuple_combinations()` and works well.

Percentage shouldn't grow / shrink any more than length should, as it's based on a percentage of the total not a percentage of the amount after length is taken out. 50%/10/50% should render (these would be good tests in the other PR):
- size = 10: 5/10/5 => [2.5, 7.5, 2.5] => [3, 5, 2]
- size = 20: 10/10/10 => [6.67, 6.67, 6.67] => [7, 6, 7]
- size = 40: 20/10/20 => [16, 8, 16]
- size = 100: 50/10/50 => [45.4, 9.1, 45.4] => [45, 10, 45]

kdheepak:
A diagram like this certainly helps:

```
positions: [start  end0  end1  end2 ]
positions: [ 0.0   1.0   3.0   10.0 ]
sizes:     [    1.0   2.0   7.0     ]
sizes:     [   size0 size1 size2    ]
```

I'm not convinced it is worth it for the extra simplification in this case. If we really wanted to remove `Element` we can create `start` and `end` variables in the function. 

I do like the way it is written right now with `Element` and with `element.size()` returning an `Expression`. It makes it very easy for me at a glance to understand exactly what is going on.

PR #423, opened 2023-08-21
chore(changelog): show full commit message

joshka:
This allows someone reading the changelog to search for information
about breaking changes or implementation of new functionality.

- refactored the commit template part to a macro instead of repeating it
- added a link to the commit and to the release
- updated the current changelog for the alpha and unreleased changes
- Automatically changed the existing * lists to - lists

---

image


CHANGELOG.md
```diff
@@ -321,34 +789,37 @@ And most importantly, special thanks to [Florian Dehau](https://github.com/fdeha
 
 ### Features
 
-* Bump `crossterm` to `0.25`
+- Bump `crossterm` to `0.25`
```
joshka, line 792: Everything below here *should* be automatic changes done by vscode.

CHANGELOG.md
joshka, line 1: The changes here are from:
```
git cliff -- v0.22.0..HEAD
```
I.e. I only updated the not released part to show how this will render. We could update the changelog for 0.19..0.22 as well to help document the previous changes as well.

joshka (commented):
This is in addition to sometime making a BREAKING_CHANGES.md, not a replacement for that idea.

CHANGELOG.md
```diff
@@ -1,5 +1,473 @@
...
+## [unreleased]
+
+### Features
+
+- *(uncategorized)* Add weak constraints to make rects closer to each other in size ✨ ([#395](https://github.com/ratatui-org/ratatui/issues/395))
+([6153371](https://github.com/ratatui-org/ratatui/commit/61533712be57f3921217a905618b319975f90330))
+
+  ````text
+  Also make `Max` and `Min` constraints MEDIUM strength for higher priority over equal chunks
+  ````
```
kdheepak, line 14: Is it possible to make these text codeblocks a html details tag?

```
  <details>
  <summary>Show more</summary>
  Also make `Max` and `Min` constraints MEDIUM strength for higher priority over equal chunks
  <details>
```

would render as 

  <details>
  <summary>Show more</summary>
  Also make `Max` and `Min` constraints MEDIUM strength for higher priority over equal chunks
  </details>
  
---

If it is possible, is that something we might want to do?

joshka: I looked at that as an option first. I couldn't get it to render in lists yesterday, but having a new look, I think indent fixes that.

The bigger problem I see with this is that it means that you can't search for text in the non-expanded regions - which limits the use case of being able to easily find when something specific changed.

Perhaps a nice way to have both would be to have a high level overview up top that links to more details of each commit underneath, but that might be overkill for now.

kdheepak:
I personally like this change, but I'm also curious if others have opinions on whether it is too verbose or not for a changelog.

kdheepak (approved):

joshka:
> I personally like this change, but I'm also curious if others have opinions on whether it is too verbose or not for a changelog.

I gave that a bit of consideration.

- The pro side is that more details = easier to search for information using Ctrl/Cmd+F. Finding what changed in a git commit history is often more difficult compared to searching a single file like this.
- The cons are that having too many details can get in the way of seeing the high level view of things overview.

I think the tradeoff is reasonable, but there's also a third approach we could have that moves the details to another section under the list of changes. It's a bit more work though.

joshka:
P.s. Don't merge this just yet - letting @orhun take some time to look at some git cliff stuff before we do that.

orhun (approved):
I just tested this and liked the changelog output. I agree with @joshka's comments about tradeoff.

Also, using a macro is a nice refactor 💯 Applied it in [git-cliff's changelog](https://github.com/orhun/git-cliff/commit/c695ca308e5aada03cbd65085497436ac5c341e9).

PR #425, opened 2023-08-22
chore(traits): add Display and FromStr traits

### LINKED ISSUE #307: Chore: implement common traits

joshka:
Use [strum](https://docs.rs/strum/latest/strum/) for most of these, with a couple of manual implementations, and
related tests

Fixes #307

note: this is mainly the enums, the structs tend not to have obvious display representations.

ping @TieWay59 (due to previous work on trait impls)

### DIFF
--- a/src/layout.rs
+++ b/src/layout.rs
@@ -2,6 +2,7 @@ use std::{
     cell::RefCell,
     cmp::{max, min},
     collections::HashMap,
+    fmt,
     rc::Rc,
 };
 
@@ -11,8 +12,9 @@ use cassowary::{
     WeightedRelation::{EQ, GE, LE},
 };
 use itertools::Itertools;
+use strum::{Display, EnumString};
 
-#[derive(Debug, Default, Clone, Copy, Eq, PartialEq, Hash)]
+#[derive(Debug, Default, Display, EnumString, Clone, Copy, Eq, PartialEq, Hash)]
 pub enum Corner {
     #[default]
     TopLeft,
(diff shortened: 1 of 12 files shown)


src/layout.rs
```diff
@@ -233,8 +268,8 @@ impl Rect {
     }
 }
 
-#[derive(Debug, Default, Clone, Eq, PartialEq, Hash)]
-pub enum SegmentSize {
+#[derive(Debug, Default, Display, EnumString, Clone, Eq, PartialEq, Hash)]
+pub(crate) enum SegmentSize {
```
kdheepak, line 272: Nice catch.

kdheepak (approved):

PR #426, opened 2023-08-24
chore(spans): remove deprecated `Spans` type

joshka:
`Spans` was replaced with `Line` in 0.21.0. This commit removes the
deprecated type and all associated code. `Buffer::set_spans` was
replaced with `Buffer::set_line`.


orhun (approved):
Haven't tested the code but general +1

joshka:
The main question on this is whether it's for 0.23 or 0.24? It's been 3 months since the deprecation, which is probably enough time to handle this sort of thing.

kdheepak:
Looking at https://crates.io/crates/ratatui/reverse_dependencies there's still quite a few projects still on `0.20.1`. I'm assuming there wouldn't even see the deprecation warning. 

That said, I'm assuming that those on v0.20.1 are probably only going to upgrade when there's a feature they want. So it's probably fine to go for it in this release.

joshka:
That's actually a reasonable check - I think perhaps it's worth holding off until after 0.23 then.

mindoodoo:
Should we maybe add mention somewhere of how long we keep things in a deprecated state before we remove them ? Do we already have that and I missed it ?

Consistency in our deprecation process would be good to have for sure

joshka:
Deprecation should be rare enough that we can mostly handle it on a case by case basis IMO.

Latest push is a rebase and removed a couple of tests that referenced `Spans`.

I think this is probably worth merging for 0.24

joshka:
Rebased on main to resolve conflicts (changes from mod.rs -> xxx.rs)

orhun (approved):

PR #427, opened 2023-08-24
docs(span): finish docs and tests for Span

joshka:
image

### DIFF
--- a/src/text/span.rs
+++ b/src/text/span.rs
@@ -6,22 +6,66 @@ use unicode_width::UnicodeWidthStr;
 use super::StyledGrapheme;
 use crate::style::{Style, Styled};
 
-/// A string where all graphemes have the same style.
+/// Represents a part of a line that is contiguous and where all characters share the same style.
+///
+/// A `Span` is the smallest unit of text that can be styled. It is usually combined in the [`Line`]
+/// type to represent a line of text where each `Span` may have a different style.
+///
+/// # Examples
+///
+/// A `Span` with `style` set to [`Style::default()`] can be created from a `&str`, a `String`, or
+/// any type convertible to [`Cow<str>`].
+///
+/// ```rust
+/// # use ratatui::prelude::*;
+/// let span = Span::raw("test content");
+/// let span = Span::raw(String::from("test content"));
+/// let span = Span::from("test content");
+/// let span = Span::from(String::from("test content"));
+/// let span: Span = "test content".into();
+/// let span: Span = String::from("test content").into();
+/// ```
+///
+/// Styled spans can be created using [`Span::styled`] or by converting strings using methods from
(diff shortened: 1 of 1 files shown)


src/text/span.rs
```diff
@@ -6,22 +6,55 @@ use unicode_width::UnicodeWidthStr;
...
+/// style.
+///
+/// It is the smallest unit of text that can be styled. It is used by the [`Line`] type to represent
+/// a line of text where each grapheme has its own style.
+///
+/// # Examples
+///
+/// `Span`a can be created from a `&str`, a `String` or a `Cow<str>`. It can also be created from a
+/// `&str` or a `String` using the `From` trait, which will create a `Span` with no style.
+/// Alternatively, a `Span` can be created from a `&str` or a `String` using the `Stylize` trait,
```
mindoodoo, line 19: All the creation options listed here except for `Cow<str>` seem redundant to the example, so maybe a little wordy considering the example follows right after ?

lmk what you think

src/text/span.rs
```diff
@@ -6,22 +6,55 @@ use unicode_width::UnicodeWidthStr;
...
+/// let span = Span::raw("test content");
+/// let span = Span::raw(String::from("test content"));
+/// let span = Span::from("test content");
+/// let span = Span::from(String::from("test content"));
+/// let span = "test content".green();
+/// let span = String::from("test content").green();
+/// ```
+///
+/// Spans can be styled using the `Stylize` trait to combine foreground and background colors and
+/// modifiers.
```
mindoodoo, line 32:
```suggestion
/// Spans can also be styled using the `Stylize` trait.
```

nit: Imho a bit wordy and doesn't need to detail what can be stylized

src/text/span.rs
```diff
@@ -54,78 +86,72 @@ impl<'a> Span<'a> {
...
+            .graphemes(true)
+            .filter(|g| *g != "\n")
             .map(move |g| StyledGrapheme {
                 symbol: g,
                 style: base_style.patch(self.style),
             })
-            .filter(|s| s.symbol != "\n")
     }
 
     /// Patches the style an existing Span, adding modifiers from the given style.
```
mindoodoo, line 130:
```suggestion
    /// Patches the style of the Span, adding modifiers from the given style.
```

mindoodoo (approved):
Couple nitpicks otherwise LGTM ! Thanks for this PR, was needed !

joshka:
Updated wording to simplify it somewhat. My reason for being more verbose than necessary is to make it easy on new rust developers (who may not have seen used `Cow`), and older tui-rs developers that may not be aware of the `Stylize` trait and the shortcut methods.

---

Screen Shot 2023-08-24 at 22 49 20

PR #430, opened 2023-08-25
refactor(barchart): reduce some calculations

karthago1:
Calculating `label_offset` is unnecessary, if we just render the group label after rendering the bars. We can just reuse `bar_y`.

this commit is getting rid of `let label_offset = bars.len() as u16 * (self.bar_width + self.bar_gap) - self.bar_gap;`

### DIFF
--- a/src/widgets/barchart/mod.rs
+++ b/src/widgets/barchart/mod.rs
@@ -156,7 +156,7 @@ impl<'a> BarChart<'a> {
         self
     }
 
-    /// set the direction ob the bars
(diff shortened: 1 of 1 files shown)


karthago1:
I am suspecting that the codecov is reduced, because we have one line less in total. If I add somewhere an empty line, I think the codecov will be happy again.

Please write, if I have to add an Empty line.

joshka (approved):

joshka:
Thanks for the PR

PR #434, opened 2023-08-25
test(test_backend): add tests for TestBackend coverage

joshka:

### DIFF
--- a/src/backend/test.rs
+++ b/src/backend/test.rs
@@ -183,3 +183,135 @@ impl Backend for TestBackend {
         Ok(())
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn new() {
+        assert_eq!(
+            TestBackend::new(10, 2),
+            TestBackend {
+                width: 10,
+                height: 2,
+                buffer: Buffer::with_lines(vec!["          "; 2]),
+                cursor: false,
+                pos: (0, 0),
+            }
+        );
+    }
+    #[test]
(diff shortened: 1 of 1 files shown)


src/backend/test.rs
```diff
@@ -183,3 +183,137 @@ impl Backend for TestBackend {
...
+            buffer_view(&buffer),
+            format!(
+                "\"{multi_byte_char}\" \
+                Hidden by multi-width symbols: \
+                [(1, \" \"), (2, \" \"), (3, \" \"), (4, \" \"), \
+                (5, \" \"), (6, \" \"), (7, \" \")]\n",
+                multi_byte_char = multi_byte_char
+            )
+        );
+    }
```
kdheepak, line 224: I don't quite understand why this test passes.

Also can you use a `#"..."#` i.e raw string here?

joshka: It's a fairly wide string (140 chars) with only a single newline at the end.
The `\` at the end of the line allows the lines to wrap
Raw strings don't support wrapping like this, so it was a choice of a shorter line vs raw string.
I'll change it to a raw string though as if it's not immediately obvious why this is so, then you're probably right that it should be changed.

joshka: Fixed in [504f3eb](https://github.com/ratatui-org/ratatui/pull/434/commits/504f3ebd2bdeb546dc1fd7e0c8f14c10b7849e99)

kdheepak (approved):

orhun (approved):

PR #438, opened 2023-08-26
docs(lib): extract feature documentation from Cargo.toml

orhun:
Utilize [document-features](https://github.com/slint-ui/document-features) crate to extract the feature documentation from `Cargo.toml`

before / after:

Screenshot 2023-08-26 at 14-32-26 ratatui - Rust

One thing to note, we can possibly make it an optional dependency: https://github.com/slint-ui/document-features/pull/8

Let me know what you think 🐻

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,11 +33,7 @@ termion = { version = "2.0", optional = true }
 ## enables the [`TermwizBackend`] backend and adds a dependency on the [Termwiz crate].
 termwiz = { version = "0.20.0", optional = true }
 
-#! The following optional features are available for all backends:
-## enables serialization and deserialization of style and color types using the [Serde crate].
-## This is useful if you want to save themes to a file.
 serde = { version = "1", optional = true, features = ["derive"] }
-
 bitflags = "2.3"
 cassowary = "0.3"
 indoc = "2.0"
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -61,6 +60,27 @@ fakeit = "1.1"
 rand = "0.8"
 pretty_assertions = "1.4.0"
 
+[features]
+default = ["crossterm"]
+serde = ["dep:serde", "bitflags/serde"]
```
Valentin271, line 65: Since it includes more than just the `serde` crate, I'd document the `serde` feature here. I think it's more natural when reading the source code.

Valentin271 (commented):
Saw this crate recently too. I think its a great idea, its easier to maintain like so.

As for making `document-feature` an optional dependency, I think it's better because users don't need to build it every time.  
That would lead to a feature named `document-feature` which makes sense to me.  
Other crates are using this pattern too like [egui](https://docs.rs/crate/egui/0.22.0/features).

As the number of crates using `document-feature` grows ([only 138](https://crates.io/crates/document-features/reverse_dependencies?page=10) right now), I think people will get used to having this feature and will instinctly know what it is, like `serde`.

Cargo.toml
orhun: Makes sense, 8bc8d5c6b925554ccd5e24d4b938f5b7d5a4d164

orhun:
Made the dependency optional in 78363c5e7054fdbf3dec645aaefbc5989ae06b71

joshka (approved):

joshka:
The https://lib.rs changes for showing features is neat too:
https://users.rust-lang.org/t/new-features-on-lib-rs-website/98560

PR #440, opened 2023-08-27
chore(changelog): ignore alpha tags

orhun:
With this change, all the changes from alpha releases will be incorporated in the stable release changelog. Also, we will have all the entries in the latest alpha release instead of separating the changelog. For example, `alpha.2` changelog will contain changes from `alpha.1` and `alpha.0`. If we do not want to do that, I can tweak the release workflow accordingly.


joshka (approved):
Tested and this works exactly as promised

Issue #386, opened 2023-08-07
Docs: improve general methods documentation


Valentin271:
## Problem

There is not enough documentation in the widgets' methods (and other relative elements such as `Title` or `Padding`). While most of them are simple and require almost no documentation, some other can be ambiguous.

This was a major pain point for me while discovering the library (as a user and as a contributor).

## Solution

Add documentation

## Alternatives
_None_

## Additional context

I'm totally down to work on it, I just want to advertise this work for users and possibly interested contributors.

Elements missing documentation, grouped by widget :

- [x] `BarChart`, `Bar` (#449)
- [x] `Block`, `Padding`, `BorderType` (#469)
- [ ] `Chart`, `Axis`, `Dataset` (#696 )
- [x] `Gauge` (#514)
- [x] `List`, `ListState`, `ListItem` (#669)
- [x] `Paragraph` (#467)
- [x] `Sparkline`, `RenderDirection` (#648)
- [x] `Table` (#471, #638)
- [x] `Tabs` (#535)

There are other elements that need documentation, but I think they can be addressed later in an other issue.

---

Would it be useful to warn (and deny, once completed) on missing doc? See https://doc.rust-lang.org/rustdoc/lints.html#missing_docs

Valentin271:
Also, I think _documentation_ and maybe _good first issue_ are good labels here. Thanks

joshka:
>Would it be useful to warn (and deny, once completed) on missing doc? See https://doc.rust-lang.org/rustdoc/lints.html#missing_docs

Yes. Definitely! (Probably worth doing per file once they are documented well)

joshka:
P.S. I recommend using an LLM (e.g. OpenAI / Llama / Claude) to help make this quicker to implement. Obviously validate the output and don't just accept AI hallucinations as gospel :)

Perhaps add some general doc guidelines to this issue or point to some existing ones if you're aware of them.
Some key points I'd worry about are (please add some more if you have them):
- keep line length to 100 chars (to make it easy to maintain) - the [vscode rewrap extension](https://marketplace.visualstudio.com/items?itemName=stkb.rewrap) is good for this
- first line is summary, second line blank, third onwards is more detail.
- the main type doc comment should talk about the general features that the widget supports and introduce the concepts pointing to the various methods. Focus on interaction with various features and giving enough information that helps understand why you might want something. Also point out alternates
- any example code is as simple as possible

[...]

PR #443, opened 2023-08-27
docs(title): add documentation to title

Valentin271:
# Description

This adds documentation for `Title` and `Position`.

Relates to #386.

# Rendered documentation

image
image

### DIFF
--- a/src/title.rs
+++ b/src/title.rs
@@ -1,25 +1,93 @@
+#![warn(missing_docs)]
+//! This module holds the [`Title`] element and its related configuration types.
+//! A title is a piece of [`Block`](crate::widgets::Block) configuration.
+
 use strum::{Display, EnumString};
 
 use crate::{layout::Alignment, text::Line};
 
+/// A [`Block`](crate::widgets::Block) title.
+///
+/// It can be aligned (see [`Alignment`]) and positioned (see [`Position`]).
+///
+/// # Example
+///
+/// Title with no style.
+/// ```
+/// # use ratatui::widgets::block::Title;
+/// Title::from("Title");
+/// ```
+///
+/// Blue title on a white background (via [`Stylize`](crate::style::Stylize) trait).
+/// ```
+/// # use ratatui::widgets::block::Title;
+/// # use ratatui::style::Stylize;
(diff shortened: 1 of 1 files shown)


src/title.rs
```diff
@@ -1,25 +1,76 @@
+#![warn(missing_docs)]
```
orhun, line 1: If we want to go with this, let's add this to the higher modules rather than just here I would say.

Valentin271: In the long run I agree, but right now if we put it in `lib.rs`, clippy will complain in every pipeline.

How should we handle this? Progressively moving the warning as the documentation grows or just add it when everything is documented?

I think moving it progressively avoids regressions (eg. someone adding a feature in title and not documenting).

joshka: I agree with this logic use tools to help drive down the missing docs and keep it there.

Slapping this on lib.rs = 359 violations, so we're a ways off yet, but I'd bet there's a few modules that are either fully documented already or close to it that we could add it to.

src/title.rs
```diff
@@ -1,25 +1,76 @@
...
+//! This module holds the [Title] element and its related configuration types.
```
joshka, line 2: nit: (Across the entire PR - add backticks in the links to code items. e.g. 
```suggestion
//! This module holds the [`Title`] element and its related configuration types.
```

src/title.rs
```diff
@@ -1,25 +1,76 @@
...
 
 use crate::{layout::Alignment, text::Line};
 
+/// A block title
+///
+/// It can be aligned (see [Alignment]) and positioned (see [Position]).
+///
+/// # Example
+///
+/// ```
```
joshka, line 15: For the example it's worth noting that a title can have multiple Spans (because it takes a line), which will can each be formatted differently. This is particularly useful for highlighting shortcut keys. E.g. `Line::from(vec!["Q".white(), "uit".gray()]` might be a good example.

joshka (changes requested):

joshka (approved):

kdheepak (approved):

PR #446, opened 2023-08-28
Update README to use ratatui 

kdheepak:
In my opinion, this change in the docs will be a better user experience for most people, for the following reasons:

- With breaking changes users can't just do this to upgrade from `tui`:

```toml
[dependencies]
tui = { package = "ratatui", version = "0.23.0", features = ["all-widgets"]} # will cause errors because of breaking changes going forward
```

At best, we should only recommend this:

```toml
[dependencies]
tui = { package = "ratatui", version = "0.19.0", features = ["all-widgets"]}
```

But I think we should save that for the docs or the ratatui-book; and the README should have one obvious way for newcomers to add the `ratatui` crate to their project.

- Third party crates linked have migrated from `tui` to `ratatui`.

- If users use `ratatui` explicitly in the code, it'll will allow for searching for `ratatui` on github / gitlab etc to be easier going forward.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,15 +55,6 @@ Or modify your `Cargo.toml`
 ratatui = { version = "0.23.0", features = ["all-widgets"]}
 ```
 
-Ratatui is mostly backwards compatible with `tui-rs`. To migrate an existing project, it may be
-easier to rename the ratatui dependency to `tui` rather than updating every usage of the crate.
-E.g.:
-
-```toml
-[dependencies]
-tui = { package = "ratatui", version = "0.23.0", features = ["all-widgets"]}
-```
-
 ## Introduction
 
 `ratatui` is a terminal UI library that supports multiple backends:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -9,7 +9,7 @@ project.
 [![Crates.io](https://img.shields.io/crates/v/ratatui?logo=rust&style=flat-square)](https://crates.io/crates/ratatui)
 [![License](https://img.shields.io/crates/l/ratatui?style=flat-square)](./LICENSE) [![GitHub CI
 Status](https://img.shields.io/github/actions/workflow/status/ratatui-org/ratatui/ci.yml?style=flat-square&logo=github)](https://github.com/ratatui-org/ratatui/actions?query=workflow%3ACI+)
-[![Docs.rs](https://img.shields.io/docsrs/ratatui?logo=rust&style=flat-square)](https://docs.rs/crate/ratatui/)  
+[![Docs.rs](https://img.shields.io/docsrs/ratatui?logo=rust&style=flat-square)](https://docs.rs/crate/ratatui/)
```
orhun, line 12: Unrelated formatting change? 😅

orhun (approved):
+1

README.md
kdheepak: Resolved in https://github.com/ratatui-org/ratatui/pull/446/commits/905fe7a21cb2c9eebae20e210d76df941f53f8f8

I did turn off my markdown formatter for this PR but I have a git pre-commit hook that trims all whitespaces, and didn't catch that while `git add` + `git commit`ing :)

Valentin271 (approved):
+1 too

joshka (approved):

joshka:
@kdheepak looks like your gpg signing was broken. Pushing a rebase should in theory allow this to merge.

kdheepak:
Thanks for catching that! It looks like I'll need to set up a post-commit gpg signing check :)

PR #449, opened 2023-08-28
docs(barchart): add documentation to `BarChart`

Valentin271:
Add documentation to the `BarChart` widgets and its sub-modules.

Relates to #386.

<details><summary>Screenshots</summary>
<p>

image
image

</p>
</details>

### DIFF
--- a/src/widgets/barchart/bar.rs
+++ b/src/widgets/barchart/bar.rs
@@ -1,11 +1,19 @@
 use crate::{buffer::Buffer, prelude::Rect, style::Style, text::Line};
 
-/// represent a bar to be shown by the Barchart
+/// A bar to be shown by the [`BarChart`](crate::widgets::BarChart) widget.
 ///
-/// # Examples
-/// the following example creates a bar with the label "Bar 1", a value "10",
-/// red background and a white value foreground
+/// Here is an explanation of a `Bar`'s components.
+/// ```plain
+/// ███                          ┐
+/// █2█  <- text_value or value  │ bar
+/// foo  <- label                ┘
+/// ```
+/// Note that every element can be styled individually.
 ///
+/// # Example
+///
+/// The following example creates a bar with the label "Bar 1", a value "10",
+/// red background and a white value foreground.
 /// ```
 /// # use ratatui::{prelude::*, widgets::*};
 /// Bar::default()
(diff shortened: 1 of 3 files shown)


src/widgets/barchart/bar.rs
```diff
@@ -30,34 +30,66 @@ pub struct Bar<'a> {
 }
 
 impl<'a> Bar<'a> {
+    /// Set the value of this bar.
+    ///
+    /// The value will be displayed inside the bar.
+    ///
+    /// # See also
```
joshka, line 37: nice - I like the cross refs

src/widgets/barchart/bar.rs
```diff
@@ -1,11 +1,11 @@
 use crate::{buffer::Buffer, prelude::Rect, style::Style, text::Line};
 
-/// represent a bar to be shown by the Barchart
+/// Represent a bar to be shown by the [`BarChart`](crate::widgets::BarChart) widget.
```
joshka, line 3: Do you think it would be worth adding a small amount of detail about the components of each bar to unambiguously differentiate the value, label, and text_value?

src/widgets/barchart/bar.rs
```diff
@@ -30,34 +30,66 @@ pub struct Bar<'a> {
...
+    /// [`Bar::value`] to set the value.
     pub fn value_style(mut self, style: Style) -> Bar<'a> {
         self.value_style = style;
         self
     }
 
-    /// set the text value printed in the bar. (By default self.value is printed)
+    /// Set the text value printed in the bar.
+    ///
+    /// By default, `value` is printed.
```
joshka, line 79:
```suggestion
    /// If `text_value` is not set, then the display representation of `value` will be shown on the bar.
```

src/widgets/barchart/bar.rs
joshka, line 3: Nit:
```suggestion
/// Represents a bar to be shown by the [`BarChart`](crate::widgets::BarChart) widget.
```
or
```suggestion
/// A bar to be shown by the [`BarChart`](crate::widgets::BarChart) widget.
```

src/widgets/barchart/bar_group.rs
```diff
@@ -5,7 +5,7 @@ use crate::{
     text::Line,
 };
 
-/// represent a group of bars to be shown by the Barchart
+/// Represent a group of bars to be shown by the Barchart.
```
joshka, line 8:
```suggestion
/// A group of bars to be shown by the Barchart.
```

src/widgets/barchart/bar_group.rs
```diff
@@ -35,7 +35,7 @@ impl<'a> BarGroup<'a> {
         self
     }
 
-    /// return the maximum bar value of this group
+    /// Return the maximum bar value of this group
```
joshka, line 38:
```suggestion
    /// The maximum bar value of this group
```

src/widgets/barchart/mod.rs
```diff
@@ -8,14 +9,16 @@ pub use bar_group::BarGroup;
 
 use super::{Block, Widget};
 
-/// Display multiple bars in a single widgets
+/// Display multiple [bars](Bar) in a single widget.
```
joshka, line 12: Here do you think it would be good to add a small amount of narrative information about the chart, groups, and bars?

src/widgets/barchart/mod.rs
```diff
@@ -97,66 +102,157 @@ impl<'a> BarChart<'a> {
...
 
+    /// Surround the [`BarChart`] with a [`Block`].
     pub fn block(mut self, block: Block<'a>) -> BarChart<'a> {
         self.block = Some(block);
         self
     }
 
+    /// Set the value necessary for a [`Bar`] to reach the maximum height.
+    ///
+    /// If not set, the maximum value in the data is taken as reference.
```
joshka, line 113: This might be good info to add to the doc comment for the type.

joshka (changes requested):
Generally looking good (none of my feedback is blocking - purely gold plating).

Would you consider adding examples even for the simple methods? My rationale for this is that examples help newer users discover things like how to write styles, do things idiomatically (e.g. `Style::new().red()` or `"foo".into()`), and discover other Ratatui modules (like the `symbols` module

src/widgets/barchart/mod.rs
Valentin271: Yep, I'll do.

A VHS of the example could be a good fit here. Maybe it's worth opening some sort of RFC or a discussion about it?

src/widgets/barchart/mod.rs
Valentin271: Tried to include it but didn't really see how.
I'm not sure this is really relevant for the majority of workflows. It can still be found while searching the `BarChart` API for advanced usage.
Please let me know if you think otherwise.

src/widgets/barchart/bar.rs
Valentin271: Yes it could be useful, I did a little schema to explain that:

image

Valentin271:
Most of it should be fixed. Here's an updated screenshot of the `BarChart` type documentation:

image

src/widgets/barchart/mod.rs
```diff
@@ -8,14 +9,28 @@ pub use bar_group::BarGroup;
 
 use super::{Block, Widget};
 
-/// Display multiple bars in a single widgets
+/// A chart showing values as [bars](Bar).
```
karthago1, line 12: Does it make sense to show somewhere a complete group chart like the following:
```
┌─────────────────────────────────┐
│                             ████│
│                             ████│
│                        ▅▅▅▅ ████│
│            ▇▇▇▇        ████ ████│
│            ████ ████   ████ ████│
│     ▄▄▄▄   ████ ████   ████ ████│
│▆10▆ 20M█   █50█ █40█   █60█ █90█│
│ C1   C2     C1   C2     C1   C2 │
│ Group1      Group2      Group3  │
└─────────────────────────────────┘
```
So that the reader knows directly, what is possible to do.
(I am not sure if this is the right place)

karthago1 (commented):
Hey I like the new documentation, I could never write such a nice one.

src/widgets/barchart/mod.rs
Valentin271: I think either a textual chart or a VHS is useful.

I can add a textual representation now and we'll decide later if we replace them with VHS.

Valentin271: Added in 09c974bce45c368ba7408c8efa729db324dec795.

I basically took your example and just made it a bit smaller.

<details><summary>Screenshot</summary>
<p>

image

</p>
</details>

karthago1 (approved):
Nice work. (approved, but I am not good at reviewing docs :grin: )

joshka (approved):

src/widgets/barchart/mod.rs
joshka: I was thinking that at some point it might be nice to make each example have a mode that displays just a small render of each feature that can be referenced from the docs easily.

PR #453, opened 2023-08-30
Support for Test-Driven Development

SirRegion:
Test-Driven Development (TDD) is widely recognized as a good way to develop software. 

Sadly, using TDD with this library has been impossible, because there is no good way to assert on the state of the UI.
To fix this, I started work on some getters that return the current state of some widgets and help with creating small and maintainable unit tests.

(See also this discussion I started a while back: https://github.com/ratatui-org/ratatui/discussions/78)

I hid these changes behind a feature flag called "unstable-testing-mode", because I'm not quite certain if this is the best way to implement this feature.  

This feature is also by no means complete yet. It's just here to gather feedback and to (maybe) get some help implementing it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,7 +57,7 @@ rand = "0.8"
 pretty_assertions = "1.4.0"
 
 [features]
-default = ["crossterm"]
+default = ["crossterm", "unstable-testing-mode"]
 #! The following optional features are available for all backends:
 ## enables serialization and deserialization of style and color types using the [Serde crate].
 ## This is useful if you want to save themes to a file.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -74,6 +74,9 @@ all-widgets = ["widget-calendar"]
 ## enables the [`calendar`] widget module and adds a dependency on the [Time crate].
 widget-calendar = ["dep:time"]
 
+## Adds a unstable API for unit testing ratatui. THIS MAY BREAK AT ANY TIME!
+unstable-testing-mode = []
+
 [package.metadata.docs.rs]
 all-features = true
 # see https://doc.rust-lang.org/nightly/rustdoc/scraped-examples.html
--- a/src/text/line.rs
+++ b/src/text/line.rs
@@ -149,6 +149,11 @@ impl<'a> Line<'a> {
             ..self
         }
     }
+    
+    #[cfg(feature = "unstable-testing-mode")]
+    pub fn get_spans(&self) -> Vec<Span>{
+        self.spans.clone()
+    }
 }
 
 impl<'a> From<String> for Line<'a> {
--- a/src/text/line.rs
+++ b/src/text/line.rs
@@ -156,6 +156,14 @@ impl<'a> Line<'a> {
     }
 }
 
+impl<'a> ToString for Line<'a> {
+    fn to_string(&self) -> String {
+        self.spans.iter().fold("".to_string(), |acc, span: &Span| {
+            acc + &span.content
+        })
+    }
+}
+
 impl<'a> From<String> for Line<'a> {
     fn from(s: String) -> Self {
         Self::from(vec![Span::from(s)])
--- a/src/text/span.rs
+++ b/src/text/span.rs
@@ -167,6 +167,12 @@ impl<'a> Span<'a> {
     pub fn reset_style(&mut self) {
         self.patch_style(Style::reset());
     }
+
+    #[cfg(feature = "unstable-testing-mode")]
+    pub fn get_content(&self) -> String{
+        self.content.to_string()
+    }
+
 }
 
 impl<'a, T> From<T> for Span<'a>
--- a/src/text/text.rs
+++ b/src/text/text.rs
@@ -147,6 +147,15 @@ impl<'a> Text<'a> {
             line.reset_style();
         }
     }
+    #[cfg(feature = "unstable-testing-mode")]
+    pub fn get_text_content(&self) -> String{
+        self.lines.iter().flat_map(|line: &Line| {
+            let lines: Vec<String> = line.get_spans().iter().map(|span: &Span| {
+                span.get_content()
+            }).collect();
+            lines
+        }).collect()
+    }
 }
 
 impl<'a> From<String> for Text<'a> {
--- a/src/text/text.rs
+++ b/src/text/text.rs
@@ -147,8 +147,13 @@ impl<'a> Text<'a> {
             line.reset_style();
         }
     }
-    #[cfg(feature = "unstable-testing-mode")]
-    pub fn get_text_content(&self) -> String{
+    
+    
+}
+
+#[cfg(feature = "unstable-testing-mode")]
+impl<'a> ToString for Text<'a> {
+    fn to_string(&self) -> String{
         self.lines.iter().flat_map(|line: &Line| {
             let lines: Vec<String> = line.get_spans().iter().map(|span: &Span| {
                 span.get_content()
--- a/src/text/text.rs
+++ b/src/text/text.rs
@@ -154,12 +154,9 @@ impl<'a> Text<'a> {
 #[cfg(feature = "unstable-testing-mode")]
 impl<'a> ToString for Text<'a> {
     fn to_string(&self) -> String{
-        self.lines.iter().flat_map(|line: &Line| {
-            let lines: Vec<String> = line.get_spans().iter().map(|span: &Span| {
-                span.get_content()
-            }).collect();
-            lines
-        }).collect()
+        self.lines.iter().fold("".to_string(), |acc: String, line: &Line| {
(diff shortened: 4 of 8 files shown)


joshka:
Hey, this seems like a lot of extra code baggage to carry around, and I'm not seeing the obvious benefit (mainly as although it's supposed to support TDD, there are no tests that show that it supports TDD).

The fields of each widget are kind of like parameter objects to the render widget call. Their primary purpose is not behavioral, but just to provide data for the render function. So generally the field methods should store the value passed in and be done with it (there are a few exceptions to this). This suggests to me that the main value in testing would be in testing the result of the render calls rather than worrying about what the fields hold.

Can we take a step back from the implementation for a minute and look at a vertical piece (i.e. a single widget / area of your app) of the problem you're solving with this? What do the tests that you want to write for an application look like?

SirRegion:
Thanks for your reply @joshka!

My usecase for this is something like this:

My application should show some tabs for some categories of data defined in a config file.
For this, I wrote the function `get_category_tabs`:
```
fn get_category_tabs<'a>(categories: &Vec<Category>, current_tab: usize) -> Tabs {
    let titles: Vec<Line> = categories
        .iter()
        .map(|category| Line::from(vec![Span::styled(&category.title, Style::default().fg(Color::Cyan))]))
        .collect();

    let tabs: Tabs = Tabs::new(titles)
        .block(Block::default().borders(Borders::ALL).title("Categories"))
        .select(current_tab)
        .style(Style::default())
        .highlight_style(Style::default().add_modifier(Modifier::BOLD).bg(Color::Yellow));

    return tabs;
}
```

Testing the output of the render call is quite complex and brittle, because I would need to either
- mock a console window and assert its exact state
- use a real console window and check each character.
This is a lot more overhead than I need in my tests, because I trust ratatuis rendering functions, I just want to assert that my code passes the correct values to the renderer.

So instead I test the state of the returned tabs object like this:
```
  #[test]
    fn categories_tabs() {
        let categories = vec![
            Category {
                title: "title 1".to_string(),
                actions: vec![],
            },
            Category {
                title: "title 2".to_string(),
                actions: vec![],
            },
        ];
        let tabs = get_category_tabs(&categories, 0);

        let tab1 = tabs.get_titles()[0].to_string();
        let tab2 = tabs.get_titles()[1].to_string();

        assert_eq!(tab1, "title 1");
        assert_eq!(tab2, "title 2");
    }
```
While this test is quite basic, it gives me confidence that the tabs are rendered with the correct titles everytime, even if I refactor some functions.

I didn't add tests for my library code, because I felt like adding tests for getters would be a lot of overhead. The idea with the derive macro is great, I will look into that.

I hope I cleared up some things, but feel free to ask more questions :)

joshka:
Something I've been pushing for more of internally is to write our unit tests for widgets like the following to cut out the backend etc:

```rust
#[test]
fn categories_tabs() {
    let categories = vec![
        Category {
            title: "title 1".to_string(),
            actions: vec![],
        },
        Category {
            title: "title 2".to_string(),
            actions: vec![],
        },
    ];
    let tabs = get_category_tabs(&categories, 0);

    let mut buffer = Buffer::empty(Rect::new(0, 0, 30, 3));
    tabs.render(buffer.area, &mut buffer);

    let mut expected = Buffer::with_lines(vec![
        "┌Categories──────────────────┐",
        "│ title 1 │ title 3          │",
        "└────────────────────────────┘",
    ]);
    expected.set_style(
        Rect::new(2, 1, 7, 1),
        Style::new().cyan().on_yellow().bold(),
    );
    expected.set_style(
        Rect::new(12, 1, 7, 1),
        Style::new().cyan().on_yellow().bold(),
    );
}
```

The output is useful for finding when there is an error in the test (e.g. title 3 and the second background color is different in the test):
```
expected: Buffer {
    area: Rect { x: 0, y: 0, width: 30, height: 3 },
    content: [
        "┌Categories──────────────────┐",
        "│ title 1 │ title 3          │",
        "└────────────────────────────┘",
    ],
    styles: [
        x: 0, y: 0, fg: Reset, bg: Reset, underline: Reset, modifier: NONE,
        x: 2, y: 1, fg: Cyan, bg: Yellow, underline: Reset, modifier: BOLD,
        x: 9, y: 1, fg: Reset, bg: Reset, underline: Reset, modifier: NONE,
        x: 12, y: 1, fg: Cyan, bg: Yellow, underline: Reset, modifier: BOLD,
        x: 19, y: 1, fg: Reset, bg: Reset, underline: Reset, modifier: NONE,
    ]
}
actual: Buffer {
    area: Rect { x: 0, y: 0, width: 30, height: 3 },
    content: [
        "┌Categories──────────────────┐",
        "│ title 1 │ title 2          │",
        "└────────────────────────────┘",
    ],
    styles: [
        x: 0, y: 0, fg: Reset, bg: Reset, underline: Reset, modifier: NONE,
        x: 2, y: 1, fg: Cyan, bg: Yellow, underline: Reset, modifier: BOLD,
        x: 9, y: 1, fg: Reset, bg: Reset, underline: Reset, modifier: NONE,
        x: 12, y: 1, fg: Cyan, bg: Reset, underline: Reset, modifier: NONE,
        x: 19, y: 1, fg: Reset, bg: Reset, underline: Reset, modifier: NONE,
    ]
}
diff:
0: at (12, 1)
  expected: Cell { symbol: "t", fg: Cyan, bg: Yellow, underline_color: Reset, modifier: BOLD, skip: false }
  actual:   Cell { symbol: "t", fg: Cyan, bg: Reset, underline_color: Reset, modifier: NONE, skip: false }
...
```

You might wonder why the `content` is displayed like an array of strings like this. It feeds pretty well into a Red Green Refactor flow. Most of the time when I write unit tests for ratatui, I'll write a stub of what will end up in the expected buffer and then just copy the actual rendering out to the test. It's not usually much more than 10-15 seconds. I'd like it to be shorter though.

There's some ergonomics missing from this that would be nice to speed that flow up perhaps:
- A way to write the buffer state and formatting in one line in a way that still makes it visually easy to detect
- More concise error reporting / perhaps something similar to pretty_assertions that does color?
- A reduction in the extraneous noise - perhaps switching `Buffer::with_lines(vec![...])` to something a little like:
```rust
let selected_style = Style::new().cyan().on_yellow().bold();
let normal_style = Style::new().cyan();
assert_widget_renders!(
    tabs,
    [
        "┌Categories──────────────────┐",
        "│ title 1 │ title 2          │",
        "└────────────────────────────┘",
    ],
    [
        (Rect::new(2, 1, 7, 1), selected_style),
        (Rect::new(12, 1, 7, 1), normal_style),
    ]
);
```

Imagine that this is the output* of the test - would that solve most of your TDD issues?

* (the styles could probably be collapsed to just the unique styles and spat out as `style_1`, `style_2` etc.)

---

Ideally with a derive macro approach we'd find a way to reduce a bunch of code for both setters and getters (i.e. perhaps all the existing builder functions could be replaced with some macro magic, which would make this something we can implement across all the widgets at once)

imagine:
```rust
#[derive(..., WidgetBuilders]
#[cfg_attr(test, derive(WidgetGetters)]
struct Tabs {
// unchanged (doc comments moved to properties from functions that are removed
}

impl Tabs {
  fn new() {
  // unchanged
  }

  // every builder function is deleted
}

**Comment by SirRegion** [comment]
I do like the way you propose to structure unit tests for ratatui, but I think we should split our discussion between unit tests internal to ratatui (that need to assert the exact render output) and tests external to ratatui (that trust the renderer and just want to assert that some properties are set correctly). 

My main issue with these sort of tests is the need of a "correct" border around the tabs. Because I trust ratatui and I don't want to test the renderer itself, I don't need to care if the Tabs look like this:
```rust
        "┌Categories──────────────────┐",
        "│ title 1 │ title 2          │",
        "└────────────────────────────┘",
``` 
or like this: 
```rust
        "┌Categories──────────────┐",
        "│ title 1 │ title 2      │",
        "└────────────────────────┘",
```

It is nice to have the option though, and it's definitely the correct / best way to test the render function output.  

The elegance of "my" way of testing is being able to just find out one field of the widget, instead of having to check each property individually.

So as long as i don't care about the exact rendering, a test like the one I proposed above is a lot more stable.

---

I think the derive macros should have their own disscussion, but I really like the way you are going with here.
 Sadly, my rust knowledge is not close to being advanced enough to implement a macro like this :/

---

BTW, i really apprechiate the way this conversation is going, if I have the time in the future I will definitely contribute more to ratatui.

joshka:
There's some related discussion on this in https://github.com/ratatui-org/ratatui/issues/619

SirRegion:
Thanks for the notification @joshka!

Feel free to close this merge request if #619 gets merged. It's aim is basically the same, but the scope isn't as broad and the implementation is actually getting worked on. (I sadly don't have a lot of time for working on this right now)

joshka:
Bad bot :D

PR #454, opened 2023-08-30
chore: check documentation lint

### LINKED ISSUE #451: Check documentation build
## Problem

Currently there is no documentation **build** check.
This leads to some small mistakes being in the documentation (see #450).

## Solution

We should add a documentation build check to use by `cargo make ci` and the pipelines.

This can also prevent documentation build from failing in `docs.rs`.

Valentin271:
# Description

Adds a `check-doc` job to the Makefile. This job checks for errors and warnings in the doc-comments.

# Makefile structure

~I've added a meta-job `check` that runs `check-code` and `check-doc`.~
~The `check` job is run in CI so now the pipelines would check for code and documentation.~

~`check-code` replaces the old `check` job.~

~For me this is not a breaking change as it concerns tooling and not public APIs. I've also checked every place the `cargo check` job could have been referenced.~

# Job implementation

I've used an unstable Rustdoc flag: `--check`, [see here](https://doc.rust-lang.org/rustdoc/unstable-features.html#--check-only-checks-the-documentation). Here is the [tracking issue](https://github.com/rust-lang/cargo/issues/10025).

I've used [`cargo rustdoc`](https://doc.rust-lang.org/cargo/commands/cargo-rustdoc.html) instead of `cargo doc` to pass flags down to `rustdoc`.
I didn't use the env variable `RUSTDOCFLAGS` because `cargo-make` [doesn't cleanup task scoped env after execution](https://github.com/sagiegurari/cargo-make#task), so it would interfere with other jobs.

---

Fixes #451

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,6 +53,8 @@ jobs:
         uses: taiki-e/install-action@cargo-make
       - name: Check formatting
         run: cargo make fmt
+      - name: Check documentation
+        run: cargo make check-doc
 
   clippy:
     runs-on: ubuntu-latest
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -21,7 +21,7 @@ dependencies = [
 
 [tasks.style-check]
 description = "Check code style"
-dependencies = ["fmt", "typos"]
+dependencies = ["fmt", "typos", "check-doc"]
 
 [tasks.fmt]
 description = "Format source code"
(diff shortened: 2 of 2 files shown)


Valentin271:
> I've added a meta-job `check` that runs `check-code` and `check-doc`.
> The `check` job is run in CI so now the pipelines would check for code and documentation.
> 
> `check-code` replaces the old `check` job.
> 
> For me this is not a breaking change as it concerns tooling and not public APIs. I've also checked every place the `cargo check` job could have been referenced.

This was failing in the pipeline since the `check` pipeline wasn't run with nightly.

I think putting `check-doc` makes more sense in the `lint` pipeline (`style-check` Makefile job).
It is already running nightly and only runs on ubuntu which is enough for checking documentation.

joshka (approved):

joshka:
Thanks for the PR!

PR #455, opened 2023-08-30
test(barchart): add benchmarks

Valentin271:
Adds benchmarks for the `BarChart` widget.

Relates to #137.

Here is the resulting graph.
image

### DIFF
--- a/benches/barchart.rs
+++ b/benches/barchart.rs
@@ -1,0 +1,73 @@
+use criterion::{criterion_group, criterion_main, Bencher, BenchmarkId, Criterion};
+use rand::Rng;
+use ratatui::{
+    buffer::Buffer,
+    layout::Rect,
+    prelude::Direction,
+    widgets::{Bar, BarChart, BarGroup, Widget},
+};
+
+/// Benchmark for rendering a barchart.
+pub fn barchart(c: &mut Criterion) {
+    let mut group = c.benchmark_group("barchart");
+    let mut rng = rand::thread_rng();
+
+    for data_count in [64, 256, 2048] {
+        let data: Vec<Bar> = (0..data_count)
+            .map(|i| {
+                Bar::default()
+                    .label(format!("B{i}").into())
+                    .value(rng.gen_range(0..data_count))
+            })
+            .collect();
+
+        let bargroup = BarGroup::default().bars(&data);
+
+        // Render a basic barchart
+        group.bench_with_input(
+            BenchmarkId::new("render", data_count),
+            &BarChart::default().data(bargroup.clone()),
+            render,
+        );
+
+        // Render an horizontal barchart
+        group.bench_with_input(
+            BenchmarkId::new("render_horizontal", data_count),
+            &BarChart::default()
+                .direction(Direction::Horizontal)
+                .data(bargroup.clone()),
+            render,
+        );
+
+        // Render a barchart with multiple groups
+        group.bench_with_input(
+            BenchmarkId::new("render_grouped", data_count),
+            &BarChart::default()
+                // We call `data` multiple time to add multiple groups.
+                // This is not a duplicated call.
+                .data(bargroup.clone())
+                .data(bargroup.clone())
+                .data(bargroup.clone()),
+            render,
+        );
+    }
+
+    group.finish();
+}
+
+/// Render the widget in a classical size buffer
+fn render(bencher: &mut Bencher, barchart: &BarChart) {
+    let mut buffer = Buffer::empty(Rect::new(0, 0, 200, 50));
+    // We use `iter_batched` to clone the value in the setup function.
+    // See https://github.com/ratatui-org/ratatui/pull/377.
+    bencher.iter_batched(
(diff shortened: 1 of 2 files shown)


benches/barchart.rs
```diff
@@ -0,0 +1,63 @@
...
+
+        // Render a barchart with multiple groups
+        group.bench_with_input(
+            BenchmarkId::new("render_grouped", data_count),
+            &BarChart::default()
+                // We call `data` multiple time to add multiple groups.
+                // This is not a duplicated call.
+                .data(bargroup.clone())
+                .data(bargroup.clone())
+                .data(bargroup.clone()),
```
karthago1, line 40: When I saw the result, I was expecting, that `barchart/render/64` and `barchart/render_grouped/64` contain both 64 bars for comparison. but the `render_grouped` has three time more bars. I think we should pass third of `bargroup` and not the full array

benches/barchart.rs
```diff
@@ -0,0 +1,63 @@
...
+            render,
+        );
+    }
+
+    group.finish();
+}
+
+/// Render the widget in a classical size buffer
+fn render(bencher: &mut Bencher, barchart: &BarChart) {
+    let mut buffer = Buffer::empty(Rect::new(0, 0, 200, 50));
```
karthago1, line 50: I am not sure if the measurement of `256` and `2048` bars are accurate. I think maybe we should make the buffer bigger. (for 2048 bars I would reserve at least `2048 * 2 = 4096`, because of the `bar_gap`. maybe a little bit more)

benches/barchart.rs
```diff
@@ -0,0 +1,63 @@
...
+    // See https://github.com/ratatui-org/ratatui/pull/377.
+    bencher.iter_batched(
+        || barchart.clone(),
+        |bench_barchart| {
+            bench_barchart.render(buffer.area, &mut buffer);
+        },
+        criterion::BatchSize::LargeInput,
+    )
+}
+
```
karthago1, line 61: I think it will be also very interesting to benchmark the horizontal `BarChart`. I expect it to be faster than the vertical one, because we have only one loop which renders everything (maybe the rendering is a little bit more simpler also)

I don't know if you have time to include it.

karthago1 (commented):
thank you for the measurements, I always wondered how fast is the rendering.

benches/barchart.rs
joshka: Most of the value we get out of the benchmarks is when they find things that are hideously excessively inefficient in normal usage, and so we test against screen sizes of similar amounts, with 200x50 being a readable size for a 16in laptop. I'd imagine that reasonable resolution goes up significantly on wide screens / 8K screens though, so perhaps it's worth picking a standard testing resolution that larger screens into account. Thoughts?

We also currently have a hard limit on cells of 65535 (i.e. u16::MAX - I'm unsure what the original limitation for this constraint was, but it's perhaps worth revisiting).

karthago1: true. It is just that I misunderstood the result as: how much it takes to render `64`, `246` and `2048` bars.

benches/barchart.rs
Valentin271: Here 64 means **datasets** of 64 elements. But the meaning is not really important, the numbers are just to satisfy criterion which needs them to display the result as a table :

image

I think passing a third of the array kind of defeats the purpose: you are then comparing the same thing with `barchart/render/64` and `barchart/render_grouped/64`.

This bench (imo) should be able to show any difference between using one or multiple datasets. In addition, I think it's good to mimic a "real world usage", thus, more datasets doesn't necessarily means less bar (in each dataset).

Hope this was clear enough, let me know your thoughts :)

benches/barchart.rs
Valentin271: It is indeed faster! I didn't think it rendered so differently so I didn't include it right away.

_NOTE: It also "renders" less bars into the buffer (since there is less space) so that might play a role._

image

benches/barchart.rs
Valentin271: > I am not sure if the measurement of 256 and 2048 bars are accurate.

I think they are, they're just not very explicit. They don't test the render of 2048 bars since there is not enough space in the buffer. It's more of "test the computation of 2048 bars", so the results are quite good, even with large datasets, it's pretty fast to compute and then render a subset.

> 200x50 being a readable size for a 16in laptop

That was my first logic, yes. But actually you're right, we should test this with the maximum available resolution (or at least maximum reasonable resolution). Otherwise we would only bench for a subset of the lib users.

> We also currently have a hard limit on cells of 65535

We'd have to test in a real world scenario I think, render a fullscreen terminal on an 8k display and see if it fits.

benches/barchart.rs
karthago1: thanks for the explanation . just resolve all issues  :smile_cat:

benches/barchart.rs
karthago1: thanks

joshka (approved):

Issue #414, opened 2023-08-18
Remove builder pattern from position and content_length methods in ScrollbarState


fritzrehde:
## Problem

Most uses of the `ScrollbarState` include setting the scroll position (anytime the user scrolls, if `positions` is used for that instead of `prev` and `next` aren't used) and setting the `content_length`. Currently, these two functions have the signature:
```
pub fn position(self, position: u16) -> Self
```
and vice versa. This ressembles the builder pattern, which enables the user to write:
```
app.vertical_scroll_state = app.vertical_scroll_state
    .position(app.vertical_scroll_index as u16)
    .content_length(text.len() as u16);
```
However, I find that most use-cases, including those presented in https://docs.rs/ratatui/latest/src/scrollbar/scrollbar.rs.html, do not utilize the builder pattern. Instead, they come in the form:
https://docs.rs/ratatui/latest/src/scrollbar/scrollbar.rs.html#71:
```
app.vertical_scroll_state = app
                            .vertical_scroll_state
                            .position(app.vertical_scroll as u16);
```
or https://docs.rs/ratatui/latest/src/scrollbar/scrollbar.rs.html#155:
```
    app.vertical_scroll_state = app.vertical_scroll_state.content_length(text.len() as u16);
```

I do not think the builder pattern makes much sense here. I am writing this issue, because I accidentally wrote:
```
app.vertical_scroll_state.content_length(text.len() as u16);
```
This seemed like the intuitive Rust way to call a setter. But this doesn't update the value correctly due to the builder pattern being required, so I was suprised to see nothing change in my TUI.

## Solution

I propose to replace the builder pattern API detailed above with the following API for both position and content_length:
```
pub fn position(&mut self, position: u16)
```
This makes it more clear that we are setting/updating values inside an existing `ScrollbarState` item, and prevents the programming mistake that I made.
I'm not sure about the general ratatui/tui-rs convetions when it comes to the builder pattern, I know it is heavily used in this crate, and I think it works very well in many cases, but I think it gets in the way and leads to unintuitive code for this specific case.

joshka:
Thanks for this - I think I can see that there are not the best ergonomics for this, and I think I was probably wrong to suggest that the scrollbar actually needed to be implemented as a stateful widget rather than just as the three builder values directly on the Widget. Come give some feedback on https://github.com/ratatui-org/ratatui/issues/174 if you can.

kdheepak:
[...]

Issue #413, opened 2023-08-18
Scrollbar without arrows


fritzrehde:
## Description

I am not sure if this is the expected behaviour of the Scrollbar feature, so I'm asking here.
I know that the Scrollbar features arrows by default at the moment, but I would like to remove them.
Therefore, I have the following code:
```
        frame.render_stateful_widget(
            Scrollbar::default()
                .orientation(ScrollbarOrientation::VerticalRight)
                .begin_symbol(None)
                .end_symbol(None),
            area,
            &mut self.help_menu.vertical_scroll_state,
        );
```
However, this produces:
```
┌help────────────────────█
│                        █
│                        ║
│                        ║
└────────────────────────║
```
The issue here is that the scrollbar also takes up the corners of the block's borders. I would prefer to have normal corners, and the scrollbar take up everything besides the corners. I read through the #228 and saw this exact issue was mentioned somewhere and temporarily implemented with `.arrows(None)` or similar, but that was later removed.
My question: Is there a way to achieve my desired functionality?

## To Reproduce

The code I wrote above.

## Expected behavior

Achieving this somehow:
```
┌help────────────────────┐
│                        █
│                        ║
│                        ║
└────────────────────────┘
```

## Environment

- OS: Arch Linux
- Terminal Emulator: Alacritty
- Font: -
- Crate version: 0.22.0
- Backend: crossterm

joshka:
I think this may be as simple as rendering the scrollbar into the area that doesn't include the corners (i.e. `Rect { y: area.y+1, height: area.height - 2, ..area }` (and turning off the arrows like you have already).

P.S. just a heads up that there's a small breaking change in the next release in https://github.com/ratatui-org/ratatui/pull/330. It's available in the alpha release if you want to test your code early.

kdheepak:
Using a vertical `Margin` of `1` will do the trick:

```diff
        frame.render_stateful_widget(
            Scrollbar::default()
                .orientation(ScrollbarOrientation::VerticalRight)
                .begin_symbol(None)
                .end_symbol(None),
-            area,
+            area.inner(&Margin {vertical: 1, horizontal: 0}),
            &mut self.help_menu.vertical_scroll_state,
        );
```

fritzrehde:
Ok, that makes sense. One more question regarding margins: I have this rendering code:
```
    pub fn render<B: Backend>(&mut self, frame: &mut Frame<B>) {
        if self.is_shown {
[...]

PR #456, opened 2023-08-30
Better ergonomics for ScrollbarState and improved documentation 

kdheepak:
This PR improves the documentation and adds a `new` constructor function to incentivize users to create a `ScrollbarState` with the appropriate `content_length` value.

Related to https://github.com/ratatui-org/ratatui/issues/414#issuecomment-1699801843

### DIFF
--- a/src/widgets/scrollbar.rs
+++ b/src/widgets/scrollbar.rs
@@ -20,10 +20,16 @@ pub enum ScrollDirection {
 
 /// A struct representing the state of a Scrollbar widget.
 ///
+/// # Important
+///
+/// It's essential to set the `content_length` field when using this struct. This field
+/// represents the total length of the scrollable content. The default value is zero
+/// which will result in the Scrollbar not rendering.
+///
 /// For example, in the following list, assume there are 4 bullet points:
 ///
-/// - the `position` is 0
 /// - the `content_length` is 4
+/// - the `position` is 0
 /// - the `viewport_content_length` is 2
 ///
 /// ```text
@@ -39,15 +45,22 @@ pub enum ScrollDirection {
 /// default of 0 and it'll use the track size as a `viewport_content_length`.
 #[derive(Debug, Default, Clone, Copy, Eq, PartialEq, Hash)]
 pub struct ScrollbarState {
-    // The current position within the scrollable content.
-    position: u16,
     // The total length of the scrollable content.
     content_length: u16,
+    // The current position within the scrollable content.
+    position: u16,
     // The length of content in current viewport.
     viewport_content_length: u16,
 }
 
 impl ScrollbarState {
+    /// Constructs a new ScrollbarState with the specified content length.
+    pub fn new(content_length: u16) -> Self {
+        Self {
+            content_length,
+            ..Default::default()
+        }
+    }
     /// Sets the scroll position of the scrollbar and returns the modified ScrollbarState.
     pub fn position(mut self, position: u16) -> Self {
         self.position = position;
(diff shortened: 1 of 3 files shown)


mindoodoo (approved):

joshka (approved):

joshka (commented):
Can you please add examples?

kdheepak:
Do you think we want examples in the docstrings or in the `examples/scrollbar.rs` file?

In `examples/scrollbar.rs`, we use the default constructor and set the `content_length` during render, which is the only time at which we know the number of items (i.e. `content_length`) that the scrollbar is tracking. So it's a little awkward to change that to use this `new` constructor.

joshka:
> Do you think we want examples in the docstrings or in the `examples/scrollbar.rs` file?

Just the doc strings should be fine. Neither `ScrollBar` or `ScrollBarState` currently have type level comments that show a simple example of how to use the ScrollBar and State.

src/widgets/scrollbar.rs
```diff
@@ -124,6 +137,21 @@ pub enum ScrollbarOrientation {
...
+/// # use ratatui::widgets::*;
+/// let items = vec![Line::from("Item 1"), Line::from("Item 2"), Line::from("Item 3")];
+/// let vertical_scroll = 0;
+/// let paragraph = Paragraph::new(items.clone()).scroll((vertical_scroll, 0));
+/// let scrollbar = Scrollbar::default()
+///     .orientation(ScrollbarOrientation::VerticalRight)
+///     .begin_symbol(Some("↑"))
+///     .end_symbol(Some("↓"));
+/// let scrollbar_state = ScrollbarState::new(items.iter().len() as u16).position(vertical_scroll);
+/// ```
```
joshka, line 154: What about making this show the render as well? That's tripped up users previously (see #413).
(side note: perhaps the position etc. should be usize not u16?)

```suggestion
/// ```rust
/// # use ratatui::prelude::*;
/// # use ratatui::widgets::*;
/// # fn render_paragraph_with_scrollbar<B: Backend>(frame: &mut Frame<B>, area: Rect) {
/// let items = vec![
///     Line::from("Item 1"),
///     Line::from("Item 2"),
///     Line::from("Item 3"),
/// ];
/// let vertical_scroll = 0; // this would be passed from the app
/// let scrollbar = Scrollbar::default()
///     .orientation(ScrollbarOrientation::VerticalRight)
///     .begin_symbol(Some("↑"))
///     .end_symbol(Some("↓"));
/// let mut scrollbar_state = ScrollbarState::new(items.len() as u16).position(vertical_scroll);
///
/// let paragraph = Paragraph::new(items)
///     .block(Block::new().borders(Borders::RIGHT)) // to hold the scrollbar
///     .scroll((vertical_scroll, 0));
///
/// frame.render_widget(paragraph, area);
/// frame.render_stateful_widget(scrollbar, area, &mut scrollbar_state);
/// # }
/// ```
```

kdheepak: I didn't realize you could put it in a function with a doc comment like that, that's neat. 

I actually considered it, but saw that all other widgets didn't have any rendering examples. I've added it in now though.

kdheepak: > side note: perhaps the position etc. should be usize not u16?

I think so too. All `position`, `content_length` and `viewport_content_length` can be greater than the max u16 value. This would be a breaking change, would it not?

joshka (approved):

src/widgets/scrollbar.rs
joshka: yep

Issue #401, opened 2023-08-12
Move all demo gifs to github


kdheepak:
## Problem

Currently a lot of the demo gifs are on vhs's website. For reliability, we might want to consider moving them to github. 

## Solution

We don't want to bloat the repo git history. So we can either:

1. Upload gifs to a separate git repo
2. Upload gifs to a gist in a comment or in a README.md file
3. Upload gifs to a release

hasezoey:
also consider using [git lfs](https://docs.github.com/en/repositories/working-with-files/managing-large-files/configuring-git-large-file-storage)

joshka:
> also consider using [git lfs](https://docs.github.com/en/repositories/working-with-files/managing-large-files/configuring-git-large-file-storage)

Good idea, but we might hit quotas pretty quickly with that:

Demo.gif is 2.8MB, the monthly quota on github's lfs is just 1GiB, and the readme was visited 5475 times in the last 2 weeks (e.g. 15.3GB just for that image), VHS also supports mp4 (651KB->3.6GB), and webm (846KB->4.6GB). I'm unsure how those would render on crates.io, but they look ok on Github. E.g.:

https://github.com/ratatui-org/ratatui/assets/381361/1df72b8c-c202-443c-b718-71b25b0b462f

https://github.com/ratatui-org/ratatui/assets/381361/c625279c-0787-43fc-8549-36bc902602f0

joshka:
+1 to the idea. I'd like this to be easy to automate both in CI and locally.

orhun:
Yeah we totally need this...

joshka:
I took the approach of creating a new branch to store the gifs. https://github.com/ratatui-org/ratatui/tree/images

joshka:
Done

PR #460, opened 2023-09-02
docs(examples): add descriptions and update theme

joshka:
Fixes https://github.com/ratatui-org/ratatui/issues/401

- docs(examples): add descriptions and update theme
Use the OceanicMaterial consistently in examples

- docs(examples): move example gifs to github
A new orphan branch named "images" is created to store the example images

See https://github.com/ratatui-org/ratatui/tree/images

### DIFF
--- a/.github/workflows/check-pr-title.yml
+++ b/.github/workflows/check-pr-title.yml
@@ -1,0 +1,23 @@
+name: Check Pull Requests
+
+on:
+  pull_request_target:
+    types:
+      - opened
+      - edited
+      - synchronize
+  merge_group:
+
+
+permissions:
+  pull-requests: read
+
+jobs:
+  check-title:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check PR title
+        if: github.event_name == 'pull_request_target'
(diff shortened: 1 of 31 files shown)


orhun (approved):
LGTM, should we also auto-generate GIFs on that branch?

joshka:
Definitely worth doing sometime, but there doesn't seem to be an easy way to only generate the gif that has changed, so manual works for the short term.

orhun:
Uhm, why this was closed?

joshka:
We were modifying the merge settings at the time (as rebasing doesn't work with signed commits), and I merged this manually (so the two commits would be separate rather than squashed).
Just https://github.com/joshka/ratatui/commit/ca9bcd3156f55cd2df4edf003aa1401abbed9b12 and https://github.com/joshka/ratatui/commit/5c785b22709fb64a0982722e4f6d0021ccf621b2 are merged (not the extraneous https://github.com/ratatui-org/ratatui/pull/460/commits/bb14d77bcd79fecb9019c392b8d981438e08d682)

Issue #379, opened 2023-08-07
Top line chart dot was styled with Y Axis color


thoongnv:
Only happen if I set `.title()` for `y_axis`

```rust
.y_axis(
      Axis::default()
          .title(Span::styled(
              "num of banners",
              Style::default().fg(Color::Red),
          ))
          .style(Style::default().fg(Color::Gray))
          .bounds([y_bounds[0], y_bounds[y_bounds.len() - 1]])
          .labels(y_labels.iter().cloned().map(Span::from).collect())
          .labels_alignment(Alignment::Center),
  );
```
<img width="1208" alt="Screenshot 2023-08-07 at 11 16 39" src="https://github.com/ratatui-org/ratatui/assets/15308084/e2af33bf-3443-4951-9fd1-b84746df2860">

These white dots seems taken from Y-Axis labels, not title style. And work fine if unset `.title()`, all line dots are red.

## Dependencies
crossterm = "0.26.1"
ratatui = { version = "0.22.0" }

PR #462, opened 2023-09-02
fix(chart): fix chart row style being incorrect when using Axis::title

aatukaj:
Fixes: #379

old behavior
image

new fixed
image

### DIFF
--- a/src/widgets/chart.rs
+++ b/src/widgets/chart.rs
@@ -560,7 +560,7 @@ impl<'a> Widget for Chart<'a> {
 
         if let Some((x, y)) = layout.title_x {
             let title = self.x_axis.title.unwrap();
-            let width = graph_area.right().saturating_sub(x);
+            let width = title.width() as u16;
             buf.set_style(
                 Rect {
                     x,
@@ -575,7 +575,7 @@ impl<'a> Widget for Chart<'a> {
 
         if let Some((x, y)) = layout.title_y {
             let title = self.y_axis.title.unwrap();
-            let width = graph_area.right().saturating_sub(x);
+            let width = title.width() as u16;
             buf.set_style(
                 Rect {
                     x,
@@ -717,5 +717,16 @@ mod tests {
         assert_eq!("Scatter".parse::<GraphType>(), Ok(GraphType::Scatter));
         assert_eq!("Line".parse::<GraphType>(), Ok(GraphType::Line));
         assert_eq!("".parse::<GraphType>(), Err(ParseError::VariantNotFound));
+    }
+
+    #[test]
+    fn it_does_not_panic_if_title_is_wider_than_buffer() {
+        let widget = Chart::default()
+            .y_axis(Axis::default().title("xxxxxxxxxxxxxxxx"))
+            .x_axis(Axis::default().title("xxxxxxxxxxxxxxxx"));
+        let mut buffer = Buffer::empty(Rect::new(0, 0, 8, 4));
+        widget.render(buffer.area, &mut buffer);
+
+        assert_eq!(buffer, Buffer::with_lines(vec![" ".repeat(8); 4]))
     }
 }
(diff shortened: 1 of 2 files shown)


src/widgets/chart.rs
```diff
@@ -560,7 +560,7 @@ impl<'a> Widget for Chart<'a> {
 
         if let Some((x, y)) = layout.title_x {
             let title = self.x_axis.title.unwrap();
-            let width = graph_area.right().saturating_sub(x);
+            let width = title.width() as u16;
```
joshka, line 563: How does this act when title.width is > buffer width? Can you please add a test to ensure that this works fine without panicking?

aatukaj: The title is hidden when it's too long. It's checked [here](https://github.com/ratatui-org/ratatui/blob/e098731d6c1a68a0319d544301ac91cf2d05ccb2/src/widgets/chart.rs#L302C13-L302C13).

aatukaj:
I'm not really sure on how to go about fixing that commit lint :(

src/widgets/chart.rs
joshka: Thanks for implementing this. The code that sets up the invariant is quite a long way (160 lines prior) to the code that relies on it.

joshka:
> I'm not really sure on how to go about fixing that commit lint :(

Mainly this would be fixed by changing the word "added" to "add"
In this case, I'll squash the two commits during the merge, and add some context to the commit message body, as the additional test really is part of the main change and doesn't really need a separate commit.

joshka (approved):

joshka:
Actually - change of plans, double checking this locally, the conventional commit lint prevented the formatting check to run, and there's an extraneous newline that the formatter corrects. Can you please squash the two commits using git rebase and then run `cargo +nightly fmt --all` and force push the result to your github branch (`git push origin main --force`)?

Do you think it would be worth adding another unit test that also checks that the data style is not overwritten by the title / axis style?

Here's a suggestion for an improved commit message. I'm not sure if I've correctly noted that it's the dataset style not the title style here.

```text
fix(chart): use graph style for top line

A bug in the rendering caused the top line of the chart to be rendered
using the style of the Y Axis, instead of the dataset style. This is
fixed by only setting the style for the width of the text, and not the
entire row.

Fixes: https://github.com/ratatui-org/ratatui/issues/379
```

joshka (changes requested):
Actually - change of plans, double checking this locally, the conventional commit lint prevented the formatting check to run, and there's an extraneous newline that the formatter corrects. Can you please squash the two commits using git rebase and then run `cargo +nightly fmt --all` and force push the result to your github branch (`git push origin main --force`)?

Do you think it would be worth adding another unit test that also checks that the data style is not overwritten by the title / axis style?

Here's a suggestion for an improved commit message. I'm not sure if I've correctly noted that it's the dataset style not the title style here.

```text
fix(chart): use graph style for top line

A bug in the rendering caused the top line of the chart to be rendered
using the style of the Y Axis, instead of the dataset style. This is
fixed by only setting the style for the width of the text, and not the
entire row.

Fixes: https://github.com/ratatui-org/ratatui/issues/379
```

aatukaj:
Added the test. Hopefully I did the rebase correctly. Thanks for the commit suggestion.

joshka (approved):

joshka:
Thanks for adding the test. LGTM. One last thing (unless the other maintainers have comments on the PR):

We use commit signature verification, which will block commits from being merged unless they are signed. To set up your machine to sign commits, see [managing commit signature verification](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification) in GitHub docs.

Also not a big deal, the commit message doesn't really need the following lines as writing tests can be generally considered as part of a feature:

```
test(chart): add test for too big titles
test(chart): add test for correct top line styling
```

orhun (approved):

aatukaj:
Should be all good now.
Thanks for all the help!

PR #463, opened 2023-09-02
ci: put commit id first in changelog

joshka:
```shell
git cliff --unreleased
```
[CHANGELOG.md](https://github.com/ratatui-org/ratatui/files/12503810/CHANGELOG.md)

image

### DIFF
--- a/cliff.toml
+++ b/cliff.toml
@@ -21,8 +21,8 @@ body = """
 {% endif -%}
 
 {% macro commit(commit) -%}
-- *({{commit.scope | default(value = "uncategorized")}})* {{ commit.message | upper_first }}
-([{{ commit.id | truncate(length=7, end="") }}]({{ "https://github.com/ratatui-org/ratatui/commit/" ~ commit.id }}))
+- [{{ commit.id | truncate(length=7, end="") }}]({{ "https://github.com/ratatui-org/ratatui/commit/" ~ commit.id }})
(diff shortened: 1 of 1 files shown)


orhun:
How do you think this helps with readability? I think the current format is good and I would say let's stick to it for a while for consistency in the changelog :D

joshka:
Mostly this is a cosmetic change backed by some subjective preferences.

Putting the commit ID first makes the visual design hierarchy more obvious by implying that each row in the changelog is strongly coupled with a commit. It also avoids the subjective messiness of having two parenthesized parts at the end of line (and avoids wrapping the commit id to the next line. By putting a fixed size piece of text at the beginning of a line, it allows someone reading to more easily ignore the first part when scanning down a list of items than if the same blue text is at the end of the line or on the following line. It's not a massive deal either way, but I think this is marginally nicer.

Current Changelog for comparison:

image

mindoodoo:
@joshka Is the commit ID always wrapped to the next line as in your screenshot ?

joshka:
> @joshka Is the commit ID always wrapped to the next line as in your screenshot ?

No - sorry. I'd pasted the output of git-cliff into a github comment and that renders this wrong. My mistake there.

mindoodoo (approved):
Aight, I guess I'm good either way, though I agree with orhun that we shouldn't change it up too often.

The change makes sense and will put an even stronger emphasis on writing clean commits.

orhun (approved):

PR #469, opened 2023-09-05
docs(Block): add documentation to Block

Valentin271:
Add documentation to the `Block` widget and its sub-components (`Padding`, `Borders`).

Relates to #386

### DIFF
--- a/src/widgets/block.rs
+++ b/src/widgets/block.rs
@@ -1,3 +1,11 @@
+#![warn(missing_docs)]
+//! Elements related to the `Block` base widget.
+//!
+//! This holds everything needed to display and configure a [`Block`].
+//!
+//! In its simplest form, a `Block` is a [border](Borders) around another widget. It can have a
+//! [title](Block::title) and [padding](Block::padding).
+
 #[path = "../title.rs"]
 pub mod title;
 
@@ -12,16 +20,60 @@ use crate::{
     widgets::{Borders, Widget},
 };
 
+/// The type of border of a [`Block`].
+///
+/// See the [`borders`](Block::borders) method of `Block` to configure its borders.
 #[derive(Debug, Default, Display, EnumString, Clone, Copy, Eq, PartialEq, Hash)]
 pub enum BorderType {
+    /// A plain, simple border.
+    ///
+    /// This is the default
+    ///
+    /// # Example
+    ///
+    /// ```plain
+    /// ┌───────┐
+    /// │       │
+    /// └───────┘
+    /// ```
     #[default]
     Plain,
+    /// A plain border with rounded corners.
+    ///
+    /// # Example
+    ///
+    /// ```plain
+    /// ╭───────╮
+    /// │       │
+    /// ╰───────╯
+    /// ```
     Rounded,
+    /// A doubled border.
+    ///
+    /// Note this uses one character that draws two lines.
+    ///
+    /// # Example
+    ///
+    /// ```plain
+    /// ╔═══════╗
+    /// ║       ║
+    /// ╚═══════╝
+    /// ```
     Double,
+    /// A thick border.
+    ///
+    /// # Example
+    ///
+    /// ```plain
+    /// ┏━━━━━━━┓
+    /// ┃       ┃
+    /// ┗━━━━━━━┛
+    /// ```
     Thick,
 }
 
 impl BorderType {
+    /// Convert this `BorderType` into the corresponding [`Set`](line::Set) of lines.
     pub const fn line_symbols(border_type: BorderType) -> line::Set {
         match border_type {
             BorderType::Plain => line::NORMAL,
@@ -32,15 +84,36 @@ impl BorderType {
     }
 }
 
+/// Defines the padding of a [`Block`].
+///
+/// See the [`padding`](Block::padding) method of [`Block`] to configure its padding.
+///
+/// This concept is similar to [CSS padding](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_box_model/Introduction_to_the_CSS_box_model#padding_area).
+///
+/// **NOTE**: Terminal cells are often taller than they are wide, so to make horizontal and vertical
+/// padding seem equal, doubling the horizontal padding is usually pretty good.
+///
+/// # Example
+///
+/// ```
+/// use ratatui::widgets::block::Padding;
+/// Padding::uniform(1);
+/// Padding::horizontal(2);
+/// ```
 #[derive(Debug, Default, Clone, Copy, Eq, PartialEq, Ord, PartialOrd, Hash)]
 pub struct Padding {
+    /// Left padding
     pub left: u16,
+    /// Right padding
     pub right: u16,
+    /// Top padding
     pub top: u16,
+    /// Bottom padding
     pub bottom: u16,
 }
 
 impl Padding {
+    /// Creates a new `Padding` by specifying every field individually.
     pub const fn new(left: u16, right: u16, top: u16, bottom: u16) -> Self {
         Padding {
             left,
@@ -50,6 +123,9 @@ impl Padding {
         }
     }
 
+    /// Creates a `Padding` of 0.
+    ///
+    /// This is also the default.
     pub const fn zero() -> Self {
         Padding {
             left: 0,
@@ -59,6 +135,9 @@ impl Padding {
         }
     }
 
+    /// Defines the [`left`](Padding::left) and [`right`](Padding::right) padding.
+    ///
+    /// This leaves [`top`](Padding::top) and [`bottom`](Padding::bottom) to `0`.
     pub const fn horizontal(value: u16) -> Self {
         Padding {
             left: value,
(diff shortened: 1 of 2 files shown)


orhun (approved):

src/widgets/block.rs
```diff
@@ -191,24 +299,30 @@ impl<'a> Block<'a> {
         self
     }
 
-    /// Applies the style to all titles. If a title already has a style, it will add on top of it.
+    /// Applies the style to all titles.
```
joshka, line 302: All titles have a style - even when that style is `Style::default()` It's worth talking about how the combination of a default style, plus the specific override is done using consistent language **across all of the docs**. Consistency on this is notable because the order of applying styles has come up in discord / issues before as a point of confusion, and we use this pattern a lot.

What about something like:
"Sets the default style for titles which may be overridden by the styles on each individual title"

I'm not quite sure that "override" is quite the right verb here as the styles are combined, not replaced, and the default is applied first.

Thoughts?

src/widgets/block.rs
```diff
@@ -191,24 +299,30 @@ impl<'a> Block<'a> {
...
+    /// If a [`Title`] already has a style, the title's style will add on top of this one.
     pub const fn title_style(mut self, style: Style) -> Block<'a> {
         self.titles_style = style;
         self
     }
 
-    /// Aligns all elements that don't have an alignment
+    /// Aligns all [titles](Title) that don't have an [`Alignment`].
+    ///
+    /// This effectively acts as a default for titles that don't explicitly set an [`Alignment`].
```
joshka, line 312: This can be stronger:

/// Sets the default alignment of titles...

Similar note as above about how to write about default / override with consistent language, though this one is less important as it's a pure replacement. It still would be handy to end up with the same language meaning the same thing across the library.

src/widgets/block.rs
```diff
@@ -176,13 +257,40 @@ impl<'a> Block<'a> {
...
+    /// # Example
+    ///
+    /// ```
+    /// # use ratatui::widgets::{Block, block::title::Title};
+    /// # use ratatui::layout::Alignment;
+    /// Block::default()
+    ///     .title("Title") // By default in the top left corner
+    ///     .title(Title::from("Right").alignment(Alignment::Right))
+    ///     .title(Title::from("Center").alignment(Alignment::Center));
+    /// ```
```
joshka, line 286: It would be good to show a rendered version of this example in the doc. The note about how having multiple titles separated by a space would be more obvious if shown visually, (also showing that it's actually filled with a border character not a space)

src/widgets/block.rs
```diff
@@ -221,38 +335,81 @@ impl<'a> Block<'a> {
         self.title_position(Position::Bottom)
     }
 
-    /// Positions all titles that don't have a position
+    /// Positions all [titles](Title) that don't have a [`Position`].
```
joshka, line 338: Sets the default...

src/widgets/block.rs
```diff
@@ -221,38 +335,81 @@ impl<'a> Block<'a> {
...
+    ///     .title_position(Position::Bottom);
     /// ```
     pub const fn title_position(mut self, position: Position) -> Block<'a> {
         self.titles_position = position;
         self
     }
 
+    /// Defines the style of the borders.
+    ///
+    /// If a [`Block::style`] is defined, `border_style` will be applied on top of it.
```
joshka, line 361: The order of the styles being applied might be worth putting in the widget level doc rather than just the method level

src/widgets/block.rs
```diff
@@ -221,38 +335,81 @@ impl<'a> Block<'a> {
...
     pub const fn border_style(mut self, style: Style) -> Block<'a> {
         self.border_style = style;
         self
     }
 
+    /// Defines the block style.
+    ///
+    /// This is the most generic [`Style`] a block can receive, it will be merged with any other
+    /// more specific style. Elements can be styled further with [`Block::title_style`] and
+    /// [`Block::border_style`].
```
joshka, line 382: Need to say something about how this will affect any widget contents that is rendered inside the block.
(See the Styled Block example at https://github.com/ratatui-org/ratatui/tree/main/examples#block)

"Merged" might be the word I was looking for above instead of override.

src/widgets/block.rs
```diff
@@ -264,15 +421,16 @@ impl<'a> Block<'a> {
     ///
     /// ```
     /// // Draw a block nested within another block
-    /// use ratatui::{backend::TestBackend, buffer::Buffer, terminal::Terminal, widgets::{Block, Borders}};
+    /// use ratatui::{
+    ///     backend::TestBackend,
+    ///     buffer::Buffer,
+    ///     terminal::Terminal,
+    ///     widgets::{Block, Borders},
+    /// };
```
joshka, line 429: This example needs to be shortened to just the interesting parts. E.g.:
```rust
/// # use ratatui::{prelude::*, widgets::Block};
/// # fn render_nested_block<B: Backend>(frame: &mut Frame) {
/// let outer = Block::default().title("Outer").borders(Borders::ALL);
/// let inner = Block::default().title("Inner").borders(Borders::ALL);
/// let outer_area = frame_size();
/// let inner_area = outer.inner(outer_area);
/// frame.render_widget(outer_block, outer_area);
/// frame.render_widget(inner_block, inner_area);
/// # }
```
renders:
```plain
┌Outer Block──┐
│┌Inner Block┐│
││           ││
│└───────────┘│
└─────────────┘
```

src/widgets/block.rs
```diff
@@ -221,38 +335,81 @@ impl<'a> Block<'a> {
...
+    /// Block::default().borders(Borders::LEFT | Borders::RIGHT);
+    /// ```
     pub const fn borders(mut self, flag: Borders) -> Block<'a> {
         self.borders = flag;
         self
     }
 
+    /// Defines the type of border to use.
+    ///
+    /// See [`BorderType`] for more information.
```
joshka, line 412: This is a little ambiguous (but that flows from the name itself) and makes it hard for a reader to search the docs for how to set this.
Perhaps "Sets the symbols used to display the border (e.g. single line, double line, thick or rounded borders)"?

joshka (approved):
Thanks for doing this. I've marked the PR as approved, and we can merge it as-is, but I'd also be happy to keep this open to take whatever changes you see as useful based on the feedback below.

I have lots of little comments on this, but none of them are super important / blocking, so please read them as discussion points.

src/widgets/block.rs
```diff
@@ -32,15 +84,36 @@ impl BorderType {
     }
 }
 
+/// Defines the padding of a [`Block`].
+///
+/// The padding unit is a character. As such, `1` horizontal ([`left`](Padding::left) or
+/// [`right`](Padding::right)) padding does not equal `1` vertical ([`top`](Padding::top) or
+/// [`bottom`](Padding::bottom)) padding.
```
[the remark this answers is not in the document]
Valentin271: > cell is usually higher than it is wide

Yes this is exactly what I was trying to say.

> I would assume that it's expressed in characters already

Some newcomers to TUIs might not know how this is rendered under the hood. So I thought it was important precise this is different from CSS.

src/widgets/block.rs
```diff
@@ -68,6 +147,9 @@ impl Padding {
         }
     }
 
+    /// Defines the [`top`](Padding::top) and [`bottom`](Padding::bottom) padding.
+    ///
+    /// This leaves [`left`](Padding::left) and [`right`](Padding::right) to `0`.
```
a-kenji, line 152: Either:
```suggestion
This leaves [`left`](Padding::left) and [`right`](Padding::right) at `0`.
```suggestion
or
```
This sets [`left`](Padding::left) and [`right`](Padding::right) to `0`.
```

File: src/widgets/block.rs [hunk 11]
```diff
@@ -77,6 +159,7 @@ impl Padding {
         }
     }
 
+    /// Applies the same value to every `Padding` component.
```
a-kenji, line 162:
```suggestion
    /// Applies the same value to every `Padding` field.
```

src/widgets/block.rs
Valentin271: I wanted to reduce it but didn't thought of using a hidden function. Anyway fixed it.

src/widgets/block.rs
Valentin271: Yep way better, fixed.

src/widgets/block.rs
Valentin271: Yes that's clearer with this phrasing.

I kept the reference to `BorderType`, but rephrased it, as this is an explicit list of all border types.

Valentin271:
I fixed most of your comments. I just left out the comments about the style and style prioritization. I agree that we should use a consistent language and document the style priorities. But I think this would be better documented in the `style` module (mainly the way we merge style), then any method could reference that as a base and add on top of it.

As for the consistent language, we could add to the guideline, for example:
- Setters that replace values use this sentence: 'Sets xxx'
- Setters that add default values use this sentence: 'Sets the default for xxx'

Put in other words, some sort of template.  
Let me know what you think. I think consistency is very important, but how do we enforce it across multiple collaborators.

joshka (approved):

joshka:
Thanks - I'll slap together a PR that fixes up the remaining wording for styles.

joshka:
I removed the commit message body, as I don't think it really added much info to the changelog that wouldn't be answered by reading the commit title.

Issue #457, opened 2023-09-01
Braille symbol not displayed


relf:
## Description
The Braille symbol is not displayed on my Windows 10 machine (replaced by a framed question mark character).
See screenshot below.

## To Reproduce

In Windows cmd, after cloning the repo:
```
cargo run --example=chart --features=crossterm
```

## Expected behavior

See [chart example](https://github.com/ratatui-org/ratatui/blob/main/examples/README.md#chart-chartrs)

## Screenshots

![ratatui_bug](https://github.com/ratatui-org/ratatui/assets/1198802/7ae6f31b-8eb4-4c66-b13b-1c94641196d7)

## Environment

- OS: Windows 10
- Terminal Emulator: cmd
- Font: Consolas, but I've tried some others (Courrier New, Lucida Console, ...) without success. Is there a recommended font?  
- Crate version: 0.23.0
- Backend: crossterm

Valentin271:
Have you tried powershell and/or in WSL? (will probably be the same result but still good to know)

Either way I found a [similar issue on gtop](https://github.com/aksakalli/gtop/issues/21) with a fix.
See also the [accepted answer on this StackOverflow post](https://stackoverflow.com/questions/33520197/print-braille-characters-to-cmd-c).

It seems like this is a font issue, so if you want a specific font, it may be worth to try a [Nerd Font](https://www.nerdfonts.com/).

_Disclaimer: I did not test any of these._

relf:
Thanks for the swift reply. I confirm it does not work with powershell nor in WSL.
I have to test but your links seem to give a solution. Too bad the Windows default font does not make it!

I do not know if I am the only user on Windows here 😅 but it might be worth to add a warning in the doc, an information about the font requirement could be added.

relf:
Ok, I confirm that it works once the Iosevka font is installed as described in the [first link](https://github.com/aksakalli/gtop/issues/21) mentionned above. 
In my case (weirdly?) I did not have to edit the registry, the font was available in the cmd properties just after the font installation (it was lucky because I do not have the admin rights to play with regedit). Once selected it becomes the default for subsequent cmd.
The second font mentioned in the stackoverflow link does not work for me. And finally I've also tried a Nerd Font font without any success as well.

I do not close the issue in case you want to keep it around at least till the font info is added somewhere in the doc, otherwise feel free to close it.

joshka:
A quick google to see if there's a good list of which fonts support braille character block I found:
[...]

PR #472, opened 2023-09-07
docs(canvas): add support note for Braille marker

joshka:
Addresses https://github.com/ratatui-org/ratatui/issues/457

### DIFF
--- a/src/symbols.rs
+++ b/src/symbols.rs
@@ -244,14 +244,21 @@ pub mod braille {
 /// Marker to use when plotting data points
 #[derive(Debug, Default, Display, EnumString, Clone, Copy, Eq, PartialEq, Hash)]
 pub enum Marker {
-    /// One point per cell in shape of dot
+    /// One point per cell in shape of dot ("•")
     #[default]
     Dot,
-    /// One point per cell in shape of a block
+    /// One point per cell in shape of a block ("█")
     Block,
-    /// One point per cell in the shape of a bar
+    /// One point per cell in the shape of a bar ("▄")
     Bar,
-    /// Up to 8 points per cell
+    /// Use the [Unicode Braille Patterns](https://en.wikipedia.org/wiki/Braille_Patterns) block to
+    /// represent data points.
+    ///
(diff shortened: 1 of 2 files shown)


joshka:
Pinging @relf for input

Valentin271:
I wonder should we link to some fonts that support this? Otherwise LGTM.

orhun (approved):
+1

relf:
The thing is, discovering `ratatui`, I just ran the demo and seing I did not get the right result out of the box, I started to dive into the code to find out, your additions would have definitely helped. 
That said, I've just noticed now (!) that I've skipped the warning just after in the [example section](https://github.com/ratatui-org/ratatui#examples) where you explain about having gliphs not showing up correctly 🤦‍♂️, maybe
some **bold font** could have helped (?) and/or links to "good" fonts would be nice right here.
The default configuration on Windows (Consolas font) does not work, I think this could be stressed as an example of "bad" configuration.
Anyway thanks and keep up the good work!

kdheepak:
Should we also add a one line text description to the `ui` in the example for `chart.rs`?:

https://github.com/ratatui-org/ratatui/blob/0c68ebed4f63a595811006e0af221b11a83780cf/examples/chart.rs#L145

Something like "Requires terminal and font support for unicode".

joshka:
> The default configuration on Windows (Consolas font) does not work, I think this could be stressed as an example of "bad" configuration.

I'd be more inclined to suggest that whatever we do with demos should generally look good out of the box for a default configuration, so perhaps we should switch to dots instead of braille for the main demo and introduce the braille markers later?

Valentin271:
Yes this seems like a good idea to me.

relf:
> demos should generally look good out of the box for a default configuration, so perhaps we should switch to dots instead of braille for the main demo and introduce the braille markers later

Yes that was my first thought but I did not dare to suggest it (more involved changes), but now that you say it...

joshka:
> I wonder should we link to some fonts that support this? Otherwise LGTM.

Perhaps a wiki page might be appropriate for this?

PR #473, opened 2023-09-08
docs: add logo and favicon to docs.rs page

joshka:
image

<br>

image


src/assets/logo.png
kdheepak, line 1: What do you think about making a separate repo for assets? We could put all the gifs there too.

joshka: These are 1kb (logo) and 15kb (icon), and should change rarely so I'm not concerned about them taking up space in the source (whereas I normally would be for larger images or things which change regularly). For comparison, style.rs is 33kb (which is subjectively large for a source file and should perhaps be trimmed / refactored).

We already have a [separate branch](https://github.com/ratatui-org/ratatui/tree/images) for the example gifs. A new repo would also work, but it doesn't solve any problem we have as far as I can tell right now (at least I can't see an obvious benefit over what we have already).

I noticed the folder was wrong however - it should be top level /assets folder not under src (hence the amended commit pushed just now).

orhun:
Speaking of image sizes, what do you guys think about setting up [`imgbot`](https://imgbot.net) to optimize the image sizes in `main`/`images` branch?

orhun (approved):

src/assets/logo.png
kdheepak: I just don't like any binary files in a git repo where the history matters :)

Issue #475, opened 2023-09-08
RGB / 24 bit true color glitching / blinking in macOS terminal.app


nx10:
## Issue Status

macOs terminal does not support TrueColor. This will be documented in the `Color::Rgb` docs (as of PR #477)

Original issue text:

---

## Description

Colors are not working properly in macOS system terminal. I am using crossterm as a backend.

## To Reproduce

Something like this when drawing the buffer

```Rust
let fg = tui::style::Color::Rgb(123, 312, 22); // any color
let bg = tui::style::Color::Rgb(123, 22, 312); // any color
buf.get_mut(0, 0).set_char('▄').set_fg(fg).set_bg(bg);
```

## Expected behavior

Instead of showing the correct colors as expected the whole buffer looks weird and blinks (see video below).

## Screenshots

This is a video of a ratatui application where most of the buffer is constructed like in the example above.

https://github.com/ratatui-org/ratatui/assets/33600480/0a650b5c-ce23-4b7f-be96-b3c60efc8aed

Same terminal with font "SF Mono Regular 11" (System default) so it does not seem to be a font issue

![image](https://github.com/ratatui-org/ratatui/assets/33600480/1ab46313-4ebc-4af1-ab40-c5b52e89c09b)

This is what it looks like in ITerm2 on the same machine (and in every other terminal I tried on linux + windows)

![image](https://github.com/ratatui-org/ratatui/assets/33600480/dd94d645-95d0-4b31-bbc5-95d49c810ecb)

## Environment

- OS: macOS Ventura 13.5 (M2)
- Terminal Emulator: macOS system terminal (Version 2.13 (447))
- Font: Monaco 18 
- Crate version: 0.23.0
- Backend: crossterm

joshka:
Do you know if this Is this just the Rgb colors or all colors?
Can you run the colors example and post the output? (I'll add a second view that shows RGB)

As a side note, looking at the image you're displaying (very cool btw), you might be interested in https://crates.io/crates/ratatu-image by @benjajaja

joshka:
Using the new colors example from https://github.com/ratatui-org/ratatui/pull/476 I am able to repro this:

<img width="1832" alt="image" src="https://github.com/ratatui-org/ratatui/assets/381361/5c6a7e99-7e25-4937-a7b8-b9eac98742fd">
expected (iterm2):
[...]

PR #476, opened 2023-09-09
docs(examples): add colors_rgb example

joshka:
Helpful for diagnosing what's happening with https://github.com/ratatui-org/ratatui/issues/475

colors_rgb

### DIFF
--- a/examples/colors_rgb.rs
+++ b/examples/colors_rgb.rs
@@ -1,0 +1,160 @@
+/// This example shows the full range of RGB colors that can be displayed in the terminal.
+///
+/// Requires a terminal that supports 24-bit color (true color) and unicode.
+use std::{
+    error::Error,
+    io::{stdout, Stdout},
+    rc::Rc,
+    time::Duration,
+};
+
+use crossterm::{
+    event::{self, Event, KeyCode, KeyEventKind},
+    terminal::{disable_raw_mode, enable_raw_mode, EnterAlternateScreen, LeaveAlternateScreen},
+    ExecutableCommand,
+};
+use itertools::Itertools;
+use ratatui::{prelude::*, widgets::*};
+
+type Result<T> = std::result::Result<T, Box<dyn Error>>;
+
+fn main() -> Result<()> {
+    install_panic_hook();
+    App::new()?.run()
+}
+
+struct App {
+    terminal: Terminal<CrosstermBackend<Stdout>>,
+    should_quit: bool,
+}
+
+impl App {
+    pub fn new() -> Result<Self> {
+        Ok(Self {
+            terminal: Terminal::new(CrosstermBackend::new(stdout()))?,
+            should_quit: false,
+        })
+    }
+
+    pub fn run(mut self) -> Result<()> {
+        init_terminal()?;
+        self.terminal.clear()?;
+        while !self.should_quit {
+            self.draw()?;
+            self.handle_events()?;
+        }
+        restore_terminal()?;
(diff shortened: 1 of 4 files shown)


examples/colors.rs
```diff
@@ -28,32 +27,149 @@ fn main() -> Result<()> {
...
-    render_indexed_grayscale(frame, layout[2]);
+    render_title(frame, layout[0]);
+    render_named_colors(frame, layout[1]);
+    render_indexed_colors(frame, layout[2]);
+    render_indexed_grayscale(frame, layout[3]);
+}
+
+fn render_title<B: Backend>(frame: &mut Frame<'_, B>, area: Rect) {
+    frame.render_widget(
+        Paragraph::new("Color example. Press Space for RGB, q to quit")
```
kdheepak, line 83: Minor comment. It'd be nice to change the title here based on `view_index`.

kdheepak:
Do you think this should just be a separate demo?

joshka:
> Do you think this should just be a separate demo?

Yes probably. I wasn't sure about it myself, so just went with adding code, but it wouldn't hurt.

orhun:
Works great!

It isn't really intuitive from the demo so I'd also say that we should have this as a separate example.

joshka:
I tweaked a bunch of stuff in the updated example:
- Modified color grid to be smooth by mirroring blue channel every 16 pixels instead of starting at 0
- Changed to `buf.cell_mut()` this is noticeably faster to render than using `Paragraph`
- App-ified and Widget-ified implementation
  - Removed manual Error printing (returning `Result<()>` handles this automatically)
  - panic hook and `Drop` impl (because I kept getting index out of bounds errors when playing around with generating the grid)
- Simplified everything as much as possible

Made with VHS

joshka:
- Added example to cargo toml and fixed commit message

examples/colors_rgb.rs
```diff
@@ -0,0 +1,159 @@
...
+    terminal::{disable_raw_mode, enable_raw_mode, EnterAlternateScreen, LeaveAlternateScreen},
+    ExecutableCommand,
+};
+use itertools::Itertools;
+use ratatui::{prelude::*, widgets::*};
+
+type Result<T> = std::result::Result<T, Box<dyn Error>>;
+
+fn main() -> Result<()> {
+    better_panic::install();
```
orhun, line 22: Wow, nice crate! Should we also use this for the other examples for backtracing errors etc.?

kdheepak: I think `better_panic` (or `human_panic`) should be part of all our examples + templates etc. It is a much nice developer experience and user experience with these installed.

I do have [a section in the book](https://ratatui-org.github.io/ratatui-book/how-to/setup-panic-hooks-better-panic.html) that describes how users can use `better_panic` so that's a start.

joshka: Yep definitely - I'm searching for a consistent approach to use for all the examples instead of something different in each one. I think this is pretty close to the right balance of simplicity and functionality for examples. It's just enough to not have to pass around too much state. 

I wonder though whether the `init_terminal()`, `restore_terminal()` and `install_panic_hook()` methods might be worth putting at the bottom of the code outside of `impl App` instead.

joshka:
moved the init/restore/panic hook to the bottom of the file to make the example really hyper focus on the application logic and example rather than the generic terminal parts.

kdheepak (approved):

PR #477, opened 2023-09-09
docs(color): document true color support

joshka:
- refactor(style): move Color to separate color mod
- docs(color): document true color support

Documents https://github.com/ratatui-org/ratatui/issues/475


src/style/color.rs
```diff
@@ -0,0 +1,360 @@
...
+    /// ANSI Color: Bright Magenta. Foreground: 95, Background: 105
+    LightMagenta,
+    /// ANSI Color: Bright Cyan. Foreground: 96, Background: 106
+    LightCyan,
+    /// ANSI Color: Bright White. Foreground: 97, Background: 107
+    /// Sometimes called `bright white` or `light white` in some terminals
+    White,
+    /// An RGB color.
+    ///
+    /// Note that Only terminals that support 24-bit true color will display this correctly.
```
orhun, line 106:
```suggestion
    /// Note that only terminals that support 24-bit true color will display this correctly.
```

joshka: Fixed in https://github.com/ratatui-org/ratatui/commit/b3553164045f69ef074ba8f00d178abf4c2627b1

orhun (approved):

PR #478, opened 2023-09-09
feat!: Table widget can now select single cells

tonogdlp:
This Pull Request adds the ability for a Table widget to select a single cell, a row, a column, or both a row and column.

---

***Breaking Change***
The main change is a new struct to represent a (row, column) selection:
```rust
pub struct TableSelection {
    pub row: usize,
    pub col: usize,
}
```

Instead of using a `Option<usize>` in `TableState`, we use the new `TableSelection` struct:

```rust
pub struct TableState {
    offset: usize,
    selected: Option<TableSelection>,         // used to be Option<usize>
}
```

This new way of "selecting" in the table generates 2 questions:
1. What if I want to select the full row as was the default behavior before?
2. Do we add spacing for the HighlightSymbol (e.g. ">> ") for all columns, or just the first one?

For the first question, after a conversation in #376, we figured that selecting both the full row and/or full column could be desirable.
With that in mind, a enum was added to represent the type of highlighting:

```rust
pub enum HighlightArea {
    /// No highlight
    None,
    /// Highlight cell only
    #[default]
    Cell,
    /// Highlight entire row
    Row,
    /// Highlight entire column
    Col,
    /// Highlight entire row and column
    RowAndCol,
}
```

Examples:
- HighlightArea::Cell

https://github.com/ratatui-org/ratatui/assets/8254361/b809a898-5377-4311-83e1-9b6f03a6c1cd

- HighlighArea::Row
https://github.com/ratatui-org/ratatui/assets/8254361/3b2f63f3-15f5-43e1-a17c-05522942a97d

- HighlightArea:Col

https://github.com/ratatui-org/ratatui/assets/8254361/16d8044c-7488-41f0-8b17-f6eee5f2c4c1

- HighlightArea::RowAndCol

https://github.com/ratatui-org/ratatui/assets/8254361/d9c46c42-cf6c-4ed3-ba5e-3e14aa461671

- HighlightArea::None

https://github.com/ratatui-org/ratatui/assets/8254361/4b44b8ed-ebe4-42b6-ba3f-689e3c47903e

---

In regards to the second question, Table has now a new field..

```rust
pub struct Table<'a> {
    ....
    /// Whether to allocate spacing for all columns.
    highlight_spacing_all_columns: bool,
}
```
Which, if set to `true`, will add a space for the `highlight_symbol` for each column, in the same way that it was added to the whole row before.

https://github.com/ratatui-org/ratatui/assets/8254361/afb242ce-f2ba-4391-9367-8988205e693c

---

### Commits
The PR is 4 commits:

 - `7db94tb` adds `struct TableSelect`, `enum HighlightArea`, and implements both of them
 - `f6741c9` adds the field `highlight_spacing_all_columns: bool`, and modifies the `get_columns_widths( )` implementation accordingly
 - `4a2a80f` adds and/or update tests in both src/widgets/table.rs and examples/widgets_table.rs
 - `47f0250` updates the example/table.rs in order to showcase the different settings added.

Any feedback you can give me will be great, as I'm not an expert on this crate, and I am most likely making a mistake somewhere.

### DIFF
--- a/src/widgets.rs
+++ b/src/widgets.rs
@@ -4,19 +4,22 @@
 //! meant to be stored but used as *commands* to draw common figures in the UI.
 //!
 //! The available widgets are:
-//! - [`Block`]
-//! - [`Tabs`]
-//! - [`List`]
-//! - [`Table`]
-//! - [`Paragraph`]
-//! - [`Chart`]
-//! - [`BarChart`]
-//! - [`Gauge`]
-//! - [`Sparkline`]
-//! - [`Scrollbar`]
-//! - [`calendar::Monthly`]
-//! - [`Clear`]
-
+//! - [`Block`]: a basic widget that draws a block with optional borders, titles and styles.
+//! - [`BarChart`]: displays multiple datasets as bars with optional grouping.
+//! - [`calendar::Monthly`]: displays a single month.
+//! - [`Canvas`]: draws arbitrary shapes using drawing characters.
+//! - [`Chart`]: displays multiple datasets as a lines or scatter graph.
+//! - [`Clear`]: clears the area it occupies. Useful to render over previously drawn widgets.
+//! - [`Gauge`]: displays progress percentage using block characters.
+//! - [`LineGauge`]: display progress as a line.
+//! - [`List`]: displays a list of items and allows selection.
+//! - [`Paragraph`]: displays a paragraph of optionally styled and wrapped text.
+//! - [`Scrollbar`]: displays a scrollbar.
+//! - [`Sparkline`]: display a single data set as a sparkline.
+//! - [`Table`]: displays multiple rows and columns in a grid and allows selection.
+//! - [`Tabs`]: displays a tab bar and allows selection.
+//!
+//! [`Canvas`]: crate::widgets::canvas::Canvas
 mod barchart;
 pub mod block;
 #[cfg(feature = "widget-calendar")]


tests/widgets_table.rs
```diff
@@ -747,11 +747,94 @@ fn widgets_table_enable_always_highlight_spacing() {
...
+            "│                            │",
+            "│>> Row11    Row12    Row13  │",
+            "│   Row21    Row22    Row23  │",
+            "│                            │",
+            "│   Row31    Row32    Row33  │",
+            "└────────────────────────────┘",
+        ]),
+    );
+
+    state.select(Some(TableSelection { row: 1, col: 0 }));
```
hasezoey, line 815: maybe consider adding another case where `col` is not `0`, just to be sure

tests/widgets_table.rs
```diff
@@ -747,11 +747,94 @@ fn widgets_table_enable_always_highlight_spacing() {
 }
 
 #[test]
-fn widgets_table_can_have_elements_styled_individually() {
+fn widgets_table_can_have_spacing_in_all_columns() {
```
joshka, line 750: It's tempting to continue with writing these integration style tests, but they have a tendency to be a bit verbose compared to an equivalent unit test. E.g.:

```rust
let buffer = Buffer::empty(Rect::new(0, 0, 15, 7));
let base_table = Table::new(vec![Row::new(vec!["C1", "C2", "C3"])]; 4)
  .header(Row::new(vec!["H1", "H2", "H3"])
  .block(Block::new().borders(Borders::ALL))
  ...

let mut state = TableState::default();
base_table.clone()
  .highlight_spacing(true)
  .render_stateful_widget(buffer.area, buffer, &mut state);
assert_buffer_eq!(
  buffer,
  buffer::with_lines(vec![
    "┌─────────────┐",
    "│H1 H2 H3     │",
    "│             │",
    "│C1 C2 C3     │",
    "│C1 C2 C3     │",
    "│C1 C2 C3     │",
    "└─────────────┘",
  ]),
);

state.select(Some(TableSelection::Row(0)));
base_table.clone()
  .highlight_spacing(true).
  .render_stateful_widget(buffer.area, buffer, &mut state);
assert_buffer_eq!(
  buffer,
  buffer::with_lines(vec![
    "┌─────────────┐",
    "│   H1 H2 H3  │",
    "│             │",
    "│>> C1 C2 C3  │",
    "│   C1 C2 C3  │",
    "│   C1 C2 C3  │",
    "└─────────────┘",
  ]),
);
```

For testing highlight styles, you can assign the expected buffer to a variable and call `expected.set_style(Rect::new(...))`). See https://github.com/ratatui-org/ratatui/blob/main/src/widgets/tabs.rs#L212 for an example of this.

Note: this is a subjective comment rather than one which blocks this in any way. It would be nicer for the long term health to head in this direction. It's possible that this is more work than you have time available, so I'd understand if it wasn't a priority.

tests/widgets_table.rs
```diff
@@ -747,11 +747,94 @@ fn widgets_table_enable_always_highlight_spacing() {
...
+    let test_case = |state: &mut TableState, space: HighlightSpacing, expected: Buffer| {
```
joshka, line 751: Use a function instead of a closure and add `#[track_caller]` to make it easier to work out which test fails (the reported failure line will be e.g. line 784 or 800 instead of always line 777)

src/widgets/table.rs
```diff
@@ -194,6 +194,21 @@ impl HighlightSpacing {
     }
 }
 
+#[derive(Debug, Display, EnumString, PartialEq, Eq, Clone, Default, Hash)]
+pub enum HighlightArea {
```
joshka, line 198: Please add a doc comment for the type (it's helpful when hovering over a type in front of a variable or field declaration).

src/widgets/table.rs
```diff
@@ -194,6 +194,21 @@ impl HighlightSpacing {
...
+    None,
+    /// Highlight cell only
+    #[default]
+    Cell,
+    /// Highlight entire row
+    Row,
+    /// Highlight entire column
+    Col,
+    /// Highlight entire row and column
+    RowAndCol,
```
joshka, line 209: Is this likely to ever have any other values? What do you think about calling this `Both` instead of `RowAndCol`?

This is not a blocker - I'm actually unsure which name works better.

joshka (changes requested):
Hey, this is cool. A few things to address and some general comments / suggestions.

joshka:
Some git notes to add:
- We use conventional commits see https://github.com/ratatui-org/ratatui/blob/main/CONTRIBUTING.md#use-conventional-commits and https://github.com/ratatui-org/ratatui/blob/main/CONTRIBUTING.md#clean-up-your-commits and https://cbea.ms/git-commit/ for some guidelines.

Think about what you'd like the CHANGELOG entry for this to look like when done and aim to make that the body, with a short title that describes the new feature (e.g. `feat(table)!: add cells / column selection`)

src/widgets/table.rs
```diff
@@ -438,10 +500,22 @@ impl<'a> Styled for Table<'a> {
     }
 }
 
+#[derive(Debug, Default, Copy, Clone, Eq, PartialEq, Hash)]
+pub struct TableSelection {
+    pub row: usize,
+    pub col: usize,
+}
```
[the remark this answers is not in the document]
tonogdlp: Yes! this would be better! Will do as soon as I have a chance!

src/widgets/table.rs
```diff
@@ -453,7 +527,7 @@ impl TableState {
         &mut self.offset
     }
 
-    pub fn with_selected(mut self, selected: Option<usize>) -> Self {
+    pub fn with_selected(mut self, selected: Option<TableSelection>) -> Self {
```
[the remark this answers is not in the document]
tonogdlp: This was what I had done at first, but if I remember correctly `cargo make` made me change it, though I'm not sure why... That is why `TableSelection` has a `From<usize>` impl, but it is not actually used.  I'll try to put it back

src/widgets/table.rs
```diff
@@ -553,20 +630,31 @@ impl<'a> StatefulWidget for Table<'a> {
...
-                buf.set_stringn(
-                    inner_offset,
-                    row,
-                    highlight_symbol,
-                    table_area.width as usize,
-                    table_row.style,
-                );
-            };
-            for ((x, width), cell) in columns_widths.iter().zip(table_row.cells.iter()) {
+
```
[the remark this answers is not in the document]
tonogdlp: I'm sorry about this.. It's my first open source contribution so I'm not really familiar with the best practices. In the case of this method, I did managed to make it smaller but it was, in my opinion, a lot more convoluted and less understandable what was happening. That's why I tried making clear what each branch was doing (HighlightArea::Cell, vs Row, vs etc.).. 
But as it occurs when you give it a look with fresh eyes, I can now see a lot of code that can be refactored into something more "human readable". I will update this! Thanks for the input.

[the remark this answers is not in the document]
joshka: Welcome to ratatui! No need for apologies. We definitely appreciate all contributions :)

src/widgets/table.rs
tonogdlp: I thought about this for a while, and decided that `RowAndCol` was more explicit, and thus better IMO. I came to this conclusion when I thought about what "HighlightArea::Both" would look like if seen out of context, vs "HighlightArea::RowAndCol".. 

I also thought about more values, and while adding a #[non_exhaustive] should shield us from having to break code in the future, I also feel that we could decide now if we want more options.

Other than `None`, `Cell`, `Row`, `Col`, and `RowAndCol`, I thought about:

- Row without Cell  
- Col without Cell
- RowAndCol without Cell

But none of them seem that relevant to me, and would make the implementation considerably more convoluted, and so I dropped them. If you think these or any other values should be added, I can definitely work on it!

joshka: Sounds like you chose right already with RowAndCol. I'm going to resolve this one for now.

tonogdlp:
.... sigh..  
Yeah.. that wasn't what I meant to do.. 





> ```shell
> git reset --hard 37d874d
> git push --force --no-verify origin main
> ```

Will this PR be opened again after this? Or should I do a new PR instead?




> Take a look at https://ohshitgit.com/ or https://sukima.github.io/GitFixUm/ if you need some really good git recovery tips.

Thank you for the resources!

joshka:
> Will this PR be opened again after this? Or should I do a new PR instead?

I think it should do automatically (not 100% sure)

joshka:
Yep.

PR #484, opened 2023-09-10
chore: Prevent PR merge with `do not merge` labels 

kdheepak:
This PR prevents merges if a `do not merge` label is added on GitHub.

This PR also fixes an issue that allows checking PR titles with backticks in the title string.

### DIFF
--- a/.github/workflows/check-pr.yml
+++ b/.github/workflows/check-pr.yml
@@ -7,7 +7,8 @@ on:
       - edited
       - synchronize
   merge_group:
-
+  pull_request:
+    types: [synchronize, opened, reopened, labeled, unlabeled]
 
 permissions:
   pull-requests: write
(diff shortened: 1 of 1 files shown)


kdheepak:
With the current title of this PR, the check-breaking-change-label step will fail; but this PR will fix it and going forward this won't be an issue. So this PR can be merged even if that step fails.

joshka (approved):

.github/workflows/check-pr.yml
```diff
@@ -49,9 +50,9 @@ jobs:
     - name: Check breaking change label
       id: check_breaking_change
       run: |
-        title="${{ github.event.pull_request.title }}"
         pattern='^(build|chore|ci|docs|feat|fix|perf|refactor|revert|style|test)(\(\w+\))?!:'
-        if echo "$title" | grep -qE "$pattern"; then
+        # Check if pattern matches
+        if echo "${{ github.event.pull_request.title }}" | grep -qE "$pattern"; then
```
orhun, line 55: We could probably do something like this here:

```
if [[ "${{ github.event.pull_request.title }}" =~ $pattern ]]; then
```

But I'm too late for the review, can you guys wait a bit before merge?

kdheepak: Sorry about that! Next time I'll wait longer for merges.

PR #486, opened 2023-09-10
docs(terminal): add docs for terminal module

joshka:
- moves the impl Terminal block up to be closer to the type definition

Pulls in doc changes from #280 in an attempt to simplify that PR (more to come for other types).

### DIFF
--- a/src/terminal.rs
+++ b/src/terminal.rs
@@ -1,3 +1,34 @@
+#![deny(missing_docs)]
+//! Provides the [`Terminal`], [`Frame`] and related types.
+//!
+//! The [`Terminal`] is the main interface of this library. It is responsible for drawing and
+//! maintaining the state of the different widgets that compose your application.
+//!
+//! The [`Frame`] is a consistent view into the terminal state for rendering. It is obtained via
+//! the closure argument of [`Terminal::draw`]. It is used to render widgets to the terminal and
+//! control the cursor position.
+//!
+//! # Example
+//!
+//! ```rust,no_run
+//! use std::io::stdout;
+//! use ratatui::{prelude::*, widgets::Paragraph};
+//!
+//! let backend = CrosstermBackend::new(stdout());
+//! let mut terminal = Terminal::new(backend)?;
+//! terminal.draw(|frame| {
+//!     let area = frame.size();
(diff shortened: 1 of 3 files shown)


joshka:
- added no_run to fix failing CI

src/terminal.rs
```diff
@@ -32,12 +81,48 @@ pub struct TerminalOptions {
...
+/// state of the buffers, cursor and viewport.
+///
+/// The [`Terminal`] is generic over a [`Backend`] implementation which is used to interface with
+/// the underlying terminal library. The [`Backend`] trait is implemented for three popular Rust
+/// terminal libraries: [Crossterm], [Termion] and [Termwiz]. See the [`backend`] module for more
+/// information.
+///
+/// The terminal has two buffers that are used to draw the application. The first buffer is the
+/// current buffer and the second buffer is the previous buffer. The two buffers are compared at
+/// the end of each draw pass to output the necessary updates to the terminal.
```
kdheepak, line 96:
```suggestion
/// the end of each draw pass to output the only the changes to the terminal.
```

This seems like good context to mention the on a terminal resize event, the developer using ratatui should clear and render everything. Also that there's a `Clear` widget that lets a developer clear an `Rect` area.

src/terminal.rs
```diff
@@ -498,6 +537,132 @@ fn compute_inline_size<B: Backend>(
     ))
 }
 
+/// A consistent view into the terminal state for rendering a single fram.
```
kdheepak, line 540:
```suggestion
/// A consistent view into the terminal state for rendering a single frame.
```

kdheepak (approved):

src/terminal.rs
joshka: The draw() method calls `autoresize()` to handle this, but the app should call draw. I added this to the doc.

src/terminal.rs
joshka: fixed in https://github.com/ratatui-org/ratatui/commit/0d6bb1073447c18ac0a9be4f7319ccb1f3d4968b

kdheepak (approved):

joshka:
One last rebase for conflicts -whew!

kdheepak (approved):

PR #487, opened 2023-09-10
fix(Layout): layout cache does not grow without bounds

### LINKED ISSUE #402: Ensure that layout cache does not grow without bounds
## Problem
`Layout::split()` stores the output in a `HashMap` cache, keyed on the layout and area. The cache is never cleared.
This may cause higher than necessary memory usage, especially in cases with dynamic layouts for widgets etc.

## Solution
Replace the HashMap with a proper cache (perhaps https://crates.io/crates/lru) that supports cache eviction based on something reasonable (e.g. time / size).
- [ ] What is the reasonable time / size constraint?

## Alternatives
There may be standard caching patterns / idioms / crates for apps that avoid the cache being hard coded as a singleton.
- [ ] What do other apps / libs do for caches?

## Additional context
https://github.com/ratatui-org/ratatui/blob/519509945be866c3b2f6a4230ee317262266f894/src/layout.rs#L143-L146
https://github.com/ratatui-org/ratatui/blob/519509945be866c3b2f6a4230ee317262266f894/src/layout.rs#L263-L272

marianomarciello:
This pull request avoids the growth of Layout cache without any bounds.
The Layout cache is now using an LruCache ([doc](https://docs.rs/lru/latest/lru/)).
The current default cache size is set to 16. It seems reasonable for me to keep this value low.
A user can set the cache size during the Layout instantiation.

fix #402

### DIFF
--- a/src/layout.rs
+++ b/src/layout.rs
@@ -331,8 +331,6 @@ pub struct Layout {
     constraints: Vec<Constraint>,
     /// option for segment size preferences
     segment_size: SegmentSize,
-    // cache size
-    cache_size: usize,
 }
 
 impl Default for Layout {
@@ -342,13 +340,13 @@ impl Default for Layout {
 }
 
 impl Layout {
+    const DEFAULT_CACHE_SIZE: usize = 16;
     /// Creates a new layout with default values.
     ///
     /// - direction: [Direction::Vertical]
     /// - margin: 0, 0
     /// - constraints: empty
     /// - segment_size: SegmentSize::LastTakesRemainder
-    /// - cache_size: 16
     pub const fn new() -> Layout {
         Layout {
             direction: Direction::Vertical,
@@ -358,8 +356,28 @@ impl Layout {
             },
             constraints: Vec::new(),
             segment_size: SegmentSize::LastTakesRemainder,
-            cache_size: 16,
         }
+    }
+
+    /// Initialize an empty cache with a custom size. The cache is keyed on the layout and area, so
+    /// that subsequent calls with the same parameters are faster. The cache is a LruCache, and
+    /// grows until `cache_size` is reached.
+    ///
+    /// Returns true if the cell's value was set by this call.
+    /// Returns false if the cell's value was not set by this call, this means that another thread
+    /// has set this value or that the cache size is already initialized.
+    ///
+    /// Note that a custom cache size will be set only if this function:
+    /// * is called before [Layout::split()] otherwise, the cache size is `DEFAULT_CACHE_SIZE`.
+    /// * is called for the first time, subsequent calls do not modify the cache size.
+    pub fn init_cache(cache_size: usize) -> bool {
+        LAYOUT_CACHE
+            .with(|c| {
+                c.set(RefCell::new(LruCache::new(
+                    NonZeroUsize::new(cache_size).unwrap(),
+                )))
+            })
+            .is_ok()
     }
 
     /// Builder method to set the constraints of the layout.
@@ -481,7 +499,8 @@ impl Layout {
     ///
     /// This method stores the result of the computation in a thread-local cache keyed on the layout
     /// and area, so that subsequent calls with the same parameters are faster. The cache is a
-    /// LruCache, and grows until cache_size is reached.
+    /// LruCache, and grows until `DEFAULT_CACHE_SIZE` is reached by default, if the cache is
+    /// initialized with the [Layout::init_cache()] grows until the initialized cache size.
     ///
     /// # Examples
     ///
@@ -502,7 +521,9 @@ impl Layout {
     pub fn split(&self, area: Rect) -> Rc<[Rect]> {
         LAYOUT_CACHE.with(|c| {
             c.get_or_init(|| {
-                RefCell::new(LruCache::new(NonZeroUsize::new(self.cache_size).unwrap()))
+                RefCell::new(LruCache::new(
+                    NonZeroUsize::new(Self::DEFAULT_CACHE_SIZE).unwrap(),
+                ))
             })
             .borrow_mut()
             .get_or_insert((area, self.clone()), || split(area, self))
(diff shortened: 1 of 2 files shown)


joshka (approved):
LGTM - happy to merge this with or without the suggestions.

src/layout.rs
```diff
@@ -352,6 +359,7 @@ impl Layout {
             },
             constraints: Vec::new(),
             segment_size: SegmentSize::LastTakesRemainder,
+            cache_size: 16,
```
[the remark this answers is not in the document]
marianomarciello: Hi @joshka , thanks for your time.

In the next commits I will change the cache_size from a field into a const.
I have also fixed the format issue stated [here](https://github.com/ratatui-org/ratatui/pull/487#pullrequestreview-1618923991)

Thanks

src/layout.rs
```diff
@@ -474,7 +482,7 @@ impl Layout {
     ///
     /// This method stores the result of the computation in a thread-local cache keyed on the layout
     /// and area, so that subsequent calls with the same parameters are faster. The cache is a
-    /// simple HashMap, and grows indefinitely (<https://github.com/ratatui-org/ratatui/issues/402>).
+    /// LruCache, and grows until cache_size is reached.
```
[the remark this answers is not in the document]
marianomarciello: Thanks again for your time.

The `DEFAULT_CACHE_SIZE` constant will be mentioned in the documentation.

src/layout.rs
```diff
@@ -355,6 +359,27 @@ impl Layout {
...
+    /// Initialize an empty cache with a custom size. The cache is keyed on the layout and area, so
+    /// that subsequent calls with the same parameters are faster. The cache is a LruCache, and
+    /// grows until `cache_size` is reached.
+    ///
+    /// Returns true if the cell's value was set by this call.
+    /// Returns false if the cell's value was not set by this call, this means that another thread
+    /// has set this value or that the cache size is already initialized.
+    ///
+    /// Note that a custom cache size will be set only if this function:
+    /// * is called before [Layout::split()] otherwise, the cache size is `DEFAULT_CACHE_SIZE`.
```
joshka, line 371: The following will link to the const correctly:
```suggestion
    /// * is called before [Layout::split()] otherwise, the cache size is [`Self::DEFAULT_CACHE_SIZE`].
```

src/layout.rs
```diff
@@ -337,6 +340,7 @@ impl Default for Layout {
 }
 
 impl Layout {
+    const DEFAULT_CACHE_SIZE: usize = 16;
```
joshka, line 343: To make this appear in the rendered docs:
```suggestion
    pub const DEFAULT_CACHE_SIZE: usize = 16;
```
(you can run `cargo doc --open` to view the docs locally)

src/layout.rs
```diff
@@ -474,7 +499,8 @@ impl Layout {
     ///
     /// This method stores the result of the computation in a thread-local cache keyed on the layout
     /// and area, so that subsequent calls with the same parameters are faster. The cache is a
-    /// simple HashMap, and grows indefinitely (<https://github.com/ratatui-org/ratatui/issues/402>).
+    /// LruCache, and grows until `DEFAULT_CACHE_SIZE` is reached by default, if the cache is
```
joshka, line 502:
```suggestion
    /// LruCache, and grows until [`Self::DEFAULT_CACHE_SIZE`] is reached by default, if the cache is
```

joshka (changes requested):
Code is great.
Docs look good - thanks for adding these.

A couple of minor changes to make the docs render the DEFAULT_CACHE_SIZE. You might also need to rewrap the doc text if these changes are made. Check the VSCode [Rewrap](https://marketplace.visualstudio.com/items?itemName=stkb.rewrap) extension or configure your IDE / editor to wrap at 100 chars.

src/layout.rs
marianomarciello: Hi @joshka 

Thanks again for your time. I will improve the documentation as suggested.

joshka: You're welcome. Thanks for the changes. This looks good now. I'll recommend the other maintainers take a second review, but generally LGTM.

joshka (approved):
Thanks again for your contribution.

src/layout.rs
```diff
@@ -355,6 +359,28 @@ impl Layout {
...
+    ///
+    /// Returns true if the cell's value was set by this call.
+    /// Returns false if the cell's value was not set by this call, this means that another thread
+    /// has set this value or that the cache size is already initialized.
+    ///
+    /// Note that a custom cache size will be set only if this function:
+    /// * is called before [Layout::split()] otherwise, the cache size is
+    ///   [`Self::DEFAULT_CACHE_SIZE`].
+    /// * is called for the first time, subsequent calls do not modify the cache size.
+    pub fn init_cache(cache_size: usize) -> bool {
```
kdheepak, line 374: What's the reason this is `pub`? Are we expecting users to call `init_cache`?

marianomarciello: Users can call `init_cache` explicitly, so we can't embed this call into `new`.
Since this cache is not bound to a  `Layout` instance but instead is shared between all the `Layout` instances, it makes sense for me to let this method `pub`.

orhun (approved):

joshka:
I rebased on main, squashed and updated the commit message so it would look nice in the Changelog:

```
fix(Layout): use LruCache for layout cache (#487)

The layout cache now uses a LruCache with default size set to 16 entries.
Previously the cache was backed by a HashMap, and was able to grow
without bounds as a new entry was added for every new combination of
layout parameters.

- Added a new method (`layout::init_cache(usize)`) that allows the cache
size to be changed if necessary. This will only have an effect if it is called
prior to any calls to `layout::split()` as the cache is wrapped in a `OnceLock`
```

Thanks for submitting this PR. We appreciate your contribution.

Issue #448, opened 2023-08-28
`BarChart` example in `docs.rs` cannot be compiled


fadeevab:
## Problem

This line causes a compilation error: https://github.com/ratatui-org/ratatui/blob/3f781cad0ad7e0d6284be0af8dc6c1e0ba7e71b4/src/widgets/barchart/mod.rs#L24

## Measure
The following syntax should be used instead (as in the actual `BarChart` example):
```rust
.bar_style(Style::default().fg(Color::Yellow))
```

Valentin271:
Modifying the `BarChart` example to the following compiles fine on my machine.

```rust
    let barchart = BarChart::default()
        .block(Block::default().title("Data1").borders(Borders::ALL))
        .data(&app.data)
        .bar_width(9)

        .bar_style(Style::new().yellow().on_red())

        .value_style(Style::default().fg(Color::Black).bg(Color::Yellow));
    f.render_widget(barchart, chunks[0]);
```

![image](https://github.com/ratatui-org/ratatui/assets/36198422/df0a1184-3635-45e0-9d4b-e26bed578314)

---

Is the compilation failing on docs.rs or on your machine?

If it fails on you machine, did you include the [`Stylize`](https://docs.rs/ratatui/0.23.0/ratatui/style/trait.Stylize.html) trait or the Ratatui prelude?

fadeevab:
@Valentin271 Right, I haven't included the `Stylize` trait, I didn't know it existed. The best thing would be to make the doc examples compilable.

joshka:
> @Valentin271 Right, I haven't included the Stylize trait, I didn't know it existed. The best thing would be to make the doc examples compilable.

This comment is interesting because it suggests a couple of things (and that the [Curse of Knowledge](https://en.wikipedia.org/wiki/Curse_of_knowledge) is present):
- The style shorthand approach (implemented via the `Stylize` trait) is not discoverable. https://docs.rs/ratatui/latest/ratatui/style/#using-style-shorthands is the current docs for this.
- Although we have a tendency to use the [prelude](https://docs.rs/ratatui/latest/ratatui/prelude/index.html) for examples, this isn't something that is necessarily discoverable or obvious either. (The `Stylize` trait is included in the prelude).

[...]

PR #490, opened 2023-09-11
docs: import prelude::* in doc examples

joshka:
This commit adds `prelude::*` all doc examples and widget::* to those
that need it. This is done to highlight the use of the prelude and
simplify the examples.

- Examples in Type and module level comments show all imports and use
  `prelude::*` and `widget::*` where possible.
- Function level comments hide imports unless there are imports other
  than `prelude::*` and `widget::*`.

Fixes https://github.com/ratatui-org/ratatui/issues/448 by ensuring that module and type level code always contains all the imports required to compile the example.

### DIFF
--- a/src/title.rs
+++ b/src/title.rs
@@ -13,22 +13,22 @@ use crate::{layout::Alignment, text::Line};
 ///
 /// Title with no style.
 /// ```
-/// # use ratatui::widgets::block::Title;
+/// use ratatui::widgets::block::Title;
+///
 /// Title::from("Title");
 /// ```
 ///
 /// Blue title on a white background (via [`Stylize`](crate::style::Stylize) trait).
 /// ```
-/// # use ratatui::widgets::block::Title;
-/// # use ratatui::style::Stylize;
+/// use ratatui::{prelude::*, widgets::block::*};
+///
 /// Title::from("Title".blue().on_white());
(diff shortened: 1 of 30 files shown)


src/title.rs
```diff
@@ -13,31 +13,31 @@ use crate::{layout::Alignment, text::Line};
...
 /// Title::from(
 ///     Line::from(vec!["Q".white().underlined(), "uit".gray()])
 /// );
 /// ```
 ///
 /// Complete example
 /// ```
-/// # use ratatui::widgets::block::{Title, Position};
-/// # use ratatui::layout::Alignment;
+/// use ratatui::{prelude::*, widgets::{*, block::*}};
```
kdheepak, line 39: `Title` is fully qualified in the above two changes but implicitly imported here?

kdheepak (approved):

src/title.rs
joshka: Good catch - I've fixed this in a few places just now.
Title is a weird one in terms of where it's defined (only affects block, but is defined outside of block and is a little too generic a name to put in the prelude nicely.

joshka:
Rebased on the other doc changes to fix conflicts

kdheepak (approved):

PR #505, opened 2023-09-15
docs(table): add documentation for `block` and `header` methods of the `Table` widget

DreadedHippy:
Additionally, add `fix(docs): don't fail rustdoc due to termion`


joshka (approved):
LGTM - thanks for the update

PR #511, opened 2023-09-17
feat(line): add `Line::raw` constructor

### LINKED ISSUE #508: Add `Line::raw` for symmetry with `Span::raw` and `Text::raw`
## Problem

There's already a `Span::raw` and `Text::raw` but we seem to be missing a `Line::raw`.

orhun:
closes #508

### DIFF
--- a/src/text/line.rs
+++ b/src/text/line.rs
@@ -25,7 +25,11 @@ impl<'a> Line<'a> {
         T: Into<Cow<'a, str>>,
     {
         Line {
-            spans: vec![Span::raw(content)],
+            spans: content
+                .into()
+                .lines()
+                .map(|v| Span::raw(v.to_string()))
+                .collect(),
             alignment: None,
         }
     }
@@ -347,6 +351,10 @@ mod tests {
     fn raw_str() {
         let line = Line::raw("test content");
         assert_eq!(line.spans, vec![Span::raw("test content")]);
+        assert_eq!(line.alignment, None);
+
+        let line = Line::raw("a\nb");
+        assert_eq!(line.spans, vec![Span::raw("a"), Span::raw("b")]);
         assert_eq!(line.alignment, None);
     }
 }
--- a/src/text/line.rs
+++ b/src/text/line.rs
@@ -11,6 +11,25 @@ pub struct Line<'a> {
 }
 
 impl<'a> Line<'a> {
+    /// Create a line with the default style.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// # use ratatui::prelude::*;
+    /// Line::raw("test content");
+    /// Line::raw(String::from("test content"));
+    /// ```
+    pub fn raw<T>(content: T) -> Line<'a>
+    where
+        T: Into<Cow<'a, str>>,
+    {
+        Line {
+            spans: vec![Span::raw(content)],
+            alignment: None,
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


orhun:
Question: should we break newlines into different spans? e.g. `test\ntest` will result in `vec![Span, Span]` etc.

joshka:
There are a few ways we could handle newlines in `Line`:
1. reject them completely (panic and consider adding some `try_` methods)
2. remove them during creation and create a single `Span` that has a single `Cow::Owned` constructed from the input with the lines removed.
3. remove them and create multiple `Span`s that each hold `Cow::Borrowed` slices.
4. ignore them and handle it on render (I don't like this approach as it means every thing that takes Line now has to understand that it might have newlines).

Option 3 makes most sense to me, but Option 1 is also possible. Option 2 allocates new memory, but is probably a reasonable approach and "performant enough" except for massive strings. Option 4 is a pretty terrible one.

kdheepak:
I think I have some naming + semantic questions about what we want `Line` to be?

From Wiki

> In computing, a line is a unit of organization for text files. A line consists of a sequence of zero or more characters, usually displayed within a single horizontal sequence.
> The term comes directly from physical printing, where a line of text is a horizontal row of characters. 

`Line` currently uses `Vec<Spans>` as the implementation. But it is only during rendering that we decide where the "end of a line" should be. Right? If someone wants to store this content in a Paragraph:

```
This is a sentence that
wraps to the next line.
This is another sentence that doesn't wrap to the next line
```

Do they make 3 lines to convert them to a `Text` object?
Do they make 2 lines with the first line having a `\n` in it? Should the first line split it into 2 spans? What if they want `This` to be bold but the rest not? Then the first line would have 3 `Spans` because `Span` is the lowest unit of styling.

orhun:
I think @joshka's 3rd option makes the most sense. If everyone agrees then I can update the PR.

joshka:
> `Line` currently uses `Vec<Spans>` as the implementation. But it is only during rendering that we decide where the "end of a line" should be. Right? If someone wants to store this content in a Paragraph:

I think it's reasonable that Line would have a `fn wrapped(params) -> impl Iterator<Type = Line<'a>>` method that helps make the distinction more obvious.

> Do they make 3 lines to convert them to a Text object?
Do they make 2 lines with the first line having a \n in it? Should the first line split it into 2 spans? What if they want This to be bold but the rest not? Then the first line would have 3 Spans because Span is the lowest unit of styling.

3 lines `Text::raw(content: T: Into<Cow<'a, str>>)` method will handle Text::raw("Line1\nLine2\nLine3") by creating 3 lines each with a single span:
https://github.com/ratatui-org/ratatui/blob/5498a889ae8bd4ccb51b04d3a848dd2f58935906/src/text/text.rs#L45-L57

So at the least for consistency, Line should ever have a newline character. This also avoids having every widget that takes a line parameter have to work manually handle it (e.g. what should `buffer.cell_mut().symbol = "\n"` do? I'm sure there are other arguments as well.

joshka (changes requested):
Please add newline handling (`"a\nb"` becomes `Line::from(vec![Span::raw("a"), Span::raw("b")])`)

orhun:
Updated in 9d4db841a334a2bba748a4f056c38a87dfa15a6f

joshka (approved):

PR #515, opened 2023-09-18
fix(barchart): fix(barchart): use width() instead of len() to get unicode string length

### LINKED ISSUE #513: Barchart doesn't render labels in entire width
## Description

The barchart widget doesn't render labels that are the full width (this affects both non-unicode and unicode)

## To Reproduce

```rust
fn render_simple_barchart(area: Rect, buf: &mut Buffer) {
    let data = [
        ("Sat", 76),
        ("Sun", 69),
        ("Mon", 65),
        ("Tue", 67),
        ("Wed", 65),
        ("Thu", 69),
        ("Fri", 73),
    ];
    let data = data
        .into_iter()
        .map(|(label, value)| {
            Bar::default()
                .value(value)
                // cannot use this until the column width bug is fixed
                .text_value(format!("{}°", value))
                .label(label.into())
        })
        .collect_vec();
    let group = BarGroup::default().bars(&data);
    BarChart::default()
        .data(group)
        .bar_width(3)
        .bar_gap(1)
        .render(area, buf);
}
```

image

with the following line removed:

```rust
                .text_value(format!("{}°", value))
```
image

And the same occurs replacing that with 
```rust
                .text_value(format!("{}x", value))
```

## Expected behavior
I think this should probably render like the following (hacking a couple of things in the barchart implementation to make it work):

image

## Environment
Mac/iterm/FiraCode(NF)/main/crossterm

## Additional context
- Is this expected behavior for some historical reason?
- If this is expected, how should it be documented?
- What significance should unicode play here (e.g. the degree symbol has string len = 2 and unicode width = 1)? Is #75 relevant to the degree symbol?
- What is the history of this? Check for any guidance from commits, and comments on PRs/ issues in Ratatui and tui-rs.

karthago1:
value_label.len() deliver byte length of the unicode strings, which are
not printed as a result.

use witdh() instead.

### DIFF
--- a/src/widgets/barchart/bar.rs
+++ b/src/widgets/barchart/bar.rs
@@ -1,3 +1,5 @@
+use unicode_width::UnicodeWidthStr;
+
 use crate::{buffer::Buffer, prelude::Rect, style::Style, text::Line};
 
 /// A bar to be shown by the [`BarChart`](crate::widgets::BarChart) widget.
@@ -154,7 +156,7 @@ impl<'a> Bar<'a> {
                 self.value.to_string()
             };
 
-            let width = value_label.len() as u16;
+            let width = value_label.width() as u16;
             if width < max_width {
                 buf.set_string(
                     x + (max_width.saturating_sub(value_label.len() as u16) >> 1),
--- a/src/widgets/barchart.rs
+++ b/src/widgets/barchart.rs
@@ -1149,4 +1149,35 @@ mod tests {
         let expected = Buffer::with_lines(vec!["  █", "▆ █", "  G"]);
         assert_buffer_eq!(buffer, expected);
     }
+
+    #[test]
+    fn test_unicode_as_value() {
+        let group = BarGroup::default().bars(&[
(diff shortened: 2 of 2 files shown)


src/widgets/barchart/bar.rs
```diff
@@ -155,7 +155,7 @@ impl<'a> Bar<'a> {
             };
 
             let width = value_label.len() as u16;
-            if width < max_width {
+            if width <= max_width {
```
kdheepak, line 158: I think this is not going to do the right thing in all instances?

For example, can you add a test where one of the labels is `"🤦🏼‍♂️".into()`?

karthago1: I reverted the comparison back to `<` , but I am using width() instead

joshka (approved):
LGTM - we can merge this and work out the rest of the problem in #513 with small values.

joshka:
I updated the commit message a bit to make it more Changelog friendly and merged this. Thanks for the PR @karthago1

PR #518, opened 2023-09-19
fix(Barchart): horizontal labels were not displayed

### LINKED ISSUE #499: Horizontal Barchart does not print labels
## Description
Horizontal Barcharts were added recently (0.23.0), but labels were not implemented.
Barcharts have two label types: label and text_value (a formatted version of the value), only the latter is rendered.

## To Reproduce
```rust
let data = vec![
    ("Jan", 10),
    ("Feb", 20),
    ("Mar", 30),
    ("Apr", 40),
    ("May", 50),
    ("Jun", 60),
];
BarChart::default()
    .data(&data)
    .render(area, buf);
```

```plain
10█████

20█████████████

30█████████████████████

40████████████████████████████

50████████████████████████████████████

60████████████████████████████████████████████
```

## Expected behavior
```plain
Jan 10█████

Feb 20█████████████

Mar 30█████████████████████

Apr 40████████████████████████████

May 50████████████████████████████████████

Jun 60████████████████████████████████████████████
```

## Environment
version 0.23-alpha (main branch as of today)

Valentin271:
Fixed horizontal bars label not displayed.
Also adds ways to configure label truncation because it would be hard to use labels horizontally otherwise.

This fixes #499

### DIFF
--- a/src/widgets/barchart.rs
+++ b/src/widgets/barchart.rs
@@ -335,6 +335,26 @@ impl<'a> BarChart<'a> {
     }
 
     fn render_horizontal_bars(self, buf: &mut Buffer, bars_area: Rect, max: u64) {
+        // get the longest label
+        let label_size = self
+            .data
+            .iter()
+            .flat_map(|group| group.bars.iter().map(|bar| &bar.label))
+            .flatten() // bar.label is an Option<Line>
+            .map(|label| label.width())
+            .max()
+            .unwrap_or(0) as u16;
+
+        let label_x = bars_area.x;
+        let bars_area = {
+            let margin = if label_size == 0 { 0 } else { 1 };
+            Rect {
+                x: bars_area.x + label_size + margin,
+                width: bars_area.width - label_size - margin,
+                ..bars_area
+            }
+        };
+
         // convert the bar values to ratatui::symbols::bar::Set
         let groups: Vec<Vec<u16>> = self
             .data
@@ -348,7 +368,7 @@ impl<'a> BarChart<'a> {
             })
             .collect();
 
-        // print all visible bars
+        // print all visible bars, label and values
         let mut bar_y = bars_area.top();
         for (group_data, mut group) in groups.into_iter().zip(self.data) {
             let bars = std::mem::take(&mut group.bars);
@@ -374,6 +394,12 @@ impl<'a> BarChart<'a> {
                     y: bar_y + (self.bar_width >> 1),
                     ..bars_area
                 };
+
+                // label
+                if let Some(label) = &bar.label {
+                    buf.set_line(label_x, bar_value_area.top(), label, label_size);
+                }
+
                 bar.render_value_with_different_styles(
                     buf,
                     bar_value_area,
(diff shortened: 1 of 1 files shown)


src/widgets/barchart.rs
```diff
@@ -335,6 +347,32 @@ impl<'a> BarChart<'a> {
...
+                        .unwrap_or(0)
+                })
+                .max()
+                .unwrap_or(0) as u16
+        };
+
+        let margin = if label_size == 0 { 0 } else { 1 };
+
+        let bars_area = Rect {
+            x: bars_area.x + label_size + margin,
```
karthago1, line 371: I would save `bar_area.x` to use it later for the labels instead of calculating it again inside for loops.
`let label_x = bars_area.x;`

src/widgets/barchart.rs
```diff
@@ -374,6 +412,15 @@ impl<'a> BarChart<'a> {
                     y: bar_y + (self.bar_width >> 1),
                     ..bars_area
                 };
+
+                // label
+                buf.set_line(
```
karthago1, line 417: I would check for the label first to avoid calling `set_line`.
also you can avoid cloning the label

```suggestion
                if let Some(label) = &bar.label {
                    buf.set_line(label_x, bar_value_area.top(), label, label_size);
                }
```

Valentin271:
Incorporated feedbacks in 57494660c05c2bdcebfb5044fe3e64c0321d753f. Thanks @karthago1!

src/widgets/barchart.rs
```diff
@@ -335,6 +335,31 @@ impl<'a> BarChart<'a> {
...
+            .unwrap_or(0) as u16;
+
+        let label_x = bars_area.x;
+        let bars_area = {
+            let margin = if label_size == 0 { 0 } else { 1 };
+            Rect {
+                x: bars_area.x + label_size + margin,
+                width: bars_area.width - label_size - margin,
+                ..bars_area
+            }
```
joshka, line 360: I find myself often writing this sort of logic for a Rect in apps and code and have been trying to think of an ergonomic method name for it - what would you think about this being `Rect::inset(x, y)` (where `Rect::offset(x, y)` would not change the width and height?

joshka (changes requested):

src/widgets/barchart.rs
```diff
@@ -335,6 +335,31 @@ impl<'a> BarChart<'a> {
...
+            .map(|group| {
+                group
+                    .bars
+                    .iter()
+                    .map(|bar| bar.label.clone().map_or(0, |label| label.width()))
+                    .max()
+                    .unwrap_or(0)
+            })
+            .max()
+            .unwrap_or(0) as u16;
```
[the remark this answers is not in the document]
Valentin271: Yep much clearer and shorter. FYI, benches did not show any performance impact either way.

src/widgets/barchart.rs
Valentin271: If I understand correctly, this would end up to something like that.

```rust
let newRect = rect.inset(label_size + margin, 0);
```

Seems very useful and less error-prone. Naming looks good to me.

joshka (approved):
LGTM. Thanks for the PR

PR #519, opened 2023-09-19
fix(table): add default column widths for when none are supplied

### LINKED ISSUE #510: Make it more difficult to render a table without `widths` set
## Problem
 
Currently, the `Table` widget initialises with an empty `widths` field. If the `widths` method isn't called before rendering, nothing is rendered in the table's allocated chunk. From an API standpoint, I find this confusing. I struggle to see a situation where not rendering any of the table's contents is desirable.

## Solution

I'm not sure that there's a way to implement this in a non-breaking way. The easiest thing to do would be to add a parameter for `widths` in the constructor of `Table`.

## Alternatives

I can think of a few other ways to implement my desired behaviour:
- Add a reasonable default for `widths`; something like each column taking up the same amount of room. I'm not sure this can be accomplished without changing `widths` to be an owned value.
- Perform an assertion before rendering that `widths` is non-empty.

If none of those changes are feasible given their breaking nature, it'd still be super nice to make the documentation clearer about the potential foot gun.

## Additional context
Nothing to add apart from I'd be very happy to implement this if approved.

joshcbrown:
Previously, if `.widths` was not called before rendering a `Table`, no content would render in the area of the table. This commit changes that behaviour to default to equal widths for each column.

Fixes #510.

### DIFF
--- a/src/widgets/table.rs
+++ b/src/widgets/table.rs
@@ -39,7 +39,7 @@ use crate::{
 ///
 /// # Setter methods
 ///
-/// These methods a fluent setters. They return a new `Table` with the specified property set.
+/// These methods are fluent setters. They return a new `Table` with the specified property set.
 ///
 /// - [`Table::rows`] sets the rows of the [`Table`].
 /// - [`Table::header`] sets the header row of the [`Table`].


src/widgets/table.rs
```diff
@@ -399,13 +399,22 @@ impl<'a> Table<'a> {
     /// Get all offsets and widths of all user specified columns
     /// Returns (x, width)
     fn get_columns_widths(&self, max_width: u16, selection_width: u16) -> Vec<(u16, u16)> {
-        let mut constraints = Vec::with_capacity(self.widths.len() * 2 + 1);
+        // Default to equal column widths if not provided
+        let widths = if self.widths.is_empty() {
+            let max_col_count = self.rows.iter().map(|r| r.cells.len()).max();
+            let header_col_count = self.header.as_ref().unwrap_or(&Row::default()).cells.len();
+            let col_count = max_col_count.unwrap_or(header_col_count);
+            vec![Constraint::Percentage(100 / col_count as u16); col_count]
```
joshka, line 407: One problem with using Percentage / Ratio is that it's a percentage of the entire space, not a percentage of the leftover space after allocations. This means this code will always over allocate space to the columns and have to choose one or more constraint to break. This can play weird with rendering expectations especially for small columns.

joshka (commented):
Comments may or may not be valid. Happy to merge this as is if not.

src/widgets/table.rs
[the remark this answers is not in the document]
joshcbrown: This behaviour makes most sense, so will do that instead. Thanks for the tip!

src/widgets/table.rs
```diff
@@ -399,13 +399,22 @@ impl<'a> Table<'a> {
...
         constraints.push(Constraint::Length(selection_width));
-        for constraint in self.widths {
+        for constraint in widths.iter() {
             constraints.push(*constraint);
             constraints.push(Constraint::Length(self.column_spacing));
         }
-        if !self.widths.is_empty() {
+        if !widths.is_empty() {
             constraints.pop();
         }
```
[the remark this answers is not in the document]
joshcbrown: Sure, that's certainly a lot cleaner.

src/widgets/table.rs
joshcbrown: Oh ok, interesting. Another option might be `Constraint::Length(max_width / col_count - col_count + 1)` in that case?

joshcbrown: On second thought, don't think this will work.

src/widgets/table.rs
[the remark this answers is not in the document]
joshcbrown: Just trying this now, here's what I have:
```rust
        let constraints: Vec<_> = once(Constraint::Length(selection_width))
            .chain(
                widths
                    .into_iter()
                    .intersperse(Constraint::Length(self.column_spacing)),
            )
            .collect();
```
This triggers E0658; apparently `.intersperse()` is unstable.

joshcbrown:
Not sure how I should squash the commit with changes from comments. Is it ok to force push after squashing?

src/widgets/table.rs
```diff
@@ -401,10 +401,15 @@ impl<'a> Table<'a> {
...
-            vec![Constraint::Percentage(100 / col_count as u16); col_count]
+            let max_col_count = self
+                .rows
+                .iter()
+                .chain(self.header.iter())
+                .map(|r| r.cells.len())
+                .max()
+                .unwrap_or(0);
+            let equal_length = max_width / max_col_count as u16 - 2 * self.column_spacing;
+            vec![Constraint::Length(equal_length); max_col_count]
```
joshka, line 412: Can you add some documentation that explains why this calculation works? Seems magical to subtract 2*column_spacing.

joshka (changes requested):
Cab you please add some docs to the width function and the struct that highlights this functionality?

You're fine to either add commits or squash them (we generally squash PRs when merging).

src/widgets/table.rs
joshcbrown: Definitely, though I'm just realising that it's actually incorrect. My thinking was that each column should share an equal portion of the `max_width`, minus the space that the column spacing takes up. The formula is wrong, though. I'll correct it and make it clearer.

src/widgets/table.rs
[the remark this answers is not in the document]
joshka: Are you using the Itertools version or the std::iter version of this?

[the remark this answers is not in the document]
joshcbrown: Ah, thanks. Missed it being a function rather than a method.

joshka:
The 2 places that need documentation are public methods (width) and the type (Table)  (the Table struct and the widths() method on the struct)

joshcbrown:
Ah, apologies.

src/widgets/table.rs
```diff
@@ -396,18 +405,28 @@ impl<'a> Table<'a> {
...
+        let widths = if self.widths.is_empty() {
+            let col_count = self
+                .rows
+                .iter()
+                .chain(self.header.iter())
+                .map(|r| r.cells.len())
+                .max()
+                .unwrap_or(0);
+            // There are `col_count - 1` spaces between the columns
+            let total_space = max_width - self.column_spacing * (col_count as u16 - 1);
```
joshka, line 421: This panics with overflow if `max_width < column_spacing * (col_count - 1)`.
Is there something we can do to avoid the panic?
If there's not, then we need a few tests and docs to show when the panicking is expected.

joshka (changes requested):
Thanks for adding the docs. I have a one more question regarding overflow.

src/widgets/table.rs
joshcbrown: Good call! In that case, I think it makes most sense to render nothing. A panic feels unreasonable. I'll try making that a saturating subtraction instead.

joshcbrown: Sorry for taking so long to respond, had to duck out for a few days. As I see it, there are multiple problems to tackle here:

1. One is doing a best effort to render something if `max_width <= self.column_spacing * (col_count - 1)`. 
2. Another is deciding what to do in the case that the sum of the widths of the columns (by which I mean the maximum unicode width entry in each column) exceeds `max_width`. 

I agree that rendering the maximum number `n` columns such that `max_width > self.column_spacing * (n - 1)` is preferable to the quick fix I did with saturating subtraction for 1.

I'm not so sure about whether it's right to drop columns in 2. To the user, it mightn't be clear that there are more columns if you were to drop the ones whose full width doesn't fit the remaining space. For example, in cases like `max_width=5` and a row that looks like `("aa", "bb", "cc")`, the choice between rendering `"aa bb"`, and `"a b c"` isn't super clear. In my view, it would be helpful to have the second rendering because it's immediately apparent that I just need to minimise the font a small amount/enlarge the window to view the full table.

I think the solution I proposed to 1. would also display the second behaviour in that example, which means the code would pretty simple. What do you think? Let me know if anything isn't clear, I might not have explained super well.

I also agree that a more visual test like the one you describe would be ideal.

joshka: If you write the visual tests how close do you think you'd get to being able to answer your questions?

There's also a reason I chose going to 15 there - for most things in software we care about how things work with 0, 1 and >1, so when truncating we want to extend that to something that truncates 0,1,>1 and leaves 0,1,>1 when truncated. So that means 3/4 for each item and 3 for the number of items. So then we look at the same values for the gaps and what happens when there isn't enough space 3*4+2(spaces)+1(selection) or 3*3+2*2(spaces)+1 etc.

joshka:
In https://github.com/ratatui-org/ratatui/pull/638 I rearranged and documented a bunch of the Table methods for top down readability, I think this PR will probably be impacted by that unfortunately, so I'd like to swing back to this once https://github.com/ratatui-org/ratatui/pull/638 is merged and get it over the line.

Valentin271:
@joshcbrown are you still interested to work on this PR now that #638 has been merged?

joshka:
ugh - sorry, I messed up the push by pushing my main branch over https://github.com/joshcbrown/ratatui/tree/main instead of the local joshcbrown/main branch - git needs to remove that footgun, I've done it a few times inadvertently
```shell
git push --force joshcbrown main:main
```
instead of 
```shell
git push --force joshcbrown joshcbrown/main:main 
```

PR #525, opened 2023-09-21
fix(barchart): divide by zero in render

joshka:
- fix(table): add default column widths for when none are supplied
- implement suggestions in comments
- add docs, correct length calculation
- simplify logic in constructing constraints
- add docs to public methods
- fix overflow panic
- fix(barchart): avoid divide by zero in rendering


orhun (approved):

joshka:
This wasn't meant to have a copy of the table branches in it, just the barchart. rebased

kdheepak (approved):

Issue #512, opened 2023-09-17
Docs - landing page Layout section is out of place -> not newcomer friendly


1oglop1:
## Description
As a newcomer to Rust and Ratatui, I go to the documentation page https://docs.rs/ratatui/latest/ratatui/
The first few lines show perfect examples of what I can do with the library, then follow the minimal working example.
But then the `Layout` section contains some function `ui()` which explains how elements are composed,
however, the function example is completely disconnected from the previous code block, which makes it hard to grasp what is going on without diving deeper into the code.

## Expected behavior
I'd expect the landing page to either provide a complete example or at least point me in the direction of examples. 
Pointing to examples should not just be a text, but rather a link to a repo or page which explains more details about how to use
the library.

joshka:
Thanks, this is some great feedback. Writing better docs is something that we're all working on (and would appreciate all the help you can spare).

If you haven't already found the following, these resources are available:
- The Readme: https://github.com/ratatui-org/ratatui#quickstart (has a more complete example)
- the Ratatui book: https://ratatui-org.github.io/ratatui-book/index.html
- the examples: https://github.com/ratatui-org/ratatui/tree/main/examples
- Several templates / samples:
  - https://github.com/ratatui-org/rust-tui-template
  - https://github.com/pmsanford/simple-tui-rs
  - https://github.com/ratatui-org/ratatui-async-template

Ideally these should be better cross linked in all the right places so concretely we need to:
- [ ] Compare the README and docs.rs main page and ensure that they are either consistent or link to each other appropriately (we could either make these exactly the same, or simplify one and point to the other, but we need to understand which bits are useful in each first). [cargo-rdme](https://crates.io/crates/cargo-rdme) is possibly something to consider here
- [ ] Ensure that mentions of "examples" actually link to the examples
- [ ] Ensure that we link to the Book more
- [ ] Highlight the available templates
- [ ] Ensure that the various docs page links all to the other docs (no dead ends)
- [ ] Document the examples more (this is a larger task)

Happy to hear more suggestions on this / take PRs if you're interested in helping out.

1oglop1:
Thanks! I will look into it.

I'd like to contribute back when I get a hang of the project. Because the examples look pretty good. It's just a pity they are a bit hidden. 
I am not a fan of a separate rust book because it is one more place to look at and maintain the parity. 

I hope that the suggested solution here will soon become a reality Because it's pretty much standard in other languages (e.g. python->Sphinx) 
[...]

PR #526, opened 2023-09-21
docs: make library and README consistent

joshka:
Generate the bulk of the README from the library documentation, so that
they are consistent using cargo-rdme.

- Removed the Contributors section, as it is redundant with the
  github contributors list.
- Removed the info about the other backends and replaced it with a
  pointer to the documentation.

Fixes: https://github.com/ratatui-org/ratatui/issues/512


joshka:
Screen Shot 2023-09-21 at 09 56 09

joshka:
ping @1oglop1

README.md
```diff
@@ -1,142 +1,333 @@
...
+use std::io;
+
+fn main() -> io::Result<()> {
+    let mut terminal = init_terminal()?;
+    let mut should_quit = false;
+    while !should_quit {
+        terminal.draw(ui)?;
+        should_quit = handle_events()?;
+    }
+    restore_terminal()?;
```
kdheepak, line 113: This feels like too much information for the landing page (or too little depending on how you look at it). 

I'd prefer to move all this content in the `Quickstart` section here into a `quickstart.md` section in the book, and link to it from this landing page. And either have no code in the landing page or the smallest "hello world" example.

In my opinion, quickstart sections are for "pattern recognition", i.e. what is the kind of code I'll have to write to use this crate. 

I'm curious to hear what others think too.

joshka: > This feels like too much information for the landing page (or too little depending on how you look at it).

I strongly disagree. When I go to a docs.rs page I want to understand how the major modules, types, traits and methods of a crate all fit together, and then jump off from there into the details. For reference level material to be effective, it needs to be well grounded with examples that help show why you're jumping to the other parts of the documentation. The least effective crate docs I've read have been those that don't take this step.

> I'd prefer to move all this content in the Quickstart section here into a quickstart.md section in the book, and link to it from this landing page. And either have no code in the landing page or the smallest "hello world" example.

I see the Ratatui book very much as a secondary source. Even when it has the best quality information I should be able to answer any reasonable question about how to use Ratatui using only the docs.rs for the version of the crate. Hard disagree on no code. And this is the smallest hello world that anyone should write (ignoring the layout and style sections)

> In my opinion, quickstart sections are for "pattern recognition", i.e. what is the kind of code I'll have to write to use this crate.

For ratatui, those critical things to understand for any app are initialization / restore, main loop, input handling, drawing the UI. Styling and layout round out the second tier of things which are perhaps not critical, but still necessary. Having a section and example method for each of those areas, with links to the rest of the docs and to the book seems "just right" to me (I'd neither want to put it all in a single file, or break it down any further)

I see tutorial sections in the book as teaching someone how to create an application, not for finding out the available modules and types that I need to use in Ratatui app. Because of this, they are low-signal high-noise pointers to reference docs. I have to read a lot of information before I find the information I need. This overview page should be the opposite of that (high-signal, low noise)

I chose the density of information pretty carefully in writing up the quickstart example here. Right now the existing docs.rs page:
- has too much detail on how to use alternate backends. replaced with a link
- Shows cargo.toml in addition to cargo add - unnecessary to show two ways of doing this
- Shows how to create a backend for multiple backends
- merges information about raw mode, alternate screen, and rendering the UI into one code block that is longer than a page. Shows some pretty questionable app choices
- jumps to layout in a way that doesn't link with the previous example
- doesn't cover the prelude, styling, style shortcuts (all of these are useful to call out as early as possible in the docs to build awareness)

My goals with the replacement doc was:
- avoid describing too much detail about why things work, preferring to show the options and jump to the pages that go into detail (for backends, etc.)
- Keep each example and section concise and focused rather than page long walls of text
- Label each section to make the docs easy to read for non-linear reading
- Ensure that we show `key_event.kind == KeyEventKind::KeyPress` to help avoid that question
- Show the prelude and style shortcuts to make user's code shorter and easier to read

I also really don't like jumping back and forth between two websites to get the information that I need to use a library. This is especially a problem as the life of the book website compared to the life of a docs.rs page is different. The docs for 0.24.0 will be the docs for 0.24.0 forever, the Ratatui book in a couple of years will probably be quite different.

In summary, I think this is probably goldilocks when it comes to docs, rather than too much / too little.

kdheepak: I think this PR is an improvement over the current state, which is great! But I'm just not convinced it is really necessary. 

> I also really don't like jumping back and forth between two websites to get the information that I need to use a library. 

I agree with this a lot and I think this is my main reasoning for having everything in the website/book. There's ways to point to different versions of the docs (e.g. different folders for different versions) if we wanted to do that.

I think of the docs.rs landing page similar to a README.md file, and I think such a page should just be the one stop shop to find all the links to useful information, not necessary provide the actual useful information. I also think this comes down to personal preference. 

If we are optimizing for new users, I don't like the idea of having code blocks that use functions that are used later. If I were reading this for the first time, I bet I'd be confused. Is `init_terminal()` a function provided by the `ratatui` crate? Why is it not imported? If I were new to rust, there'd be even more things that would confuse me: what's `ui<B: Backend>()`? Do I have to write scary code like that?

I also agree that we should avoid going into too much detail which is why I think the _smallest_ example possible would be best. This might just be me but most of the time when I'm scanning through the docs.rs landing page for a crate, I'm skimming through all the code in the page to get a sense for what the crate is about. All text surrounding the code is secondary to me when I'm just exploring. If I'm actually evaluating using it, I start looking through the detailed docs and the examples. The landing page is not useful in that regard.

See for example [`serde`'s README.md](https://github.com/serde-rs/serde/blob/ddc1ee564b33aa584e5a66817aafb27c3265b212/README.md) as an example of a short and sweet README.
It has a fully working example in 20 or so lines of code, including comments that show that you'd expect to get if you ran the code. 

Other landing pages also show a simple program from start to finish, and show the output when you run that program. [clap](https://docs.rs/clap/4.4.4/clap/) for example shows what happens when you run `demo --help`. I'd argue that with the current PR's quickstart section, it is not easy to predict what the output would be. It would benefit from a screenshot or the output copy pasted into a codeblock.

All that said, this is not something I feel too strongly about and I do like the content quickstart section from the PR :) I'm just not convinced it is doing the most useful work as part of the landing page on docs.rs. But in case someone doesn't care, they'll just scroll past it all, so it's fine with having it if you think it is important. I'm sure people read docs differently and we'd want to cater to all needs, and my perspective here might be biased :)

joshka:
With discussed tweaks (rebased on the frame PR):
Screen Shot 2023-09-22 at 17 08 43

README.md
```diff
@@ -1,142 +1,330 @@
...
-    terminal: &mut Terminal<CrosstermBackend<Stdout>>,
-) -> Result<(), Box<dyn Error>> {
     disable_raw_mode()?;
-    execute!(terminal.backend_mut(), LeaveAlternateScreen,)?;
-    Ok(terminal.show_cursor()?)
+    stdout().execute(LeaveAlternateScreen)?;
+    Ok(())
+}
+
+fn ui(frame: &mut Frame) {
```
kdheepak, line 155: What do you think about moving the `ui` defintion to the bottom of the code block? i.e. after the `handle_events()` definition. 

And also surrounding the remaining two code blocks with `fn ui(frame: &mut Frame) {`  and `}`? 

When I scrolled through the page and then scrolled back up, I think these these changes it'd be easier to spot the pattern for me (i.e. `ui` function at the bottom of the code block + whatever is inside the `ui` function is required to generate the contents in the screenshot).

joshka: I prefer it how it is (because a) natural reading order, and b) not wanting to double down on the ui method being the "pattern"), but not strongly enough to matter. https://github.com/ratatui-org/ratatui/pull/526/commits/2eef36bf273adcd5293d5308c82e8f577b341d20

kdheepak (dismissed):

joshka:
Rebased to address lib.rs conflicts, squashed and updated commit message.

kdheepak (approved):

PR #527, opened 2023-09-21
Don't include datasets with no name in legend

alemidev:
Currently, creating a `Dataset` to be used with the `Chart` widget will put an empty name by default. When not overriding it with the `.name()` builder method, it will still appear in the resulting chart legend as an empty space.

In a project of mine I draw more specific reference grids as lines with no names, and the legend field either gets filled with empty spaces or not drawn at all due to its length.

With this change, legend won't be drawn for datasets with no name.

To make the name optional in the struct, I changed the `Cow<str>` into an `Option<Cow<str>>` and excluded datasets with empty names using `.filter_map()` in place of simple `.maps()` while iterating.

To properly calculate legend length, getting `datasets.len()` is no longer sufficient and I iterate them one more time to count which ones have names. This looks cleaner and should have a negligible impact, but could be improved by counting them and finding the longest one in just one pass.

### DIFF
--- a/src/widgets/chart.rs
+++ b/src/widgets/chart.rs
@@ -90,7 +90,7 @@ pub enum GraphType {
 #[derive(Debug, Default, Clone, PartialEq)]
 pub struct Dataset<'a> {
     /// Name of the dataset (used in the legend if shown)
-    name: Cow<'a, str>,
+    name: Option<Cow<'a, str>>,
     /// A reference to the actual data
     data: &'a [(f64, f64)],
     /// Symbol used for each points of this dataset
@@ -106,7 +106,7 @@ impl<'a> Dataset<'a> {
     where
         S: Into<Cow<'a, str>>,
     {
-        self.name = name.into();
+        self.name = Some(name.into());
         self
     }
 
@@ -301,9 +301,15 @@ impl<'a> Chart<'a> {
             }
         }
 
-        if let Some(inner_width) = self.datasets.iter().map(|d| d.name.width() as u16).max() {
+        if let Some(inner_width) = self
+            .datasets
+            .iter()
+            .filter_map(|d| Some(d.name.as_ref()?.width() as u16))
+            .max()
+        {
             let legend_width = inner_width + 2;
-            let legend_height = self.datasets.len() as u16 + 2;
+            let legend_height =
+                self.datasets.iter().filter(|d| d.name.is_some()).count() as u16 + 2;
             let max_legend_width = self
                 .hidden_legend_constraints
                 .0
@@ -545,12 +551,17 @@ impl<'a> Widget for Chart<'a> {
             Block::default()
                 .borders(Borders::ALL)
                 .render(legend_area, buf);
-            for (i, dataset) in self.datasets.iter().enumerate() {
+            for (i, (dataset_name, dataset_style)) in self
+                .datasets
+                .iter()
+                .filter_map(|d| Some((d.name.as_ref()?, d.style)))
+                .enumerate()
+            {
                 buf.set_string(
                     legend_area.x + 1,
                     legend_area.y + 1 + i as u16,
-                    &dataset.name,
-                    dataset.style,
+                    dataset_name,
+                    dataset_style,
                 );
             }
         }


src/widgets/chart.rs
```diff
@@ -545,12 +551,17 @@ impl<'a> Widget for Chart<'a> {
             Block::default()
                 .borders(Borders::ALL)
                 .render(legend_area, buf);
-            for (i, dataset) in self.datasets.iter().enumerate() {
+            for (i, (dataset_name, dataset_style)) in self
+                .datasets
+                .iter()
+                .filter_map(|d| Some((d.name.as_ref()?, d.style)))
```
joshka, line 557: I think this returns Some regardless of whether the item has a name and hence doesn't filter out empty names.

src/widgets/chart.rs
```diff
@@ -301,9 +301,15 @@ impl<'a> Chart<'a> {
...
+        if let Some(inner_width) = self
+            .datasets
+            .iter()
+            .filter_map(|d| Some(d.name.as_ref()?.width() as u16))
+            .max()
+        {
             let legend_width = inner_width + 2;
-            let legend_height = self.datasets.len() as u16 + 2;
+            let legend_height =
+                self.datasets.iter().filter(|d| d.name.is_some()).count() as u16 + 2;
```
joshka, line 312: There's some complexity to this that's probably worth calling to make it easier to understand for a future reader out by adding a comment / intermediate variable with a good name / extracted method with a good name.

joshka (changes requested):
Can you please add docs for the rendering decisions for the legend on the Chart type and on the Dataset name method?

Can you also please add some unit tests for this? Use the approach from `it_does_not_panic_if_title_is_wider_than_buffer()` but use `assert_buffer_eq!` to help highlight failures easily during and after testing.

joshka:
> This looks cleaner and should have a negligible impact, but could be improved by counting them and finding the longest one in just one pass.

No need as the count should be generally small as you noted.

alemidev:
> Can you please add docs for the rendering decisions for the legend on the Chart type and on the Dataset name method?
> Can you also please add some unit tests for this?

Sure thing! Will look into it in the coming days

alemidev:
Sorry that "in the coming days" became "in the next three weeks" but I eventually got to it

I added concise docs on Chart and Dataset::name explaining the behavior. I also updated a little the Chart examples to show potential usage. I tried to be consistent with what I saw around but let me know if you'd rather reword things.

I added a test checking that unnamed charts don't increase legend height.

I also rewrote the part filtering out unnamed datasets by collecting names into a vector, hoping this will make it clearer

src/widgets/chart.rs
```diff
@@ -102,11 +102,14 @@ pub struct Dataset<'a> {
 }
 
 impl<'a> Dataset<'a> {
+    /// Assigns a name to this dataset, making it appear on the [Chart] legend.
+    /// Datasets don't require a name and can be created without specifying one.
+    /// Once assigned, a name can't be removed, only changed.
```
joshka, line 107: Nit: doc comments generally have a header and body like:

```suggestion
    /// Assigns a name to this dataset
    ///
    /// Named datasets will appear in the [Chart] legend. This is an optional property, Datasets
    /// don't require a name and can be created without specifying one.
```

src/widgets/chart.rs
```diff
@@ -154,6 +157,7 @@ struct ChartLayout {
 }
 
 /// A widget to plot one or more dataset in a cartesian coordinate system
+/// It will draw a legend box listing only datasets with a name set
```
joshka, line 160:
```suggestion
/// A widget to plot one or more dataset in a Cartesian coordinate system
///
/// It will draw a legend box listing every datasets that has a name set
```

src/widgets/chart.rs
```diff
@@ -90,7 +90,7 @@ pub enum GraphType {
 #[derive(Debug, Default, Clone, PartialEq)]
 pub struct Dataset<'a> {
     /// Name of the dataset (used in the legend if shown)
-    name: Cow<'a, str>,
+    name: Option<Cow<'a, str>>,
```
joshka, line 93: Observation for future - this could be changed to `Line` to support styling the items.

joshka (changes requested):
Some small doc suggestions to apply, but looks good to me otherwise

joshka:
Hi there. Are you still interested in getting this across the line? If so, it will need to be rebased on the recent changes to the Chart widget. These don't seem to apply cleanly, so you'll need to take some care when doing this to ensure the behavior is maintained.

PR #530, opened 2023-09-22
feat(frame)!: Remove generic Backend parameter

joshka:
The purpose of this change is to simplify UI code that uses the Frame
type. E.g.:

```
fn draw<B: Backend>(frame: &mut Frame<B>) {
    // ...
}
```

Frame was generic over Backend because it stored a reference to the
terminal in the field. Instead it now directly stores the viewport area
and current buffer. These are provided at creation time and are valid
for the duration of the frame.

BREAKING CHANGE: Frame is no longer generic over Backend. Code that
accepted a Frame<Backend> will now need to accept a Frame.

### DIFF
--- a/examples/calendar.rs
+++ b/examples/calendar.rs
@@ -16,7 +16,7 @@ fn main() -> Result<(), Box<dyn Error>> {
     let mut terminal = Terminal::new(backend)?;
 
     loop {
-        let _ = terminal.draw(|f| draw(f));
+        let _ = terminal.draw(draw);
 
         if let Event::Key(key) = event::read()? {
             #[allow(clippy::single_match)]
@@ -35,7 +35,7 @@ fn main() -> Result<(), Box<dyn Error>> {
     Ok(())
 }
 
-fn draw<B: Backend>(f: &mut Frame<B>) {
+fn draw(f: &mut Frame) {
     let app_area = f.size();
 
     let calarea = Rect {
--- a/examples/barchart.rs
+++ b/examples/barchart.rs
@@ -136,7 +136,7 @@ fn run_app<B: Backend>(
     }
 }
 
-fn ui<B: Backend>(f: &mut Frame<B>, app: &App) {
+fn ui(f: &mut Frame, app: &App) {
     let chunks = Layout::default()
         .direction(Direction::Vertical)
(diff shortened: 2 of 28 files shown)


examples/calendar.rs
```diff
@@ -16,7 +16,7 @@ fn main() -> Result<(), Box<dyn Error>> {
     let mut terminal = Terminal::new(backend)?;
 
     loop {
-        let _ = terminal.draw(|f| draw(f));
+        let _ = terminal.draw(draw);
```
joshka, line 19: Clippy dictated this change, which makes sense as the terminal method is concretely parameterized while the app method was generic and couldn't be called directly. Now it can.

kdheepak:
I haven't checked all the tests but what happens if a terminal resize occurs? Does this handle that?

joshka:
> I haven't checked all the tests but what happens if a terminal resize occurs? Does this handle that?

Frames are only created by `terminal.get_frame(&mut self)` which borrows terminal mutably and prevents the resize method from being called. The viewport and buffer size can only be changed by something that can mutate the terminal. The following does not compile:
```rust
let backend = TestBackend::new(10, 10);
let mut terminal = Terminal::new(backend).unwrap();
let frame = terminal.get_frame();
terminal.resize(Rect::new(0, 0, 5, 5)).unwrap();
```
```
 1  error[E0499]: cannot borrow `terminal` as mutable more than once at a time
    --> src/terminal.rs:687:9
     |
 686 |         let frame = terminal.get_frame();
     |                     -------------------- first mutable borrow occurs here
 687 |         terminal.resize(Rect::new(0, 0, 5, 5)).unwrap();
     |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ second mutable borrow occurs here
 688 |         assert_eq!(frame.size(), Rect::new(0, 0, 10, 10));
     |                    ------------ first borrow later used here
 
 For more information about this error, try `rustc --explain E0499`.
```

kdheepak (approved):

orhun (approved):
Breaking but good. We should add "migration" instructions to the final commit message.

joshka:
> Breaking but good. We should add "migration" instructions to the final commit message.

Will do

PR #531, opened 2023-09-23
chore(rustfmt): Make empty items not single line

kdheepak:
I have format on save enabled. During prototyping, if I type the following:

```rust
impl<'a> Widget for Histogram<'a> {
    fn render(mut self, area: Rect, buf: &mut Buffer) {
    }
}
```

and hit save, it gets formatted to this:

```rust
impl<'a> Widget for Histogram<'a> {
    fn render(mut self, area: Rect, buf: &mut Buffer) {}
}
```

It would be nice if we can disable this.

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -3,3 +3,4 @@ group_imports = "StdExternalCrate"
 imports_granularity = "Crate"
 wrap_comments = true
 comment_width = 100
+empty_item_single_line = false


joshka:
I prefer empty functions to be annotated with an empty end of line block like that. It's harmonious with the match empty block syntax:
```rust
match foo {
  bar => {}
  _ => {}
}

For items, it's mostly useful when you have a trait with functions that are overridable by an implementor, or multiple empty functions.

As a workaround, perhaps there is a vim shortcut that automatically positions the cursor in between those characters adds a new line and starts writing the function?

kdheepak:
Yeah, I tried it out locally and it makes a number of changes that we might not prefer. I'll close this PR.

PR #532, opened 2023-09-23
feat(barchart): allow to render charts smaller than 3 lines

### LINKED ISSUE #513: Barchart doesn't render labels in entire width

karthago1:
Add an internal structure `LabelInfo`, which stores the reserved height for the labels (0, 1 or 2) and also the state of the labels, whether they will be shown or not.

Add internal property to the Bar `transformed_value`, which is used only for in the vertical charts. This value is needed in 2 reasons:
1. to draw the vertical bar
2. to draw the bar value or text_value. If the transformed_value is below 8 and the bar_width is 1, then the value will not be printed

transformed_value is line height multiplied by 8.
(8 is the number of element of symbols::bar::Set struct)

fixes #513
[asciicast](https://asciinema.org/a/s69ZQrsjNDO8syNuA5415aVgp)

### DIFF
--- a/src/widgets/barchart.rs
+++ b/src/widgets/barchart.rs
@@ -283,6 +283,12 @@ impl<'a> BarChart<'a> {
     }
 }
 
+struct LabelInfo {
+    group_label_visible: bool,
+    bar_label_visible: bool,
+    height: u16,
+}
+
 impl<'a> BarChart<'a> {
     /// Check the bars, which fits inside the available space and removes
     /// the bars and the groups, which are outside of the available space.
@@ -306,23 +312,43 @@ impl<'a> BarChart<'a> {
         }
     }
 
-    /// Get the number of lines needed for the labels.
+    /// Get label information.
     ///
-    /// The number of lines depends on whether we need to print the bar labels and/or the group
-    /// labels.
-    /// - If there are no labels, return 0.
-    /// - If there are only bar labels, return 1.
-    /// - If there are only group labels, return 1.
-    /// - If there are both bar and group labels, return 2.
-    fn label_height(&self) -> u16 {
-        let has_group_labels = self.data.iter().any(|e| e.label.is_some());
-        let has_data_labels = self
+    /// height is the number of lines, which depends on whether we need to print the bar
+    /// labels and/or the group labels.
+    /// - If there are no labels, height is 0.
+    /// - If there are only bar labels, height is 1.
+    /// - If there are only group labels, height is 1.
+    /// - If there are both bar and group labels, height is 2.
+    fn label_info(&self, available_height: u16) -> LabelInfo {
+        if available_height == 0 {
+            return LabelInfo {
+                group_label_visible: false,
+                bar_label_visible: false,
+                height: 0,
+            };
+        }
+
+        let bar_label_visible = self
             .data
             .iter()
             .any(|e| e.bars.iter().any(|e| e.label.is_some()));
 
-        // convert true to 1 and false to 0 and add the two values
-        u16::from(has_group_labels) + u16::from(has_data_labels)
+        if available_height == 1 && bar_label_visible {
+            return LabelInfo {
+                group_label_visible: false,
+                bar_label_visible: true,
+                height: 1,
+            };
+        }
+
+        let group_label_visible = self.data.iter().any(|e| e.label.is_some());
+        LabelInfo {
+            group_label_visible,
+            bar_label_visible,
+            // convert true to 1 and false to 0 and add the two values
+            height: u16::from(group_label_visible) + u16::from(bar_label_visible),
+        }
     }
 
     /// renders the block if there is one and updates the area to the inner area
@@ -425,9 +451,16 @@ impl<'a> BarChart<'a> {
         }
     }
 
-    fn render_vertical_bars(&self, buf: &mut Buffer, bars_area: Rect, max: u64) {
+    fn render_vertical(self, buf: &mut Buffer, area: Rect, max: u64) {
+        let label_info = self.label_info(area.height - 1);
+
+        let bars_area = Rect {
+            height: area.height - label_info.height,
+            ..area
+        };
+
         // convert the bar values to ratatui::symbols::bar::Set
-        let mut groups: Vec<Vec<u64>> = self
+        let group_ticks: Vec<Vec<u64>> = self
             .data
             .iter()
             .map(|group| {
@@ -439,11 +472,17 @@ impl<'a> BarChart<'a> {
             })
             .collect();
 
+        self.render_vertical_bars(bars_area, buf, &group_ticks);
+        self.render_labels_and_values(area, buf, label_info, &group_ticks);
+    }
+
+    fn render_vertical_bars(&self, area: Rect, buf: &mut Buffer, group_ticks: &[Vec<u64>]) {
         // print all visible bars (without labels and values)
-        for j in (0..bars_area.height).rev() {
-            let mut bar_x = bars_area.left();
-            for (group_data, group) in groups.iter_mut().zip(&self.data) {
-                for (d, bar) in group_data.iter_mut().zip(&group.bars) {
+        let mut bar_x = area.left();
+        for (ticks, group) in group_ticks.iter().zip(&self.data) {
+            for (d, bar) in ticks.iter().zip(&group.bars) {
+                let mut d = *d;
+                for j in (0..area.height).rev() {
                     let symbol = match d {
                         0 => self.bar_set.empty,
                         1 => self.bar_set.one_eighth,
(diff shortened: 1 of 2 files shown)


joshka:
Hey,

I'd really like to see the `BarChart` logic to be simplified, so took a stab at partially refactoring this into the following. It's incomplete and untested (more of an idea than a full implementation). I think this approach it's this leads to the code being a lot more  (subjectively) maintainable and readable as there's less reliance on temporal changes to state (e.g. `remove_invisible_groups_and_bars` and `calculate_transformed_value`). The other core simplification is explicitly splitting the layout and rendering tasks so that each part is much clearer, this avoids spreading the layout between `remove_invisible_groups_and_bars` and the `render_{horizontal,vertical}_bars()` functions.

The other part that I'd suggest is moving away from is testing private methods like `remove_invisible_groups_and_bars` and `label_info`. Instead, it's more important to test the rendering of each scenario than the internals of how we got there. By testing from the outside of the unit, rather than testing the internals, it makes it easier to refactor the internals for maintainability / readability / performance as necessary.

```rust
impl<'a> Widget for BarChart<'a> {
    fn render(mut self, mut area: Rect, buf: &mut Buffer) {
        self.render_block(area, buf);
        let area = self.layout_block(area);
        if area.is_emtpy() || self.groups.is_empty() || self.bar_width == 0 {
            return;
        }
        match self.direction {
            Direction::Vertical => self.render_vertical(area, buf),
            Direction::Horizontal => self.render_horizontal(area, buf),
        }
    }
}

impl<'a> BarChart<'a> {
    fn render_block(&self, area: Rect, buf: &mut Buffer) {
        self.block.as_ref().map(|&block| block.render(area, buf));
    }

    fn layout_block(&self, area: Rect) -> Rect {
        self.block
            .as_ref()
            .map_or_else(|| area, |&block| block.inner(area))
    }

    fn render_vertical(self, area: Rect, buf: &mut Buffer) {
        let max_value = self.max_value();
        let layout = self.layout_vertical(area);
        let (bar_height, label_height, group_label_height) = self.heights(area);
        for (group, group_area) in self.groups.iter().zip(layout) {
            group.render_vertical_bars(
                max_value,
                bar_height,
                self.bar_style,
                self.value_style,
                group_area,
                buf,
            );
            group.render_vertical_labels(label_height, self.label_style, group_area, buf);
            group.render_vertical_group_label(
                group_label_height,
                self.label_style,
                group_area,
                buf,
            );
        }
    }

    /// Returns a Rect for each group that will contain all the bars of this group.
    fn layout_vertical(&self, area: Rect) -> Vec<Rect> {
        self.groups
            .iter()
            .map(|g| g.layout_vertical(area, self.bar_width, self.bar_gap))
            .scan(0, |x, mut group_area| {
                // layout the groups from left to right ensuring that the group width is not
                // greater than the available width. Groups beyond the available width are
                // truncated
                group_area.x += *x;
                group_area.width = group_area.width.min(area.width.saturating_sub(*x));
                *x = x
                    .saturating_add(group_area.width)
                    .saturating_add(self.group_gap);
                Some(group_area)
            })
            .collect()
    }

    fn heights(&self, area: Rect) -> (u16, u16, u16) {
        let has_group_labels = self.groups.iter().any(BarGroup::has_label);
        let has_bar_labels = self.groups.iter().any(BarGroup::has_bar_labels);
        // ensure that the height of the leftover area is at least 1 so that the bars can be
        // rendered, and that labels get precedence over group labels
        let bars_height = area
            .height
            .saturating_sub(has_group_labels as u16 + has_bar_labels as u16)
            .max(1);
        let labels_height = area.height.saturating_sub(bars_height).min(1);
        let group_label_height = area
            .height
            .saturating_sub(bars_height)
            .saturating_sub(labels_height)
            .min(1);
        (bars_height, labels_height, group_label_height)
    }

    fn max_value(&self) -> u64 {
        self.max
            .or_else(|| self.groups.iter().flat_map(BarGroup::max).max())
            .unwrap_or_default()
            .max(1)
    }
```

Does this make more sense?

karthago1:
> I think this approach it's this leads to the code being a lot more (subjectively) maintainable and readable as there's less reliance on temporal changes to state (e.g. remove_invisible_groups_and_bars and calculate_transformed_value).

I agree. I removed `calculate_transformed_value`. But I am still in love with the `remove_invisible_groups_and_bars`. As we agreed once, you will someday eliminate my beloved anti pattern function :cry: .

> The other part that I'd suggest is moving away from is testing private methods like remove_invisible_groups_and_bars and label_info. Instead, it's more important to test the rendering of each scenario than the internals of how we got there. By testing from the outside of the unit, rather than testing the internals, it makes it easier to refactor the internals for maintainability / readability / performance as necessary.

done

> Does this make more sense?

maybe :smile_cat: 
I don't like the suggested `heights` function much. It is always iterating over everything even if it is unnecessary. Readability is maybe depending on the person not better than the `label_info`

Anyway this is a second try, it is for sure not what you expect. But please, if it is far from your expectation or making the code worse, then don't hesitate to close the PR. Of course I will be happy to tweek/change small things, if this PR is more or less still acceptable.

(I appreciate very much your taking the time and maintaining this wonderful library :heart:  )

src/widgets/barchart.rs
```diff
@@ -1238,4 +1108,193 @@ mod tests {
...
+                "          ▂ ▄ ▆ █",
+                "  ▂ ▄ ▆ 4 5 6 7 8",
+                "a b c d e f g h i",
+                "      Group      ",
+            ])
+        );
+    }
+
+    #[test]
+    fn tow_lines_without_bar_labels() {
```
joshka, line 1272:
```suggestion
    fn two_lines_without_bar_labels() {
```

src/widgets/barchart.rs
```diff
@@ -1238,4 +1108,193 @@ mod tests {
...
+            .bars(&[
+                Bar::default().value(0),
+                Bar::default().value(1),
+                Bar::default().value(2),
+                Bar::default().value(3),
+                Bar::default().value(4),
+                Bar::default().value(5),
+                Bar::default().value(6),
+                Bar::default().value(7),
+            ]);
```
joshka, line 1284: nit: let's go to 8 here for consistency? If we want to check how the rounding of values that don't exactly divide into the max, perhaps do so in a separate test method?

joshka (approved):
Thanks - looks good to me.

I'm approving this, (with a couple of quick nits to fix).

--- 

Regarding the "beloved antipattern" ;)

There's a strong rationale behind why I find it easier to read the refactored code. In general, code that is immutable is easier to reason about, as you never have to keep in mind which things have happened earlier to fields. The subjective readability measure comes from not having to read all of the entire code in order to understand what each field will be set to at various points in the creation / rendering of the widget, and not having to look for mutations inside several levels of for loops.

The remove invisible groups and bars approach (which mutates the state) works because the widget is consumed on render. If we change the widget to be borrowed in the future, then the entire logic for the widget needs to rewritten (this is something that we're considering).

joshka:
> I don't like the suggested `heights` function much. It is always iterating over everything even if it is unnecessary. Readability is maybe depending on the person not better than the `label_info`

I don't think I quite understand this comment. The `heights` function has the exact effect and negligible difference in perfo for small N (which is what we're dealing with at the terminal) as the label_info function.

src/widgets/barchart.rs
karthago1: > let's go to 8 here for consistency

done

> If we want to check how the rounding of values that don't exactly divide into the max, perhaps do so in a separate test method?

I am not sure if I understood you well, can you check whether the test function `one_lines_with_more_bars` does what asked?

karthago1:
@joshka I agree with everything you said. Maybe this weekend I will try to create a PR which removes my nice function `remove_invisible_groups_and_bars`. I will try to get your suggestions implemented.

joshka (approved):
LGTM. Thanks for the PR and updates.

joshka:
Merged with a small rewording of the commit message

PR #534, opened 2023-09-24
feat(rect): add is_empty() to simplify some common checks

joshka:
- refactor(rect): extract Rect to layout/rect.rs
- feat(rect): add is_empty() to simplify some common checks

### DIFF
--- a/src/layout.rs
+++ b/src/layout.rs
@@ -1,12 +1,4 @@
-use std::{
-    cell::RefCell,
-    cmp::{max, min},
-    collections::HashMap,
-    fmt,
-    num::NonZeroUsize,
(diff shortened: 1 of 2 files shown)


orhun (approved):
LGTM, we should probably note the module changes in the commit description.

joshka:
> LGTM, we should probably note the module changes in the commit description.

There's no public api change (just a developer happiness change). `rect` is a private module and the `Rect` type is re-exported under the layout module. I'll note this in the commit.

https://github.com/ratatui-org/ratatui/blob/f23677376652b40b55a7c9efde04d46ab20a1f90/src/layout.rs#L28-L29

PR #535, opened 2023-09-25
docs(tabs): add documentation to `Tabs`

Valentin271:
<details><summary>Screenshot</summary>
<p>

image

</p>
</details>

### DIFF
--- a/src/widgets/tabs.rs
+++ b/src/widgets/tabs.rs
@@ -1,3 +1,4 @@
+#![deny(missing_docs)]
 use crate::{
     buffer::Buffer,
     layout::Rect,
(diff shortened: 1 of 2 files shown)


orhun (approved):

src/widgets/tabs.rs
```diff
@@ -9,16 +10,20 @@ use crate::{
 
 /// A widget to display available tabs in a multiple panels context.
 ///
-/// # Examples
+/// Tabs titles can be styled, as well as the currently selected tab ([`Tabs::highlight_style`]).
+/// The divider can be customized with [`Tabs::divider`]. The current tab is selected with the use
+/// of [`Tabs::select`].
```
joshka, line 15: There could be a little ambiguity between a tab (which often refers to the full space that contains other elements) compared to a Tab Bar, which doesn't. What do you think about:

```
A widget that displays a horizontal set of Tabs with a single tab selected."

Each tab title is stored as a `Line` which can be individually styled.
The selected tab is set using [`select()`] and styled using [`highlight_style`].
The divider...
```

joshka (approved):
Happy to take with / without the suggested change - let me know what you think.

joshka (approved):

PR #544, opened 2023-09-28
refactor(barchart): replace remove_invisible_groups_and_bars by group…

karthago1:
It is maybe still not what you expect. but maybe it is a small step. (At least the internal state is not changed)

`remove_invisible_groups_and_bars()` is replaced by `group_ticks()`, which calculates the visible bar length in ticks. (A cell contains 8 ticks).

It is used for 2 purposes:
1. to get the bar length in ticks for rendering
2. since it delivers only the values of the visible bars, If we zip these values with the groups and bars, then we will filter out the invisible groups and bars

Since this patch, doesn't add any new feature, it can be closed, if it doesn't add any value regarding readability.

### DIFF
--- a/src/widgets/barchart.rs
+++ b/src/widgets/barchart.rs
@@ -290,26 +290,43 @@ struct LabelInfo {
 }
 
 impl<'a> BarChart<'a> {
-    /// Check the bars, which fits inside the available space and removes
-    /// the bars and the groups, which are outside of the available space.
-    fn remove_invisible_groups_and_bars(&mut self, mut width: u16) {
-        for group_index in 0..self.data.len() {
-            let n_bars = self.data[group_index].bars.len() as u16;
-            let group_width = n_bars * self.bar_width + n_bars.saturating_sub(1) * self.bar_gap;
-
-            if width > group_width {
-                width = width.saturating_sub(group_width + self.group_gap + self.bar_gap);
-            } else {
-                let max_bars = (width + self.bar_gap) / (self.bar_width + self.bar_gap);
-                if max_bars == 0 {
-                    self.data.truncate(group_index);
-                } else {
-                    self.data[group_index].bars.truncate(max_bars as usize);
-                    self.data.truncate(group_index + 1);
+    /// Returns the visible bars length in ticks. A cell contains 8 ticks.
+    /// `available_space` used to calculate how many bars can fit in the space
+    /// `bar_max_length` is the maximal length a bar can take.
+    fn group_ticks(&self, available_space: u16, bar_max_length: u16) -> Vec<Vec<u64>> {
+        let max: u64 = self.maximum_data_value();
+        self.data
+            .iter()
+            .scan(available_space, |space, group| {
+                if *space == 0 {
+                    return None;
                 }
-                break;
-            }
-        }
+                let n_bars = group.bars.len() as u16;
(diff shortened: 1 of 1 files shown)


karthago1:
also I noticed, that I should replace the removed unit tests in the previous commit. (I almost made some breaking changes)

src/widgets/barchart.rs
```diff
@@ -290,26 +290,43 @@ struct LabelInfo {
...
+                let n_bars = if *space > group_width {
+                    *space = space.saturating_sub(group_width + self.group_gap + self.bar_gap);
+                    Some(n_bars)
+                } else {
+                    let max_bars = (*space + self.bar_gap) / (self.bar_width + self.bar_gap);
+                    if max_bars > 0 {
+                        *space = 0;
+                        Some(max_bars)
+                    } else {
+                        None
```
joshka, line 316: Does this line need a test?

src/widgets/barchart.rs
```diff
@@ -371,35 +388,25 @@ impl<'a> BarChart<'a> {
...
-                    .map(|bar| (bar.value * u64::from(bars_area.width) / max) as u16)
-                    .collect()
-            })
-            .collect();
+        let group_ticks = self.group_ticks(bars_area.height, bars_area.width);
 
         // print all visible bars, label and values
         let mut bar_y = bars_area.top();
-        for (group_data, mut group) in groups.into_iter().zip(self.data) {
+        for (ticks_vec, mut group) in group_ticks.into_iter().zip(self.data) {
```
joshka, line 405: What makes group mutable here?

src/widgets/barchart.rs
```diff
@@ -371,35 +388,25 @@ impl<'a> BarChart<'a> {
...
             let bars = std::mem::take(&mut group.bars);
 
-            for (bar_length, bar) in group_data.into_iter().zip(bars) {
+            for (ticks, bar) in ticks_vec.into_iter().zip(bars) {
+                let bar_length = (ticks >> 3) as u16;
```
joshka, line 409: Prefer this for clarity unless there's a measured performance difference (I'd guess this is probably automatically compiler optimized)
```suggestion
                let bar_length = (ticks /8 ) as u16;
```

joshka (approved):
Generally LGTM - with a couple of quick questions

src/widgets/barchart.rs
karthago1: yes. I don't think that a slightly slower cpu instruction can be measured in this context.
done

src/widgets/barchart.rs
karthago1: yes. added this function `first_bar_of_the_group_is_half_outside_view()`

src/widgets/barchart.rs
karthago1: because of the following reasons:
1. the bars and group needs to be consumed by their render methods
2. bars are owned by the group
3. so if we render bars first, they will be consumed
4. then if we try to render the group, we get this error: `value used here after partial move`

to solve the problem, I need to take the bars out of the group first:
`let bars = std::mem::take(&mut group.bars);`

joshka: got it - thanks.

joshka (approved):

joshka:
Thanks for the PR

PR #545, opened 2023-09-29
docs: Improve clarity in documentation for `Frame`

kdheepak:
The documentation currently says `This allows for widgets to be drawn in any order.` which I believe is incorrect.

I removed that line and tweaked the remaining text for more clarity.

### DIFF
--- a/src/terminal.rs
+++ b/src/terminal.rs
@@ -91,9 +91,11 @@ pub struct TerminalOptions {
 /// terminal libraries: [Crossterm], [Termion] and [Termwiz]. See the [`backend`] module for more
 /// information.
 ///
-/// The terminal has two buffers that are used to draw the application. The first buffer is the
-/// current buffer and the second buffer is the previous buffer. The two buffers are compared at
-/// the end of each draw pass to output only the changes to the terminal.
+/// The `Terminal` struct maintains two buffers: the current and the previous.
+/// When the widgets are drawn, the changes are accumulated in the current buffer.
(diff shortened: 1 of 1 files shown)


joshka:
Is there a way to rephrase this without "changes drawn"? I don't get from that wording that the entire frame should be drawn rather than the full frame needs to be refreshed.

kdheepak:
How about something like this?

```
/// The `Terminal` struct maintains two buffers: the current and the previous.
/// When the widgets are drawn, the changes are accumulated in the current buffer.
/// At the end of the draw pass, the two buffers are compared, and only the changes
/// between these buffers are written to the terminal, avoiding any redundant operations.
/// After flushing these changes, the buffers are swapped to prepare for the next draw cycle.
```

joshka:
> How about something like this?
> 
> ```
> /// The `Terminal` struct maintains two buffers: the current and the previous.
> /// When the widgets are drawn, the changes are accumulated in the current buffer.
> /// At the end of each draw pass, the two buffers are compared, and only the changes
> /// between these buffers are written to the terminal, avoiding any redundant operations.
> /// After flushing these changes, the buffers are swapped to prepare for the next draw cycle.
> ```

LGTM

kdheepak:
Looking at the code again. This change is for the `Frame` struct, but the comment I shared above is better documentation for the `Terminal` struct. 

I'm pushing another change that rephrases the comment for the `Frame` struct and updating the `Terminal` struct's docstring with the text above. Let me know what you think?

joshka (approved):

PR #548, opened 2023-09-29
feat: Add `buffer_mut` method on `Frame`

kdheepak:
`current_buffer_mut` already exists on `Terminal`. It would be nice to expose the same thing to `Frame` as well.

### DIFF
--- a/src/terminal.rs
+++ b/src/terminal.rs
@@ -650,6 +650,11 @@ impl Frame<'_> {
     pub fn set_cursor(&mut self, x: u16, y: u16) {
         self.cursor_position = Some((x, y));
     }
+
+    /// Gets the buffer that this `Frame` draws into as a mutable reference.
+    pub fn buffer_mut(&mut self) -> &mut Buffer {
+        self.buffer
+    }
 }
 
 /// `CompletedFrame` represents the state of the terminal after all changes performed in the last


src/terminal.rs
```diff
@@ -650,6 +650,11 @@ impl Frame<'_> {
     pub fn set_cursor(&mut self, x: u16, y: u16) {
         self.cursor_position = Some((x, y));
     }
+
+    /// Gets the current buffer as a mutable reference.
+    pub fn current_buffer_mut(&mut self) -> &mut Buffer {
```
joshka, line 655: Probably:
```suggestion
    pub fn buffer_mut(&mut self) -> &mut Buffer {
```
As there aren't multiple buffers from the perspective of a Frame - just a single one.

joshka:
I think this is a good idea - as it enables working with the frame in sub functions (e.g. things like components).
Let's wait for a second review by one of the other maintainers on this though.

src/terminal.rs
kdheepak: I did consider this, but opted for `current_buffer_mut` for two reasons:

1. Consistency with the `Terminal` struct
1. It communicates more clearly that is a only currently available buffer, i.e. it is the current buffer that the frame has access to. It also reinforces the idea that ratatui uses a double buffer rendering method.

For example, I've written code accidentally before when I call `draw` twice in the same loop to render different things. 

To me it is slightly more clear that we are getting a temporary buffer here:

```
let terminal = Terminal::new();
terminal.draw(|f| {
  let buf = f.current_buffer_mut();
  // use buf here to render widget 1
}
terminal.draw(|f| {
  let buf = f.current_buffer_mut();
  // use buf here to render widget 2
}
```

than in this code:

```
let terminal = Terminal::new();
terminal.draw(|f| {
  let buf = f.buffer_mut();
  // use buf here to render widget 1
}
terminal.draw(|f| {
  let buf = f.buffer_mut();
  // use buf here to render widget 2
}
```

In both cases above you should never do this because the widget 2 render will overwrite widget 1's render. Which one of these is names for the method is more likely to convey this? If a new user picks up ratatui, vaguely knows that it uses a double buffer techique and has never read the implementation details, seeing `f.current_buffer_mut()` will convey more clearly what is going on.

Thoughts?

joshka: I see the idea and can understand how you get to that point, but I still disagree.

To me, the fact that you only get at the frame is enough to indicate the temporary-ness you're suggesting is added by `current_`

>It also reinforces the idea that ratatui uses a double buffer rendering method.

From the perspective of a single draw call, which is the only place that Frame can exist, there is only a single buffer. And that is the buffer for the frame for the duration of drawing that single frame. No other buffer exists from the perspective of the application. While we know that the terminal is backed by a double buffer, this is something that is an implementation detail rather than something that applications should know about.

Termwiz has a BufferedTerminal implementation which does much the same buffer diffing we do in the Terminal. We might consider disabling the double buffer in the terminal and pushing it down to the backends to take advantage of that rather than performing this twice (or removing the BufferedTerminal). Another future implementation approach could be to switch to a single buffer mode in order to let the application handle its own update timeing somehow. This might be useful for games (not a fully thought out approach though).

TLDR, `current` here is wrong IMO. It implies temporality that only exists at the Terminal level, not the frame level.

kdheepak: Okay, that's fair. I agree that it should be an implementation detail but at the moment I don't think it is. Users have to be aware of the double buffer implementation detail otherwise they might write code like this:

```
let terminal = Terminal::new();
while true {
  terminal.draw(|f| {
    let buf = f.buffer_mut();
    // use buf here to render widget 1
  }
  terminal.draw(|f| {
    let buf = f.buffer_mut();
    // use buf here to render widget 2
  }
}
```

and expect it to work. 

But I'm fine with using `buffer_mut()` to set up future changes in this space.

kdheepak: Resolved in https://github.com/ratatui-org/ratatui/pull/548/commits/1a16fc486a1ced2574e536bf7b49860e54f06ac7.

joshka (approved):

orhun (approved):

PR #550, opened 2023-09-30
feat(canvas): implement half block marker

joshka:
A useful technique for the terminal is to use half blocks to draw a grid
of "pixels" on the screen. Because we can set two colors per cell, and
because terminal cells are about twice as tall as they are wide, we can
draw a grid of half blocks that looks like a grid of square pixels.

This commit adds a new `HalfBlock` marker that can be used in the Canvas
widget and the associated HalfBlockGrid.

  Made with VHS

  Made with VHS

### DIFF
--- a/src/widgets/canvas.rs
+++ b/src/widgets/canvas.rs
@@ -226,7 +226,7 @@ impl Grid for CharGrid {
 
 /// The HalfBlockGrid is a grid made up of cells each containing a half block character.
 ///
-/// In terminals, each character is usually twice as tall as it is wide. Unicode has a ccouple of
+/// In terminals, each character is usually twice as tall as it is wide. Unicode has a couple of
 /// vertical half block characters, the upper half block '▀' and lower half block '▄' which take up
 /// half the height of a normal character but the full width. Together with an empty space ' ' and a
 /// full block '█', we can effectively double the resolution of a single cell. In addition, because
@@ -288,9 +288,9 @@ impl Grid for HalfBlockGrid {
         // not the same as the background reset color (i.e. default background color), we need to
         // swap around the colors for that state (2 reset/color).
         //
-        // If color1 == color2, we could just use the upper half block here but replacing this with
-        // a full block character allows us to write unit tests that treat the cell as a single
-        // character instead of two half block characters.
+        // When the upper and lower colors are the same, we could continue to use an upper half
+        // block, but we choose to use a full block instead. This allows us to write unit tests that
+        // treat the cell as a single character instead of two half block characters.
 
         // first we join each adjacent row together to get an iterator that contains vertical pairs
         // of pixels, with the lower row being the first element in the pair
@@ -344,8 +344,8 @@ impl Grid for HalfBlockGrid {
 
 /// Painter is an abstraction over the [`Context`] that allows to draw shapes on the grid.
 ///
-/// It is used by the Shape trait to draw shapes on the grid. It can be useful to think of this as
-/// similar to the [`Buffer`] struct that is used to draw widgets on the terminal.
+/// It is used by the [`Shape`] trait to draw shapes on the grid. It can be useful to think of this
+/// as similar to the [`Buffer`] struct that is used to draw widgets on the terminal.
 #[derive(Debug)]
 pub struct Painter<'a, 'b> {
     context: &'a mut Context<'b>,
@@ -356,9 +356,10 @@ impl<'a, 'b> Painter<'a, 'b> {
     /// Convert the (x, y) coordinates to location of a point on the grid
     ///
     /// (x, y) coordinates are expressed in the coordinate system of the canvas. The origin is in
-    /// the lower left corner of the canvas. The x and y bounds of the canvas define the specific
-    /// area of some coordinate system that will be drawn on the canvas. The resolution of the grid
-    /// is used to convert the (x, y) coordinates to the location of a point on the grid.
+    /// the lower left corner of the canvas (unlike most other coordinates in Ratatui where the
+    /// origin is the upper left corner). The x and y bounds of the canvas define the specific area
+    /// of some coordinate system that will be drawn on the canvas. The resolution of the grid is
+    /// used to convert the (x, y) coordinates to the location of a point on the grid.
     ///
     /// The grid coordinates are expressed in the coordinate system of the grid. The origin is in
     /// the top left corner of the grid. The x and y bounds of the grid are always [0, width - 1]
--- a/src/widgets/canvas.rs
+++ b/src/widgets/canvas.rs
@@ -5,7 +5,9 @@ mod points;
 mod rectangle;
 mod world;
 
-use std::fmt::Debug;
+use std::{fmt::Debug, iter::zip};
+
+use itertools::Itertools;
 
 pub use self::{
     circle::Circle,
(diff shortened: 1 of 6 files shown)


src/widgets/canvas.rs
```diff
@@ -176,6 +227,106 @@ impl Grid for CharGrid {
...
+        )
+    }
+
+    fn save(&self) -> Layer {
+        // We use the following 4 states to represent the 4 possible combinations of foreground and
+        // background colors for each cell:
+        // - off: ' ' colors: reset/reset
+        // - top: '▀' colors: upper/reset
+        // - bottom: '▄' colors: lower/reset (use lower half block set the fg to the lower color)
+        // - both: '▀' colors upper/lower
```
kdheepak, line 280: It took me a minute to understand this comment. I think it's not going to be super clear what is going on here for someone that doesn't understand this technique. Can we expand this comment further? 

Here's my attempt at expanding it:

```suggestion
        // In terminals, text is usually twice as tall as it is wide,
        // Let's assume that the smallest pixel is called a "Cell".
        // In Unicode, there are a couple of vertical half block characters
        // Namely, the upper half block '▀' and lower half block '▄'. 
        // Together with an empty space ' ' and a full block '█', we can
        // effectively double the resolution of a single Cell.
        // Specifically, for the half block characters, we can use 
        // different foreground and background colors to double the 
        // resolution of a single Cell
        //
        // This technique requires scanning the grid and comparing two
        // elements in the vertical axis to decide what character should be
        // printed in a particular cell.
        //
        // In this scan comparison, there are the 4 states we will encounter:
        // - off: ' ' colors - fg: doesn't matter / bg: reset
        // - top: '▀' colors - fg: upper / bg: reset
        // - bottom: '▀' colors - fg: reset / bg: lower
        // - both: '▀' colors - fg: upper / bg: lower
        //
        // However, because the foreground reset color (i.e. default foreground color) is usually not the same as the background reset color (i.e. default background color), we need to swap around the colors for the `bottom` state. So these are the 4 state we use to represent each cell:
        // - off: ' ' colors - fg: doesn't matter / bg: reset
        // - top: '▀' colors - fg: upper / bg: reset
        // - bottom: '▄' colors: fg: lower / bg: reset (use lower half block instead and set the fg to the lower color)
        // - both: '▀' colors - fg: upper / bg: lower
```

src/widgets/canvas.rs
```diff
@@ -176,6 +227,106 @@ impl Grid for CharGrid {
...
+/// This allows for more flexibility than the BrailleGrid which only supports a single foreground
+/// color for each 2x4 dots cell.
+#[derive(Debug, Default, Clone, Eq, PartialEq, Hash)]
+struct HalfBlockGrid {
+    /// width of the grid in number of terminal columns
+    width: u16,
+    /// height of the grid in number of terminal rows
+    height: u16,
+    /// represents a single color for each cell
+    cells: Vec<Vec<Color>>,
```
kdheepak, line 243: Above in the comment, there's this text:

> /// ... This grid type supports a foreground and background color for each terminal cell.

I think the name of `cells` field is not matching the usage of `cells` in the comment.

`Cell` refers to a single "pixel" or single "character" in the terminal, but this `Vec<Vec<Color>>` is the color of each half block, right? Maybe this field name should be called `blocks`?

joshka: pixels sounds right

src/widgets/canvas.rs
joshka: Sounds good - I'll tighten that up a little and push a change

src/widgets/canvas.rs
kdheepak: Thinking about it some more, the name `half_blocks` would be clearest imo. But pixels sounds good too!

joshka: I find calling it `half block` confusing as the half block character takes up a full cell. It's a weird ambiguous semantic thing where I think pixels makes more sense.

joshka: https://github.com/ratatui-org/ratatui/pull/550/commits/30fbec7f5c91f59e5a0a20d8eccb4a3bae5c916b

src/widgets/canvas.rs
joshka: https://github.com/ratatui-org/ratatui/pull/550/commits/30fbec7f5c91f59e5a0a20d8eccb4a3bae5c916b

src/widgets/canvas.rs
```diff
@@ -176,6 +224,128 @@ impl Grid for CharGrid {
     }
 }
 
+/// The HalfBlockGrid is a grid made up of cells each containing a half block character.
+///
+/// In terminals, each character is usually twice as tall as it is wide. Unicode has a ccouple of
```
kdheepak, line 229:
```suggestion
/// In terminals, each character is usually twice as tall as it is wide. Unicode has a couple of
```

src/widgets/canvas.rs
```diff
@@ -176,6 +224,128 @@ impl Grid for CharGrid {
...
+        // 2. upper: reset, lower: color => '▄' fg: lower color / bg: reset
+        // 3. upper: color, lower: reset => '▀' fg: upper color / bg: reset
+        // 4. upper: color, lower: color => '▀' fg: upper color / bg: lower color
+        //
+        // Note that because the foreground reset color (i.e. default foreground color) is usually
+        // not the same as the background reset color (i.e. default background color), we need to
+        // swap around the colors for that state (2 reset/color).
+        //
+        // If color1 == color2, we could just use the upper half block here but replacing this with
+        // a full block character allows us to write unit tests that treat the cell as a single
```
kdheepak, line 292:
```suggestion
        // If uppercolor == lowercolor, we could use the upper half block with fg color == bg color, but instead we replace this with
        // a full block character as it allows us to write unit tests that treat the cell as a single
```

src/widgets/canvas.rs
```diff
@@ -176,6 +224,128 @@ impl Grid for CharGrid {
...
+    }
+
+    fn paint(&mut self, x: usize, y: usize, color: Color) {
+        self.pixels[y][x] = color;
+    }
+}
+
+/// Painter is an abstraction over the [`Context`] that allows to draw shapes on the grid.
+///
+/// It is used by the Shape trait to draw shapes on the grid. It can be useful to think of this as
```
kdheepak, line 347:
```suggestion
/// It is used by the [`Shape`] trait to draw shapes on the grid. It can be useful to think of this as
```

src/widgets/canvas.rs
```diff
@@ -185,6 +355,16 @@ pub struct Painter<'a, 'b> {
 impl<'a, 'b> Painter<'a, 'b> {
     /// Convert the (x, y) coordinates to location of a point on the grid
     ///
+    /// (x, y) coordinates are expressed in the coordinate system of the canvas. The origin is in
+    /// the lower left corner of the canvas. The x and y bounds of the canvas define the specific
```
kdheepak, line 359:
```suggestion
/// the lower left corner of the canvas (unlike most terminals where the origin is the upper left corner). The x and y bounds of the canvas define the specific
```

src/widgets/canvas.rs
```diff
@@ -146,27 +196,25 @@ impl Grid for CharGrid {
     }
 
     fn resolution(&self) -> (f64, f64) {
-        (f64::from(self.width) - 1.0, f64::from(self.height) - 1.0)
+        (f64::from(self.width), f64::from(self.height))
     }
```
kdheepak, line 200: Does this need a `breaking change` label?

joshka: No.

The grid trait is private, and the grids themselves are also, so nothing of this is actually user facing. 

I just moved the -1 outside the resolution calculation because I have a 3840x2160 4K monitor, not a 3839x2159 maximum pixel index monitor ;) I should probably document it as a note in the commit however.

joshka:
Comments addressed in https://github.com/ratatui-org/ratatui/pull/550/commits/c8314489716e5e8240caf5a4bbdc6882f6ffbd08

src/widgets/canvas.rs
kdheepak: oops, didn't notice it was not `pub fn`!

kdheepak (approved):

kdheepak:
Awesome! Excited to see what people make with this!

PR #557, opened 2023-10-04
feat(layout)!: add parameters to Layout::new()

joshka:
Adds a convenience function to create a layout with a direction and a
list of constraints which are the most common parameters that would be
generally configured using the builder pattern. The constraints can be
passed in as any iterator of constraints.

```rust
let layout = Layout::new(Direction::Horizontal, [
    Constraint::Percentage(50),
    Constraint::Percentage(50),
]);
```

BREAKING CHANGE:
Layout::new() now takes a direction and a list of constraints instead of
no arguments. This is a breaking change because it changes the signature
of the function. Layout::new() is also no longer const because it takes
an iterator of constraints.

Layout::new() was added in a recent release, so the amount of code broken by this is reasonably small. Tagging the owners of affected repos as a headsup:
https://github.com/search?q=ratatui+%22Layout%3A%3Anew%22&type=code
- https://github.com/sxyazi/yazi @sxyazi
- https://github.com/masterned/cygnus @masterned
- https://github.com/tokisakiyuu/pound @tokisakiyuu
- https://github.com/Shadorain/shadobeam @Shadorain
- https://github.com/Fabus1184/ramp @Fabus1184
- https://github.com/TheEmeraldBee/widgetui @TheEmeraldBee

### DIFF
--- a/src/layout.rs
+++ b/src/layout.rs
@@ -196,10 +196,11 @@ pub(crate) enum SegmentSize {
 /// use ratatui::{prelude::*, widgets::*};
 ///
 /// fn render(frame: &mut Frame, area: Rect) {
-///     let layout = Layout::default()
-///         .direction(Direction::Vertical)
-///         .constraints([Constraint::Length(5), Constraint::Min(0)])
-///         .split(Rect::new(0, 0, 10, 10));
+///     let layout = Layout::new(
+///         Direction::Vertical,
+///         [Constraint::Length(5), Constraint::Min(0)],
+///     )
+///     .split(Rect::new(0, 0, 10, 10));
 ///     frame.render_widget(Paragraph::new("foo"), layout[0]);
 ///     frame.render_widget(Paragraph::new("bar"), layout[1]);
 /// }
@@ -223,7 +224,7 @@ pub struct Layout {
 
 impl Default for Layout {
     fn default() -> Layout {
-        Layout::new()
+        Layout::new(Direction::Vertical, [])
     }
 }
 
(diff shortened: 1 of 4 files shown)


src/layout.rs
```diff
@@ -292,11 +288,8 @@ impl Layout {
     ///     Rect::new(0, 8, 10, 2),
     /// ]);
     /// ```
-    pub fn constraints<C>(mut self, constraints: C) -> Layout
-    where
-        C: Into<Vec<Constraint>>,
-    {
-        self.constraints = constraints.into();
+    pub fn constraints<C: AsRef<[Constraint]>>(mut self, constraints: C) -> Layout {
```
kdheepak, line 291: I'm curious. What is the advantage or difference of making this `AsRef`?

sxyazi: Both `[T]` and `&[T]` are accepted

kdheepak: The previous code had `Into<Vec<Constraint>>` and that seemed to accept `&[T]` as well, right? 

I guess my question is with this function being `AsRef<[Constaint]>` and the `Layout::new` being `Into<Vec<Constraint>>`, is there any difference from a user perspective?

joshka: It was a mistake. This should be IntoIterator like the new() parameter

joshka: I am going to go with `AsRef<[Constraint]>` here due to AsRef allowing slices of fixed sized arrays (which Into<Vec<Constraint>> does not. IntoIterator has an issue with another case.

```
   Compiling ratatui v0.23.0 (/Users/joshka/local/ratatui)
error[E0277]: the trait bound `Vec<layout::Constraint>: From<&[layout::Constraint; 2]>` is not satisfied
   --> src/layout.rs:634:43
    |
634 |             Layout::default().constraints(&CONSTRAINTS).constraints,
    |                               ----------- ^^^^^^^^^^^^ the trait `From<&[layout::Constraint; 2]>` is not implemented for `Vec<layout::Constraint>`
    |                               |
    |                               required by a bound introduced by this call
    |
    = note: required for `&[layout::Constraint; 2]` to implement `Into<Vec<layout::Constraint>>`
note: required by a bound in `layout::Layout::constraints`
   --> src/layout.rs:291:27
    |
291 |     pub fn constraints<C: Into<Vec<Constraint>>>(mut self, constraints: C) -> Layout {
    |                           ^^^^^^^^^^^^^^^^^^^^^ required by this bound in `Layout::constraints`
help: consider dereferencing here
    |
634 |             Layout::default().constraints(*&CONSTRAINTS).constraints,
    |                                           +

For more information about this error, try `rustc --explain E0277`.
warning: build failed, waiting for other jobs to finish...
```

mindoodoo (approved):

orhun (approved):

PR #565, opened 2023-10-06
feat(LineGauge): allow LineGauge background styles

nowNick:
## Description

LineGauge can be styled with regard to the line itself. There was no option to style the background behind the line. There was a community issue request to add such a feature.

## How to test

I didn't want to add another example to the project as this is really simple to test. Basically replace `render_app` of `examples/hello_world.rs` with the code below:

```rust
fn render_app(frame: &mut Frame) {
        .constraints([Constraint::Ratio(1, 3), Constraint::Ratio(2, 3)])
        .split(frame.size());

    let line_gauge = LineGauge::default()
        .ratio(50.0 / 100.0)
        .label("Test")
        .style(Style::new().light_blue())
        .gauge_style(Style::new().fg(Color::Red).bg(Color::Blue))
        .gauge_background_style(Style::new().fg(Color::Blue).bg(Color::Red))
        .line_set(symbols::line::THICK);
    frame.render_widget(line_gauge, chunks[1]);
}
```

## Screenshot
image

### Issue link

Implements ratatui-org#424


src/widgets/gauge.rs
```diff
@@ -12,9 +12,9 @@ use crate::{
 ///
 /// A `Gauge` renders a bar filled according to the value given to [`Gauge::percent`] or
 /// [`Gauge::ratio`]. The bar width and height are defined by the [`Rect`] it is
-/// [rendered](Widget::render) in.
```
nowNick, line 15: My editor autoremoved trailing whitespaces. I'm not sure if they are needed 🤔

src/widgets/gauge.rs
```diff
@@ -365,7 +378,7 @@ impl<'a> Widget for LineGauge<'a> {
                 .set_symbol(self.line_set.horizontal)
                 .set_style(Style {
                     fg: self.gauge_style.fg,
-                    bg: None,
+                    bg: self.gauge_background_style.fg,
```
nowNick, line 381: I wanted to keep it backwards compatible so I decided to introduce a new field `gauge_background_style`. However I'll admit I'm a bit uneasy with this line as it might be confusing to set `bg` to `fg`. It's just in the context of `LineGauge` foreground is used to style filled area and background is used to style empty area.

Fabus1184:
Very nice, I too think the name `gauge_background_style` is still confusing because of the word "background" which does not clearly indicate that this refers to the "unfilled" area of the gauge, it would be better to have `LineGauge::filled_style` and `LineGauge::unfilled_style` or some similar term. IMO because of compatibility `LineGauge::gauge_style` can be left and just sets the `Style::fg` field of the `LineGauge::filled_style` and the `Style::bg` field of the `LineGauge::unfilled_style` so it won't alter the behavior.

nowNick:
Oh I like that! We can mark `LineGauge::gauge_style` as deprecated and introduce `LineGauge::filled_style` and `LineGauge::unfilled_style` with proper `fg` and `bg`. Then in future we can remove `LineGauge::gauge_style`

src/widgets/gauge.rs
orhun: They are needed for the newlines, you should probably add them back :D

With trailing whitespaces:

image

Without trailing whitespaces:

image

joshka: As a general rule, I strongly prefer to avoid end of line spaces like this as editors and other tooling often bork on them, and they're invisible to users until the output. Use a blank line instead here and a `<br>` in markdown if you really need a newline rather than a new paragraph.

TieWay59:
`LineGauge::filled_style` is neat. I'm always confused by the `bg/fg` somehow, I favor this wording.

And saw your screenshot above, it's great. Hope the new API design did not slow you down, it is always hard to give backward compatibility.

joshka:
Another +1 to filled / unfilled from me.

The remaining tasks on this are something like:
- [ ] Fix the end of line spacing in doc comments (replace with multiple new lines to break the paragraphs)
- [ ] Update to use filled_style and unfilled_style
- [ ] Mark gauge style as deprecated with a message to use filled and unfilled
- [ ] Write / update tests
- [ ] Create a small demo and VHS tape file

examples/line_gauge.rs
```diff
@@ -0,0 +1,218 @@
+//! # [Ratatui] Line Gauge example
```
nowNick, line 1: This is practically copied & adjusted from `examples/gauge.rs`

src/widgets/gauge.rs
nowNick: I've added additional line and I think it fixed the issue. Now it looks like this:
image

nowNick:
Hey all! :)

It took me some time but finally I got around to wrapping everything up. I tried to check all of the boxes on the todo list you provided @joshka 

> [x] Fix the end of line spacing in doc comments (replace with multiple new lines to break the paragraphs)
> [x] Update to use filled_style and unfilled_style
> [x] Mark gauge style as deprecated with a message to use filled and unfilled
> [x] Write / update tests
> [x] Create a small demo and VHS tape file

---

I'm not 100% if vhs demo will be recorded on the main branch in the repository itself or should I do it from my computer and somehow upload it? 
Also I couldn't quite run the vhs tape file according to the instructions in the comments.
They all say something like:

```
# To run this script, install vhs and run `vhs ./examples/paragraph.tape`
```

and I don't think that the path is quite right. Shouldn't it be `vhs ./examples/vhs/paragraph.tape` ?

src/widgets/gauge.rs
```diff
@@ -341,9 +349,36 @@ impl<'a> LineGauge<'a> {
     ///
     /// `style` accepts any type that is convertible to [`Style`] (e.g. [`Style`], [`Color`], or
     /// your own type that implements [`Into<Style>`]).
+    #[deprecated(
+        since = "0.27.0",
+        note = "You should use `LineGauge::filled_style and LineGauge::unfilled_style` instead."
```
EdJoPaTo, line 354: Backticks are missing:
```suggestion
        note = "You should use `LineGauge::filled_style` and `LineGauge::unfilled_style` instead."
```

Not sure about the phrasing as `filled_style` is the direct equivalent and unfilled is new? Maybe tell the user to replace with filled and consider the new unfilled?

examples/README.md
```diff
@@ -164,6 +164,17 @@ cargo run --example=gauge --features=crossterm
...
+## Line Gauge
+
+Demonstrates the [`Line Gauge`](https://docs.rs/ratatui/latest/ratatui/widgets/struct.LineGauge.html) widget.
+Source: [line_gauge.rs](./line_gauge.rs).
+
+```shell
+cargo run --example=line_gauge --features=crossterm
+```
+
+![LineGauge][line_gauge.gif]
```
EdJoPaTo, line 176: Also needs to link the exact file in the bottom of the file.

(I don't like this Markdown feature as it makes it easy to forget… I would prefer `LineGauge` and be done with it.)

nowNick:
So happy it got merged! 🎉 
Thanks for everybody's help! Especially @joshka for pushing it over the last hump 🙇

PR #566, opened 2023-10-08
docs: Replace colons with dashes in README.md for consistency

tz629:
Replaced 4 bullet points with colons under the "third-party libraries" header in README.md with dashes. This is because most other bullet points had a link to the library and then a dash, not a colon.


joshka (approved):

joshka:
Thanks for the PR :)

Issue #555, opened 2023-10-03
Windows 7, crossterm, SetUnderlineColor error


rigelminer:
## Description
When using ratatui with crossterm backend on Windows 7, the application fails to render with an error `Custom { kind: Other, error: "SetUnderlineColor not supported by winapi." }`

## To Reproduce
Render UI with ratatui 0.23.0 on Windows 7

## Expected behavior
The UI renders successfully

## Environment
- OS: Windows 7
- Terminal Emulator: cmd.exe
- Crate version: 0.23.0
- Backend: crossterm

## Additional context
The error originates here https://github.com/crossterm-rs/crossterm/blob/master/src/style.rs#L251. `SetUnderlineColor` command is called in ratatui here: https://github.com/ratatui-org/ratatui/blob/main/src/backend/crossterm.rs#L167 <-- commenting out this line fixes the issue.
I'm not experiencing this problem with the original tui-rs as it doesn't call `SetUnderlineColor`.

Is there a way to make the code call `SetUnderlineColor` only when the underline color functionality is used? I promise not to use on Win 7 :smile:

kdheepak:
Thanks for reporting! I'm using Ratatui on Windows and I'm surprised I didn't run into this yet. Does this happen for you when you style a cell a particular way or does it happen even if you apply no styling?

rigelminer:
No worries!
I believe this is specific to Windows 7. Crossterm is probably using ansi for setting colours on newer versions of Windows.

> Does this happen for you when you style a cell a particular way or does it happen even if you apply no styling?

Looks like it fails even with applying Style::default()

joshka:
How is Windows 7 still a thing? LOL

The colored underline behavior was added recently with a feature flag behind crossterm. We might have to replace that with a "colored-underline" flag or something like that instead.

rigelminer:
> How is Windows 7 still a thing? LOL

Haha, I know right? Some of our users are running Windows 7. I had to install it myself to troubleshoot.

> We might have to replace that with a "colored-underline" flag or something like that instead.

That would be great, thanks! The less intrusive the fix the better - otherwise it's not worth the effort, considering the issue is only present on Win 7. It's that unconditional `SetUnderlineColor(CColor::Reset)` call that's causing the problem.

kdheepak:
You'll also run into the issue if you manually set underline color, right? For example this line:

[...]

PR #570, opened 2023-10-18
fix(backend): add feature flag for underline-color

joshka:
Windows 7 doesn't support the underline color attribute, so we need to
make it optional. This commit adds a feature flag for the underline
color attribute - it is enabled by default, but can be disabled by
passing `--no-default-features` to cargo.

We could specically check for Windows 7 and disable the feature flag
automatically, but I think it's better for this check to be done by the
crossterm crate, since it's the one that actually knows about the
underlying terminal.

To disable the feature flag in an application that supports Windows 7,
add the following to your Cargo.toml:

```toml
ratatui = { version = "0.24.0", default-features = false, features = ["crossterm"] }
```

Fixes https://github.com/ratatui-org/ratatui/issues/555

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,14 +23,8 @@ rust-version = "1.67.0"
 [badges]
 
 [dependencies]
-#! The crate provides a set of optional features that can be enabled in your `cargo.toml` file.
-#!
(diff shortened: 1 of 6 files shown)


joshka:
Ping @rigelminer

kdheepak (approved):

Issue #501, opened 2023-09-14
Gauge: renders percentage inverted since ratatui 0.22.0


LeoniePhiline:
## Description

In ratatui 0.21.0, a full gauge is rendered correctly:
![ratatui-0 21 0](https://github.com/ratatui-org/ratatui/assets/22329650/1f7cffc1-7682-4058-9d0a-13e438ac1016)

However, in ratatui 0.22.0 and 0.23.0, a full gauge is rendered with inverted fg/bg at the percentage label part:
![ratatui-0 22 0](https://github.com/ratatui-org/ratatui/assets/22329650/99258c74-935b-4524-ac28-5d1b8f39fb60)

[Diff v0.21.0...v0.22.0](https://github.com/ratatui-org/ratatui/compare/v0.21.0...v0.22.0#diff-f154dfc8f60440dd00333e94b2b925d6701dac0c874ae7cad6be32440ba5ab4e)

## To Reproduce

Render a gauge (with crossterm) with ratatui 0.21.0, then with 0.22.0+.

Example:

```rust
let gauge = Gauge::default()
    .gauge_style(
        Style::default()
            .fg(Color::LightGreen)
            .add_modifier(Modifier::BOLD)
    )
    .use_unicode(true)
    .ratio(100.0);
```

## Expected behavior

"100%" Percentage label being rendered with gauge color as background.

## Screenshots

See above.

## Environment

- OS: OpenSUSE Tumbleweed
- Terminal Emulator: Konsole
- Font: Hasklug Nerd Font Mono
- Crate version: 0.21, 0.22 and 0.33
- Backend: crossterm 0.26.1 / 0.27.0

## Additional context

*None*

joshka:
I recall writing that code to fix some rendering issues with VHS a while back. I'll fix it back soon.

PR #572, opened 2023-10-18
fix(gauge): fix gauge widget colors

joshka:
The background colors of the gauge had a workaround for the issue we had
with VHS / TTYD rendering the background color of the gauge. This
workaround is no longer necessary in the updated versions of VHS / TTYD.

Fixes https://github.com/ratatui-org/ratatui/issues/501

### DIFF
--- a/examples/gauge.rs
+++ b/examples/gauge.rs
@@ -1,17 +1,32 @@
 use std::{
-    error::Error,
-    io,
-    time::{Duration, Instant},
+    io::{self, stdout, Stdout},
+    rc::Rc,
+    time::Duration,
 };
 
+use anyhow::Result;
 use crossterm::{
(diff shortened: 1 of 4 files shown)


joshka:
Ping @LeoniePhiline

Can you please test this in your application using a patch override in your cargo.toml:
```toml
[patch.crates-io]
ratatui = { git = "https://github.com/ratatui-org/ratatui.git", branch = "fix-gauge" }
```

Note that if you're using the non-alpha version of ratatui there may be a few other unreleased breaking changes. See https://github.com/ratatui-org/ratatui/blob/main/BREAKING-CHANGES.md for more details.

LeoniePhiline:
Thanks for pinging me! Unfortunately I'm currently ill and cannot perform any tests.

kdheepak (approved):

PR #583, opened 2023-10-22
docs(examples): add animation and FPS counter to colors_rgb

joshka:
This doesn't render nicely in VHS yet unfortunately (there's some odd screen tearing that happens)

https://github.com/ratatui-org/ratatui/assets/381361/5cd99151-0511-4501-a4b6-3704b82317dd

### DIFF
--- a/examples/colors_rgb.rs
+++ b/examples/colors_rgb.rs
@@ -19,74 +19,61 @@ fn main() -> color_eyre::Result<()> {
     App::run()
 }
 
+#[derive(Debug, Default)]
 struct App {
     should_quit: bool,
     colors: Vec<Vec<Color>>,
+    fps: Fps,
     frame_count: usize,
-    last_fps_frame_count: usize,
-    last_fps_instant: Instant,
-    fps: f64,
+}
+
+#[derive(Debug)]
+struct Fps {
+    frame_count: usize,
+    last_instant: Instant,
+    fps: Option<f32>,
+}
+
+struct AppWidget<'a> {
+    title: Paragraph<'a>,
+    fps_widget: FpsWidget<'a>,
+    rgb_colors_widget: RgbColorsWidget<'a>,
+}
+
+struct FpsWidget<'a> {
+    fps: &'a Fps,
+}
+
+struct RgbColorsWidget<'a> {
+    /// The colors to render - should be double the height of the area
+    colors: &'a Vec<Vec<Color>>,
+    /// the number of elapsed frames that have passed - used to animate the colors
+    frame_count: usize,
 }
 
 impl App {
-    fn new() -> Self {
-        Self {
-            should_quit: false,
-            colors: vec![],
-            frame_count: 0,
-            last_fps_frame_count: 0,
-            last_fps_instant: Instant::now(),
-            fps: 0.0,
-        }
-    }
-
     pub fn run() -> color_eyre::Result<()> {
         install_panic_hook()?;
 
         let mut terminal = init_terminal()?;
-        let mut app = Self::new();
+        let mut app = Self::default();
 
         let size = terminal.size()?;
         app.setup_colors(size.width, size.height * 2);
         while !app.should_quit {
+            app.tick();
             terminal.draw(|frame| {
-                app.render(frame);
+                frame.render_widget(AppWidget::new(&app), frame.size());
             })?;
             app.handle_events()?;
-            app.frame_count += 1;
         }
         restore_terminal()?;
         Ok(())
     }
 
-    fn render(&mut self, frame: &mut Frame) {
-        let layout = Layout::default()
-            .direction(Direction::Vertical)
-            .constraints([Constraint::Length(1), Constraint::Min(0)])
-            .split(frame.size());
-        self.calculate_fps();
-        let title = format!(
-            "colors_rgb example. Press q to quit. {fps:.2} fps",
-            fps = self.fps
-        );
-        let title = Paragraph::new(title).alignment(Alignment::Center);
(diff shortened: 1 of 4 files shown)


examples/colors_rgb.rs
```diff
@@ -1,72 +1,130 @@
...
-};
+use palette::{convert::FromColorUnclamped, Okhsv, Srgb};
 use ratatui::{prelude::*, widgets::*};
 
-type Result<T> = std::result::Result<T, Box<dyn Error>>;
-
 fn main() -> Result<()> {
-    install_panic_hook();
-    App::new()?.run()
+    // stdout runs ~ 2x faster than stderr on a 2023 Macbook Pro M2 Max (~22-26fps vs ~11-12fps)
```
Valentin271, line 20: Oh that's interesting, `stdout` is also faster on my machine 
- from ~1.60 to ~2 fps in fullscreen
- from ~15 to ~90 fps in 80x24!

If this is consistent, we might want to keep that in mind. We currently use stderr in the template. And we should mention that in the [FAQ](https://ratatui.rs/faq/#should-i-use-stdout-or-stderr)

joshka: Yep - definitely agree. I've been slowly changing things to stdout in various example code etc.

joshka: https://github.com/ratatui-org/website/issues/274

examples/colors_rgb.rs
```diff
@@ -1,72 +1,130 @@
...
+    colors: Vec<Vec<Color>>,
+    frame_count: usize,
+    last_fps_frame_count: usize,
+    last_fps_instant: Instant,
+    fps: f64,
 }
 
 impl App {
-    pub fn new() -> Result<Self> {
+    fn new() -> Result<Self> {
```
Valentin271, line 34: `new` doesn't need to return a result

examples/colors_rgb.rs
```diff
@@ -75,49 +133,22 @@ impl Drop for App {
...
-    fn render(self, area: Rect, buf: &mut Buffer) {
-        let layout = Self::layout(area);
-        Self::render_title(layout[0], buf);
-        Self::render_colors(layout[1], buf);
-    }
+struct RgbColors<'a> {
+    // The colors to render - should be double the height of the area
+    colors: &'a Vec<Vec<Color>>,
+    // the number of elapsed frames that have passed - used to animate the colors
+    frame_count: usize,
```
Valentin271, line 140: nit: these should probably be doc comment

examples/colors_rgb.rs
```diff
@@ -1,72 +1,130 @@
...
             }
         }
         Ok(())
     }
+
+    fn setup_colors(&mut self, width: u16, height: u16) {
+        self.colors.clear();
+        for y in 0..height {
+            let mut row = Vec::new();
+            for x in 0..width {
```
Valentin271, line 115: I couldn't find a real fix for the crash we talked about earlier, however a workaround is to add 1 to the width. This provides to kind of security gap.

```suggestion
            for x in 0..width+1 {
```

Weirdly enough, the crash only occurs while extending the width, not the height.

I think it has to do with the window being resized between colors calculations and rendering.
If that were the case, moving `handle_event` call inside `draw` or reducing the input latency should have fixed it, but it did not so ...

Valentin271 (changes requested):

examples/colors_rgb.rs
orhun: I will test the crash again with this change before merge.

joshka: Changed the for loops to 0..=width and 0..=height

joshka:
https://github.com/ratatui-org/ratatui/assets/381361/26014c24-ebb0-48ab-9355-ed881df54e8d

joshka:
https://github.com/ratatui-org/ratatui/assets/381361/485e775a-e0b5-4133-899b-1e8aeb56e774

joshka:
Ok - I think this is in the right place now.

Valentin271:
Did you tested this? I still got it to panic quite easily.

I'm afraid this is related to fps and computer performance. If you didn't got it to panic that's probably the case

Valentin271:
I really think this is because the frame changes size between the resize event and the next draw. So I included the `setup_colors` call in the draw loop where the `frame.size()` doc says "This is guaranteed not to change when rendering". I got it to work with that, even got a little bit more fps on some occasions.

Could you try these changes? [fix-panic-colors_rgb.patch](https://github.com/ratatui-org/ratatui/files/13695122/fix-panic-colors_rgb.txt) (had to rename it to .txt to upload on github). If you include these, feel free to rearrange as you see fit.
Might also require a comment somewhere to explain all this.

joshka:
The problem is cause by Terminal::draw internally calling autoresize before creating the frame (this is intentiona). If the resize event occurred twice since the previous draw, then the second event may not have been seen by the app (but the size is changed regardless). So the app really can't rely on the size from the resize event to be the size to use for sizing the colors cache. It has to instead use the size from the Frame.

Valentin271:
Yep that's what I did in the patch. Ultimately this is going to be a problem in every app that want to use the resize event then ...

joshka:
> Yep that's what I did in the patch. Ultimately this is going to be a problem in every app that want to use the resize event then ...

yep

Valentin271 (approved):
Looks good, happy to merge that

joshka:
> Yep that's what I did in the patch. Ultimately this is going to be a problem in every app that want to use the resize event then ...

https://github.com/ratatui-org/ratatui/pull/697

PR #597, opened 2023-10-26
refactor(widgets): remove unnecessary dynamic dispatch and heap allocation

rhysd:
This PR refactors implementation of `Paragraph::render`. It allocates `LineComposer` trait object on heap memory and calls its methods via dynamic dispatch. However it can be easily changed to static dispatch with allocating the object on stack.

This PR does the change as refactoring. No rendering logic has changed.

### DIFF
--- a/src/widgets/paragraph.rs
+++ b/src/widgets/paragraph.rs
@@ -226,49 +226,50 @@ impl<'a> Widget for Paragraph<'a> {
             return;
         }
 
-        let style = self.style;
         let styled = self.text.lines.iter().map(|line| {
-            (
-                line.spans
-                    .iter()
-                    .flat_map(|span| span.styled_graphemes(style)),
-                line.alignment.unwrap_or(self.alignment),
-            )
+            let graphemes = line
+                .spans
+                .iter()
+                .flat_map(|span| span.styled_graphemes(self.style));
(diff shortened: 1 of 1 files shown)


src/widgets/paragraph.rs
```diff
@@ -208,6 +208,34 @@ impl<'a> Paragraph<'a> {
         self.alignment = alignment;
         self
     }
+
+    fn render_text<C: LineComposer<'a>>(&self, mut composer: C, area: Rect, buf: &mut Buffer) {
```
joshka, line 212: nit: I'm a big fan of top to bottom reading order to avoid having to scroll up to read an implementation that's used in code below. Can you please move this to a new `impl Paragraph` section below the `impl Widget for Paragraph`?

joshka (approved):
I'm a big fan of anything that simplifies complexity like this.
I can merge as-is and apply my nit after if you don't have time to make the suggested change.

src/widgets/paragraph.rs
rhysd: Thank you for the review. I applied the change and rebased this branch onto the latest main branch at 3f76164.

joshka (approved):
Thanks for the PR and the requested fixes :)

Issue #257, opened 2023-06-12
Allow alignment of table columns or cells


utopiabound:
## Problem
Right aligning data in Table Cells

## Solution(s)
* Offer `fn align()` on `Text` (possibly moving aligment logic from `Paragraph` to `Text`)
* Add alignment logic to `Cell`.

## Alternatives
Adding alignment data to `Line`s in a Cell does nothing.

## Additional context
This is a "port" of the Feature Request from https://github.com/fdehau/tui-rs/issues/295

PR #599, opened 2023-10-27
feat(list): support line alignment

orhun:
related: #257

### DIFF
--- a/src/widgets/list.rs
+++ b/src/widgets/list.rs
@@ -1180,8 +1180,11 @@ mod tests {
             .map(ListItem::new)
             .collect::<Vec<ListItem>>();
         let list = List::new(items);
-        let buffer = render_widget(list, 4, 0);
-        let expected = Buffer::with_lines::<&str>(vec![]);
+        // assert_buffer_eq! doesn't handle zero height buffers so we call this test manually
+        // rather than using render_widget
+        let mut buffer = Buffer::empty(Rect::new(0, 0, 4, 1));
+        Widget::render(list, Rect::new(0, 0, 4, 0), &mut buffer);
+        let expected = Buffer::with_lines(vec!["    "]);
         assert_buffer_eq!(buffer, expected);
     }
 
--- a/src/widgets/list.rs
+++ b/src/widgets/list.rs
@@ -1095,4 +1095,135 @@ mod tests {
         ]);
         assert_buffer_eq!(buffer, expected);
     }
+
+    #[test]
+    fn test_render_list_alignment_odd_line_odd_area() {
+        let items = [
+            Line::from("Odd").alignment(Alignment::Left),
+            Line::from("Even").alignment(Alignment::Center),
+            Line::from("Width").alignment(Alignment::Right),
+        ]
+        .into_iter()
+        .map(ListItem::new)
+        .collect::<Vec<ListItem>>();
+        let list = List::new(items);
+        let buffer = render_widget(list, 7, 5);
+        let expected =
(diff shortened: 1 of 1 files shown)


src/widgets/list.rs
```diff
@@ -286,7 +286,14 @@ impl<'a> StatefulWidget for List<'a> {
...
                 };
-                buf.set_line(elem_x, y + j as u16, line, max_element_width);
+                let x_offset = match line.alignment {
+                    Some(Alignment::Center) => {
+                        (area.width / 2).saturating_sub(line.width() as u16 / 2)
+                    }
+                    Some(Alignment::Right) => area.width.saturating_sub(line.width() as u16),
+                    _ => 0,
+                };
+                buf.set_line(elem_x + x_offset, y + j as u16, line, max_element_width);
```
joshka, line 296: Do you think this could be handled by buffer::set_line so that we don't have to write this code everywhere a line is written? Not blocking, but worth checking for a future refactor.

joshka, line 296: I think this may overwrite the bounds as the width isn't changed while the line start position is.
We probably need a few more combinations of tests that cover the edge cases:
- Odd/even for line length and area (to cover the 2 places that do integer division) (4 tests)
- Zero width for both line and area (2 tests)
- Line <, =, > width (3 tests)
- ?

joshka (changes requested):
I think this probably is missing some unit tests, docs, and breaking changes notes

src/widgets/list.rs
orhun: Yes, we can probably add an alignment parameter to the `set_line` or create `set_line_with_alignment` for refactoring. What do you think? This should be the scope of another PR though.

orhun: I added some tests about this but `test_render_list_alignment_zero_area_width` is panicking for some reason. Can you check?

orhun:
Not sure if this is a breaking change but will update the docs after I figure out the test failure.

src/widgets/list.rs
```diff
@@ -1065,4 +1073,157 @@ mod tests {
...
+
+    #[test]
+    fn test_render_list_alignment_zero_area_width() {
+        let items = [Line::from("Text").alignment(Alignment::Left)]
+            .into_iter()
+            .map(ListItem::new)
+            .collect::<Vec<ListItem>>();
+        let list = List::new(items);
+        let buffer = render_widget(list, 4, 0);
+        let expected = Buffer::with_lines::<&str>(vec![]);
```
joshka, line 1184: Yeah, this is a weird one. assert_buffer_eq! has problems with buffers with zero height. It's probably best to just handle it by avoiding the issue:

```suggestion
        // assert_buffer_eq! doesn't handle zero height buffers so we call this test manually
        // rather than using render_widget
        let mut buffer = Buffer::empty(Rect::new(0, 0, 4, 1));
        Widget::render(list, Rect::new(0, 0, 4, 0), &mut buffer);
        let expected = Buffer::with_lines(vec!["    "]);
```

src/widgets/list.rs
joshka: Yep - definitely worth doing in another PR, but I suspect it should be the default behavior of set_line (if a line has an alignment, and we try to render the line, then it should be rendered aligned.

src/widgets/list.rs
orhun: f247e48ed6bd06fc944514344652dffe1e679903

src/widgets/list.rs
orhun: Makes sense, let's create an issue for refactoring it. (probably I can work on it)

joshka (approved):

Issue #293, opened 2023-07-02
Conslidated wrapping design


joshka:
# Request for comment

This is stub of a design to be completed so that the implementation actually builds what we need for wrapping.

## Problem

The requirements of wrapping / truncation / boundaries / trimming and how they should interact with text, lines, and widgets are unclear. This Issue aims to address that.

We have a reflow module that we've bolted on some features, and are trying to add / refactor away from. That's turning out to be difficult because we don't have a clear destination in mind.

## Requirements

1. Should be minimally breaking existing code. Code that uses the current methods should receive a deprecated warning for some amount of versions (at least 1 - and probably more) make it easy for developers to upgrade their code.
2. We should support setting lines to wrap
3. We should support setting Text to wrap
4. We should support setting widgets that contain text / lines to wrap
5. We should be able to configure the boundary of where to wrap (e.g. Character / Word)
6. We should be able to just wrap visible lines (e.g. scrolled paragraphs should be reasonably fast)
7. We should be able to trim leading space from lines that are wrapped
8. We should be able to indent lines that are wrapped by a certain amount
9. The complexity of the implementation should approach the complexity of the algorithm (right now the code to algorithm size ratio is about 10:1)
10. Scrolling should be possible without having to wrap every line of a Text

## Questions

1. Search existing issues / PRs for requirements / concrete needs from users.

## Solution

1. Each item that supports wrapping (e.g. Line, Text, Paragraph, ListItem, Cell, etc.) will have a wrap field that takes a Wrap struct with configuration. (this is the current state)
2. The wrap struct will be defined in the text module
3. Existing methods that take the current wrap will transition to taking an Into<text::wrap>
4. Wrapping shall occur in the render call and be the combination of the line / text / widget (in precedence order).
5. 

Please continue adding to this.

## Links
- https://github.com/tui-rs-revival/ratatui/issues/242
- https://github.com/tui-rs-revival/ratatui/issues/201
- https://github.com/tui-rs-revival/ratatui/issues/284
- https://github.com/tui-rs-revival/ratatui/pull/259
- https://github.com/tui-rs-revival/ratatui/pull/290
- https://github.com/tui-rs-revival/ratatui/pull/33
-

joshka:
Pinging @lthoerner @TimerErTim @topaxi

topaxi:
[...]

PR #607, opened 2023-11-02
feat: make widgets::reflow public

progval:
This makes it easier for users to define their own `Paragraph` with custom rendering logic.

### DIFF
--- a/src/widgets.rs
+++ b/src/widgets.rs
@@ -30,7 +30,7 @@ mod clear;
 mod gauge;
 mod list;
 mod paragraph;
-mod reflow;
+pub mod reflow;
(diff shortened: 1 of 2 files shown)


joshka:
Hi there - this PR has some unintentional baggage that you were probably not aware of. The reflow module is complex and we'd like to deprecate it (and reimplement something that is easier to maintain) instead of making it part of the Ratatui public API. See https://github.com/ratatui-org/ratatui/issues/293 for a very long list of things related to this.

I'd suggest copying the code rather than making it public.

progval:
> I'd suggest copying the code rather than making it public.

ack, I just did that

Issue #617, opened 2023-11-07
include `fn centered_rect` from `examples/popup.rs` in API


wookietreiber:
**Note:** This is more or less a copy of my issue on the original `tui-rs` project: https://github.com/fdehau/tui-rs/issues/498.

## Problem

I find the `fn centered_rect` in `examples/popup.rs` to be extremely useful, so much, that I would like it to be part of the API. I'd probably copy this function to every TUI project :sweat_smile: 

## Solution

- add the function to the public API, possibly on `Rect` itself, as it always takes a `Rect` as an argument
- keep the `popup.rs` example and use the function from the public API instead of defining it there again

## Alternatives

- keep it as is
- have a `ratatui-util` crate to contain functions like these

## Additional context

Whether to include this in the public API depends on your inclination towards having these kinds of utility functions in the API. I have heard different opinions on this and I'd respect not adding this. IMHO if enough members of the community use this function anyway, I'd include it in the API, as it would be a good way to do cross-project code deduplication and to have a reference to the usecase in the API documentation.

joshka:
@kdheepak wrote in #618

---
Thanks for the PR! I also tend to have this in almost every application, and I believe this is a good candidate for inclusion. 

There is already a `inner` method:

https://github.com/ratatui-org/ratatui/blob/59b9c32fbc2bc6725bdec42e63216024fab71493/src/layout/rect.rs#L93

that people can use like this:

```rust
Rect::new(1, 2, 3, 4).inner(&Margin::new(1, 2)),
```

I'm not sure if this can be done in a non-breaking manner, but it would have been nice to use the same function for different purposes.

e.g.


```rust
Rect::new(1, 1, 10, 10).inner(Margin::px(1, 2))
Rect::new(1, 1, 10, 10).inner(Margin::percent(10, 20))
```

joshka:
>I'm not sure if this can be done in a non-breaking manner, but it would have been nice to use the same function for different purposes.

Regarding non-breaking often replacing a parameter that takes `T` (e.g. Margin) with one that takes `Into<U>` (e.g.  `Into<InnerRectConstraint>`) is a pretty easy way to maintain mostly backwards compatible things.

> There is already a inner method:

I definitely agree with @kdheepak's implication here - centered seems too specific / narrow. For context the definition in #617 is:

```rust
pub fn centered_rect(self, percent_x: u16, percent_y: u16) -> Rect
```

[...]

PR #618, opened 2023-11-07
feat: adds centered to public API

wookietreiber:
I'm leaving this in draft for now and won't be adding tests or fixing coverage just yet. Wanted to communicate my idea for #617. If you want this to happen, let me know, then I'll add tests for coverage, etc. 

I think the function name should be changed to something that reads better. Haven't come up with any good ideas though :thinking: 

---

fixes #617

### DIFF
--- a/examples/popup.rs
+++ b/examples/popup.rs
@@ -86,29 +86,8 @@ fn ui(f: &mut Frame, app: &App) {
 
     if app.show_popup {
         let block = Block::default().title("Popup").borders(Borders::ALL);
-        let area = centered_rect(60, 20, size);
+        let area = size.centered_rect(60, 20);
         f.render_widget(Clear, area); //this clears out the background
         f.render_widget(block, area);
     }
-}
-
-/// helper function to create a centered rect using up certain percentage of the available rect `r`
-fn centered_rect(percent_x: u16, percent_y: u16, r: Rect) -> Rect {
-    let popup_layout = Layout::default()
-        .direction(Direction::Vertical)
-        .constraints([
-            Constraint::Percentage((100 - percent_y) / 2),
-            Constraint::Percentage(percent_y),
-            Constraint::Percentage((100 - percent_y) / 2),
-        ])
-        .split(r);
-
-    Layout::default()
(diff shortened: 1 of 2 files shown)


kdheepak:
Thanks for the PR! I also tend to have this in almost every application, and I believe this is a good candidate for inclusion. 

There is already a `inner` method:

https://github.com/ratatui-org/ratatui/blob/59b9c32fbc2bc6725bdec42e63216024fab71493/src/layout/rect.rs#L93

That people can use like this:

```rust
Rect::new(1, 2, 3, 4).inner(&Margin::new(1, 2)),
```

I'm not sure if this can be done in a non-breaking manner, but it would have been nice to use the same function for different purposes.

e.g.


```rust
Rect::new(1, 1, 10, 10).inner(Margin::px(1, 2))
Rect::new(1, 1, 10, 10).inner(Margin::percent(10, 20))
```

joshka:
Definitely something more generic than this would be preferable. I'm not sure exactly how the API would look, perhaps using `Constraint`s for specifying things? Let's chat on that in #617

joshka:
See notes in #617 about using flex instead of this approach.

joshka:
Given that this is too specific (i.e. it only deals with creating a rectangle that is a certain percentage high / wide, and not a rectangle with any other constraints, nor multiple rectangles), I'm going to close this out. Thank you for the PR even if we didn't merge it.

The replacement code for this is effectively the following, which clearly expresses the intended effect.
```rust
let [area] = Layout::horizontal([Constraint::Percentage(50)]).flex(Flex::Center).areas(area);
let [area] = Layout::vertical([Constraint::Percentage(20)]).flex(Flex::Center).areas(area);
```

The only part that is really missing here is a neat syntax for working in two dimensions at once in a single statement. This seems like an orthogonal (pun intended) concern to this PR.

I'll update https://ratatui.rs/recipes/layout/center-a-rect/ shortly.

wookietreiber:
Sorry about not responding sooner, thanks for keeping up with this :+1:

joshka:
No problem :)

Issue #621, opened 2023-11-08
RFC: Experimental API features


joshka:
## Problem
It would be nice if we could create changes to the API that we mark as experimental so that features can be added in a way that doesn't need the same amount of care taken for getting it exactly right the first time round. This would allow us to deliver features like #618 with less concern about breaking customer code in the future.

## Solution
This is likely a combination of:
- an `experimental` module - unsure of this or whether we just put the code where it will live - there's pros and cons to both approaches.
- an `experimental` feature - do we want one feature per experiment or one for all experiments?

Does anyone have some ideas about how other crates handle this?

mindoodoo:
I'm curious to know what are some of the pros you see in having an `experimental` module ?

I think the main thing is that experimental features should be opt in. There's an [open rfc issue](https://github.com/rust-lang/rfcs/issues/1491) asking for an `#[unstable]` attribute for library authors, however it does seem quite stale. Alternatively there is also the [`stability`](https://github.com/sagebind/stability) crate which we could try out or create a similar macro. Something quite nice it does is that it expands some doc comments warning that the feature might change.

joshka:
The rationale is that it makes things explicit to the user that the interface may change wildly between releases as the needs of the code changes. It's a "there be dragons" approach. I've seen this used in various libs in other languages to allow for things which will become core to be stabilized close to core.

mindoodoo:
Oh no I mean specifically for the having a specific module for experimental features ? I would tend to think that a feature flag would be enough, and that on the contrary we would probably prefer to have the code for experimental features live in the place that they will live once / if stabilized.

joshka:
Got it. The difference is that if you have an experimental module, then you're able to migrate existing code to the stabilized form without breaking the last version of the experimental (and can mark it as deprecated between that time). Whether that's useful / necessary is the thing to think about

joshka:
@mindoodoo I just saw your star on https://docs.rs/stability/latest/stability/index.html in my GH feed - that looks promising.

mindoodoo:
> Got it. The difference is that if you have an experimental module, then you're able to migrate existing code to the stabilized form without breaking the last version of the experimental

Interesting idea, though I think anyone opting into an experimental part of the crate would be ready to make the changes necessary once the experimental parts are stabilized, especially since they'll need to do it sooner or later if we deprecate and eventually remove the experimental feature in the experimental module.

Yeah stability seems quite nice, I think it does the two main things we want to have :
[...]

PR #623, opened 2023-11-11
feat: add popup to layout

wookietreiber:
Same as with #618: I'm leaving this in draft for now and won't be adding tests or fixing coverage just yet. I'm putting in the effort of accompanying the PR/variant with tests/coverage when the decision is made. 

---

With this variant, the idea is to clearly communicate the intent for creating popups with a distinct struct in the layout module. Alternatively, if you're opposed to the struct, this could be done with plain functions and without the re-export in `src/layout.rs`, as in `ratatui::layout::popup::percentage(x, y, rect)`, but IMO the struct, which can easily be re-exported in `prelude.rs` like I did here, serves as a better, more stream-lined entrypoint for users than a module with plain functions would. 

The other use cases mentioned in https://github.com/ratatui-org/ratatui/issues/617#issuecomment-1801173192, i.e. popup with explicit size, based on margins, I can add later. With this variant though I'd avoid using generics and rather have individual functions, for the same reason as mentioned before: clearly communicate intent and thus more easily understandable API. 

--- 

fixes #617

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -94,7 +94,11 @@ underline-color = ["dep:crossterm"]
 #! The following features are unstable and may change in the future:
 
 ## Enable all unstable features.
-unstable = ["unstable-segment-size", "unstable-rendered-line-info"]
+unstable = ["unstable-popup-layout", "unstable-segment-size", "unstable-rendered-line-info"]
+
+## Enables [`Popup`](crate::layout::Popup) as a layout which is experimental and may change in the
+## future. See [Issue #617](https://github.com/ratatui-org/ratatui/issues/617) for more details.
+unstable-popup-layout = []
 
 ## Enables the [`Layout::segment_size`](crate::layout::Layout::segment_size) method which is experimental and may change in the
 ## future. See [Issue #536](https://github.com/ratatui-org/ratatui/issues/536) for more details.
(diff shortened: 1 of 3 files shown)


joshka:
I think at this point, this seems too hyper-specific (or put another way, it solves just one problem which I'm not sure is the right problem to solve). To validate that I'd ask several questions:
- why is centered the right way to render this and not right / left aligned? 
- why is this called popup?
- what inherently makes a popup centered?
- more questions generally about how the API evolves from here

I suspect the right way to prototype this would be to do so in a second crate and bring the more complete ideas into the core once shown to work well for several apps. Alternatively consider helping spec https://github.com/ratatui-org/ratatui/issues/621 first

Right now the methods in Layout are mostly 1 dimensional rather than 2D - we should probably fix that by moving to a taffy based implementation that does flex/grid layout rather than going more specific like this.

wookietreiber:
I've updated this draft to use `stability::unstable`.

Valentin271:
There is now [tui-popup](https://github.com/joshka/tui-popup) which is dedicated to building popups. Maintained by @joshka.

I'm closing this. Feel free to reopen if you think it is still relevant.

wookietreiber:
This is awesome! Thanks @joshka! I already started my own **tui-popup** crate, including the three variants I proposed here, but I wasn't ready to publish yet :sweat_smile: I'll try out your **tui-popup**.

joshka:
There's also https://github.com/sephiroth74/tui-confirm-dialog to take a look at - not mine - but recent also. Uses channels to communicate back to the app, which is a neat but more complex approach.

PR #628, opened 2023-11-15
fix(table): widths() now accepts AsRef<[Constraint]>

joshka:
This allows passing an array, slice or Vec of constraints, which is more
ergonomic than requiring this to always be a slice.

The following calls now all succeed:

```rust
Table::new(rows).widths([Constraint::Length(5), Constraint::Length(5)]);
Table::new(rows).widths(&[Constraint::Length(5), Constraint::Length(5)]);

// widths could also be computed at runtime
let widths = vec![Constraint::Length(5), Constraint::Length(5)];
Table::new(rows).widths(widths.clone());
Table::new(rows).widths(&widths);
```

Fixes: https://discord.com/channels/1070692720437383208/1072907135664529508/1174137200909221988

---

> I have a ratatui table widget which I return from a function, but due to the widths function requiring a reference to a value, it yells at me for referencing a temporary value
```rs
error[E0515]: cannot return value referencing temporary value
  --> src\ui.rs:89:5
   |
89 | /     Table::new(rows)
90 | |         .header(header)
91 | |         .block(Block::default().borders(Borders::ALL).title("Table"))
92 | |         .highlight_symbol(">> ")
93 | |         .widths(&(widths.clone()))
   | |__________________----------------^ returns a value referencing data owned by the
 current function
   |                    |
   |                    temporary value created here
```

```rs
fn draw_table<'a>(items: &'a project::Items, fields: &'a project::Fields) -> Table<'a> {
    let header_cells = fields.fields
        .iter()
        .map(|h| Cell::from(h.name.clone())
            .style(Style::default().fg(Color::Red)));

    let header = Row::new(header_cells)
        .height(1)
        .bottom_margin(0);

    let empty_cell = Value::String(String::from(""));

    let rows = items.items.iter().map(|item| {
        let cells = fields.fields.iter()
            .map(move |f|
                item.fields
                .get(&*f.name.to_ascii_lowercase()).unwrap_or(&Value::Bool(false))
                .as_str().unwrap_or(""))
            .collect::<Vec<&str>>();

        Row::new(cells).height(1).bottom_margin(1)
    });

    let split = (100 / fields.fields.len()) as u16;
    let widths: Vec<Constraint> = (0..fields.fields.len()).map(
        |n| Constraint::Percentage(split)
    ).collect::<Vec<Constraint>>();

    Table::new(rows)
        .header(header)
        .block(Block::default().borders(Borders::ALL).title("Table"))
        .highlight_symbol(">> ")
        .widths(&(widths.clone()))
}
```
> this is the whole function

### DIFF
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -10,6 +10,9 @@ github with a [breaking change] label.
 
 This is a quick summary of the sections below:
 
+- Unreleased (0.24.1)
+  -
+
 - [v0.24.0](#v0240)
   - MSRV is now 1.70.0
   - `ScrollbarState`: `position`, `content_length`, and `viewport_content_length` are now `usize`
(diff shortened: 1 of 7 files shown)


joshka:
(added note to BREAKING-CHANGES.md about clippy lints.

orhun (approved):

Issue #502, opened 2023-09-14
Tabs: Remove initial space prior to `Tabs` widget and around tab dividers inside


joshka:
## Problem
The Tabs widget creates a non configurable space prior to the beginning of the tab, and non removable spaces between the tab text and the divider. E.g. in the following, the space prior to "Inbox" and two of the spaces between "Inbox" and "Sent" are non-removable.

<img width="340" alt="image" src="https://github.com/ratatui-org/ratatui/assets/381361/251f404b-c0ec-4f26-8323-296f59d778e6">

```rust
Tabs::new(vec![" Inbox ", " Sent ", " Drafts "])
    .style(Style::new().fg(Color::Indexed(244)).bg(Color::Indexed(232)))
    .highlight_style(
        Style::new()
            .bold()
            .fg(Color::Indexed(232))
            .bg(Color::Rgb(64, 96, 192)),
    )
    .select(0)
    .divider("")
    .render(area[0], buf);
```
Both of these spaces can be got to by alternative means (rendering to an area with a margin, or changing the divider to include spaces either side of the divider)

## Solution

- Remove the initial space
- Change the default divider to `" │ "` from `"│"` and remove the excess spaces

This breaks rendering in two places:
- Apps that want the extra initial space can fix this by putting a margin around the tabs. E.g.: `let area = area.inner(&Margin::(new(1,0)));`
- Apps that use a custom divider can fix this by adding spaces either side of the divider. E.g.: `.divider(" • ")` 

## Alternatives
1. Make the amount of extra space a configuration option with the default set to the current amount
2. Make the extra space characters configurable with the default option set to a space

I think the main solution is probably reasonable - but so are the alternatives that don't break existing code, so I'd like to get consensus before implementing this.

bhavuk2002:
i can do that, assign me

joshka:
Hey there - I'd like us wait a little bit for some feedback from users before we jump to PRs on this:

> I think the main solution is probably reasonable - but so are the alternatives that don't break existing code, so I'd like to get consensus before implementing this.

bhavuk2002:
okay, what can i do i the mean while? as you have changed the title, i have done the same in my PR. can you suggest some good first issue on which i can work on? or for this issue i can do something?

joshka:
[...]

PR #629, opened 2023-11-16
fix(tabs): Tab widget now supports custom padding (Fixes issue #502))

rhaskia:
The tab widget now contains the values padding_left/padding_right and the function padding() that sets those values.
Should fix issue #502.


joshka:
Cool. I didn't realize that the spacing before the tabs was coming from padding left / right. My mental model without having looked at the code was that the spaces "belonged to the divider. This solution is better than the ones on the Issue.

A question though, what if we made the padding accept  `&str` instead of just a number?

A follow up to this might be to make padding accept `Line` to make the divider / padding able to be styled easily.
Imagine:
```
🭊🭁Tab 1🭌🬿 🭊🭁Tab 2🭌🬿
```
But with the background of the label part the same color as the tab padding parts.

(copied from the discord feedback so that it's recorded for posterity)

src/widgets/tabs.rs
```diff
@@ -143,22 +143,18 @@ impl<'a> Tabs<'a> {
...
+    /// Nothing on either side of the tabs.
     /// ```
     /// # use ratatui::{prelude::*, widgets::Tabs, symbols};
-    /// let tabs = Tabs::new(vec!["Tab 1", "Tab 2"]).padding(0, 0);
+    /// let tabs = Tabs::new(vec!["Tab 1", "Tab 2"]).padding("", "");
     /// ```
-    pub fn padding(mut self, left: u16, right: u16) -> Tabs<'a> {
-        self.padding_left = left;
-        self.padding_right = right;
+    pub fn padding<T>(mut self, left: T, right: T) -> Tabs<'a> where T: Into<Line<'a>> {
```
joshka, line 155: Looking at both left and right having to be the same `T` here, I think this would work better as two methods rather than one. This won't otherwise allow `padding("a", "b".red())`

joshka (approved):

joshka (changes requested):
approved by mistake

src/widgets/tabs.rs
rhaskia: Oh true, I am away for a few days but I will correct that once I am back.

src/widgets/tabs.rs
```diff
@@ -44,6 +44,8 @@ pub struct Tabs<'a> {
     padding_left: Line<'a>,
     /// Tab Right Padding
     padding_right: Line<'a>,
+    /// Whether padding gets highlighted when selected or not
+    highlight_padding: bool,
```
joshka, line 48: Let's use an enum for this. https://rust-lang.github.io/api-guidelines/type-safety.html#c-custom-type

src/widgets/tabs.rs
```diff
@@ -131,6 +140,70 @@ impl<'a> Tabs<'a> {
...
         self
     }
+
+    /// Sets the padding between tabs.
+    ///
+    /// # Examples
+    ///
+    /// A space on either side of the tabs.
+    /// ```
+    /// # use ratatui::{prelude::*, widgets::Tabs, symbols};
```
joshka, line 150:
```suggestion
    /// # use ratatui::{prelude::*, widgets::Tabs};
```

src/widgets/tabs.rs
```diff
@@ -131,6 +140,70 @@ impl<'a> Tabs<'a> {
...
+    /// let tabs = Tabs::new(vec!["Tab 1", "Tab 2"]).padding(" ", " ");
+    /// ```
+    /// Nothing on either side of the tabs.
+    /// ```
+    /// # use ratatui::{prelude::*, widgets::Tabs, symbols};
```
joshka, line 155:
```suggestion
    /// # use ratatui::{prelude::*, widgets::Tabs};
```

src/widgets/tabs.rs
```diff
@@ -131,6 +140,70 @@ impl<'a> Tabs<'a> {
...
+    /// Sets the left side padding between tabs.
+    ///
+    /// # Example
+    ///
+    /// An arrow on the left of tabs.
+    /// ```
+    /// # use ratatui::{prelude::*, widgets::Tabs, symbols};
+    /// let tabs = Tabs::new(vec!["Tab 1", "Tab 2"]).left_padding("->");
+    /// ```
+    pub fn left_padding<T>(mut self, padding: T) -> Tabs<'a>
```
joshka, line 177: nit: I think I prefer left/right at the end. What do you think?
```suggestion
    pub fn padding_left<T>(mut self, padding: T) -> Tabs<'a>
```

joshka (commented):
We can include the padding part of the tabs in the highlight by adding it to the tab titles instead of the padding. So I wonder whether it would be the worst thing to omit the highlight part of this?

Can you please add some docs about padding to the main rustdoc comment for the type? The doc needs info on the defaults for the padding and divider.

I'm going to make a breaking change for highlight_style in another PR as it seems natural to make that actually do something reasonable in the default configuration.

joshka:
Hey, would you mind setting up commit signing?
https://github.com/ratatui-org/ratatui/blob/main/CONTRIBUTING.md#sign-your-commits

rhaskia:
I did, I think something went wrong as it says the email does not match the signature.

joshka:
Github mangles your email address to `94507929+RhaskiaGameDev@users.noreply.github.com` is that setup as being valid for signing?

joshka (approved):

joshka:
Thanks for fixing up the commit signing - Github still complains that it can't merge them as it needs all the commits to be signed. Can you please squash the commits into a single signed commit and force push the single commit as an update to your branch?

joshka:
Just merged this - thanks again for the PR :)

PR #633, opened 2023-11-18
feat(layout): allow layout fill to be configurable

joshka:
Method docs:

image

Type docs:

image

Note: I fixed the link to the `Table` widget - just didn't re-screenshot.


joshka:
Updated commit message to work better in the changelog

PR #635, opened 2023-11-19
fix(tabs)!: set the default highlight_style

joshka:
Previously the default highlight_style was set to `Style::default()`,
which meant that the highlight style was the same as the normal style.
This change sets the default highlight_style to reversed text.

BREAKING CHANGE: The `Tab` widget now renders the highlight style as
reversed text by default. This can be changed by setting the
`highlight_style` field of the `Tab` widget.


orhun (approved):
I like this change, LGTM!

joshka:
Added commit just fixes up conflict in breaking_changes.md

orhun:
Ping @kdheepak for additional review.

PR #645, opened 2023-11-25
style: reformat using rust nightly

joshka:
This is a PR to highlight the upcoming formatting changes in nightly, which will impact anyone that is currently working on any ratatui changes (and likely cause our CI to start failing with issue unrelated to the actual code). We should look into the status of the formatting changes to understand how likely they are to end up in stable. 

Most of the changes look nice, but I'm not sure I understand why a bunch of docs are being removed. This might be a bug.


src/backend/crossterm.rs
```diff
@@ -88,14 +88,6 @@ where
...
 {
     /// Creates a new `CrosstermBackend` with the given writer.
-    ///
-    /// # Example
-    ///
-    /// ```rust,no_run
-    /// # use std::io::stdout;
-    /// # use ratatui::prelude::*;
-    /// let backend = CrosstermBackend::new(stdout());
-    /// ```
```
joshka, line 98: rustfmt bug?

joshka: https://github.com/rust-lang/rustfmt/issues/5965

joshka:
Looks like the rustfmt bug is fixed in the latest nightly

PR #646, opened 2023-11-25
docs(Table): documented Table

Valentin271:
Added documentation to the Table widget.


joshka:
As they say "great minds think alike (or fools never differ ;))" - https://github.com/ratatui-org/ratatui/pull/638

Would you mind reviewing #638 for me instead of this?

Valentin271:
Didn't see that one, yep sure!

PR #600, opened 2023-10-28
fix: add #[must_use] attributes on (self: Style) methods

SOF3:

PR #647, opened 2023-11-26
feat(span): add setters for content and style

joshka:

### DIFF
--- a/src/text/span.rs
+++ b/src/text/span.rs
@@ -11,6 +11,26 @@ use crate::style::{Style, Styled};
 /// A `Span` is the smallest unit of text that can be styled. It is usually combined in the [`Line`]
 /// type to represent a line of text where each `Span` may have a different style.
 ///
+/// # Constructor Methods
+///
+/// - [`Span::default`] creates an span with empty content and the default style.
+/// - [`Span::raw`] creates an span with the specified content and the default style.
+/// - [`Span::styled`] creates an span with the specified content and style.
+///
+/// # Setter Methods
+///
+/// These methods are fluent setters. They return a new `Span` with the specified property set.
+///
+/// - [`Span::content`] sets the content of the span.
+/// - [`Span::style`] sets the style of the span.
+///
+/// # Other Methods
+///
+/// - [`Span::patch_style`] patches the style of the span, adding modifiers from the given style.
+/// - [`Span::reset_style`] resets the style of the span.
+/// - [`Span::width`] returns the unicode width of the content held by this span.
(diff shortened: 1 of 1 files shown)


src/text/span.rs
```diff
@@ -100,6 +120,78 @@ impl<'a> Span<'a> {
...
+    /// Accepts any type that can be converted to [`Cow<str>`] (e.g. `&str`, `String`, `&String`,
+    /// etc.).
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// # use ratatui::prelude::*;
+    /// let mut span = Span::default().content("content");
+    /// ```
+    #[must_use = "builder methods are lazy and do not modify the underlying object"]
```
Valentin271, line 134: I like this! Never thought of using it on builder methods.

Valentin271 (approved):

src/text/span.rs
joshka: Credit to @SOF3 in https://github.com/ratatui-org/ratatui/pull/600 for this

SOF3: For builders, there is no reason to construct a builder without using it anyway. Maybe better put `#[must_use]` on the `struct Builder` rather than its methods?

SOF3:
Actually these should be called "fluent setters" not "builder methods". Builder methods usually refer to a separate builder object that may have incomplete fields and need to be consumed to become the built object, e.g. https://docs.rs/typed-builder.

src/text/span.rs
```diff
@@ -11,6 +11,26 @@ use crate::style::{Style, Styled};
...
 ///
+/// # Constructor Methods
+///
+/// - [`Span::default`] creates an span with empty content and the default style.
+/// - [`Span::raw`] creates an span with the specified content and the default style.
+/// - [`Span::styled`] creates an span with the specified content and style.
+///
+/// # Setter Methods
+///
+/// These methods a fluent setters. They return a new `Span` with the specified property set.
```
Valentin271, line 22: I think you meant "are" here
```suggestion
/// These methods are fluent setters. They return a new `Span` with the specified property set.
```

Valentin271 (approved):
Other than the typo, looks good

joshka:
Rebased and squashed - should be good to go

joshka:
Tweaked must_use message to:
```rust
```
Based on discussion in https://github.com/ratatui-org/ratatui/pull/655

Valentin271 (approved):
Fine by me, approved

PR #651, opened 2023-11-27
docs: remove simple-tui-rs

joshka:
This has not been recently and doesn't lead to good code

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,6 @@ render "Hello World!". The full code for this example which contains a little mo
 
 - [rust-tui-template]
 - [ratatui-async-template] (book and template)
-- [simple-tui-rs]
 
 Every application built with `ratatui` needs to implement the following steps:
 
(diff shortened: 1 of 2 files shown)


orhun:
ping @pmsanford

joshka:
The context on this is that a user in discord mentioned they were using this repo as a base because it seemed like the right idea. It may have been way back when tui-rs was just a young project, but it's got a few things would be better not to use in a newer project.
- Termion
- `Spans` instead of `Line`
- etc.

pmsanford:
Thanks for the ping; I'd be happy to update it, but I can't make any commitments on a timeline at the moment. I'd suggest going ahead with merging this PR and when/if I get it updated I'll open one to re-add it. I'll also add a note to my repo saying it's not up to date and link to this PR.

orhun:
Sounds good, thank you!

orhun (approved):

PR #657, opened 2023-12-02
fix(block): make `inner` aware of title positions

### LINKED ISSUE #658: bottom-positioned title in block with bottom border prevents rendering of paragraph content when drawing area has height 2
## Description

Rendering a block with title and border at the bottom in an area of height 2 makes the associated paragraphs' content disappear.

## To Reproduce

``` rust
let block = Block::default()
    .title("Title")
    .title_position(Position::Bottom)
    .borders(Borders::BOTTOM);
let paragraph = Paragraph::new("Hello, world!").block(block);

let area = Rect::new(0, 0, 15, 2);
paragraph.render(area, &mut buffer);
println!("{buffer:?}");
```

```
Buffer {
    …
    content: [
        "               ",
        "Title──────────",
    ],
    …
}
```

## Expected behavior

```
Buffer {
    …
    content: [
        "Hello, world!  ",
        "Title──────────",
    ],
    …
}
```

## Environment

- Crate version: 0.24.1-alpha.5

Since I'm using the `ratatui::buffer::Buffer` in the minimal working example, I feel like other environment information is not necessary. Please let me know if that assumption is incorrect.

## Additional context

- A fix is proposed in #657

jan-ferdinand:
Previously, when computing the inner rendering area of a block, all titles were assumed to be positioned at the top. If some or all titles were at the bottom, this caused various incorrect computations of the inner area, depending on which borders were set; no border at the bottom would lead to an area that's too big, no border at the top but at the bottom and _all_ titles at the bottom would lead to an area that's too small.

Now, when computing the inner area, the positions of the titles are taken into account.

fix #658

### DIFF
--- a/src/widgets/block.rs
+++ b/src/widgets/block.rs
@@ -508,14 +508,15 @@ impl<'a> Block<'a> {
             inner.x = inner.x.saturating_add(1).min(inner.right());
             inner.width = inner.width.saturating_sub(1);
         }
-        if self.borders.intersects(Borders::TOP) || !self.titles.is_empty() {
+        if self.borders.intersects(Borders::TOP) || self.have_title_at_position(Position::Top) {
             inner.y = inner.y.saturating_add(1).min(inner.bottom());
             inner.height = inner.height.saturating_sub(1);
         }
         if self.borders.intersects(Borders::RIGHT) {
             inner.width = inner.width.saturating_sub(1);
         }
-        if self.borders.intersects(Borders::BOTTOM) {
+        if self.borders.intersects(Borders::BOTTOM) || self.have_title_at_position(Position::Bottom)
+        {
             inner.height = inner.height.saturating_sub(1);
         }
 
@@ -530,6 +531,12 @@ impl<'a> Block<'a> {
             .saturating_sub(self.padding.top + self.padding.bottom);
 
         inner
+    }
+
+    fn have_title_at_position(&self, position: Position) -> bool {
+        self.titles
+            .iter()
+            .any(|title| title.position.unwrap_or(self.titles_position) == position)
     }
 
     /// Defines the padding inside a `Block`.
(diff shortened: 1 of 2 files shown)


orhun (commented):
Can you also add a test case where you cover the issue in #658? (you can assert the contents of the buffer with `assert_buffer`)

jan-ferdinand:
> add a test case where you cover the issue in #658

Good suggestion. I have amended the PR accordingly. Let me know if anything can be improved. :relaxed:

src/widgets/block.rs
```diff
@@ -532,6 +533,16 @@ impl<'a> Block<'a> {
...
     }
 
+    fn have_title_at_position(&self, position: Position) -> bool {
+        if self.titles.iter().any(|t| t.position == Some(position)) {
+            return true;
+        }
+        if self.titles.iter().any(|t| t.position.is_none()) {
+            return self.titles_position == position;
+        }
+        false
```
joshka, line 543: slightly simpler (nit):

```suggestion
        self.titles
            .iter()
            .any(|title|
                title.position == Some(position)
                || title.position.is_none() && self.titles_position == position)
```

joshka (approved):
Can you please suggest a simple and concise explanation of the bug?
If you're able, add it to the body of a commit message to make it automatically fill the changelog. If not, we can add the wording when squashing the commit.

src/widgets/block.rs
jan-ferdinand: How about:

``` rust
        self.titles.iter().any(|title| match title.position {
            Some(title_position) => title_position == position,
            None => self.titles_position == position,
        })
```

jan-ferdinand:
I have added an explanation of the bug and the fix to both the description of this PR and the commit message. Let me know should brevity be lacking, or if the level of detail could be improved upon (in either direction).

Also, I have reworked the part that @joshka had a nit about, without accepting the suggestion. If you prefer to have your suggestion exactly, I'll incorporate that instead.

src/widgets/block.rs
joshka: Good idea, but you could go one step futher still: :)
```rust
self.titles.iter.any(|title| position == title_position.unwrap_or(self.titles_position))
```

joshka (approved):

jan-ferdinand:
Thanks for the shortened bug description as well as the improved code snippet. :+1: I've incorporated both and amended the PR accordingly.

orhun:
Thanks for your contribution!

PR #660, opened 2023-12-03
feat(table): Add a Table::segment_size method

### LINKED ISSUE #370: Table: Add option to make `expand_to_fill` configurable for `Table::widths`

asomers:
It works just like Layout::segment_size, but for Tables.  Previously, it was impossible to distribute extra space anywhere in a Table.

Fixes #370

### DIFF
--- a/src/widgets/table.rs
+++ b/src/widgets/table.rs
@@ -267,6 +267,8 @@ pub struct Table<'a> {
     rows: Vec<Row<'a>>,
     /// Decides when to allocate spacing for the row selection
     highlight_spacing: HighlightSpacing,
+    /// Controls how to distribute extra space among the columns
+    segment_size: SegmentSize,
 }
 
 impl<'a> Table<'a> {
@@ -304,6 +306,7 @@ impl<'a> Table<'a> {
             header: None,
             rows: rows.into_iter().collect(),
             highlight_spacing: HighlightSpacing::default(),
+            segment_size: SegmentSize::None,
         }
     }
 
@@ -433,7 +436,7 @@ impl<'a> Table<'a> {
         let layout = Layout::default()
             .direction(Direction::Horizontal)
             .constraints(constraints)
-            .segment_size(SegmentSize::None)
+            .segment_size(self.segment_size)
             .split(Rect::new(0, 0, max_width, 1));
         layout
             .iter()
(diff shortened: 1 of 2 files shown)


src/widgets/table.rs
```diff
@@ -267,6 +267,7 @@ pub struct Table<'a> {
     rows: Vec<Row<'a>>,
     /// Decides when to allocate spacing for the row selection
     highlight_spacing: HighlightSpacing,
+    segment_size: SegmentSize,
```
joshka, line 270: Nit: Add a comment here

src/widgets/table.rs
```diff
@@ -475,6 +477,21 @@ impl<'a> Table<'a> {
         }
         (start, end)
     }
+
+    /// Builder method to control how extra space is distributed amongst columns.
+    ///
+    /// This determines how the space is distributed when the constraints are satisfied. By default,
+    /// the extra space is not distributed at all.  But this can be changed to distribute all extra
+    /// space to the last column or to distribute it equally.
```
joshka, line 485: Add examples something like...

```rust
let area = Rect::new(0, 0, 62, 1);
let widths = [Min(10), Min(10, Min(10)];
// results in column widths 10, 10, 10
Table::new(rows).widths(widths).segment_size(SegmentSize::None);
// ...
```

joshka (changes requested):
LGTM - 1 small nit.

Can you please update the commit message (or provide one in a comment) that describes the behavior directly rather than in reference to `Layout`. This will end up in the Changelog, so think about how you'd announce the functionality rather than just describing the change and perhaps add 3 small code examples like

```rust
let area = Rect::new(0, 0, 62, 1);
let widths = [Min(10), Min(10, Min(10)];
// results in column widths 10, 10, 10
Table::new(rows).widths(widths).segment_size(SegmentSize::None);
// ...
```

asomers:
Does your CI process merge PRs with the main branch before running tests?  Because the test that's failing is one that isn't in my branch.  Neat, if true.

joshka:
> Does your CI process merge PRs with the main branch before running tests? Because the test that's failing is one that isn't in my branch. Neat, if true.

Yep :)
If it was the tabs - I just fixed that issue, so if you rebase it should be fine

asomers:
Rebased.  And I squashed my commits to satisfy the commit style checker.

orhun (approved):

joshka (approved):

PR #663, opened 2023-12-04
feat(layout): Accept IntoIterator for constraints

joshka:
- style(tabs): fix doc formatting
- feat(layout): Accept IntoIterator for constraints

### DIFF
--- a/src/layout.rs
+++ b/src/layout.rs
@@ -97,6 +97,12 @@ impl Default for Constraint {
     }
 }
 
+impl AsRef<Constraint> for Constraint {
+    fn as_ref(&self) -> &Constraint {
+        self
+    }
+}
+
 impl fmt::Display for Constraint {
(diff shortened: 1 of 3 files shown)


src/layout.rs
```diff
@@ -246,23 +252,41 @@ pub struct Layout {
     segment_size: SegmentSize,
 }
 
-impl Default for Layout {
-    fn default() -> Layout {
-        Layout::new(Direction::Vertical, [])
-    }
-}
```
orhun, line 253: Do we make sure with any tests that the default `Default` implementation and this implementation produce the same results?

joshka: Yes now in 9c732ca9

orhun:
Not sure 7b0ef3d85caf47cd5027f8517a7ba1cdfa80a760 belongs to this PR 🐻

orhun (approved):

PR #667, opened 2023-12-04
fix(list): Modify List and List example to support saving offsets.

bblsh:
The current `List` example will unselect and reset the position of a list.

This PR will save the last selected item, and updates `List` to honor its offset, preventing the list from resetting when the user `unselect()`s a `StatefulList`.

### DIFF
--- a/src/widgets/list.rs
+++ b/src/widgets/list.rs
@@ -738,6 +738,7 @@ impl<'a> List<'a> {
         self.items.is_empty()
     }
 
+    /// Given an offset, calculate which items can fit in a given area
     fn get_items_bounds(
         &self,
         selected: Option<usize>,
@@ -745,35 +746,69 @@ impl<'a> List<'a> {
         max_height: usize,
     ) -> (usize, usize) {
         let offset = offset.min(self.items.len().saturating_sub(1));
-        let mut start = offset;
-        let mut end = offset;
-        let mut height = 0;
+
+        // Note: visible here implies visible in the given area
+        let mut first_visible_index = offset;
+        let mut last_visible_index = offset;
+
+        // Current height of all items in the list to render, beginning at the offset
+        let mut height_from_offset = 0;
+
+        // Calculate the last visible index and total height of the items
+        // that will fit in the available space
         for item in self.items.iter().skip(offset) {
-            if height + item.height() > max_height {
+            if height_from_offset + item.height() > max_height {
                 break;
             }
-            height += item.height();
-            end += 1;
+
+            height_from_offset += item.height();
+
+            last_visible_index += 1;
         }
 
-        let selected = selected.unwrap_or(0).min(self.items.len() - 1);
-        while selected >= end {
-            height = height.saturating_add(self.items[end].height());
-            end += 1;
-            while height > max_height {
-                height = height.saturating_sub(self.items[start].height());
-                start += 1;
+        // Get the selected index, but still honor the offset if nothing is selected
+        // This allows for the list to stay at a position after select()ing None.
+        let index_to_display = selected.unwrap_or(offset).min(self.items.len() - 1);
+
+        // Recall that last_visible_index is the index of what we
+        // can render up to in the given space after the offset
+        // If we have an item selected that is out of the viewable area (or
+        // the offset is still set), we still need to show this item
+        while index_to_display >= last_visible_index {
+            height_from_offset =
+                height_from_offset.saturating_add(self.items[last_visible_index].height());
+
+            last_visible_index += 1;
+
+            // Now we need to hide previous items since we didn't have space
+            // for the selected/offset item
+            while height_from_offset > max_height {
+                height_from_offset =
+                    height_from_offset.saturating_sub(self.items[first_visible_index].height());
+
+                // Remove this item to view by starting at the next item index
+                first_visible_index += 1;
             }
         }
-        while selected < start {
-            start -= 1;
-            height = height.saturating_add(self.items[start].height());
-            while height > max_height {
-                end -= 1;
-                height = height.saturating_sub(self.items[end].height());
+
+        // Here we're doing something similar to what we just did above
+        // If the selected item index is not in the viewable area, let's try to show the item
+        while index_to_display < first_visible_index {
+            first_visible_index -= 1;
+
+            height_from_offset =
+                height_from_offset.saturating_add(self.items[first_visible_index].height());
+
+            // Don't show an item if it is beyond our viewable height
+            while height_from_offset > max_height {
+                last_visible_index -= 1;
+
+                height_from_offset =
+                    height_from_offset.saturating_sub(self.items[last_visible_index].height());
             }
         }
-        (start, end)
+
+        (first_visible_index, last_visible_index)
     }
 }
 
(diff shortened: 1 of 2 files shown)


joshka:
I personally think this PR is probably the right thing to do (i.e. don't reset offset when unselecting). It was mentioned recently in another PR / issue, but I don't know the history of why this current behavior is like it is. Would you mind digging into the tui-rs history to find some details on the existing behavior if possible? The questions I'd generally ask about odd behavior like this are (a non-exhaustive list):

1. Was it always like this?
2. When and why did it change?
3. Was there any discussion about it?
4. Are there any bug reports about the behavior?
5. What does it break when we change it (is this a load bearing bug)?

Assuming that we go ahead with this change, this PR needs a few extra things:
- BREAKING CHANGES info
- Docs for the new functionality
- Tests for the change

orhun:
I like this change and I think it is something nice to have in the long run.

Seconding Josh here and I would like to add `git blame` would probably help figuring out the historical changes that led to the current behavior.

bblsh:
I've looked through the discussions and issues in tui-rs and found nothing. I've also browsed the git history, which indicates this behavior was added when `ListItem`s were introduced. The commit message for that change was logged as `add new text primitives`.

That said I'm still unsure if this was intentional behavior after reading through these bits.

However, this change to `list.rs`is _not_ breaking and does _not_ change existing behavior unless the user makes changes to their own `StatefulList` as shown in the PR. Take a look at the existing `ListState`'s `select()`method:
```
pub fn select(&mut self, index: Option<usize>) {
    self.selected = index;
    if index.is_none() {
        self.offset = 0;
    }
}
```
Currently, the offset will _always_ be 0 when `unselect()`ing. 
Now let's look at the `unselect()` example method historically provided for a `StatefulList`:
```
fn unselect(&mut self) {
    self.state.select(None);
}
```
Finally, let's look at the pre-PR line in `list.rs:191`:
```
let selected = selected.unwrap_or(0).min(self.items.len() - 1);
```

With the pre-PR setup, whenever the user `unselect()`s a list, changing the offset will not work correctly since `selected` will be `None`. With this change, the user can still `unselect()` and keep behavior that currently exists, but also have the ability to retain position by resetting the offset when `unselect()`ing as shown below (in the PR):
```
    fn unselect(&mut self) {
        let offset = self.state.offset();
        self.last_selected = self.state.selected();
        self.state.select(None);
        *self.state.offset_mut() = offset;
    }
```

Because of this I do not believe this change is breaking or changes pre-existing behavior.

joshka:
Thanks for doing the history dive on. Your findings make me much less concerned that it was intentional and more like something that just seemed right at the time.

I have to admit that I don't know this code deeply (you're more of an expert on it right now than anyon), but if you can change a line in a private method without breaking any unit tests, I think there's a problem somewhere - possibly in the existing unit test coverage of that line. Incidentally I can change it to 1 or 2 and all the tests still pass (but not 3).

Can you add a test that shows how this change impacts things (rendering / state)? It might be helpful to the next person to look at that code to also to add a comment that explains why that method exists. If you can discern the intent, writing it down helps validate that the change is correct.

bblsh:
Two things:
- The `List`'s `StatefulWidget::render()` method now has plenty of comments to explain what is happening.
- A unit test was added to display this functionality in a couple ways.

The gist of why that test was previously failing is because of the way rendering happens for the `List`.
The new unit test added here in the PR shows why your mentioned test was failing, and what should have been expected.

src/widgets/list.rs
```diff
@@ -186,35 +186,82 @@ impl<'a> List<'a> {
...
     ) -> (usize, usize) {
+        // In case the offset is greater than the number of items, set the offset
+        // to be the index of the last item. Otherwise, use the provided offset
         let offset = offset.min(self.items.len().saturating_sub(1));
+
+        // Index of the first list item we're rendering
         let mut start = offset;
+
+        // Index of the last item item we're able to render in the given area
         let mut end = offset;
```
orhun, line 197: I'm not a big fan of documenting everything and I think some of the variables/logic (e.g. this one) explains itself. I would like to get the opinions of other maintainers (cc @ratatui-org/core) but I would prefer if we have a doc/function comment explaining the general purpose/flow of the function rather than having distinct comments throughout the function (unless they are really necessary).

bblsh: I agree not all of the comments added are necessary and pieces like what you are highlighting here could absolutely be consolidated at a minimum. Leaving comments was helpful with trying to fully understand how lists are rendered (considering there was nothing prior to this) and I figured someone else may find it useful if changes need to be made down the line.

Agreed that a doc/function comment should be added regardless - would be happy to add that.

bblsh:
Thank you very much for the detailed explanations and suggestions.

In this push I've removed some of the comments and made some changes to variable names as suggested.
Specifically, renaming `offset`, `start` and `end` has made the code much more understandable without overly verbose explanations. A function doc comment was also added for `get_item_bounds()`.

I will address potential optimizations/readability with iterators and ranges in a future PR if that is okay with everyone.

orhun:
Sounds good.

There seems to be a conflict with `main`, can you fix it?

bblsh:
Done.

joshka (approved):
Can you write a small message that explains the change for the Changelog? (i.e. a one liner for the title of the change and a more full message for the body). If you're familiar with rebasing and squashing your changes, feel free to force push a single commit with the updated message, otherwise we can apply one for you using the github squash and merge.

orhun (approved):

src/widgets/list.rs
```diff
@@ -1598,6 +1642,48 @@ mod tests {
...
+        // in the list
+        let items = list_items(vec![
+            "Item 0", "Item 1", "Item 2", "Item 3", "Item 4", "Item 5",
+        ]);
+        *state.offset_mut() = 4;
+        let list = List::new(items);
+        let buffer = render_stateful_widget(list, &mut state, 10, 3);
+        let expected = Buffer::with_lines(vec!["Item 4    ", "Item 5    ", "          "]);
+        assert_buffer_eq!(buffer, expected);
+    }
```
Valentin271, line 1685: IMO, this does not clearly communicate the intent of your change and the fixed behavior. The underlying problem (that you fixed) is the offset not being taken into account while rendering.
I think testing the correct render of a list with a given offset is enough.

Valentin271 (changes requested):
Appart from the test, would be happy to merge that

src/widgets/list.rs
bblsh: I half-agree, with the reasoning being because rendering a List manipulates the list's state, taking the offset change into account. Each `assert_eq` here verifies each step of list manipulation with offset modifications either through the List's code or by user intervention. Since `List::get_item_bounds()` uses this offset change and determines a new offset, it seemed fair/correct to include each of these tests.

Previously Josh mentioned why one test was failing - this unit test covers that scenario with the last `assert_eq` and is expected behavior for setting an offset and rendering it, without having rendered the list before (since the List's render code changes the offset as mentioned earlier).

Happy to remove one or more of these tests if you still consider it unnecessary, but I still lean towards each of these tests being relevant. Perhaps moving one of those `assert_eq`s would make things clearer.

Valentin271: I agree that this test is useful, but I think the underlying thing you fixed is just the offset not being taken into account (selection or not). Previously setting `offset` explicitly would do nothing, I think this is the testcase we're missing (your testcase covers it but too broadly, it's not its main focus). Hope you get what I mean here.

Overall, I'd propose moving this to an integration test (it seems more like testing a usage rather than one particular feature) and testing that the `offset` setter works as expected (i.e. renders a shifted list).

Valentin271:
```rust
#[test]
fn test_offset_renders_shifted(
) {
    let items = list_items(vec![
        "Item 0", "Item 1", "Item 2", "Item 3", "Item 4", "Item 5", "Item 6",
    ]);
    let list = List::new(items);
    let mut state = ListState::default().with_offset(3);
    let buffer = render_stateful_widget(list, &mut state, 6, 3);

    let expected = Buffer::with_lines(vec!["Item 3", "Item 4", "Item 5"]);
}
```

Setting an offset didn't work previously and is now fixed with your changes. This test demonstrate that.

bblsh: I understand what you're saying, and agree this is an appropriate test.

Updated and pushed the test to reflect this. Thanks!

Valentin271 (approved):
Looks good. Do you want to move your initial test to an integration test?

orhun (approved):

Valentin271:
Hey @bblsh, do you still want to work on this. We couldn't merge it at the time because of unsigned commit and kind of forgot to tell you :sweat:.

If you still want to work on this, it will need to be rebased on main.

If you could sign your commits that's be great (we need all commits to be signed so the easiest is to squash them).
Otherwise if you don't want to setup commit signing, ask us to merge this manually.

Don't hesitate to ask for help if needed!

bblsh:
Rebased and signed! Thanks for the reminder.

Valentin271:
Thanks!

PR #668, opened 2023-12-05
feat(Paragraph): add `line_count` and `line_width` unstable helper methods

TylerBloom:
This PR adds two methods to the `Paragraph` widget to help get information about the rendered paragraph. The first method, `Paragraph::line_count`, takes a bounding width and returns the number of lines necessary to display the full paragraph when rendered. My motivation for this method is to use it with scrollbars. If a paragraph has wrapped text, it is currently impossible to get the number of lines it will need.

The second method is a natural extension of this, `Paragraph::line_width`. This returns the smallest line width needed to fully display every line of a paragraph. This is simply the maximum line widths. Together, these two methods can be used to calculate the minimum bounding box for a paragraph.

### DIFF
--- a/src/widgets/paragraph.rs
+++ b/src/widgets/paragraph.rs
@@ -213,6 +213,77 @@ impl<'a> Paragraph<'a> {
         self.alignment = alignment;
         self
     }
+
+    /// Calculates the number of lines needed to fully render.
+    ///
+    /// Given a max line width, this method calculates the number of lines that a paragraph will
+    /// need in order to be fully rendered. For paragraphs that do not use wrapping, this count is
+    /// simply the number of lines present in the paragraph.
+    ///
+    /// # Example
+    ///
+    /// ```ignore
+    /// # use ratatui::{prelude::*, widgets::*};
+    /// let paragraph = Paragraph::new("Hello World")
+    ///     .wrap(Wrap { trim: false });
+    /// assert_eq!(paragraph.line_count(20), 1);
+    /// assert_eq!(paragraph.line_count(10), 2);
+    /// ```
+    #[stability::unstable(
+        feature = "rendered-line-info",
+        reason = "The design for text wrapping is not stable and might affect this API.",
+        issue = "https://github.com/ratatui-org/ratatui/issues/293"
+    )]
+    pub fn line_count(&self, width: u16) -> usize {
+        if width < 1 {
+            return 0;
+        }
+
+        if let Some(Wrap { trim }) = self.wrap {
+            let styled = self.text.lines.iter().map(|line| {
+                let graphemes = line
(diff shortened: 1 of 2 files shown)


joshka:
Nice. I think this probably is a good interim approach to getting info about wrapping until we handle it properly in Line and Text.

src/widgets/paragraph.rs
```diff
@@ -208,6 +208,65 @@ impl<'a> Paragraph<'a> {
...
+    ///
+    /// let paragraph = Paragraph::new("Hello World\nhi\nHello World!!!");
+    /// assert_eq!(paragraph.line_width(), 14);
+    /// ```
+    pub fn line_width(&self) -> usize {
+        self.text
+            .lines
+            .iter()
+            .map(|l| l.width())
+            .max_by(|a, b| a.cmp(b))
```
joshka, line 267: nit: can this be `.max()`?

tests/widgets_paragraph.rs
```diff
@@ -364,3 +364,18 @@ fn widgets_paragraph_can_align_spans() {
         ]),
     );
 }
+
+#[test]
+fn widgets_paragraph_counts_rendered_lines() {
```
joshka, line 369: can you please add some tests for the lines that codecov annotates as not covered (width < 1, no wrap and line_width function)

joshka (changes requested):
LGTM generally. Needs a couple of extra tests and there's a couple of small nits.

I'm not sure whether we should put this behind an unstable flag (because of https://github.com/ratatui-org/ratatui/issues/293) or whether it should just be something that hangs around forever. What do the other @ratatui-org/core maintainers think about this?

Valentin271:
> put this behind an unstable flag (because of https://github.com/ratatui-org/ratatui/issues/293)

Yep definitely agree with that. We should sort out text wrapping pretty soon as I feel it's one of the most requested feature.

Otherwise we'd probably have to remove it at some point with a breaking change.

This unstable feature idea could also be useful to test designs and new APIs.

kdheepak:
I like the feature! It's a particularly useful building block for making a feature like "jump to $N line" or similar features. I haven't had a chance to review the code yet, but in favor of tagging it as unstable.

joshka:
Cool - Sounds like stability is the go - see `layout::SegmentSize` for an example of how to tag it.

joshka (approved):

Valentin271:
Ok so looks like you're just missing a rebase on main then @TylerBloom

Valentin271 (approved):
Looks good, thanks for your contribution!

orhun:
Too late to review but definitely a +1 from me. Thanks!

PR #669, opened 2023-12-06
docs(list): add documentation to the List widget

Valentin271:
Adds documentation to the `List` widget and all its sub components like `ListState` and `ListItem`

Relates to #386

# Screenshots

<details><summary>List</summary>
<p>

image

</p>
</details>

<details><summary>ListItem</summary>
<p>

image

</p>
</details> 

<details><summary>ListState</summary>
<p>

image

</p>
</details>

### DIFF
--- a/src/widgets/list.rs
+++ b/src/widgets/list.rs
@@ -1,3 +1,4 @@
+#![warn(missing_docs)]
 use unicode_width::UnicodeWidthStr;
 
 use crate::{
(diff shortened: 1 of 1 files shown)


src/widgets/list.rs
```diff
@@ -159,23 +515,76 @@ impl<'a> List<'a> {
...
         self.start_corner = corner;
         self
     }
 
+    /// Returns the number of [`ListItem`]s in the list
     pub fn len(&self) -> usize {
         self.items.len()
     }
 
+    /// Tells if the `List` is empty
```
orhun, line 587: We can use maybe something more explicit such as "Returns "true" if the list is empty" - it also aligns with the Rust stdlib docs

Valentin271: Good idea, I literally copied the `Vec` documentation

joshka (approved):

Issue #671, opened 2023-12-06
Clarify `List::start_corner`


Valentin271:
## Problem

What [`List::start_corner`](https://github.com/ratatui-org/ratatui/blob/dd22e721e3aed24538eb08e46e40339cec636bcb/src/widgets/list.rs#L170) does is not clear at first glance. Looking into it, the name doesn't correspond to the behavior.
This method actually sets the direction of the list.

## Solution

I suggest renaming the method and attribute, and maybe creating an corresponding enum.

This is a breaking change

## Additional context

I suspect this was done to re-use the `Corner` enum. In my opinion this doesn't make sense since the `Corner` enum designate a corner and not a direction.

The original commit for this is 
```
commit 5de571fb037fa02c2a67db1110c1838aea68000d
Author: Florian Dehau <work@fdehau.com>
Date:   Wed Sep 27 08:40:28 2017 +0200

    feat(widgets): Add start_corner option to List

diff --git a/src/widgets/list.rs b/src/widgets/list.rs
--- a/src/widgets/list.rs
+++ b/src/widgets/list.rs
@@ -68,0 +72,1 @@
+    pub fn start_corner(&'b mut self, corner: Corner) -> &mut List<'b, 'i, L, D> {
```

PR #672, opened 2023-12-07
feat(List)!: `new` now accepts `IntoIterator<Item = Into<ListItem>>`

### LINKED ISSUE #670: Make `List::new` accept `IntoIterator`
## Problem

Currently we have to write something like this to create a list.

```rust
List::new([ListItem::new("Item 1"), ListItem::new("Item 2"), ListItem::new("Item 3")]);
```

Would be nice and natural to be able to write

```rust
List::new(["Item 1", "Item 2", "Item 3"]);
```

## Solution

Make `List::new` accept `IntoIterator<Item = Into<ListItem>>`.

Unfortunately this seems to be breaking bug, especially for empty lists as the compiler cannot infer `::Item`'s type

## Alternatives

Leave as is

## Additional context

This should probably be worked on after #669 is merged.

Valentin271:
This allows to build list like

```rust
List::new(["Item 1", "Item 2"])
```

Fixes #670 

I think this should be added for 0.26 to not add too much breaking change for 0.25. Moreover there's another breaking change for list coming with #671. It makes sense to release them together.

### DIFF
--- a/src/widgets/list.rs
+++ b/src/widgets/list.rs
@@ -34,7 +34,7 @@ use crate::{
 /// # use ratatui::{prelude::*, widgets::*};
 /// # fn ui(frame: &mut Frame) {
 /// # let area = Rect::default();
-/// # let items = vec![];
+/// # let items = vec!["Item 1"];
 /// let list = List::new(items);
 ///
 /// // This should be stored outside of the function in your application state.
@@ -173,12 +173,22 @@ impl ListState {
 /// # Examples
 ///
 /// You can create [`ListItem`]s from simple `&str`
+///
 /// ```rust
 /// # use ratatui::{prelude::*, widgets::*};
 /// let item = ListItem::new("Item 1");
 /// ```
 ///
+/// Anything that can be converted to [`Text`] can be a [`ListItem`].
+///
+/// ```rust
+/// # use ratatui::{prelude::*, widgets::*};
+/// let item1: ListItem = "Item 1".into();
+/// let item2: ListItem = Line::raw("Item 2").into();
+/// ```
+///
 /// A [`ListItem`] styled with [`Stylize`]
+///
 /// ```rust
 /// # use ratatui::{prelude::*, widgets::*};
 /// let item = ListItem::new("Item 1").red().on_white();
@@ -186,6 +196,7 @@ impl ListState {
 ///
 /// If you need more control over the item's style, you can explicitly style the underlying
 /// [`Text`]
+///
 /// ```rust
 /// # use ratatui::{prelude::*, widgets::*};
 /// let mut text = Text::default();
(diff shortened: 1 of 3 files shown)


src/widgets/list.rs
```diff
@@ -178,6 +178,13 @@ impl ListState {
 /// let item = ListItem::new("Item 1");
 /// ```
 ///
+/// Anything that can be converted to [`Text`] can be a [`ListItem`].
```
joshka, line 181: Just as a stylistic thing, I prefer to leave a newline between text and code fences (there's a default markdown lint error that picks this up for actual markdown files). It's not a big deal, mainly a consistency thing.

joshka (changes requested):
I think this breaks things in a good way (i.e. on an empty list of items with indeterminate  type), so we can probably take it in the next major (0.25.0).

Needs a couple more things:
- some tests to show it works for a few of the assorted types.
- an entry in the breaking changes doc to explain how old code will fail (and corresponding changelog message in the commit

Let's also add an `items()` method so that `List::default().items(...)` works. What do you think?

src/widgets/list.rs
Valentin271: I don't mind adding a newline, I prefer consistency.
What markdown linter/formatter do you use? I hate formatting manually and checking every single line I write.

Valentin271:
Let's go for 0.25 then, it shouldn't really break any code anyway since there weren't an `items` method before. So creating an empty list was pointless.

> Let's also add an items() method so that List::default().items(...) works. What do you think?

Agree, so it lines up with table once #638 is merged

Valentin271:
Should be all good now. Hope the commit message is good enough, took inspiration from this one 37c70dbb8e19c0fb35ced16b29751933514a441e

src/widgets/list.rs
```diff
@@ -411,6 +457,28 @@ impl<'a> List<'a> {
...
+    ///
+    /// This is a fluent setter method which must be chained or used as it consumes self.
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// # use ratatui::{prelude::*, widgets::*};
+    /// let list = List::default().items(["Item 1", "Item 2"]);
+    /// ```
+    pub fn items<T>(mut self, items: T) -> Self
```
joshka, line 473: should this have a must_use attribute?

tests/widgets_list.rs
```diff
@@ -20,7 +20,7 @@ fn list_should_shows_the_length() {
     assert_eq!(list.len(), 3);
     assert!(!list.is_empty());
 
-    let empty_list = List::new(vec![]);
+    let empty_list = List::new::<Vec<&str>>(vec![]);
```
joshka, line 23: nit: The `::<Vec<&str>>` part looks a bit odd. Does something like this look better?
```suggestion
    let empty_list = List::new(iter::empty::<ListItem>());
```
or `vec::new::<ListItem>()` perhaps?

Valentin271: Or even better, we can now use the Default implementation to create an empty list. I should add documentation for that

Valentin271:
I also fixed some typos and missing links in the breaking-change file. Weird that `typos-cli` didn't catch them.

orhun (approved):

joshka (approved):

PR #673, opened 2023-12-07
refactor(List): `start_corner` is now `direction`

### LINKED ISSUE #671: Clarify `List::start_corner`

Valentin271:
The previous name `start_corner` did not communicate clearly what the function was doing. The method was renamed to `direction` and a new enum `ListDirection` was added.

Fixes #671 

Same as #672, I feel like this should be for 0.26.

### DIFF
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -11,6 +11,7 @@ github with a [breaking change] label.
 This is a quick summary of the sections below:
 
 - Unreleased (0.24.1)
+  - `List::start_corner` is renamed to `List::direction`
   - `List::new()` now accepts `IntoIterator<Item = Into<ListItem<'a>>>`
   - `Table::new()` now requires specifying the widths
   - `Table::widths()` now accepts `IntoIterator<Item = AsRef<Constraint>>`
@@ -41,6 +42,29 @@ This is a quick summary of the sections below:
 
 ## Unreleased (v0.24.1)
 
+### `List::start_corner` is renamed to `List::direction` ([#673])
+
+[#673]: https://github.com/ratatui-org/ratatui/pull/673
+
+Previously `List::start_corner` didn't communicate the intent of the method. It also used an
+inadequate `Corner` enum. The method is now renamed `direction` and a new enum
+`ListDirection` has been added.
+
+```diff
+- List::new(/* items */).start_corner(Corner::TopLeft);
+- List::new(/* items */).start_corner(Corner::TopRight);
+// This is not an error, BottomRight rendered top to bottom previously
+- List::new(/* items */).start_corner(Corner::BottomRight);
+// all becomes
++ List::new(/* items */).direction(ListDirection::TopToBottom);
+```
+
+```diff
+- List::new(/* items */).start_corner(Corner::BottomLeft);
+// becomes
++ List::new(/* items */).direction(ListDirection::BottomToTop);
+```
+
 ### `List::new()` now accepts `IntoIterator<Item = Into<ListItem<'a>>>` ([#672])
 
 [#672]: https://github.com/ratatui-org/ratatui/pull/672
@@ -50,10 +74,10 @@ error for `IntoIterator`s with an indeterminate item (e.g. empty vecs).
 
 E.g.
 
-```rust
-let list = List::new(vec![]);
+```diff
+- let list = List::new(vec![]);
 // becomes
-let list = List::default();
++ let list = List::default();
 ```
 
 ### The default `Tabs::highlight_style` is now `Style::new().reversed()` ([#635])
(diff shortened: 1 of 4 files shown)


src/widgets/list.rs
```diff
@@ -384,6 +385,20 @@ pub struct List<'a> {
     highlight_spacing: HighlightSpacing,
 }
 
+/// Defines the direction in which the list will be rendered.
+///
+/// If there is too few items to fill the screen, the list will stick to the starting edge.
```
a-kenji, line 390:
```suggestion
/// If there are too few items to fill the screen, the list will stick to the starting edge.
```

Valentin271: Thanks!

BREAKING-CHANGES.md
```diff
@@ -40,6 +41,29 @@ This is a quick summary of the sections below:
...
+List::new(/* items */).start_corner(Corner::BottomRight);
+// all becomes
+List::new(/* items */).direction(ListDirection::TopToBottom);
+```
+
+```rust
+List::new(/* items */).start_corner(Corner::BottomLeft);
+// becomes
+List::new(/* items */).direction(ListDirection::BottomToTop);
+```
```
orhun, line 65: Just an idea, we can probably use a diff here like below to highlight the changes:

```diff
- List::new(/* items */).start_corner(Corner::BottomLeft);
+ List::new(/* items */).direction(ListDirection::BottomToTop);
```

What do you think?

Valentin271: That's a good idea, I'll add it when that makes sense

joshka (changes requested):
To make this more of a user friendly change, instead of renaming the existing setter method and enum, keep both and mark them deprecated with a note on the replacement. Make the implementation of the old code call translate the corner to direction. This allows consuming libraries to be notified of changes as a warning in their build process rather than breaking immediately.

Valentin271:
Rebased on #672

Valentin271:
Right I completely forgot about the depreciation warning. I added back `start_corner` and its tests with the warning.

I'm wondering if this is still a breaking change then?

orhun:
> I'm wondering if this is still a breaking change then?

If we keep the old behavior as deprecated, I don't think so.

BREAKING-CHANGES.md
joshka, line 1: Perhaps stash this file change and include it in the second part of the PR when we remove start_corner.

Alternatively, perhaps it might be better to keep it and add a line at the beginning that the file will also deprecations as well as breaking changes. What do you think?

src/widgets/list.rs
```diff
@@ -458,7 +473,7 @@ impl<'a> List<'a> {
             block: None,
             style: Style::default(),
             items: items.into_iter().map(|i| i.into()).collect(),
-            start_corner: Corner::TopLeft,
+            direction: ListDirection::default(),
             highlight_style: Style::default(),
             highlight_symbol: None,
             repeat_highlight_symbol: false,
```
joshka, line 479: This can probably be simplified to:
```
Self {
  items: items.into_iter().map(|i| i.into()).collect(),
  .. Self::default()
}
```

joshka (changes requested):
2 Small things. A small simplification to new() and another about how we want to communicate breaking changes.

Otherwise this is great.

BREAKING-CHANGES.md
Valentin271: I think we should inform users of changes and help them migrate as soon as possible.

So either keep the file and mix breaking and deprecation (which would duplicate a lot in the end). Or maybe just document how to migrate in the commit message?

We could also have a separate DEPRECATION file (essentially its content would be copied over to BREAKING CHANGE when the API is removed).

I'd like to here @ratatui-org/core's opinion on this.

joshka: That's reasonable. :)

orhun: Maintaining 3 separate files for changes (changelog, deprecation, breaking changes) is too much in my opinion. We can simply mention the deprecation notice in the commit body and it will appear in the changelog. Considering that we are already using `deprecated` annotations, that should be enough for users to do a gradual switch to the new API.

joshka: I should have been more specific about what i agreed with there.
I agree with @orhun here, no need for an extra file for deprecation.
My point was deprecation is not a breaking change, it's an intent to make a breaking change in the future. It's no biggy though

PR #677, opened 2023-12-09
fix(Rect): Fixed underflow in the `Rect.intersection` method

bugnano:
The `Rect.intersection` method contained the simple `-` operator, which panics on underflow.
This is strange, as all the other methods in `Rect` use `saturating_add` and `saturating_sub`.
So I changed the `-` operation to a `saturating_sub`, so that `Rect.intersection` never panics.

### DIFF
--- a/src/layout/rect.rs
+++ b/src/layout/rect.rs
@@ -155,8 +155,8 @@ impl Rect {
         Rect {
             x: x1,
             y: y1,
-            width: x2 - x1,
-            height: y2 - y1,
+            width: x2.saturating_sub(x1),
+            height: y2.saturating_sub(y1),
         }
     }
 


Valentin271 (changes requested):
Looks like a sane change. Could you add a test validating your changes.

Also your commits need to be signed (see [CONTRIBUTING.md](https://github.com/ratatui-org/ratatui/blob/main/CONTRIBUTING.md#sign-your-commits) for help)

bugnano:
OK, so I'm not doing all the GPG signing stuff just for this simple commit.
Can somebody with commit rights to this repo make a change similar to mine, and close this PR as fixed in a different commit, please?

Valentin271:
Done in #678, thanks for reporting that!

PR #678, opened 2023-12-09
fix(Rect): fixes underflow in the `Rect::intersection` method

Valentin271:
From @bugnano in #677

> The `Rect.intersection` method contained the simple `-` operator, which panics on underflow.
> This is strange, as all the other methods in `Rect` use `saturating_add` and `saturating_sub`.
> So I changed the `-` operation to a `saturating_sub`, so that `Rect.intersection` never panics.

With a signed commit and a unit test.

### DIFF
--- a/src/layout/rect.rs
+++ b/src/layout/rect.rs
@@ -155,8 +155,8 @@ impl Rect {
         Rect {
             x: x1,
             y: y1,
-            width: x2 - x1,
-            height: y2 - y1,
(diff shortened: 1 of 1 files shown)


orhun (approved):

orhun:
Not sure if we need more unit tests for this, let's wait for review from other maintainers.

joshka (approved):
Seems like enough tests to me

PR #685, opened 2023-12-12
fix: Render table with overflow

### LINKED ISSUE #684: Debug assertion on too-wide, empty, right-aligned table cell
## Description

When trying to render a particular table which is wider than the available area, ratatui panics on a debug assertion. This is similar to #474, but with a somewhat different backtrace.

The following seems to need to be true for the repro below to work:

* Terminal should be narrower than the table's width constraint (e.g., 140 in the example)
* There should be an empty, right-aligned line in the table
* Also, using `Line::raw` in place of `<Line as From<&str>>::from` doesn't work; I suspect that's because Line::raw uses str::lines to normalize its content

```
thread 'main' panicked at 'Trying to access position outside the buffer: x=138, y=0, area=Rect { x: 0, y: 0, width: 138, height: 45 }', /home/wbain/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ratatui-0.24.0/src/buffer.rs:240:9
stack backtrace:
   0: rust_begin_unwind
             at /rustc/eb26296b556cef10fb713a38f3d16b9886080f26/library/std/src/panicking.rs:593:5
   1: core::panicking::panic_fmt
             at /rustc/eb26296b556cef10fb713a38f3d16b9886080f26/library/core/src/panicking.rs:67:14
   2: ratatui::buffer::Buffer::index_of
             at /home/wbain/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ratatui-0.24.0/src/buffer.rs:240:9
   3: ratatui::buffer::Buffer::set_stringn
             at /home/wbain/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ratatui-0.24.0/src/buffer.rs:309:25
   4: ratatui::buffer::Buffer::set_line
             at /home/wbain/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ratatui-0.24.0/src/buffer.rs:343:23
   5: ratatui::widgets::table::render_cell
             at /home/wbain/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ratatui-0.24.0/src/widgets/table.rs:640:9
   6: <ratatui::widgets::table::Table as ratatui::widgets::StatefulWidget>::render
             at /home/wbain/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ratatui-0.24.0/src/widgets/table.rs:558:17
   7: ratatui::terminal::Frame::render_stateful_widget
             at /home/wbain/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ratatui-0.24.0/src/terminal.rs:632:9
   8: bad_table::main::{{closure}}
             at ./examples/bad_table.rs:20:13
   9: ratatui::terminal::Terminal<B>::draw
             at /home/wbain/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ratatui-0.24.0/src/terminal.rs:334:9
  10: bad_table::main
             at ./examples/bad_table.rs:19:9
  11: core::ops::function::FnOnce::call_once
             at /rustc/eb26296b556cef10fb713a38f3d16b9886080f26/library/core/src/ops/function.rs:250:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

## To Reproduce

```rust
use std::io;

use crossterm::{terminal, ExecutableCommand};
use ratatui::{
    backend::CrosstermBackend,
    layout::{Alignment, Constraint},
    text::Line,
    widgets::{Row, Table},
    Terminal,
};

fn main() -> io::Result<()> {
    io::stdout().execute(terminal::EnterAlternateScreen)?;
    terminal::enable_raw_mode()?;
    let mut terminal = Terminal::new(CrosstermBackend::new(io::stdout()))?;
    terminal.clear()?;

    loop {
        terminal.draw(|frame| {
            frame.render_stateful_widget(
                Table::new(vec![])
                    .header(Row::new([Line::from("").alignment(Alignment::Right)]))
                    .widths(&[Constraint::Min(140); 1]),
                frame.size(),
                &mut Default::default(),
            );
        })?;

        // Check for user input every 250 milliseconds
        if crossterm::event::poll(std::time::Duration::from_millis(250))? {
            // If a key event occurs, handle it
            if let crossterm::event::Event::Key(key) = crossterm::event::read()? {
                if key.kind == crossterm::event::KeyEventKind::Press {
                    match key.code {
                        crossterm::event::KeyCode::Char('q') => break,
                        _ => {}
                    }
                }
            }
        }
    }

    io::stdout().execute(terminal::LeaveAlternateScreen)?;
    terminal::disable_raw_mode()?;

    Ok(())
}
```

## Expected behavior

Trying to render a table which doesn't fit should resolve in some form of truncation rather than a panic.

## Environment

- OS: Linux
- Terminal Emulator: vscode (TERM=xterm-256color)
- Font: N/A
- Crate version: 0.24.0
- Backend: crossterm

## Additional context

In [ratatui::buffer::Buffer::set_stringn](https://github.com/ratatui-org/ratatui/blob/d19b266e0eabdb0fb00660439a1818239c94024b/src/buffer.rs#L318) the first call is to `Buffer::index_of`, which asserts. It looks like `Buffer::set_stringn` is supposed to handle truncating strings which are too long—if it weren't for the debug assert I believe that the call would hit the bound check [in the loop](https://github.com/ratatui-org/ratatui/blob/d19b266e0eabdb0fb00660439a1818239c94024b/src/buffer.rs#L329) and avoid actually writing anything. So it seems like a fix here would be to defer the `index_of` call until after a bounds check.

YeungKC:
This PR aims to fix an issue in the `render_cell` function that could lead to an index out of range error. The problem was caused by incorrect positioning of the `x` coordinate in the original code.

Fixes #684

### DIFF
--- a/src/widgets/table.rs
+++ b/src/widgets/table.rs
@@ -706,7 +706,12 @@ fn render_cell(buf: &mut Buffer, cell: &Cell, area: Rect) {
             _ => 0,
         };
 
-        buf.set_line(area.x + x_offset, area.y + i as u16, line, area.width);
+        let x = area.x + x_offset;
+        if x >= area.right() {
+            continue;
+        }
+
+        buf.set_line(x, area.y + i as u16, line, area.width);
     }
 }
 
@@ -1033,6 +1038,15 @@ mod tests {
         ]);
 
         assert_eq!(buf, expected);
+    }
+
+    #[test]
+    fn test_render_table_when_overflow() {
+        let mut buf = Buffer::empty(Rect::new(0, 0, 20, 3));
+        let table = Table::new(vec![], [Constraint::Min(20); 1])
+            .header(Row::new([Line::from("").alignment(Alignment::Right)]));
+
+        Widget::render(table, Rect::new(0, 0, 20, 3), &mut buf);
     }
 
     #[test]


orhun (approved):
Can confirm that the panic is fixed with this. Not sure if we need more tests / if there is any edge cases though. General approval +

YeungKC:
I'm not very know this lib, but I think, normally, only an empty string would trigger that issue.

Valentin271 (approved):
I think that's enough test. I agree with @YeungKC, only an empty string, right aligned and with no border can trigger this.

Issue #679, opened 2023-12-09
Implement `From` for `Style`


Valentin271:
## Problem

There is no way to create a `Style` from crossterm, termion or termwiz style.

This is impossible to implement on the user's side because of the orphan rule. Or super cumbersome with a new-type pattern (not to mention underlying `Color` and attributes have the exact same problem).

It is also impossible to use raw ansi code when creating `Text`, `Line` or `Span` as they will render 
![image](https://github.com/ratatui-org/ratatui/assets/36198422/6e40ed52-dd9c-44e6-ab4a-234d5959e7c9)
Code reference:
![image](https://github.com/ratatui-org/ratatui/assets/36198422/7ffb71ac-3f26-4cfb-aff3-6c0a0cdbd995)

## Solution

Add a `From` impl for `Style`, `Color` and `Modifier`.
These should be under their corresponding terminal feature.

~~We could also authorize raw ansi codes when creating `Text`, `Line`, `Span` and others (if this is possible, don't know the underlying limitations here).~~

EDIT for future readers: [ansi-to-tui](https://crates.io/crates/ansi-to-tui) does exactly that.

## Additional context

I wanted to use https://github.com/sharkdp/lscolors to retrieve `LS_COLORS` and print a list of files.

This could somewhat relate to #493. 

# Tasks

- [x] `crossterm`
- [x] `termion`
- [ ] `termwiz`

joshka:
Definitely a good idea

mobley-trent:
Hello @joshka Is this issue still open for `termwiz` ?

joshka:
Yes. And termion.

Valentin271:
Termion was already done in #692.

Do you want to work on termwiz @mobley-trent?

---

[added a checklist to the original comment]

joshka:
> Termion was already done in https://github.com/ratatui-org/ratatui/pull/692.

Apologies - I forgot.

Valentin271:
Sorry, you didn't reply so I thought you were only checking on :sweat_smile: . I was already working on this at this point ...

If you want to help there is still some interesting work left in the issues list.

PR #686, opened 2023-12-12
feat(style): implement `From` trait for crossterm to `Style` related structs

Valentin271:
This PR includes the following `From` implementations:

- from `Attribute` to `Modifier`
- from `Attributes` to `Modifier`
- from crossterm `Color` to `Color`
- from `ContentStyle` to `Style`

This PR relates to #679. I've decided to split the work across multiple PRs to facilitate the reviewing process.

### DIFF
--- a/src/backend/crossterm.rs
+++ b/src/backend/crossterm.rs
@@ -10,8 +10,8 @@ use crossterm::{
     cursor::{Hide, MoveTo, Show},
     execute, queue,
     style::{
-        Attribute as CAttribute, Color as CColor, Print, SetAttribute, SetBackgroundColor,
-        SetForegroundColor,
+        Attribute as CAttribute, Attributes as CAttributes, Color as CColor, ContentStyle, Print,
+        SetAttribute, SetBackgroundColor, SetForegroundColor,
     },
     terminal::{self, Clear},
 };
@@ -21,7 +21,7 @@ use crate::{
     buffer::Cell,
     layout::Size,
     prelude::Rect,
-    style::{Color, Modifier},
+    style::{Color, Modifier, Style},
 };
 
 /// A [`Backend`] implementation that uses [Crossterm] to render to the terminal.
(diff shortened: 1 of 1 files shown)


src/style.rs
```diff
@@ -110,6 +110,60 @@ impl fmt::Debug for Modifier {
     }
 }
 
+#[cfg(feature = "crossterm")]
```
joshka, line 113: These might belong in crossterm.rs to avoid the feature checks and having an impl for each backend in style.rs

src/style.rs
```diff
@@ -110,6 +110,60 @@ impl fmt::Debug for Modifier {
...
+            res |= Modifier::DIM;
+        }
+        if value.has(CAttribute::Italic) {
+            res |= Modifier::ITALIC;
+        }
+        if value.has(CAttribute::Underlined)
+            || value.has(CAttribute::DoubleUnderlined)
+            || value.has(CAttribute::Undercurled)
+            || value.has(CAttribute::Underdotted)
+            || value.has(CAttribute::Underdashed)
```
joshka, line 143: Aside: If these really work, do you think we should add them to Ratatui at some point?

src/style.rs
```diff
@@ -631,4 +826,67 @@ mod tests {
...
+            ContentStyle {
+                attributes: CAttributes::from([CAttribute::NoBold, CAttribute::NoItalic].as_ref()),
+                ..Default::default()
+            },
+        ];
+        let styles = styles();
+
+        for (content_style, style) in content_styles.into_iter().zip(styles) {
+            assert_eq!(Style::from(content_style), style);
+        }
```
joshka, line 876: When reading it I had to read down, then up to work out what fn styles was doing and then for each style had to read up and down to keep them in sync. This is a test smell. I'd prefer to replace this approach with many asserts like:
```rust
assert_eq!(ContentStyle::new().dark_yellow().into(), Style::new().yellow());
```

joshka (changes requested):

src/style.rs
Valentin271: Makes much more sense now that I'm writing it, to have it in crossterm with the other From impl

src/style.rs
Valentin271: You're right I also had to do that while writing the test. Fixed when moved to crossterm

src/style.rs
Valentin271: I think we either implement everything (see [crossterm](https://docs.rs/crossterm/latest/crossterm/style/enum.Attribute.html) and [ANSI wiki](https://en.wikipedia.org/wiki/ANSI_escape_code#SGR_(Select_Graphic_Rendition)_parameters)) or stay as is.

It seems those are rarely supported features. For example my `xfce4-terminal` supports underlined and double underlined bot not fraktur or framed. And tbh `Underdotted`, `Undercurled` and `Underdashed` doesn't even show up on the wiki.

tldr; overall no one has requested that, let's keep it as is and implement it later if really needed

src/backend/crossterm.rs
```diff
@@ -274,6 +274,32 @@ impl From<Color> for CColor {
     }
 }
 
+impl From<CColor> for Color {
+    fn from(value: CColor) -> Self {
+        match value {
+            CColor::Reset => Self::Reset,
```
joshka, line 280: nit: is it worth making these in the same order as above (it's that way because it aligns with the numeric index of the colors)?

src/backend/crossterm.rs
```diff
@@ -344,3 +370,300 @@ impl ModifierDiff {
         Ok(())
     }
 }
+
+impl From<CAttribute> for Modifier {
+    fn from(value: CAttribute) -> Self {
+        Modifier::from(CAttributes::from(value))
```
joshka, line 376: Can you add a doc comment for this line - it's a bit weird / requires understanding of crossterm Attribute vs Attributes

joshka (approved):
Approve - with nits

src/backend/crossterm.rs
Valentin271: Sure, I put them in the order of the enum originally

src/backend/crossterm.rs
Valentin271: Sure, plus the difference is subtle

orhun (approved):
I likey

PR #692, opened 2023-12-15
feat: implement `From` trait for termion to `Style` related structs

Valentin271:
Relates to #679

### DIFF
--- a/src/backend/termion.rs
+++ b/src/backend/termion.rs
@@ -379,6 +379,30 @@ impl fmt::Display for ModifierDiff {
     }
 }
 
+macro_rules! from_termion_for_modifier {
+    ($termion_modifier:ident, $modifier: ident) => {
+        impl From<termion::style::$termion_modifier> for Modifier {
+            fn from(_: termion::style::$termion_modifier) -> Self {
+                Modifier::$modifier
+            }
+        }
+    };
+}
+
+from_termion_for_modifier!(Invert, REVERSED);
+from_termion_for_modifier!(Bold, BOLD);
+from_termion_for_modifier!(Italic, ITALIC);
+from_termion_for_modifier!(Underline, UNDERLINED);
+from_termion_for_modifier!(Faint, DIM);
+from_termion_for_modifier!(CrossedOut, CROSSED_OUT);
+from_termion_for_modifier!(Blink, SLOW_BLINK);
+
+impl From<termion::style::Reset> for Modifier {
+    fn from(_: termion::style::Reset) -> Self {
+        Modifier::empty()
+    }
+}
+
 #[cfg(test)]
 mod tests {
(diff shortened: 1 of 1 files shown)


src/backend/termion.rs
```diff
@@ -338,3 +416,231 @@ impl fmt::Display for ModifierDiff {
...
+        use tcolor::Fg;
+
+        assert_eq!(
+            Style::from(Fg(tcolor::Reset)),
+            Style::default().fg(Color::Reset)
+        );
+        assert_eq!(
+            Style::from(Fg(tcolor::Black)),
+            Style::default().fg(Color::Black)
+        );
```
orhun, line 564: We can potentially have a loop with a tuple values for reducing the code repetition but I guess then it is harder to track which test has failed from the output. Either way, I just wanted to ask your opinion on this.

orhun (commented):
Generally LGTM, just had one comment.

src/backend/termion.rs
Valentin271: I thought of doing that. 

> harder to track which test has failed

True, but for ~100 lines of basically enum testing I would have done it.

Concerning the loop, termion colors are not an enum, each color is a different type and they all implement the `Color` trait.
So I could have tuples of `(dyn Color, _)` or similar. But the `From` trait is implemented on each color struct, not on the `Color` trait.

That's the reason for the macro, will be the same for termwiz and I must say this is annoying.

orhun: > So I could have tuples of (dyn Color, _) or similar. But the From trait is implemented on each color struct, not on the Color trait.

So you have to use dyn Color then you need to convert it back to a struct, yeah, this is annoying.

> Alternative is to test just a few combinations 

Or we can just keep it like this 😅

Valentin271: > convert it back to a struct, yeah, this is annoying

I don't think I could even do it in a loop, because they're all different struct, and downcast take a template parameter ...

Anyway I agree on this :smile:

> Or we can just keep it like this

orhun (approved):

src/backend/termion.rs
joshka: Perhaps consider whether each assert can be a single line by using shorter import aliases (e.g TC::) and stylize shortcuts (e.g. Style::new().on_red())

Valentin271: Good idea! It worked, made the tests a lot cleaner and easier to read

joshka: Or perhaps use .into() to avoid Style::From()

Valentin271: This won't work, rustc is unable to infer the target type and requires an annotation.

joshka (approved):

PR #695, opened 2023-12-16
chore(rustfmt): enable format_code_in_doc_comments

Valentin271:
This enables more consistently formatted code in doc comments, especially since ratatui heavily uses fluent setters.

See https://rust-lang.github.io/rustfmt/?version=v1.6.0#format_code_in_doc_comments

We're already using rustfmt nightly so this shouldn't change the pipelines or toolchains.

See the diff for the benefits this brings.

What do you guys think about that?


orhun (approved):
I think this is a reasonable change, thanks!

PR #696, opened 2023-12-16
docs(chart): document chart module

### LINKED ISSUE #386: Docs: improve general methods documentation

Valentin271:
This fixes #386 and completes the documentation for the widgets module apart from canvas.

<details><summary>doc screenshots</summary>
<p>

image
image
image
image
image

</p>
</details>

### DIFF
--- a/src/widgets/chart.rs
+++ b/src/widgets/chart.rs
@@ -1,3 +1,4 @@
+#![warn(missing_docs)]
 use std::{borrow::Cow, cmp::max};
 
 use strum::{Display, EnumString};
@@ -8,18 +9,37 @@ use crate::{
     layout::{Alignment, Constraint, Rect},
     style::{Color, Style, Styled},
     symbols,
-    text::{Line as TextLine, Span},
+    text::{Line, Span},
     widgets::{
-        canvas::{Canvas, Line, Points},
+        canvas::{Canvas, Line as CanvasLine, Points},
         Block, Borders, Widget,
     },
 };
 
-/// An X or Y axis for the chart widget
+/// An X or Y axis for the [`Chart`] widget
+///
+/// An axis can have a [title](Axis::title) which will be displayed at the end of the axis. For an
+/// X axis this is the right, for a Y axis, this is the top.
+///
+/// You can also set the bounds and labels on this axis using respectively [`Axis::bounds`] and
+/// [`Axis::labels`].
+///
+/// See [`Chart::x_axis`] and [`Chart::y_axis`] to set an axis on a chart.
+///
+/// # Example
+///
+/// ```rust
+/// use ratatui::{prelude::*, widgets::*};
+/// let axis = Axis::default()
+///     .title("X Axis")
+///     .style(Style::default().gray())
+///     .bounds([0.0, 50.0])
+///     .labels(vec!["0".bold(), "25".into(), "50".bold()]);
+/// ```
 #[derive(Debug, Default, Clone, PartialEq)]
 pub struct Axis<'a> {
     /// Title displayed next to axis end
-    title: Option<TextLine<'a>>,
+    title: Option<Line<'a>>,
     /// Bounds for the axis (all data points outside these limits will not be represented)
     bounds: [f64; 2],
     /// A list of labels to put to the left or below the axis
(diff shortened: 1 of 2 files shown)


src/widgets/chart.rs
```diff
@@ -8,18 +9,37 @@ use crate::{
     layout::{Alignment, Constraint, Rect},
     style::{Color, Style, Styled},
     symbols,
-    text::{Line as TextLine, Span},
+    text::{Line, Span},
     widgets::{
-        canvas::{Canvas, Line, Points},
+        canvas::{Canvas, Line as CanvasLine, Points},
```
Valentin271, line 14: Renamed the canvas line instead of the text line because otherwise that would show up on the doc as `TextLine`

src/widgets/chart.rs
```diff
@@ -71,23 +141,36 @@ impl<'a> Axis<'a> {
...
 #[derive(Debug, Default, Display, EnumString, Clone, Copy, Eq, PartialEq, Hash)]
 pub enum GraphType {
-    /// Draw each point
+    /// Draw each point. This is the default.
     #[default]
     Scatter,
-    /// Draw each point and lines between each point using the same marker
+    /// Draw a line between each following point.
+    ///
+    /// This means your dataset can go back and this will draw a line backward.
```
Valentin271, line 149: I'm not sure this is clear enough, I want to communicate the fact that you can do this sort of thing
image

joshka: "The order of the lines will be the same as the order of the points in the dataset, which allows this widget to draw lines both left-to-right and right-to-left."

src/widgets/chart.rs
joshka: I've been generally using the approach of using `text::Line` instead of the rename - it's close enough to the same amount of code. No big deal though.

src/widgets/chart.rs
```diff
@@ -31,36 +51,86 @@ pub struct Axis<'a> {
...
     pub fn title<T>(mut self, title: T) -> Axis<'a>
     where
-        T: Into<TextLine<'a>>,
+        T: Into<Line<'a>>,
     {
         self.title = Some(title.into());
         self
     }
 
+    /// Sets the bound of this axis
```
joshka, line 69:
```suggestion
    /// Sets the bounds of this axis
```

src/widgets/chart.rs
Valentin271: That's way better, I'll update

src/widgets/chart.rs
```diff
@@ -186,6 +269,34 @@ impl LegendPosition {
 }
 
 /// A group of data points
+///
+/// This is the main element composing a [`Chart`].
+///
+/// A dataset can be [named](Dataset::name) to be referenced in the legend (NOTE: Currently,
+/// datasets with an empty name will show an empty line in the legend, see [PR 527]).
+///
+/// After that, you can pass it data with [`Dataset::data`]. Data consist in an array of `f64`
```
joshka, line 278: Grammer nit:
```suggestion
/// After that, you can pass it data with [`Dataset::data`]. Data is an array of `f64`
```

src/widgets/chart.rs
```diff
@@ -326,6 +498,16 @@ pub struct Chart<'a> {
 }
 
 impl<'a> Chart<'a> {
+    /// Creates a chart with the given [datasets](Dataset)
+    ///
+    /// # Example
+    ///
+    /// Creates an empty chart. See the [struct level doc](Chart) for more details.
```
joshka, line 505: Is this empty? What does the dataset passed here do?

joshka (approved):
Just nits

src/widgets/chart.rs
Valentin271: It was yes. I thought it was better because filling the points is done and documented on the dataset. Reading this again I feel like I want to know quickly how to create a chart, so I just added the `data` setter.

I also forgot to mention that you can pass multiple datasets, so I added that just now.

What do you think?

src/widgets/chart.rs
Valentin271: That's a good alternative, I'll remember that. Let's keep it that way since its already done?

src/widgets/chart.rs
joshka: LGTM

joshka (approved):

PR #698, opened 2023-12-17
docs: remove deprecated table constructor from breaking changes

orhun:
We have 2 breaking changes that effects `Table` so I think we should recommend using the new constructor instead of the old one.


Valentin271 (approved):
Yes definitely!

PR #703, opened 2023-12-21
docs(examples): refactor chart example to showcase scatter

Valentin271:
I reworked the chart example to showcase the scatter chart which was missing before. While at it, I also used some real data.

| Before | After |
|--------|--------|
| image | image |

### DIFF
--- a/examples/chart.rs
+++ b/examples/chart.rs
@@ -9,18 +9,10 @@ use crossterm::{
     execute,
     terminal::{disable_raw_mode, enable_raw_mode, EnterAlternateScreen, LeaveAlternateScreen},
 };
-use ratatui::{prelude::*, widgets::*};
-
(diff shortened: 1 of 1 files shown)


orhun (approved):

examples/chart.rs
```diff
@@ -194,64 +200,164 @@ fn ui(f: &mut Frame, app: &App) {
...
         )
-        .legend_position(Some(LegendPosition::TopLeft))
         .hidden_legend_constraints((Constraint::Ratio(1, 2), Constraint::Ratio(1, 2)));
-    f.render_widget(chart, chunks[2]);
+
+    f.render_widget(chart, area);
 }
+
+// Data from https://ourworldindata.org/space-exploration-satellites
+const HEAVY_PAYLOAD_DATA: [(f64, f64); 9] = [
```
orhun, line 295: Not sure if it makes sense to move constant declarations to the top.

Valentin271: Usually they're at the top but here they're 100 lines long. I felt it was a lot of scrolling and its not that important.

joshka (approved):

PR #705, opened 2023-12-23
feat(color): use the FromStr implementation for deserialization

### LINKED ISSUE #704: Suggestion: When using serde to deserialize a Color, it would be great to use the FromStr implementation
## Problem
I have a config file that allows configuring colors via strings.
If I use the "serde" feature of the crate, the only accepted color names are the string repr of the Enum variants.

Given that Color implements beautifully the FromStr trait, it would be awesome to be able to use any of the strings supported by FromStr in the config file.

## Solution

Something like:

```rust
impl<'de> Deserialize<'de> for Color {
    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
        where D: Deserializer<'de>
    {
        let s = String::deserialize(deserializer)?;
        FromStr::from_str(&s).map_err(de::Error::custom)
    }
}

```

## Alternatives
An alternative that I considered is to use the serde_with crate and its DisplayFromStr annotation.
However, annotating every single color is tedious.

orhun:
The deserialize implementation for Color used to support only the enum
names (e.g. Color, LightRed, etc.) With this change, you can use any of
the strings supported by the FromStr implementation (e.g. black,
light-red, #00ff00, etc.)

closes #704

### DIFF
--- a/src/style/color.rs
+++ b/src/style/color.rs
@@ -61,7 +61,7 @@ use std::{
 ///
 /// [ANSI color table]: https://en.wikipedia.org/wiki/ANSI_escape_code#Colors
 #[derive(Debug, Default, Clone, Copy, Eq, PartialEq, Hash)]
-#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serde", derive(serde::Serialize))]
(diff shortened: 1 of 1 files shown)


Valentin271 (approved):
Definitely a good idea, approved.

PR #706, opened 2023-12-23
add unicode support for user_input example

alexniver:
user_input

### DIFF
--- a/examples/user_input.rs
+++ b/examples/user_input.rs
@@ -61,7 +61,9 @@ impl App {
     }
 
     fn enter_char(&mut self, new_char: char) {
-        self.input.insert(self.cursor_position, new_char);
+        let mut chars = self.input.chars().collect::<Vec<char>>();
+        chars.insert(self.cursor_position, new_char);
+        self.input = chars.into_iter().collect();
 
         self.move_cursor_right();
     }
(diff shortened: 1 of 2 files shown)


examples/user_input.rs
```diff
@@ -61,35 +61,26 @@ impl App {
...
 
     fn delete_char(&mut self) {
         let is_not_cursor_leftmost = self.cursor_position != 0;
         if is_not_cursor_leftmost {
-            // Method "remove" is not used on the saved text for deleting the selected char.
-            // Reason: Using remove on String works on bytes instead of the chars.
-            // Using remove would require special care because of char boundaries.
+            let mut chars = self.input.chars().collect::<Vec<char>>();
+            chars.remove(self.cursor_position - 1);
+            self.input = chars.into_iter().collect();
```
kdheepak, line 76: I don't quite understand why the previous code doesn't work as expected? It is using `.chars()` too, right?

Also, if you try this example on this emoji:

```
🤦🏼‍♂️
```

I believe you'll have to hit backspace 5 times.

```rust
// main.rs

fn main() {

    let s = "🤦🏼‍♂️";

    dbg!(s.chars().count());

    for (i, c) in s.chars().enumerate() {
        println!("s.chars()[{}] = {:?}", i, c);
    }

}
```

```
[main.rs:19] s.chars().count() = 5
s.chars()[0] = '🤦'
s.chars()[1] = '🏼'
s.chars()[2] = '\u{200d}'
s.chars()[3] = '♂'
s.chars()[4] = '\u{fe0f}'
```

I think we should just document that this example isn't going to handle any kind of user input, because it's not trivial to write an example that covers all the corner cases.

alexniver: in my test, for emoji
`🤦`
just need hit backspace 1 time.

Peek 2023-12-24 13-00


use this tape
```
Output "target/user_input.gif"
# Set Theme "Aardvark Blue"
Set Width 1200
Set Height 600
Hide
Type "cargo run --example=user_input --features=crossterm"
Enter
Sleep 1s
Show
Sleep 2s
Type e
Sleep 1s
Type "Hello, world!你好，世界！🤦"
Enter
Sleep 2s
Backspace 13
Sleep 1s
Type "Goodbye, world!再见，世界！"
Enter
Sleep 5s
Type "你好，世界！🤦"
Backspace 7
```
result is 
user_input

kdheepak: The emoji you pasted is in your comment is 🤦 but the emoji I pasted in my comment is 🤦🏼‍♂️. Do you see the difference? The second emoji has modifiers for skin tone and gender.

alexniver: oh, my bad.

PR #709, opened 2023-12-26
feat(span): implement Widget on Span

joshka:
This allows us to use Span as a child of other widgets, and to use
Span::render() to render it rather than calling buffer.set_span().

```rust
frame.render_widget(Span::raw("Hello, world!"), area);
// or
Span::raw("Hello, world!").render(frame, area);
// or even
"Hello, world!".green().render(frame, area);
```

### DIFF
--- a/src/text/span.rs
+++ b/src/text/span.rs
@@ -4,7 +4,7 @@ use unicode_segmentation::UnicodeSegmentation;
 use unicode_width::UnicodeWidthStr;
 
 use super::StyledGrapheme;
-use crate::style::{Style, Styled};
+use crate::{prelude::*, widgets::Widget};
 
 /// Represents a part of a line that is contiguous and where all characters share the same style.
 ///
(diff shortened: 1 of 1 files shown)


joshka:
`Line` will get the same next (once https://github.com/ratatui-org/ratatui/pull/708 is merged)

The end goal of this is to fully remove the `Buffer::set_line` and `Buffer::set_span` methods, and implement various alignment type things in the line itself instead of in other widgets (like block title, various chart labels etc.).

Valentin271 (approved):
Seems like a good idea, approved

orhun (approved):

kdheepak:
Should we also implement `Widget` on `&str`? i.e. `"Hello, world!".render(frame, area);`

joshka:
That might be a little too magical for my taste

kdheepak:
If I have code that is like `"Hello, world!".green().render(frame, area);` and I want to remove the color, how would you recommend I do it?

joshka:
I see your point there. `.into::<Span>()` might work?

PR #710, opened 2023-12-26
fix(table): render missing widths as equal

### LINKED ISSUE #510: Make it more difficult to render a table without `widths` set

joshka:
Previously, if `.widths` was not called before rendering a `Table`, no
content would render in the area of the table. This commit changes that
behaviour to default to equal widths for each column.

Fixes #510.

(Resubmission of https://github.com/ratatui-org/ratatui/pull/519 by @joshcbrown due to git footgun)

### DIFF
--- a/src/widgets/table.rs
+++ b/src/widgets/table.rs
@@ -26,6 +26,8 @@ use crate::{
 /// [`Table`] is also a [`StatefulWidget`], which means you can use it with [`TableState`] to allow
 /// the user to scroll through the rows and select one of them.
 ///
+/// Note: if the `widths` field is empty, the table will be rendered with equal widths.
+///
 /// See the [table example] and the recipe and traceroute tabs in the [demo2 example] for a more in
 /// depth example of the various configuration options and for how to handle state.
 ///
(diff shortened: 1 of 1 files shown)


Valentin271 (approved):
Looks good

orhun (approved):

Issue #711, opened 2023-12-26
State of stateful widgets should be serializable


MultisampledNight:
## Problem

I'm building a viewer for a relatively long schedule, and want to base that upon `Table`. Since the schedule is soooo looooong and the user is likely to close and open the app several times, I want the user to return to the exact spot of where they were.

However, [`TableState`] is relatively opaque, and doesn't seem to have anything I could use to serialize it.

[`TableState`]: https://docs.rs/ratatui/0.25.0/ratatui/widgets/struct.TableState.html

## Solution
New cargo features `rkyv` and `serde` which implement the respective {de,}serialization traits if enabled. 

## Alternatives

- Vendoring `TableState` into one's own code, implementing the respective traits, and using `TableState::with_offset` + `TableState::with_selected` anytime it needs to be restored
- <missing string table entry since I'm severely out of ideas>

## Additional context

None. I want to implement this though, at least for rkyv.

MultisampledNight:
Turns out there _already is_ a `serde` feature: https://github.com/ratatui-org/ratatui/blob/b0ed658970e8a94f25948c80d511102c197a8f6a/Cargo.toml#L74-L76

I'll extend that one to include state then.

joshka:
Regardingthe rkyv support part of this, it sounds to me that the reason for choosing this over serde is purely a performance thing.

Can you quantify the performance difference here? Is this likely to be a real benefit or one where we go from 0.010ms to 0.005ms for something that is called once in an app's lifetime?

Is there some other benefit I'm missing (i.e. do your already have an app that uses rkyv more generally that is blocked by not having this, or you have a use case where you do need to choose rkyv for wider compatibility reasons and this prevents you from doing so.

These questions might sound a bit pedantic. I'm trying to understand the tradeoff of added maintenance burden vs benefit.
Adding features that are for less commonly used libraries (serde:rkyv ~= 10:1 monthly downloads) and where the maintainers aren't as familiar with them (I'd never heard of rkyv before this issue) adds some extra maintenance burden. E.g. things like breaking serialization formats in new versions of Ratatui, or keeping track of downstream changes.

MultisampledNight:
(seen this pop up and decided it jumps the queue because otherwise I'll forget)

It's not only a performance thing, but also a space/memory efficiency thing! An overview about performance and size benchmarks is available at https://github.com/djkoloski/rust_serialization_benchmark, though it'd be interesting to perform a specific one for ratatui!

[...]

PR #712, opened 2023-12-26
feat(state): serialization, deserialization and rkyv support

MultisampledNight:
Adds `serde` for the state of stateful widgets as well, and introduces `rkyv` at every almost place where `serde` is found.

Closes https://github.com/ratatui-org/ratatui/issues/711.

**Note:** I did not test this one bit _yet_. I'm only putting it up to have it somewhere and to save someone else the effort.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,7 @@ termion = { version = "2.0", optional = true }
 termwiz = { version = "0.20.0", optional = true }
 
 serde = { version = "1", optional = true, features = ["derive"] }
+rkyv = { version = "0.7.43", optional = true }
 bitflags = "2.3"
 cassowary = "0.3"
 indoc = "2.0"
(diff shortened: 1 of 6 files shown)


MultisampledNight:
Looks like `bitflags` and `rkyv` don't play well with each other. I'll remove the rkyv attribute from everywhere they both are atm.

MultisampledNight:
Which was only in `State`, done. The push hook is currently running through, force push should appear soon.

MultisampledNight:
...and of course the effect from the missing `Style` impl propagates throughout `Buffer` and `TestBackend`, I'll remove those, too.

joshka:
Hey, how about splitting this into two PR, finishing the serde bits and introducing rkyv?
It would be nice to have some tests that show how the serialization works and which helps us understand whether there are backwards compatibility problems to deal with.

MultisampledNight:
Sure! I'm the next few days on a con, but should have time to do so and also add tests after new year.

MultisampledNight:
Will close this PR after I also made the PR about adding rkyv support.

MultisampledNight:
Superseded by #723 and #724.

PR #716, opened 2023-12-27
feat(table): Add a Table::footer method

yanganto:
The behavior is the same as the header of the table.

### DIFF
--- a/src/widgets/table.rs
+++ b/src/widgets/table.rs
@@ -45,6 +45,7 @@ use crate::{
 ///
 /// - [`Table::rows`] sets the rows of the [`Table`].
 /// - [`Table::header`] sets the header row of the [`Table`].
+/// - [`Table::footer`] sets the footer row of the [`Table`].
 /// - [`Table::widths`] sets the width constraints of each column.
 /// - [`Table::column_spacing`] sets the spacing between each column.
 /// - [`Table::block`] wraps the table in a [`Block`] widget.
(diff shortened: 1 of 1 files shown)


Valentin271 (approved):
Looks good. I can't test it right now so I'll wait for another maintainer to take a look or test.

joshka:
I like the idea, but there's some conflicting work (that needs to be updated). https://github.com/ratatui-org/ratatui/pull/516. Recently Table was refactored to move around all the methods for better top to bottom readability, which made the above PR difficult to merge (my guess is tat it will be a manual re-write of the functionality rather than just a rebase).

src/widgets/table.rs
```diff
@@ -1078,54 +1106,79 @@ impl<'a> StatefulWidget for Table<'a> {
         }
 
         // Draw rows
-        if self.rows.is_empty() {
-            return;
+        if !self.rows.is_empty() {
```
joshka, line 1109: Here, like in https://github.com/ratatui-org/ratatui/pull/516 I'd prefer to make the `render()` method call several smaller methods (`render_header()`, `render_rows()`, `render_footer()`) rather than increasing the size of this one method.

yanganto:
This is a tui using the current feature branch, the red line is the table footer.
footer

@joshka 
Please kindly let me know the next step on this PR.
Is it possible to merge, once the `render()` into three methods?
Do we need another PR based on #516, or just wait for #516 to complete?
Thanks in advance. :pray:

joshka:
@yanganto I've done the refactor I mentioned and split the tables module up into multiple files in https://github.com/ratatui-org/ratatui/pull/718. No need to wait on #516, but #718 makes things a bit easier to maintain long-term.

joshka:
718 is merged - feel free to update your code based on the changes. I would expect the additions of the extra field will apply cleanly, but the changes to draw the footer will not. That said, it's likely much easier to make the change now as the methods are small and obvious.

yanganto:
move to #719

PR #717, opened 2023-12-27
chore(ci): use cargo-nextest for running tests

orhun:
Switch to [nextest](https://nexte.st/) for running tests.

Motivation:

- `cargo test`: 19.332s
- `cargo nextest run`: 0.981s

Please note:

- `cargo-nextest` does not support running doctests yet: https://github.com/nextest-rs/nextest/issues/16
- The `cargo-nextest` binary is installed automatically via `cargo-make`, see https://github.com/sagiegurari/cargo-make#crates

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -151,6 +151,8 @@ jobs:
           toolchain: ${{ matrix.toolchain }}
       - name: Install cargo-make
         uses: taiki-e/install-action@cargo-make
+      - name: Install cargo-make
+        uses: taiki-e/install-action@nextest
       - name: Test ${{ matrix.backend }}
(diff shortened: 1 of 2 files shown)


joshka (approved):

joshka:
It would be good to be have the fast unit tests run before the slow doc tests.

orhun:
Good idea, done in 39e5d4581a8325ed18accfcedcd421cc61b4263c

What we can also do is running them in parallel, but then we get a bunch of "Blocking waiting for file lock on package cache" and the speedup is not significant.

joshka (approved):

joshka:
>What we can also do is running them in parallel, but then we get a bunch of "Blocking waiting for file lock on package cache" and the speedup is not significant.

Probably best to avoid that.

PR #718, opened 2023-12-27
refactor(table): split table into multiple files

joshka:
At close to 2000 lines of code, the table widget was getting a bit
unwieldy. This commit splits it into multiple files, one for each
struct, and one for the table itself.

Also refactors the table rendering code to be easier to maintain.

---

This is a fairly large PR to review, and I'd expect that reviewing the entire thing is not the right approach (as the majority of the code is just moved).

The main refactoring changes that are not pure moves or renames are at `impl StatefulWidget for Table` in `table/table.rs` as well as a few added unit tests in `mod render` in the same file. These tests are not fully comprehensive over the rendering method (as we have existing integration tests that help round out the coverage - these continue to pass).

https://github.com/ratatui-org/ratatui/blob/f6070a1a8d521c2a3696440b20289213cd185608/src/widgets/table/table.rs#L513-L536

The implementation reduces the amount of state that is passed in the function by using `Layout` rather than row / column indexes tracking.

https://github.com/ratatui-org/ratatui/blob/f6070a1a8d521c2a3696440b20289213cd185608/src/widgets/table/table.rs#L539-L550

The other changes of note are splitting up all the test methods into the right file and dropping unnecessary prefixes from test names (e.g. `table::tests::table_new()` -> `table::tests::new()`, `table::tests::cell_new()` -> `table::cell::tests::new()`)


kdheepak (approved):
Thanks for summarizing the changes. I'm approving without reading through every line.

PR #719, opened 2023-12-28
feat(table): Add Table::footer and Row::top_margin methods

yanganto:
The behavior is similar to the header of the table.

Due to comity #718, the PR is reopened with the same ideal as #716.

### DIFF
--- a/src/widgets/table/row.rs
+++ b/src/widgets/table/row.rs
@@ -59,6 +59,7 @@ use crate::prelude::*;
 pub struct Row<'a> {
     pub(crate) cells: Vec<Cell<'a>>,
     pub(crate) height: u16,
+    pub(crate) top_margin: u16,
     pub(crate) bottom_margin: u16,
     pub(crate) style: Style,
 }
@@ -141,6 +142,25 @@ impl<'a> Row<'a> {
         self
     }
 
+    /// Set the top margin. By default, the top margin is `0`.
+    ///
+    /// The top margin is the number of blank lines to be displayed before the row.
+    ///
+    /// This is a fluent setter method which must be chained or used as it consumes self
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// # use ratatui::{prelude::*, widgets::*};
+    /// # let cells = vec!["Cell 1", "Cell 2", "Cell 3"];
+    /// let row = Row::default().top_margin(1);
+    /// ```
+    #[must_use = "method moves the value of self and returns the modified value"]
+    pub fn top_margin(mut self, margin: u16) -> Self {
+        self.top_margin = margin;
+        self
+    }
+
     /// Set the bottom margin. By default, the bottom margin is `0`.
     ///
     /// The bottom margin is the number of blank lines to be displayed after the row.
@@ -194,7 +214,9 @@ impl<'a> Row<'a> {
 impl Row<'_> {
     /// Returns the total height of the row.
     pub(crate) fn height_with_margin(&self) -> u16 {
-        self.height.saturating_add(self.bottom_margin)
+        self.height
+            .saturating_add(self.top_margin)
+            .saturating_add(self.bottom_margin)
     }
 }
 
@@ -235,6 +257,12 @@ mod tests {
     fn height() {
         let row = Row::default().height(2);
         assert_eq!(row.height, 2);
+    }
+
+    #[test]
+    fn top_margin() {
+        let row = Row::default().top_margin(1);
+        assert_eq!(row.top_margin, 1);
     }
 
     #[test]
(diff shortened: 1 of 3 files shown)


yanganto:
Hi @Valentin271 @joshka 
Does this PR look good for you?
By the way, I believe the failure of CI is not related to the PR.  Please kindly review this PR.  Many thanks.

src/widgets/table/row.rs
```diff
@@ -141,6 +142,25 @@ impl<'a> Row<'a> {
         self
     }
 
+    /// Set the top margin. By default, the top margin is `0`.
+    ///
+    /// The top margin is the number of blank lines to be displayed after the row.
```
joshka, line 147:
```suggestion
    /// The top margin is the number of blank lines to be displayed before the row.
```

src/widgets/table/table.rs
```diff
@@ -561,21 +596,42 @@ impl Table<'_> {
 
     fn render_header(&self, area: Rect, buf: &mut Buffer, column_widths: &[(u16, u16)]) {
         if let Some(ref header) = self.header {
+            let area = Rect {
+                x: area.x,
+                y: area.y + header.top_margin,
+                width: area.width,
+                height: area.height - header.top_margin,
+            };
```
joshka, line 604: Do you think it might be clearer to instead add some extra unused space in the layout for margins above?
e.g.
```rust
.constraints([
  Constraint::Length(header.top_margin),
  Constraint::Length(header_height)
  Constraint::Length(header_bottom_margin),
  ...
```
(and then ignore those rows and just grab layout[1], [3], and [5])

Doing this could collapse the header / footer methods to a single method:
```rust
fn render_row(row: row, ...)
```
and could be called with
```rust
self.header.map(|row| render_row(row, header_area, ...);
self.footer.map(|row| render_row(row, footer_area, ...);
```

joshka (approved):
Generally LGTM, a quick fix for a typo, and a question about whether we might improve the code, but otherwise happy with things.

src/widgets/table/table.rs
yanganto: Agreed.

yanganto: Sounds good. Let me try first.

orhun:
I'd say let's update the examples to make these new methods more visible.

orhun (approved):

joshka (approved):

orhun:
We use [gpg signing](https://github.com/ratatui-org/ratatui/blob/main/CONTRIBUTING.md#sign-your-commits) for this project which means you need to sign your commits before the merge. Are you able to set it up / sign your commits?

yanganto:
I use `trezor-gpg` and it seems difficult to sign a commit with this. 
Also I try to generate a new one with gpg my machine(NixOS)

And it always hangs in this process.
```
gpg (GnuPG) 2.3.7; Copyright (C) 2021 Free Software Foundation, Inc.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Note: Use "gpg --full-generate-key" for a full featured key generation dialog.

GnuPG needs to construct a user ID to identify your key.

Real name: Antonio Yang
Email address: yanganto@gmail.com
You selected this USER-ID:
    "Antonio Yang <yanganto@gmail.com>"

Change (N)ame, (E)mail, or (O)kay/(Q)uit? O
We need to generate a lot of random bytes. It is a good idea to perform
some other action (type on the keyboard, move the mouse, utilize the
disks) during the prime generation; this gives the random number
generator a better chance to gain enough entropy.

```

Could I use "squash merge" on the GitHub web UI when I get all approvals?  Such that the commit in the commit history page will still be verified.  Is that good for you?

joshka:
> I use `trezor-gpg` and it seems difficult to sign a commit with this. Also I try to generate a new one with gpg my machine(NixOS)

This isn't something a combination of tools I'm familiar with.

> Could I use "squash merge" on the GitHub web UI when I get all approvals? Such that the commit in the commit history page will still be verified. Is that good for you?

Github prevents the merge due to the signing (and doesn't enable the squash merge button).
image

What I can do is pull this down, sign it and force push it myself if you're unable to sign the commits. Usually we like to encourage the original committer to sign the commits when possible though, so it's up to you.

yanganto:
Hmm... This is based on the setting of the repo.
You can see in my working repo every PR using `Squash Merge` will still have a verify tag, because the operation is based on GitHub web UI with authentication.
https://github.com/kuutamolabs/kld/commits/main/
And there is no sign of the commits of each feature PR
https://github.com/kuutamolabs/kld/pull/834

Will signing for others' commitment change the author? 
If it changes, there is no meaning to the signature anymore. This is different from a cherry pick, I think nobody should do that, and I feel bad for this kind of workaround, so please don't.
I will try to find another machine to sign these if the `Squash Merge` is forbidden for other reasons here.

joshka:
> Hmm... This is based on the setting of the repo.

Yep. We've intentionally turned that setting on.

image


>Will signing for others' commitment change the author?

The commit author and committer are two separate things. You'd be marked as the author, and I'd be marked as the committer. I'll wait and let you take a look at getting things signed properly, but the workaround is there if you need it.

PR #722, opened 2023-12-29
feat(table): Add Table::footer and Row::top_margin methods

yanganto:
The behavior is similar to the header of the table.

Due to comity #718 and gpg sign requirement in #719, the PR is reopened with the same ideal as https://github.com/ratatui-org/ratatui/pull/716.


joshka (approved):

joshka:
Thanks :)

yanganto:
@joshka  Thanks for the review.

joshka:
No problem :)

PR #725, opened 2024-01-02
feat: layout! macro for shorter layout construction ✨

kdheepak:


kdheepak:
From @joshka:

> 1. Can something like this work? 
> 
> ```rust
> let [header, main] = vertical_layout!(area, [1, *]);
> let [sidebar, content] = horizontal_layout!(main, [30%, 70%]);
> let [a, b, c] = horizontal_layout!(content, [1/3, 1/3, 1/3]);
> let [min, max] = vertical_layout!(a, [min(1), max(10)]);
> ```
> 
> 2. Can we use the star to indicate whatever is left (-> min(0))
> 3. Can a bare percent symbol work like that?
> 4. Can we write ratios like that?
> 5. Can the min/max work like this?
> 6. How would you specify the segment size?
> 7. does that syntax look about right?
> 
> A full example that should work would be:
> 
> ```
> let [a,b,c,d,e,f,g] = layout_horizontal!(area, [1, min(1), max(5), 10%, 1/2, *]);
> ```

kdheepak:
I'd like to get something like this working:

```rust
let [a,b,c,d,e,f] = layout_vertical!(area, [==1, >=1, <=5, ~=10%, ==1/2, >=0]);
```

The advantages of this is it is a more generalized syntax and we can add more constraints in the future and this syntax would be able to support it.

Thoughts?

joshka:
> I'd like to get something like this working:
> 
> ```rust
> let [a,b,c,d,e,f] = layout_vertical!(area, [==1, >=1, <=5, ~=10%, ==1/2, >=0]);
> ```
> 
> The advantages of this is it is a more generalized syntax and we can add more constraints in the future and this syntax would be able to support it.
> 
> Thoughts?

https://discord.com/channels/1070692720437383208/1072879985762762812/1191857716658962604
Has some ideas on this:

```rust
let [a,b,c,d,e,f,g] = layout_horizontal!(area, [1, min(1), max(5), 10%, 1/2, *]);
```

```rust
let [header, main] = vertical_layout!(area, [1, *]);
let [sidebar, content] = horizontal_layout!(main, [30%, 70%]);
let [a, b, c] = horizontal_layout!(content, [1/3; 3]);
let [min, max] = vertical_layout!(a, [min(1), max(10)]);
```

kdheepak:
I am not able to get `100%` to work with declarative macros:

```
error: expected expression, found `,`
   --> src/layout.rs:875:43
    |
810 |     ([$($constraint:expr),+]) => {{
    |         ---------------- while parsing argument for this `expr` macro fragment
...
875 |         let [rect1, rect2] = layout!([100%, >=3])
    |                                           ^ expected expression

warning: build failed, waiting for other jobs to finish...
```

I tried using `tt` as well but it treats it as two tokens, i.e. `100` and `%`. I'm looking into this more but I _think_ I would need to use procedural macros if we want the user facing syntax without double quotes.

kdheepak:
I figured out a way to get it to work without proc macros! 

We should discuss the following

1. Name for the macro.
  - I choose `layout!` but maybe we want it to be different from the `layout` module name? 
  - Right now, you can create vertical layouts using `layout!(v, 100%, >=3)` and horizontal layouts using `layout!(h, 100%, >=3)`.
  - If we use `vertical_layout!` and `horizontal_layout!`, we'll have to have three macros defined, the verbose names would forward to an internal only macro.
2. Finalize syntax for the macro for future compatibility.
  - I've chosen syntax that is as close to the mathematical representation of the constraints as possible
  - Because of that I think this syntax will be future proof, i.e. if we move to a different layout algorithm we'll still be able to use the same syntax in the macro by just changing the implementation details.

Valentin271:
1. I like the name, I think we should keep it.
As for `{vertical,horizontal}_layout` i'm not too fond of. I'm not too sure of the benefits of two additional macros. If we do keep them, what about `{v,h}layout`

2. At a glance the syntax looks nice. Will look more into it when on my computer though.

Overall I like this idea a lot.

kdheepak:
`@Sera` on Discord suggested the following syntax:

```rust
layout_horizontal!(area; min(1) => a, max(5) => b, 10% => c, 1/2 => d, * => e)
```

which would expand to

```rust
let [a,b,c,d,e] = Layout::horizontal().constraints([Constraints::Min(1), Constraints::Max(5), Constraints::Percentage(10), Constraints::Ratio(1, 2), Constraints::Min(0)]).split_into_array(area);
```

Unfortunately, I believe this is much harder to do or even impossible to do using declarative macros. The reason is that (in my understanding) declarative macros "match" on syntax and return valid rust expressions. 

Here's an implementation that matches on the intended syntax:

```rust
#[macro_export]
macro_rules! named_layout {
    // Horizontal layout variant
    ($area:ident, h, $( $constraint:tt )+ ) => {
        named_layout!(@construct $area, $crate::prelude::Direction::Horizontal, named_layout!(@parse () $($constraint)+))
    };
    // Vertical layout variant
    ($area:ident, v, $( $constraint:tt )+ ) => {
        named_layout!(@construct $area, $crate::prelude::Direction::Vertical, named_layout!(@parse () $($constraint)+))
    };
    // Internal parsing rule for constraints
    // This rule checks if `,` exists after the `head` token
    (@parse ($($acc:tt)*) $head:tt , $($tail:tt)*) => {
        // Combines the head constraint with the tail constraints into a vector
        std::iter::once(
            named_layout!(@process ($($acc)* $head)) // -> Constraint
        ).chain(
            named_layout!(@parse () $($tail)*).into_iter() // -> Iterator with type Constraint
        )
    };
    // If there is no `,` then accumulate `next` token into existing `acc` tokens
    // and return `Iterator`
    (@parse ($($acc:tt)*) $next:tt $($tail:tt)*) => {
        named_layout!(@parse ($($acc)* $next) $($tail)*)
    };
    // At the end there will a set of tokens left after the last `,`
    // Process that as a `Constraint`
    (@parse ($($acc:tt)*)) => {
        vec![named_layout!(@process ($($acc)*))]
    };
    // Process different types of constraints into a `Constraint`
    (@process ($variable:ident == $token1:tt / $token2:tt)) => {
        // Ratio constraint
        {
        let t1: u32 = $token1;
        let t2: u32 = $token2;
        (stringify!($variable), $crate::prelude::Constraint::Ratio(t1, t2))
        }
    };
    (@process ($variable:ident == $token:tt %)) => {
        // Percentage constraint
        (stringify!($variable), $crate::prelude::Constraint::Percentage($token))
    };
    (@process ($variable:ident >= $token:tt)) => {
        // Minimum size constraint
       (stringify!($variable), $crate::prelude::Constraint::Min($token))
    };
    (@process ($variable:ident <= $token:tt)) => {
        // Maximum size constraint
        (stringify!($variable), $crate::prelude::Constraint::Max($token))
    };
    (@process ($variable:ident == $token:tt)) => {
        // Fixed size constraint
        (stringify!($variable), $crate::prelude::Constraint::Length($token))
    };
    // Construct the final `Layout` object
    (@construct $area:ident, $direction:expr, $constraints:expr) => {
        {
        let _variables = $constraints.map(|t| t.0); // Iterator of &str
        let constraints = $constraints.map(|t| t.1); 
        $crate::prelude::Layout::default()
            .direction($direction)
            .constraints(constraints)
            .split($area)
            .to_vec()
            .try_into()
            .unwrap()
        }
    };
}

#[test]
fn named_layout_macro_constraints() {

    let rect = Rect {
        x: 0,
        y: 0,
        width: 10,
        height: 10,
    };

    let [_rect1, _rect2] = named_layout!(rect, v, rect1 == 7, rect2 <= 3);
}
```

Unfortunately, there doesn't seem to be a way to retrieve the identifier from the inner process match. So I don't know how to get it to work with this as the user facing syntax:

```rust
    named_layout!(rect, v, rect1 == 7, rect2 <= 3);
```

This would be possible to do with a proc macro, since we can build the desired token tree and return that to the user.
I'll wait for input from others to see if we want to go in that direction.

kdheepak:
Should we also have a macro to make it easier to construct just the constraints? i.e.`constraints!( == 100%, >= 5 )` will return `[Constraint::Percentage(100), Constraint::Min(5)]`. This will be useful for table widths.

kdheepak:
Also for discussion, currently this PR does:

```rust
let [a, b, c, d, e, f] = layout!(h, ==1, >=1, <=1, ==1/2, ==10%, >=0).split_into_array(rect)
```

Another option would be:

```rust
let [a, b, c, d, e, f] = layout!([==1, >=1, <=1, ==1/2, ==10%, >=0], orientation = horizontal).split_into_array(rect)
```

I think this better allows space in the design for more options to be added in the future, e.g.

```rust
let [a, b, c, d, e, f] = layout!(
    [==1, >=1, <=1, ==1/2, ==10%, >=0], 
    orientiation = horizontal, 
    segment_size = distribute_evenly
).split_into_array(rect)
```

deifactor:
```rs
macro_rules! do_layout {
    ($area:expr; $var:ident $($rest:tt)*) => { do_layout!($area; @vars () @constraints () $var $($rest)* ) };

    ($area:expr; @vars ($($var:ident)*) @constraints ($($cons:expr)*) $variable:ident == $tok1:tt / $tok2:tt , $($rest:tt)*) =>
    { do_layout!($area; @vars ($($var)* $variable) @constraints ($($cons)* Constraint::Ratio($tok1, $tok2)) $($rest)*) };

    ($area:expr; @vars ($($var:ident)*) @constraints ($($cons:expr)*) $variable:ident == $tok:expr , $($rest:tt)*) =>
    { do_layout!($area; @vars ($($var)* $variable) @constraints ($($cons)* Constraint::Length($tok)) $($rest)*) };

    ($area:expr; @vars ($($var:ident)*) @constraints ($($cons:expr)*) $variable:ident >= $tok:expr , $($rest:tt)*) =>
    { do_layout!($area; @vars ($($var)* $variable) @constraints ($($cons)* Constraint::Min($tok)) $($rest)*) };

    ($area:expr; @vars ($($var:ident)*) @constraints ($($cons:expr)*) $variable:ident <= $tok:expr , $($rest:tt)*) =>
    { do_layout!($area; @vars ($($var)* $variable) @constraints ($($cons)* Constraint::Max($tok)) $($rest)*) };

    ($area:expr; @vars ($($var:ident)*) @constraints ($($cons:expr)*)) => {
        use std::ops::Deref;

        let [$($var),*] = ::ratatui::prelude::Layout::default().direction(Direction::Horizontal).constraints([$($cons),*]).split($area).deref().try_into().unwrap();
    }
}

#[cfg(test)]
mod tests {
    use ratatui::prelude::Rect;

    use super::*;

    #[test]
    fn test_macro() {
        let rect = Rect { x: 0, y: 0, width: 10, height: 10 };
        let one = 1;
        do_layout!(rect; x == one, y >= (one + one), z <= 3, w == 1 / 10,);
        assert_eq!(x, Rect::new(0, 0, 1, 10));
        assert_eq!(y, Rect::new(1, 0, 5, 10));
        assert_eq!(z, Rect::new(6, 0, 3, 10));
        assert_eq!(w, Rect::new(9, 0, 1, 10));
    }
}
```

Didn't add support for directionality, % constraints but it should be workable. I think not requiring the trailing comma will require duplicating all the 'constraint-builder' arms, annoyingly.

I'm not sure if the niceness of having the variables next to the constraints outweighs the inflexibility of not being able to manipulate the layout as an object. Maybe have both? I have more thoughts but my flight is literally boarding so I'll send them later.

kdheepak:
Is it even possible to remove the trailing comma in your rules? Because `tok:expr` must be followed by a `=>`, `,` or `;`. I imagine we'd have to rearrange it for that to work, like in your original post, i.e.: `do_layout!( area; 100% => a )`

src/layout.rs
```diff
@@ -790,12 +790,183 @@ impl Element {
...
+/// - `== $token:tt`: Sets a fixed size constraint for the token.
+///
+/// # Examples
+///
+/// ```
+/// // Vertical layout with fixed size and percentage constraints
+/// use ratatui::layout;
+/// use ratatui::constraints;
+/// layout!([== 50, == 30%]);
+/// ```
```
kdheepak, line 895: Oddly though, you have to import both `layout` and `constraints` to use `layout!`.

Doing this was throwing an error for the doctests:

```rust
use ratatui::layout;
layout!([==50, ==30%]);
```

kdheepak:
> Also for discussion, currently the syntax of this PR is:
> 
> ```rust
> let [a, b, c, d, e, f] = layout!(h, ==1, >=1, <=1, ==1/2, ==10%, >=0).split_into_array(rect)
> ```
> 
> Another option for the syntax would be to introduce `[]` for the constraints, i.e.:
> 
> ```rust
> let [a, b, c, d, e, f] = layout!([==1, >=1, <=1, ==1/2, ==10%, >=0], orientation = horizontal).split_into_array(rect)
> ```
> 

I changed the PR to use the syntax in the second option. I also added a `constraints!` macro.

Valentin271:
I like the idea of having a independent `constraints!` macro that can be used elsewhere and internally by `layout!`.
I also like the fact you moved the `orientation` as a second optional named parameter. Though I'd call it `direction`, to stay consistent with `Layout::new`.

Also it'd be nice if we could destructure the resulting constraints like so (probably linked to https://github.com/ratatui-org/ratatui/pull/729)
```rust
let [a, b] = constraints!(==50%, ==50%);
```

---

> ```rust
> layout_horizontal!(area; min(1) => a, max(5) => b, 10% => c, 1/2 => d, * => e)
> ```

I don't like this syntax, it feels unclear to me. Additionally, you can't store the resulting layout, i.e. `let layout = layout!(...)`  which makes the macro less versatile.

src/layout.rs
kdheepak: Resolved by adding a prefix everywhere, i.e. `$crate::constraints!`

kdheepak:
Thanks for the feedback! I renamed `orient` to `direction`. 

I tried this and it works:

```rust
let [a, b] = constraints!([==50%, ==50%]);
```

All I had to do was add these lines to inside the macro:

```rust
macro_rules! constraints {
    ([ $( $constraint:tt )+ ]) => {
        $crate::constraints!(@parse () $($constraint)+)
            .collect::<Vec<Constraint>>() // add this line 
            .try_into()                   // add this line
            .unwrap();                    // add this line
    };
```

However, when I added these lines, I was getting errors elsewhere. For example, when doing:

```rust
let constraints = constraints!([==50%, ==50%]);
```

I was getting an compile error where Rust wasn't able to infer the types. Maybe you or someone else knows a better way to solve that? I reverted this change.

Personally, I think it is less common that someone is going to want to split the constraints, right?

kdheepak:
I do like the idea of having the rect names next to the constraints, i.e.

```rust
do_layout!(rect; x == 1, y >= 2, z <= 3, w == 1 / 10,);
```

I think there's more magic going on there for users to understand it though. And the equivalent with the current macro in this PR isn't too far off. 

```rust
let [x, y, z, w] = layout!(
    [==1, >=2, <=3, ==1/10]
).split_into_array(rect);
```

For me, it is easy to pattern match the `[]` and the number of `,`s in the LHS and RHS with this syntax, and is a good compromise imho.

Valentin271:
> Personally, I think it is less common that someone is going to want to split the constraints, right?

Oh right it was an oversight of me. I meant (destructuring **layout**)

```rust
let [a, b] = layout!([==50%, ==50%]);
```

I agree constraints are not going to be split.

deifactor:
You should be able to rework the macro so that it just builds an array literal instead of chaining iterators, like in my example up above. In that case you wouldn't need anything special because the output would just be [Constraints::Max(1), Constraints::Min(2)] or what have you.

kdheepak:
We will need `area` to be part of layout as well, then, right? I think we can get this syntax working:

```
let [a, b] = layout!(area, [==50%, ==50%]);
```

deifactor:
Yeah, that's workable. Though I'm somewhat less certain about the usefulness of a layout macro that splits for you since it doesn't "compose"; if you add more builder methods then you have to add more code to the macro, whereas if it just evaluated to a Layout object you could keep chaining.

On Wed, Jan 3, 2024, at 10:20 PM, Dheepak Krishnamurthy wrote:
> 
> 
> We will need `area` to be part of layout as well, then, right? I think we can get this syntax working:
> 
> `let [a, b] = layout!(area, [==50%, ==50%]);
`

Valentin271:
If anyone wants to try this out and give feedback, see https://github.com/kdheepak/ratatui-macros

src/layout.rs
```diff
@@ -4,6 +4,7 @@ mod corner;
 mod direction;
 #[allow(clippy::module_inception)]
 mod layout;
+mod macros;
```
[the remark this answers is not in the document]
kdheepak: Good idea! Done in https://github.com/ratatui-org/ratatui/pull/725/commits/508bc1e89b49025a55d342f433653076f4c2178b

orhun:
Can you also update the description of the `macros` feature?

orhun (approved):
looks neat to me

joshka (approved):
I still don't 100% jive with the syntax for specifying the constraints here. To me they feel a bit weird. But that's more of a subjective perspective than objective.

Do we still want to test this externally a bit more before rolling it into the core?

kdheepak:
I think the implementation now is easy to understand ( thanks to @deifactor ) but I want to point out a couple of things.

1. This works fine:

```rust
vertical![
    >=3,
    ==100%
]
```

but the following will fail to compile:

```rust
vertical![
    >=3,
    ==100%,
]
```

because of a trailing comma. I _think_ this is quite hard to fix with just a declarative macro but would be possible with a proc macro.

2. If we add more features to the constraints will the macro be able to accommodate it? For example, what's a good syntax for `Constraint::Fixed`?

I think it may be a good idea to see what the adoption is like as a third party crate. I imagine it'll get more adoption if it were in the library itself, but I'd be more comfortable if I were completely happy with how layouts are handled and didn't think there were new features needed or if I thought I could accommodate any new feature or new syntax.

joshka:
The other thing that keeping it in another crate would help is if we want layout to go 2D at some point (distinctly possible), it might be nice to have the crate `layout!` macro handle horizontal and vertical layout without being constrained into a syntax that makes it difficult to change. I'm not sure if that would be the case here and YAGNI would apply if we weren't already thinking about some of the Taffy things.

Valentin271:
I'd like to wait a little bit for this, I guess aim this for 0.27. So people (including myself) have time to try ratatui-macro and give feedback.
I don't want to engage in something that will change every realease because it wasn't tested well enough.
(Or pur it behind an unstable flag for 1 release, your call)

deifactor:
Did some experimenting and you can get the trailing comma to work if you move the `[$(,)?]` rule to be the very first one.

I definitely like the idea of having it in a separate crate for now.

PR #726, opened 2024-01-02
feat(termwiz): add `From` termwiz style impls

### LINKED ISSUE #679: Implement `From` for `Style`

Valentin271:
Closes #679

Important note: this also fixes a wrong mapping between ratatui's gray and termwiz's grey. `ratatui::Color::Gray` now maps to `termwiz::color::AnsiColor::Silver`

### DIFF
--- a/src/backend/termwiz.rs
+++ b/src/backend/termwiz.rs
@@ -405,3 +405,145 @@ impl From<LinearRgba> for Color {
 fn u16_max(i: usize) -> u16 {
     u16::try_from(i).unwrap_or(u16::MAX)
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    mod into_color {
+        use Color as C;
+
+        use super::*;
+
+        #[test]
+        fn from_linear_rgba() {
+            // full black + opaque
+            assert_eq!(C::from(LinearRgba(0., 0., 0., 1.)), Color::Rgb(0, 0, 0));
+            // full black + transparent
+            assert_eq!(C::from(LinearRgba(0., 0., 0., 0.)), Color::Rgb(0, 0, 0));
+
+            // full white + opaque
+            assert_eq!(C::from(LinearRgba(1., 1., 1., 1.)), C::Rgb(254, 254, 254));
+            // full white + transparent
+            assert_eq!(C::from(LinearRgba(1., 1., 1., 0.)), C::Rgb(254, 254, 254));
+
+            // full red
+            assert_eq!(C::from(LinearRgba(1., 0., 0., 1.)), C::Rgb(254, 0, 0));
+            // full green
+            assert_eq!(C::from(LinearRgba(0., 1., 0., 1.)), C::Rgb(0, 254, 0));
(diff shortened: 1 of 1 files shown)


src/backend/termwiz.rs
```diff
@@ -249,12 +249,73 @@ impl Backend for TermwizBackend {
...
+}
+
 impl From<Color> for ColorAttribute {
     fn from(color: Color) -> ColorAttribute {
         match color {
             Color::Reset => ColorAttribute::Default,
             Color::Black => AnsiColor::Black.into(),
-            Color::Gray | Color::DarkGray => AnsiColor::Grey.into(),
+            Color::DarkGray => AnsiColor::Grey.into(),
+            Color::Gray => AnsiColor::Silver.into(),
```
Valentin271, line 318: | ANSI | ratatui | termwiz |
|--------|--------|--------|
| white (37) | Gray <sup>1</sup> | Silver <sup>2</sup> |
| bright black (90) | DarkGray <sup>3</sup> | Grey | 

1: As stated in our doc "_this is sometimes called `silver` or `white` but we use `white` for bright white_"
2: As stated in [termwiz doc](https://docs.rs/termwiz/latest/termwiz/color/enum.AnsiColor.html#variant.Silver), silver is "dark white". "dark" in termwiz actually means normal in ANSI

3: As stated in our doc "_this is sometimes called `light black` or `bright black` but we use `dark gray`_"
4: According to [termwiz doc](https://docs.rs/termwiz/latest/termwiz/color/enum.AnsiColor.html#variant.Grey), grey is "bright black"

joshka: This would probably be a breaking change, or worth noting in the changelog specifically.

joshka (approved):
LGTM, though when merging this make sure to document the color change for the changelog.

src/backend/termwiz.rs
Valentin271: It *is* a breaking changing, but unlikely to hit anyone in a breaking way. Definitely worth mentionning in the changelog though 👍

PR #727, opened 2024-01-02
feat(layout): rename SegmentSize to Spread

joshka:
For some time we've been using the term "segment size" to refer to the
way extra space is distributed when the constraints are satisfied. This
is a very confusing name, and it's not clear what it means. This commit
renames it to "spread", which is a more common term for this concept.

Removed the unstable feature gate for this API now that a good name has
been found.

BREAKING CHANGE:
- SegmentSize has been renamed to Spread
- unstable-segment-size feature has been removed
- Layout::segment_size() has been renamed to Layout::spread()
- Table::segment_size() has been renamed to Table::spread()

### DIFF
--- a/src/layout.rs
+++ b/src/layout.rs
@@ -33,7 +33,7 @@ thread_local! {
 /// - a set of constraints (length, ratio, percentage, min, max)
 /// - a margin (horizontal and vertical), the space between the edge of the main area and the split
 ///   areas
-/// - extra options for segment size preferences
+/// - extra options for spread preferences
 ///
 /// The algorithm used to compute the layout is based on the [`cassowary-rs`] solver. It is a simple
 /// linear solver that can be used to solve linear equations and inequalities. In our case, we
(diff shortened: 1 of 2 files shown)


joshka:
I want a couple of approvals on this so we have consensus on the name.

kdheepak:
Is it a breaking change if this was gated behind a unstable flag?

joshka:
> Is it a breaking change if this was gated behind a unstable flag?

Yeah, but not one I'd worry about too much. Doesn't hurt to mark it as such.

kdheepak:
The title probably needs a `!` then, right? 

For some reason the `check-breaking-change-label` action isn't catching it.

Valentin271 (approved):
I like this name a lot more.

> For some reason the `check-breaking-change-label` action isn't catching it.

If I'm not mistaken the action only checks for `!` in the title and, if found, adds the `breaking-change` label.

orhun (approved):
`Spread` stills does not sound like the exact definition of this - or maybe it is because I got used to `SegmentSize` at this point. I will give it some time to grow on me.

kdheepak:
Can we hold off on merging this? Or still keep it behind the unstable flag? 

I like the name `Spread` as a verb describing what we are doing with the remaining segment size, but the variant names don't quite make sense now, right?

orhun:
Agree with kd here!

joshka:
The enum describes an action that happens (the allocation of remaining space once all constraints are satisfied). I think this would suggest a verb name is more appropriate than a noun for that concept. SegmentSize has never been clear to me. What is a segment? Why is being sized?

If you do feel more nounish, then something that notifies a verb is good. Eg allocate extra space -> ExtraSpaceAllocation. Distribute-> distribution ("spread" -> "spread" as it is bough a verb and the result of that verb having been done - this seems convenient).

I like the variant renames suggested.

joshka:
#804 makes this no longer necessary.

PR #729, opened 2024-01-02
feat(layout): add Rect::split method

joshka:
This method splits a Rect and returns a fixed-size array of the
resulting Rects. This allows the caller to use array destructuring
to get the individual Rects.

```rust
use Constraint::*;
let layout = &Layout::vertical([Length(1), Min(0)]);
let [top, main] = area.split(&layout);
```

### DIFF
--- a/src/layout/rect.rs
+++ b/src/layout/rect.rs
@@ -167,6 +167,32 @@ impl Rect {
             && self.y < other.bottom()
             && self.bottom() > other.y
     }
+
+    /// Split the rect into a number of sub-rects according to the given [`Layout`]`.
+    ///
+    /// An ergonomic wrapper around [`Layout::split`] that returns an array of `Rect`s instead of
+    /// `Rc<[Rect]>`.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the number of constraints is not equal to the length of the returned array.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// # use ratatui::prelude::*;
+    /// # fn render(frame: &mut Frame) {
+    /// let area = frame.size();
+    /// let layout = Layout::vertical([Constraint::Length(1), Constraint::Min(0)]);
+    /// let [top, main] = area.split(&layout);
+    /// # }
+    pub fn split<const N: usize>(self, layout: &Layout) -> [Rect; N] {
+        layout
+            .split(self)
+            .to_vec()
+            .try_into()
+            .expect("invalid number of rects")
+    }
 }
 
 #[cfg(test)]
@@ -352,5 +378,20 @@ mod tests {
         const _TOP: u16 = RECT.top();
         const _BOTTOM: u16 = RECT.bottom();
         assert!(RECT.intersects(RECT));
+    }
+
+    #[test]
+    fn split() {
+        let layout = Layout::horizontal([Constraint::Percentage(50), Constraint::Percentage(50)]);
+        let [a, b] = Rect::new(0, 0, 2, 1).split(&layout);
+        assert_eq!(a, Rect::new(0, 0, 1, 1));
+        assert_eq!(b, Rect::new(1, 0, 1, 1));
+    }
+
+    #[test]
+    #[should_panic(expected = "invalid number of rects")]
+    fn split_invalid_number_of_recs() {
+        let layout = Layout::horizontal([Constraint::Percentage(50), Constraint::Percentage(50)]);
+        let [_a, _b, _c] = Rect::new(0, 0, 2, 1).split(&layout);
     }
 }


orhun (approved):

joshka:
I want two approvals on this one - mainly for working out whether the naming makes sense.

kdheepak:
In terms of names, I would have preferred `split_into()` if `area` was stored in the layout, i.e.

```rust
let [top, main] =
    Layout::vertical([Constraint::Length(1), Constraint::Min(0)])
    .split_array(area);
```

versus

```rust
let [top, main] =
    Layout::vertical([Constraint::Length(1), Constraint::Min(0)])
    .area(area)
    .split_into();
```

With the functionality as is, I think the name `split_array()` is great!

kdheepak:
I have a small preference for `.split_into_array(area)` as it would be more descriptive and representative of what the function does, i.e.

```rust
layout.split(area)
    .to_vec()
    .try_into()
    .unwrap()
```

is replaced by this:

```rust
layout.split_into_array(area)
```

Valentin271:
I like `split_into_array` as a descriptive name but perhaps it feels too verbose.

> it feels too verbose

Once merged, I assume this will be the main splitting method used.

kdheepak:
Maybe it is worth introducing a breaking change for this and replacing `.split(area)` with this?

joshka:
I'm 👎 on breaking .spilt - 1.1K calls
- https://github.com/search?q=ratatui+layout%3A%3A+.split%28+lang%3Arust+NOT+repo%3Aratatui-org%2Fratatui&type=code

But what about split_area() as a compromise on verbosity?

>Once merged, I assume this will be the main splitting method used.

There's a PR that shows how the code would look using this https://github.com/ratatui-org/ratatui/pull/731

kdheepak:
I like in order of preference:

1. `.split_into(area)`: it's as short as we can get while being descriptive, it hints at what's going on inside the function (i.e. `try_into`)
2. `.split_into_array(area)`: it's descriptive but verbose for something that might be the default way people use this function
3. `.split_array(area)`: it's short but I don't think I'd be able to guess what is going on in the function? It could be misinterpreted as `split the argument which is an array` which is why I like `.split_into_array(area)` more than this
4. `.split_area(area)`: it's short but repeats the argument name in the function name so it falls down the list

joshka:
I think 1,2, and 3 are all problematic because they imply that the area is the (linguistic) object associated with the verb (split_into, split_into_array, split_array). Area is not the object of the sentence its the subject. Only a bare verb, or a verb phrase that repeats the subject communicates this properly.

Put in a sentence we want:
- The area is split by the layout
Not:
- The layout is split into the area
- The layout is split into an array that is the area
- The layout splits an array that is the area

joshka:
Chatting on Discord - @kdheepak made the observation that this then leads to:
```rust
area.split(Layout::horizontal([constraints]));
```
which I've added.
This now also includes the helper methods too.

src/layout.rs
```diff
@@ -338,9 +338,12 @@ impl Layout {
     pub fn horizontal<I>(constraints: I) -> Layout
     where
         I: IntoIterator,
-        I::Item: AsRef<Constraint>,
+        I::Item: Into<Constraint>,
```
joshka, line 341: I don't think we can change the new() impl to use Into instead of AsRef, as it would break too much code. Need to investigate the impact. These methods are not yeat released, so we can break code easily.

Making this accpet Into<Constraint> allows for:
```rust
Layout::horizontal([1,2,3,4])
```

src/layout.rs
```diff
@@ -806,6 +809,61 @@ impl Constraint {
...
+///
+/// Equivalent to `Constraint::Length(length)`
+pub fn eq(length: u16) -> Constraint {
+    Constraint::Length(length)
+}
+
+/// Helper function to create a Min constraint
+///
+/// Equivalent to `Constraint::Min(length)`
+pub fn ge(length: u16) -> Constraint {
```
joshka, line 822: Do we want both this and min?

joshka:
Needs:
- docs
- updated changelog message
- examples in website to be changed to use this approach

kdheepak (approved):
This looks good to me! 

But maybe useful to get one more review for feedback on `area.split(&layout)`?

joshka:
> This looks good to me!
> 
> But maybe useful to get one more review for feedback on `area.split(&layout)`?

Yep - definitely want one on this. I'm going to rebase and rewrite the commit for the CHANGELOG also.

joshka:
rebased and commit message / PR text updated.
There's a build failure that is fixed in #743

Valentin271 (approved):
Would be nice if we could write

```rust
area.split([Constraint::Length(10), Constraint::Length(10)]);
area.split([10, 10]);
```

Not too sure about the second one but the first one is definitely something I'd like.

Approving as this can be part of another PR.

orhun (approved):

joshka:
> Would be nice if we could write
> 
> ```rust
> area.split([Constraint::Length(10), Constraint::Length(10)]);
> area.split([10, 10]);
> ```
> 
> Not too sure about the second one but the first one is definitely something I'd like.
> 
> Approving as this can be part of another PR.

Direction to split is necessary, so perhaps `Rect::split_horizontal()`:
```rust
fn split_horizontal<const N: usize, T: IntoIterator<Item = Into<Constraint>>(self, constraints: T) -> [Rect; N] {
    self.split(Layout::horizontal(constraints));
}
```
But at that point we're just losing a single word (Layout), which is a good one to keep as it makes it a little more obvious what the [1,2,3] does. I like there to be a solid obvious way to do something. Do you think having both `split(layout)` and `split_horizontal(constraints)` might add one too many ways that would make it difficult to choose which method is the right one to use?

Valentin271:
> it makes it a little more obvious what the [1,2,3] does

I wasn't too sure about that, let's forget that for now.

> Do you think having both split(layout) and split_horizontal(constraints) might add one too many ways that would make it difficult to choose which method is the right one to use?

Probably yes, this is confusing.

---

I though of my suggestion as

```rust
fn split<const N: usize, L: Into<Layout>>(self, layout: L)
```

That way you can still write `.split(layout)` which is the "solid obvious way", and is good if you want to store layout (for configuration perhaps). And for quick and easy split use `Into`.

That would require implementing `From<Vec<Constraint>> for Layout` or similar. But overall it makes sense to convert a bunch of constraint to a layout.

> Direction to split is necessary

As for direction I'd just use the default one. Use layout if you want something else or

joshka:
I'd prefer to avoid defaulting the direction implicitly as it makes the code difficult to read unless you know what that default is. Happy to do it with `default(), but it should be explicit otherwise.

Valentin271:
Alright lets leave it as is for now. On second thought after reading the updated examples it seems fine to me.

PR #731, opened 2024-01-03
docs(examples): simplify docs using new layout methods

joshka:
Use the new `Layout::horizontal` and `vertical` constructors and
`Rect::split_array` through all the examples.

### DIFF
--- a/examples/block.rs
+++ b/examples/block.rs
@@ -103,20 +103,14 @@ fn ui(frame: &mut Frame) {
 ///
 /// Returns a tuple of the title area and the main areas.
 fn calculate_layout(area: Rect) -> (Rect, Vec<Vec<Rect>>) {
-    let layout = Layout::default()
-        .direction(Direction::Vertical)
-        .constraints([Constraint::Length(1), Constraint::Min(0)])
-        .split(area);
-    let title_area = layout[0];
-    let main_areas = Layout::default()
-        .direction(Direction::Vertical)
-        .constraints([Constraint::Max(4); 9])
-        .split(layout[1])
+    let main_layout = Layout::vertical([Constraint::Length(1), Constraint::Min(0)]);
+    let block_layout = &Layout::vertical([Constraint::Max(4); 9]);
+    let [title_area, main_area] = area.split(&main_layout);
+    let main_areas = block_layout
+        .split(main_area)
         .iter()
         .map(|&area| {
-            Layout::default()
-                .direction(Direction::Horizontal)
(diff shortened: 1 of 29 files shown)


joshka:
Removed the helper methods. They don't make things particularly clearer and have problems with polluting the imports. They're tough to put in the prelude due to the common names (min/max/length), which overall suggests that we're better without them. Diff when removed was 254 lines added, 258 removed, so the benefit on LoC is negligible.

joshka:
Rebased on the current version of main with Rect:split()

joshka:
Regarding testing this, the easiest way is to run `makers run-examples` which runs each example one after the other. All still work fine.

joshka:
Rebased now that #729 is merged.

Valentin271 (approved):
Looks good, I can confirm all examples are working and layouts are visually ok.

kdheepak:
I think it is a little confusing that we have 

```
let rects = layout.split(area);
```

and we have

```
let [rect1, rect2] = area.split(layout);
```

Is there a reason we couldn't always use the second form?
If there is a good reason for choosing the first form, I think at the very least it should be documented on the website.

joshka:
> Is there a reason we couldn't always use the second form? If there is a good reason for choosing the first form, I think at the very least it should be documented on the website.

The area.split(layout) form requires knowing the number of rects that will be returned beforehand. The layout.split(area) form does not. The places where I've left that are places where there are a dynamic number of areas in which to split to.

I also left a couple of more complex examples alone (IIRC it was in the old demo).

You're right - this could do with some docs in `Layout::split`

examples/block.rs
```diff
@@ -103,20 +103,14 @@ fn ui(frame: &mut Frame) {
...
-    let title_area = layout[0];
-    let main_areas = Layout::default()
-        .direction(Direction::Vertical)
-        .constraints([Constraint::Max(4); 9])
-        .split(layout[1])
+    let main_layout = Layout::vertical([Constraint::Length(1), Constraint::Min(0)]);
+    let block_layout = &Layout::vertical([Constraint::Max(4); 9]);
+    let [title_area, main_area] = area.split(&main_layout);
+    let main_areas = block_layout
+        .split(main_area)
```
kdheepak, line 110: Just for my understanding, here even though we know that the number of `Rect`s after the split has to be `2`, the compiler is not able to figure it out at compile time so we have to use `block_layout.split(main_area)`?

Presumably the following would work?

```
main_area.split::<2>(block_layout)
```

joshka: yep

kdheepak (approved):

PR #734, opened 2024-01-04
feat(layout): helper methods

joshka:
This is just a demo of some ideas we're discussing in discord.

- feat(layout): add split_array method
- feat: add layout helper methods
- feat: add layout helper methods


joshka:
Consensus after chatting a bit on this one is that it doesn't add a lot of value over doing something like `use Constraint as C` in the methods that need constraints as it would still require extra imports, and the method names are bad to put in the prelude as they would tend to clash with common method and variable names.

PR #744, opened 2024-01-05
feat(layout): accept Into<Constraint> for constructors

joshka:
This allows Layout constructors to accept any type that implements
Into<Constraint> instead of just AsRef<Constraint>. This is useful when
you want to specify a fixed size for a layout, but don't want to
explicitly create a Constraint::Length yourself.

```rust
Layout::new(Direction::Vertical, [1, 2, 3]);
Layout::horizontal([1, 2, 3]);
Layout::vertical([1, 2, 3]);
Layout::default().constraints([1, 2, 3]);
```

### DIFF
--- a/src/layout/constraint.rs
+++ b/src/layout/constraint.rs
@@ -205,6 +205,32 @@ impl Constraint {
     }
 }
 
+impl From<u16> for Constraint {
+    /// Convert a u16 into a [Constraint::Length]
+    ///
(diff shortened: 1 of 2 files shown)


kdheepak (approved):

kdheepak:
@a-kenji feel free to provide feedback on what you think about this PR, even if it is merged! We'd definitely want to hear everyone's opinions.

PR #749, opened 2024-01-05
feat(layout): add `Rect::clamp()` method

joshka:
This ensures a rectangle does not end up outside an area. This is useful
when you want to be able to dynamically move a rectangle around, but
keep it constrained to a certain area.

For example, this can be used to implement a draggable window that can
be moved around, but not outside the terminal window.

```rust
let window_area = Rect::new(state.x, state.y, 20, 20).clamp(area);
state.x = rect.x;
state.y = rect.y;
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,6 +40,7 @@ unicode-width = "0.1"
 document-features = { version = "0.2.7", optional = true }
 lru = "0.12.0"
 stability = "0.1.1"
+rstest = "0.18.2"
 
 [dev-dependencies]
 anyhow = "1.0.71"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,7 +40,6 @@ unicode-width = "0.1"
 document-features = { version = "0.2.7", optional = true }
 lru = "0.12.0"
 stability = "0.1.1"
-rstest = "0.18.2"
 
 [dev-dependencies]
 anyhow = "1.0.71"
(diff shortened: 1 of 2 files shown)


joshka:
A generic solution for:
https://github.com/joshka/tui-popup/blob/6c742c99362849d7d2df12361e946ec566b71cd7/src/widget.rs#L30-L34

kdheepak (approved):

src/layout/rect.rs
```diff
@@ -193,10 +193,44 @@ impl Rect {
             .try_into()
             .expect("invalid number of rects")
     }
+
+    /// Clamp the rect to fit inside the given rect.
+    ///
+    /// If the width or height of the rect is larger than the given rect, it will be clamped to
+    /// the given rect's width or height.
```
Valentin271, line 200: Looking at this it is very unclear to me which rect you're talking about. Maybe reference to them as `self` and `parameter`.

joshka: Makes sense - will change it to `this` and `other`

joshka: fixed

Valentin271 (approved):

Cargo.toml
```diff
@@ -40,6 +40,7 @@ unicode-width = "0.1"
 document-features = { version = "0.2.7", optional = true }
 lru = "0.12.0"
 stability = "0.1.1"
+rstest = "0.18.2"
```
orhun, line 43: Can we move this to `dev-dependencies`?

orhun (changes requested):

Cargo.toml
joshka: yep

orhun (approved):

PR #750, opened 2024-01-05
docs(layout): document the difference in the split methods

joshka:

### DIFF
--- a/src/layout/layout.rs
+++ b/src/layout/layout.rs
@@ -378,6 +378,11 @@ impl Layout {
     /// LruCache, and grows until [`Self::DEFAULT_CACHE_SIZE`] is reached by default, if the cache
     /// is initialized with the [Layout::init_cache()] grows until the initialized cache size.
     ///
+    /// There is a helper method on Rect that can be used to split the whole area into smaller ones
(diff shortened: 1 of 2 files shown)


src/layout/layout.rs
```diff
@@ -378,6 +378,11 @@ impl Layout {
     /// LruCache, and grows until [`Self::DEFAULT_CACHE_SIZE`] is reached by default, if the cache
     /// is initialized with the [Layout::init_cache()] grows until the initialized cache size.
     ///
+    /// There is a helper method on Rect that can be used to split the whole area into smaller ones
+    /// based on the layout: [`Rect::split()`]. That method is a shortcut for calling this method.
+    /// It allows you to destructure the result directly into variables, which is useful when you
+    /// know the number of areas that will be created.
```
kdheepak, line 384: I see it is mentioned below as well but I think it is better to clarify "at compile time" everywhere.

```suggestion
    /// know at compile time the number of areas that will be created.
```

kdheepak (approved):

src/layout/layout.rs
joshka: done

orhun (approved):

PR #751, opened 2024-01-05
fix(table)!: ensure that default and new() match

joshka:
In https://github.com/ratatui-org/ratatui/pull/660 we introduced the
segment_size field to the Table struct. However, we forgot to update
the default() implementation to match the new() implementation. This
meant that the default() implementation picked up SegmentSize::default()
instead of SegmentSize::None.

Additionally the introduction of Table::default() in an earlier PR,
https://github.com/ratatui-org/ratatui/pull/339, was also missing the
default for the column_spacing field (1).

This commit fixes the default() implementation to match the new()
implementation of these two fields by implementing the Default trait
manually.

BREAKING CHANGE: The default() implementation of Table now sets the
column_spacing field to 1 and the segment_size field to
SegmentSize::None. This will affect the rendering of a small amount of
apps.


joshka:
Affected repos:
https://github.com/search?q=ratatui%20AND%20%22Table%3A%3Adefault()%22%20AND%20(NOT%20org%3Aratatui-org)%20AND%20NOT%20(is%3Afork)&type=code

- https://github.com/Adi-df/foucault - @Adi-df
- https://github.com/SOF3/lpl - @SOF3
- https://github.com/LAPKB/NPcore - @LAPKB
- https://github.com/sarub0b0/kubetui - @sarub0b0
- https://github.com/sadovsf/RadioLog - @sadovsf
- https://github.com/asomers/gstat-rs - @asomers

Note the suggested changes above.

The impact of this PR is that the column spacing will change unless you've manually specified this, and the last column will not fill the remaining space if your width constraints do not specify to do so.

src/widgets/table/table.rs
```diff
@@ -209,6 +209,24 @@ pub struct Table<'a> {
...
+            header: Default::default(),
+            footer: Default::default(),
+            widths: Default::default(),
+            column_spacing: 1,
+            block: Default::default(),
+            style: Default::default(),
+            highlight_style: Default::default(),
+            highlight_symbol: Default::default(),
+            highlight_spacing: Default::default(),
+            segment_size: SegmentSize::None,
```
kdheepak, line 225: Does this not work?

```suggestion
            segment_size: SegmentSize::None,
            column_spacing: 1,
            ..Default::default(),
```

Adi-df:
Thanks for notifying the changes !

src/widgets/table/table.rs
joshka: No - that recursively calls the method that we're defining. I tried the same thing ;)

tranzystorekk: unfortunately this would require `Table` to have a `Default` in the first place :/

kdheepak: It would be nice to do something like this:

```rust
#[derive(Default)]
struct Foo {
    alpha: u8,
    #[default = 1]
    beta: u8,
}
```

which is listed under future possibilities in this RFC:

https://rust-lang.github.io/rfcs/3107-derive-default-enum.html#overriding-default-fields

Valentin271 (approved):

joshka:
I'd like a second reviewer on this one before merging - mainly as a quick sanity check that this breaking change makes sense.

Valentin271 (approved):
Makes sense to me, approved.

orhun (approved):
LGTM

joshka:
Resolved conflicts in BREAKING-CHANGES.md

Valentin271 (approved):

kdheepak:
It was because this PR wasn't rebased on the latest `main`. I thought GitHub's queue feature would catch this but that doesn't seem to be the case?

joshka:
Squash doesn't do a rebase, CI, then merge. it just does the rebase, merge part. The changes were textually non conflicting, just semantically.

asomers:
Thanks for notifying.  But FTR, the warning for gstat-rs is a false alarm.

PR #754, opened 2024-01-06
refactor!: make `patch_style` & `reset_style` chainable

### LINKED ISSUE #644: .patch_style() methods should be chainable (BREAKING CHANGE)
## Problem
The patch_style methods on Text/Line/Span should be chainable. Currently they all take &mut self and do not return a value.
https://github.com/search?q=ratatui+patch_style&type=code suggests that close to all the usages of these methods tend to allocate a mutable variable, then call patch style directly on that variable (there's a couple of outliers, but this holds for 95%+). Changing these methods to `fn patch_style(mut self, style: Style) -> Text` (and `Line` / `Span`) will make the code more consistent with the rest of Ratatui's builder style methods.

## Solution
- Change patch_style in `Text`, `Line` and `Span` to take `mut self` and return `Self`
- Update examples
- Update doc comments
- Update BREAKING CHANGES doc
- Ensure there is a good commit message for the changelog

Valentin271:
Previously, `patch_style` and `reset_style` in `Text`, `Line` and `Span` were using a mutable reference to `Self`. To be more consistent with the rest of `ratatui`, which is using fluent setters, these now take ownership of `Self` and return it.

Closes #644

### DIFF
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -11,6 +11,7 @@ github with a [breaking change] label.
 This is a quick summary of the sections below:
 
 - [v0.26.0 (unreleased)](#v0260-unreleased)
+  - `patch_style` & `reset_style` now consume and return `Self`
   - Removed deprecated `Block::title_on_bottom`
   - `Line` now has an extra `style` field which applies the style to the entire line
   - `Block` style methods cannot be created in a const context
@@ -44,6 +45,23 @@ This is a quick summary of the sections below:
   - `List` no longer ignores empty strings
 
 ## v0.26.0 (unreleased)
+
+### `patch_style` & `reset_style` now consumes and returns `Self` ([#754])
+
+[#754]: https://github.com/ratatui-org/ratatui/pull/754
+
+Previously, `patch_style` and `reset_style` in `Text`, `Line` and `Span` were using a mutable
+reference to `Self`. To be more consistent with the rest of `ratatui`, which is using fluent
+setters, these now take ownership of `Self` and return it.
+
+The following example shows how to migrate for `Line`, but the same applies for `Text` and `Span`.
+
+```diff
+- let mut line = Line::from("foobar");
+- line.patch_style(style);
+// becomes
++ let line = Line::new("foobar").patch_style(style);
+```
 
 ### Remove deprecated `Block::title_on_bottom` ([#757])
 
@@ -81,7 +99,7 @@ the `Span::style` field.
   let line = Line {
       spans: vec!["".into()],
       alignment: Alignment::Left,
-+     ..Default::default()    
++     ..Default::default()
   };
 
   // or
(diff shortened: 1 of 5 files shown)


kdheepak:
I use `.patch_style` on a field in struct:

```rust
struct Config {
    calendar: Style
}

fn user_config() {
  // ...
  config.calendar.patch_style(Style::default().fg(color_from_user_input))
  // ...
}
```

I wouldn't be able to do this though if `.patch_style` takes an owned value.

joshka:
> ```rust
> calendar: Style
> ```
>I wouldn't be able to do this though if .patch_style takes an owned value.

This doesn't change `Style.patch_style`, which already takes an owned value for the style parameter. Style implements Copy, so it doesn't really matter if it's owned.

I think maybe you're talking about not being able to do this on Lines/Spans that are stored?

BREAKING-CHANGES.md
```diff
@@ -11,6 +11,7 @@ github with a [breaking change] label.
 This is a quick summary of the sections below:
 
 - [v0.26.0 (unreleased)](#v0260-unreleased)
+  - `patch_style` & `reset_style` now consumes and returns `Self`
```
joshka, line 14:
```suggestion
  - `patch_style` & `reset_style` now consume and return `Self`
```

joshka (approved):

BREAKING-CHANGES.md
Valentin271: Fixed, thanks

joshka (approved):
Approved, but I want to make sure that @kdheepak's concern is addressed before merging.

src/text/line.rs
```diff
@@ -293,37 +295,47 @@ impl<'a> Line<'a> {
...
         let style = style.into();
-        for span in &mut self.spans {
-            span.patch_style(style);
-        }
+        self.spans = self
+            .spans
+            .into_iter()
+            .map(|span| span.patch_style(style))
+            .collect();
+        self
```
joshka, line 309: It's worth considering whether to change this to `self.style = self.style.patch(style)`
And likewise the same for reset, and Text patch / reset.

The impact of this is that the styles are not stored in the Span - this both a pro and con.

Valentin271: I think `patch_style` on a struct should patch the style _of this struct_ and not its internal structs.

Won't be able to do that for Text though, as you noted in https://github.com/ratatui-org/ratatui/issues/758 it doesn't have style.

joshka: yup. The grand plan is to gradually build that out so the hierarchy is right. :)

Valentin271:
Kind of messed up the rebase but its up to date now.

joshka (approved):

kdheepak:
My bad, didn't review the code changes properly. I don't store `Span` in a struct, so this wouldn't be a problem for me. I'm assuming storing `Span` and then patching it is uncommon.

joshka:
@Valentin271 the doc tests are failing.

Valentin271:
I forgot to update the tests after the `Line` update.

I'd like another review on this because I modified the examples to be more concise and better demonstrate the effect. Before that the examples were the exact same as the tests.

orhun (approved):

PR #664, opened 2023-12-04
fix(table)!: Add widths parameter to new()

joshka:
- style(tabs): fix doc formatting
- feat(layout): Accept IntoIterator for constraints
- fix(table)!: Add widths parameter to new()

PR #755, opened 2024-01-06
feat(table): Implement FromIterator for widgets::Row

Lunderberg:
The `Row::new` constructor accepts a single argument that implements `IntoIterator`.  This commit adds an implementation of `FromIterator`, as a thin wrapper around `Row::new`.  This allows `.collect::<Row>()` to be used at the end of an iterator chain, rather than wrapping the entire iterator chain in `Row::new`.

Examples in `examples/table.rs` that used an intermediate variable for the cell iterator have been updated to instead collect directly into a `Row`.  Examples that have an explicit `vec![cell1, ..., cellN]` have been kept as-is.

### DIFF
--- a/src/widgets/table/row.rs
+++ b/src/widgets/table/row.rs
@@ -45,6 +45,14 @@ use crate::prelude::*;
 /// ]);
 /// ```
 ///
+/// An iterator whose item type is convertible into [`Text`] can be collected into a row.
+///
+/// ```rust
+/// use ratatui::widgets::Row;
+///
+/// (0..10).map(|i| format!("{i}")).collect::<Row>();
+/// ```
+///
 /// `Row` implements [`Styled`] which means you can use style shorthands from the [`Stylize`] trait
 /// to set the style of the row concisely.
 ///
@@ -235,6 +243,15 @@ impl<'a> Styled for Row<'a> {
     }
 }
 
+impl<'a, Item> FromIterator<Item> for Row<'a>
+where
+    Item: Into<Cell<'a>>,
+{
+    fn from_iter<IterCells: IntoIterator<Item = Item>>(cells: IterCells) -> Self {
+        Row::new(cells)
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use std::vec;
@@ -246,6 +263,13 @@ mod tests {
     fn new() {
         let cells = vec![Cell::from("")];
         let row = Row::new(cells.clone());
+        assert_eq!(row.cells, cells);
+    }
+
+    #[test]
+    fn collect() {
+        let cells = vec![Cell::from("")];
+        let row: Row = cells.iter().cloned().collect();
         assert_eq!(row.cells, cells);
     }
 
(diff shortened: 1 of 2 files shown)


joshka (approved):
Neat! The only thing I'd suggest adding would be some docs to the `Row` type so there's some example to point people in this direction.

Happy to take this as is though if you don't want to spend the time of docs.

Valentin271 (approved):
I like that too, very good idea!

src/widgets/table/row.rs
```diff
@@ -45,6 +45,15 @@ use crate::prelude::*;
 /// ]);
 /// ```
 ///
+/// An iterator whose item type is convertible into [`Text`] can be
+/// collected into a row.
```
joshka, line 49: nit: wrap at 100
```suggestion
/// An iterator whose item type is convertible into [`Text`] can be collected into a row.
```

joshka:
> Oh and your commit needs to be signed for us to merge if that's not too much to ask.

https://github.com/ratatui-org/ratatui/blob/main/CONTRIBUTING.md#sign-your-commits has some info on this.

Lunderberg:
> The only thing I'd suggest adding would be some docs to the Row type so there's some example to point people in this direction.

Sounds good, and I've updated the docstring for `Row` to include an example, alongside the `Row::new` usage.

> Oh and your commit needs to be signed for us to merge if that's not too much to ask.

No problem, and it should now be signed.

src/widgets/table/row.rs
Lunderberg: Thank you, and updated.

joshka (approved):

joshka:
Thanks for this PR :)

joshka:
I wonder if it makes sense to do this for any other types

Lunderberg:
> I wonder if it makes sense to do this for any other types

I took a brief look at `Table` when implementing `FromIterator` for `Row`, but hadn't looked at it in detail.  From a grep for `fn new` inside the widgets directory, there's two definite cases where it would be useful, and one maybe.

* `Table`: Maybe.  Prior to https://github.com/ratatui-org/ratatui/pull/664, then `impl<'a> FromIterator<Row<'a>> for Table<'a>`.  However, there isn't a good way for `FromIterator` to receive the `widths` argument.  It could be called as `iter.collect::<Table>().widths(widths)`, but that would have the possibility of forgetting to call `.widths(widths)`.

* `Tabs`: Definitely.  The `titles: Vec<T>` argument is converted to a `Vec<Line<'a>>`.  This argument could be changed to `titles: IntoIterator<Item: Into<Line<'a>>`, and `Tabs` could implement `FromIterator<T> where T: Into<Line<'a>>`.

* `List`: Definitely.  The `List::new` method accepts `items: IntoIterator<Item: Into<ListItem<'a>>>` so there could easily be `impl<'a> FromIterator<Item> for List<'a> where Item: Into<ListItem<'a>>`.

joshka:
Cool. Do you want to raise a couple of issues for these?

Lunderberg:
@joshka Sure!  I've opened https://github.com/ratatui-org/ratatui/pull/774, https://github.com/ratatui-org/ratatui/pull/775, and https://github.com/ratatui-org/ratatui/pull/776 for the three separate updates.  There are some slight variations in how each is implemented, in order to avoid breaking types that may be inferred from `Table::new` or `Tabs::new`.

PR #759, opened 2024-01-06
refactor(list)!: remove deprecated `start_corner`

Valentin271:
`List::start_corner` was deprecated in v0.25. Use `List::direction` and `ListDirection` instead.

```diff
- list.start_corner(Corner::TopLeft);
- list.start_corner(Corner::TopRight);
// This is not an error, BottomRight rendered top to bottom previously
- list.start_corner(Corner::BottomRight);
// all becomes
+ list.direction(ListDirection::TopToBottom);
```

```diff
- list.start_corner(Corner::BottomLeft);
// becomes
+ list.direction(ListDirection::BottomToTop);
```


joshka:
Perhaps we could let this one live for a minute? There's not a huge amount of users, but 0.25 is the most recent release. 
https://github.com/search?q=ratatui+start_corner+AND+%28NOT+org%3Aratatui-org%29+AND+%28NOT+is%3Afork%29&type=code

Valentin271:
I think one version is enough for an easy fix like that. But I don't mind holding it for a while.

Let's see what the others have to say.

Maybe we should state in the contributing guide how much we want to wait before removing deprecated items.

Valentin271:
To add to my previous comment:

When the deprecated feature was clear and didn't have problem in itself, I agree to keep it a few version.
But when it's something that wasn't clear how to use or its effects, I think basically people won't miss it because they did not understand what it was in the first place.

joshka:
I think 2 releases is a good idea generally. This gives our users enough time to update to a release, test on that release and have users test it for a bit. We're releasing Ratatui about every 6-8 weeks, so a 2 version deprecation gives 3 months notice of breaking changes.

There might be a few changes that we want to move users away from more forcefully so we can implement a new feature where the 2 release approach doesn't work, but for things like this, where the downside of keeping it is only that the API is messy, I'm fine with it.

^Weak opinion - if you do feel more strongly than I do about removing this sooner, then I'm happy to support that.

Valentin271:
Weak opinion for me too, It's literally that I don't like keeping a messy API for too long.

So conclusion for me:

- If a deprecation is blocking for us to implement a new feature we may *consider* removing it in a one version notice
- Otherwise if the deprecation is not blocking for anything, wait at least 2 version

Valentin271:
Updated CONTRIBUTING.md

https://github.com/ratatui-org/ratatui/pull/761

orhun:
+1 from me, I think having a 2 release notice is good enough for removing deprecated items. Let's also add these types of issues to the future milestones (which I saw it is already added!).

joshka:
Let's merge this right before 0.27.0 releases to avoid putting breaking changes in the repo right after release. Marking it draft until then.

Valentin271:
Rebased on main. It is unlikely we will release 0.26.2 in my opinion.

joshka:
> Rebased on main. It is unlikely we will release 0.26.2 in my opinion.

Probably not, but it's also not a huge thing to hold back from merging. Merging this means that we can't do 26.2 if needed.
I'd still caution about merging this until we're closer to a major release.

Valentin271:
Works for me, was easy to rebase and probably still will be later on. We can wait.

joshka (approved):

EdJoPaTo:
Is Corner used somewhere else? Otherwise, it can be removed too.

Issue #752, opened 2024-01-05
Add Rows and Columns iterators to Rect


joshka:
## Problem
Often we write code in rendering functions that does something for each row or each column.
These tend to use some sort of index to iterate over the bounds of the area.  E.g:

https://github.com/ratatui-org/ratatui/blob/f13fd73d9ec108af723a9cd11f4262f2b09c9d25/src/widgets/clear.rs#L27-L35

Many of these sorts of things would benefit from iterating over a Rect as the Rows or Columns instead.

```rust
for row in area.rows() {
  for cell in row.columns() {
    buf.get_mut(cell.x, cell.y).reset();
  }
}
```

Additionally, by returning iterators, these can be composed using th iterator operations (like zip, merge, chain etc.)

E.g.:

```
self.lines.iter()
  .zip(area.rows())
  .foreach(|(line, row)| line.render(row, &mut buf));
```

## Solution
1. Implement `rows()` and `columns()` method on `Rect`. These should each return a custom iterator. I think these might be called `Rects`, though there is probably a benefit to calling these `Rows` and `Columns` instead as I suspect it will be easier to keep state smaller with two separate iterator types.
2. Add Unit tests
3. Add Docs
4. Update widget code to use these methods to prove them out.
5. Update any examples where these might be useful.

## Additional context
Marked as a good first issue. Happy to provide any guidance necessary on this one.

BogdanPaul15:
Hello! I would like to work on this issue. Do I have to create a struct called Iterator and add some logic to iterate over Columns and Rows structs, and then add rows() and columns() method to the impl of Rect?

joshka:
> Hello! I would like to work on this issue.

Welcome. I've assigned this to you.

> Do I have to implement an custom Iterator and add next() to iterate over Columns and Rows structs, and then add rows() and columns() method to the impl of Rect?

Yes

> Should I make a separate file? Or work in the rect.rs

I expect that each iterator is probably not more than about 15 LoC + tests and docs, so adding these to rect.rs is probably the right place to put the code.

BogdanPaul15:
Hi! I am encountering a problem when I try to push.
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.

PR #765, opened 2024-01-07
feat(rect.rs): implement Rows and Columns iterators in Rect

BogdanPaul15:
Signed-off-by: Prisacaru Bogdan-Paul <paulbp68@gmail.com>

#752

This PR will allow users to use iterators for Rows and Columns in Rects in order to manage rows and columns. During iteration, each structure produces `Rect` instances representing each row and column within the given Rect.

### DIFF
--- a/src/layout/rect.rs
+++ b/src/layout/rect.rs
@@ -25,6 +25,58 @@ pub struct Rect {
     pub height: u16,
 }
 
+/// Manages row divisions within a `Rect`.
+///
+/// The `Rows` struct is an iterator that allows iterating through rows of a given `Rect`.
+pub struct Rows {
+    /// The `Rect` associated with the rows.
+    pub rect: Rect,
+    /// The y coordinate of the row within the `Rect`.
+    pub current_row: u16,
+}
+
+impl Iterator for Rows {
+    type Item = Rect;
+
+    /// Retrieves the next row within the `Rect`.
+    ///
+    /// Returns `None` when there are no more rows to iterate through.
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.current_row >= self.rect.bottom() {
+            return None;
+        }
+        let row = Rect::new(self.rect.x, self.current_row, self.rect.width, 1);
+        self.current_row += 1;
+        Some(row)
+    }
+}
+
+/// Manages column divisions within a `Rect`.
+///
+/// The `Columns` struct is an iterator that allows iterating through columns of a given `Rect`.
+pub struct Columns {
+    /// The `Rect` associated with the columns.
+    pub rect: Rect,
+    /// The x coordinate of the column within the `Rect`.
+    pub current_column: u16,
+}
+
+impl Iterator for Columns {
+    type Item = Rect;
+
+    /// Retrieves the next column within the `Rect`.
+    ///
+    /// Returns `None` when there are no more columns to iterate through.
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.current_column >= self.rect.right() {
+            return None;
+        }
+        let column = Rect::new(self.current_column, self.rect.y, 1, self.rect.height);
+        self.current_column += 1;
+        Some(column)
+    }
+}
+
 impl fmt::Display for Rect {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         write!(f, "{}x{}+{}+{}", self.width, self.height, self.x, self.y)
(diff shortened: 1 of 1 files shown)


Valentin271 (commented):
I'm not sure this is what joshka meant when creating the issue.

I think you wanted to do the following right?
```diff
- for x in area.left()..area.right() { 
+ for x in area.columns() { 
-   for y in area.top()..area.bottom() { 
+   for y in area.rows() { 
    buf.get_mut(x, y).reset(); 
  } 
} 
```

The way I understood the issue, the iterators were supposed to return `Rect`s representing the rows and column. So we could work on that and, for example, get the columns of a row (so a `Rect`).

```rust
let area = Rect(0, 0, 10, 5);
for row in area.rows() {
  // row is Rect { x: 0, y: 0, width: 10, height: 1}, note height of one because it's a row
  // on second iteration this will become Rect { x: 0, y: 1, width: 10, height: 1}
  // up to Rect { x: 0, y: 9, width: 10, height: 1}, y of 9 because there is 10 rows starting from 0

  // and then here because row is a `Rect` we can call `columns` again
  for cell in row.columns() {
    // which will give Rect { x: 0, y: 0, width: 1, height: 1}, width is one because it's a columns, height is one because we iterate over a row
    // then Rect { x: 1, y: 0, width: 1, height: 1}, incrementing x up to 9
    // and at the end Rect { x: 9, y: 9, width: 1, height: 1}, the last cell on the bottom right

    // here we can call rect.x and rect.y because the resulting rect, when iterating over rows and columns, is 1 by 1, so a cell
    buf.get_mut(cell.x, cell.y).reset();
  }
}
```

That was my understanding, let's wait for @joshka's input

BogdanPaul15:
I understand your point. I will wait for @joshka. Thank you!

joshka (changes requested):
@Valentin271 was right. I'd expect each element of the iterator to be a single height or single width Rect.

For the tests, simplify them to compare rows().collect() to a hard coded vec of rects. This will visually show that the x or y index is changing, but the width or height is always 1.

BogdanPaul15:
From my understanding, this is what you needed. Please let me know if I am wrong.

src/layout/rect.rs
```diff
@@ -25,6 +25,62 @@ pub struct Rect {
...
+/// Manages row divisions within a `Rect`.
+///
+/// The `Rows` struct is an iterator that allows iterating through rows of a given `Rect`.
+pub struct Rows {
+    /// The `Rect` associated with the rows.
+    pub rect: Rect,
+    /// The y coordinate of the row within the `Rect`.
+    pub current_row: u16,
+    /// The x coordinate of the row within the `Rect`.
+    pub current_column: u16,
```
Valentin271, line 37: You don't need the current column in the row iterator.

src/layout/rect.rs
```diff
@@ -25,6 +25,62 @@ pub struct Rect {
...
+}
+
+/// Manages column divisions within a `Rect`.
+///
+/// The `Columns` struct is an iterator that allows iterating through columns of a given `Rect`.
+pub struct Columns {
+    /// The `Rect` associated with the columns.
+    pub rect: Rect,
+    /// The y coordinate of the column within the `Rect`.
+    pub current_row: u16,
```
Valentin271, line 63: Same here you don't need the current_row in the columns iterator.

src/layout/rect.rs
```diff
@@ -25,6 +25,62 @@ pub struct Rect {
...
+}
+
+impl Iterator for Rows {
+    type Item = Rect;
+
+    /// Retrieves the next row within the `Rect`.
+    ///
+    /// Returns `None` when there are no more rows to iterate through.
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.current_row >= self.rect.y + self.rect.height {
```
Valentin271, line 47: You can use `Rect::bottom()` here to make this easier to maintain and avoid panics.
```suggestion
        if self.current_row >= self.rect.bottom() {
```

src/layout/rect.rs
```diff
@@ -25,6 +25,62 @@ pub struct Rect {
...
+}
+
+impl Iterator for Columns {
+    type Item = Rect;
+
+    /// Retrieves the next column within the `Rect`.
+    ///
+    /// Returns `None` when there are no more columns to iterate through.
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.current_column >= self.rect.x + self.rect.width {
```
Valentin271, line 75: Similarly, you can use `Rect::right()` here.
```suggestion
        if self.current_column >= self.rect.right() {
```

Valentin271 (changes requested):
Other than my few comments looks good.

Valentin271:
It looks like you tried to sign your commit (as shows the PR description), but the github interface doesn't show it, could check that?

You have some information here https://github.com/ratatui-org/ratatui/blob/main/CONTRIBUTING.md#sign-your-commits.

If you need help feel free to ask.

BogdanPaul15:
@Valentin271 thank you for your feedback. Should I commit these changes? Also, yes, I've tried to sign my commits, my machine is already configured with gpg keys, but I don't know why it doesn't work.

Valentin271:
Have you looked at https://commitizen-tools.github.io/commitizen/getting_started/#sign-off-the-commit? (weird though I don't remember having to do that when I used it)

You can either commit the proposed changes from the interface (will be committed as me though) or commit them yourself as usual, as you wish.

Valentin271 (approved):
Looks all good, code and signing. Thanks!

I'll wait for a second approval on this.

joshka (approved):
Looks great.

joshka:
To get this to merge, all the commits need to be signed, not just the last one. Can you please squash the commits locally into a single commit and force push to your branch?

image

Thanks for the PR :)

BogdanPaul15:
Ok, I'm trying to squash into a single commit and force push.

BogdanPaul15:
I need some help. When I try to squash my commits, I am getting others commits than mine. Do I have to squash them too?

Valentin271:
I'd suggest

```sh
// this will go back to your latest relevant commit
git reset --hard d9195580cc9eb2c941f66e34bb4630eee33d14c3

git rebase -i HEAD~4
// replace pick with squash for the last 3 commits, see attached screenshot 
// then the commit message will show up, edit as needed

// just to save the changes
git push -f --no-verify
```

Then sign your commit and force push.
If you can't/don't know how to sign an already existing commit, you can

```sh
// deletes the commit and put the changes back to staged
git reset --soft HEAD^
// then commit+sign as usual
// and force push
git push -f --no-verify
```

Might not be the most straightforward way but at least it should work.

---

`rebase -i` interface
image

BogdanPaul15:
Thank you. I really don't lnow why this failed.

Valentin271:
You're welcome, and thanks for your contribution!

PR #766, opened 2024-01-07
feat: Add frame count ✨

### LINKED ISSUE #764: Add tick counter to `Frame`
## Problem

Every time `Frame` is constructed, it could increment a field `frame_tick` or a `frame_count` variable. This will make it easier to do animations "every 4 frames". This should be a read only field, exposed using a getter function.

`frame_count` could be a `usize` and could roll over to the beginning when it hits it's maximum value.

kdheepak:
Resolves #764

### DIFF
--- a/src/terminal/terminal.rs
+++ b/src/terminal/terminal.rs
@@ -71,6 +71,8 @@ where
     /// Last known position of the cursor. Used to find the new area when the viewport is inlined
     /// and the terminal resized.
     last_known_cursor_pos: (u16, u16),
+    /// Number of frames rendered up until current time.
+    frame_count: u64,
 }
 
 /// Options to pass to [`Terminal::with_options`]
@@ -149,15 +151,18 @@ where
             viewport_area,
             last_known_size: size,
             last_known_cursor_pos: cursor_pos,
+            frame_count: 0,
         })
     }
 
     /// Get a Frame object which provides a consistent view into the terminal state for rendering.
     pub fn get_frame(&mut self) -> Frame {
+        let count = self.frame_count.wrapping_add(1);
         Frame {
             cursor_position: None,
             viewport_area: self.viewport_area,
             buffer: self.current_buffer_mut(),
+            count,
         }
     }
 
@@ -279,9 +284,12 @@ where
         // Flush
         self.backend.flush()?;
 
+        self.frame_count = self.frame_count.wrapping_add(1);
+
         Ok(CompletedFrame {
             buffer: &self.buffers[1 - self.current],
             area: self.last_known_size,
+            count: self.frame_count,
         })
     }
 
--- a/src/terminal/terminal.rs
+++ b/src/terminal/terminal.rs
@@ -157,7 +157,7 @@ where
 
     /// Get a Frame object which provides a consistent view into the terminal state for rendering.
     pub fn get_frame(&mut self) -> Frame {
-        let count = self.frame_count.wrapping_add(1);
+        let count = self.frame_count;
         Frame {
             cursor_position: None,
             viewport_area: self.viewport_area,
@@ -261,6 +261,9 @@ where
         // and the terminal (if growing), which may OOB.
         self.autoresize()?;
 
+        // increment frame count before getting frame
+        self.frame_count = self.frame_count.wrapping_add(1);
+
         let mut frame = self.get_frame();
         f(&mut frame);
         // We can't change the cursor position right away because we have to flush the frame to
@@ -283,8 +286,6 @@ where
 
         // Flush
         self.backend.flush()?;
-
-        self.frame_count = self.frame_count.wrapping_add(1);
 
         Ok(CompletedFrame {
             buffer: &self.buffers[1 - self.current],
--- a/src/terminal/terminal.rs
+++ b/src/terminal/terminal.rs
@@ -72,7 +72,7 @@ where
     /// and the terminal resized.
     last_known_cursor_pos: (u16, u16),
     /// Number of frames rendered up until current time.
-    frame_count: u64,
+    frame_count: usize,
 }
 
 /// Options to pass to [`Terminal::with_options`]
--- a/src/terminal/terminal.rs
+++ b/src/terminal/terminal.rs
@@ -261,9 +261,6 @@ where
         // and the terminal (if growing), which may OOB.
         self.autoresize()?;
 
-        // increment frame count before getting frame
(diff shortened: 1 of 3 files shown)


src/terminal/terminal.rs
```diff
@@ -71,6 +71,8 @@ where
     /// Last known position of the cursor. Used to find the new area when the viewport is inlined
     /// and the terminal resized.
     last_known_cursor_pos: (u16, u16),
+    /// Number of frames rendered up until current time.
+    frame_count: u64,
```
kdheepak, line 75: Should we use `u64`, `u32` or `usize` here? 

I think it is should be `usize` so it is more ergonomic to do `f.count() / vector.len()`

Valentin271: Yep let's use the most convenient type imo

kdheepak: I changed it to `usize` since that is still plenty of time even at 60 fps.

image

Valentin271: Just thought of that, this is going to be u16 on 16bits devices (~18mins). Is it a problem?
Probably not, it won't crash, just go back to zero.

kdheepak: I think it is not a problem since I expect people using it will write code assuming it rolls over. It is also documented to roll over. I think it is worth using `usize` just for the ergonomic benefits since I expect most commonly they are going to doing modulo division.

Valentin271: That's what I thought, seems reasonable.

Valentin271 (approved):
Looks all good.

src/terminal/terminal.rs
```diff
@@ -149,15 +151,18 @@ where
             viewport_area,
             last_known_size: size,
             last_known_cursor_pos: cursor_pos,
+            frame_count: 0,
         })
     }
 
     /// Get a Frame object which provides a consistent view into the terminal state for rendering.
     pub fn get_frame(&mut self) -> Frame {
+        let count = self.frame_count.wrapping_add(1);
```
joshka, line 160: It looks like you're incrementing twice. Here and in flush. Am I mistaken?

kdheepak: Yes, but this adds and stores the result in the `Frame` struct which is passed to the `FnOnce` callback the user provides. The second increments the field in the `Terminal` struct.

I can combine the two by doing `self.frame_count = self.frame_count.wrapping_add(1)` just in this function.

kdheepak: Simplified implementation in https://github.com/ratatui-org/ratatui/pull/766/commits/842ef80fb7acca1f8660b886864d70ac649507f5.

src/terminal/terminal.rs
```diff
@@ -256,6 +261,9 @@ where
         // and the terminal (if growing), which may OOB.
         self.autoresize()?;
 
+        // increment frame count before getting frame
+        self.frame_count = self.frame_count.wrapping_add(1);
+
```
joshka, line 266: Would it make sense to increment after rendering? so the first frame is `0`?

joshka (approved):

src/terminal/terminal.rs
kdheepak: The intended the first frame to be 1 since I wanted to include the count of the current render into the frame count.

```
    /// This method provides access to the frame count, which is a sequence number indicating
    /// how many frames have been rendered up to and including this one. 
```

When a user calls `f.count()` in the first frame, I thought it made more sense to return `1` then. 

Also, if we increment after rendering, the `f.count()` will be `0` but `CompletedFrame`, i.e. `let completed_frame = terminal.draw(|f| {...});` will have `completed_frame.count` be `1`.

With this implementation both `f.count()` and `completed_frame.count` will be the same.

Thoughts?

joshka: The second part of this is a non-issue as it's simple to avoid. Create the `CompletedFrame` and store it in a local variable before incrementing the value and then return the completed frame. So only the first part is relevant here.

In my mind I kinda have this idea that it's the index of the frame, not the number of frames rendered. I don't think either is the incorrect perspective. Though perhaps calling it index instead of count might make sense.

Where it might make (a very small) difference is in animations - if you have an animation that starts from frame 0 at the beginning of a program, then you'd need to subtract 1 from the frame count. Not sure if there are other calculations that would potentially have off by one bugs, so I guess it's no big deal.

kdheepak: Should we rename it to frame index?

joshka: I defer to you on this one. Either is fine and so I can't decide. You've go the use cases in mind for it though.

kdheepak: I like starting at `0` because you'll the first element in the array when you use `f.count() % vector.len()`. I guess what's the best name for the variable and the method on the `Frame` struct for this.

I asked chatgpt for alternative name suggestions.

```
When considering alternative names for the count field in the Frame struct, it's important to choose a name that accurately reflects its purpose and usage in the context of frame rendering or tracking. Here are some alternative names that might be suitable, depending on the exact role and functionality of this field:

    frameNumber: This name clearly indicates that the field represents the sequence number of the current frame.

    renderCount: This name suggests that the field tracks the number of times rendering has occurred.

    sequenceIndex: This implies that the field is an index in a sequence, which in this case is the sequence of rendered frames.

    frameSequence: Similar to sequenceIndex, this indicates a sequential ordering of frames.

    renderSequence: This name emphasizes the rendering process in the sequence tracking.

    frameIndex: This straightforwardly indicates an index value for frames.

    iterationCount: If the field is used to count iterations of a rendering loop, this name would be appropriate.

    updateCount: This could be used if the field is incremented every time the frame is updated.

    frameId: This suggests a unique identifier for each frame, though 'ID' often implies a broader uniqueness than a simple count.

    renderIteration: This name combines the rendering process with the notion of iterative counts.
```

I like `f.count()` more than `f.number()` or `f.index()`.

joshka: I'd be inclined to keep frame in there possibly (`frame_number`)? Frame number is pretty common when dealing with video / audio frames.

joshka: Alternatively just accept that count is the count of frames previous to this one.

kdheepak: I left the name as frame count but changed it to start from zero. I've updated the tests and the docstring too.

joshka (approved):

kdheepak:
Thanks for the reviews!

PR #767, opened 2024-01-07
chore(deps): update termion to v3.0

Valentin271:
This PR updates termion to its latest version. There is only one breaking change https://gitlab.redox-os.org/redox-os/termion#200-to-300-guide.

We don't really use this but our users do in when initializing the terminal. I'm not sure then if this is a breaking change. I'd say no because users can probably still specify termion 2 right?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ rust-version = "1.70.0"
 
 [dependencies]
 crossterm = { version = "0.27", optional = true }
-termion = { version = "2.0", optional = true }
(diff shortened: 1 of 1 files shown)


joshka:
Perhaps link to the changes in the commit to make it easy for users.

joshka (approved):

orhun (approved):

Valentin271:
> Perhaps link to the changes in the commit to make it easy for users.

Good idea, done.

Issue #763, opened 2024-01-07
Support hsluv color constructor


kdheepak:
## Problem

This will allows developers to define more uniform color palettes like so:

```rust
let GRAY = [
    Color::from_hsluv(235.5, 22.1, 98.2),
    Color::from_hsluv(248.2, 18.5, 96.2),
    Color::from_hsluv(248.3, 16.7, 91.6),
    Color::from_hsluv(243.9, 14.4, 85.1),
    Color::from_hsluv(246.8, 12.5, 66.8),
    Color::from_hsluv(249.7, 15.7, 47.9),
    Color::from_hsluv(245, 25.3, 35.8),
    Color::from_hsluv(248.1, 32.2, 27.3),
    Color::from_hsluv(247.2, 42, 16.3),
    Color::from_hsluv(254, 48.3, 8.3),
]
```

We could use palette as a dependency to handle the conversions

https://crates.io/crates/palette

## Solution

This can be implemented as a constructor on `Color`:

```rust
impl Color {
    fn from_hsluv(h: f64, s: f64, l: f64, a: f64) -> Self {
        let h = h.clamp(0.0, 360.0);
        let s = s.clamp(0.0, 100.0);
        let l = l.clamp(0.0, 100.0);
        let a = a.clamp(0.0, 100.0);
        // TODO: convert to RGB
        Self::Rgb(r, g, b)
    }
}
```

## Alternatives

I'm open to other suggestions for how to enable such a feature.

Valentin271:
I like this idea. But I'd prefer to avoid using palette (or any other deps) just for that.

I'd suggest putting this behind a `palette` feature. I believe most people will still use rgb.

If not to hard, we could also implement it by hand, but didn't really find a clear formula so maybe that's too much.

kdheepak:
My understanding is that it is a series of transformations to go from hsluv to rgb, and I'm assuming we wouldn't want to maintain that.

joshka:
Definitely a feature flash. I recall seeing a create that does more generic conversions like this. I'll find a link later today.

orhun:
I also think that this is a nice feature and we can gate it behind a feature flag for avoiding additional dependencies as default.

joshka:
https://github.com/Canop/coolor was the crate I was thinking of. It's not quite what we're thinking of here though.
I mentioned perhaps doing a Ratatui feature for it in https://github.com/Canop/coolor/issues/2, but never got back to that.

Are there any other crates that already do the conversions between libraries?

PR #772, opened 2024-01-09
feat: Add `Color::from_hsl` ✨

kdheepak:
This PR adds `Color::from_hsl` that returns a valid `Color::Rgb`. 

```rust
let color: Color = Color::from_hsl(360.0, 100.0, 100.0);
assert_eq!(color, Color::Rgb(255, 255, 255));

let color: Color = Color::from_hsl(0.0, 0.0, 0.0);
assert_eq!(color, Color::Rgb(0, 0, 0));
```

HSL stands for Hue (0-360 deg), Saturation (0-100%), and Lightness (0-100%) and working with HSL the values can be more intuitive. For example, if you want to make a red color more orange, you can change the Hue closer toward yellow on the color wheel (i.e. increase the Hue).

Related #763


joshka:
Can we import an optional library for this functionality instead?

orhun (approved):

orhun:
I think the logic/conversion is pretty simple and it does not pose a huge maintenance burden (and it is not going to change anytime soon - `hsl-rs` was updated 5 years ago). On the other hand, if there are more benefits to using an optional library (such as more functionality / faster code) then I'm up for it.

kdheepak:
If we import a library like https://crates.io/crates/palette we can convert from any color space to RGB and it would be worth it in my opinion. HSLuv is the main one that people recommend for perceptually uniform color spaces.

https://www.hsluv.org/comparison/

If we are only going to stick to HSL, I agree with @orhun, this implementation is standard and there's little to no maintenance cost.

joshka:
Are we just going to stick with HSL? Why that over HSV or some other approach?

kdheepak:
My guess for why HSL is more popular than HSB/HSV is that in HSB, 100% Brightness/Value gives you the white color only when the Saturation is 0. But in HSL, 100% Lightness will give you white independent of the value of Saturation.

So HSL is more intuitive, since you can increase Lightness to get closer to white and decrease Lightness to get closer to black. With HSB/HSV you have to change two things, and in opposite directions.

joshka:
I guess my point is that there are already so many versions of this sort of function in various crates:
- https://docs.rs/sorbet-color/0.3.1/sorbet_color/
- https://crates.io/crates/colortypes
- https://crates.io/crates/palette
- etc.

No problem with merging this as-is though.

joshka:
- https://crates.io/crates/colorsys/ seems to have a pretty decent API

kdheepak:
I don't see people needing to use cmyk (because it is primarily used for print and not digital). Also, for terminals, the `alpha` property is pointless. 

imo only `HSL` and `HSLuv` (and maybe other equivalent perceptually uniform) color spaces are probably the most useful to have in the library.

joshka:
Makes sense.

PR #773, opened 2024-01-09
chore: Updated Layout in tabs.rs example

Libroru:
I updated the tabs.rs example to use the new syntax for tabs.
Other examples might need to be changed as well.

### DIFF
--- a/examples/tabs.rs
+++ b/examples/tabs.rs
@@ -80,8 +80,9 @@ fn run_app<B: Backend>(terminal: &mut Terminal<B>, mut app: App) -> io::Result<(
 
 fn ui(f: &mut Frame, app: &App) {
     let area = f.size();
-    let vertical = Layout::vertical([Constraint::Length(3), Constraint::Min(0)]);
-    let [tabs_area, inner_area] = area.split(&vertical);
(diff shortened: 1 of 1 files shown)


kdheepak:
This PR changes it from the new to the old syntax. Can you clarify where you read about this? If anything has to be changed to the new syntax, it is probably our documentation on the website.

joshka:
> If anything has to be changed to the new syntax, it is probably our documentation on the website.

Ah, that would make sense. I didn't think about how this sort of thing would affect users when updating the examples. Oops.

Libroru:
> This PR changes it from the new to the old syntax. Can you clarify where you read about this? If anything has to be changed to the new syntax, it is probably our documentation on the website.

That's how I got it working in my personal projects with version `0.25.0`. But yeah in addition to that the documentation [also shows the old syntax](https://docs.rs/ratatui/latest/ratatui/layout/struct.Layout.html#example). I figured that that would be the new syntax since it worked on a local installation of `0.25.0`.

joshka:
I'm going to close this PR out, and move it to an issue instead.

joshka:
See https://github.com/ratatui-org/ratatui/issues/779

PR #774, opened 2024-01-10
feat(widgets): Collect iterator of `Row` into `Table`

Lunderberg:
A follow-up from https://github.com/ratatui-org/ratatui/pull/755, allowing any iterator whose item is convertible into `Row` to be collected into a `Table`.

### DIFF
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -16,6 +16,7 @@ This is a quick summary of the sections below:
   - `Line` now has an extra `style` field which applies the style to the entire line
   - `Block` style methods cannot be created in a const context
   - `Tabs::new()` now accepts `IntoIterator<Item: Into<Line<'a>>>`
+  - `Table::new` now accepts `IntoIterator<Item: Into<Row<'a>>>`.
 - [v0.25.0](#v0250)
   - Removed `Axis::title_style` and `Buffer::set_background`
   - `List::new()` now accepts `IntoIterator<Item = Into<ListItem<'a>>>`
@@ -47,20 +48,37 @@ This is a quick summary of the sections below:
 
 ## v0.26.0 (unreleased)
 
+### `Table::new()` now accepts `IntoIterator<Item: Into<Row<'a>>>` ([#774])
+
+[#774]: https://github.com/ratatui-org/ratatui/pull/774
+
+Previously, `Table::new()` accepted `IntoIterator<Item=Row<'a>>`.  The argument change to
+`IntoIterator<Item: Into<Row<'a>>>`, This allows more flexible types from calling scopes, though it
+can some break type inference in the calling scope for empty containers.
+
+This can be resolved either by providing an explicit type (e.g. `Vec::<Row>::new()`), or by using
+`Table::default()`.
+
+```diff
+- let table = Table::new(vec![], widths);
+// becomes
++ let table = Table::default().widths(widths);
+```
+
 ### `Tabs::new()` now accepts `IntoIterator<Item: Into<Line<'a>>>` ([#776])
 
 [#776]: https://github.com/ratatui-org/ratatui/pull/776
 
 Previously, `Tabs::new()` accepted `Vec<T>` where `T: Into<Line<'a>>`.  This allows more flexible
-types from calling scopes, though it can break type inference when the calling scope.
+types from calling scopes, though it can break some type inference in the calling scope.
 
 This typically occurs when collecting an iterator prior to calling `Tabs::new`, and can be resolved
 by removing the call to `.collect()`.
 
 ```diff
-- let table = Tabs::new((0.3).map(|i| format!("{i}")).collect());
+- let tabs = Tabs::new((0.3).map(|i| format!("{i}")).collect());
 // becomes
-+ let table = Tabs::new((0.3).map(|i| format!("{i}")));
++ let tabs = Tabs::new((0.3).map(|i| format!("{i}")));
 ```
 
 ### Table::default() now sets segment_size to None and column_spacing to ([#751])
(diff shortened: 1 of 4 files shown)


src/widgets/table/table.rs
```diff
@@ -799,6 +821,19 @@ impl<'a> Styled for Table<'a> {
...
 }
 
+impl<'a, Item> FromIterator<Item> for Table<'a>
+where
+    Item: Into<Row<'a>>,
+{
+    fn from_iter<Iter: IntoIterator<Item = Item>>(rows: Iter) -> Self {
+        // When collecting from an iterator into a table, the user
+        // must provide the widths using `Table::widths` after
+        // construction.
```
Valentin271, line 831: This is great, what about moving it as doc comment of the `from_iter` function?

src/widgets/table/table.rs
```diff
@@ -133,6 +133,23 @@ use crate::{
 /// Cell::from(Text::from("text"));
 /// ```
 ///
+/// Just as rows can be collected from iterators of `Cell`s, tables can be collected from iterators
+/// of `Row`s.  Note: While `Table::new` requires the column widths to be provided on construction,
+/// collection of an iterator into a `Table` does not have this safeguard.  If the column widths are
+/// not set with the `Table::widths`, the table will appear empty when drawn.
```
Valentin271, line 139: I think this is no longer true, tables without width sould have columns of equal widths now.

Valentin271 (changes requested):
LGTM, just had a few comments. Only the table widths one needs to be checked and addressed if needed. (On mobile rn, hard to check for me)

src/widgets/table/table.rs
joshka: Also - wrap at 100 chars please for consistency

src/widgets/table/table.rs
Lunderberg: Ooh, nice!  I added this caveat since it was mentioned in #664.  Since #710 resolved the issue entirely, I'll remove this comment.

I'm wondering if #664 should be reverted entirely, now that the failure mode it guarded against is no longer present.  It would be a breaking change from 0.25 to 0.26, but would remove a breaking change from 0.24 to 0.26.

Valentin271: I think we should keep it because the equal widths thing is just a failsafe. Most of the time you'll want to specify widths. Moreover, `default` already provide a behavior where widths are not set.

joshka: I like having `new()` push users to provide the values for the most common use case for a type, so I want 664 to stay too.

src/widgets/table/table.rs
Lunderberg: Good call, and moved it to the docstring, wrapped at 100 chars.

src/widgets/table/table.rs
```diff
@@ -233,6 +250,11 @@ impl<'a> Table<'a> {
     /// ```
     pub fn new<R, C>(rows: R, widths: C) -> Self
     where
+        // The `Table::new` function could be made more flexible by accepting `R: IntoIterator`,
+        // with `R::Item: Into<Row<'a>>`.  This may be desirable at some point, but would be a
+        // breaking change, as type inference could no longer use `Table::new` to infer the type of
+        // elements of `R`.  For example, `Table::new(vec![], [Constraint::Min(10)])` would not
+        // compile, because the type of `vec![]` would be unknown.
```
joshka, line 257: I think we might want to break the empty vec use case. I don't recall why we didn't already.
The caller can use `Table::default().widths([Constraint::Min(10)])` for that.

This would imply an updated commit message and an update in the breaking changes doc.

src/widgets/table/table.rs
Lunderberg: Good point, and that makes sense for guiding users to good practices.

src/widgets/table/table.rs
Lunderberg: Agreed.  The empty vec/array case seems like it is primarily used in the test cases, but an empty table wouldn't show up as often in actual usage.

Updated here, and in the breaking changes doc.

src/widgets/table/table.rs
```diff
@@ -133,6 +133,22 @@ use crate::{
...
+/// ```rust
+/// use ratatui::{prelude::*, widgets::*};
+///
+/// let text = "Mary had a\nlittle lamb.";
+///
+/// let table = text
+///     .split("\n")
+///     .map(|line: &str| -> Row { line.split_ascii_whitespace().collect() })
+///     .collect::<Table>()
+///     .widths([Constraint::Length(10); 3]);
```
joshka, line 149: Good example!

BREAKING-CHANGES.md
```diff
@@ -15,6 +15,7 @@ This is a quick summary of the sections below:
   - Removed deprecated `Block::title_on_bottom`
   - `Line` now has an extra `style` field which applies the style to the entire line
   - `Block` style methods cannot be created in a const context
+  - `Table::new` now accepts `IntoIterator<Item: Into<Row<'a>>>`.
```
joshka, line 18: Also need to add a full description of this below (around line 49)

joshka (changes requested):
LGTM - just missing the full breaking change description.

commit message for the changelog should probably be something like

```
feat(table)!: accept IntoItertor<Into<Row>> for constructors

Table::new() now accepts ...
Table::from_iterator() ...

BREAKING CHANGE:
Calls to Table::new() passing rows as as an infered type
(e.g. `vec![]`) will no longer compile because the compiler
cannot to infer the type. (**reword this a bit better**)
```

BREAKING-CHANGES.md
Lunderberg: Can do, and updated!

joshka (approved):
LGTM

joshka:
Thanks again for the PR

PR #775, opened 2024-01-10
feat(widgets): Collect iterator of `ListItem` into `List`

Lunderberg:
A follow-up from https://github.com/ratatui-org/ratatui/pull/755, allowing any iterator whose item is convertible into `ListItem` to be collected into a `List`.


Valentin271 (approved):
Lgtm

joshka (approved):

Issue #619, opened 2023-11-08
please add getters to all settable things


pm100:
## Problem

My immediate example is that I need to look inside a `Block` being passed to me (I am a library that a tui app uses). I need to read the `border_style` and `title`. But I think this is a general point, especially for a library that adds tui features  (tui_textarea in my case)

It would be nice if rust had c# style properties, but it does not, so this is a lot of sloggy typing

## Solution

Add `get_xxx` fn to every struct that has an `xxx` fn that sets the xxx property.

joshka:
This has come up before (https://github.com/ratatui-org/ratatui/pull/453 and https://github.com/ratatui-org/ratatui/discussions/78 are worth a quick read). I think this use case makes a much better argument for why we need to add getters than for unit testing. I'm not sure whether to consolidate this into the discussion on #78 / #453 or whether to treat it as a new thing do you think your request is enough alike #78 to be useful there - if so, let's copy it over and close this?

There aren't C# style getters in rust, mainly because idiomatically rust types that need them tend to just make the fields public - which might be the right option here (particularly if we mark the types non-exhaustive). The alternatives are crates that macro up getters like https://crates.io/crates/derive-getters or https://crates.io/crates/getset

pm100:
#78 is much larger in scope so, no, please don't merge this. This one is fundamentally simple

Would you accept a PR using derive-getters?

joshka:
> (Currently disappearing down a open source rabbit hole, originally working on gitui, it needs textarea, turns out that needs tweaks, turns out those fixes need tui changes..just posted issue to windows terminal needed by crossterm.. happy days)

Yak shaving is a funny business.

> Would you accept a PR using derive-getters?

I think this would be good to see.

Because the builder methods currently use the field name (e.g. `Block::title()`) you'll need to add an attribute on every field to rename the getter. I suspect getset will be easier. E.g:
```rust
#[derive(Debug, Default, Clone, Eq, PartialEq, Hash, Getters)]
#[getset(get = "pub with_prefix")]
```

Make sure to pay some attention to how the API docs render. I suspect the field level comments will not be good enough for the getters in some cases as we've often commented just the builder methods. Perhaps it's sufficient to just link to the setters from each field?

pm100:
Yup I saw that about names. But I like getters simplicity, you can just ask to generate all getters with a derive at the struct level. So now my next rabbit hole to get derive-getters to accept a prefix attribute at the struct level

[...]

PR #782, opened 2024-01-10
feat(tabs): add a function that returns the current tab

AlexCul:
### Problem

It was not possible to get the current selected tab

### DIFF
--- a/src/widgets/tabs.rs
+++ b/src/widgets/tabs.rs
@@ -112,6 +112,11 @@ impl<'a> Tabs<'a> {
         self
     }
 
+    /// Returns the current selected tab
+    pub fn selected(self) -> usize {
+        self.selected
+    }
+
     /// Sets the style of the tabs.
     ///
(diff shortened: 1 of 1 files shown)


orhun:
Hello, why this was closed? 🐻

joshka:
Hi there - this sort of change rolls up to something we've discussed more generally in https://github.com/ratatui-org/ratatui/discussions/78 and https://github.com/ratatui-org/ratatui/issues/619. For the moment we're holding off on doing this, mainly because any code that you use to set a field is accessible to the user already.

E.g. your code to create the `Tabs` will at some point have called `Tabs::select()` passing it a variable or calculation. That variable or calculation can be used to work out what the selected tab is.

AlexCul:
Hi there. Some CI tests on this pull request failed so I was afraid it's something with my request. Therefore I closed it and decided to check it out.
3 CI tests were failed with the error 255, so I think my pull request was fine. I'll create a new one with no changes but link to issue #619 soon

PR #783, opened 2024-01-10
feat: Add `Constraint::Fixed(x)` and `Constraint::Proportional(x)`

kdheepak:
This PR adds `Constraint::Fixed(x)` and `Constraint::Proportional(x)`. I've also added a simple example showing how some constraints interact with each other:

constraints

This PR also deprecates the unused `Constraint::apply` method.

This PR allows users to create more advanced layouts. For example, we can force column spacing in the table layouts with a `Fixed` constraint.

Without this PR:

image

With this PR:

image

In the second screenshot, there's a column spacing after the description column. This does not exist in the first screenshot.

I've also improved the docstring for the enum variants (with some static ascii illustrations from the example):

image

**Additional information**

Currently we have the following levels of priorities:

- `Constraint::Min(lower)` and `Constraint::Max(upper)`
    - `segment >= lower` and `segment <= upper` have weights `STRONG`
    - `segment == lower` and `segment == upper` have weights `MEDIUM`
- `Constraint::Length(v)`, `Constraint::Percentage(v)` and `Constraint::Ratio(v)`
    - `segment == v` has weight `STRONG`

With this PR it becomes the following:

- (NEW) `Constraint::Fixed(v)`
    - `segment == v` has weight `REQUIRED - 1`
- (unchanged) `Constraint::Min(lower)` and `Constraint::Max(upper)`
    - `segment >= lower` and `segment <= upper` have weights `STRONG`
    - `segment == lower` and `segment == upper` have weights `MEDIUM`
- (unchanged) `Constraint::Length(v)`, `Constraint::Percentage(v)` and `Constraint::Ratio(v)`
    - `segment == v` has weight `STRONG`
- (NEW) `Constraint::Proportional(scaling_factor)`
    - `segment == total_available_area` has weight `WEAK`
    - `fill_segment / scaling_factor == any_other_fill_segment / any_other_fill_segment_scaling_factor` has weight `REQUIRED - 1`

### DIFF
--- a/examples/constraints.rs
+++ b/examples/constraints.rs
@@ -1,0 +1,495 @@
+use std::{error::Error, io};
+
+use crossterm::{
+    event::{self, Event, KeyCode},
+    execute,
+    terminal::{disable_raw_mode, enable_raw_mode, EnterAlternateScreen, LeaveAlternateScreen},
+};
+use itertools::Itertools;
+use ratatui::{layout::Constraint::*, prelude::*, widgets::*};
+
+fn main() -> Result<(), Box<dyn Error>> {
+    // setup terminal
+    enable_raw_mode()?;
+    let mut stdout = io::stdout();
+    execute!(stdout, EnterAlternateScreen)?;
+    let backend = CrosstermBackend::new(stdout);
+    let mut terminal = Terminal::new(backend)?;
+
+    // create app and run it
+    let res = run_app(&mut terminal);
+
+    // restore terminal
+    disable_raw_mode()?;
+    execute!(terminal.backend_mut(), LeaveAlternateScreen)?;
+    terminal.show_cursor()?;
+
+    if let Err(err) = res {
+        println!("{err:?}");
+    }
+
+    Ok(())
+}
+
+fn run_app<B: Backend>(terminal: &mut Terminal<B>) -> io::Result<()> {
+    let mut selection = ExampleSelection::Fixed;
+    loop {
+        terminal.draw(|f| f.render_widget(selection, f.size()))?;
+
+        if let Event::Key(key) = event::read()? {
+            use KeyCode::*;
+            match key.code {
+                Char('q') => break Ok(()),
+                Char('j') | Char('l') | Down | Right => {
+                    selection = selection.next();
+                }
+                Char('k') | Char('h') | Up | Left => {
+                    selection = selection.previous();
+                }
+                _ => (),
+            }
+        }
+    }
+}
+
+#[derive(Debug, Copy, Clone)]
+enum ExampleSelection {
+    Fixed,
+    Length,
+    Percentage,
+    Ratio,
+    Proportional,
+    Min,
+    Max,
+}
+
+impl ExampleSelection {
+    fn previous(&self) -> Self {
+        use ExampleSelection::*;
+        match *self {
+            Fixed => Fixed,
+            Length => Fixed,
+            Percentage => Length,
+            Ratio => Percentage,
+            Proportional => Ratio,
+            Min => Proportional,
+            Max => Min,
+        }
+    }
+
+    fn next(&self) -> Self {
+        use ExampleSelection::*;
+        match *self {
+            Fixed => Length,
+            Length => Percentage,
+            Percentage => Ratio,
+            Ratio => Proportional,
+            Proportional => Min,
+            Min => Max,
+            Max => Max,
+        }
+    }
+
+    fn selected(&self) -> usize {
+        use ExampleSelection::*;
+        match self {
+            Fixed => 0,
+            Length => 1,
+            Percentage => 2,
+            Ratio => 3,
+            Proportional => 4,
+            Min => 5,
+            Max => 6,
+        }
+    }
+}
+
+impl Widget for ExampleSelection {
+    fn render(self, area: Rect, buf: &mut Buffer) {
+        let [tabs, area] = area.split(&Layout::vertical([Fixed(3), Proportional(0)]));
+        let [area, _] = area.split(&Layout::horizontal([Fixed(80), Proportional(0)]));
+
+        self.render_tabs(tabs, buf);
+
+        match self {
+            ExampleSelection::Fixed => self.render_fixed_example(area, buf),
+            ExampleSelection::Length => self.render_length_example(area, buf),
+            ExampleSelection::Percentage => self.render_percentage_example(area, buf),
+            ExampleSelection::Ratio => self.render_ratio_example(area, buf),
+            ExampleSelection::Proportional => self.render_proportional_example(area, buf),
+            ExampleSelection::Min => self.render_min_example(area, buf),
+            ExampleSelection::Max => self.render_max_example(area, buf),
+        }
+    }
+}
+
+impl ExampleSelection {
+    fn render_tabs(&self, area: Rect, buf: &mut Buffer) {
+        // ┌Constraints───────────────────────────────────────────────────────────────────┐
+        // │  Fixed  │  Length  │  Percentage  │  Ratio  │  Proportional  │  Min  │  Max  │
+        // └──────────────────────────────────────────────────────────────────────────────┘
+        Tabs::new(
+            [
+                ExampleSelection::Fixed,
+                ExampleSelection::Length,
+                ExampleSelection::Percentage,
+                ExampleSelection::Ratio,
+                ExampleSelection::Proportional,
+                ExampleSelection::Min,
+                ExampleSelection::Max,
+            ]
+            .iter()
+            .map(|e| format!("{:?}", e)),
+        )
+        .block(Block::bordered().title("Constraints"))
+        .highlight_style(Style::default().yellow())
+        .select(self.selected())
+        .padding("  ", "  ")
+        .render(area, buf);
+    }
+
+    fn render_fixed_example(&self, area: Rect, buf: &mut Buffer) {
+        let [example1, example2, _] = area.split(&Layout::vertical([Fixed(8); 3]));
+
+        // Fixed(40), Proportional(0)
+        //
+        // <---------------------50 px---------------------->
+        //
+        // ┌──────────────────────────────────────┐┌────────┐
+        // │                 40 px                ││  10 px │
+        // └──────────────────────────────────────┘└────────┘
+        Example::new([Fixed(40), Proportional(0)]).render(example1, buf);
+
+        // Fixed(20), Fixed(20), Proportional(0)
+        //
+        // <---------------------50 px---------------------->
+        //
+        // ┌──────────────────┐┌──────────────────┐┌────────┐
+        // │       20 px      ││       20 px      ││  10 px │
+        // └──────────────────┘└──────────────────┘└────────┘
+        Example::new([Fixed(20), Fixed(20), Proportional(0)]).render(example2, buf);
+    }
+
+    fn render_length_example(&self, area: Rect, buf: &mut Buffer) {
+        let [example1, example2, example3, example4, _] =
+            area.split(&Layout::vertical([Fixed(8); 5]));
+
+        // Length(20), Fixed(20)
+        //
+        // <---------------------50 px---------------------->
+        //
+        // ┌────────────────────────────┐┌──────────────────┐
+        // │            30 px           ││       20 px      │
+        // └────────────────────────────┘└──────────────────┘
+        Example::new([Length(20), Fixed(20)]).render(example1, buf);
+
+        // Length(20), Length(20)
+        //
+        // <---------------------50 px---------------------->
+        //
+        // ┌──────────────────┐┌────────────────────────────┐
+        // │       20 px      ││            30 px           │
+        // └──────────────────┘└────────────────────────────┘
+        Example::new([Length(20), Length(20)]).render(example2, buf);
+
+        // Length(20), Min(20)
+        //
+        // <---------------------50 px---------------------->
+        //
+        // ┌──────────────────┐┌────────────────────────────┐
+        // │       20 px      ││            30 px           │
+        // └──────────────────┘└────────────────────────────┘
+        Example::new([Length(20), Min(20)]).render(example3, buf);
+
+        // Length(20), Max(20)
+        //
+        // <---------------------50 px---------------------->
+        //
+        // ┌────────────────────────────┐┌──────────────────┐
+        // │            30 px           ││       20 px      │
+        // └────────────────────────────┘└──────────────────┘
+        Example::new([Length(20), Max(20)]).render(example4, buf);
+    }
+
+    fn render_percentage_example(&self, area: Rect, buf: &mut Buffer) {
+        let [example1, example2, example3, example4, example5, _] =
+            area.split(&Layout::vertical([Fixed(8); 6]));
+
+        // Percentage(75), Proportional(0)
+        //
+        // <---------------------50 px---------------------->
+        //
+        // ┌────────────────────────────────────────────────┐
+        // │                      50 px                     │
+        // └────────────────────────────────────────────────┘
+        //
+        Example::new([Percentage(75), Proportional(0)]).render(example1, buf);
+
+        // Percentage(25), Proportional(0)
+        //
+        // <---------------------50 px---------------------->
+        //
+        // ┌────────────────────────────────────────────────┐
+        // │                      50 px                     │
+        // └────────────────────────────────────────────────┘
+        Example::new([Percentage(25), Proportional(0)]).render(example2, buf);
+
+        // Percentage(50), Min(20)
+        //
+        // <---------------------50 px---------------------->
+        //
+        // ┌───────────────────────┐┌───────────────────────┐
+        // │         25 px         ││         25 px         │
+        // └───────────────────────┘└───────────────────────┘
+        Example::new([Percentage(50), Min(20)]).render(example3, buf);
+
+        // Percentage(0), Max(0)
+        //
+        // <---------------------50 px---------------------->
+        //
+        // ┌────────────────────────────────────────────────┐
+        // │                      50 px                     │
+        // └────────────────────────────────────────────────┘
+        Example::new([Percentage(0), Max(0)]).render(example4, buf);
+
+        // Percentage(0), Proportional(0)
+        //
+        // <---------------------50 px---------------------->
+        //
+        // ┌────────────────────────────────────────────────┐
+        // │                      50 px                     │
+        // └────────────────────────────────────────────────┘
+        Example::new([Percentage(0), Proportional(0)]).render(example5, buf);
+    }
+
+    fn render_ratio_example(&self, area: Rect, buf: &mut Buffer) {
+        let [example1, example2, example3, example4, _] =
+            area.split(&Layout::vertical([Fixed(8); 5]));
+
+        // Ratio(1, 2), Ratio(1, 2)
+        //
+        // <---------------------50 px---------------------->
+        //
+        // ┌───────────────────────┐┌───────────────────────┐
+        // │         25 px         ││         25 px         │
+        // └───────────────────────┘└───────────────────────┘
+        Example::new([Ratio(1, 2); 2]).render(example1, buf);
+
+        // Ratio(1, 4), Ratio(1, 4), Ratio(1, 4), Ratio(1, 4)
+        //
+        // <---------------------50 px---------------------->
+        //
+        // ┌───────────┐┌──────────┐┌───────────┐┌──────────┐
+        // │   13 px   ││   12 px  ││   13 px   ││   12 px  │
+        // └───────────┘└──────────┘└───────────┘└──────────┘
+        Example::new([Ratio(1, 4); 4]).render(example2, buf);
+
+        // Ratio(1, 2), Ratio(1, 3), Ratio(1, 4)
+        //
+        // <---------------------50 px---------------------->
+        //
+        // ┌───────────────────────┐┌───────────────┐┌──────┐
+        // │         25 px         ││     17 px     ││ 8 px │
+        // └───────────────────────┘└───────────────┘└──────┘
+        Example::new([Ratio(1, 2), Ratio(1, 3), Ratio(1, 4)]).render(example3, buf);
+
+        // Ratio(1, 2), Percentage(25), Length(10)
+        //
+        // <---------------------50 px---------------------->
+        //
+        // ┌───────────────────────┐┌───────────┐┌──────────┐
+        // │         25 px         ││   13 px   ││   12 px  │
+        // └───────────────────────┘└───────────┘└──────────┘
+        Example::new([Ratio(1, 2), Percentage(25), Length(10)]).render(example4, buf);
+    }
+
+    fn render_proportional_example(&self, area: Rect, buf: &mut Buffer) {
+        let [example1, example2, _] = area.split(&Layout::vertical([Fixed(8); 3]));
+
+        // Proportional(1), Proportional(2), Proportional(3)
+        //
+        // <---------------------50 px---------------------->
+        //
+        // ┌──────┐┌───────────────┐┌───────────────────────┐
+        // │ 8 px ││     17 px     ││         25 px         │
+        // └──────┘└───────────────┘└───────────────────────┘
+        Example::new([Proportional(1), Proportional(2), Proportional(3)]).render(example1, buf);
+
+        // Proportional(1), Percentage(50), Proportional(1)
+        //
+        // <---------------------50 px---------------------->
+        //
+        // ┌───────────┐┌───────────────────────┐┌──────────┐
+        // │   13 px   ││         25 px         ││   12 px  │
+        // └───────────┘└───────────────────────┘└──────────┘
+        Example::new([Proportional(1), Percentage(50), Proportional(1)]).render(example2, buf);
+    }
+
+    fn render_min_example(&self, area: Rect, buf: &mut Buffer) {
+        let [example1, example2, example3, example4, example5, _] =
+            area.split(&Layout::vertical([Fixed(8); 6]));
+        // Percentage(100), Min(0)
+        //
+        // <---------------------50 px---------------------->
+        //
+        // ┌────────────────────────────────────────────────┐
+        // │                      50 px                     │
+        // └────────────────────────────────────────────────┘
+        Example::new([Percentage(100), Min(0)]).render(example1, buf);
+
+        // Percentage(100), Min(20)
+        //
+        // <---------------------50 px---------------------->
+        //
+        // ┌────────────────────────────┐┌──────────────────┐
+        // │            30 px           ││       20 px      │
+        // └────────────────────────────┘└──────────────────┘
+        Example::new([Percentage(100), Min(20)]).render(example2, buf);
+
+        // Percentage(100), Min(40)
+        //
+        // <---------------------50 px---------------------->
+        //
+        // ┌────────┐┌──────────────────────────────────────┐
+        // │  10 px ││                 40 px                │
+        // └────────┘└──────────────────────────────────────┘
+        Example::new([Percentage(100), Min(40)]).render(example3, buf);
+
+        // Percentage(100), Min(60)
+        //
+        // <---------------------50 px---------------------->
+        //
+        // ┌────────────────────────────────────────────────┐
+        // │                      50 px                     │
+        // └────────────────────────────────────────────────┘
+        Example::new([Percentage(100), Min(60)]).render(example4, buf);
+
+        // Percentage(100), Min(80)
+        //
+        // <---------------------50 px---------------------->
+        //
+        // ┌────────────────────────────────────────────────┐
+        // │                      50 px                     │
+        // └────────────────────────────────────────────────┘
+        Example::new([Percentage(100), Min(80)]).render(example5, buf);
+    }
+
+    fn render_max_example(&self, area: Rect, buf: &mut Buffer) {
+        let [example1, example2, example3, example4, example5, _] =
+            area.split(&Layout::vertical([Fixed(8); 6]));
+
+        // Percentage(0), Max(0)
+        //
+        // <---------------------50 px---------------------->
+        //
+        // ┌────────────────────────────────────────────────┐
+        // │                      50 px                     │
+        // └────────────────────────────────────────────────┘
+        Example::new([Percentage(0), Max(0)]).render(example1, buf);
+
+        //
+        // Percentage(0), Max(20)
+        //
+        // <---------------------50 px---------------------->
+        //
+        // ┌────────────────────────────┐┌──────────────────┐
+        // │            30 px           ││       20 px      │
+        // └────────────────────────────┘└──────────────────┘
+        Example::new([Percentage(0), Max(20)]).render(example2, buf);
+
+        // Percentage(0), Max(40)
+        //
+        // <---------------------50 px---------------------->
+        //
+        // ┌────────┐┌──────────────────────────────────────┐
+        // │  10 px ││                 40 px                │
+        // └────────┘└──────────────────────────────────────┘
+        Example::new([Percentage(0), Max(40)]).render(example3, buf);
+
+        // Percentage(0), Max(60)
+        //
+        // <---------------------50 px---------------------->
+        //
+        // ┌────────────────────────────────────────────────┐
+        // │                      50 px                     │
+        // └────────────────────────────────────────────────┘
+        Example::new([Percentage(0), Max(60)]).render(example4, buf);
+
+        // Percentage(0), Max(80)
+        //
+        // <---------------------50 px---------------------->
+        //
+        // ┌────────────────────────────────────────────────┐
+        // │                      50 px                     │
+        // └────────────────────────────────────────────────┘
+        Example::new([Percentage(0), Max(80)]).render(example5, buf);
+    }
+}
+
+struct Example {
+    constraints: Vec<Constraint>,
+}
+
+impl Example {
+    fn new<C>(constraints: C) -> Self
+    where
+        C: Into<Vec<Constraint>>,
+    {
+        Self {
+            constraints: constraints.into(),
+        }
+    }
+}
+
+impl Widget for Example {
+    fn render(self, area: Rect, buf: &mut Buffer) {
+        let [title, legend, area] = area.split(&Layout::vertical([Ratio(1, 3); 3]));
+        let blocks = Layout::horizontal(&self.constraints).split(area);
+
+        self.heading().render(title, buf);
+
+        self.legend(legend.width as usize).render(legend, buf);
+
+        for (i, block) in blocks.iter().enumerate() {
+            let text = format!("{} px", block.width);
+            let fg = Color::Indexed(i as u8 + 1);
+            self.illustration(text, fg).render(*block, buf);
+        }
+    }
+}
+
+impl Example {
+    fn heading(&self) -> Paragraph {
+        // Renders the following
+        //
+        // Fixed(40), Proportional(0)
+        let spans = self.constraints.iter().enumerate().map(|(i, c)| {
+            let color = Color::Indexed(i as u8 + 1);
+            Span::styled(format!("{:?}", c), color)
+        });
+        let heading =
+            Line::from(Itertools::intersperse(spans, Span::raw(", ")).collect::<Vec<Span>>());
+        Paragraph::new(heading).block(Block::default().padding(Padding::vertical(1)))
+    }
+
+    fn legend(&self, width: usize) -> Paragraph {
+        // a bar like `<----- 80 px ----->`
+        let width_label = format!("{} px", width);
+        let width_bar = format!(
+            "<{width_label:-^width$}>",
+            width = width - width_label.len() / 2
+        );
+        Paragraph::new(width_bar.dark_gray()).alignment(Alignment::Center)
+    }
+
+    fn illustration(&self, text: String, fg: Color) -> Paragraph {
+        // Renders the following
+        //
+        // ┌─────────┐┌─────────┐
+        // │  40 px  ││  40 px  │
+        // └─────────┘└─────────┘
+        Paragraph::new(text)
+            .alignment(Alignment::Center)
+            .block(Block::bordered().style(Style::default().fg(fg)))
+    }
+}
--- a/examples/constraints.tape
+++ b/examples/constraints.tape
@@ -1,0 +1,14 @@
+# This is a vhs script. See https://github.com/charmbracelet/vhs for more info.
+# To run this script, install vhs and run `vhs ./examples/constraints.tape`
+Output "target/constraints.gif"
+Set Theme "Aardvark Blue"
+Set FontSize 18
+Set Width 1200
+Set Height 1200
+Hide
+Type "cargo run --example=constraints --features=crossterm"
+Enter
+Sleep 2s
+Show
+Sleep 5s
+Right @5s 7
--- a/src/layout/constraint.rs
+++ b/src/layout/constraint.rs
@@ -2,97 +2,216 @@ use std::fmt::{self, Display};
 
 use itertools::Itertools;
 
-/// A constraint that can be applied to a layout
+/// A constraint that defines the size of a layout element.
 ///
-/// Constraints are used to define the size of a layout. They can be used to define a fixed size, a
-/// percentage of the available space, a ratio of the available space, or a minimum or maximum size.
+/// Constraints can be used to specify a fixed size, a percentage of the available space, a ratio of
+/// the available space, a minimum or maximum size or a proportional value for a layout element.
 ///
 /// Relative constraints (percentage, ratio) are calculated relative to the entire space being
-/// split, not the space available after applying the more fixed constraints (min, max, length).
+/// divided, rather than the space available after applying more fixed constraints (min, max,
+/// length).
+///
+/// Constraints are prioritized in the following order:
+///
+/// 1. [`Constraint::Fixed`]
+/// 2. [`Constraint::Min`] / [`Constraint::Max`]
+/// 3. [`Constraint::Length`] / [`Constraint::Percentage`] / [`Constraint::Ratio`]
+/// 4. [`Constraint::Proportional`]
 ///
 /// # Examples
 ///
-/// `Constraint` has some helper methods to create lists of constraints from anything that can be
-/// converted into an iterator of u16s ((u16, u16) for ratios).
+/// `Constraint` provides helper methods to create lists of constraints from various input formats.
 ///
 /// ```rust
 /// # use ratatui::prelude::*;
-/// // a fixed layout
+/// // Create a layout with specified lengths for each element
 /// let constraints = Constraint::from_lengths([10, 20, 10]);
 ///
-/// // a centered layout
+/// // Create a layout with specified fixed lengths for each element
+/// let constraints = Constraint::from_fixed_lengths([10, 20, 10]);
+///
+/// // Create a centered layout using ratio or percentage constraints
 /// let constraints = Constraint::from_ratios([(1, 4), (1, 2), (1, 4)]);
 /// let constraints = Constraint::from_percentages([25, 50, 25]);
 ///
-/// // a centered layout with a minimum size
+/// // Create a centered layout with a minimum size constraint for specific elements
 /// let constraints = Constraint::from_mins([0, 100, 0]);
 ///
-/// // a sidebar layout specifying maximum sizes of the columns
+/// // Create a sidebar layout specifying maximum sizes for the columns
 /// let constraints = Constraint::from_maxes([30, 170]);
+///
+/// // Create a layout with proportional sizes for each element
+/// let constraints = Constraint::from_proportional_lengths([1, 2, 1]);
 /// ```
 #[derive(Debug, Clone, Copy, Eq, PartialEq, Hash)]
 pub enum Constraint {
-    /// Apply a percentage to a given amount
+    /// Applies a percentage of the available space to the element
     ///
-    /// Converts the given percentage to a f32, and then converts it back, trimming off the decimal
-    /// point (effectively rounding down)
+    /// Converts the given percentage to a floating-point value and multiplies that with area.
+    /// This value is rounded back to a integer as part of the layout split calculation.
+    ///
+    /// # Examples
+    ///
+    /// `[Percentage(75), Proportional(1)]`
+    ///
+    /// ```plain
+    /// ┌────────────────────────────────────┐┌──────────┐
+    /// │                38 px               ││   12 px  │
+    /// └────────────────────────────────────┘└──────────┘
     /// ```
-    /// # use ratatui::prelude::*;
-    /// assert_eq!(0, Constraint::Percentage(50).apply(0));
-    /// assert_eq!(2, Constraint::Percentage(50).apply(4));
-    /// assert_eq!(5, Constraint::Percentage(50).apply(10));
-    /// assert_eq!(5, Constraint::Percentage(50).apply(11));
+    ///
+    /// `[Percentage(50), Proportional(1)]`
+    ///
+    /// ```plain
+    /// ┌───────────────────────┐┌───────────────────────┐
+    /// │         25 px         ││         25 px         │
+    /// └───────────────────────┘└───────────────────────┘
     /// ```
     Percentage(u16),
-    /// Apply a ratio
+    /// Applies a ratio of the available space to the element
     ///
-    /// Converts the given numbers to a f32, and then converts it back, trimming off the decimal
-    /// point (effectively rounding down)
+    /// Converts the given ratio to a floating-point value and multiplies that with area.
+    /// This value is rounded back to a integer as part of the layout split calculation.
+    ///
+    /// # Examples
+    ///
+    /// `[Ratio(1, 2) ; 2]`
+    ///
+    /// ```plain
+    /// ┌───────────────────────┐┌───────────────────────┐
+    /// │         25 px         ││         25 px         │
+    /// └───────────────────────┘└───────────────────────┘
     /// ```
-    /// # use ratatui::prelude::*;
-    /// assert_eq!(0, Constraint::Ratio(4, 3).apply(0));
-    /// assert_eq!(4, Constraint::Ratio(4, 3).apply(4));
-    /// assert_eq!(10, Constraint::Ratio(4, 3).apply(10));
-    /// assert_eq!(100, Constraint::Ratio(4, 3).apply(100));
     ///
-    /// assert_eq!(0, Constraint::Ratio(3, 4).apply(0));
-    /// assert_eq!(3, Constraint::Ratio(3, 4).apply(4));
-    /// assert_eq!(7, Constraint::Ratio(3, 4).apply(10));
-    /// assert_eq!(75, Constraint::Ratio(3, 4).apply(100));
+    /// `[Ratio(1, 4) ; 4]`
+    ///
+    /// ```plain
+    /// ┌───────────┐┌──────────┐┌───────────┐┌──────────┐
+    /// │   13 px   ││   12 px  ││   13 px   ││   12 px  │
+    /// └───────────┘└──────────┘└───────────┘└──────────┘
     /// ```
     Ratio(u32, u32),
-    /// Apply no more than the given amount (currently roughly equal to [Constraint::Max], but less
-    /// consistent)
+    /// Applies a fixed size to the element
+    ///
+    /// The element size is set to the specified amount.
+    /// [`Constraint::Fixed`] will take precedence over all other constraints.
+    ///
+    /// # Examples
+    ///
+    /// `[Fixed(40), Proportional(1)]`
+    ///
+    /// ```plain
+    /// ┌──────────────────────────────────────┐┌────────┐
+    /// │                 40 px                ││  10 px │
+    /// └──────────────────────────────────────┘└────────┘
     /// ```
-    /// # use ratatui::prelude::*;
-    /// assert_eq!(0, Constraint::Length(4).apply(0));
-    /// assert_eq!(4, Constraint::Length(4).apply(4));
-    /// assert_eq!(4, Constraint::Length(4).apply(10));
+    ///
+    /// `[Fixed(20), Fixed(20), Proportional(1)]`
+    ///
+    /// ```plain
+    /// ┌──────────────────┐┌──────────────────┐┌────────┐
+    /// │       20 px      ││       20 px      ││  10 px │
+    /// └──────────────────┘└──────────────────┘└────────┘
+    /// ```
+    Fixed(u16),
+    /// Applies a length constraint to the element
+    ///
+    /// The element size is set to the specified amount.
+    ///
+    /// # Examples
+    ///
+    /// `[Length(20), Fixed(20)]`
+    ///
+    /// ```plain
+    /// ┌────────────────────────────┐┌──────────────────┐
+    /// │            30 px           ││       20 px      │
+    /// └────────────────────────────┘└──────────────────┘
+    /// ```
+    ///
+    /// `[Length(20), Length(20)]`
+    ///
+    /// ```plain
+    /// ┌──────────────────┐┌────────────────────────────┐
+    /// │       20 px      ││            30 px           │
+    /// └──────────────────┘└────────────────────────────┘
     /// ```
     Length(u16),
-    /// Apply at most the given amount
+    /// Applies the scaling factor proportional to all other [`Constraint::Proportional`] elements
+    /// to fill excess space
     ///
-    /// also see [std::cmp::min]
+    /// The element will only expand into excess available space, proportionally matching other
+    /// [`Constraint::Proportional`] elements while satisfying all other constraints.
+    ///
+    /// # Examples
+    ///
+    ///
+    /// `[Proportional(1), Proportional(2), Proportional(3)]`
+    ///
+    /// ```plain
+    /// ┌──────┐┌───────────────┐┌───────────────────────┐
+    /// │ 8 px ││     17 px     ││         25 px         │
+    /// └──────┘└───────────────┘└───────────────────────┘
     /// ```
-    /// # use ratatui::prelude::*;
-    /// assert_eq!(0, Constraint::Max(4).apply(0));
-    /// assert_eq!(4, Constraint::Max(4).apply(4));
-    /// assert_eq!(4, Constraint::Max(4).apply(10));
+    ///
+    /// `[Proportional(1), Percentage(50), Proportional(1)]`
+    ///
+    /// ```plain
+    /// ┌───────────┐┌───────────────────────┐┌──────────┐
+    /// │   13 px   ││         25 px         ││   12 px  │
+    /// └───────────┘└───────────────────────┘└──────────┘
+    /// ```
+    Proportional(u16),
+    /// Applies a maximum size constraint to the element
+    ///
+    /// The element size is set to at most the specified amount.
+    ///
+    /// # Examples
+    ///
+    /// `[Percentage(100), Min(20)]`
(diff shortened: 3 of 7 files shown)


joshka:
To give a concrete example of this, If I have a table width constraint [Min(20), Length(79)], with column spacing = 1, and resize my terminal to 50, what would the expected result be?

I'd want both columns to visible still, but probably truncated as [20, 1, 29] including the column space.
Alternatively the overflow to be taken out of all equally [10, 1, 39]. (I'm guessing this or [10, 0, 40]) is the current behavior).
I think this PR will probably result in that scenario being [0, 1, 49]. or possibly [0, 0, 50]. I'm not sure this is desireable.

kdheepak:
Currently because `Constraint::Length`, `Constraint::Percentage` and `Constraint::Ratio` all have the same "weight", i.e. `STRONG`. This means if a user can pre-calculate these values (which usually is possible), then there's no practical difference in the results between pre-calculating an equivalent `Constraint::Percentage()` or `Constraint::Ratio()` for the intended `Length` versus using the actual `Constraint::Length()`. `Length` right now is just a convenience instead of `Percentage` or `Ratio`. 

More concretely, in your example, if I wanted to handle resizing I could do this:

```rust
[Min(20), Length(1), Length(f.size().width * 79 / 100))]
```

Furthermore, in the `main` branch implementation, `Percentage` and `Ratio` are in effect just syntax differences for the same thing. Anything you can accomplish with `Ratio` you can accomplish with `Percentage`. 

With the `main` branch, these tests pass:

```rust
      .constraints([Min(20), Length(1), Percentage(79)])
      .split(Rect::new(0, 0, 50, 1));
  assert_eq!(
      layout[..],
      [
          Rect::new(0, 0, 20, 1),
          Rect::new(20, 0, 1, 1),
          Rect::new(21, 0, 29, 1)
      ]
  );
      .constraints([Min(20), Length(1), Ratio(79, 100)])
      .split(Rect::new(0, 0, 50, 1));
  assert_eq!(
      layout[..],
      [
          Rect::new(0, 0, 20, 1),
          Rect::new(20, 0, 1, 1),
          Rect::new(21, 0, 29, 1)
      ]
  );
```

I guess the question is what do you want `Constraint::Length` to mean? 

Currently it means try your best to make the length in this segment match this value but there's no strong guarantee for what happens if it cannot match. Percentage and Ratio are also the same, in effect making these redundant. 

I would like `Constraint::Length` to mean something different that is not possible to accomplish with `Constraint::Percentage` or `Constraint::Ratio`, giving us more flexibility in creating the layouts we want.

This PR is my short and easy workaround to being able to get any specific layout I want because if pre calculate `Constraint::Length` that will take priority and then other segments can flex into the values we provide.

If I were to make a more intrusive change that would allow us more flexibility in layouts, without adding any new variants, I would make the priorities the following:

1. `Length`: REQUIRED - 1
2. `Min`: REQUIRED - 1 _or_ STRONG 
3. `Max`: REQUIRED - 1 _or_ STRONG
4. `Percentage`: STRONG
5. `Ratio`: WEAK

If I could add more variants to `Constraint`, I would introduce a `Constraint::Fixed(x)` which does what this PR does for `Constraint::Length(x)` and leave `Length` alone. I'd also take the opportunity to introduce a `Constraint::Flex()` which will take up any remaining space equally matching any other `Flex` if possible. I'd love to write this: `[Fixed(3), Flex(), Flex(), Flex(), Fixed(3)]`

1. `Fixed`: REQUIRED - 1
2. `Length`: STRONG
3. `Min`: STRONG 
4. `Max`: STRONG
5. `Percentage`: STRONG
6. `Flex`: STRONG
7. `Ratio`: WEAK

Alternatively, we could add a variant I can specify the type of constraint and strength of the constraint when I create the enum, e.g. `[Min(5), Manual(5, ConstraintType::Equality, ConstraintStrength::Weak), Min(5)]`.

joshka:
I think I like being able to tweak the priority as the best fix for this.

Would it make sense to extract the match block at 459 that converts Constraint to the cassowary constrain into a method on Constraint?

Is there any way we could add a priority to the `Constraint` itself easily so that we can pass in higher priority constraints like this where the table spacers should not be made smaller? I think it probably would have to be a new type like WeightedConstraint, which we convert into, but I'm not sure.

kdheepak:
For me, the current implementation is confusing because of stuff like this:

```rust
    .constraints([Min(20), Length(1), Length(100)])
    .split(Rect::new(0, 0, 50, 1));

assert_eq!(layout[0].width, 20);
assert_eq!(layout[1].width, 1);
assert_eq!(layout[2].width, 29);

    .constraints([Length(100), Length(1), Min(20)])
    .split(Rect::new(0, 0, 50, 1));

assert_eq!(layout[0].width, 30);
assert_eq!(layout[1].width, 0);
assert_eq!(layout[2].width, 20);
```

Just introducing a `Fixed` would solve my gripes, because I could get it to behave like this:

```rust
    .constraints([Min(20), Fixed(1), Length(100)])
    .split(Rect::new(0, 0, 50, 1));

assert_eq!(layout[0].width, 20);
assert_eq!(layout[1].width, 1);
assert_eq!(layout[2].width, 29);

    .constraints([Length(100), Fixed(1), Min(20)])
    .split(Rect::new(0, 0, 50, 1));
assert_eq!(layout[0].width, 29);
assert_eq!(layout[1].width, 1);
assert_eq!(layout[2].width, 20);
```

kdheepak:
MEDIUM, REQUIRED, STRONG, WEAK are just floating point numbers and the cassowary algorithm accepts any f64. `REQUIRED` is special but the others are just smaller numbers in order:

```
    pub const REQUIRED: f64 = 1_001_001_000.0;
    pub const STRONG: f64 = 1_000_000.0;
    pub const MEDIUM: f64 = 1_000.0;
    pub const WEAK: f64 = 1.0;
```

The simplest change that will address the most features is this:

1. Add `Constraint::Fixed` as a variant
2. Add these lines to the solver

```
Constraint::Fixed(l) => {
    solver.add_constraint(element.size() | EQ(REQUIRED - 1.0) | f64::from(l))?
}
```

Without any tests, this is the full patch:

```diff
diff --git a/src/layout/constraint.rs b/src/layout/constraint.rs
index 3f330de..1c2fac6 100644
--- a/src/layout/constraint.rs
+++ b/src/layout/constraint.rs
@@ -70,6 +70,15 @@ pub enum Constraint {
     /// assert_eq!(4, Constraint::Length(4).apply(10));
     /// ```
     Length(u16),
+    /// Apply no more than the given amount (currently roughly equal to [Constraint::Max], but less
+    /// consistent)
+    /// ```
+    /// # use ratatui::prelude::*;
+    /// assert_eq!(0, Constraint::Fixed(4).apply(0));
+    /// assert_eq!(4, Constraint::Fixed(4).apply(4));
+    /// assert_eq!(4, Constraint::Fixed(4).apply(10));
+    /// ```
+    Fixed(u16),
     /// Apply at most the given amount
     ///
     /// also see [std::cmp::min]
@@ -108,6 +117,7 @@ impl Constraint {
                 (percentage * length).min(length) as u16
             }
             Constraint::Length(l) => length.min(l),
+            Constraint::Fixed(l) => length.min(l),
             Constraint::Max(m) => length.min(m),
             Constraint::Min(m) => length.max(m),
         }
@@ -249,6 +259,7 @@ impl Display for Constraint {
             Constraint::Percentage(p) => write!(f, "Percentage({})", p),
             Constraint::Ratio(n, d) => write!(f, "Ratio({}, {})", n, d),
             Constraint::Length(l) => write!(f, "Length({})", l),
+            Constraint::Fixed(l) => write!(f, "Fixed({})", l),
             Constraint::Max(m) => write!(f, "Max({})", m),
             Constraint::Min(m) => write!(f, "Min({})", m),
         }
diff --git a/src/layout/layout.rs b/src/layout/layout.rs
index 1111bde..b2989a3 100644
--- a/src/layout/layout.rs
+++ b/src/layout/layout.rs
@@ -469,6 +469,9 @@ impl Layout {
                 Constraint::Length(l) => {
                     solver.add_constraint(element.size() | EQ(STRONG) | f64::from(l))?
                 }
+                Constraint::Fixed(l) => {
+                    solver.add_constraint(element.size() | EQ(REQUIRED - 1.0) | f64::from(l))?
+                }
                 Constraint::Max(m) => {
                     solver.add_constraints(&[
                         element.size() | LE(STRONG) | f64::from(m),
@@ -1212,5 +1215,33 @@ mod tests {
                 ]
             );
         }
```

This will allow forcing table spacing, this wouldn't change how `Length` behaves right now, and is the best simplest solution imho. 

It adds a new variant to the constraints, so we have to document that as part of the tutorials etc.

joshka:
I get that solving this particular problem is necessary. But I wonder whether we can't solve some of the related problems at the same time (while avoiding solutions that make it difficult to solve those problems).

If we just add a single priority to Fixed, what happens if a particular case need more priorities? E.g. for a table we have column spacers, but highlight column is also significant, and we might also want to allocate space for scrollbar position instead of pre-splitting the layout. What happens when the user provides Fixed to the table constraints?

Another approach (spit balling an idea here) might be to add a `Constraint::Weighted(Constraint, f64)` variant that wraps any constraint and adds a weight.

BTW, was there any reason to choose REQUIRED - 1.0 instead of STRONG + 1.0? (I intuitively leaned towards the latter when thinking about this problem, so I'm curious if there's a reason that makes one better than the other)

joshka:
> The caveat is that I'm basing this of my understanding of the theory behind cassowary, and not the cassowary algorithm itself or this specific implementation. I haven't tested what I mentioned in this comment. I think there's a chance it behaves differently from how I mentioned based on the order in which it calculates the weights, and what the weights are specifically.

Got it. My intuition of the values was that they were acting as an ordering function rather than a cost, and that they are f64 so that you can arbitrarily insert items in the order where needed and not so that you can use math to choose how much to break the constraint by. I'll have to read a bit more at some point.

kdheepak:
> If we just add a single priority to Fixed, what happens if a particular case need more priorities? E.g. for a table we have column spacers, but highlight column is also significant, and we might also want to allocate space for scrollbar position instead of pre-splitting the layout. What happens when the user provides Fixed to the table constraints?
> 
> Another approach (spit balling an idea here) might be to add a Constraint::Weighted(Constraint, f64) variant that wraps any constraint and adds a weight.

Can you do `Constraint::Weighted(Constraint, f64)` in Rust? I think we'd have to use `Box` here, right?

My hesitation with `Constraint::Weighted` is the following:

1. We would be exposing implementation details of the cassowary algorithm to users.
2. Weights can behave in odd ways and this would give users a way to shoot themselves in the foot
3. It's harder to move to taffy and maintain backward compatibility
4. It doesn't solve all the problems (e.g. equal width rects) 

I don't think we want `Fixed` to have multiple weights, we want `Fixed` to basically _always_ allocate space. The only time it won't allocate space is if it would otherwise panic. So `Fixed` is an example of no priorities or always highest priorities.

I think there's a case to be made for controlling priorities for other constraints but I feel less strongly about those. I am pretty sure I can get every layout I want with just `Fixed(x)` in addition to the current existing variants. Additional variants would be convenience for me.

joshka:
>Can you do Constraint::Weighted(Constraint, f64) in Rust? I think we'd have to use Box here, right?

You're right - it would need a box or something similar.

> I am pretty sure I can get every layout I want with just Fixed(x) in addition to the current existing variants. Additional variants would be convenience for me.

This is the strongest argument for this. I think this probably also fixes the [Min(0), Length(1), Min(0)] problem we have that doesn't center properly yet.

kdheepak:
>  I think this probably also fixes the [Min(0), Length(1), Min(0)] problem we have that doesn't center properly yet.

I'm not certain it will work the way you think or at least I'm not certain this will consistently behave this way in every case. 

I tested it in this particular case now and it doesn't do what you expect:

```rust
    .constraints([Min(0), Fixed(1), Min(0)])
    .split(Rect::new(0, 0, 50, 1));
assert_eq!(layout[0].width, 0);
assert_eq!(layout[1].width, 1);
assert_eq!(layout[2].width, 49);
```

`Min(0)` means is that the rect's width is >=0 AND the rect's width is as close to zero as possible. 
This means in `[Min(0), Fixed(1), Min(0)]` the first and last rect's widths can be any of the following:

```
0, 49
1, 48
2, 47
...
49, 0
```

and the "cost" will be the same in all the cases above (`49 * (STRONG + MEDIUM)`). Additionally, the constraints as we formulated them do not say anything else about the widths in relation to each other. 

Because there's so many solutions with the same cost, you can think about it as all these solutions are equally likely. 
In practice, because of the cassowary implementation, the solution will typically be one of the extremes, i.e. either `0` and `49` or `49` and `0`.

I think the only way to center is to use this:

```rust
    .constraints([Percentage(50), Fixed(1), Percentage(50)])
    .split(Rect::new(0, 0, 50, 1));
assert_eq!(layout[0].width, 25);
assert_eq!(layout[1].width, 1);
assert_eq!(layout[2].width, 24);
```

In this result, when the first and last widths are `25` and `24` the "cost" is just `1 * STRONG`. Any other value would result result in a higher "cost", e.g. `10` and `34` would be (`15 * STRONG + 11 * STRONG`). 

So in this case, I would expect only two possible solutions, either `25` and `24` or `24` and `25`.

kdheepak:
Even that doesn't work as expected always. Let me share a solution for centering that will always work.

kdheepak:
I think centering is only possible consistently with a new type called `Flex` (or what I've called `Auto` in the past):

```rust
    .constraints([Flex(1), Fixed(10), Flex(1)])
    .split(Rect::new(0, 0, 50, 1));
assert_eq!(layout[0].width, 20);
assert_eq!(layout[1].width, 10);
assert_eq!(layout[2].width, 20);
```

`Flex` adds a dependency between every other `Flex` constraint, the inner number is how much that segment should be scaled in proportion to another.

```rust
    .constraints([Flex(1), Fixed(10), Flex(2)])
    .split(Rect::new(0, 0, 50, 1));
assert_eq!(layout[0].width, 13);
assert_eq!(layout[1].width, 10);
assert_eq!(layout[2].width, 27);
```

Maybe an alternative name could be `Proportional`? This allows making segments that are relatively sized. 

```
    .constraints([Flex(1), Flex(2), Flex(1), Flex(1)])
    .split(Rect::new(0, 0, 50, 1));
assert_eq!(layout[0].width, 10);
assert_eq!(layout[1].width, 20);
assert_eq!(layout[2].width, 10);
assert_eq!(layout[3].width, 10);

    .constraints([Flex(1), Flex(2), Flex(1), Flex(1)])
    .split(Rect::new(0, 0, 100, 1));
assert_eq!(layout[0].width, 20);
assert_eq!(layout[1].width, 40);
assert_eq!(layout[2].width, 20);
assert_eq!(layout[3].width, 20);
```

kdheepak:
I changed this PR to include `Fixed` and `Flex`. I still need to update the documentation but figured you could see the implementation.

kdheepak:
I don't quite understand why we have this function:

```rust
impl Constraint {
    pub fn apply(&self, length: u16) -> u16 {
        ...
```

We only seem to use it once in `chart.rs`.

kdheepak:
I changed the name of `Flex` to `Fill` since I believe that will be more a more intuitive way to describe how it works. I also changed `Percentage` and `Ratio` to be lower weights than `Length`.

`Ratio` is completely redundant imo, particularly so if we change `Percentage(u16)` to be `Percentage(f64)` or `Percentage(f32)`.

kdheepak:
Alternate names for `Fill` would be

- `Spread`
- `Stretch`
- `Auto`
- `Flex`

These alternate names have similar analogs in `flexbox` from css which may be a good reason for choosing them.

src/layout/constraint.rs
```diff
@@ -70,6 +70,23 @@ pub enum Constraint {
     /// assert_eq!(4, Constraint::Length(4).apply(10));
     /// ```
     Length(u16),
+    /// Apply no more or no less than the given amount
+    /// ```
+    /// # use ratatui::prelude::*;
+    /// assert_eq!(0, Constraint::Fixed(4).apply(0));
+    /// assert_eq!(4, Constraint::Fixed(4).apply(4));
+    /// assert_eq!(4, Constraint::Fixed(4).apply(10));
```
kdheepak, line 78: I'm not exactly sure what `.apply` is supposed to do, so this docstring can be improved.

joshka:
> I don't quite understand why we have this function:

It's a leftover from the original code prior to refactoring the split method to properly understand the Constraint variants. I think we left it in because I didn't have a robust understanding of how to deprecate code at that point. Deprecate and remove would be worth it.

>I changed the name of Flex to Fill since I believe that will be more a more intuitive way to describe how it works.

Sounds good. I like that better than Flex.

> I also changed Percentage and Ratio to be lower weights than Length.

Do you think it makes sense to make this change in another PR, with some associated tests to show the impact. I'm concerned that it might be one of those difficult to quantify changes?

>Ratio is completely redundant imo, particularly so if we change Percentage(u16) to be Percentage(f64) or Percentage(f32).

I've used Ratio a lot. It's good in particular for e.g. `[Ratio (1,x) ; x]`, which would otherwise not be able to properly split the full amount.  An example of this 33% vs 1/3.

src/layout/layout.rs
```diff
@@ -1177,8 +1201,8 @@ mod tests {
             assert_eq!(
                 layout[..],
                 [
-                    Rect::new(0, 0, 1, 0),
                     Rect::new(0, 0, 1, 1),
+                    Rect::new(0, 1, 1, 0),
```
kdheepak, line 1205: This is the only test that changed as a result of changing the weights of `Percentage` and `Ratio`. Either we are implicitly assuming these new weights in our tests or our tests are under specified (likely the latter?).

kdheepak:
> Do you think it makes sense to make this change in another PR, with some associated tests to show the impact. I'm concerned that it might be one of those difficult to quantify changes?

Currently there aren't any tests that break (except for one `edge_case` test). Let me know if you still want it as a separate PR.

> I've used Ratio a lot. It's good in particular for e.g. [Ratio (1,x) ; x], which would otherwise not be able to properly split the full amount. An example of this 33% vs 1/3.

I also like using `[Ratio(1,x); x]` as a shorthand, but `[Percentage(1.0 / x * 100.0); x]` should give you the exact same result provided we change `Percentage` to use `f64` or `f32`.

src/layout/layout.rs
joshka: Most Likely under specified. that's the other part of why I'd suggest moving this part of the change to another PR. It allows us to see which parts of the change are related to fixed and file and which parts are changes to the existing code.

kdheepak: Removed changes to the weights for `Percentage` and `Ratio`.

kdheepak:
> Do you think it makes sense to make this change in another PR, with some associated tests to show the impact. I'm concerned that it might be one of those difficult to quantify changes?

I removed the weight changes from this here and made a separate PR.

src/layout/constraint.rs
kdheepak: Improved the docstring in the latest commit.

examples/constraints.rs
```diff
@@ -0,0 +1,139 @@
...
+        if let Event::Key(key) = event::read()? {
+            if let KeyCode::Char('q') = key.code {
+                return Ok(());
+            }
+            if let KeyCode::Char('j') = key.code {
+                selection += 1
+            }
+            if let KeyCode::Char('k') = key.code {
+                selection = selection.saturating_sub(1)
+            }
```
joshka, line 51:
```suggestion
            use KeyCode::*;
            match key.code {
                Esc => break Ok(()),
                Left | Char('h') | Char('k') => selection = selection.saturating_sub(1),
                Right | Char('l') | Char('j') => selection = selection.saturating_add(1).min(6),
                _ => {}
            }
```

src/layout/constraint.rs
```diff
@@ -2,97 +2,204 @@ use std::fmt::{self, Display};
...
 /// ```
 #[derive(Debug, Clone, Copy, Eq, PartialEq, Hash)]
 pub enum Constraint {
-    /// Apply a percentage to a given amount
+    /// Applies a percentage of the available space to the element
     ///
-    /// Converts the given percentage to a f32, and then converts it back, trimming off the decimal
-    /// point (effectively rounding down)
+    /// Converts the given percentage to a floating-point value, then back to an integer, trimming
+    /// off the decimal point (effectively rounding down).
```
joshka, line 38: I think this description might be wrong. It applied when we used the `apply` method (which did truncate the area), but that changed when we moved the conversion to an area inside the split method, so splitting 3 into 50/50 works fine - it rounds up for the first element and down for the second.

```rust
    solver.add_constraint(element.size() | EQ(STRONG) | (area_size * percent))?;
}
```

src/layout/constraint.rs
```diff
@@ -2,97 +2,204 @@ use std::fmt::{self, Display};
...
+    ///
+    /// # Examples
+    ///
+    /// `[Percentage(75), Min(0)]`
```
joshka, line 42: Perhaps use the `Fill` in these examples for the leftover parts.

src/layout/constraint.rs
```diff
@@ -203,6 +332,23 @@ impl Constraint {
...
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// # use ratatui::prelude::*;
+    /// # let area = Rect::default();
+    /// let constraints = Constraint::from_mins([1, 2, 3]);
+    /// let layout = Layout::default().constraints(constraints).split(area);
+    /// ```
+    pub fn from_fills<T>(fills: T) -> Vec<Constraint>
```
joshka, line 346: I think this makes sense calling it from relative lengths.

src/layout/constraint.rs
```diff
@@ -323,40 +482,213 @@ mod tests {
...
     #[test]
     fn apply() {
-        assert_eq!(Constraint::Percentage(0).apply(100), 0);
-        assert_eq!(Constraint::Percentage(50).apply(100), 50);
-        assert_eq!(Constraint::Percentage(100).apply(100), 100);
-        assert_eq!(Constraint::Percentage(200).apply(100), 100);
-        assert_eq!(Constraint::Percentage(u16::MAX).apply(100), 100);
+        assert_eq!(
+            {
+                #[allow(deprecated)]
```
joshka, line 500: Slap this on the test itself.

src/layout/layout.rs
```diff
@@ -481,6 +475,67 @@ impl Layout {
                         element.size() | EQ(MEDIUM) | f64::from(m),
                     ])?;
                 }
+                Constraint::Length(l) => {
+                    solver.add_constraint(element.size() | EQ(STRONG) | f64::from(l))?
```
joshka, line 479: Question about the strength of Min/Max being higher because we have two constraints (GE and EQ). Would setting the length etc. to STRONG * 2 be worth an experiment?

Like what if all the Mins had an additional weak constraint that was `EQ(WEAK) | area` ? Would this get us somewhere good by default?

src/layout/layout.rs
```diff
@@ -481,6 +475,67 @@ impl Layout {
...
+        // │abcdef││abcdef│
+        // └──────┘└──────┘
+        //
+        // [Fill(1), Fill(2)]
+        // ┌──────┐┌────────────┐
+        // │abcdef││abcdefabcdef│
+        // └──────┘└────────────┘
+        //
+        // l_size / l_scale == r_size / r_scale
+        for ((&l_constraint, &l_element), (&r_constraint, &r_element)) in layout
```
joshka, line 516: Can this be implemented by adding one extra element, which each element gets EQ to instead of constraining every combinations of elements?

e.g.
```rust
let relative_element = ...;
solver.add_constraint(
    f64::from(scaling_factor) * relative_element.size() | EQ (..) | element.size()
)
```

src/layout/layout.rs
```diff
@@ -1212,5 +1267,176 @@ mod tests {
...
+            //
+            // So we check just that the last value is what we want it to be
+            let [_a, _b, c] = Rect::new(0, 0, 100, 1).split(&Layout::horizontal([
+                Length(25),
+                Length(25),
+                Fixed(25),
+            ]));
+            assert_eq!(c.width, 25);
+
+            let layout = Layout::default()
```
joshka, line 1325: i haven't read all these test cases, but it would be worth adding comments to describe the intent of each test. This helps understand why this test is true.

src/layout/layout.rs
```diff
@@ -1212,5 +1267,176 @@ mod tests {
...
+                .constraints([Min(20), Fixed(1), Length(100)])
+                .direction(Direction::Horizontal)
+                .split(Rect::new(0, 0, 50, 1));
```
joshka, line 1328: add `.map(Rect::width)` to simplify the tests to a single assertion over an array of u16

src/widgets/chart.rs
```diff
@@ -745,10 +745,12 @@ impl<'a> Chart<'a> {
             if let Some(inner_width) = legends.clone().max() {
                 let legend_width = inner_width + 2;
                 let legend_height = legends.count() as u16 + 2;
+                #[allow(deprecated)]
```
joshka, line 748: Add an issue to fix this up

kdheepak: Thanks for opening an issue: https://github.com/ratatui-org/ratatui/issues/793

src/layout/layout.rs
kdheepak: Some of the tests are just to capture the current behavior. If we don't have such tests, then our behavior will be under specified by the tests.

I'll mark those with comments stating as such.

src/layout/layout.rs
kdheepak: I think that would work. Do you think that would be easier to understand? We are introducing an external free variable that isn't part of the layout, so at first glance it might be a little confusing.

src/layout/layout.rs
kdheepak: Typically you want an order of magnitude difference at the very least to be safe. Otherwise you can end up switching behavior when you scale the number of a certain kind of constraints. This can be hard to predict, hard to test and hard to debug. 

I have some ideas for better weights for all the constraints that we can discuss in the other PR.

examples/constraints.rs
```diff
@@ -0,0 +1,139 @@
...
+        }
```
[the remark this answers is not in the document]
kdheepak: I hit commit suggestion here but GitHub said the comment was stale and when I refreshed it was gone. 
I reimplemented it and pushed, resolved in latest commit.

examples/constraints.rs
```diff
@@ -0,0 +1,139 @@
...
+        selection = selection.clamp(0, 6)
+    }
+}
+
+fn paragraph<T: Into<Text<'static>>>(text: T, bordered: bool) -> Paragraph<'static> {
```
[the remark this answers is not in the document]
kdheepak: I removed this and added a `Widget` struct per your suggestion.

examples/constraints.rs
```diff
@@ -0,0 +1,139 @@
...
+    for b in blocks.iter() {
+        f.render_widget(paragraph(format!("{} px", b.width), true), *b);
+    }
+}
+
+macro_rules! example {
+    ($f:ident, $area:ident, [$($args:tt)*]) => {
+        example($f, $area, vec![$($args)*], format!("`{}`", stringify!([$($args)*])));
+    };
+}
```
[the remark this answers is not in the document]
kdheepak: good idea! Done it latest commit.

examples/constraints.rs
```diff
@@ -0,0 +1,139 @@
...
+        }
+
+        4 => {
+            example!(f, example1, [Fill(1), Fill(2), Fill(3)]);
+            example!(f, example2, [Fill(1), Percentage(50), Fill(1)]);
+        }
+
+        5 => {
+            example!(f, example1, [Percentage(100), Min(20)]);
+            example!(f, example2, [Percentage(100), Min(10)]);
```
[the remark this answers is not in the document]
kdheepak: Do you think users would want to do that? 

That seems like it would not be well defined and might be one of those things that would change with weights in an odd way. I'm happy to add a test case for that.

examples/constraints.tape
```diff
@@ -0,0 +1,28 @@
...
+Set Theme "Aardvark Blue"
+Set Width 1200
+Set Height 600
+Hide
+Type "cargo run --example=constraints --features=crossterm"
+Enter
+Sleep 5s
+Show
+Sleep 5s
+Type j
```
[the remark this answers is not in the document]
kdheepak: Nice idea!

src/layout/layout.rs
joshka: I think this makes the underlying concept more obvious, that there is some base value item that these scaling factors all relate to, rather than them relating to every other constraint.

src/layout/layout.rs
joshka: I meant specifically * 2 because we're looking at 2 constraints for the min/max.
Strictly,  STRONG*2 is always > STRONG+MEDIUM

examples/constraints.rs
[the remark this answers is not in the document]
joshka: I deleted the comment to ensure that I was slapping them in a review. Fat fingers ;)

examples/constraints.rs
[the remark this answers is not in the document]
joshka: Not sure - it just seemed like a question to ask when I used the example and noticed that the examples showed 100%/min(0) and 0%/max(20).

src/layout/constraint.rs
```diff
@@ -2,97 +2,204 @@ use std::fmt::{self, Display};
...
 ///
-/// `Constraint` has some helper methods to create lists of constraints from anything that can be
-/// converted into an iterator of u16s ((u16, u16) for ratios).
+/// `Constraint` provides helper methods to create lists of constraints from various input formats.
 ///
 /// ```rust
 /// # use ratatui::prelude::*;
-/// // a fixed layout
+/// // Create a fixed layout with specified sizes for each element
 /// let constraints = Constraint::from_lengths([10, 20, 10]);
```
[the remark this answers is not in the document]
kdheepak: Added examples for these as well.

src/layout/constraint.rs
kdheepak: Done

src/layout/constraint.rs
kdheepak: Thanks! I tried it on the `assert_eq!` and that didn't work.

src/layout/layout.rs
kdheepak: Will be discussed here: https://github.com/ratatui-org/ratatui/pull/788

src/layout/layout.rs
kdheepak: I've added more comments.

src/layout/layout.rs
kdheepak: > I think this makes the underlying concept more obvious, that there is some base value item that these scaling factors all relate to, rather than them relating to every other constraint.

I think what you are suggesting is introducing a new concept, i.e. a "base value item" and then saying that the scaling factors relate to the base value item. The underlying concept as implemented right now is that the segments are "proportional" to each other.

kdheepak: I tried implementing it this way. It breaks some of the tests. Specifically when there's just one `Fill(0)`.

What do we want `[Fill(0), Percentage(10)]` and `[Fill(0), Max(0)]` to do? 

When only one `Fill` is present, currently, it behaves the same regardless of the number inside it, i.e. `[Fill(0), Percentage(10)]` is the same as `[Fill(1), Percentage(10)]`, i.e. the `Percentage` constraint is respected:

```rust
    .constraints([Fill(1), Percentage(20)])
    .split(Rect::new(0, 0, 100, 1));
assert_eq!(layout[0].width, 80);
assert_eq!(layout[1].width, 20);
```

If I change the implementation to what you are suggesting, when there's just one `Fill(0)`, `Fill(0)` is more strong than even a `Max` constraint, and this breaks the `Max` constraint.

```rust
    .constraints([Fill(0), Max(20)])
    .split(Rect::new(0, 0, 100, 1));
assert_eq!(layout[0].width, 0);
assert_eq!(layout[1].width, 100);
```

This does not occur with `Fill(1)`:

```rust
    .constraints([Fill(1), Max(20)])
    .split(Rect::new(0, 0, 100, 1));
assert_eq!(layout[0].width, 80);
assert_eq!(layout[1].width, 20);
```

When introducing a new free variable, I have to special case when there's just one `Fill(0)` to get it to do the old behavior. It feels less clean to have to do this introducing a new variable. 

I'm going to leave it as it is right now. We can revisit this and consider a refactor in the future if we find this is harder to understand.

joshka: Does adding weak constraints EQ area.size work?
To me, based on how you've explained Fill to be needed, Fill(0) says to me "This constraint should be 0 times any other constraints that fill, it should effectively aim to be the smallest element possible."

How is Fill(0), Max(20) -> 0, 100 wrong?

src/layout/constraint.rs
joshka: Does it work on the test function?

src/layout/layout.rs
joshka: >The underlying concept as implemented right now is that the segments are "proportional" to each other.

Kinda but not - [Fill(1), Fill(2)] means proportional to a floating amount that is shared, that just happens to be the same as the first constraint. Swap it out with `Fill(2), Fill(4)` for what should have the same result, but doesn't have a unity amount to compare to. I think you can probably make the variable not free by adding [element[0] + element[1] +... | EQ(WEAK) area.size()] or something like that.

examples/constraints.rs
```diff
@@ -0,0 +1,217 @@
...
+        C: Into<Vec<Constraint>>,
+    {
+        Self {
+            constraints: constraints.into(),
+        }
+    }
+}
+
+impl Widget for Example {
+    fn render(self, area: Rect, buf: &mut Buffer) {
```
joshka, line 73: Lots of code with no discernable gaps, which means you have to grok the whole instead of just a small slice.

examples/constraints.rs
```diff
@@ -0,0 +1,217 @@
...
+        let [title, legend, area] = area.split(&Layout::vertical([Ratio(1, 3); 3]));
+        let heading = Line::from(
+            Itertools::intersperse(
+                self.constraints.iter().enumerate().map(|(i, c)| {
+                    Span::styled(
+                        format!("{:?}", c),
+                        Style::default().fg(Color::Indexed(i as u8 + 1)),
+                    )
```
joshka, line 81:
```suggestion
                    let color = Color::Indexed(i as u8 + 1);
                    Span::styled(format!("{:?}", c), color)
```
Also, wonder it it might be worth putting a small macro implementation of this (without the need to discuss style interpolations because spans only have a single style) so we could just have written:
```rust
span!("{:?}", c, color);
```

examples/constraints.rs
```diff
@@ -0,0 +1,217 @@
...
+            constraints: constraints.into(),
+        }
+    }
+}
+
+impl Widget for Example {
+    fn render(self, area: Rect, buf: &mut Buffer) {
+        let [title, legend, area] = area.split(&Layout::vertical([Ratio(1, 3); 3]));
+        let heading = Line::from(
+            Itertools::intersperse(
```
joshka, line 76:
```rust
// color each constraint in order Red, Green, Yellow, Blue, ...
let colored_spans = ...
```
would go a ways to help simplify this section

examples/constraints.rs
```diff
@@ -0,0 +1,217 @@
...
+                        format!("{:?}", c),
+                        Style::default().fg(Color::Indexed(i as u8 + 1)),
+                    )
+                }),
+                Span::raw(", "),
+            )
+            .collect::<Vec<Span>>(),
+        );
+        Paragraph::new(heading)
+            .block(Block::default().padding(Padding::new(0, 0, 1, 0)))
```
joshka, line 88: Why pad this - it's a single line right?
Give it an area 2 high to write into in the layout.

examples/constraints.rs
joshka, line 1: crashes for small terminal heights.

examples/constraints.rs
```diff
@@ -0,0 +1,217 @@
...
+            legend.y,
+            "<",
+            Style::default().fg(Color::DarkGray),
+        );
+        buf.set_string(
+            legend.x + legend.width - 1,
+            legend.y,
+            ">",
+            Style::default().fg(Color::DarkGray),
+        );
```
joshka, line 107:
```suggestion
        /// a bar like `<----- 80 px ----->`
        width = format!(" {} px ", legend.width);
        let width_bar = format!("<{width_label:-^width$}>", width = legend.width)
        Paragraph::new(width_bar.dark_gray())
            .alignment(Alignment::Center)
            .render(legend, buf);
```

examples/constraints.rs
```diff
@@ -0,0 +1,217 @@
...
+        .block(Block::bordered().title("Constraints"))
+        .highlight_style(Style::default().yellow())
+        .select(tab)
+        .padding("  ", "  "),
+        tabs,
+    );
+    let [area, _] = area.split(&Layout::horizontal([Fixed(80), Proportional(0)]));
+
+    match tab {
+        0 => {
```
joshka, line 142: Can this be an enum?
Small methods would help here too.
0 => render_fixed_examples(...),
1 => render_length_examples(...),

src/layout/layout.rs
kdheepak: I wrote a long comment explaining why I think it shouldn't work that way but after arguing for it, I decided to include a section on the benefits of working the way you described and I think I've convinced myself that what you suggested is what it should do. 

I can share the comment later if you'd like.

examples/constraints.rs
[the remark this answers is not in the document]
kdheepak: I've added these examples in there.

src/layout/layout.rs
kdheepak: I've changed the implementation to use an extra `base_element`.

examples/constraints.rs
kdheepak: Fixed! I realized it was crashing only when I tried to generate the vhs gif.

examples/constraints.rs
kdheepak: I like the idea!

examples/constraints.rs
kdheepak: If I remove the padding I get this:

image

I think this looks nicer:

image

It's a padding from the top, here it is with padding `2`:

image

I changed it to `Padding::vertical(1)` so it is clearer from the code.

Related:

Should `Padding` get some additional functions?

Instead of this:

```
Block::default().padding(Padding::new(0, 0, 2, 0))
```

It should be:

```
Block::default().padding(Padding::top(2))
```

examples/constraints.rs
kdheepak: Done!

kdheepak: Thanks for the suggestion. Let me know if it is over engineered / over refactored now :)

examples/constraints.rs
kdheepak: Refactored

src/layout/constraint.rs
kdheepak: I changed the name to `Proportional` now and updated these functions.

kdheepak:
This is an example of why I don't like `Fill(0)` (`Fill` now renamed to `Proportional`) being `0` in the example we discussed earlier. With our current formulation, there's no guarantees that it will be `0` always. This test passes.

```rust
let [a, b, c] = Rect::new(0, 0, 100, 1).split(&Layout::horizontal([
    Proportional(0),
    Fixed(1),
    Proportional(0),
]));
assert_eq!([a.width, b.width, c.width], [0, 1, 99]);
```

kdheepak:
I'm tempted to make `Proportional(0)` behave identically to `Proportional(1)` using `.clamp(1, u16::MAX)`. We already do something similar for `Ratio`.

kdheepak:
Another test that has me worried:

```rust
let [a, b] = Rect::new(0, 0, 100, 1)
    .split(&Layout::default().constraints([Proportional(0), Length(20)]));
assert_eq!([a.width, b.width], [100, 100]);
```

I'm going to change the implementation back to the old way because I don't have the confidence that this will behave the way we expect it to. I am confident however that the way we did it in the old way will always do exactly what I intend it to do.

src/layout/layout.rs
kdheepak: Done.

src/layout/layout.rs
kdheepak: Here's my current thoughts on this discussion.

We are discussing two things here:

1. What should the semantics of `Proportional(0)` do?
2. What is the implementation we should use? i.e. should we use an additional variable `base_element` or not?

I'm fine with either implementation for number 2 but I have some thoughts related to mathematical numerical stability and code maintenance about number 1. So, I'd like to narrow down number 1, and then we can discuss number 2. 

There are 4 cases to consider when using `Proportional` constraints.

1. When you have code like this:

```rust
[Proportional(x), Proportional(y), ..., any_constraints_here]
```

and if the value of `x` and `y` is greater than `0`. In this case, `Proportional` works as you'd expect. It as scaled w.r.t to each other OR scaled w.r.t to a base value, the result is the same.

4. When you have code like this:

```rust
[Proportional(_), any_other_constraint_here]
```

the value inside `Proportional` shouldn't matter because there's nothing to be proportional to or relative to. Right now, i.e. with the `tuple_windows()` implementation, no constraints are added for `Proportional(_)` when there's just one of its kind.

In the `tuple_windows()` implementation, a single `Proportional(x)` behaves the same way for any value of `x`. This intuitively makes sense to me because the `x` shouldn't matter since there's nothing it can be proportional to. Or in other words, it is proportional to itself. It also behaves the same way it does when there are multiple `Proportional` constraints, i.e. they always fill up excess space. 
In the `base_element` implementation, a single `Proportional(x)` behaves differently for x equal to `0` and `x` greater than `0`. There's a special case here for `0` that I'm not particularly a fan off. Writing the code to make the semantics of this implementation behave the same as the previous requires some additional checks.

As a clarification, we can definitely get both the `tuple_windows()` implementation and the `base_element` implementation to behave in both of the above ways. We just have to choose which one we want. This is why I want to discussion the semantics first. The above is the default behavior with no additional constraints or changes. 

5. When you have code like this:

```rust
[Proportional(0), Proportional(y), ..., any_constraints_here]
```

i.e. there is just one `0` `Proportional` constraint (there is at least one other `Proportional` constraint and it and every other `Proportional` constraint is non-zero). 

There's many solutions to this problem and the solver will not be numerically stable for this formulation. Not having a stable result will cause problems with reproducibility and burden us with maintenance in the future if we want to add new constraints. I've solved this particular problem of numerical stability by forcing the `Proportional(0)` to actually be a very small value, which results in a zero sized segment in this particular case.  Even without the numerical stability fix, this would behave the same way in both `tuple_windows()` and `base_element`, i.e. it would be zero sized.

6. When you have code like this:

```rust
[Proportional(0), Proportional(0), ..., any_constraints_here]
```

and `x` and `y` are `0`, i.e. at least 2 zeros. We have to decide what we want to do here since it is technically undefined. Different solvers may return different results here. The same workaround as before works here because I've forced it to be a very small value. This makes it fill any excess space while also being proportional to any other non zero proportional constraint. Without the workaround, `tuple_windows()` behaves consistently but `base_element` does not.

`tuple_windows()`

```rust
let [a, b, c] = Rect::new(0, 0, 100, 1).split(&Layout::horizontal([
    Proportional(0),
    Fixed(1),
    Proportional(0),
]));
assert_eq!([a.width, b.width, c.width], [0, 1, 99]);

let [a, b, c] = Rect::new(0, 0, 100, 1).split(&Layout::horizontal([
    Proportional(0),
    Percentage(1),
    Proportional(0),
]));
assert_eq!([a.width, b.width, c.width], [0, 1, 99]);
```

`base_element`

```rust
let [a, b, c] = Rect::new(0, 0, 100, 1).split(&Layout::horizontal([
    Proportional(0),
    Fixed(1),
    Proportional(0),
]));
assert_eq!([a.width, b.width, c.width], [0, 1, 99]);

let [a, b, c] = Rect::new(0, 0, 100, 1).split(&Layout::horizontal([
    Proportional(0),
    Percentage(1),
    Proportional(0),
]));
assert_eq!([a.width, b.width, c.width], [0, 100, 0]);
```

`tuple_windows()` with numerical stability fix

```rust
let [a, b, c] = Rect::new(0, 0, 100, 1).split(&Layout::horizontal([
    Proportional(0),
    Fixed(1),
    Proportional(0),
]));
assert_eq!([a.width, b.width, c.width], [50, 1, 49]);

let [a, b, c] = Rect::new(0, 0, 100, 1).split(&Layout::horizontal([
    Proportional(0),
    Percentage(1),
    Proportional(0),
]));
assert_eq!([a.width, b.width, c.width], [50, 1, 49]);
```

examples/constraints.rs
joshka, line 1: This is super neat!

I really like how the selection enum is both the item for the tab widget and the widget itself. That's clever without being complex.

A couple of match statements could be replaced with Strum derives, and I'd probably reorder a few methods to read top to bottom, but this is really easy to follow.

src/layout/constraint.rs
```diff
@@ -203,6 +338,26 @@ impl Constraint {
...
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// # use ratatui::prelude::*;
+    /// # let area = Rect::default();
+    /// let constraints = Constraint::from_mins([1, 2, 3]);
+    /// let layout = Layout::default().constraints(constraints).split(area);
+    /// ```
+    pub fn from_proportionals<T>(proportionals: T) -> Vec<Constraint>
```
joshka, line 352: Would from_proportions make more sense? Or proportional lengths?

src/layout/layout.rs
joshka: How about we make 0 not allowed?

I think it's generally ok to have a special case that only applies when there is one element that satisfies some condition like this, so I'm not too concerned. It's good that we call out out as we'd want to test the edge cases like this thoroughly.

I'm confused why there's multiple solutions for the prop(0), prop(1), ... case in list item 5.

src/layout/layout.rs
```diff
@@ -1212,5 +1273,288 @@ mod tests {
                 ]
             );
         }
+
+        #[test]
+        fn proportional_fixed() {
```
joshka, line 1278: These are super clear and maintainable. Will read them more thoroughly later, but just wanted to call that out as really nice.

joshka:
> I'm tempted to make `Proportional(0)` behave identically to `Proportional(1)` using `.clamp(1, u16::MAX)`. We already do something similar for `Ratio`.

I wonder if this should take a float?

examples/constraints.rs
joshka: #796

examples/constraints.rs
joshka: Ahh my confusion was that this was bottom padding 😀 
Good idea about adding better constructors

joshka: https://github.com/ratatui-org/ratatui/issues/798

kdheepak:
> I wonder if this should take a float?

Unfortunately, we cannot make a enum variant contain a float without changing a whole bunch of stuff. `f64` cannot implement `Eq` (only `PartialEq`) and does not implement `Hash`, which has implications of its own. Any struct that then holds a `Constraint` cannot implement `Eq` and `Hash` too. 

I haven't tried using an `Option<f64>`, maybe it would work if it were `Proportional(None)` and `Proportional(Some(f_64_value))`?

examples/constraints.rs
kdheepak: I looked at `Strum` and would probably use that if I were doing this in my own app. But I figured as an example for a user this is probably easier to understand.

kdheepak: I'm curious what order you'd choose, happy to refactor it that way.

src/layout/constraint.rs
kdheepak: I chose `proportional_lengths`

kdheepak:
Using `Option<f64>` doesn't work either:

image

joshka:
Do we need the Eq / Hash impls for this? What would the use case for them be? If we've added these and they aren't really in use, I'd not be opposed to removing them to get the float behavior.

src/layout/layout.rs
kdheepak: We wouldn't be able to make `0` not allowed from the user perspective, right? 

I can absolutely clamp it to `1..u16::MAX` inside the split function but if I did, these tests would change this way:

```diff
  let [a, b, c] = Rect::new(0, 0, 100, 1).split(&Layout::horizontal([
      Proportional(0),
      Proportional(1),
      Proportional(0),
  ]));
- assert_eq!([a.width, b.width, c.width], [0, 100, 0]);
+ assert_eq!([a.width, b.width, c.width], [33, 34, 33]);

  let [a, b, c, d] = Rect::new(0, 0, 100, 1).split(&Layout::horizontal([
      Proportional(0),
      Proportional(2),
      Proportional(0),
      Proportional(1),
  ]));
- assert_eq!([a.width, b.width, c.width, d.width], [0, 67, 0, 33]);
+ assert_eq!([a.width, b.width, c.width, d.width], [20, 40, 20, 20]);
```

The reason I like the behavior the way it is right now is because it is consistent with how I think about stable solutions. For example both these tests pass with the current implementation:

```rust
  let [a, b, c] = Rect::new(0, 0, 100, 1).split(&Layout::horizontal([
      Proportional(u16::MAX),
      Proportional(1),
      Percentage(20),
  ]));
  assert_eq!([a.width, b.width, c.width], [80, 0, 20]);

  let [a, b, c] = Rect::new(0, 0, 100, 1).split(&Layout::horizontal([
      Proportional(u16::MAX),
      Proportional(0),
      Percentage(20),
  ]));
  assert_eq!([a.width, b.width, c.width], [80, 0, 20]);
```

examples/constraints.rs
joshka: > I'm curious what order you'd choose, happy to refactor it that way.

Optimize for never having to scroll up to understand what something does.

`impl Widget for ExampleSelection` can be above a second `impl ExampleSelection` section with all the render functions. Probably move render_tabs first in that block as it is rendered visually first.

Same deal with `Example`

I like to keep the layout calculations together up top in the render method. 

The comments in the impl Widget for Example could be doc comments on the methods.

---

A somewhat related observation, thinking about the way the example methods return widgets, and the ExampleSelection methods call render multiple times, there's a missing harmony that could be filled with a LayoutContainer widget perhaps.

e.g. I wonder if something like this could work:

```rust
fn fixed_example(&self) -> LayoutContainer {
    LayoutContainer::new(
        Layout::vertical([Fixed(8); 2]),
        [
            Example::new([Fixed(40), Proportional(0)]),
            Example::new([Fixed(20), Fixed(20), Proportional(0)]),
        ],
    )
}
```

I suspect there might be problems with boxing being needed for this and render calls not working otherwise.

kdheepak:
> Do we need the Eq / Hash impls for this? What would the use case for them be?

I believe we wouldn't be able to store the `Constraint` in a `HashMap`. And also a user wouldn't be able to do this:

```rust
#[derive(Debug, Clone, Eq, PartialEq, Hash)]
pub struct MyCustomWidget<'a> {
    ...
    widths: Vec<Constraint>,
    ...
}
```

I don't think it is worth it just to expose `f64` values to a user. The results that a user gets from more accurately expressing the `Constraint` in `f64` is still going to have a +/- 1 `u16` from the true result at the end of the day. 

Users can still express `Ratio` and `Proportional` values as an approximate decimal value they want. If someone wanted to express `0.0078125`, they could `Ratio(1, 128)` or `[Proportional(1), Proportional(128)]`. 

If were were designing layouts for use not in a terminal, I'd recommend changing it. Now I don't see it being all that meaningful.

joshka:
What if I want my layouts to be always the golden ratio regardless of terminal size? (And yeah I know that an integer approx would be good enough ;))

> I believe we wouldn't be able to store the Constraint in a HashMap

Right now I doubt anyone is doing that already (a github search would confirm, but I'd expect this to be 0), so my question is whether there is a real example of why someone might need to do that.

> And also a user wouldn't be able to do this:

Hrm. I'm curious about this one too. Given that widgets are consumed, why would they need to be stored / compared for Eq? Mostly I'm trying to work out if that use case is more important than making this accept floats.

Could we implement Hash / Eq manually to avoid this problem?

examples/constraints.rs
kdheepak: I agree about the layout container. It lends itself to thinking about it as HTML-like containers.

If something like that existed, hypothetically we'd be able to write purely declarative UI like:

```
render!(

   <vertical class="fixed-8 fixed-8">
      
      <Example> { [Fixed(40), Proportional(0)] } </Example>
      <Example> { [Fixed(20), Fixed(20), Proportional(0)] } </Example>

   </vertical>

)
```

kdheepak: Thanks for the comments! I like that way of thinking. I made the changes in the example file.

kdheepak:
I think by expressing constraints in f64, the results won't change at all or all that much, so I'm fine with leaving it as is. It would be nice to have `Percentage`, `Ratio` and `Proportional` represented in f64 values though.

I'm not sure of the full implications of `Hash` and `Eq`, so I'd like to leave a `Hash` and `Eq` refactor out of this PR if that's possible. We'd probably want to change it for the other variants too so we can discuss it then?

examples/constraints.rs
joshka: Assuming we only have one type in the container, this works:
```rust
use crate::{prelude::*, widgets::*};

pub struct LayoutWidget<T: Widget> {
    layout: Layout,
    children: Vec<T>,
}

impl<T: Widget> LayoutWidget<T> {
    pub fn new<Iter: IntoIterator<Item = T>>(layout: Layout, children: Iter) -> Self {
        Self {
            layout,
            children: children.into_iter().collect(),
        }
    }
}

impl<T: Widget> Widget for LayoutWidget<T> {
        let areas = self.layout.split(area);
        for (child, area) in self.children.into_iter().zip(areas.iter()) {
            child.render(*area, buf);
        }
    }
}
```

```rust
fn render_fixed_example(&self, area: Rect, buf: &mut Buffer) {
        // let [example1, example2, _] = area.split(&Layout::vertical([Fixed(8); 3]));
        // Example::new([Fixed(40), Proportional(0)]).render(example1, buf);
        // Example::new([Fixed(20), Fixed(20), Proportional(0)]).render(example2, buf);
        LayoutWidget::new(
            Layout::vertical([Fixed(8); 3]),
            [
                Example::new([Fixed(40), Proportional(0)]),
                Example::new([Fixed(20), Fixed(20), Proportional(0)]),
            ],
        )
        .render(area, buf);
    }
```

joshka:
> I think by expressing constraints in f64, the results won't change at all or all that much, so I'm fine with leaving it as is. It would be nice to have `Percentage`, `Ratio` and `Proportional` represented in f64 values though.
> 
> I'm not sure of the full implications of `Hash` and `Eq`, so I'd like to leave a `Hash` and `Eq` refactor out of this PR if that's possible. We'd probably want to change it for the other variants too so we can discuss it then?

Sounds good.

src/layout/constraint.rs
```diff
@@ -2,97 +2,210 @@ use std::fmt::{self, Display};
...
+    /// │   13 px   ││   12 px  ││   13 px   ││   12 px  │
+    /// └───────────┘└──────────┘└───────────┘└──────────┘
     /// ```
     Ratio(u32, u32),
-    /// Apply no more than the given amount (currently roughly equal to [Constraint::Max], but less
-    /// consistent)
+    /// Applies a fixed size to the element
+    ///
+    /// The element size is set to the specified amount as
+    /// a strict requirement.
```
joshka, line 90: Nit: wrapping (here and below)

src/layout/constraint.rs
joshka, line 1: Can you add some docs about the meaning of "strength" in this context (strong, medium, etc.) Assume the reader doesn't know what cassowary is (and shouldn't need to).

src/layout/layout.rs
```diff
@@ -1212,5 +1277,288 @@ mod tests {
...
+            ]));
+            assert_eq!([a.width, b.width, c.width], [33, 33, 34]);
+
+            // cassowary implementation tends to put excess in last variable
+            let [a, b, c] = Rect::new(0, 0, 100, 1).split(&Layout::horizontal([
+                Fixed(25),
+                Fixed(25),
+                Fixed(25),
+            ]));
+            assert_eq!([a.width, b.width, c.width], [25, 25, 50]);
```
joshka, line 1297: This still seems a bit confusing. I know we have segment_size, but it really feels weird that Fixed doesn't have higher priority than that.

src/layout/layout.rs
```diff
@@ -1212,5 +1277,288 @@ mod tests {
...
+
+            // 3 lengths for reference
+            // cassowary implementation tends to put excess in last variable
+            let [a, b, c] = Rect::new(0, 0, 100, 1).split(&Layout::horizontal([
+                Length(25),
+                Length(25),
+                Length(25),
+            ]));
+            assert_eq!([a.width, b.width, c.width], [25, 25, 50]);
```
joshka, line 1319: It might be useful to pull all of the tests that currently already pass out from this PR into one that we can merge directly. That way this PR will clearly make the change in behavior more easy to discern from the unchanged behavior.

src/layout/layout.rs
```diff
@@ -1212,5 +1277,288 @@ mod tests {
...
+
+            // fixed with length
+            // this test gives different results when run over and over again
+            //
+            // assert_eq!([a.width, b.width], [25, 50])
+            // assert_eq!([a.width, b.width], [50, 25])
```
joshka, line 1325: ugh - could we clean that problem up somehow in another PR?

joshka (approved):
I think this is pretty close to being done. Docs would be the main thing to get it over the line. Otherwise I'm not too concerned about the remaining issues.

src/layout/constraint.rs
kdheepak: Done.

src/layout/constraint.rs
kdheepak: I added this to the docstring:

```
/// Constraints are prioritized in the following order:
///
/// 1. [`Constraint::Fixed`]
/// 2. [`Constraint::Min`] / [`Constraint::Max`]
/// 3. [`Constraint::Length`] / [`Constraint::Percentage`] / [`Constraint::Ratio`]
/// 4. [`Constraint::Proportional`]
```

I can repeat it in each variant if we want. Let's leave this documentation as it is right now. I think in a separate PR, I'd like to revisit all the weights and we can write the docs accordingly for that.

src/layout/layout.rs
kdheepak: What are you intending this should do? Something like this?

```rust
let [a, b, c] = Rect::new(0, 0, 100, 1).split(&Layout::horizontal([
    Fixed(25),
    Fixed(25),
    Fixed(25),
]));
assert_eq!([a.width, b.width, c.width], [33, 34, 33]);
```

I have a solution for this (by introducing a invisible `Rect` at the end) but it'll have to be a separate PR. That will  change the behavior of all the tests. We would also define an alignment on the layout itself. Something like `Layout::vertical(constraints).aligment(Alignment::Start)` where alignment could be `Start`, `Center` or `End`.

src/layout/layout.rs
kdheepak: Yes, I have an idea for this by introducing a single `Rect` at the end. It would build on top of this PR.

src/layout/layout.rs
kdheepak: I've added this test and more new tests into a separate PR

https://github.com/ratatui-org/ratatui/pull/802

src/layout/layout.rs
kdheepak: I'll address the new behavior in the next PR.

kdheepak:
>  Docs would be the main thing to get it over the line. 

I'll improve docs in the PR where I rework the weights.

joshka (approved):
LGTM

kdheepak:
Thank you for all the feedback!

PR #786, opened 2024-01-11
feat(style): add Material Design colors

joshka:
Enable the `material` feature flag to use the colors from the Material
design color scheme. This adds a dependency on the `matdesign-color`
crate. The colors are available in the `style::material` module.

See https://m2.material.io/design/color/the-color-system.html#tools-for-picking-colors
for more information.

```rust
use ratatui::style::material::BLUE_GRAY;
Line::styled("Hello", BLUE_GRAY.c500);
```

### DIFF
--- a/src/style/palette.rs
+++ b/src/style/palette.rs
@@ -1,3 +1,4 @@
 //! A module for defining color palettes.
 
+pub mod material;
 pub mod tailwind;
--- a/src/style/palette/material.rs
+++ b/src/style/palette/material.rs
@@ -1,0 +1,606 @@
+//! Material design color palettes.
+//!
+//! Represents the colors from the 2014 [Material design color palettes][palettes] by Google.
+//!
+//! [palettes]: https://m2.material.io/design/color/the-color-system.html#tools-for-picking-colors
+//!
+//! There are 16 palettes with accent colors, and 3 palettes without accent colors. Each palette
+//! has 10 colors, with variants from 50 to 900. The accent palettes also have 4 accent colors
+//! with variants from 100 to 700. Black and White are also included for completeness and to avoid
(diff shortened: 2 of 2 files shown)


joshka:
image

Cargo.toml
```diff
@@ -72,18 +73,24 @@ termion = ["dep:termion"]
...
 serde = ["dep:serde", "bitflags/serde"]
 
 ## enables the [`border!`] macro.
 macros = []
 
-## enables all widgets.
-all-widgets = ["widget-calendar"]
+## Material design color scheme. Enables the [`material`](crate::style::material) module and adds a
+## dependency on the [`matdesign-color`](https://crates.io/crates/matdesign-color) crate.
+material = ["dep:matdesign-color"]
```
kdheepak, line 87: I'm assuming we can include a feature for tailwindcss in the future?

kdheepak (approved):

Cargo.toml
joshka: on the way...

joshka:
I think it might be worth dropping the external dependency for this one as the documentation to make it pretty weighs more than the included code..

kdheepak:
I was going to suggest the same thing! I think if it doesn't use a crate, we don't even need to put it behind a feature flag, right?

orhun (approved):

Valentin271 (commented):
Themes don't seem to add to much weight to the crate. I know I don't need themes at all so it's your call to put it behind a feature flag.

But I'd like to have this in a `theme` submodule or similar.

joshka:
I removed the feature flags.

Valentin271 (approved):

Valentin271 (approved):

PR #788, opened 2024-01-11
feat: Change priority of constraints and add `split_with_spacers` ✨

kdheepak:
Follow up to https://github.com/ratatui-org/ratatui/pull/783

This PR introduces different priorities for each kind of constraint. This PR also adds tests that specifies this behavior. This PR resolves a number of broken tests. 

Fixes https://github.com/ratatui-org/ratatui/issues/827

With this PR, the layout algorithm will do the following in order:

1. Ensure that all the segments are within the user provided area and ensure that all segments and spacers are aligned next to each other
2. if a user provides a `layout.spacing`, it will enforce it.
3. if a user provides a `Fixed(v)` constraint, it will enforce it. 
4. `Min` / `Max` binding inequality constraints
5. ensure proportional elements are all proportional to each other
6. `Length`
7. `Percentage`
8. `Ratio`
9. collapse `Min` or collapse `Max`
10. grow spacers as much as possible
11. grow `Proportional` as much as possible

This PR also returns the space areas as `Rects` to the user. Users can then draw into the spacers as they see fit (thanks @joshka for the idea). Here's a screenshot with the modified flex example:

image

## Motivation

Cassowary is an [constraint solving](https://en.wikipedia.org/wiki/Constraint_solving) algorithm that solves systems of linear equalities and [inequalities](https://en.wikipedia.org/wiki/Linear_inequality).

Numerical stability is important in solving systems of equations, especially when using iterative methods or numerical algorithms.

We want to strive for numerical stability, i.e. "stable solutions". Stable solutions are ones that are always reproducible on different architectures, ones where a small perturbation in the input does not lead to a large change in the output, etc.

For terminal user interfaces, we are bound to integer values as the solution, so a truly stable solution for all situations is never going to be possible.

Take this test for example:

```rust
let [_a, _b, c] = Rect::new(0, 0, 100, 1).split(&Layout::horizontal([
    Length(25),
    Length(25),
    Fixed(25),
]));
assert_eq!(c.width, 25);
```

Running this multiple times gives me different results. This is an example of an unstable solution.

Some of these are unavoidable but we can greatly minimize how number of times a user can create a layout that has an unstable solution.

We also want to allow users to be expressive in how they make layouts, and allow for making any kind of layout they want. Finally, we want layout algorithm's solution to be predictable and intuitive. 

The easiest way to think about cassowary layouts is that there is a "cost" associated with not satisfying a constraint and the cassowary algorithm tries to minimize that "cost". In cassowary, the "cost" is proportional to the strength of the constraint. 

Take the following example:

```rust
x == 20 // cost = 100
y == 20 // cost = 1
x + y == 50 // cost = 100
```

Cassowary will choose the values `x = 20` and `y = 30` to solve this system since that minimizes the "cost".

```rust
#[test]
fn test_solver() {
    use super::*;
    let mut solver = Solver::new();
    let x = Variable::new();
    let y = Variable::new();

    solver.add_constraint(x | EQ(100.0) | 20.0).unwrap();
    solver.add_constraint(y | EQ(1.0) | 20.0).unwrap();
    solver.add_constraint(x + y | EQ(100.0) | 50.0).unwrap();

    let changes: HashMap<Variable, f64> = solver.fetch_changes().iter().copied().collect();

    let x = changes.get(&x).unwrap_or(&0.0).round() as u16;
    let y = changes.get(&y).unwrap_or(&0.0).round() as u16;

    println!("x = {x}");
    println!("y = {y}");
}
```

image

Changing the costs can change the results entirely, for example if we switched the costs around like so:

```rust
x == 20 // cost = 100
y == 20 // cost = 100
x + y == 50 // cost = 1
```

i.e.

```diff
         let y = Variable::new();
 
         solver.add_constraint(x | EQ(100.0) | 20.0).unwrap();
-        solver.add_constraint(y | EQ(1.0) | 20.0).unwrap();
-        solver.add_constraint(x + y | EQ(100.0) | 50.0).unwrap();
+        solver.add_constraint(y | EQ(100.0) | 20.0).unwrap();
+        solver.add_constraint(x + y | EQ(1.0) | 50.0).unwrap();
 
         let changes: HashMap<Variable, f64> = solver.fetch_changes().iter().copied().collect();
```

image

When the solver can't satisfy all of the constraints, it will pick at least one of the weakest constraints to violate (i.e. the lowest "cost" to violate). 
In this case, there's just one constraint that is clearly weaker than the others, so that is violated.

However if all the constraints are equally weighted, cassowary can, for efficiency reasons, pick any of the lowest equally weighted constraints to violate nondeterministically. 

Take this set of equations for example:

```rust
        solver.add_constraint(x | EQ(100.0) | 20.0).unwrap();
        solver.add_constraint(y | EQ(100.0) | 20.0).unwrap();
        solver.add_constraint(x + y | EQ(100.0) | 50.0).unwrap();
```

Solving it gives:

image

Switching the order around however:

```rust
        solver.add_constraint(x | EQ(100.0) | 20.0).unwrap();
        solver.add_constraint(x + y | EQ(100.0) | 50.0).unwrap();
        solver.add_constraint(y | EQ(100.0) | 20.0).unwrap();
```

gives:

image

here again with a different order:

```rust
        solver.add_constraint(x + y | EQ(100.0) | 50.0).unwrap();
        solver.add_constraint(y | EQ(100.0) | 20.0).unwrap();
        solver.add_constraint(x | EQ(100.0) | 20.0).unwrap();
```

we get a different result:

image

In this case we are deterministically able to violate a different constraint based on the order in which we add constraints. 

When cassowary violates constraints, it _tends_ to violate the last constraint of the lowest weight. However, this is not deterministic and not guaranteed to happen. This unstable non-deterministic implementation detail is the current basis of `Flex::StretchLast` (this is changed to be more deterministic in this PR).

Due to the nature of the algorithm, "in-between" solutions, although just as valid, are not picked.

## Goal

We want the layout solution to provide:

- numerical stable solution
- intuitive behavior
- predictable behavior

Given a simple set of rules to remember, a user should be able to predict exactly what a layout is going to do.

For example, if `[Length(x), Fixed(y)]` results in two rects of widths `x` and `y`, `[Fixed(y), Length(x)]` should result in two rects of widths `y` and `x`. For this, we need to ensure that all constraints have different "strengths" or "costs".

This PR introduces a `strengths` module that has "default" weights that give stable solutions as well as predictable behavior.


joshka:
Seems like fixed and flex are still in this.

kdheepak:
Ah, I assumed we wanted to merge the other one first. I'm pushing the change with just the weights changed in this one. We'll have to resolve some conflicts to merge both of them but I can do that regardless of the order.

joshka (changes requested):
The commit message should document why we're doing this and the impact that would likely occur to layouts.

kdheepak:
I want to clarify want I mean by unstable. If you have a width of a `100` and split it using these constraints:

```
[Length(25), Length(25)]
```

you will likely always get the exact same solution, i.e. widths of `[25, 75]`. But this solution is technically unstable.

> (If solver can't satisfy all of the constraints) it will pick at least one of the weakest constraints to violate. In this case it will be one or both of the preferred widths. For efficiency reasons this is picked nondeterministically, so there are two possible results. 
> Due to the nature of the algorithm, "in-between" solutions, although just as valid, are not picked.

In this case there are `51` solutions that all are valid and possible that have equal costs according to the solver. The cost of the solution `[25, 75]` is the same as `[26, 74]`, `[27, 73]`, ..., `[75, 25]`. 

Because of an implementation detail, the dependency we are using tends to pick `[25, 75]` (i.e. what `StretchLast` does) in this case consistently. But for more complex cases that is not guaranteed to happen.

All of that is to say that `StretchLast` is currently an implementation detail and is unstable, and I recommend that after this PR is merged, we should change the default to `Stretch` and remove `StretchLast`.

joshka:
> All of that is to say that StretchLast is currently an implementation detail and is unstable, and I recommend that after this PR is merged, we should change the default to Stretch and remove StretchLast.

If I were to guess about a default layout, I'd intuitively think that  [Length(10) ; 3] would only allocate 30 cells, it would not stretch those.
And [Length(10), Min(0), Length(10)] would stretch regardless of Flex

So perhaps there's a missing Flex option?

kdheepak:
> And [Length(10), Min(0), Length(10)] would stretch regardless of Flex (Edit: I could be wrong about this)

Unfortunately this will not stretch when using `Flex::Start` because we currently have a constraint that tries to make `Min(0)` as close to `0` as possible if all other constraints are satisfied. 

I personally think `Stretch` is a better default for the way all the constraints we have work with each other. More importantly, it will also not break any UI for existing users because it will behave almost identically to `StretchLast`. 

When we move to taffy I think we should consider using `Start` as the default.

src/layout/layout.rs
```diff
@@ -812,53 +939,184 @@ impl Layout {
...
+                .borrow_mut()
+                .get_or_insert((area, self.clone()), || {
+                    Self::try_split(area, self).expect("failed to split")
+                })
+                .clone()
+            })
+            .0
+    }
+
+    pub fn split_with_spacers(&self, area: Rect) -> (Rects, Rects) {
```
kdheepak, line 1097: I added a new `split_with_spacers` that returns both the `Rects` for the constraints the user passed in and spacers. This is also stored in the cache.

```rust
let blocks = Layout::horizontal([Length(20), Min(0)]).spacing(5).split_with_spacers(area);

let (blocks, spacers) = Layout::horizontal([Length(20), Min(0)]).spacing(5).split_with_spacers(area);
```

This function is in this PR for PR reviewers to be able to review how spacers work using the flex example. 

For the upcoming release, we can either:

1. remove this function and prevent users from accessing spacer rects
1. rename this function and leave it in
1. break `layout.split(area)` and have it return both blocks and spacers
1. Make `area.split(&Layout)` return both blocks and spacers

Valentin271: I like the name and the usage, I think we should keep it and make it available to the public API at some point.

For the upcoming release, maybe make it private or unstable just to have the time to test it. But if we're confident this will work for .26 then include it now.

src/layout/layout.rs
```diff
@@ -812,53 +939,184 @@ impl Layout {
...
+                        NonZeroUsize::new(Self::DEFAULT_CACHE_SIZE).unwrap(),
+                    ))
+                })
+                .borrow_mut()
+                .get_or_insert((area, self.clone()), || {
+                    Self::try_split(area, self).expect("failed to split")
+                })
+                .clone()
+            })
+            .0
```
Valentin271, line 1094: If we keep `split_with_spacers` (which I think we should), this could be

```rust
self.split_with_spacer(area).0;
```

kdheepak: Good idea!

src/layout/layout.rs
kdheepak: I'm fairly happy with the implementation now. The previous approach of implementing spacing was isolated from flex (and was taking advantage of `Fixed`) but now spacers are a core part of how the `Flex` features work, and now it makes sense to expose this internal state to users.

examples/flex.rs
```diff
@@ -98,15 +109,9 @@ enum SelectedTab {
 
 fn main() -> Result<()> {
     // assuming the user changes spacing about a 100 times or so
-    Layout::init_cache(EXAMPLE_DATA.len() * SelectedTab::iter().len() * 100);
```
kdheepak, line 101: TODO: revert this change before merging.

src/layout/layout.rs
```diff
@@ -889,12 +1095,138 @@ impl Element {
...
+    pub proportional_scaling_equality: f64,
+    pub fixed_value_equality: f64,
+    pub min_max_inequality: f64,
+    pub length_equality: f64,
+    pub percentage_equality: f64,
+    pub ratio_equality: f64,
+    pub min_max_equality: f64,
+    pub proportional_grower: f64,
+    pub grower: f64,
+    pub space_grower: f64,
```
kdheepak, line 1110: Better names for these fields are welcome

src/layout/constraint.rs
```diff
@@ -364,6 +366,18 @@ impl Constraint {
             .map(Constraint::Proportional)
             .collect_vec()
     }
+
+    pub fn rank(&self) -> usize {
```
Valentin271, line 370: Is this really supposed to be public? If so could we document it?

Valentin271 (approved):
I like the diagrams you put all around comments, they look really nice!

Looks good overall but as stated earlier I'm no specialist in this part of the lib.
I've tested it though (both examples and with my projects) and this works as expected.

Don't forget to put back the cache in the example.

joshka (commented):
Hey, this is 1.3K lines changed in a 2.5K line file (plus a few others), and the changes are spread across quite a few methods. This makes it pretty difficult to review well. If it was just a refactoring, then this would be fine because the metric would be that its still passes the same tests, but this is refactoring combined with new features.

In getting this over the line, I wonder if there's a way to break this into a few smaller coherent check points that implement part of the problem. Particularly, splitting this into small additive changes only, or refatoring changes only, and each perhaps a quarter or less the size of this change.

Perhaps another approach would be provide a map to review?

The problem behind this PR is that the code here is already complex, which means that understanding the start point is difficult, then adding a bunch more changes on top of that makes that many times more complex.

Maybe you have other suggestions to make it easier to review?

src/layout/constraint.rs
kdheepak: Good catch, it shouldn't be public.

src/layout/constraint.rs
```diff
@@ -364,6 +366,18 @@ impl Constraint {
...
+        match self {
+            Constraint::Fixed(_) => 0,
+            Constraint::Min(_) => 1,
+            Constraint::Max(_) => 2,
+            Constraint::Length(_) => 3,
+            Constraint::Percentage(_) => 4,
+            Constraint::Ratio(_, _) => 5,
+            Constraint::Proportional(_) => 6,
+        }
+    }
```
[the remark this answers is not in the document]
kdheepak: I'm not able to change it to `self as usize`, presumably because there's a payload attached to the enum variant? I'm getting an error that implies that I can only convert from a primitive type when I use `self as usize`.

I added a test for this.

kdheepak:
Some of the refactoring changes are related to each other, so it's harder to break it down into simpler chunks. I'll write up a map here, and open to suggestions to make it easier to understand.

Previously we had one big method called `try_split` in which a number of things happened:

https://github.com/ratatui-org/ratatui/blob/f383625f0e1cae320ae56af615f3b05c59700f93/src/layout/layout.rs#L488

---

1. `layout.spacing`:

This method conditionally added "spacers", i.e interspersed `Fixed(layout.spacing)` into the constraints:

https://github.com/ratatui-org/ratatui/blob/f383625f0e1cae320ae56af615f3b05c59700f93/src/layout/layout.rs#L498-L514

These `Fixed` constraints were only added in between constraints. At the end of the function, we `step_by(2)` to skip over these `Fixed` constraints to return the result to the user.

https://github.com/ratatui-org/ratatui/blob/f383625f0e1cae320ae56af615f3b05c59700f93/src/layout/layout.rs#L835-L857

**Takeaways** 

- The highest priority constraint we had was `Fixed`
- This was used for `layout.spacing`
- After interspersing the `Fixed` there was no way to distinguish between a user provided `Fixed` and a interspersed `Fixed` segment.

This is why a test like this was broken (note the `ne` in `assert_ne!`):

https://github.com/ratatui-org/ratatui/blob/1e755967c53e9a1803cc7fcc46ad0946c78f0eda/src/layout/layout.rs#L2173-L2186

---

2. `Flex`

This `try_split` method, also added "spacers" internally for different `Flex` layouts. Let's look at the current implementation of `Flex::Center` as an example:

https://github.com/ratatui-org/ratatui/blob/f383625f0e1cae320ae56af615f3b05c59700f93/src/layout/layout.rs#L650-L678

```plain
<------------------------------80 px (gap: 5 px)------------------------------->
┌   ┐┌──────────────────┐     ┌──────────────────┐     ┌──────────────────┐┌   ┐
     │     Fixed(20)    │     │      Min(20)     │     │      Max(20)     │
└   ┘└──────────────────┘     └──────────────────┘     └──────────────────┘└   ┘
^^^^^                                                                      ^^^^^
  └─────────────────────────flex elements────────────────────────────────────┘
```

For `Flex::Center` two spacers were added, one at the beginning and one at the end.
For `Flex::Start` and `Flex::End`, only one was added, either at the beginning or at the end.

For `Flex::SpaceBetween`, we create new spacers between every user provided segment (for `Flex::SpaceAround` we create new spacers around every segment).

https://github.com/ratatui-org/ratatui/blob/f383625f0e1cae320ae56af615f3b05c59700f93/src/layout/layout.rs#L560-L587
 
**Takeaways** 

- These "spacers" are lower strengths than `layout.spacing`, because if for example a user specifies `layout.spacing = 2` and uses `Flex::Center`, we want the user provided spacing to take priority over the filler spacers at the beginning or the end:

```plain
     ┌──────────────────┐┌   ┐┌──────────────────┐┌   ┐┌──────────────────┐
     │     Fixed(20)    │     │      Min(20)     │     │      Max(20)     │
     └──────────────────┘└   ┘└──────────────────┘└   ┘└──────────────────┘
                         ^^^^^                    ^^^^^
                           └────────────────────────┴─────────────────────Spacers
```

Put another way, even if the user has a ultra wide monitor, we want the beginning and end to take any value such that all the segments are in the center with a space of `2` between each segment.

3. `Constraint`

Finally we added the constraints for the `Constraint` segments:

https://github.com/ratatui-org/ratatui/blob/f383625f0e1cae320ae56af615f3b05c59700f93/src/layout/layout.rs#L720-L765

Before `layout.spacing`, `Flex`, `Fixed` and `Proportional` were introduced, we really used just one weights here for all the constraints (`STRONG`):

https://github.com/ratatui-org/ratatui/blob/425a65140b61695169c996784974488ad2fd16ea/src/layout/layout.rs#L458-L485

This resulted in suggestions like this, saying increasing the priority of `Length` made it more intuitive:

https://github.com/ratatui-org/ratatui/issues/354#issuecomment-1675936938

More pertinently, because of the weights being the same, we relied on cassowary implementation details a lot. This means layouts were not stable and were inconsistent, e.g. in both these cases the last constraint would be violated:

```
let area = Rect(0, 0, 100, 0);

(width_25, width_75) = area.split(&Layout::horizontal([Length(25), Percentage(25)]))

(width_25, width_75) = area.split(&Layout::horizontal([Percentage(25), Length(25)]))
```

We decided to call this `LastTakesRemainder` or `StretchLast` but what's actually happening is that the last constraint of lowest priority is the one that is least binding. Previously we basically had just one weight (`STRONG`) so everything was of equal priority. This is not guaranteed behavior though, and we would see jumps in the layout because of this. See the `flex` example from `main`: 

https://github.com/ratatui-org/ratatui/assets/1813121/06efaca3-60b3-4e57-89d1-0b85d9d3272a

When increasing spacing, when the solver recalculates the layouts, sometimes it makes `Percentage` take priority and some times it makes `Length` take priority. This effect is more pronounced when caching is small.

---

In an upcoming comment, I'll describe how this PR addresses these issues.

src/layout/layout.rs
```diff
@@ -11,7 +11,8 @@ use lru::LruCache;
 use super::{Flex, SegmentSize};
 use crate::prelude::*;
 
-type Cache = LruCache<(Rect, Layout), Rc<[Rect]>>;
+type Rects = Rc<[Rect]>;
+type Cache = LruCache<(Rect, Layout), (Rects, Rects)>;
```
kdheepak, line 15: Previously `Cache` mapped `(Rect, Layout)` to `[Rect]` but now it maps `(Rect, Layout)` to `([Rect], [Rect])`.

The first is the rects as a solution to the segments that the user provided constraints for, and the second is the rects that are a spacers.

The second number of `Rect`s is always the first's length + 1, because spacers are added around all user provided constraints.

src/layout/layout.rs
```diff
@@ -812,53 +961,164 @@ impl Layout {
...
+    ///
+    /// let layout = Layout::default()
+    ///     .direction(Direction::Horizontal)
+    ///     .constraints([Constraint::Ratio(1, 3), Constraint::Ratio(2, 3)])
+    ///     .split(Rect::new(0, 0, 9, 2));
+    /// assert_eq!(layout[..], [Rect::new(0, 0, 3, 2), Rect::new(3, 0, 6, 2)]);
+    /// ```
+    pub fn split(&self, area: Rect) -> Rects {
+        self.split_with_spacers(area).0
+    }
```
kdheepak, line 1097: `split` returns just the first element in the tuple, which contains the same results as before.

src/layout/constraint.rs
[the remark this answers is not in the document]
joshka: Got it. I didn't realize that, but it makes sense.

src/layout/layout.rs
joshka: Would it be worth pulling this out into a small named type instead of a tuple?

joshka:
Thanks for the map of the changes, I feel like I have a much better idea about how to review this now.

src/layout/layout.rs
```diff
@@ -436,333 +443,474 @@ impl Layout {
...
+    ///
+    /// ```plain
+    /// <----------------------------80 px (spacer: 10 px)----------------------------->
+    /// ┌──────────────────┐┌        ┐┌──────────────────┐┌        ┐┌──────────────────┐
+    /// │     Fixed(20)    │          │      Min(20)     │          │      Max(20)     │
+    /// └──────────────────┘└        ┘└──────────────────┘└        ┘└──────────────────┘
+    /// ^^^^^^^^^^^^^^^^^^^^          ^^^^^^^^^^^^^^^^^^^^          ^^^^^^^^^^^^^^^^^^^^
+    ///          │                              │                             │
+    ///          └──────────────────────────────┴───────────────────────────Segments
+    /// ```
```
joshka, line 466: This picture is is a lot of lines just to explain the concept of what a segment is, but doesn't provide any information about the result or actions that this method does.

I find it a bit confusing given the name "build" that this actually both creates and also adds the elements to the solver and then returns them. The interaction between layout, solver, and each element is a bit difficult to visualize here.

Working through this method and the Element::constrain call, I think the net effect of this method is to create variables to track the start and end of the element areas, and then to and cassowary constraints that ensure those variables are within the range.

I think this might be a case where overly splitting and documenting makes it more difficult to sense what's going on here.

How about something like the following, which I think would go a long way to making this generally super easy to follow:

```rust
let (start, end, length) = match (self.direction) {  
    Direction::Horizontal => (inner.left().into(), inner.right().into(), inner.width.into()),  
    Direction::Vertical => (inner.top().into(), inner.bottom().into(), inner.height.into()),  
};
let segment_count = layout.constraints.len();
let spacer_count = segment_count.saturating_add(1);
let segments = iter::repeat_with(Element::new).take(segment_count).collect();
let spacers = iter::repeat_with(Element::new).take(spacer_count).collect();

// ensure that segments and spacers are within the area
solver.add_constraints(segment.iter().map(|e| e.to_within_area_constraint(start, end)));
solver.add_constraints(spacers.iter().map(|e| e.to_within_area_constraint(start, end)));

// ensure the segments and spacers are arranged left to right and interleaved
solver.add_constraints(spacers.iter().interleave(segments.iter()).tuple_windows().map((a,b) colinear_constraint(a, b)));
solver.add_constraints(spacers.iter().first().map(|e| e.start_constraint(start)));
solver.add_constraints(spacers.iter().last().map(|e| e.end_constraint(end)));
```

joshka (commented):
Partial review to highlight a possible simplification before continuing the review

src/layout/layout.rs
```diff
@@ -889,12 +1149,138 @@ impl Element {
...
+            // │<= Min(x) =>│
+            // └────────────┘
+            grower,
+            // ┌       ┐
+            //  <= x =>
+            // └       ┘
+            space_grower,
+        ) = (
+            strength::create(1000.0, 1000.0, 999.0, 1.0), // REQUIRED - 1.0
+            strength::create(1000.0, 1000.0, 999.0, 1.0), // REQUIRED - 1.0
```
joshka, line 1232: Why not just use the comment version directly? They seem more readable / understanable?

src/layout/layout.rs
```diff
@@ -889,12 +1149,138 @@ impl Element {
...
+            && self.ratio_equality > self.min_max_equality
+            && self.min_max_inequality > self.proportional_grower
+            && self.proportional_grower > self.grower
+            && self.grower > self.space_grower
+    }
+}
+
+impl Default for StrengthSet {
+    fn default() -> Self {
+        let (
```
joshka, line 1185: I think this method should probably just be 13 LoC (if you keep it as a struct and not a module):
```rust
Self {
    spacer_size_equality: REQUIRED - 1,
    proportional_scaling_equality: REQUIRED - 1,
    fixed_value_equality: REQUIRED / 10.0,
    min_max_inequality: STRONG * 10.0
    length_equality: STRONG / 10.0
    percentage_equality: MEDIUM * 10.0,  
    ratio_equality: MEDIUM,
    min_max_equality: MEDIUM / 10.0,
    grower: WEAK * 10.0,
    space_grower: WEAK,
    proportional_grower: WEAK / 10.0,
}
```

src/layout/layout.rs
```diff
@@ -889,12 +1149,138 @@ impl Element {
...
+            proportional_scaling_equality,
+            fixed_value_equality,
+            min_max_inequality,
+            length_equality,
+            percentage_equality,
+            ratio_equality,
+            min_max_equality,
+            grower,
+            space_grower,
+            proportional_grower,
```
joshka, line 1255: nit: out of order

src/layout/layout.rs
```diff
@@ -889,12 +1149,138 @@ impl Element {
...
+    pub fixed_value_equality: f64,
+    pub min_max_inequality: f64,
+    pub length_equality: f64,
+    pub percentage_equality: f64,
+    pub ratio_equality: f64,
+    pub min_max_equality: f64,
+    pub proportional_grower: f64,
+    pub grower: f64,
+    pub space_grower: f64,
+}
```
joshka, line 1165: Given that these are basically constant and not user configurable (YAGNI for now?):
```rust
mod strengths {
    /// The strength used to constrain spacers to approximately equal size
    /// 
    /// ```plain
    /// ┌      ┐┌───┐┌      ┐┌───┐┌       ┐
    ///    eq   │   │   eq   │   │   eq
    /// └      ┘└───┘└      ┘└───┘└       ┘
    /// ```
    SPACER_SIZE_EQ: f64 = REQUIRED - 1.0;
    PROPORTIONAL_SCALING_EQ: f64 = REQUIRED - 1.0;
    FIXED_SIZE_EQ: f64 = REQUIRED / 10.0;
    MIN_SIZE_GE: f64 = STRONG * 10.0;
    MAX_SIZE_LE: f64 = STRONG * 10.0;
    LENGTH_SIZE_EQ: f64 = STRONG / 10.0;
    PERCENTAGE_SIZE_EQ: f64 = MEDIUM * 10.0;
    RATIO_SIZ_EQ: f64 = MEDIUM;
    MIN_SIZE_EQ: f64 = MEDIUM / 10.0;
    MAX_SIZE_EQ: f64 = MEDIUM / 10.0;
    PROPORTIONAL_GROWER: f64 = WEAK * 10.0
    GROWER: f64 = WEAK * 10.0;
    SPACE_GROWER: f64 = WEAK / 10.0;
}

File: src/layout/layout.rs [hunk 14]
```diff
@@ -1213,30 +1599,30 @@ mod tests {
             test(Rect::new(0, 0, 1, 1), &[Length(0), Length(1)], "b"); // zero, exact
             test(Rect::new(0, 0, 1, 1), &[Length(0), Length(2)], "b"); // zero, overflow
             test(Rect::new(0, 0, 1, 1), &[Length(1), Length(0)], "a"); // exact, zero
-            test(Rect::new(0, 0, 1, 1), &[Length(1), Length(1)], "a"); // exact, exact
-            test(Rect::new(0, 0, 1, 1), &[Length(1), Length(2)], "a"); // exact, overflow
+            test(Rect::new(0, 0, 1, 1), &[Length(1), Length(1)], "b"); // exact, exact with stretchlast
```
joshka, line 1602: This result seems not good.

src/layout/layout.rs
```diff
@@ -1213,30 +1599,30 @@ mod tests {
...
+            test(Rect::new(0, 0, 1, 1), &[Length(1), Length(2)], "b"); // exact, overflow with stretchlast
             test(Rect::new(0, 0, 1, 1), &[Length(2), Length(0)], "a"); // overflow, zero
-            test(Rect::new(0, 0, 1, 1), &[Length(2), Length(1)], "a"); // overflow, exact
-            test(Rect::new(0, 0, 1, 1), &[Length(2), Length(2)], "a"); // overflow, overflow
+            test(Rect::new(0, 0, 1, 1), &[Length(2), Length(1)], "b"); // overflow, exact with stretch last
```
joshka, line 1605: This seems not good either.

src/layout/layout.rs
```diff
@@ -1213,30 +1599,30 @@ mod tests {
...
             test(Rect::new(0, 0, 2, 1), &[Length(0), Length(0)], "bb"); // zero, zero
             test(Rect::new(0, 0, 2, 1), &[Length(0), Length(1)], "bb"); // zero, underflow
             test(Rect::new(0, 0, 2, 1), &[Length(0), Length(2)], "bb"); // zero, exact
             test(Rect::new(0, 0, 2, 1), &[Length(0), Length(3)], "bb"); // zero, overflow
             test(Rect::new(0, 0, 2, 1), &[Length(1), Length(0)], "ab"); // underflow, zero
             test(Rect::new(0, 0, 2, 1), &[Length(1), Length(1)], "ab"); // underflow, underflow
-            test(Rect::new(0, 0, 2, 1), &[Length(1), Length(2)], "ab"); // underflow, exact
-            test(Rect::new(0, 0, 2, 1), &[Length(1), Length(3)], "ab"); // underflow, overflow
+            test(Rect::new(0, 0, 2, 1), &[Length(1), Length(2)], "bb"); // underflow, exact with stretchlast
+            test(Rect::new(0, 0, 2, 1), &[Length(1), Length(3)], "bb"); // underflow, overflow with stretchlast
```
joshka, line 1615: Same

src/layout/layout.rs
```diff
@@ -1213,30 +1599,30 @@ mod tests {
...
             test(Rect::new(0, 0, 2, 1), &[Length(2), Length(0)], "aa"); // exact, zero
-            test(Rect::new(0, 0, 2, 1), &[Length(2), Length(1)], "aa"); // exact, underflow
-            test(Rect::new(0, 0, 2, 1), &[Length(2), Length(2)], "aa"); // exact, exact
-            test(Rect::new(0, 0, 2, 1), &[Length(2), Length(3)], "aa"); // exact, overflow
+            test(Rect::new(0, 0, 2, 1), &[Length(2), Length(1)], "ab"); // exact, underflow with stretch last
+            test(Rect::new(0, 0, 2, 1), &[Length(2), Length(2)], "bb"); // exact, exact with stretchlast
```
joshka, line 1618: bad

src/layout/layout.rs
```diff
@@ -1213,30 +1599,30 @@ mod tests {
...
+            test(Rect::new(0, 0, 2, 1), &[Length(2), Length(3)], "bb"); // exact, overflow with stretchlast
```
joshka, line 1619: bad

src/layout/layout.rs
```diff
@@ -1213,30 +1599,30 @@ mod tests {
...
             test(Rect::new(0, 0, 2, 1), &[Length(3), Length(0)], "aa"); // overflow, zero
-            test(Rect::new(0, 0, 2, 1), &[Length(3), Length(1)], "aa"); // overflow, underflow
-            test(Rect::new(0, 0, 2, 1), &[Length(3), Length(2)], "aa"); // overflow, exact
-            test(Rect::new(0, 0, 2, 1), &[Length(3), Length(3)], "aa"); // overflow, overflow
+            test(Rect::new(0, 0, 2, 1), &[Length(3), Length(1)], "ab"); // overflow, underflow with stretchlast
+            test(Rect::new(0, 0, 2, 1), &[Length(3), Length(2)], "bb"); // overflow, exact with stretchlast
+            test(Rect::new(0, 0, 2, 1), &[Length(3), Length(3)], "bb"); // overflow, overflow with stretchlast
```
joshka, line 1623: bad

src/layout/layout.rs
```diff
@@ -1213,30 +1599,30 @@ mod tests {
...
 
-            test(Rect::new(0, 0, 3, 1), &[Length(2), Length(2)], "aab");
+            test(Rect::new(0, 0, 3, 1), &[Length(2), Length(2)], "abb"); // with stretchlast
```
joshka, line 1625: bad

src/layout/layout.rs
```diff
@@ -1254,30 +1640,30 @@ mod tests {
             test(Rect::new(0, 0, 1, 1), &[Max(0), Max(1)], "b"); // zero, exact
             test(Rect::new(0, 0, 1, 1), &[Max(0), Max(2)], "b"); // zero, overflow
             test(Rect::new(0, 0, 1, 1), &[Max(1), Max(0)], "a"); // exact, zero
-            test(Rect::new(0, 0, 1, 1), &[Max(1), Max(1)], "a"); // exact, exact
-            test(Rect::new(0, 0, 1, 1), &[Max(1), Max(2)], "a"); // exact, overflow
+            test(Rect::new(0, 0, 1, 1), &[Max(1), Max(1)], "b"); // exact, exact with stretchlast
+            test(Rect::new(0, 0, 1, 1), &[Max(1), Max(2)], "b"); // exact, overflow with stretchlast
```
joshka, line 1644: Lots more bad results. I know these are edge cases, but they tend to be intuitively hitting the wrong side of what I'd expect, which suggests that they will also make off by one decisions all over the place. "Centered" layout being 1 column to the right etc. I'm ommitting further comments about bad changes below this point. Also note, this is probably subjective, but I can't find an obvious justification for why these results should be preferred over the alternatives.

src/layout/constraint.rs
```diff
@@ -15,8 +15,10 @@ use itertools::Itertools;
 ///
 /// 1. [`Constraint::Fixed`]
 /// 2. [`Constraint::Min`] / [`Constraint::Max`]
```
joshka, line 17: Is it worth splitting these too?

joshka (commented):
Finished review except tests - which need some feedback before continuing to review.

I think there's some strong refactoring towards readability to be had here, but overall I like where this is going.

I'm concerned about how the default edge cases have changed for the worse (e.g. showing the last character in an over constrained ["aa", "bb"] test instead of the first sems intuitively wrong.

src/layout/layout.rs
kdheepak: In all of these bad results, what you are saying is "intuitive" doesn't match the what a user would expect from the name `StretchLast`. `StretchLast` is supposed to mean excess into last.

fwiw, if I add a new `test_with_stretch` like below, I get the previous behavior:

```rust
        #[track_caller]
        fn test_with_stretch(area: Rect, constraints: &[Constraint], expected: &str) {
                .constraints(constraints)
                .flex(Flex::Stretch) // solve with `Flex::Stretch`
                .split(area);
            for (i, c) in ('a'..='z').take(constraints.len()).enumerate() {
                let s: String = c.to_string().repeat(area.width as usize);
                Paragraph::new(s).render(layout[i], &mut buffer);
            }
            let expected = Buffer::with_lines(vec![expected]);
        }
```

What you are calling intuitive is unfortunately an implementation defined behavior, and relying on that can lead to unstable solutions unfortunately.

My recommendation is either

- accept the new behavior as `StretchLast` (it is stable)
- OR make `Flex::Stretch` the default

joshka: The issues are:
- this changes existing behavior, which means apps with painfully tweaked layouts now change and that annoys devs
- the behavior of stretch last doesn't imply last area is more important, it implies that this applies only once all other constraints have been met - I can't work out a way where you would describe how `Length(3), Length(2) -> "bb"` makes sense ever. The first area is bigger than the second, this should result in either "ab" or "aa", and never "bb".

> What you are calling intuitive is unfortunately an implementation defined behavior, and relying on that can lead to unstable solutions unfortunately.

This was stable previously though, so this is a regression without a way to fix right? Don't neither of your options result in behavior that matches the existing layout, or am I missing something?

src/layout/layout.rs
```diff
@@ -812,53 +961,164 @@ impl Layout {
...
+        // Solves will just one constraint different would not need to resolve the entire layout.
+        //
+        // The disadvantage of this approach is that it requires tracking which constraints were
+        // added, and which variables they correspond to.
+        // This will also require introducing and maintaining the API for users to do so.
+        //
+        // Currently we don't support that use case and do not intend to support it in the future,
+        // and instead we require that the user re-solve the layout every time they call `split`.
+        // To minimize the time it takes to solve the same problem over and over again, we
+        // cache the `Layout` struct along with the results.
```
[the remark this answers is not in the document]
kdheepak: I put this in here mainly because if someone wants to refactor cassowary in the future or improve the implementation, they may try to make it such that we create an instance of the solver just once; and I wanted to preemptively hedge against that.

src/layout/layout.rs
kdheepak: It wasn't a stable solution in general, it just happened to be reproducible for this particular test. There's no guarantee that for any `[Length(x), Length(y)]` or `[Length(x), Length(y), Length(z)]` or any number of constraints for any size of the rect that you'll get the same behavior. 

I also think using two or multiple `Length`s like this is unlikely? Users that have tweaked their layouts with this are depending on an implementation specific behavior. Similarly, we have tweaked the priorities for `Percentage` and `Ratio` in this PR but our tests do not capture all the different ways this could behave. I think commonly used patterns will be unaffected by this change in this PR.

> The first area is bigger than the second, this should result in either "ab" or "aa", and never "bb".

Can I ask you to think what should the following do in these 3 cases? (think about what a user coming to Ratatui completely new to layouts / terminal user interfaces etc would expect):

```rust
for width in [10, 50, 100] {
    let area = Rect(0, 0, width, 0);
    let (a1, a2) = area.split(Layout::horizontal([Length(30), Length(70)]))
}
```

fwiw, if you comment these lines in `StretchLast`, you'll get the old behavior:

```rust
if let Some((segment, _)) = segments
    .iter()
    .zip(&layout.constraints)
    .sorted_by(|(_, c1), (_, c2)| Ord::cmp(&c1.rank(), &c2.rank()))
    .next_back()
{
    solver.add_constraint(segment.grows_to(size))?;
}
```

If you do comment it, users will get unstable layouts. An example where that removing this block of code will result in this test pass / fail randomly:

```rust
#[case::previously_unstable_test(vec![50, 25, 25], vec![Length(25), Length(25), Fixed(25)])] 
#[case::previously_unstable_test(vec![25, 50, 25], vec![Length(25), Length(25), Fixed(25)])]
```

In my opinion, having a layout that is random and unpredictable like this is worse than a breaking change in behavior.

---

If we absolutely don't want to break existing user behavior, I'd recommend changing `Flex::StretchLast` to `Flex::Unstable` and removing the block of code from above, and working toward changing the default in the future.

src/layout/layout.rs
```diff
@@ -436,333 +443,267 @@ impl Layout {
...
+                //     AddConstraintError::InternalSolverError("Unable to get last segment"),
+                // )?)?;
             }
             Flex::StretchLast => {
-                // this is the default behavior
-                // by default cassowary tends to put excess into the last constraint of the lowest
-                // priority.
-                if let Some(first) = elements.first() {
-                    solver.add_constraint(first.start | EQ(REQUIRED) | area_start)?;
+                for spacer in spacers.iter().skip(1).rev().skip(1) {
```
joshka, line 557: What's the intent of this? It seems a bit magical (opaque in operation).

src/layout/layout.rs
```diff
@@ -436,333 +443,267 @@ impl Layout {
...
+                    solver.add_constraint(last_segment.end_sticks_to(end))?
+                }
+                if let Some((segment, _)) = segments
+                    .iter()
+                    .zip(&layout.constraints)
+                    .sorted_by(|(_, c1), (_, c2)| Ord::cmp(&c1.rank(), &c2.rank()))
+                    .next_back()
+                {
+                    solver.add_constraint(segment.grows_to(size))?;
                 }
```
joshka, line 575: This seems hard to parse too.

src/layout/layout.rs
kdheepak: We have to skip the first and last spacers when adding constraints for everything except `SpaceAround`. Would a comment suffice in clarifying intent?

src/layout/layout.rs
kdheepak: Good idea! I was hoping to make it configurable for easier comparison between `main` and this branch but that never came to fruition.

src/layout/constraint.rs
kdheepak: `Min` and `Max` are equal priority because I don't expect users to do something like `[Min(100), Max(0)]` and expect a stable solution. But because I chose `Min`'s rank as higher than `Max`, I've updated the docstring to indicate that `Min` has a higher priority.

src/layout/layout.rs
joshka: yep, something like `//ignore first and last spacer` would do it, and perhaps combine with `spacers[1..spacers.len-1]` as this is a slice.

joshka: note - I just realized why that doesn't work :)
https://docs.rs/itertools/0.11.0/itertools/trait.Itertools.html#method.with_position is an approach that also could help:
```rust
spacers.with_position().filter(|(_,position) position = Position::Middle).map((e,_) e)
```
Maybe it's just as complex?

src/layout/layout.rs
kdheepak: > My takeaway on that hypothetical is that intuitively we should consider lengths as being proportional when constraints cannot be met.

The problem is cassowary (or linear programming in general) cannot handle "conditionals". There's no way in the formulation to say "if this, do that". We have to preprocess those cases and selectively add different constraints if we want it to behave differently.

> is there a test that fails in 0.25 like this that would be a better example?

Here's an example of an unstable test on the v0.25.0 tag:

```rust
#[test]
fn unstable_test() {
    let area = Rect::new(0, 0, 100, 0);
        .constraints([
            Constraint::Length(25),
            Constraint::Length(25),
            Constraint::Max(25),
        ])
        .split(area);
    dbg!(chunks.iter().map(|r| r.width).collect::<Vec<u16>>());
}
```

image

I think my point is that you are drawing "meaning" from these simple tests but that's not what you should do, because that's not how cassowary will behave all the time with the constraints we have in v0.25.0. In `v0.25.0` and in the current `main`, how the default solve behaves is undefined. And we've named that undefined behavior `LastTakesRemainder` / `StretchLast` because that's behavior we observed in simple test cases. But our tests are under specified. 

I think we should strive for solutions that are reproducible and stable, particularly more so because we have now included `Fixed` and `Proportional`.

kdheepak: > I'd strongly want that to give stable [38,37,25] regardless of which flex option is chosen

This is kind of why I want `Stretch` to be the default, since it is more intuitive that what happens with `StretchLast` in this PR OR in `main`.

joshka: While we can't have conditionals, doesn't this just highlight that there's probably a missing rule here though?
Let's constrain our discussion to an over-allocation that cannot be met: 70/30 or 30/70 into width 50:
Here's what that looks like from a cassowary perspective (I'm going to simplify this to ignore spacers, and just use
3 variables (in fact we only need 1 as everything else is fixed, but let's go with 3 for now - a, b, c (in the 0.25 real impl this would be a b c d with b = c = required)
1.  a = 0 (start) required
2. a <= b required
3. b <= c required
4. c = 50 (end) required
5. b - a = 70 strong
6. c - b = 30 strong
7. a >= 0, b >= 0, c >=0
8. a <=50, b <=50, c<=50

This doesn't have a solution that doesn't violate either 5 or 6, but we can add a nice default that should apply regardless of what flex type we choose:
(b - a) / 30 = (c-b / 70) medium

This is a gross simplification, but reasonable. This constraint only kicks in once the solver tries to break one of the more important constraints right? (This is basically the example in the main doc btw.)

---

Incidentally, this discussion suggests that it would be really useful to split the *split methods at the point where all the constraints are calculated, so that we can just test that part (and not from an add a `dbg!` macro approach, but an actual testable function approach - I think this might help visualize some of the odd edge case behavior better.

For that to work, we'd have to collect the constraints outside of the solver instead of directly in it as it has no getters. That's a simple but really useful thing that makes some of our talk earlier where should the call to solver.add_constraint belong lie kinda moot - it belongs we've created all the constraints.

kdheepak: > This constraint only kicks in once the solver tries to break one of the more important constraints right? (This is basically the example in the main doc btw.)

I like that change, i.e. make `Length` behave more like `Proportional`. (All the more reason to get rid of `Proportional` then?)

kdheepak: I pushed a new commit where `Length`, `Percentage` and `Ratio` are proportional.

https://github.com/ratatui-org/ratatui/pull/788/commits/5b7792fa7f6a72c9f13a7d666a75f0e482373f60

These tests still change, so it would be breaking the UI.

src/layout/layout.rs
```diff
@@ -1250,64 +1564,67 @@ mod tests {
...
+            test(Rect::new(0, 0, 2, 1), &[Min(0), Min(0)], "ab"); // zero, zero
             test(Rect::new(0, 0, 2, 1), &[Min(0), Min(1)], "bb"); // zero, underflow
             test(Rect::new(0, 0, 2, 1), &[Min(0), Min(2)], "bb"); // zero, exact
             test(Rect::new(0, 0, 2, 1), &[Min(0), Min(3)], "bb"); // zero, overflow
-            test(Rect::new(0, 0, 2, 1), &[Min(1), Min(0)], "ab"); // underflow, zero
+            test(Rect::new(0, 0, 2, 1), &[Min(1), Min(0)], "aa"); // underflow, zero
             test(Rect::new(0, 0, 2, 1), &[Min(1), Min(1)], "ab"); // underflow, underflow
             test(Rect::new(0, 0, 2, 1), &[Min(1), Min(2)], "ab"); // underflow, exact
-            test(Rect::new(0, 0, 2, 1), &[Min(1), Min(3)], "ab"); // underflow, overflow
+            test(Rect::new(0, 0, 2, 1), &[Min(1), Min(3)], "ab"); // underflow, overflow // unstable
```
kdheepak, line 1616: image

kdheepak:
Unfortunately, I'm not able to get this proportional behavior to work as expected in a stable manner.

image

Weirdly, I only get the error almost consistently when running `cargo nextest run`.  Running `cargo nextest run --verbose -- widgets::table::table::tests::render::render_with_footer_margin` fails occasionally but not often enough.

If I run `cargo test --lib -- widgets::table::table::tests::render::render_with_footer_margin --nocapture` it never seems to fail. 

I tried to reproduce the error in a separate test like this and I'm never able to get it to fail:

```rust
#[rstest]
#[case::proportional( vec![0, 0, 0, 1, 1, 1, 0],   vec![Length(0), Length(0), Length(0), Min(0), Length(1), Length(1), Length(0)])]
fn table_unstable_test(#[case] expected: Vec<u16>, #[case] constraints: Vec<Constraint>) {
    let rect = Rect::new(0, 0, 3, 1);
    let r = Layout::horizontal(constraints)
        .split(rect)
        .iter()
        .cloned()
        .map(|r| r.width)
        .collect::<Vec<u16>>();
    assert_eq!(expected, r);
}
```

kdheepak:
I've decided to decrease the scope of this PR. Previously, I was trying to make `StretchLast` also stable, but it seems like there's not a clear consensus on what we want `Flex::StretchLast` OR what we want the `Flex::default()` to do. Instead of trying to figure that out in this PR, I'm hoping we can discuss that in a future PR. We can also discuss if we want to remove `Fixed()` and `Proportional()` as part of that discussion.

This PR now just the following:

1. refactors the layout to make spacers higher priority (Fixes https://github.com/ratatui-org/ratatui/issues/827)
2. adds `split_with_spacers` to return these spacers
3. ranks `Length` higher than `Percentage` higher than `Ratio` for more stable solutions

This also means there are fewer tests that change between `main` and this PR.

src/layout/layout.rs
kdheepak: I've reverted the change for these bad results. I'm resolving these conversations.

src/layout/layout.rs
kdheepak: This code has been removed.

src/layout/layout.rs
kdheepak: I added a comment. `skip(1).rev().skip(1)` seems as clear to me as `with_position`, so I'm indifferent.

src/layout/layout.rs
```diff
@@ -436,333 +443,474 @@ impl Layout {
...
+    fn add_spacer_constraints(
+        solver: &mut Solver,
+        strengths: &StrengthSet,
+        area: Rect,
+        layout: &Layout,
+        spacers: &[Element],
+    ) -> Result<(), AddConstraintError> {
+        let (_area_start, _area_end, area_size) = layout.get_start_end_size(area);
+
+        match layout.flex {
```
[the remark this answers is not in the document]
kdheepak: I've oriented it around `flex`. I've also moved `solver.add_constraint` calls up.

src/layout/layout.rs
```diff
@@ -436,333 +443,474 @@ impl Layout {
...
-        match flex {
-            Flex::SpaceBetween => {
-                let spacers: Vec<Element> = std::iter::repeat_with(|| {
-                    Element::constrain(&mut solver, (area_start, area_end))
-                })
-                .take(elements.len().saturating_sub(1)) // one less than the number of elements
+        // <------------------------------------80 px------------------------------------->
+        // ┌   ┐┌──────────────────┐┌   ┐┌──────────────────┐┌   ┐┌──────────────────┐┌   ┐
+        //   1  │         a        │  2  │         b        │  3  │         c        │  4
+        // └   ┘└──────────────────┘└   ┘└──────────────────┘└   ┘└──────────────────┘└   ┘
```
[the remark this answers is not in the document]
kdheepak: I've removed the inline comments and changed the constraint creation using `Element` methods.

src/layout/layout.rs
```diff
@@ -436,333 +443,474 @@ impl Layout {
...
+
+        Ok(spacers)
+    }
+
+    /// Adds constraint for spacers in the given layout area using the specified solver.
+    ///
+    /// All spacers added in this function either
+    ///
+    /// 1. almost required to be the size as the spacing requirement provided by the user
+    /// 2. strongly incentivized to be the same size
```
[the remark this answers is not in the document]
kdheepak: This was from language I had previously to describe `REQUIRED - 1.0`.

src/layout/layout.rs
```diff
@@ -812,53 +961,164 @@ impl Layout {
...
+        // To minimize the time it takes to solve the same problem over and over again, we
+        // cache the `Layout` struct along with the results.
+        //
+        // `try_split` is the inner method in `split` that is called only when the LRU cache doesn't
+        // match the key. So inside `try_split`, we create a new instance of the solver.
+        //
+        // This is equivalent to storing the solver in `Layout` and calling `solver.reset()` here.
+        let mut solver = Solver::new();
+
+        let segments = Layout::build_segment_elements(&mut solver, area, layout)?;
```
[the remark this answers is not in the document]
kdheepak: I figured these are methods that will only be used in the `Layout` struct, so maybe using associated functions makes it easier to understand that?

src/layout/layout.rs
kdheepak: I've updated the code to use `Element` methods.

src/layout/layout.rs
kdheepak: I've added a docstring and tests for this function.

src/layout/layout.rs
kdheepak: Do you mean like this?

```rust
type Segments = Rc<[Rect]>;
type Spacers = Rc<[Rect]>;
type Cache = LruCache<(Rect, Layout), (Segments, Spacers)>;
```

src/layout/layout.rs
```diff
@@ -479,383 +539,379 @@ impl Layout {
...
+    // all variables are in ascending order
+    for (&left, &right) in variables.iter().tuple_windows() {
+        solver.add_constraint(left | LE(REQUIRED) | right)?;
+    }
+
+    // the first and last variables are at the start and end of the area
+    if let (Some(&first), Some(&last)) = (variables.first(), variables.last()) {
+        solver.add_constraint(first | EQ(REQUIRED) | area.start)?;
+        solver.add_constraint(last | EQ(REQUIRED) | area.end)?;
+    }
```
joshka, line 665: We could also move this to the main split method as something like
```rust
let area = Element::from((first.start, last.end))
```

src/layout/layout.rs
joshka: Interesting, but I actually meant:
```rust
struct { segments, spacers }
```
Not a big deal, as this is internal only. :) I just don't like having to look up what the order of some tuple is when the items have distinctly useful names.

src/layout/layout.rs
kdheepak: I did this:

```rust
let area_size = Element::from((*variables.first().unwrap(), *variables.last().unwrap()));
```

using `.unwrap()` since they always exist. `variables[0]` and `variables[variables.len()-1]` also works for me. Is there one you prefer?

src/layout/layout.rs
```diff
@@ -27,7 +29,7 @@ type Rects = Rc<[Rect]>;
 // └   ┘└──────────────────┘└   ┘└──────────────────┘└   ┘└──────────────────┘└   ┘
 //
 // Number of spacers will always be one more than number of segments.
-type Cache = LruCache<(Rect, Layout), (Rects, Rects)>;
+type Cache = LruCache<(Rect, Layout), (Segments, Spacers)>;
```
joshka, line 32: Does this need to be public?

joshka (approved):

joshka:
LGTM

src/layout/layout.rs
joshka: I'm fine with either.

src/layout/layout.rs
kdheepak: Let's leave it private for now till we figure out if there's a better API for returning these `Rects` to users.

joshka: k

joshka (approved):

kdheepak:
What's a good commit message for the changelog? Just first part of the PR description?

kdheepak:
Thanks for all the comments!

PR #789, opened 2024-01-11
feat(layout): add Rect -> Size conversion methods

joshka:
- add Size::new() constructor
- add Rect::as_size()
- impl From<Rect> for Size
- document and add tests for Size

### DIFF
--- a/src/layout/rect.rs
+++ b/src/layout/rect.rs
@@ -8,6 +8,7 @@ use crate::prelude::*;
 
 mod offset;
 
+use layout::Size;
 pub use offset::*;
 
 /// A simple rectangle used in the computation of the layout and to give widgets a hint about the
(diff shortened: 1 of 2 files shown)


Valentin271 (approved):
Looks good. Why do we need that? Maybe its worth adding an example somewhere if its relevant to users.

joshka:
In [Oatmeal](https://github.com/dustinblackman/oatmeal), I noticed some code where app was storing the most recent height and width of a rect and then later comparing those with each other (paraphrased a bit)

```rust
if area.width != state.width || area.height != state.height {
    state.width = area.width;
    state.height = area.height;
    do_some_update();
}
```

This simplifies that use case to:
```rust
if area.as_size() != state.size {
    state.size = area.as_size();
    do_some_update();
}
```

> Maybe it's worth adding an example somewhere if its relevant to users.

I think this is more of a utility type method. Something that can be used optimistically when needed rather than something that needs a demo or examples.

I'm also going to add a `Position { x, y }` struct as a followup PR for similar reasons.

Valentin271:
> I think this is more of a utility type method. Something that can be used optimistically when needed rather than something that needs a demo or examples.

Seeing the usecase I agree. All good to me then.

PR #790, opened 2024-01-11
feat(layout): add Position struct

joshka:
This stores the x and y coordinates (columns and rows)

- add conversions from Rect
- add conversion with Size to Rect
- add Rect::as_position

### DIFF
--- a/src/layout/position.rs
+++ b/src/layout/position.rs
@@ -1,0 +1,81 @@
+#![warn(missing_docs)]
+use crate::layout::Rect;
+
+/// Position in the terminal
+///
+/// The position is relative to the top left corner of the terminal window, with the top left corner
+/// being (0, 0). The x axis is horizontal increasing to the right, and the y axis is vertical
+/// increasing downwards.
+pub struct Position {
+    /// The x coordinate of the position
+    ///
+    /// The x coordinate is relative to the left edge of the terminal window, with the left edge
+    /// being 0.
+    pub x: u16,
+
+    /// The y coordinate of the position
+    ///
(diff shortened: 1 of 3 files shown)


src/layout/position.rs
```diff
@@ -0,0 +1,81 @@
+#![warn(missing_docs)]
+use crate::layout::Rect;
+
+/// Position in the terminal
+///
+/// The position is relative to the top left corner of the terminal window, with the top left corner
+/// being (0, 0). The x axis is horizontal increasing to the right, and the y axis is vertical
+/// increasing downwards.
+pub struct Position {
```
Valentin271, line 9: Could you add the derive macros here

Valentin271 (changes requested):

Valentin271:
Could you also add a test for `From<(Position, size)> for Rect`? (don't forget to fix formatting)

Valentin271:
Check out https://github.com/ratatui-org/ratatui/pull/789#issuecomment-1887535265.

Mainly this allows easy manipulation and especially storage of positions.

joshka:
One place I'd like this is instead of a (u16,u16) tuple in a state object

joshka:
realized I pushed, but didn't see that it failed due to the a missing import in a doc test. Pushed now

Valentin271:
It seems like the doc-tests are still failing due to missing import @joshka.

joshka:
> It seems like the doc-tests are still failing due to missing import @joshka.

Missed the commit. Fixed now.

Valentin271 (approved):

PR #795, opened 2024-01-12
fix(examples/scrollbar): title wasn't displayed because of background reset

Valentin271:
This has always been like that since scrollbar was introduced so the title was never displayed(?).


orhun (approved):

Valentin271:
At some point it was black (shown on the example gif), but I have a black terminal, and some other people too so ... its good to change anyway.

PR #804, opened 2024-01-13
feat: Add flex to layout ✨

kdheepak:
This PR adds a new way to space elements in a `Layout`.

Loosely based on [flexbox](https://css-tricks.com/snippets/css/a-guide-to-flexbox/), this PR adds a `Flex` enum with the following variants:

- Start
- Center
- End
- SpaceAround
- SpaceBetween

image

It also adds two more variants, to make this backward compatible and to make it replace `SegmentSize`:

- StretchLast (default in the `Flex` enum, also behavior matches old default `SegmentSize::LastTakesRemainder`)
- Stretch (behavior matches `SegmentSize::EvenDistribution`)

The `Start` variant from above matches `SegmentSize::None`.

This allows `Flex` to be a complete replacement for `SegmentSize`, hence this PR also deprecates the `segment_size` constructor on `Layout`. `SegmentSize` is still used in `Table` but under the hood `segment_size` maps to `Flex` with all tests passing unchanged.

I also put together a simple example for `Flex` layouts so that I could test it visually, shared below:

https://github.com/ratatui-org/ratatui/assets/1813121/c8716c59-493f-4631-add5-feecf4bd4e06


src/layout/layout.rs
```diff
@@ -426,34 +437,83 @@ impl Layout {
...
-            solver.add_constraint(first.start | EQ(REQUIRED) | area_start)?;
-        }
-        if layout.segment_size != SegmentSize::None {
-            // ensure the last element touches the right/bottom edge of the area
-            if let Some(last) = elements.last() {
-                solver.add_constraint(last.end | EQ(REQUIRED) | area_end)?;
+
+        match layout.flex {
+            Flex::StretchLast => {
+                if let Some(first) = elements.first() {
```
joshka, line 486: I wonder if there's a refactoring of this method that makes there always be an element (i.e. have one more element than constraints). This would make first and last always have a value, and remove all the if statements.

src/layout/layout.rs
```diff
@@ -602,13 +656,30 @@ impl Layout {
...
+        (area_start, area_end): (f64, f64),
+    ) -> Result<Self, AddConstraintError> {
+        let e = Element {
+            start: Variable::new(),
+            end: Variable::new(),
+        };
+        solver.add_constraints(&[
+            e.start | GE(REQUIRED) | area_start,
+            e.end | LE(REQUIRED) | area_end,
+            e.start | LE(REQUIRED) | e.end,
```
joshka, line 731: Is this constraint necessary?

src/layout/layout.rs
```diff
@@ -1719,5 +1789,120 @@ mod tests {
             ]));
             assert_eq!([a.width, b.width, c.width, d.width], [0, 0, 0, 100]);
         }
+
+        #[test]
+        fn flex() {
```
joshka, line 1847: These test would probably be a good rstest test (there's a couple already in the repo for reference, but there's other methods that might suit this test.

src/layout/layout.rs
```diff
@@ -1719,5 +1789,120 @@ mod tests {
...
+            // length should stretch to the very end
+            let [a] = Rect::new(0, 0, 100, 1).split(&Layout::horizontal([Length(50)]));
+            assert_eq!([a.x, a.width], [0, 100]);
```
joshka, line 1850: This test forces the reader to have to look up what the default for Flex is rather than just seeing that directly in front of them. There's two ways to fix this:
1. Create one test that checks that when you create a Layout, the default for flex is a particular value. Then always specify the flex for these tests, knowing that the default is properly handled already. This will make the tests less implicit (and they won't fail when we change the default).
2. Duplicate this test with the same results and with the flex specified

(I prefer option 1)

joshka (commented):
LGTM

src/layout/layout.rs
```diff
@@ -443,19 +443,56 @@ impl Layout {
...
-            solver.add_constraint(pair[0].end | EQ(REQUIRED) | pair[1].start)?;
-        }
-
         match layout.flex {
+            Flex::SpaceBetween => {
+                let mut spacers: Vec<Element> = elements
+                    .iter()
+                    .map(|_| Element::constrain(&mut solver, (area_start, area_end)))
+                    .try_collect()?;
+                spacers.pop();
```
joshka, line 452: looks like you're just duplicating a certain number here. can this be something like :
```rust
let spaces = vec![Element::constrain(...) ; elements.len() - 1];

Comment (kdheepak) [reply src/layout/layout.rs:452]: This doesn't work unfortunately, because this expands to something like the following:

```rust
let mut spaces = vec![];
let only_one_element_instantiated = Element::constrain(...);
for i in 0..elements.len() - 1 {
  spaces.push(only_one_element_instantiated)
}
```

We need this instead:

```
let mut spaces = vec![];
for i in 0..elements.len() - 1 {
  spaces.push(Element::constrain(...))
}
```

I can probably use `repeat_with` and `take`.

joshka: https://doc.rust-lang.org/std/vec/struct.Vec.html#method.fill_with perhaps?

src/layout/layout.rs
kdheepak: I think it is required because it is a base assumption about how we measure width. 

```
    fn size(&self) -> Expression {
        self.end - self.start
    }
```

without this constraint, the solver may find a solution where width is negative. This is probably more true now with the `Proportional` constraints and the `Flex` layouts.

joshka: But if area_start < area_end, and e.start >= area_start, and e.end <= area.end, I'm not sure how e.start > e.end unless math is broken

kdheepak: If the solver finds this `Rect` for whatever reason:

```
area_start == 0.0
e.end == 5.0
e.start == 8.0
area_end == 10.0
```

- [x] `area_start < area_end`, 0 < 10
- [x] `e.start >= area_start`, 8 >= 0
- [x] `e.end` <= area_end`, 5 <= 10

We need this constraint to ensure that this doesn't happen:

- [ ] `e.start <= e.end`, 8 >= 5

kdheepak: `area_start` and `area_end` are the total area start and end. Maybe you are thinking it is the area of that particular rect?

src/layout/layout.rs
```diff
@@ -142,6 +144,7 @@ impl Layout {
             margin: Margin::new(0, 0),
             constraints: constraints.into_iter().map(Into::into).collect(),
             segment_size: SegmentSize::LastTakesRemainder,
```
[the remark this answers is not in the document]
kdheepak: Done.

src/layout/layout.rs
```diff
@@ -91,7 +92,7 @@ pub struct Layout {
     direction: Direction,
     constraints: Vec<Constraint>,
     margin: Margin,
-    /// option for segment size preferences
+    flex: Flex,
     segment_size: SegmentSize,
```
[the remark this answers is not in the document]
kdheepak: Done.

src/layout/layout.rs
joshka: oh yeah, my bad.

src/layout/layout.rs
```diff
@@ -360,9 +369,13 @@ impl Layout {
         issue = "https://github.com/ratatui-org/ratatui/issues/536"
     )]
     #[must_use = "method moves the value of self and returns the modified value"]
-    pub const fn segment_size(mut self, segment_size: SegmentSize) -> Layout {
-        self.segment_size = segment_size;
-        self
+    pub const fn segment_size(self, segment_size: SegmentSize) -> Layout {
```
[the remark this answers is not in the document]
kdheepak: Should I deprecate this?

src/layout/flex.rs
```diff
@@ -0,0 +1,16 @@
...
+///
+/// This controls how the space is distributed when the constraints are satisfied.
+#[derive(Copy, Debug, Default, Display, EnumString, Clone, Eq, PartialEq, Hash)]
+pub enum Flex {
+    #[default]
+    StretchLast, // Always fill available space within the container
+    Stretch, // Always fill available space within the container
+    Start,   // Align items to the start of the container
+    End,     // Align items to the end of the container
+    Center,  // Center items within the container
```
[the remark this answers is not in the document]
kdheepak: Done.

src/layout/layout.rs
```diff
@@ -426,34 +437,83 @@ impl Layout {
...
+                // prefer equal chunks if other constraints are all satisfied
+                for (left, right) in elements.iter().tuple_combinations() {
+                    solver.add_constraint(left.size() | EQ(WEAK) | right.size())?;
+                }
+            }
+            Flex::Center => {
+                let flex_start_element = Element::constrain(&mut solver, (area_start, area_end))?;
+                let flex_end_element = Element::constrain(&mut solver, (area_start, area_end))?;
+                solver.add_constraint(
+                    flex_start_element.size() | EQ(STRONG) | flex_end_element.size(),
```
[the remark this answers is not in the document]
kdheepak: Yes it does change, because we want this to be exactly equal to `Min` and `Max` when they are at their limit for good solutions. 

In this PR I didn't change any weights of any constraints but in another PR I will change weights which affect how constraints behave.

src/layout/layout.rs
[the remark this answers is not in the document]
joshka: yep

src/layout/flex.rs
```diff
@@ -1,19 +1,183 @@
...
+///   last element.
+/// - `Stretch`: Always fills the available space within the container.
+/// - `Start`: Aligns items to the start of the container.
+/// - `End`: Aligns items to the end of the container.
+/// - `Center`: Centers items within the container.
+/// - `SpaceBetween`: Adds excess space between each element.
+/// - `SpaceAround`: Adds excess space around each element.
 #[derive(Copy, Debug, Default, Display, EnumString, Clone, Eq, PartialEq, Hash)]
 pub enum Flex {
+    /// Fills the available space within the container, putting excess space into the last element.
```
joshka, line 17: Add a note about this being the default for matching existing behavior of ratatui / tui apps?

src/layout/flex.rs
```diff
@@ -1,19 +1,183 @@
...
+    ///
+    /// ```plain
+    /// 
+    /// Length(20), Length(10)
+    ///
+    /// <------------------------------------80 px------------------------------------->
+    ///
+    /// ┌──────────────────┐┌──────────────────────────────────────────────────────────┐
+    /// │       20 px      ││                           60 px                          │
+    /// └──────────────────┘└──────────────────────────────────────────────────────────┘
```
joshka, line 55: This is a weird result. Could do with a comment about why this is this way.

Is this comment flipped with StretchLast above?

Perhaps leave a line between the variants to communicate more obviously that the doc comment belongs to the variant below not the one above.

joshka (changes requested):

src/layout/layout.rs
```diff
@@ -357,10 +357,7 @@ impl Layout {
     /// equal chunks or to not distribute extra space at all (which is the default used for laying
     /// out the columns for [`Table`] widgets).
     ///
-    /// Note: If you're using this feature please help us come up with a good name. See [Issue
-    /// #536](https://github.com/ratatui-org/ratatui/issues/536) for more information.
-    ///
-    /// [`Table`]: crate::widgets::Table
+    /// This function exists for backwards compatibility reasons. Use [`Layout::flex`] instead.
```
joshka, line 360: Note the equivalent values here or in the doc string

src/layout/flex.rs
kdheepak: Oops my bad. It should be the other way round.

src/layout/layout.rs
kdheepak: repeat_with and take seem idiomatic so I'll leave it at that.

src/layout/flex.rs
kdheepak: Fixed in latest commit.

src/layout/layout.rs
kdheepak: Changed it to rstest.

src/layout/layout.rs
kdheepak: Changed it to rstest. I'll add a separate test for the default.

src/layout/layout.rs
```diff
@@ -1842,237 +1843,88 @@ mod tests {
...
+        #[case::min_right_justified([Min(50)], Flex::End, [50, 50])]
+        #[case::min_center_justified([Min(50)], Flex::Center, [25, 50])]
+        #[case::max_stretches_to_end([Max(50)], Flex::StretchLast, [0, 100])]
+        #[case::max_left_justified([Max(50)], Flex::Start, [0, 50])]
+        #[case::max_right_justified([Max(50)], Flex::End, [50, 50])]
+        #[case::max_center_justified([Max(50)], Flex::Center, [25, 50])]
+        fn flex_one_constraint(
+            #[case] constraints: [Constraint; 1],
+            #[case] flex: Flex,
+            #[case] expected_widths: [u16; 2],
```
joshka, line 1876: Perhaps make this either a tuple or multiple values - (start, width)

src/layout/layout.rs
```diff
@@ -1842,237 +1843,88 @@ mod tests {
...
+        #[case::min_stretches_to_end([Min(50)], Flex::StretchLast, [0, 100])]
+        #[case::min_left_justified([Min(50)], Flex::Start, [0, 50])]
+        #[case::min_right_justified([Min(50)], Flex::End, [50, 50])]
+        #[case::min_center_justified([Min(50)], Flex::Center, [25, 50])]
+        #[case::max_stretches_to_end([Max(50)], Flex::StretchLast, [0, 100])]
+        #[case::max_left_justified([Max(50)], Flex::Start, [0, 50])]
+        #[case::max_right_justified([Max(50)], Flex::End, [50, 50])]
+        #[case::max_center_justified([Max(50)], Flex::Center, [25, 50])]
+        fn flex_one_constraint(
+            #[case] constraints: [Constraint; 1],
```
joshka, line 1874: This could just be a single Constraint to the case list easier to read (less nested brackets)

src/layout/layout.rs
```diff
@@ -1842,237 +1843,88 @@ mod tests {
...
+        #[case::min_stretches_to_end([Min(25), Min(25)], Flex::Stretch, [[0, 50], [50, 50]])]
+        #[case::min_justifies_to_start([Min(25), Min(25)], Flex::Start, [[0, 25], [25, 25]])]
+        #[case::min_justifies_to_center([Min(25), Min(25)], Flex::Center, [[25, 25], [50, 25]])]
+        #[case::min_justifies_to_end([Min(25), Min(25)], Flex::End, [[50, 25], [75, 25]])]
+        #[case::length_spaced_between([Length(25), Length(25)], Flex::SpaceBetween, [[0, 25], [75, 25]])]
+        #[case::length_spaced_around([Length(25), Length(25)], Flex::SpaceAround, [[17, 25], [58, 25]])]
+        fn flex_two_constraints(
+            #[case] constraints: [Constraint; 2],
+            #[case] flex: Flex,
+            #[case] expected_widths: [[u16; 2]; 2],
```
joshka, line 1906: Definitely make this an array of tuples instead of array of arrays - `[(17, 25), (58, 25)]` is easier to read than `[[17, 25], [58, 25]]`

src/layout/layout.rs
kdheepak: Done.

src/layout/layout.rs
kdheepak: Changed everything to tuples.

src/layout/layout.rs
kdheepak: Good idea. Done.

src/layout/layout.rs
kdheepak: I added a docstring with the mapping.

src/layout/flex.rs
kdheepak: I added a note documenting the default.

src/layout/layout.rs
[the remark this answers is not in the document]
kdheepak: Done.

src/layout/layout.rs
kdheepak: This can and probably should be refactored, but I'd like to do it when I add more tests to document the implementation.

kdheepak:
Follow on work will: 

1. Add more tests to specify behavior of flex layouts in tests
2. Revisit illustrations docstrings for Constraint and Flex
3. Rework priorities for potentially different behavior
4. Refactor long split function
5. Format test cases so they are more readable

joshka (approved):
LGTM

joshka (approved):

Valentin271 (approved):
Looks good to me. I think `SegmentSize` itself could be deprecated.

I'd like if we could scroll on the flex example because I can't see everything (this can be part of another PR and add scroll to constraints as well)

image

PR #811, opened 2024-01-14
refactor(example): add scroll to flex example

Valentin271:
On a standard size terminal, flex example is cut off which is a shame because of how great it is.

image

I added scroll to the example.

`constraints` will also need that but I first want to validate this kind of janky implementation.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -220,7 +220,7 @@ doc-scrape-examples = false
 [[example]]
 name = "flex"
 required-features = ["crossterm"]
-doc-scrape-examples = false
+doc-scrape-examples = true
 
 [[example]]
 name = "list"
--- a/examples/flex.rs
+++ b/examples/flex.rs
@@ -5,7 +5,6 @@ use crossterm::{
     execute,
     terminal::{disable_raw_mode, enable_raw_mode, EnterAlternateScreen, LeaveAlternateScreen},
 };
-use itertools::Itertools;
 use ratatui::{
     layout::{Constraint::*, Flex},
     prelude::*,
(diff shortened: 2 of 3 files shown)


kdheepak:
Neat hack!

Valentin271:
> Neat hack!

It was easier than refactoring the example to draw offset and all ...

I think something similar could be added to `Buffer` itself.

kdheepak:
I moved the name of the constraints into the box itself (thanks for @joshka for the idea) and changed the scrolling to make it never go past the last example.

default:

image

max scroll:

image

I also changed some of the constraints such that it shows how flex behaves even with just 80 px width (and shows the name of the constraints well).

Valentin271:
Looks great!

I tweak it a bit more:

- replaced example height value with a constant to avoid desync and be self documenting
- replaced `clamp(0, ...)` with `min(...)`
- Increased `Length(10)` to 12 so it's not truncated as below

image

kdheepak:
I was making almost the exact same changes lol 

image

Pushing mine now

kdheepak:
Previously colors were just incremented in order to match a title. With the title moved into the blocks, we don't need that.

Now, I made the colors dependent on the type of constraint:

image

I think this makes it easier to visually parse what is going on.

kdheepak (approved):

kdheepak (approved):

kdheepak:
I'm going to add a couple more examples here, now that scroll is implemented, we can show more behavior.

kdheepak:
This is good to go imo.

kdheepak:
Here's a demo of it as a gif. I'm adding the tape file to this PR as well.

flex

kdheepak:
here's the latest gif 

flex

PR #813, opened 2024-01-14
fix: make SpaceBetween with one element Stretch 🐛

kdheepak:
When there's just one element, `SpaceBetween` should do the same thing as `Stretch`.

### DIFF
--- a/src/layout/layout.rs
+++ b/src/layout/layout.rs
@@ -472,7 +472,6 @@ impl Layout {
         // If we changed our default layout to use `Flex::Start`, there is a case to be made for
         // this to do `Flex::Start` as well.
         //
-        // TODO: add test for this
         let flex = if layout.constraints.len() == 1 && layout.flex == Flex::SpaceBetween {
             Flex::Stretch
         } else {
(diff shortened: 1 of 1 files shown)


Valentin271:
I don't think converting `SpaceBetween` to `Stretch` is a good idea.

I agree the current behavior (sometimes on the left sometimes on the right) is a bug, but instinctively, setting `SpaceBetween` with one element I expect it to respect its constraint and stay either left or right. (Looking at [this page](https://developer.mozilla.org/en-US/docs/Web/CSS/justify-content), it's probably left).

kdheepak:
In CSS the default when you use `flex` is justify to the start. So when there's just one element that's what they do. 

For us, our default is `Stretch`. 

Additionally, there's two reasons I think `SpaceBetween` should be `Stretch`. 

1. Intuitively it makes sense because we are saying for every element add a spacer between adjacent element but make the start of the first element at the start of the area and make the end of the last element at the end of the area.
2. This above is exactly is what constraints are added in the `SpaceBetween` match but we are using `tuple_combinations` and windows so when there's just one element and no spacers, it doesn't do anything. If I make that code work for one element, it'll do the same thing as `Stretch`.

Valentin271:
> In CSS the default when you use `flex` is justify to the start. So when there's just one element that's what they do.
> 
> For us, our default is `Stretch`.

Alright this convinced me. I see your reasoning.

> If we changed our default layout to do `Start`, I can see the rationale for making this do `Start` as well.

Can we then make it `Flex::default()` so it expresses that intent? It'll be `StretchLast` but this should produce the same output right?

kdheepak:
> Can we then make it Flex::default() so it expresses that intent? It'll be StretchLast but this should produce the same output right?

I like this idea and almost made the change to do it that way. But I'd like to leave it as `Flex::Stretch` for now. I want to play around with all the constraint weights in this PR https://github.com/ratatui-org/ratatui/pull/788 because I think that might affect what we choose here as well. If I don't see any issues after messing around with the weights, then we can do that.

Valentin271 (approved):
Works for me then!

PR #814, opened 2024-01-14
chore(layout): comment tests that may fail on occasion

kdheepak:
These fails seem to fail on occasion, locally and on CI.

This issue will be revisited in the PR on constraint weights: https://github.com/ratatui-org/ratatui/pull/788


Valentin271 (approved):
Merging once the pipeline is completed

PR #817, opened 2024-01-14
refactor(example): improve constraints and flex examples

Valentin271:
This PR is a follow up to https://github.com/ratatui-org/ratatui/pull/811.

It improves the UI of the layouts by

- thoughtful accessible color that represent priority in constraints resolving
- using QUADRANT_OUTSIDE symbol set for block rendering
- adding a scrollbar
- panic handling
- refactoring for readability

to name a few. Here are some example gifs of the outcome:

constraints

flex

### DIFF
--- a/examples/flex.rs
+++ b/examples/flex.rs
@@ -45,17 +45,15 @@ const EXAMPLE_DATA: &[(&str, &[Constraint])] = &[
     ("", &[Min(20), Max(20), Length(20), Fixed(20)]),
     ("`Proportional(u16)` always fills up space in every `Flex` layout", &[Proportional(0), Proportional(0)]),
     (
-        "`Proportional()` can be to scale with respect to other `Proportional()`",
-        &[Proportional(1), Proportional(2), Length(20), Fixed(20)],
+        "`Proportional(1)` can be to scale with respect to other `Proportional(2)`",
+        &[Proportional(1), Proportional(2)],
     ),
     (
-        "",
+        "`Proportional(0)` collapses if there are other non-zero `Proportional(_)`\nconstraints. e.g. `[Proportional(0), Proportional(0), Proportional(1)]`:",
         &[
-            Min(10),
-            Proportional(3),
-            Proportional(2),
-            Length(15),
-            Fixed(15),
+            Proportional(0),
+            Proportional(0),
+            Proportional(1),
         ],
     ),
 ];
--- a/examples/flex.rs
+++ b/examples/flex.rs
@@ -11,7 +11,9 @@ use ratatui::{
     widgets::{block::Title, *},
 };
 
-const EXAMPLE_HEIGHT: u16 = 5;
+const SPACER_HEIGHT: u16 = 1;
+const ILLUSTRATION_HEIGHT: u16 = 3;
+const EXAMPLE_HEIGHT: u16 = ILLUSTRATION_HEIGHT + SPACER_HEIGHT;
 const N_EXAMPLES_PER_TAB: u16 = 11;
 
 fn main() -> Result<(), Box<dyn Error>> {
@@ -91,6 +93,34 @@ impl App {
             .min(self.max_scroll_offset)
     }
 
+    fn render_tabs_and_axis(&self, area: Rect, buf: &mut Buffer) {
+        let [tabs, axis] = area.split(&Layout::vertical([
+            Constraint::Fixed(3),
+            Constraint::Fixed(3),
+        ]));
+        self.render_tabs(tabs, buf);
+        self.render_axis(axis, buf);
+    }
+
+    fn render_axis(&self, area: Rect, buf: &mut Buffer) {
+        let width = area.width as usize;
+        // a bar like `<----- 80 px ----->`
+        let width_label = format!("{} px", width);
+        let width_bar = format!(
+            "<{width_label:-^width$}>",
+            width = width - width_label.len() / 2
+        );
+        Paragraph::new(width_bar.dark_gray())
+            .alignment(Alignment::Center)
+            .block(Block::default().padding(Padding {
+                left: 0,
+                right: 0,
+                top: 1,
+                bottom: 0,
+            }))
+            .render(area, buf);
+    }
+
     fn render_tabs(&self, area: Rect, buf: &mut Buffer) {
         Tabs::new(
             [
@@ -119,7 +149,8 @@ impl App {
 
 impl Widget for App {
     fn render(self, area: Rect, buf: &mut Buffer) {
-        let [tabs_area, demo_area] = area.split(&Layout::vertical([Fixed(3), Proportional(0)]));
+        let [tabs_and_axis_area, demo_area] =
+            area.split(&Layout::vertical([Fixed(6), Proportional(0)]));
 
         // render demo content into a separate buffer so all examples fit
         let mut demo_buf = Buffer::empty(Rect::new(
@@ -132,12 +163,12 @@ impl Widget for App {
         self.selected_example.render(demo_buf.area, &mut demo_buf);
 
         // render tabs into a separate buffer
-        let mut tabs_buf = Buffer::empty(tabs_area);
-        self.render_tabs(tabs_area, &mut tabs_buf);
+        let mut tabs_and_axis_buf = Buffer::empty(tabs_and_axis_area);
+        self.render_tabs_and_axis(tabs_and_axis_area, &mut tabs_and_axis_buf);
 
         // Assemble both buffers
         // NOTE: You shouldn't do this in a production app
-        buf.content = tabs_buf.content;
+        buf.content = tabs_and_axis_buf.content;
         buf.content.append(
             &mut demo_buf
                 .content
@@ -273,12 +304,9 @@ impl Example {
 
 impl Widget for Example {
     fn render(self, area: Rect, buf: &mut Buffer) {
-        let [legend, area] = area.split(&Layout::vertical([Ratio(1, 3); 2]));
         let blocks = Layout::horizontal(&self.constraints)
             .flex(self.flex)
             .split(area);
-
-        self.legend(legend.width as usize).render(legend, buf);
 
         for (block, constraint) in blocks.iter().zip(&self.constraints) {
             let text = format!("{} px", block.width);
@@ -291,29 +319,16 @@ impl Widget for Example {
                 Constraint::Fixed(_) => Color::Indexed(6),
                 Constraint::Proportional(_) => Color::Indexed(7),
             };
-            self.illustration(*constraint, text, fg).render(*block, buf);
+            let [block, _] = block.split(&Layout::vertical([
+                Fixed(ILLUSTRATION_HEIGHT),
+                Fixed(SPACER_HEIGHT),
+            ]));
+            self.illustration(*constraint, text, fg).render(block, buf);
         }
     }
 }
 
 impl Example {
-    fn legend(&self, width: usize) -> Paragraph {
-        // a bar like `<----- 80 px ----->`
-        let width_label = format!("{} px", width);
-        let width_bar = format!(
-            "<{width_label:-^width$}>",
-            width = width - width_label.len() / 2
-        );
-        Paragraph::new(width_bar.dark_gray())
-            .alignment(Alignment::Center)
-            .block(Block::default().padding(Padding {
-                left: 0,
-                right: 0,
-                top: 1,
-                bottom: 0,
-            }))
-    }
-
     fn illustration(&self, constraint: Constraint, text: String, fg: Color) -> Paragraph {
         Paragraph::new(format!("{:?}", constraint))
             .alignment(Alignment::Center)
--- a/examples/flex.rs
+++ b/examples/flex.rs
@@ -125,7 +125,7 @@ impl App {
         use KeyCode::*;
         match event::read()? {
             Event::Key(key) if key.kind == KeyEventKind::Press => match key.code {
-                Char('q') => self.quit(),
+                Char('q') | Esc => self.quit(),
                 Char('l') | Right => self.next(),
                 Char('h') | Left => self.previous(),
                 Char('j') | Down => self.down(),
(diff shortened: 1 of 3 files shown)


kdheepak:
Looks great! Thanks for making the changes!

kdheepak (approved):

joshka:
Could it be worth putting the width bar at the top / bottom and compressing the examples into half as many lines?

kdheepak:
I tried it and I think it does look better. I ended up making the width bar sticky and only at the top. Scroll doesn't affect it so we don't need it at the bottom.

image

kdheepak:
Here's what it looks like with the flex examples

image

kdheepak:
Here's what it looks like with a one line spacer:

image

kdheepak:
Once this is merged I'll make a documentation PR to update the code snippets with this style.

joshka:
what about putting the pixels inside and the constraints on the border?

kdheepak:
> what about putting the pixels inside and the constraints on the border?

Interesting idea.

image

I didn't like it as much because the constraint types don't stand out. But changing the colors makes it better I think:

image

kdheepak:
I like this better:

image

joshka:
What about the second option, but with centered headings?

examples/flex.rs
```diff
@@ -291,29 +319,16 @@ impl Widget for Example {
                 Constraint::Fixed(_) => Color::Indexed(6),
                 Constraint::Proportional(_) => Color::Indexed(7),
             };
-            self.illustration(*constraint, text, fg).render(*block, buf);
+            let [block, _] = block.split(&Layout::vertical([
+                Fixed(ILLUSTRATION_HEIGHT),
+                Fixed(SPACER_HEIGHT),
+            ]));
+            self.illustration(*constraint, text, fg).render(block, buf);
```
joshka, line 326: Moving the text and fg calculation into illustration (and just passing constraint and area.width) makes this a bit neater.

joshka:
Using background colors instead of borders looks pretty nice:

  Made with VHS

Though constraints that match and are next to each other need an indicator

joshka:
image

kdheepak:
I considered it but in addition to what you said I also want something that is copy paste friendly, e.g.:

```plain
┌Flex Layouts ─ Use h l or ◄ ► to change tab and j k or ▲ ▼  to scroll─────────┐
│ Stretch │ StretchLast │ Start │ Center │ End │ SpaceAround │ SpaceBetween    │
└──────────────────────────────────────────────────────────────────────────────┘

<------------------------------------80 px------------------------------------->

┌─────────Length(20)─────────┐┌────Fixed(20)─────┐┌───────Percentage(20)───────┐
│            30 px           ││       20 px      ││            30 px           │
└────────────────────────────┘└──────────────────┘└────────────────────────────┘

┌────Fixed(20)─────┐┌───────Percentage(20)───────┐┌─────────Length(20)─────────┐
│       20 px      ││            30 px           ││            30 px           │
└──────────────────┘└────────────────────────────┘└────────────────────────────┘
```

joshka:
https://github.com/joshka/ratatui/tree/refactor/scroll-constraints-example

joshka:
Copied...
```
▐              Fixed(40)               ▌▐              Proportional(0)              ▌
▐                 40 px                ▌▐                   45 px                   ▌

▐    Fixed(20)     ▌▐    Fixed(20)     ▌▐              Proportional(0)              ▌
▐       20 px      ▌▐       20 px      ▌▐                   45 px                   ▌
```

joshka:
image

copied:

```plain
<---------------------------------------85 px--------------------------------------->

┌──────────────────────────────────────────────────────────────┐┌───────────────────┐
│                             64 px                            ││       21 px       │
│                        Percentage(75)                        ││  Proportional(0)  │
└──────────────────────────────────────────────────────────────┘└───────────────────┘
┌───────────────────┐┌──────────────────────────────────────────────────────────────┐
│       21 px       ││                             64 px                            │
│  Percentage(25)   ││                        Proportional(0)                       │
└───────────────────┘└──────────────────────────────────────────────────────────────┘
┌─────────────────────────────────────────┐┌────────────────────────────────────────┐
│                  43 px                  ││                  42 px                 │
│             Percentage(50)              ││                 Min(20)                │
└─────────────────────────────────────────┘└────────────────────────────────────────┘
┌───────────────────────────────────────────────────────────────────────────────────┐
│                                       85 px                                       │
│                                   Percentage(0)                                   │
└───────────────────────────────────────────────────────────────────────────────────┘
┌───────────────────────────────────────────────────────────────────────────────────┐
│                                       85 px                                       │
│                                  Proportional(0)                                  │
└───────────────────────────────────────────────────────────────────────────────────┘

```

kdheepak:
Do you think users will find this easier to read in the docs:

```
▐              Fixed(40)               ▌▐              Proportional(0)              ▌
▐                 40 px                ▌▐                   45 px                   ▌

▐    Fixed(20)     ▌▐    Fixed(20)     ▌▐              Proportional(0)              ▌
▐       20 px      ▌▐       20 px      ▌▐                   45 px                   ▌
```

or this:

```
┌─────────Length(20)─────────┐┌────Fixed(20)─────┐┌───────Percentage(20)───────┐
│            30 px           ││       20 px      ││            30 px           │
└────────────────────────────┘└──────────────────┘└────────────────────────────┘

┌────Fixed(20)─────┐┌───────Percentage(20)───────┐┌─────────Length(20)─────────┐
│       20 px      ││            30 px           ││            30 px           │
└──────────────────┘└────────────────────────────┘└────────────────────────────┘
```

I have a strong preference for the borders in the docs. 

I can make the example default to using your background colors method with a hidden keyboard toggle to switch into bordered so I can copy paste the examples. What do you think?

joshka:
image

copied:

```
<---------------------------------------85 px--------------------------------------->

┌──────────────────────────────────────────────────────────────┐┌───────────────────┐
│                             64 px                            ││       21 px       │
│                        Percentage(75)                        ││  Proportional(0)  │
└──────────────────────────────────────────────────────────────┘└───────────────────┘

┌───────────────────┐┌──────────────────────────────────────────────────────────────┐
│       21 px       ││                             64 px                            │
│  Percentage(25)   ││                        Proportional(0)                       │
└───────────────────┘└──────────────────────────────────────────────────────────────┘

┌─────────────────────────────────────────┐┌────────────────────────────────────────┐
│                  43 px                  ││                  42 px                 │
│             Percentage(50)              ││                 Min(20)                │
└─────────────────────────────────────────┘└────────────────────────────────────────┘

┌───────────────────────────────────────────────────────────────────────────────────┐
│                                       85 px                                       │
│                                   Percentage(0)                                   │
└───────────────────────────────────────────────────────────────────────────────────┘

┌───────────────────────────────────────────────────────────────────────────────────┐
│                                       85 px                                       │
│                                  Proportional(0)                                  │
└───────────────────────────────────────────────────────────────────────────────────┘
```

kdheepak:
Your most recent example looks good to me! I think you can even draw the border but color the foreground for the border with the same color as the terminal background, and we'll get the best of both worlds.

joshka:
bold borders kinda works for the non-light colors:
image

joshka:
> Your most recent example looks good to me! I think you can even draw the border but color the foreground for the border with the same color as the terminal background, and we'll get the best of both worlds.

I've always struggled to find a way to make the foreground color be the default background color - I think there might be a missing piece to make this work in either our code or in crossterm. But I'd agree, hidden borders could work.

joshka:
image
```rust
            .border_style(Style::reset().fg(bg).reversed())
```
Is about as best I can do.. (all in the https://github.com/joshka/ratatui/tree/refactor/scroll-constraints-example if you want it)

joshka:
using the QUADRANT_OUTSIDE border set:

image

```plain
▛▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▜▛▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▜
▌                 40 px                ▐▌                   45 px                   ▐
▌               Fixed(40)              ▐▌              Proportional(0)              ▐
▙▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▟▙▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▟

▛▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▜▛▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▜▛▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▜
▌       20 px      ▐▌       20 px      ▐▌                   45 px                   ▐
▌     Fixed(20)    ▐▌     Fixed(20)    ▐▌              Proportional(0)              ▐
▙▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▟▙▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▟▙▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▟
```

kdheepak:
I'm playing around with background colors locally and I don't find that it adds that much value to these examples. The jumping of large blobs of colors when moving between flex types hurts my eyes. As a static image I think they are fine.

I also think getting the colors right is hard, and I think it takes away from being able to easily read the constraint type, which is the essence of this demo.

I also have general concern with the background colors and having text on them for readability. Getting good contrast is not straightforward, and is also highly dependent on the users terminal palette. I cycled through my themes and I found a number of themes (although it was a minority) that has readability issues.

image

image

image

image

image

Can we leave these examples with just bordered the way they are? It's also easier for me to work with while I'm adding features to layouts. 

I'll work on a `constraint-visualizer-ui` that can use the background color method to distinguish boxes.

kdheepak:
I love the `QUADRANT_OUTSIDE` border set effect for screenshots! They look great.

joshka:
I changed the colors in my version away from your indexed defaults as they looked kinda bad in the aardvark blue theme.
Are those examples from my version, or yours? See the tweaked background colors to account for contrast.

joshka:
But black works better for aardvark too on this.
image

kdheepak:
With your colors I think it is better, I still found a few where the readability was off:

image

ignore the jagged edges, I was in the middle of changing something

joshka:
https://github.com/joshka/ratatui/commit/8d2d2dc446a7c8a18e65b38982b36e9744cebe5e

Aardvark Blue:

image

image

Frappe:

image

joshka:
We could always use the new palette modules if this was a huge concern

joshka:
https://github.com/joshka/ratatui/commit/b870e0f580182c0bcdc0b4a8253fed932697d18f
image

joshka:
Moving Proportional before Perc / Ration might also make it look nice.
I picked colors that make proportional / fixed darker than (perc and ratio), (length)

joshka:
After some discord chats, we came to:
image

kdheepak:
image

joshka:
Lots of little things changed in those refactoring changes:
- color eyre
- moved examples to a const
- made height calcs a method
- moved app state / running stuff around, so the main call is just `App::default().run(terminal)?;`
- don't render the spacers when the description is empty
- tailwind colors for tabs, highlighted is reverse
- tabs and axis methods just return `impl Widget` and let the render method call it.
- render_demo creates a buffer the same size of the area and renders into it and then splices that into the main buffer
- next/prev tab functions use strum::from_repr 
- renamed exampleSelection to SelectedTab
- SelectedTab to_tab_title uses strum::Display to get to_string()
- dynamically calculate example heights
- simplify the Example widget::render and illustration methods
- move color for constraint to a method

joshka:
image

image

joshka:
> I think the only thing left is to reorder the code in `constraints` right?

@kdheepak was going to add comments to the examples. I think it could do with a ScrollBar as well.

I'll reorder the constraints example to match flex, and add the ScrollBar shortly.

> Damn, that new `flex` example is so good!

A joint effort of bouncing ideas back and forth with @kdheepak last night :)

Valentin271:
Alright. Agree for the scrollbar if you can make it work without too much hassle. I couldn't really because of the hacky way I implemented the scroll.

joshka:
constraints

flex

examples/flex.rs
```diff
@@ -1,326 +1,452 @@
...
+
+        if !self.description.is_empty() {
+            Paragraph::new(
+                self.description
+                    .split('\n')
+                    .map(|s| {
+                        format!("// {}", s)
+                            .italic()
+                            .fg(Color::from_str("#908caa").unwrap())
+                    })
```
joshka, line 371: What about:
```suggestion
                    .map(|s| format!("// {}", s).italic().fg(tailwind::SLATE.c400))
```
which gives:
image

instead of:

image

joshka (approved):

examples/flex.rs
joshka: It could be worth also manually setting the background color of the app if we're drawing on that directly.
I just realized this probably looks fine on a black bg (or something with less color in it than my blue terminal)

joshka:
Can we summarize this change in the first comment and then squash it?

examples/flex.rs
```diff
@@ -1,326 +1,446 @@
-use std::{error::Error, io};
+use std::{
+    io::{self, stdout},
+    str::FromStr,
+};
```
joshka, line 4:
```suggestion
use std::io::{self, stdout};
```

kdheepak:
I've updated the PR description. This should be ready to squash. @joshka you want to do the honors?

Issue #497, opened 2023-09-13
Layout: Configure padding / gaps between areas in layouts


joshka:
## Problem
Often I want an empty space (row or column) in between areas. Usually these are the same width or height.
The way to do this currently is to add a `Constraint::Length(1)` between each of the other constraints and then go and bump every index up by the amount that this shifts the output array. E.g.:
```rust
Paragraph::new("foo").render(areas[0], buf);
Paragraph::new("bar").render(areas[2], buf); // etc.
```
It would be nice if instead we could ignore these gaps.

- [ ] I'm not sure what the best name for this is (padding / gap / other?)
- [ ] I'm not sure if we should support different gaps between elements (i.e. a constraint that is a gap constraint) or just a single gap amount for the entire layout (I'd like to hear other people's perspective on this before anyone implements this)

## Solution
The simplest solution is to add
```rust
gap: u16,
```
and the appropriate builder method (`fn gap()`)
and then factor that into the distance between the end and start of areas:
https://github.com/ratatui-org/ratatui/blob/a7bf4b3f36f3281017d112ac1a67af7e82308261/src/layout.rs#L561-L564

## Alternatives
A more advanced option would add an extra variant to `Constraint` that inserts an ignored gap (I'm unsure how complex we'd want to go for this, but this might be useful for more complex layouts.

## Additional context
This would probably simplify some of the rendering logic in Table.

Valentin271:
I think this can be closed after #821. They might be some other cases:

- different spacing size for each space
- space around areas

But we may create more specific issues for these if it turns out we need it.

PR #821, opened 2024-01-15
feat: Add layout spacing ✨

kdheepak:
This PR adds a `spacing` feature for layouts

Spacing: 0

image

Spacing: 5

image

### DIFF
--- a/src/layout/layout.rs
+++ b/src/layout/layout.rs
@@ -91,6 +91,7 @@ pub struct Layout {
     constraints: Vec<Constraint>,
     margin: Margin,
     flex: Flex,
+    spacing: u16,
 }
 
 /// A container used by the solver inside split
@@ -137,9 +138,8 @@ impl Layout {
     {
         Layout {
             direction,
-            margin: Margin::new(0, 0),
             constraints: constraints.into_iter().map(Into::into).collect(),
-            flex: Flex::default(),
+            ..Layout::default()
         }
     }
 
@@ -347,6 +347,12 @@ impl Layout {
         self
     }
 
+    /// Sets spacing options for adding padding
+    pub const fn spacing(mut self, spacing: u16) -> Layout {
+        self.spacing = spacing;
+        self
+    }
+
     /// Set whether chunks should be of equal size.
     ///
     /// This determines how the space is distributed when the constraints are satisfied. By default,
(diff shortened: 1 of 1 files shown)


src/layout/layout.rs
```diff
@@ -347,6 +347,12 @@ impl Layout {
         self
     }
 
+    /// Sets spacing options for adding padding
```
Valentin271, line 350: Can we specify padding is added `between` items and not around them.

src/layout/layout.rs
```diff
@@ -434,10 +440,27 @@ impl Layout {
...
+            )
+        };
+        let constraints = if spacers_added {
+            Itertools::intersperse(
+                layout.constraints.iter().cloned(),
+                Constraint::Fixed(layout.spacing),
+            )
+            .collect_vec()
+        } else {
+            layout.constraints.iter().cloned().collect_vec()
```
Valentin271, line 458: `.clone()` seems to be enough here
```suggestion
            layout.constraints.clone()
```

Valentin271 (commented):
I like this, I think this is a good first and easy step.

Relates to #497.

src/layout/layout.rs
```diff
@@ -341,12 +341,66 @@ impl Layout {
...
+    /// ```
     pub const fn flex(mut self, flex: Flex) -> Layout {
         self.flex = flex;
         self
     }
 
+    /// Sets the spacing between items in the layout.
+    ///
+    /// The `spacing` method sets the spacing between items in the layout. The spacing is applied
+    /// evenly between all items. The spacing value represents the number of pixels between each
```
Valentin271, line 383:
```suggestion
    /// evenly between all items. The spacing value represents the number of cells between each
```

src/layout/layout.rs
```diff
@@ -341,12 +341,66 @@ impl Layout {
...
+    /// item.
+    ///
+    /// # Examples
+    ///
+    /// In this example, the spacing between each item in the layout is set to 2 pixels.
```
Valentin271, line 388:
```suggestion
    /// In this example, the spacing between each item in the layout is set to 2 cells.
```

Valentin271 (approved):
`pixel` is going to be confusing I think, let's change it for `cell`?

Otherwise looks good to me!

joshka:
That "broken" test seems irksome. Can we add an issue to track it?

kdheepak:
I added a issue: https://github.com/ratatui-org/ratatui/issues/827

PR #823, opened 2024-01-15
docs(scrollbar): complete scrollbar documentation

Valentin271:

### DIFF
--- a/src/widgets/scrollbar.rs
+++ b/src/widgets/scrollbar.rs
@@ -1,3 +1,4 @@
+#![warn(missing_docs)]
 use strum::{Display, EnumString};
 
 use super::StatefulWidget;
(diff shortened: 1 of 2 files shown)


src/widgets/scrollbar.rs
```diff
@@ -117,19 +137,33 @@ impl ScrollbarState {
...
-/// Scrollbar Orientation
+/// This is the position of the scrollbar around a given area.
+///
+/// ```plain
+///           HorizontalTop
+///             ┌───────┐
+/// VerticalLeft│       │VerticalRight
+///             └───────┘
+///          HorizontalBottom
+/// ```
```
kdheepak, line 148: I think scrollbar orientation should be:

- `Horizontal`
- `Vertical`

With another `ScrollbarPosition` as `Start` and `End`.

Maybe leave this undocumented for now?

orhun (approved):

src/widgets/scrollbar.rs
Valentin271: For me scrollbar _orientation_ isn't even a good name. There is definitely room for improvement here. However, in the meantime those improvement are being thought/worked on I'm for keeping the documentation so that users and future maintainers know what this is.

kdheepak: Makes sense! We should open an issue to track that discussion.

kdheepak (approved):

PR #824, opened 2024-01-15
feat: Table column calculation uses layout spacing ✨

kdheepak:
This uses the new `spacing` feature of the `Layout` struct to allocate columns spacing in the `Table` widget.
This changes the behavior of the table column layout in the following ways:

1. Selection width is always allocated.
    - if a user does not want a selection width ever they should use `HighlightSpacing::Never`
2. Column spacing is prioritized over other constraints
    - if a user does not want column spacing, they should use `Table::new(...).column_spacing(0)`

### DIFF
--- a/src/widgets/table/table.rs
+++ b/src/widgets/table/table.rs
@@ -1197,21 +1197,15 @@ mod tests {
         #[test]
         fn length_constraint() {
             // without selection, more than needed width
-            let table = Table::default()
-                .widths([Length(4), Length(4)])
-                .segment_size(SegmentSize::None);
+            let table = Table::default().widths([Length(4), Length(4)]);
             assert_eq!(table.get_columns_widths(20, 0), [(0, 4), (5, 4)]);
 
             // with selection, more than needed width
-            let table = Table::default()
-                .widths([Length(4), Length(4)])
-                .segment_size(SegmentSize::None);
+            let table = Table::default().widths([Length(4), Length(4)]);
             assert_eq!(table.get_columns_widths(20, 3), [(3, 4), (8, 4)]);
 
             // without selection, less than needed width
-            let table = Table::default()
-                .widths([Length(4), Length(4)])
-                .segment_size(SegmentSize::None);
+            let table = Table::default().widths([Length(4), Length(4)]);
             assert_eq!(table.get_columns_widths(7, 0), [(0, 4), (5, 2)]);
 
             // with selection, less than needed width
@@ -1220,36 +1214,26 @@ mod tests {
             // │ (3, 3) │x│ (7, 0) │
             // └────────┘x└────────┘
             // column spacing (i.e. `x`) is always prioritized
-            let table = Table::default()
-                .widths([Length(4), Length(4)])
-                .segment_size(SegmentSize::None);
+            let table = Table::default().widths([Length(4), Length(4)]);
             assert_eq!(table.get_columns_widths(7, 3), [(3, 3), (7, 0)]);
         }
 
         #[test]
         fn max_constraint() {
             // without selection, more than needed width
-            let table = Table::default()
-                .widths([Max(4), Max(4)])
-                .segment_size(SegmentSize::None);
+            let table = Table::default().widths([Max(4), Max(4)]);
             assert_eq!(table.get_columns_widths(20, 0), [(0, 4), (5, 4)]);
 
             // with selection, more than needed width
-            let table = Table::default()
-                .widths([Max(4), Max(4)])
-                .segment_size(SegmentSize::None);
+            let table = Table::default().widths([Max(4), Max(4)]);
             assert_eq!(table.get_columns_widths(20, 3), [(3, 4), (8, 4)]);
 
             // without selection, less than needed width
-            let table = Table::default()
-                .widths([Max(4), Max(4)])
-                .segment_size(SegmentSize::None);
+            let table = Table::default().widths([Max(4), Max(4)]);
             assert_eq!(table.get_columns_widths(7, 0), [(0, 4), (5, 2)]);
 
             // with selection, less than needed width
-            let table = Table::default()
-                .widths([Max(4), Max(4)])
-                .segment_size(SegmentSize::None);
+            let table = Table::default().widths([Max(4), Max(4)]);
             assert_eq!(table.get_columns_widths(7, 3), [(3, 3), (7, 0)]);
         }
 
@@ -1260,58 +1244,42 @@ mod tests {
             // constraint and not split it with all available constraints
 
             // without selection, more than needed width
-            let table = Table::default()
-                .widths([Min(4), Min(4)])
-                .segment_size(SegmentSize::None);
+            let table = Table::default().widths([Min(4), Min(4)]);
             assert_eq!(table.get_columns_widths(20, 0), [(0, 4), (5, 4)]);
 
             // with selection, more than needed width
-            let table = Table::default()
-                .widths([Min(4), Min(4)])
-                .segment_size(SegmentSize::None);
+            let table = Table::default().widths([Min(4), Min(4)]);
             assert_eq!(table.get_columns_widths(20, 3), [(3, 4), (8, 4)]);
 
             // without selection, less than needed width
             // allocates spacer
-            let table = Table::default()
-                .widths([Min(4), Min(4)])
-                .segment_size(SegmentSize::None);
+            let table = Table::default().widths([Min(4), Min(4)]);
             assert_eq!(table.get_columns_widths(7, 0), [(0, 4), (5, 2)]);
 
             // with selection, less than needed width
             // always allocates selection and spacer
-            let table = Table::default()
-                .widths([Min(4), Min(4)])
-                .segment_size(SegmentSize::None);
+            let table = Table::default().widths([Min(4), Min(4)]);
             assert_eq!(table.get_columns_widths(7, 3), [(3, 3), (7, 0)]);
         }
 
         #[test]
         fn percentage_constraint() {
             // without selection, more than needed width
-            let table = Table::default()
-                .widths([Percentage(30), Percentage(30)])
-                .segment_size(SegmentSize::None);
+            let table = Table::default().widths([Percentage(30), Percentage(30)]);
             assert_eq!(table.get_columns_widths(20, 0), [(0, 6), (7, 6)]);
 
             // with selection, more than needed width
-            let table = Table::default()
-                .widths([Percentage(30), Percentage(30)])
-                .segment_size(SegmentSize::None);
+            let table = Table::default().widths([Percentage(30), Percentage(30)]);
             assert_eq!(table.get_columns_widths(20, 3), [(3, 5), (9, 5)]);
 
             // without selection, less than needed width
             // rounds from positions: [0.0, 0.0, 2.1, 3.1, 5.2, 7.0]
-            let table = Table::default()
-                .widths([Percentage(30), Percentage(30)])
-                .segment_size(SegmentSize::None);
+            let table = Table::default().widths([Percentage(30), Percentage(30)]);
             assert_eq!(table.get_columns_widths(7, 0), [(0, 2), (3, 2)]);
 
             // with selection, less than needed width
             // rounds from positions: [0.0, 3.0, 5.1, 6.1, 7.0, 7.0]
-            let table = Table::default()
-                .widths([Percentage(30), Percentage(30)])
-                .segment_size(SegmentSize::None);
+            let table = Table::default().widths([Percentage(30), Percentage(30)]);
             assert_eq!(table.get_columns_widths(7, 3), [(3, 1), (5, 1)]);
         }
 
@@ -1319,39 +1287,29 @@ mod tests {
         fn ratio_constraint() {
             // without selection, more than needed width
             // rounds from positions: [0.00, 0.00, 6.67, 7.67, 14.33]
-            let table = Table::default()
-                .widths([Ratio(1, 3), Ratio(1, 3)])
-                .segment_size(SegmentSize::None);
+            let table = Table::default().widths([Ratio(1, 3), Ratio(1, 3)]);
             assert_eq!(table.get_columns_widths(20, 0), [(0, 7), (8, 6)]);
 
             // with selection, more than needed width
             // rounds from positions: [0.00, 3.00, 10.67, 17.33, 20.00]
-            let table = Table::default()
-                .widths([Ratio(1, 3), Ratio(1, 3)])
-                .segment_size(SegmentSize::None);
+            let table = Table::default().widths([Ratio(1, 3), Ratio(1, 3)]);
             assert_eq!(table.get_columns_widths(20, 3), [(3, 6), (10, 5)]);
 
             // without selection, less than needed width
             // rounds from positions: [0.00, 2.33, 3.33, 5.66, 7.00]
-            let table = Table::default()
-                .widths([Ratio(1, 3), Ratio(1, 3)])
-                .segment_size(SegmentSize::None);
+            let table = Table::default().widths([Ratio(1, 3), Ratio(1, 3)]);
             assert_eq!(table.get_columns_widths(7, 0), [(0, 2), (3, 3)]);
 
             // with selection, less than needed width
             // rounds from positions: [0.00, 3.00, 5.33, 6.33, 7.00, 7.00]
-            let table = Table::default()
-                .widths([Ratio(1, 3), Ratio(1, 3)])
-                .segment_size(SegmentSize::None);
+            let table = Table::default().widths([Ratio(1, 3), Ratio(1, 3)]);
             assert_eq!(table.get_columns_widths(7, 3), [(3, 1), (5, 2)]);
         }
 
         /// When more width is available than requested, the behavior is controlled by segment_size
         #[test]
         fn underconstrained() {
-            let table = Table::default()
-                .widths([Min(10), Min(10), Min(1)])
-                .segment_size(SegmentSize::None);
+            let table = Table::default().widths([Min(10), Min(10), Min(1)]);
             assert_eq!(
                 table.get_columns_widths(62, 0),
                 &[(0, 10), (11, 10), (22, 1)]
--- a/src/widgets/table/table.rs
+++ b/src/widgets/table/table.rs
@@ -1386,7 +1386,7 @@ mod tests {
         }
 
         #[test]
-        fn spacers_priority_when_excess_column_width_available() {
+        fn excess_area_highlight_symbol_and_column_spacing_allocation() {
             // no highlight_symbol rendered ever
             assert_buffer_eq!(
                 test_table_with_selection(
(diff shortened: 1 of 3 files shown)


joshka:
> Selection width is always allocated. 

Does this make it breaking?

kdheepak:
I don’t consider it breaking because previously what the solver did was unstable and not well defined. There is a change in behavior from what it mostly likely did to what it now will always do. I would be very surprised if a user was relying on this behavior because it would probably flicker a lot. 

That said I’m happy to mark it as breaking so it stands out in the changelogs.

joshka:
Ah, I think I misinterpreted what you were saying. You're saying that we allocate the space even if it's zero (due to being no selection). I thought you meant we always allocate the non-zero width event if there's no selection. Am I mistaken on that?

kdheepak:
> we allocate the space even if it's zero (due to being no selection)

Yes. Specifically, we allocate the space for the selection using 

```rust
let [_selection_area, columns_area] =
    Rect::new(0, 0, max_width, 1).split(&Layout::horizontal([
        Constraint::Fixed(selection_width),
        Constraint::Proportional(0),
    ]));
```

and only render columns to the `columns_area`. 

So the column for `highlight_symbol` is always allocated when it should be (when selections are made and when highlight spacing is `WhenSelected` OR when highlight spacing is `WhenSelected`). 

`column_spacing` is added as a `Fixed` constraint into the `columns_area`, so it will be prioritized the most when the column widths are being calculated.

kdheepak:
I added more visual tests for specifying this exact behavior into the tests.

src/widgets/table/table.rs
```diff
@@ -727,30 +725,23 @@ impl Table<'_> {
                 .map(|r| r.cells.len())
                 .max()
                 .unwrap_or(0);
-            // There are `col_count - 1` spaces between the columns
-            let total_space =
-                max_width.saturating_sub(self.column_spacing * col_count.saturating_sub(1) as u16);
-            // Divide the remaining space between each column equally
-            vec![Constraint::Length(total_space / col_count.max(1) as u16); col_count]
+            // Divide the space between each column equally
+            vec![Constraint::Length(max_width / col_count.max(1) as u16); col_count]
```
joshka, line 729: Q: if we made the spacers take away space before the relative constraints (ratio/percentage) are applied (or if this is the way things work already), then could this be expressed as `[Ratio(1, col_count); col_count]`?

src/widgets/table/table.rs
```diff
@@ -1248,12 +1240,17 @@ mod tests {
...
+            // └────────┘x└────────┘
+            // column spacing (i.e. `x`) is always prioritized
             test(
                 &[Length(4), Length(4)],
                 SegmentSize::None,
                 7,
                 3,
-                &[(3, 4), (7, 0)],
+                &[(3, 3), (7, 0)],
             );
```
joshka: This is an example where using DRY for tests makes the tests harder to understand, not easier.
```suggestion
            // with selection, less than needed width
            // <--------7px-------->
            // ┌────────┐x┌────────┐
            // │ (3, 3) │x│ (7, 0) │
            // └────────┘x└────────┘
            // column spacing (i.e. `x`) is always prioritized
            let table = Table::default()
                .widths([Length(4), Length(4)])
                .segment_size(SegmentSize::None);
            assert_eq!(table.get_columns_widths(7, 3), [(3, 3), (7, 0)]);
```

joshka: fixed in 3475ea92409b03fb1a9ac2bc3f41f3889b4f402d

joshka: and 161492f1996b5af1042e2cf1104343832d3c07c2

src/widgets/table/table.rs
```diff
@@ -1484,7 +1364,264 @@ mod tests {
                 .rows(vec![])
                 .footer(Row::new(vec!["h", "i"]))
                 .column_spacing(0);
-            assert_eq!(table.get_columns_widths(10, 0), &[(0, 5), (5, 5)])
+            assert_eq!(table.get_columns_widths(10, 0), [(0, 5), (5, 5)])
+        }
+
+        #[test]
+        fn spacers_priority_unconstrained() {
```
joshka, line 1371: For a reader looking at this code for the first time, neither priority or unconstrained have any meaning with respect to the wider ratatui context, and so are floating undefined abstract concepts. Choosing a test name that is more grounded, defined, concrete would help make it obvious what this test is testing.

src/widgets/table/table.rs
```diff
@@ -1484,7 +1364,264 @@ mod tests {
...
+                let mut buf = Buffer::empty(area);
+                if let Some(s) = selection {
+                    let mut state = TableState::default().with_selected(Some(s));
+                    StatefulWidget::render(table, area, &mut buf, &mut state);
+                    assert_buffer_eq!(buf, expected);
+                } else {
+                    Widget::render(table, area, &mut buf);
+                    assert_buffer_eq!(buf, expected);
+                }
+            }
```
joshka, line 1392: Continuing, here there's a chunk of code which provides context for each thing tested below. The reader has to memorize how this works and then apply that to the tests below in order to understand what's being tested.

This would be better pulled out as a function that just does the rendering.

Given that there are two paths, and the path chosen is based on passing either `None` or `Some(_)` those paths would be better expressed with two obvious methods - `render()`, `render_with_selection()`

src/widgets/table/table.rs
```diff
@@ -1484,7 +1364,264 @@ mod tests {
...
+
+        #[test]
+        fn spacers_priority_unconstrained() {
+            fn test(
+                highlight_spacing: HighlightSpacing,
+                selection: Option<usize>,
+                expected: Buffer,
+            ) {
+                let table = Table::default()
+                    .rows(vec![Row::new(vec!["hello", "world"])])
```
joshka, line 1378: nit: it might be better to pick two words that don't tend to be presented as a single string, as it highlights that these are separate columns rather than a single common sentence. E.g. "ABCDE", "12345" might be more obvious.

src/widgets/table/table.rs
kdheepak: Right now `[Ratio(1, col_count); col_count]` does the same thing as `[Length(max_width / col_count); col_count]`, but that may not be true in the future. In #788, `Length` is higher priority than `Ratio`, so I'd prefer we use `Length`.

> spacers take away space before the relative constraints

I'm not sure if I'm not following what you asking here specifically. I think it won't be possible to make spacers be higher priority because then spacers would allocate the entire space over ratio or percentage.

src/widgets/table/table.rs
```diff
@@ -1484,7 +1364,264 @@ mod tests {
...
+            // no highlight_symbol rendered ever
+            test(
+                HighlightSpacing::Never,
+                None,
+                Buffer::with_lines(vec![
+                    "hello  world   ", // row 1
+                    "               ", // row 2
+                    "               ", // row 3
+                ]),
+            );
```
joshka, line 1403: Taking the advice from above, this might look better as:
```rust

assert_buffer_eq!(
  render_test_table(HighlightSpacing::Never),
  Buffer::with_lines(vec![
     "ABCDE  12345   ",
     "               ",
     "               "
  ]),
);
```

src/widgets/table/table.rs
```diff
@@ -1484,7 +1364,264 @@ mod tests {
...
+                    assert_buffer_eq!(buf, expected);
+                }
+            }
+
+            // no highlight_symbol rendered ever
+            test(
+                HighlightSpacing::Never,
+                None,
+                Buffer::with_lines(vec![
+                    "hello  world   ", // row 1
```
joshka, line 1399: why does this have two spaces if column spacing is 0?

src/widgets/table/table.rs
```diff
@@ -1484,7 +1364,264 @@ mod tests {
...
+                    let mut state = TableState::default().with_selected(Some(s));
+                    StatefulWidget::render(table, area, &mut buf, &mut state);
+                    assert_buffer_eq!(buf, expected);
+                } else {
+                    Widget::render(table, area, &mut buf);
+                    assert_buffer_eq!(buf, expected);
+                }
+            }
+
+            // no highlight_symbol rendered ever
```
joshka, line 1394: It wouldn't hurt to pull these out to individual methods

src/widgets/table/table.rs
```diff
@@ -1484,7 +1364,264 @@ mod tests {
...
+                Some(0),
+                Buffer::with_lines(vec![
+                    "hello  world   ", // row 1
+                    "               ", // row 2
+                    "               ", // row 3
+                ]),
+            );
+            // highlight_symbol rendered because selection is made
+            test(
+                HighlightSpacing::WhenSelected,
```
joshka, line 1437: it probably makes more sense to group the tests by HighlightSpacing value to be able to contrast how for each setting interacts with the selection

src/widgets/table/table.rs
```diff
@@ -1484,7 +1364,264 @@ mod tests {
...
+                None,
+                Buffer::with_lines(vec![
+                    "hello 1234", // row 1
+                    "          ", // row 2
+                    "          ", // row 3
+                ]),
+            );
+            // highlight_symbol space always allocated
+            // because column widths is calculated after
+            // when column widths are calculated, column spacing is prioritized
```
joshka, line 1502: This comment requires some implied understanding of the implementation. Taking a step back from this, how would you explain it if someone didn't know about the implementation detail.

Th comment is difficult to parse in a way that makes grammatical sense "highlight_symbol space always allocated because _column widths is calculated after when column widths are calculated_, column spacing is prioritized". I'm assuming those are three sentences perhaps?

What is the actual property that is being tested for correctness with this test? highlight symbol or column spacing priority? (It's not clear)

src/widgets/table/table.rs
```diff
@@ -1484,7 +1364,264 @@ mod tests {
...
+                    assert_buffer_eq!(buf, expected);
+                }
+            }
+
+            // column spacing is prioritized over everything constraint
+            test(
+                HighlightSpacing::Never,
+                None,
+                Buffer::with_lines(vec![
+                    "hello 1234", // row 1
```
joshka, line 1486: Is the use of truncation in these tests significant with respect to the tested properties, if so call it out, if not avoid it as it puts two behaviors on display and it's difficult to understand which is the behavior that is being tested and which is incidental.

tests/widgets_table.rs
```diff
@@ -399,26 +399,28 @@ fn widgets_table_columns_widths_can_use_mixed_constraints() {
...
-            "│Row11             Row12     │",
-            "│Row21             Row22     │",
-            "│Row31             Row32     │",
-            "│Row41             Row42     │",
-            "│                            │",
-            "│                            │",
-            "└────────────────────────────┘",
-        ]),
-    );
+    // This test is unstable and should not be in the test suite
```
joshka, line 402: Was it unstable before this change? I don't think we've seen test failures, but perhaps it has been always cached and so never indeterminate? Can you cut an issue for this to investigate it at some point?

joshka (approved):
General approval of the change.

The tests could bear improving if they are designed to help a future reader to understand the change. I think I'd find it difficult to understand what to do if a future change broke one of these tests. If they are purely to test the implementation right now, then they are fine.

src/widgets/table/table.rs
joshka: What I mean is that if you have [Percentage(50); 2] and spacer = 10 splitting a space of 100, then you really don't have a 50% constraint, the constraints are 45% now. And in fact you can likely never write a percentage that meaningfully makes sense when spacers are taking some of that space. So that implies that the spacers should remove the space available to split before the constraints get applied. The split impl would then need to first remove the spacer space leaving 90 to split up, and then when that calculation of percentage -> solver constraint happens it would be a percentage of that instead of the total area.
```rust
                    solver.add_constraint(element.size() | EQ(STRONG) | (area_size * percent))?;
```

I haven't looked at the implementation to see if that's how this currently works though, or just how I'd intuitively think about it naively (which is perhaps the right way to think of this for our users).

tests/widgets_table.rs
kdheepak: It was unstable but rarely failed. But if you look through the recent PRs this test has changed a number of times by one column. I talk more about unstable solutions in #788

joshka: Makes sense.

src/widgets/table/table.rs
joshka: Right now isn't that just because of how the constraints interact and then fail when they can't all be met though?
That [p(100), min(20) is 3 cassowary constraints: [a=100(strong), b>=20(strong), b=20(medium)], which just happens to end up with [80,20]. Same deal with spacers I think right [a = 100(strong), sp = 1(required-1),..]

kdheepak: > isn't that just because of how the constraints interact

Yes. But also, I don't see how it can work well any other way.

As context, I don't think it is possible to make `Percentage` (or `Length` or `Ratio`) higher priority than `Min` / `Max` because if you do you'll end up in cases where `Min` and `Max` stop working as intended. 

At best, `Length`, `Percentage` and `Ratio` can be the same priority as `Min` and `Max` (which is what we have now). In #788 `Length` is the same priority as `Min` and `Max` but `Percentage` and `Ratio` are lower priority. 

Because `Percentage` can't have a higher priority, and because it can have a lower priority, it is easier to think of `Percentage(x)` as filling x % of available space after higher priority constraints are resolved.

joshka: Got it - resolving this for now - it was mostly curiosity driven.

src/widgets/table/table.rs
kdheepak: Good idea. I've updated the test name now.

src/widgets/table/table.rs
kdheepak: Changed to `ABCDE` and `12345`.

src/widgets/table/table.rs
kdheepak: When widths are not given, column constraints are calculated as `vec![Constraint::Length(max_width / col_count.max(1) as u16); col_count]`, i.e. columns are split equally among available space. Here we get `Length(7)`.

This is a little odd behavior considering default layout for the table is `Flex::Start`. It would be nice to use some kind of `Proportional` constraint by default based on number of characters in each row on average. But that is out of scope for this PR and can be discussed later.

src/widgets/table/table.rs
kdheepak: good catch, this comment was garbled. I've updated it with something that is easier to parse. Also added more tests to describe what is going on.

src/widgets/table/table.rs
kdheepak: Done

kdheepak: Thanks for this suggestion. I think we should add examples to the website for how to write tests when submitting a PR.

src/widgets/table/table.rs
kdheepak: I think it would be better to change the default to be `Flex::Stretch` here as well.

src/widgets/table/table.rs
kdheepak: I think these tests all fall into the same category, i.e. how is space allocated when there's lots of room to do so. Thoughts?

src/widgets/table/table.rs
kdheepak: Good idea! Done in most recent commit.

src/widgets/table/table.rs
kdheepak: I'm going to leave this the way it is for now.

kdheepak:
I've improved the tests. I think once the priorities are set in stone, we should add a lot more tests and cover all the corner cases.

src/widgets/table/table.rs
joshka: Changing the defaults for layout type things require an proper understanding of how people are using these widgets and what and how they'd expect stability in their UI between versions. I'm generally willing to break compatibility in some things, but not without understanding the impact of that change. This seems like it would be problematic (even if it is the right thing to do)

joshka: Regarding the tests, perhaps it would be the right thing to do to specify the widths properly so that the columns are clear and obvious. If I had to ask about why this is 2 spaces, then later when these tests break it will be difficult to understand why the 2 space or 1 space made sense.

src/widgets/table/table.rs
joshka: This was the lowest priority comment - it's only a thing because each assertion has a comment, which would make a perfect name for the test. That said, the extra comments on each individual assertion suggest that there's enough complexity to generally do this I think.

src/widgets/table/table.rs
joshka: This still introduces two concepts that aren't actually defined anywhere.
Spacers - is a layout concept, but doesn't appear in the public API of Table at all, so it's not clear how this relates.
Priority - many of the comments introduce or reference priority, but the reference is single sided comparison e.g. "highlight_symbol and spacing are prioritized". Prioritized compared to what?

kdheepak: Trying these names:

- excess_area_highlight_symbol_and_column_space_allocation
- insufficient_area_highlight_symbol_and_column_space_allocation
- insufficient_area_highlight_symbol_allocation_with_no_column_spacing

joshka: what is the verb? (i.e. the behavior that is being tested)

kdheepak:
This was good feedback! 

I'm going to merge this and revisit when I work on the tests for constraint priorities. I'll resolve the conversations, merge and then unresolve them again.

src/widgets/table/table.rs
kdheepak: We are testing that "highlight_symbol and column spaces are allocated" in three cases

- when there's excess area to do so
- when there's insufficient area to do so

We are also testing that highlight_symbol is allocated when column spacing is zero

- when there's insufficient area to do so

joshka: - renders_columns_with_column_spacing_and_enough_space
- renders_columns_with_column_spacing_and_insufficient_space
- renders_columns_with_no_column_spacing_and_insufficient_space

joshka: If you're leaving the multiple tests per function the "how does this render the columns" part becomes the comment as well as the "with these extra relevant constraints that modify the behavior"

kdheepak: Specifically we are checking that highlight spacing is working the way we expect it to. Your names don’t have anything about highlight spacing.

src/widgets/table/table.rs
```diff
@@ -1486,6 +1483,376 @@ mod tests {
...
+        #[test]
+        fn spacers_priority_when_excess_column_width_available() {
+            // no highlight_symbol rendered ever
+            assert_buffer_eq!(
+                test_table_with_selection(
+                    HighlightSpacing::Never,
+                    15,   // width
+                    0,    // spacing
+                    None, // selection
+                ),
```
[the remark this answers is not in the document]
joshka: additionally, if made into an rstest, the default table could be a fixture:
```rust
#[fixture]
fn abcde_table() -> Table {
...
}

#[fixture]
fn buf() -> &mut Buffer {
...
}

#[rstest]
fn with_highlight_never_and_no_column(abcde_table: Table, buf: &mut Buffer) {
    abcde_table
        .highlight_spacing(HighlightSpacing::Never)
        .column_spacing(0)
        .render(buf.area, buf);
    assert_buffer_eq!(
        buf,
        ...expected...
    );
}
```

src/widgets/table/table.rs
joshka: - renders_highlight_spacing_when...

Issue #816, opened 2024-01-14
implement `Display` for `Text`, `Line`, `Span`


Valentin271:
## Problem
Someone on discord wanted to check the actual content of a `Text`, ignoring the style.

One solution is to loop over the inner lines and span manually. I'm wondering if implementing `Display` is worth it to get a string representation of `Text`?

## Solution

Implement `Display` for `Text`, `Line` and `Span` at least.

## Alternatives
Just let users code it themselves as the fields are all public and this might not be a wide use case.

orhun:
I remember hitting the same case in the past. I think that the `Display` implementation or a method such as `raw_text()` would be nice.

Valentin271:
Sure go ahead @Emivvvvv

PR #825, opened 2024-01-15
feat: implement Display for Text, Line, Span

Emivvvvv:
Issue: https://github.com/ratatui-org/ratatui/issues/816

This PR adds:

`std::fmt::Display` for `Text`, `Line`, and `Span` structs.

Display implementation displays actual content while ignoring style.

### DIFF
--- a/src/text/line.rs
+++ b/src/text/line.rs
@@ -381,6 +381,15 @@ impl Widget for Line<'_> {
     }
 }
 
+impl std::fmt::Display for Line<'_> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        for span in &self.spans {
+            write!(f, "{span}")?;
+        }
+        Ok(())
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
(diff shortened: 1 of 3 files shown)


src/text/line.rs
joshka, line 1: Can you please add two tests:
- a line that has a style is rendered plain
- a line with spans that have styles is rendered plain

I think only really the first test "display_line_from_vec" is needed below. The other tests seem like they test the internals of how strings get converted to a line, so it seems like they test code that is already tested.

src/text/text.rs
```diff
@@ -225,6 +225,19 @@ where
...
 
+impl std::fmt::Display for Text<'_> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let vec_len = self.lines.len();
+        for (i, line) in self.lines.iter().enumerate() {
+            write!(f, "{line}")?;
+            if (i + 1) != vec_len {
+                writeln!(f)?
+            }
+        }
```
Valentin271, line 236: This can be simplified to
```suggestion
        for line in &self.lines {
            writeln!(f, "{line}")?;
        }
```

This will add a trailing `\n`, but is it a problem really?

joshka, line 236: One alternate way of doing this is to use https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.with_position
```suggestion
        for (position, line) in self.lines.iter().with_positions() {
            if position == Position::Last {
                write(f, "{line}")?;
            } else {
                writeln(f, "{line}")?;
            }
        }
```

joshka (approved):
LGTM

src/text/text.rs
Valentin271: That works too, it just depends on whether we want an ending `\n`. I don't know what makes the most sense.

joshka: We likely don't - think of the single line case - that should render "Hello", two lines "Hello\nWorld".

joshka:
What happened here?

orhun:
I've been talking with @Emivvvvv on Discord for guidance, asked him what happened :D

orhun:
There was a misunderstanding about the gpg signing and now we can't reopen this PR since the branch is re-created.

Emivvvvv:
I'm sorry because of the mess. This is my first time 😬. Like @orhun said I wanted to make the commit verified but because of my limited knowledge about how git works I removed the branch to be able to re-send a verified commit. But apparently I could have used `--amend`. I'm opening a new PR shortly. Again, sorry about this mess.

joshka:
No worries :)

PR #826, opened 2024-01-15
feat: implement Display for Text, Line, Span

Emivvvvv:
Issue: https://github.com/ratatui-org/ratatui/issues/816

This PR adds:

`std::fmt::Display` for `Text`, `Line`, and `Span` structs.

Display implementation displays actual content while ignoring style.


joshka (approved):
LGTM

PR #830, opened 2024-01-16
feat: Add spacing feature to flex example ✨

kdheepak:
This adds the `spacing` using `+` and `-` to the flex example:

https://github.com/ratatui-org/ratatui/assets/1813121/1980206d-8558-47f0-8631-c74ba98b224d

This was useful to find and debug spacing related bugs like this: https://github.com/ratatui-org/ratatui/pull/829

### DIFF
--- a/examples/flex.rs
+++ b/examples/flex.rs
@@ -30,6 +30,7 @@ const EXAMPLE_DATA: &[(&str, &[Constraint])] = &[
     ("", &[Fixed(20), Percentage(20), Length(20)]),
     ("", &[Percentage(20), Length(20), Fixed(20)]),
     ("", &[Length(20), Length(15)]),
+    ("Spacing has no effect in `SpaceAround` and `SpaceBetween`", &[Proportional(1), Proportional(1)]),
     ("", &[Length(20), Fixed(20)]),
     (
         "When not using `Flex::Stretch` or `Flex::StretchLast`,\n`Min(u16)` and `Max(u16)` collapse to their lowest values",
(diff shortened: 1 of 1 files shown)


orhun (approved):

Valentin271:
I like this a lot, could we display the current spacing value somewhere? And maybe add a note that `Space*` don't support spacing.

I see you fixed the proportional bug in https://github.com/ratatui-org/ratatui/pull/829 (I rebased locally and tested). I still can see a bug with `Percentage` or maybe `Length`.

https://github.com/ratatui-org/ratatui/assets/36198422/90d58a36-c8d7-4130-bf52-4f26cbdd66fe

I also recommend bumping the layout cache to 200 because every `+` or `-` is a lot of "layouting" added. I can really feel it on my machine.

kdheepak:
> I like this a lot, could we display the current spacing value somewhere?

Good idea. I added it to the axis.

image

I also fixed the axis to show the correct width of the blocks when there is a scrollbar or not.

> And maybe add a note that Space* don't support spacing.

Done.

image

> I still can see a bug with Percentage or maybe Length.

Yes, that can only be fixed with this PR though: https://github.com/ratatui-org/ratatui/pull/788
Trying to get other PRs through and will work on the constraint weights at the end.

> I also recommend bumping the layout cache to 200

I bumped it to a higher number because I was seeing lag too.

Thanks for the comments!

Valentin271 (approved):
Works for me then!

PR #831, opened 2024-01-16
feat: Add one eighth inner and outer border set ✨

kdheepak:
This PR adds the [`McGugan`](https://www.willmcgugan.com/blog/tech/post/ceo-just-wants-to-draw-boxes/) border set, which allows for tighter borders.

For example, with the `flex` example you can get this effect (top is mcgugan wide, bottom is mcgugan tall):

image

image

As of this PR, `MCGUGAN_WIDE` has to be styled manually, like so:

```rust
            let main_color = color_for_constraint(*constraint);
            let cell = buf.get_mut(block.x, block.y + 1);
            cell.set_style(Style::reset().fg(main_color).reversed());
            let cell = buf.get_mut(block.x, block.y + 2);
            cell.set_style(Style::reset().fg(main_color).reversed());
            let cell = buf.get_mut(block.x + block.width.saturating_sub(1), block.y + 1);
            cell.set_style(Style::reset().fg(main_color).reversed());
            let cell = buf.get_mut(block.x + block.width.saturating_sub(1), block.y + 2);
            cell.set_style(Style::reset().fg(main_color).reversed());

```

`MCGUGAN_TALL` has to be styled manually, like so:

```rust
            let main_color = color_for_constraint(*constraint);
            for x in block.x + 1..(block.x + block.width).saturating_sub(1) {
                let cell = buf.get_mut(x, block.y);
                cell.set_style(Style::reset().fg(main_color).reversed());
                let cell = buf.get_mut(x, block.y + block.height - 1);
                cell.set_style(Style::reset().fg(main_color).reversed());
            }

```

### DIFF
--- a/src/symbols.rs
+++ b/src/symbols.rs
@@ -381,6 +381,49 @@ pub mod border {
         horizontal_top: QUADRANT_BOTTOM_HALF,
         horizontal_bottom: QUADRANT_TOP_HALF,
     };
+
+    pub const ONE_EIGHTH_TOP_EIGHT: &str = "▔";
+    pub const ONE_EIGHTH_BOTTOM_EIGHT: &str = "▁";
+    pub const ONE_EIGHTH_LEFT_EIGHT: &str = "▏";
+    pub const ONE_EIGHTH_RIGHT_EIGHT: &str = "▕";
+
+    /// Wide border set based on McGugan box technique
+    ///
+    /// ```text
+    /// ▁▁▁▁▁▁▁
+    /// ▏xxxxx▕
+    /// ▏xxxxx▕
+    /// ▔▔▔▔▔▔▔
+    /// ```
+    pub const ONE_EIGHTH_WIDE: Set = Set {
+        top_right: ONE_EIGHTH_BOTTOM_EIGHT,
+        top_left: ONE_EIGHTH_BOTTOM_EIGHT,
+        bottom_right: ONE_EIGHTH_TOP_EIGHT,
+        bottom_left: ONE_EIGHTH_TOP_EIGHT,
+        vertical_left: ONE_EIGHTH_LEFT_EIGHT,
+        vertical_right: ONE_EIGHTH_RIGHT_EIGHT,
+        horizontal_top: ONE_EIGHTH_BOTTOM_EIGHT,
+        horizontal_bottom: ONE_EIGHTH_TOP_EIGHT,
+    };
+
+    /// Tall border set based on McGugan box technique
+    ///
+    /// ```text
+    /// ▕▔▔▏
+    /// ▕xx▏
+    /// ▕xx▏
+    /// ▕▁▁▏
+    /// ```
+    pub const ONE_EIGHTH_TALL: Set = Set {
+        top_right: ONE_EIGHTH_LEFT_EIGHT,
+        top_left: ONE_EIGHTH_RIGHT_EIGHT,
+        bottom_right: ONE_EIGHTH_LEFT_EIGHT,
+        bottom_left: ONE_EIGHTH_RIGHT_EIGHT,
+        vertical_left: ONE_EIGHTH_RIGHT_EIGHT,
+        vertical_right: ONE_EIGHTH_LEFT_EIGHT,
+        horizontal_top: ONE_EIGHTH_TOP_EIGHT,
+        horizontal_bottom: ONE_EIGHTH_BOTTOM_EIGHT,
+    };
 }
 
 pub const DOT: &str = "•";


orhun (approved):

Valentin271:
I like this idea but I have a few concerns/questions about this:

1. I can clearly see here that the left and right parts are not of the same size. Is it just on my terminal?

image
Using the one quarter instead of the one eighth block fixes this (at least for me). 
image

There's also a difference on top vs left but there are no alternative so that's fine by me.

2. This is not the same as in the article, I'm fine with that I just want to make it clear. The article have corners, this does not (see below, plus you can clearly see the size difference here too).

image

The article actually is like this. (see again the border difference on the left)

image

The point of the article (as I understood it) was to have a new way of drawing boxes with (1) full borders, corners included, (2) put a background only inside the box.

I like the possibilities the second point gives, with the current implementation this is not possible. I also like this way of drawing because it leaves more padding horizontally than vertically, which is what we tend to see in most UIs. (the first example I found wasn't too far :smile: )
image

kdheepak:
Here's what the flex example looks like when I hack it together:

image

```rust
            let cell = buf.get_mut(block.x, block.y + 1);
            cell.set_style(Style::reset().fg(main_color).reversed());
            let cell = buf.get_mut(block.x, block.y + 2);
            cell.set_style(Style::reset().fg(main_color).reversed());
            let cell = buf.get_mut(block.x + block.width.saturating_sub(1), block.y + 1);
            cell.set_style(Style::reset().fg(main_color).reversed());
            let cell = buf.get_mut(block.x + block.width.saturating_sub(1), block.y + 2);
            cell.set_style(Style::reset().fg(main_color).reversed());
```

Here's the copy paste version:

```

<-------------------------------------------------107 px------------------------------------------------->

// Min(u16) takes any excess space when using `Stretch` or `StretchLast`                                  ▲
▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█
▏     Fixed(20)    ▕▏                             Min(20)                            ▕▏      Max(20)     ▕█
▏       20 px      ▕▏                              66 px                             ▕▏       20 px      ▕█
▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔█
// Proportional(u16) takes any excess space in all `Flex` layouts                                         █
▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█
▏    Length(20)    ▕▏  Percentage(20)   ▕▏    Ratio(1, 5)    ▕▏              Proportional(1)             ▕█
▏       20 px      ▕▏       21 px       ▕▏       21 px       ▕▏                   44 px                  ▕█
▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔█
// In `StretchLast`, last constraint of lowest priority takes excess space                                ║
▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁║
▏    Length(20)    ▕▏     Fixed(20)    ▕▏                         Percentage(20)                         ▕║
▏       20 px      ▕▏       20 px      ▕▏                              66 px                             ▕║
▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔║
▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁║
▏     Fixed(20)    ▕▏  Percentage(20)   ▕▏                          Length(20)                           ▕║
▏       20 px      ▕▏       21 px       ▕▏                             65 px                             ▕║
▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔║
▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁║
▏                         Percentage(20)                         ▕▏    Length(20)    ▕▏     Fixed(20)    ▕║
▏                              66 px                             ▕▏       20 px      ▕▏       20 px      ▕║
▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔▔║
```

kdheepak:
I've added both the inner and outer version.

> I can clearly see here that the left and right parts are not of the same size. Is it just on my terminal?

Maybe I copy pasted the wrong one when I first opened the PR? In the most recent commit, I searched on https://www.compart.com/en/unicode/ for the one eight left, right, top and bottom and copied them into my editor:

image

Can you check now?

Valentin271:
> I've added both the inner and outer version.

I think it's a good idea!

> Maybe I copy pasted the wrong one when I first opened the PR? In the most recent commit, I searched on https://www.compart.com/en/unicode/ for the one eight left, right, top and bottom and copied them into my editor

You did copy the left 1/8. That's what I was worried about ... on my machine left 1/8 is not the same size as right 1/8, it's actually left 1/4 which is the "right" size for right 1/8.
Semantically the characters you picked are better so lets keep them and just add a note for users about font support.

> I think to do this automatically, we have to implement styling borders specifically for this set. Because we have to color the left and right borders differently

We do, but looking at the code this doesn't look to complex or hard to maintain: https://github.com/ratatui-org/ratatui/blob/f71bf182975526aa2eca9ee710361f39db2d666d/src/widgets/block.rs#L613-L643

> Here's what the flex example looks like when I hack it together

I think what you're trying to do is not possible with `MCGUGAN_OUTER`, you want spacing between items but these borders end on their side of the block (i.e. left border is left of the block).

TLDR; I suggest

- keeping both borders set (if it wasn't already planned)
- keep the semantically correct characters
- Style `MCGUGAN_OUTER` so that left and right borders can't actually have a background and it is instead the background from the block.
- use `MCGUGAN_INNER` for the example

kdheepak:
> I think what you're trying to do is not possible with MCGUGAN_OUTER, you want spacing between items but these borders end on their side of the block (i.e. left border is left of the block).

Can you clarify what is not possible? If I want the effect that these rects are next to each other to showcase spacing, e.g.:

https://github.com/ratatui-org/ratatui/assets/1813121/67fec3de-d693-4303-8d4a-1d7b23739640

`MCGUGAN_OUTER` with a manual hack works well imo.

>    - keeping both borders set (if it wasn't already planned)
>    - keep the semantically correct characters
>    - Style MCGUGAN_OUTER so that left and right borders can't actually have a background and it is instead the background from the block.
>    - use MCGUGAN_INNER for the example, EDIT: you cound also use outer but put their style as something that hides them.

I agree with all of this! 

> You did copy the left 1/8. That's what I was worried about ... on my machine left 1/8 is not the same size as right 1/8, it's actually left 1/4 which is the "right" size for right 1/8.

I didn't know about that! What font are you using? I cycled through a few fonts in my terminal and they look the same for me.

Valentin271:
> I didn't know about that! What font are you using? I cycled through a few fonts in my terminal and they look the same for me.

Me neither and that's super weird, 1/8 should always be 1/8, that's the whole point it's relative.
I'm using [JetBrainsMonoNL Nerd Font](https://github.com/ryanoasis/nerd-fonts/blob/master/patched-fonts/JetBrainsMono/NoLigatures/Regular/JetBrainsMonoNLNerdFont-Regular.ttf) with `xfce4-terminal` (maybe it's terminal related but I don't think so).

> Can you clarify what is not possible? If I want the effect that these rects are next to each other to showcase spacing

Sure, what I meant was you can't have a gap between blocks with `MCGUGAN_OUTER` because they will touch each other. As opposed to the `PLAIN` border where two side by side blocks would leave a gap.

But that's not really relevant as I figured out after writing that that you could invert the borders and it would leave the gap as wanted.

I think the example could look even better by making all the `MCGUGAN_OUTER` appear hidden, just like the left and right ones.

joshka:
I'd suggest naming the characters just after their Unicode names directly rather than McGugan. Left eighth block is more informative than Michigan McGugan right as it highlights how the border touching actually works.

+1 for making this use quarter left and right for proportionality.

There's also a set of corner characters that allows for quarter vertical quarter and horizontal eighth coming in that supplemental legacy computing block. So someday this won't be necessary.

Trying to work out ways to make those chars more widely available in fonts prior to then (like getting the block into nerd fonts).

kdheepak:
> I'd suggest naming the characters just after their Unicode names 

Renamed!

> +1 for making this use quarter left and right for proportionality.

I'm not sure I follow:

The left one quarter block and right one quarter block are not the same width:

image

This is what it looks like in the terminal:

image

The one eighth blocks are all even:

image

kdheepak:
I am seeing that the quarter block left and right characters renders as the same width in Iosevka but the `flex` still looks weird:

With quarter block:

image

image

With one-eight block:

image

It seems like one eight block is a better default?

Valentin271:
I looks like the quarter block don't render exactly on the side of the cell. I do prefer personally the eight block for this PR.
But can actually include both.

joshka:
> I looks like the quarter block don't render exactly on the side of the cell.

This is also a font bug, it should render on the side of the cell with a little bit of overlap so that adjoining cells look contiguous.

I've been thinking about whether it's worth some time to progress https://github.com/ryanoasis/nerd-fonts/issues/1301 so that all the nerd fonts handle block drawing characters properly. I recently fixed up some missing glyphs for FiraCode in https://github.com/tonsky/FiraCode/pull/1566. The process was a bit fiddly, but it's reasonable.

It's unlikely that these changes would be different per font, which means that any implementation could be shared across all the SIL OFL licensed fonts that tend to have nerd font implementations.

kdheepak:
I'll plan to merge this as is, we can add quarter inner and outer border sets later when fonts support it better.

joshka (commented):
What about making this McGugan Wide ( corners use upper/lower blocks) and McGugan Tall (corners use left/right blocks)

```plain
▁▁▁▁
▏xx▕
▏xx▕
▔▔▔▔

▕▔▔▏
▕xx▏
▕xx▏
▕▁▁▏
```

image

kdheepak:
I like the wide and tall idea!

PR #836, opened 2024-01-17
docs(examples): refactor demo2

joshka:
Simplified a bunch of the logic in the demo2 example
- Moved destroy mode to its own file.
- Moved error handling to its own file.
- Removed AppContext
- Implemented Widget for &App. The app state is small enough that it
  doesn't matter here and we could just copy or clone the app state on
  every frame, but for larger apps this can be a significant performance
  improvement.
- Made the tabs stateful
- Made the term module just a collection of functions rather than a
  struct.
- Changed to use color_eyre for error handling.
- Changed keyboard shortcuts and rearranged the bottom bar.
- Use strum for the tabs enum.


joshka:
Includes #833

joshka:
Changed this to the `impl Widget for &App` approach. works fine.

examples/demo2/app.rs
```diff
@@ -1,252 +1,223 @@
...
             })
-            .context("terminal.draw")?;
+            .wrap_err("terminal.draw")?;
         Ok(())
     }
 
+    /// Handle events from the terminal.
+    ///
+    /// This function is called once per frame, The events are polled from the stdin with timeout of
+    /// 1/50th of a second. This was chosen to try to match the default frame rate of a GIF.
```
Valentin271, line 78: Is it the default or maximum framerate?

```suggestion
    /// 1/50th of a second. This was chosen to try to match the maximum frame rate of a GIF.
```

examples/demo2/app.rs
```diff
@@ -1,252 +1,223 @@
...
-            }
-        } else {
-            // move the pixel down one row
-            let dest_x = src_x;
-            let dest_y = src_y.saturating_add(1).min(area.bottom() - 2);
-            // copy the cell to the new location
-            let dest = buf.get_mut(dest_x, dest_y);
-            *dest = src;
-        }
+/// Implement RefWidget for &App rather than for App as we would otherwise have to clone or copy the
```
Valentin271, line 135:
```suggestion
/// Implement Widget for &App rather than for App as we would otherwise have to clone or copy the
```

examples/demo2/app.rs
```diff
@@ -1,252 +1,223 @@
...
+        let spans = keys
+            .iter()
+            .flat_map(|(key, desc)| {
+                let key = Span::styled(format!(" {} ", key), THEME.key_binding.key);
+                let desc = Span::styled(format!(" {} ", desc), THEME.key_binding.description);
+                [key, desc]
+            })
+            .collect_vec();
+        Line::from(spans)
+            .alignment(Alignment::Center)
```
Valentin271, line 198: If up to date with main.
```suggestion
        Line::from(spans)
            .centered()
```

Valentin271 (approved):
A few minor comments, otherwise looks good

examples/demo2/app.rs
joshka: changed to "...the default frame rate of a GIF in VHS."

examples/demo2/app.rs
joshka: done

examples/demo2/app.rs
joshka: done

PR #837, opened 2024-01-17
refactor(padding): move to padding.rs and update doc comments

joshka:


src/widgets/block.rs
```diff
@@ -6,18 +6,21 @@
...
-pub use self::title::{Position, Title};
 use crate::{
     prelude::*,
     symbols::border,
     widgets::{Borders, Widget},
 };
 
+mod padding;
+#[path = "../title.rs"]
+pub mod title;
```
Valentin271, line 19: What about moving `title.rs` to the right place?

Valentin271 (approved):
Apart from my comment, happy to merge.

Issue #800, opened 2024-01-12
Update table example


joshka:
## Problem
In #781 we tweaked the table example a bit to use a nicer layout and colors.
The example image has not been updated yet in the readme.
Before we do, this example needs to be a bit less frantic (mostly just slow down the key presses in the tape file for this).

  ![Made with VHS](https://vhs.charm.sh/vhs-6z75EkofPu13czwIq1ZhYw.gif)

It would be nice to add a scrollbar to the example too, and tweak the colors from the new material design or tailwind palette modules.

## Solution

- [ ] update the example
- [ ] tweak the vhs tape and generate a new image
- [ ] PR for the main repo updating the example
- [ ] PR for the /images branch with the new image
- [ ] Update the website example https://ratatui.rs/showcase/widgets/#table

Emivvvvv:
I did the first three in the PR: https://github.com/ratatui-org/ratatui/pull/840. 

However, I can't update the gif in the images branch. I get this error when I try to push my changes.
```
[cargo-make] INFO - cargo make 0.37.7
[cargo-make] INFO - Build File: Makefile.toml
[cargo-make] INFO - Task: ci
[cargo-make] INFO - Profile: development
[cargo-make] ERROR - Task ci not found
[cargo-make] WARN - Build Failed.
error: failed to push some refs to 'https://github.com/Emivvvvv/ratatui.git'
```

I also wanted to update website's example but `style::palette::material` is not available for rataui 0.25.0 but I'm using that for new table example
```
const EXAMPLE_TABLE_COLORS: [material::AccentedPalette; 4] = [material::RED, ...];
...
let normal_row_color = curr_color.c600;
let alt_row_color = curr_color.c400;
```

orhun:
My guess is you probably have some git hooks set up and `git push` is unable to run them on the `images` branch.

Emivvvvv:
`git push --no-verify` worked 👍🏻

Emivvvvv:
And for the website, I guess I need to add more things to put my code because of that color change ability.

> though if they are more advanced than the ones already in this repo (i.e. require anything beyond just rendering), you may need to add another binary to the source project.

joshka:
Apologies. The workflow for updating example images isn't great - it's optimized for maintainers over contributors because it isn't an area that has had much external contribution. We're happy to run the vhs scripts if needed to regen the images, or take PRs on that branch.

[...]

PR #840, opened 2024-01-17
feat: update table example and table.tape

Emivvvvv:
Issue: https://github.com/ratatui-org/ratatui/issues/800

This PR adds:

for table.rs
- scrollbar to the table
- colors changed to style::palette::material::AccentedPalette
- now colors can be changed with keys (l or →) for the next color, (h or ←) for the previous color
- added a footer for key info

for table.tape
- typing speed changed to 0.75s from 0.5s
- screen size changed to 1600x864 from 1200x800
- pushed keys changed to show the current example better

I'll push the new GIF to the images branch. Edit: https://github.com/ratatui-org/ratatui/pull/841

### DIFF
--- a/examples/table.rs
+++ b/examples/table.rs
@@ -6,22 +6,49 @@ use crossterm::{
     terminal::{disable_raw_mode, enable_raw_mode, EnterAlternateScreen, LeaveAlternateScreen},
 };
 use itertools::Itertools;
-use ratatui::{prelude::*, widgets::*};
+use ratatui::{prelude::*, style::palette::material::AccentedPalette, widgets::*};
 use style::palette::material;
 
-const EXAMPLE_TABLE_COLORS: [material::AccentedPalette; 4] = [
+const PALETTES: [material::AccentedPalette; 4] = [
+    material::BLUE,
     material::RED,
     material::GREEN,
-    material::BLUE,
     material::ORANGE,
 ];
 const INFO_TEXT: &str =
-    "(q) to quit / (j, ↑) to go up / (k, ↓) to go down / (l, →) for next color / (h, ←) for previous color";
+    "(Esc) quit | (↑) move up | (↓) move down | (→) next color | (←) previous color";
+
+const ITEM_HEIGHT: usize = 4;
+
+struct TableColors {
+    buffer_bg: Color,
+    header_bg: Color,
+    header_fg: Color,
+    row_fg: Color,
+    normal_row_color: Color,
+    alt_row_color: Color,
+    footer_border_color: Color,
+}
+
+impl TableColors {
+    fn new(color: AccentedPalette) -> Self {
+        Self {
+            buffer_bg: color.c900,
+            header_bg: color.c900,
+            header_fg: material::WHITE,
+            row_fg: material::GRAY.c100,
+            normal_row_color: color.c600,
+            alt_row_color: color.c400,
+            footer_border_color: color.c400,
+        }
+    }
+}
 
 struct App {
     state: TableState,
     items: Vec<Vec<String>>,
     scroll_state: ScrollbarState,
+    colors: TableColors,
     color_index: usize,
 }
 
@@ -30,7 +57,8 @@ impl App {
         let items = generate_fake_names();
         App {
             state: TableState::default().with_selected(0),
-            scroll_state: ScrollbarState::new((items.len() - 1) * 4),
+            scroll_state: ScrollbarState::new((items.len() - 1) * ITEM_HEIGHT),
+            colors: TableColors::new(PALETTES[0]),
             color_index: 0,
             items,
         }
@@ -47,7 +75,7 @@ impl App {
             None => 0,
         };
         self.state.select(Some(i));
-        self.scroll_state = self.scroll_state.position(i * 4);
+        self.scroll_state = self.scroll_state.position(i * ITEM_HEIGHT);
     }
 
     pub fn previous(&mut self) {
@@ -62,23 +90,20 @@ impl App {
             None => 0,
         };
         self.state.select(Some(i));
-        self.scroll_state = self.scroll_state.position(i * 4);
+        self.scroll_state = self.scroll_state.position(i * ITEM_HEIGHT);
     }
 
     pub fn next_color(&mut self) {
-        if self.color_index >= EXAMPLE_TABLE_COLORS.len() - 1 {
-            self.color_index = 0;
-        } else {
-            self.color_index += 1;
-        }
+        self.color_index = (self.color_index + 1) % PALETTES.len();
     }
 
     pub fn previous_color(&mut self) {
-        if self.color_index == 0 {
-            self.color_index = EXAMPLE_TABLE_COLORS.len() - 1;
-        } else {
-            self.color_index -= 1
-        }
+        let count = PALETTES.len();
+        self.color_index = (self.color_index + count - 1) % count;
+    }
+
+    pub fn set_colors(&mut self) {
+        self.colors = TableColors::new(PALETTES[self.color_index])
     }
 }
 
@@ -136,12 +161,13 @@ fn run_app<B: Backend>(terminal: &mut Terminal<B>, mut app: App) -> io::Result<(
 
         if let Event::Key(key) = event::read()? {
             if key.kind == KeyEventKind::Press {
+                use KeyCode::*;
                 match key.code {
-                    KeyCode::Char('q') => return Ok(()),
-                    KeyCode::Down | KeyCode::Char('j') => app.next(),
-                    KeyCode::Up | KeyCode::Char('k') => app.previous(),
-                    KeyCode::Right | KeyCode::Char('l') => app.next_color(),
-                    KeyCode::Left | KeyCode::Char('h') => app.previous_color(),
+                    Char('q') | Esc => return Ok(()),
+                    Char('j') | Down => app.next(),
+                    Char('k') | Up => app.previous(),
+                    Char('l') | Right => app.next_color(),
+                    Char('h') | Left => app.previous_color(),
                     _ => {}
                 }
             }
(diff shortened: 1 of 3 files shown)


examples/table.tape
```diff
@@ -1,19 +1,30 @@
 # This is a vhs script. See https://github.com/charmbracelet/vhs for more info.
 # To run this script, install vhs and run `vhs ./examples/table.tape`
 Output "target/table.gif"
-Set Theme "Aardvark Blue"
-Set Width 1200
-Set Height 800
+Set Width 1600
+Set Height 864
```
joshka, line 5: These gifs end up going in the examples readme, so I like to make them look a bit consistent. Changing the width here might make things look a bit weird (or may be fine).

Can you publish a gif with your changes using `vhs publish` and update the example readme so we can see this in context of what it looks like on GitHub?

examples/table.tape
```diff
@@ -1,19 +1,30 @@
...
+Set TypingSpeed 0.75s
 Down 2
 Up 2
-Down 8
-Up 12
-Down 4
+Down 2
+Right 1
+Down 2
+Right 1
```
joshka, line 18: These color changes make this subjectively still a bit frantic for my liking.

It's difficult to get a good balance for something that you can look at and get an understanding of how the example is made up. In most of the examples, you'll see lots of ~2 second sleeps for that reason any time the view changes (this color change counts), or at the end of movement / typing / scrolling to allow the user to process things.

The opposite side of this is going too slow so the UI looks clunky.

examples/table.rs
```diff
@@ -141,16 +183,16 @@ fn ui(f: &mut Frame, app: &mut App) {
...
     let bar = " █ ";
     let t = Table::new(
         rows,
         [
-            Constraint::Length(15),
+            Constraint::Length(25),
+            Constraint::Min(40),
             Constraint::Min(30),
-            Constraint::Min(25),
         ],
```
joshka, line 196: It's probably worth calculating the maximum item lengths at least for the name and address here, and then letting the email take up the rest.

joshka (commented):
I like where this is heading. I started reviewing and then got stuck in a bit of a loop of "what if we did this..." :D

examples/table.rs
```diff
@@ -7,17 +7,32 @@ use crossterm::{
...
+use style::palette::material;
+
+const EXAMPLE_TABLE_COLORS: [material::AccentedPalette; 4] = [
+    material::RED,
+    material::GREEN,
+    material::BLUE,
+    material::ORANGE,
+];
+const INFO_TEXT: &str =
+    "(q) to quit / (j, ↑) to go up / (k, ↓) to go down / (l, →) for next color / (h, ←) for previous color";
```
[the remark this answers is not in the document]
joshka: Oh - I should have written why on this:
l is too easy to confuse for 1 an i, so caps for shortcuts makes sense
Making this narrower is useful because our demos often tend to be rendered into around 78 chars wide.

examples/table.rs
```diff
@@ -7,17 +7,32 @@ use crossterm::{
...
+    color_index: usize,
 }
 
 impl App {
     fn new() -> App {
+        let items = generate_fake_names();
         App {
             state: TableState::default().with_selected(0),
-            items: generate_fake_names(),
+            scroll_state: ScrollbarState::new((items.len() - 1) * 4),
```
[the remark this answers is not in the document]
joshka: or `const ITEM_HEIGHT = 4;`

examples/table.rs
```diff
@@ -115,15 +150,22 @@ fn run_app<B: Backend>(terminal: &mut Terminal<B>, mut app: App) -> io::Result<(
...
+        .split(f.size());
+
+    let curr_color = EXAMPLE_TABLE_COLORS[app.color_index];
 
-    // colors from https://tailwindcss.com/docs/customizing-colors
-    let header_bg = Color::from_str("#1e3a8a").unwrap();
-    let header_fg = Color::from_str("#eff6ff").unwrap();
+    let buffer_bg = curr_color.c900;
+    let header_bg = curr_color.c900;
+    let header_fg = material::WHITE;
```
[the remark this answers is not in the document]
joshka: I don't think the 2014 material color palettes have a particularly dark colors: compare:
https://docs.rs/ratatui/0.26.0-alpha.1/ratatui/style/palette/material/index.html
image

https://docs.rs/ratatui/0.26.0-alpha.1/ratatui/style/palette/tailwind/index.html
image

examples/table.rs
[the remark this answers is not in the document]
kdheepak: I agree that we should make the just `ESC` and arrow keys, and hide `qhjkl` from messages. Terminal / vim users will naturally try it, so it is enough if the examples support it.

examples/table.tape
Emivvvvv: I'll play with table.tape to see what could be done

examples/table.rs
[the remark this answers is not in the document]
Emivvvvv: Maybe we can let users change `palette::material` to `palette::tailwind` with pressing `TAB` key? Also, they can change between fully colored bg to a darker version.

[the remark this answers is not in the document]
Emivvvvv: I agree that this dark background version is better for eyes, however, the real implementations might be much smaller, so the colored version might be preferred. So I think besides choosing one implementation for colors we can let users explore them with the `TAB` key.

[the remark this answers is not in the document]
joshka: The examples are really about showing a specific behavior. The table example should be less about showing off tailwind / material / being able to switch schemes. and More about showing off the features of the table. I'd suggest keep it muted, and keep the color changes to the end of the gif, cycliing through with at least 1 sec and perhaps 2 secs. To avoid "jumping" when the user reaches the end of the loop, you'll either need a longer pause at the end or make sure to make it back to where the loops starts.

examples/table.tape
Emivvvvv: Looks good imo. I also set the theme as `"Aardvark Blue"` to not break the consistency. You can see it in the last commit.

examples/table.rs
Emivvvvv: I wrote the calculation function and added + 1 to all of the lengths before sending them `Constraint::Length `because they were looking too close.

```
    let (longest_name_len, longest_address_len, longest_email_len) =
        constraint_len_calculator(&app.items);
    let t = Table::new(
        rows,
        [
            // + 1 is for padding.
            Constraint::Length(longest_name_len + 1),
            Constraint::Min(longest_address_len + 1),
            Constraint::Min(longest_email_len + 1),
        ],
```

```
fn constraint_len_calculator(items: &[Vec<String>]) -> (u16, u16, u16) {
    let len_vec = items
        .iter()
        .map(|row| {
            (
                row[0].chars().count() as u16,
                row[1]
                    .split('\n')
                    .map(|address_line| address_line.chars().count())
                    .max()
                    .unwrap() as u16,
                row[2].chars().count() as u16,
            )
        })
        .collect::<Vec<(u16, u16, u16)>>();

    let longest_name_len = len_vec.iter().map(|row| row.0).max().unwrap();
    let longest_address_len = len_vec.iter().map(|row| row.1).max().unwrap();
    let longest_email_len = len_vec.iter().map(|row| row.2).max().unwrap();

    (longest_name_len, longest_address_len, longest_email_len)
}
```

examples/table.tape
Emivvvvv: I think It's better now what do you think?

examples/table.rs
```diff
@@ -115,15 +150,22 @@ fn run_app<B: Backend>(terminal: &mut Terminal<B>, mut app: App) -> io::Result<(
 }
 
 fn ui(f: &mut Frame, app: &mut App) {
```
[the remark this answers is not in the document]
Emivvvvv: Separated ui to `render_table`, `render_scrollbar` and `render_footer`

examples/table.rs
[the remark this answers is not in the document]
Emivvvvv: Now it looks like this

```
const INFO_TEXT: &str =
    "(Esc) quit | (↑) move up | (↓) move down | (→) next color | (←) previous color";
```

Screenshot 2024-01-18 at 16 00 06

examples/table.rs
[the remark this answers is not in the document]
Emivvvvv: First implementation (PR):

table

`palette::material` changed to `palette::tailwind` and has darker background https://github.com/ratatui-org/ratatui/pull/840#discussion_r1456842572:

table

Colors changed to `[tailwind::BLUE, tailwind::EMERALD, tailwind::INDIGO, tailwind::RED]` https://github.com/ratatui-org/ratatui/pull/840#discussion_r1456704953:

table

I'm not sure which one should we choose because both versions look good to me.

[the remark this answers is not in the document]
Emivvvvv: Now I realize if we put Sleep 2s or 1s before going down again to the first place it will look better

examples/table.rs
Emivvvvv: I can add a test function for `fn constraint_len_calculator()` and I'm open to name suggestions :D

examples/table.rs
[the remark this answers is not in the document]
Emivvvvv: Referring to this issue: https://github.com/ratatui-org/ratatui/issues/846

 I'll push the third option

Emivvvvv:
This is how the table looks in the last commit:
table

examples/table.rs
```diff
@@ -161,6 +236,82 @@ fn ui(f: &mut Frame, app: &mut App) {
...
-    f.render_stateful_widget(t, rects[0], &mut app.state);
+    f.render_stateful_widget(t, area, &mut app.state);
+}
+
+fn constraint_len_calculator(items: &[Vec<String>]) -> (u16, u16, u16) {
+    let len_vec = items
+        .iter()
+        .map(|row| {
+            (
+                row[0].chars().count() as u16,
```
joshka, line 249: Does .width work here?

examples/table.rs
```diff
@@ -161,6 +236,82 @@ fn ui(f: &mut Frame, app: &mut App) {
...
+                row[1]
+                    .split('\n')
```
joshka, line 251: And .lines().map(UnicodeSeg::width).max() ?

examples/table.rs
```diff
@@ -161,6 +236,82 @@ fn ui(f: &mut Frame, app: &mut App) {
...
+                    .map(|address_line| address_line.chars().count())
+                    .max()
+                    .unwrap() as u16,
+                row[2].chars().count() as u16,
+            )
+        })
+        .collect::<Vec<(u16, u16, u16)>>();
+
+    let longest_name_len = len_vec.iter().map(|row| row.0).max().unwrap();
```
joshka, line 260: If you're iterating and collecting then iterating on the result 3 times or might be neater to iterate and collect the max for each column in a single iteration.

examples/table.rs
```diff
@@ -161,6 +236,82 @@ fn ui(f: &mut Frame, app: &mut App) {
...
+    (longest_name_len, longest_address_len, longest_email_len)
+}
+
+fn render_scrollbar(f: &mut Frame, app: &mut App, area: Rect) {
+    f.render_stateful_widget(
+        Scrollbar::default()
+            .orientation(ScrollbarOrientation::VerticalRight)
+            .begin_symbol(None)
+            .end_symbol(None),
+        area.inner(&Margin {
```
joshka, line 273: Maybe rethink this margin. It looks a bit weird on the bottom.

examples/table.rs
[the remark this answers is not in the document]
joshka: Either is fine

examples/table.tape
joshka: Looks good.

examples/table.rs
joshka: What about adding a small struct that has name, address, email and which gets generated in the app constructor, and then also calculating the column width values there and storing them in an app field rather than in each render?

This swaps indexing the columns implicitly for
```rust
let name_len =
  people.map(Person::name)
    .map(UnicodeWidth::width)
    .max();
let address_len
  people.map(Person::address)
    .flatmap(String::lines)
    .map(UnicodeWidth::width)
    .max();
...

```

Obviously this is just gold plating things at this point. But it does give a nice way of showing how to map real world data into table columns.

joshka (approved):
Looking good. Nothing blocking here. Just things that would improve it a little.

joshka (approved):
LGTM

examples/table.rs
joshka: Marking as resolved to merge this.

joshka:
Thanks again for making these changes.

Issue #835, opened 2024-01-17
Deprecate `SegmentSize` on `Table`


kdheepak:
Currently table uses `SegmentSize`: 

https://github.com/ratatui-org/ratatui/blob/9df6cebb58e97ac795868fa0af96a8aaf9c794c0/src/widgets/table/table.rs#L540-L569

but we would like to deprecate that in favor of `Flex`. 

- Add new `flex` method with `flex` field
- Change `segment_size` method to use `flex` method
- Deprecate `segment_size` method and remove `segment_size` field
- Update documentation 

See this PR for reference:

https://github.com/ratatui-org/ratatui/pull/804

Emivvvvv:
This also means

```
/// Controls how to distribute extra space among the columns
segment_size: SegmentSize,
```
```
fn default() -> Self {
        Self {
            ...
            segment_size: SegmentSize::None,
        }
    }
```

will also be changed to (including all other usages of segment_size and SegmentSize)

```
/// Controls how to distribute extra space among the columns
flex: Flex,
```
```
fn default() -> Self {
        Self {
            ...
            flex: Flex::Start,
        }
    }
```
right? Or I understand wrongly.

Emivvvvv:
Yeah makes sense. Thanks for explaining! 🚀

PR #842, opened 2024-01-17
feat(table): deprecate SegmentSize on table

Emivvvvv:
Issue: https://github.com/ratatui-org/ratatui/issues/835

This PR adds:

for table

- Added new flex method with flex field
- Changed segment_size method to use flex method
- Deprecated segment_size method and removed segment_size field
- Updated documentation
- modified tests to use flex instead of segment_size except for one test.
    -test `fn underconstrained_segment_size()` still uses segment_size to test if it still works.

### DIFF
--- a/src/widgets/table/table.rs
+++ b/src/widgets/table/table.rs
@@ -539,8 +539,6 @@ impl<'a> Table<'a> {
         self
     }
 
-    /// NOTE: segment_size is deprecated. Use flex method instead.
-    ///
     /// Set how extra space is distributed amongst columns.
     ///
     /// This determines how the space is distributed when the constraints are satisfied. By default,
@@ -566,7 +564,7 @@ impl<'a> Table<'a> {
         reason = "The name for this feature is not final and may change in the future",
         issue = "https://github.com/ratatui-org/ratatui/issues/536"
     )]
-    #[deprecated(note = "please use flex instead of segment_size")]
+    #[deprecated(since = "0.26.0", note = "You should use Table::flex instead.")]
     pub const fn segment_size(self, segment_size: SegmentSize) -> Self {
         let translated_to_flex = match segment_size {
             SegmentSize::None => Flex::Start,
@@ -589,13 +587,15 @@ impl<'a> Table<'a> {
     ///
     /// Create a table that needs at least 30 columns to display.  Any extra space will be assigned
     /// to the last column.
-    #[cfg_attr(feature = "unstable", doc = " ```")]
-    #[cfg_attr(not(feature = "unstable"), doc = " ```ignore")]
+    /// ```
     /// # use ratatui::layout::{Constraint, Flex};
     /// # use ratatui::widgets::{Table, Row};
-    /// let widths = [Constraint::Min(10), Constraint::Min(10), Constraint::Min(10)];
-    /// let table = Table::new(Vec::<Row>::new(), widths)
-    ///     .flex(Flex::StretchLast);
+    /// let widths = [
+    ///     Constraint::Min(10),
+    ///     Constraint::Min(10),
+    ///     Constraint::Min(10),
+    /// ];
+    /// let table = Table::new(Vec::<Row>::new(), widths).flex(Flex::StretchLast);
     /// ```
     pub const fn flex(mut self, flex: Flex) -> Self {
         self.flex = flex;
--- a/src/widgets/table/table.rs
+++ b/src/widgets/table/table.rs
@@ -2,7 +2,7 @@ use itertools::Itertools;
 
 use super::*;
 use crate::{
-    layout::SegmentSize,
+    layout::{Flex, SegmentSize},
     prelude::*,
     widgets::{Block, StatefulWidget, Widget},
 };
@@ -221,7 +221,7 @@ pub struct Table<'a> {
     highlight_spacing: HighlightSpacing,
 
     /// Controls how to distribute extra space among the columns
-    segment_size: SegmentSize,
+    flex: Flex,
 }
 
 impl<'a> Default for Table<'a> {
@@ -237,7 +237,7 @@ impl<'a> Default for Table<'a> {
             highlight_style: Default::default(),
             highlight_symbol: Default::default(),
             highlight_spacing: Default::default(),
-            segment_size: SegmentSize::None,
+            flex: Flex::Start,
         }
     }
 }
@@ -539,6 +539,8 @@ impl<'a> Table<'a> {
         self
     }
 
+    /// NOTE: segment_size is deprecated. Use flex method instead.
+    ///
     /// Set how extra space is distributed amongst columns.
     ///
     /// This determines how the space is distributed when the constraints are satisfied. By default,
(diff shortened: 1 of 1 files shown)


src/widgets/table/table.rs
```diff
@@ -564,8 +566,39 @@ impl<'a> Table<'a> {
         reason = "The name for this feature is not final and may change in the future",
         issue = "https://github.com/ratatui-org/ratatui/issues/536"
     )]
-    pub const fn segment_size(mut self, segment_size: SegmentSize) -> Self {
-        self.segment_size = segment_size;
+    #[deprecated(note = "please use flex instead of segment_size")]
```
kdheepak, line 569: We should add a `since = "v0.26.0"`.

```suggestion
#[deprecated(note = "please use Table::flex instead of Table::segment_size", since = "v0.26.0")]
```

src/widgets/table/table.rs
```diff
@@ -237,7 +237,7 @@ impl<'a> Default for Table<'a> {
             highlight_style: Default::default(),
             highlight_symbol: Default::default(),
             highlight_spacing: Default::default(),
-            segment_size: SegmentSize::None,
+            flex: Flex::Start,
```
kdheepak, line 240: I personally think the default should be `Flex::Stretch` for now. We should change it along with the default for `Layout` to be `Flex::Start` in a future PR.

src/widgets/table/table.rs
```diff
@@ -564,8 +566,39 @@ impl<'a> Table<'a> {
...
+    /// space to the last column or to distribute it equally.
+    ///
+    /// This is a fluent setter method which must be chained or used as it consumes self
+    ///
+    /// # Examples
+    ///
+    /// Create a table that needs at least 30 columns to display.  Any extra space will be assigned
+    /// to the last column.
+    #[cfg_attr(feature = "unstable", doc = " ```")]
+    #[cfg_attr(not(feature = "unstable"), doc = " ```ignore")]
```
kdheepak, line 593: I'm curious, why is this required?

kdheepak:
Thank you for the PR!

src/widgets/table/table.rs
Emivvvvv: I don't know. I copy pasted this part from the Table::segment_size's documentation.

kdheepak: Do the tests pass without it?

src/widgets/table/table.rs
Emivvvvv: But isn't `Flex::Start` the equivalent of `SegmentSize::None`? I think the opposite should be done.

src/widgets/table/table.rs
Emivvvvv: Yeah, you're right. It's definitely better

src/widgets/table/table.rs
Emivvvvv: For `Table::flex` yes, but `Table::segment_size` fails. I think we can remove it from `Table::flex`.

```
---- src/widgets/table/table.rs - widgets::table::table::Table<'a>::segment_size (line 556) stdout ----
error[E0603]: enum `SegmentSize` is private
  --> src/widgets/table/table.rs:558:35
   |
5  | use ratatui::layout::{Constraint, SegmentSize};
   |                                   ^^^^^^^^^^^ private enum
   |
note: the enum `SegmentSize` is defined here
  --> /Users/emivvvvv/Documents/GitHub/ratatui/src/layout.rs:24:16
   |
24 | pub(crate) use segment_size::SegmentSize;
   |                ^^^^^^^^^^^^^^^^^^^^^^^^^

error[E0624]: method `segment_size` is private
   --> src/widgets/table/table.rs:562:6
    |
9   |       .segment_size(SegmentSize::LastTakesRemainder);
    |        ^^^^^^^^^^^^ private method
    |
   ::: /Users/emivvvvv/Documents/GitHub/ratatui/src/widgets/table/table.rs:563:5
    |
563 | /     #[stability::unstable(
564 | |         feature = "segment-size",
565 | |         reason = "The name for this feature is not final and may change in the future",
566 | |         issue = "https://github.com/ratatui-org/ratatui/issues/536"
567 | |     )]
    | |______- private method defined here
    ```

File: src/widgets/table/table.rs (hunk 1 above)
Comment (Emivvvvv) [reply src/widgets/table/table.rs:569]: Clippy didn't like something but I don't know what exactly

```
error: the since field must contain a semver-compliant version
   --> src/widgets/table/table.rs:570:9
    |
570 |         since = "v0.26.0",
    |         ^^^^^^^^^^^^^^^^^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#deprecated_semver
    = note: `#[deny(clippy::deprecated_semver)]` on by default
```

Comment (kdheepak) [reply src/widgets/table/table.rs:569]: Sorry, remove the `v`.

File: src/widgets/table/table.rs [hunk 4]
```diff
@@ -539,6 +539,8 @@ impl<'a> Table<'a> {
         self
     }
 
+    /// NOTE: segment_size is deprecated. Use flex method instead.
+    ///
```
orhun, line 543: Don't we already have deprecated notice on the type? I don't think having this as an additional comment is not very useful. I'm not sure we have done the same thing for the other deprecated types though.

src/widgets/table/table.rs
orhun, line 569: Nit: 
```suggestion
    #[deprecated(note = "You should use flex instead.")]
```

We can also have a more detailed message here. See the messages of other deprecated types.

src/widgets/table/table.rs
Valentin271: Yes it should be removed from `Table::flex` and kept for `segment_size`. That's to indicate this is subject to change, `flex` isn't really.

src/widgets/table/table.rs
kdheepak: `Flex::Start` is the equivalent of `SegmentSize::None` but imho that's not a good default. 

Let's see what others think.

joshka: I think we need to split the task of deprecating the existing behavior and changing the defaults into two separate tasks. Combining the two is problematic. The first is a mechanical change that is either right or wrong, the second is an exercise in opinion, and relies on specifying the existing behavior and understanding the impact of changing it.

kdheepak: Fair point. I'll resolve this conversation.

src/widgets/table/table.rs
Emivvvvv: Removed those two lines 👍🏻

src/widgets/table/table.rs
Emivvvvv: Yeah, now it works! Thanks! I updated that line to `#[deprecated(since = "0.26.0", note = "You should use Table::flex instead.")]`

Valentin271 (approved):
Everything seems resolved here and the code looks good. Approved. Thanks for this!

PR #844, opened 2024-01-18
docs(examples): document incompatible examples better

joshka:
Examples often take advantage of unreleased API changes, which makes them not copy-paste friendly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,9 +74,10 @@ for more info.
 
 The following example demonstrates the minimal amount of code necessary to setup a terminal and
 render "Hello World!". The full code for this example which contains a little more detail is in
-[hello_world.rs]. For more guidance on different ways to structure your application see the
-[Application Patterns] and [Hello World tutorial] sections in the [Ratatui Website] and the various
-[Examples]. There are also several starter templates in the [templates] repository.
+the [Examples] directory. For more guidance on different ways to structure your application see
+the [Application Patterns] and [Hello World tutorial] sections in the [Ratatui Website] and the
+various [Examples]. There are also several starter templates available in the [templates]
+repository.
 
 Every application built with `ratatui` needs to implement the following steps:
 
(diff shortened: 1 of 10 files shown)


orhun (approved):
LGTM

README.md
```diff
@@ -289,8 +290,8 @@ Running this example produces the following output:
 [Handling Events]: https://ratatui.rs/concepts/event-handling/
 [Layout]: https://ratatui.rs/how-to/layout/
 [Styling Text]: https://ratatui.rs/how-to/render/style-text/
-[templates]: https://github.com/ratatui-org/templates
-[Examples]: https://github.com/ratatui-org/ratatui/tree/main/examples
+[templates]: https://github.com/ratatui-org/templates/
+[Examples]: https://github.com/ratatui-org/ratatui/tree/main/examples/README.md
```
kdheepak, line 294: An idea for discussion? We can make this line point to the latest release:

```suggestion
[Latest Release]: https://github.com/ratatui-org/ratatui/releases/stable
```

Then we add a link in the release notes:

image

kdheepak (approved):

examples/README.md
```diff
@@ -1,22 +1,24 @@
...
-> To use incompatible example code as is, you can also use the most recent alpha release of Ratatui
-> (we release these weekly).
-
-These gifs were created using [VHS](https://github.com/charmbracelet/vhs). Each example has a
-corresponding `.tape` file that holds instructions for how to generate the images. Note that the
-images themselves are stored in a separate git branch to avoid bloating the main repository.
+> - View the examples as they were when the latest version was release by selecting the tag that
+>   matches that version. E.g. <https://github.com/ratatui-org/ratatui/tree/v0.25.0/examples>. There
+>   is a combo box at the top of this page which allows you to select any previous tagged version.
+> - To view the code locally, checkout the tag using `git switch --detach v0.25.0`.
```
Valentin271, line 13: nit/question:

`git checkout <tag>` also work here and is more common.

I'd be interested to learn if there is actually any difference/reason to chose this over `checkout`?

Valentin271 (approved):

README.md
Valentin271: > Alternatively, we can link `https://github.com/ratatui-org/ratatui/tree/v0.25.0/examples` here

That was though process too in #779, it's not that hard to maintain, a quick `grep` over the previous version would show where to make changes pretty easily.

joshka: Why not put that link at the top of the examples README - before even the warning...?
I don't want the links to the examples pointing at release versions as those links go stale. I'd much prefer people see up to date examples, with the option to go to previous versions, than out of date with the option to go to updated versions.

examples/README.md
joshka: `git switch` is the newer command. It was created to have fewer edge cases to deal with and should generally be the command to recommend to most users. `checkout` does too many things, which makes it easy to do the wrong thing with the wrong combinations of parameters.

joshka: https://github.com/git/git/blob/master/Documentation/RelNotes/2.23.0.txt#L61

README.md
joshka: (Oh and yes to putting links to the examples in the release notes definitely - it's worth updating those retrospectively even)

joshka: https://github.com/ratatui-org/ratatui/pull/844/commits/1339d526001973462863c539d63b390834e1aa74

image

joshka:
Merging this - we can do the other things in addition to this if needed.

examples/README.md
Valentin271: TIL; thanks!

PR #847, opened 2024-01-18
refactor(Scrollbar): Rewrite scrollbar implementation 📦

kdheepak:
This is a WIP PR to refactor the scrollbar widget


src/widgets/scrollbar.rs
```diff
@@ -446,198 +449,177 @@ impl<'a> Scrollbar<'a> {
...
-            ScrollbarOrientation::VerticalRight => {
-                (area.top(), area.bottom(), area.right().saturating_sub(1))
-            }
-            ScrollbarOrientation::VerticalLeft => (area.top(), area.bottom(), area.left()),
-            ScrollbarOrientation::HorizontalBottom => {
-                (area.left(), area.right(), area.bottom().saturating_sub(1))
-            }
-            ScrollbarOrientation::HorizontalTop => (area.left(), area.right(), area.top()),
+            ScrollbarOrientation::VerticalRight => area.x,
+            ScrollbarOrientation::VerticalLeft => area.x + area.width.saturating_sub(1),
```
joshka, line 488: Seems flipped. Should Left be Right here?

src/widgets/scrollbar.rs
```diff
@@ -446,198 +449,177 @@ impl<'a> Scrollbar<'a> {
...
+    /// ```plain
+    ///         ┌──────────── thumb_len
+    ///       vvvvv
+    ///    ═══█████═══════
+    ///    ^^^     ^^^^^^^
+    ///     │         └────── track_end_len
+    ///     │
+    ///     └──────────────── track_start_len
+    /// ```
+    fn get_track_lens(&self, area: Rect, state: &mut ScrollbarState) -> (usize, usize, usize) {
```
joshka, line 537: Abbreviating le
```suggestion
    fn get_part_lengths(&self, area: Rect, state: &mut ScrollbarState) -> (usize, usize, usize) {
```

src/widgets/scrollbar.rs
```diff
@@ -446,198 +449,177 @@ impl<'a> Scrollbar<'a> {
...
+        let (mut track_start, mut track_end) = match self.orientation {
+            ScrollbarOrientation::VerticalRight => (area.y, (area.y + area.height)),
+            ScrollbarOrientation::VerticalLeft => (area.y, (area.y + area.height)),
+            ScrollbarOrientation::HorizontalBottom => (area.x, (area.x + area.width)),
+            ScrollbarOrientation::HorizontalTop => (area.x, (area.x + area.width)),
         };
```
joshka, line 543: Pass this value into the method (or calculate it by calling some other small method). Just deal in calculating the lengths not the direction.

src/widgets/scrollbar.rs
```diff
@@ -446,198 +449,177 @@ impl<'a> Scrollbar<'a> {
...
+        //
+        // <═══█████═══════>
+        //
+        // then increment and decrement track_start and track_end respectively
+        if let Some(s) = self.begin_symbol {
+            track_start = track_start.saturating_add(s.width() as u16);
+        };
+        if let Some(s) = self.end_symbol {
+            track_end = track_end.saturating_sub(s.width() as u16);
+        };
```
joshka, line 554: Simplify this: pass in the length to this method without the arrows already calculated.
They don't have any impact on the calculation and hence don't belong in this method.

src/widgets/scrollbar.rs
```diff
@@ -446,198 +449,177 @@ impl<'a> Scrollbar<'a> {
...
+        let position = state.position.min(state.content_length - 1) as f64;
 
-        let thumb_start = track_start + (scroll_position_ratio * track_size as f64).round() as u16;
+        // vscode style scrolling behavior
+        let scrollable_content_len = content_length + viewport_len - 1.0;
+        let thumb_start = position * track_len / scrollable_content_len;
+        let thumb_end = (position + viewport_len) * track_len / scrollable_content_len;
 
-        let thumb_end = thumb_start + thumb_size;
+        // round() as usize gives closest int, as opposed to `floor` or `ceil`
```
joshka, line 569: Make a note that this intentionally rounds the positions and not the sizes, as rounding the sizes leads to subtle off by 1 errors. e.g. 6.4 (position) +3.4 (length) = 9.8 which rounds to 10, but 6 (rounded position) + 3 (rounded length) = 9.

src/widgets/scrollbar.rs
```diff
@@ -446,198 +449,177 @@ impl<'a> Scrollbar<'a> {
...
+    /// For ScrollbarOrientation::HorizontalBottom
+    ///
+    /// ```plain
+    ///   ┌───────────────┐
+    ///   │               │
+    ///   │               │
+    ///   │               │
+    ///   └═══███═════════┘<──────── track_axis
+    /// ```
+    fn get_track_axis(&self, area: Rect) -> u16 {
```
joshka, line 485: To avoids introducing an unnecessary extra concept that exists only for the purpose of rendering (and all the associated commentary), this would be much simpler:

```rust
fn get_scollbar_area(&self, area: Rect) -> Rect {
  match self.orientation {
    ScrollbarOrientation::VerticalLeft => Rect { width: 1, ..area },
    ScrollbarOrientation::VerticalRight => Rect { x: area.right().saturating_sub(1), width: 1, ..area },
    ScrollbarOrientation::HorizontalTop => Rect { height: 1, ..area },
    ScrollbarOrientation::HorizontalBottom => Rect { y: area.bottom().saturating_sub(1), height: 1, ..area },
  }
}
```

Then this just gets used as:
```rust
for x in area.left()..area.right() {
  for y in area.top()..area..bottom() {
    let (symbol, style) = bar.next();
    buf.set_string(x, y, symbol, style);
  }
}
```
or an equivalent functional approach:
```rust
for (position, (symbol, style)) in area.rows().flatmap(Rect::columns).zip(bar) {
  buf.set_string(position.x, position.y, symbol, style);
}
```

src/widgets/scrollbar.rs
```diff
@@ -446,198 +449,177 @@ impl<'a> Scrollbar<'a> {
...
-            ScrollbarOrientation::HorizontalTop => (area.left(), area.right(), area.top()),
+            ScrollbarOrientation::VerticalRight => area.x,
+            ScrollbarOrientation::VerticalLeft => area.x + area.width.saturating_sub(1),
+            ScrollbarOrientation::HorizontalBottom => area.y + area.height.saturating_sub(1),
+            ScrollbarOrientation::HorizontalTop => area.y,
         }
     }
 
-    /// Calculate the starting and ending position of a scrollbar thumb.
+    /// Returns length segment information about scrollbar track
```
joshka, line 494: Simplify this doc a bunch by naming these as either "parts" or "sections" of a scrollbar.
```
/// Returns the lengths of the parts of a scrollbar
///
/// The scrollbar has 3 parts of note:
/// - the beginning track part
/// - the thumb part
/// - the end track part
///
/// ```plain
///         ┌──────────── thumb_len
///       vvvvv
///    ═══█████═══════
///    ^^^     ^^^^^^^
///     │         └────── track_end_len
///     │
///     └──────────────── track_start_len
/// ```
///
/// This method the length of each part returns these as a tuple:
/// `(track_start_len, thumb_len, track_end_len)`.
```

src/widgets/scrollbar.rs
```diff
@@ -446,198 +449,177 @@ impl<'a> Scrollbar<'a> {
...
+
+        let begin = (self.begin_symbol, self.begin_style);
+        let track = (self.track_symbol, self.track_style);
+        let thumb = (Some(self.thumb_symbol), self.thumb_style);
+        let end = (self.end_symbol, self.end_style);
+        let bar = iter::once(begin)
+            .chain(iter::repeat(track).take(track_start_len))
+            .chain(iter::repeat(thumb).take(thumb_len))
+            .chain(iter::repeat(track).take(track_end_len))
+            .chain(iter::once(end));
```
joshka, line 599: Making begin and end `Option<(symbol, style)>` instead of making these (`Option<symbol>, style)` allows this to be 
```suggestion
        let bar = begin.into_iter()
            .chain(iter::repeat(track).take(track_start_len))
            .chain(iter::repeat(thumb).take(thumb_len))
            .chain(iter::repeat(track).take(track_end_len))
            .chain(end.into_iter());
```
This allows you to just iterate over the area, without having to keep an iterator variable manually (i) as the empty arrows are omitted from the iterator if they are `None`.

src/widgets/scrollbar.rs
kdheepak: Oops my bad. Fixed in recent commit with tests to check for this.

src/widgets/scrollbar.rs
```diff
@@ -446,198 +450,203 @@ impl<'a> Scrollbar<'a> {
...
+        //
+        // then increment and decrement track_start and track_end respectively
+        if let Some(s) = self.begin_symbol {
+            track_start = track_start.saturating_add(s.width() as u16);
+        };
+        if let Some(s) = self.end_symbol {
+            track_end = track_end.saturating_sub(s.width() as u16);
+        };
+        track_end.saturating_sub(track_start)
     }
```
joshka, line 507:
```suggestion
    /// Calculates length of the track excluding the arrow heads
    /// ```plain
    ///        ┌────────── track_length
    ///  vvvvvvvvvvvvvvv
    /// <═══█████═══════>
    /// ```
    ///
    /// # Returns
    ///
    /// `(track_length)`
    fn track_length(&self, area: Rect) -> u16 {
        let start_len = self.begin_symbol.map(|s| s.width() as u16).unwrap_or(0);
        let end_len = self.end_symbol.map(|s| s.width() as u16).unwrap_or(0);
        let arrows_len = start_len + end_len;
        if self.orientation.is_vertical() {
            area.height.saturating_sub(arrows_len)
        } else {
            area.width.saturating_sub(arrows_len)
        }
    }
```

src/widgets/scrollbar.rs
joshka, line 1: Replace is_vertical with

```rust
impl ScrollbarOrientation {
    fn is_vertical(&self) -> bool {
        self == &Self::VerticalLeft || self == &Self::VerticalRight
    }
}
```

joshka (commented):
Simplications:
https://github.com/ratatui-org/ratatui/compare/1648e877aa80f29b8773143ec1fa8ccc26765549...joshka:ratatui:refactor-scrollbar

src/widgets/scrollbar.rs
```diff
@@ -446,198 +450,203 @@ impl<'a> Scrollbar<'a> {
...
+        let end = self.end_symbol.map(|s| (s, self.end_style));
+
+        iter::once(begin)
+            // Current state of the iterator:
+            //
+            // ```plain
+            // <________________
+            // ^
+            // └─────────────────── begin
+            // ```
```
joshka, line 582: These are too much noise and detract from the code

src/widgets/scrollbar.rs
```diff
@@ -446,198 +450,203 @@ impl<'a> Scrollbar<'a> {
...
-            } else {
-                // For horizontal scrollbar, reduce the width by one
-                Rect::new(
-                    area.x + 1,
-                    area.y,
-                    area.width.saturating_sub(1),
-                    area.height,
-                )
-            }
+    fn get_viewport_len(&self, area: Rect) -> u16 {
```
joshka, line 453: https://rust-lang.github.io/api-guidelines/naming.html?highlight=get_#getter-names-follow-rust-convention-c-getter

kdheepak (commented):
LGTM but I can't approve because I'm the PR author

src/widgets/scrollbar.rs
```diff
@@ -438,206 +358,227 @@ impl<'a> Scrollbar<'a> {
...
+
+    /// Returns the lengths of the parts of a scrollbar
+    ///
+    /// The scrollbar has 3 parts of note:
+    /// - `<═══█████═══════>`: full scrollbar
+    /// - ` ═══             `: track start part
+    /// - `    █████        `: thumb part part
+    /// - `         ═══════ `: track end part
+    ///
+    /// This method returns the length of the start, thunb, and end as a tuple.
```
kdheepak, line 498:
```suggestion
    /// This method returns the length of the start, thumb, and end as a tuple.
```

src/widgets/scrollbar.rs
```diff
@@ -438,206 +358,227 @@ impl<'a> Scrollbar<'a> {
...
+        } else {
+            area.width.saturating_sub(arrows_len)
+        }
+    }
+
+    fn viewport_length(&self, area: Rect) -> u16 {
+        if self.orientation.is_vertical() {
+            area.height
+        } else {
+            area.width
```
kdheepak, line 555: This should probably use `viewport_content_length`, right?

src/widgets/scrollbar.rs
joshka: fixed

src/widgets/scrollbar.rs
joshka: I'm not sure that I'm clear enough by what the original intent of the viewport content length was enough to make the change.
The example on ScrollbarState confuses me as it leaves several ambiguous interpretations:
- Is 2 the number of items that are visible in the viewport, or the size of each item?
- what happens if the multi-line items are not a fixed size?

joshka (approved):

src/widgets/scrollbar.rs
kdheepak: I think the original intent was "viewport size" but it was poorly named and documented by me to mean number of items in the viewport. 

For example, a viewport of size 2 where each item is 1 line in height should behave the same as a "2 items in the viewport". 

I didn't map those concepts on to each other though. The implementation under the hood does exactly that though.

joshka: ok - I'm going to add a viewport_size method, make the underlying Option<usize> instead of 0 and use it

kdheepak: That works!

joshka: actually ended up not doing this. Option breaks serialization, and we need to discuss this further.

Issue #834, opened 2024-01-17
Add wide and tall border set for half blocks


kdheepak:
Also worth considering for an add is 
```
 ▄▄▄▄
 █xx█
 █xx█
 ▀▀▀▀
 ```
Half blocks for top/bottom which renders nicely due to the aspect ratio.
<img width="53" alt="image" src="https://github.com/ratatui-org/ratatui/assets/381361/45c93528-c762-4fa6-9f0f-7b8db7fc325f">

And the tall variant:
```
 █▀▀█
 █xx█
 █xx█
 █▄▄█
 ```
<img width="59" alt="image" src="https://github.com/ratatui-org/ratatui/assets/381361/00b475d5-d0f9-4f46-bbe5-e5619f194da2">

_Originally posted by @joshka in https://github.com/ratatui-org/ratatui/issues/831#issuecomment-1894526134_

PR #848, opened 2024-01-18
feat(symbols): add wide and tall proportional border set

Emivvvvv:
Adds `PROPORTIONAL_WIDE` and `PROPORTIONAL_TALL` border sets.

`symbols::border::PROPORTIONAL_WIDE`
```
▄▄▄▄
█xx█
█xx█
▀▀▀▀
```

`symbols::border::PROPORTIONAL_TALL`
```
█▀▀█
█xx█
█xx█
█▄▄█
```

Fixes: https://github.com/ratatui-org/ratatui/issues/834

### DIFF
--- a/src/symbols.rs
+++ b/src/symbols.rs
@@ -425,8 +425,9 @@ pub mod border {
         horizontal_bottom: ONE_EIGHTH_BOTTOM_EIGHT,
     };
 
-    /// Wide proportional border with using set of quadrants.
-    /// The border is created using half blocks for top and bottom, and full
+    /// Wide proportional (visually equal width and height) border with using set of quadrants.
+    ///
+    /// The border is created by using half blocks for top and bottom, and full
     /// blocks for right and left sides to make horizontal and vertical borders seem equal.
     ///
     /// ```text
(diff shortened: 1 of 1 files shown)


joshka (approved):
It's convention to leave a blank line between the title of the comment and the body.

It would be nice to define proportional here as visually equal width and height. 

I think the name is right.

Thanks for this.

Emivvvvv:
How about now? And you're welcome but I think I'm the one who should thank because you and the other maintainers are providing the best environment for open-source beginners like me :)

joshka:
LGTM. Approving. Will wait on another maintainer to confirm the name makes sense to them as well.

joshka (approved):

joshka:
Reformatted first comment for neater changelog.

kdheepak:
I think the name is good, it has the same meaning to the naming in `Padding`. There's `Contraints::Proportional` that we might want to consider renaming to minimize confusion.

kdheepak (approved):

orhun (approved):

PR #849, opened 2024-01-19
feat: Adds message to `assert_buffer_eq!` ✨

kdheepak:
Makes it easier for debugging.

### DIFF
--- a/src/buffer/assert.rs
+++ b/src/buffer/assert.rs
@@ -4,6 +4,9 @@
 #[macro_export]
 macro_rules! assert_buffer_eq {
     ($actual_expr:expr, $expected_expr:expr) => {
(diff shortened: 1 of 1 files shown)


src/buffer/assert.rs
```diff
@@ -46,7 +49,7 @@ macro_rules! assert_buffer_eq {
                 }
                 // shouldn't get here, but this guards against future behavior
                 // that changes equality but not area or content
-                assert_eq!(actual, expected, "buffers not equal");
+                assert_eq!(actual, expected, $message);
```
joshka, line 52: This is just the fall through case (in case we didn't already fail for not equal buffers). The main failures will occur on lines 10 and 35.

EdJoPaTo:
I close this in favor of #1007

Issue #820, opened 2024-01-14
Increase Layout cache size


joshka:
When we reimplemented the cache to not grow infinitely we chose an arbitrarily small size of 16. This is too small.

https://github.com/ratatui-org/ratatui/blob/11e4f6a0ba71b7adad44af5866a2b0789175aafa/src/layout/layout.rs#L104

The cache helps apps have stable user interfaces when the layout is provided with constraints that could be satisfied multiple ways.

We should bump this up about an order of magnitude (and measure the memory impact of it - it's likely small)

Valentin271:
For reference, on my machine, the flex example starts at about 1.4MB, a cache of 9800 fully grown makes it go to 3.5MB.

That's ~2.3 times the initial memory, but this is a ridiculous value and in general not that much memory.

Concerning by how much we should increase the value I'd go for `32`ish. `flex` and  `constraints` run about fine with 100 and there are ridiculous/overcomplicated layouts.

I want to go more in depth on this and check out some projects to see how much layout to they use.

joshka:
It would be good to add some sort of metric around the cache size and hits. Possibly just logged using the tracing library, but perhaps worth investigating https://docs.rs/metrics/0.22.0/metrics/

joshka:
> 9800

that's also about how many cells I have on my terminal screen. I suspect a good ballpark figure to use is Rows + Columns, not Rows * Columns. This allows for a layout to be specified for every row and column of the screen (which is likely rare, but may become more common if we start moving to using layout for e.g. each row of a table or similar.

So that's 171+51 = 222 for my laptop so why not make this arbitrarily double that amount and a bit more (500)?

PR #850, opened 2024-01-19
feat(layout): increase default cache size to 500

joshka:
This is a somewhat arbitrary size for the layout cache based on adding
the columns and rows on my laptop's terminal (171+51 = 222) and doubling
it for good measure and then adding a bit more to make it a round
number. This gives enough entries to store a layout for every row and
every column, twice over, which should be enough for most apps. For
those that need more, the cache size can be set with
`Layout::init_cache()`.

Fixes: https://github.com/ratatui-org/ratatui/issues/820

### DIFF
--- a/src/layout/layout.rs
+++ b/src/layout/layout.rs
@@ -103,7 +103,13 @@ struct Element {
 }
 
 impl Layout {
-    pub const DEFAULT_CACHE_SIZE: usize = 16;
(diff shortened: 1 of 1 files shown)


kdheepak (approved):

Valentin271 (approved):
Seems reasonable to me too.

PR #851, opened 2024-01-19
test(Scrollbar): Backport new scrollbar implementation's tests

kdheepak:
This PR backports the tests from https://github.com/ratatui-org/ratatui/pull/847
The backported tests are made to pass with the current (i.e. old) implementation, so that a reference can be established for the changes in the new implementation.

### DIFF
--- a/src/widgets/scrollbar.rs
+++ b/src/widgets/scrollbar.rs
@@ -631,7 +631,10 @@ impl<'a> StatefulWidget for Scrollbar<'a> {
 
 #[cfg(test)]
 mod tests {
+    use itertools::Itertools;
+    use rstest::rstest;
     use strum::ParseError;
+    use unicode_width::UnicodeWidthStr;
 
     use super::*;
     use crate::{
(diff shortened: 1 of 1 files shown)


src/widgets/scrollbar.rs
```diff
@@ -1128,4 +1131,341 @@ mod tests {
             assert_buffer_eq!(buffer, Buffer::with_lines(expected.clone()));
         }
     }
+
+    #[rstest]
+    #[case("██", 0, 2, "position_0")]
+    #[case("██", 1, 2, "position_1")]
```
Valentin271, line 1137: Instead of having generic cases and pass in an `assertion_message` (which I guess is only to debug which case failed), cases could be [named](https://docs.rs/rstest/latest/rstest/attr.rstest.html#optional-case-description).

EDIT: That's also a better description when the test succeeds.

src/widgets/scrollbar.rs
```diff
@@ -1128,4 +1131,341 @@ mod tests {
...
+    #[case("##########", 0, 10, "position_0")]
+    #[case("##########", 1, 10, "position_1")]
+    #[case("##########", 2, 10, "position_2")]
+    #[case("##########", 3, 10, "position_3")]
+    #[case("##########", 4, 10, "position_4")]
+    #[case("##########", 5, 10, "position_5")]
+    #[case("##########", 6, 10, "position_6")]
+    #[case("##########", 7, 10, "position_7")]
+    #[case("##########", 8, 10, "position_8")]
+    #[case("##########", 9, 10, "position_9")]
```
Valentin271, line 1172: Maybe a comment here (and for the above test) to indicate why the scrollbar is completely filled. It confused me at first and it took me some time to figure it out.

src/widgets/scrollbar.rs
```diff
@@ -1128,4 +1131,341 @@ mod tests {
...
+            buffer,
+            Buffer::with_lines(vec![expected]),
+            "{}",
+            assertion_message,
+        );
+    }
+
+    #[rstest]
+    #[case("          ", 0, 0, "position_0")]
+    fn render_scrollbar_nobar(
```
Valentin271, line 1201: ~~Is there any point to this test? Testing that is renders empty instead of the block border?~~

Nit but it confused me, really this tests an `empty` bar and not an invisible bar (with no track or thumb).

src/widgets/scrollbar.rs
```diff
@@ -1128,4 +1131,341 @@ mod tests {
...
+    #[case("██████████", 2, 10, "position_2")]
+    #[case("██████████", 3, 10, "position_3")]
+    #[case("██████████", 4, 10, "position_4")]
+    #[case("██████████", 5, 10, "position_5")]
+    #[case("██████████", 6, 10, "position_6")]
+    #[case("██████████", 7, 10, "position_7")]
+    #[case("██████████", 8, 10, "position_8")]
+    #[case("██████████", 9, 10, "position_9")]
+    #[case("██████████", 100, 10, "position_out_of_bounds")]
+    fn render_scrollbar_twoline_horizontal(
```
Valentin271, line 1372: So here you're testing when there is a top and bottom bar right? Have you noticed weird behavior related to that?

Valentin271 (commented):
Using named cases feels necessary to me. Other comments are just little improvements.

src/widgets/scrollbar.rs
```diff
@@ -1128,4 +1131,341 @@ mod tests {
...
+        let mut buffer = Buffer::empty(Rect::new(0, 0, size as u16, 1));
+        let mut state = ScrollbarState::default()
+            .position(position)
+            .content_length(content_length);
+        Scrollbar::default()
+            .orientation(ScrollbarOrientation::HorizontalTop)
+            .begin_symbol(None)
+            .end_symbol(None)
+            .track_symbol(Some("-"))
+            .thumb_symbol("#")
```
joshka, line 1189: extract the repetition to an rstest fixture:

```rust
#[fixture]
fn scrollbar_without_arrows() -> Scrollbar { ... }

fn render_scrollbar_simple(
        #[case] expected: &str,
        #[case] position: usize,
        #[case] content_length: usize,
        #[case] assertion_message: &str,
        scrollbar_without_arrows: Scrollbar,
    ) { ... }
```

The main benefit of this is that it gives the test data a consistent name, which represents that item.
Generally I'm not a fan of DRY in tests, but for exact equivalent test objects, I am definitely for it.

src/widgets/scrollbar.rs
kdheepak: If we use named tests this makes it uneven, e.g.:

image

joshka: yeah, the alignment is good to have for very visual style tests like this.

src/widgets/scrollbar.rs
kdheepak: For context this is what the old implementation does, so it is merely a "specification test". 

This entire PR is just to show how the new tests behave in the old implementation. I'm planning to revert the commit after this PR is merged to use the new tests.

Here's what the new tests do here:

image

src/widgets/scrollbar.rs
kdheepak: What do you think about merging this and then making improvements in the new PR? It'd be easier for me to not have to jump back and forth between branches.

src/widgets/scrollbar.rs
Valentin271: Hm I see ... too bad. Your call what you want to do.

src/widgets/scrollbar.rs
joshka: So is this perhaps due to the fact that content length and viewport are about the same size? If so, need to add more tests where that isn't the case.

src/widgets/scrollbar.rs
joshka: +1 comment / more tests around this.

src/widgets/scrollbar.rs
joshka: +1 this is confusing

src/widgets/scrollbar.rs
kdheepak: This old implementation showed full scrollbars whenever `content_length` was less than or equal to viewport area:

image

src/widgets/scrollbar.rs
```diff
@@ -705,29 +708,34 @@ mod tests {
...
+    #[rstest]
+    #[case("        ", 0)]
+    #[case("████████", 1)]
+    #[case("████████", 2)]
+    #[case("████████", 3)]
+    #[case("████████", 4)]
+    #[case("████████", 5)]
+    #[case("████████", 6)]
+    #[case("████████", 7)]
+    #[case("████████", 8)]
```
joshka, line 720: It's not obvious why this is rendered like this, so it's worth adding a longer comment for this test that documents what's going on here.

Additionally, it would more intuitive to read these tests as input (content_length) -> output (expected) rather than the reverse.

I don't quite understand the meaning of the name of this test. " test_renders_empty_with_content_length_is_zero"

Edit:
So, this is a test of the vertical rendering. It might be useful to use a vertical looking element for the bars, and have this represented as an array e.g. `[" #", " #", ...]`.
That said, the tests all have the same data except the first one. So it's could be better to split this into two tests rather than going for the visual approach.

src/widgets/scrollbar.rs
```diff
@@ -1128,4 +1136,341 @@ mod tests {
             assert_buffer_eq!(buffer, Buffer::with_lines(expected.clone()));
         }
     }
+
+    #[rstest]
+    #[case("██", 0, 2, "position_0")]
+    #[case("██", 1, 2, "position_1")]
+    fn render_scrollbar_simplest(
```
joshka, line 1143: What does simplest mean here?
What concept does this test communicate?

src/widgets/scrollbar.rs
```diff
@@ -1128,4 +1136,341 @@ mod tests {
...
+    #[case("##########", 1, 10, "position_1")]
+    #[case("##########", 2, 10, "position_2")]
+    #[case("##########", 3, 10, "position_3")]
+    #[case("##########", 4, 10, "position_4")]
+    #[case("##########", 5, 10, "position_5")]
+    #[case("##########", 6, 10, "position_6")]
+    #[case("##########", 7, 10, "position_7")]
+    #[case("##########", 8, 10, "position_8")]
+    #[case("##########", 9, 10, "position_9")]
+    fn render_scrollbar_simple(
```
joshka, line 1178: This name communicates very little about why this test is how it is.

joshka (changes requested):
Generally, these tests need to be in a state where even if they are characterization tests, the character of the existing implementation is explained somehow.

This is sort of a [Chesterton's fence](https://fs.blog/chestertons-fence/) issue. i know you want to change this for the better, but I don't know what it currently does and why. The changes you want to make are a complete rewrite, so having tests that make it easy to see what's changed allows the reviewer of the rewrite to understand the changes better.

kdheepak:
Closing in favor of original PR.

PR #852, opened 2024-01-19
docs(terminal): document buffer diffing better

Valentin271:


Valentin271:
Fixed formatting, and renamed as suggested. Thanks for the review!

Issue #846, opened 2024-01-18
Make the examples look pretty


joshka:
## Problem
In the new flex and constraints examples, we used the tailwind palettes to make this look pretty.
It would be nice to have a consistent look for our examples in a similar way.

<img width="976" alt="image" src="https://github.com/ratatui-org/ratatui/assets/381361/4007f4cb-ef79-467f-9401-ba45ae1d22d7">

## Solution
Some guidelines for what looks nice (subjectively)
- tailwind seems to have a better set of colors that work well in terminal apps than does the material palette.
- use dark background, lighter text (e.g. the SLATE colors from tailwind)
- not too much text on colored background. Save this for headers, footers, and things which need to be highlighted in an app
- pay attention to contrast

## Example Guidelines
- App should be a struct, with a state field that is an AppState enum
- AppState has Running, Quitting (and any other states that the example needs
- If there is animation that should start once the application is fully running, start it by pressing a key (Enter / whatever makes sense), and set the `app.state` to e.g. `Started`
- Arrange the code so that it reads in a natural top-down outside-in order.
  - Structs and Enums first
  - impl main
  - impl App
  - impl Widget for App (or for `&App` / `&mut App` as appropriate
  - impl App again (with just the functions for rendering)
  - functions for install_error_hooks, init_terminal, restore_terminal (these are common to all the examples, so just copy them from one of the recently changed examples like https://github.com/ratatui-org/ratatui/blob/main/examples/constraints.rs)
- No methods longer than a page, and prefer less than ~20 lines
  
Main should looks something like:
```rust
fn main() -> Result<(), Box<dyn Error>> {
    init_error_hooks()?;
    let terminal = init_terminal()?;
    App::default().run(terminal)?;
    restore_terminal()?;
    Ok(())
}
```
App::run should look something like:
```rust
fn run(&mut self, mut terminal: Terminal<impl Backend>) -> io::Result<()> {
    while !self.is_quitting() {
        self.draw(&mut terminal)?;
        self.handle_events()?;
    }
    Ok(())
}
```

Issue #819, opened 2024-01-14
Improve Tabs examples


joshka:
## Problem
In the new Constraints example, a really neat approach to implementing tabs using enums was used. I like it because it avoids using strings, so it's easy to follow how the pieces work together.

This would make a really good example for the API docs, and could be expanded into a howto article on the website easily.

https://github.com/ratatui-org/ratatui/blob/cc6737b8bc09d254413adc1cbf2bc62d2f93792d/examples/constraints.rs#L114-L131

## Solution
Write this up for:
- [ ] api docs on the `Tabs` widget
- [ ] more detailed how-to article on the website

Emivvvvv:
We can also rewrite `Tabs` example. I think I can work on this task in this order.

- [ ] API docs on the Tabs widget
- [ ] Rewrite `Tabs` example following this Issue: https://github.com/ratatui-org/ratatui/issues/846
- [ ] More detailed how-to article on the website

joshka:
Go for it. Sounds great.

The caveat is I'd focus on making the Tabs usage clear and clean be the primary goal. Make it Pretty is secondary to that.

It wouldn't hurt to pick something interesting for the tab contents, but only if it's also fairly short.

On the pretty front, I've been trying to get away from using borders where they're unnecessary. Think about how web design uses borders rarely. Look at demo2, flex, and constraints examples for some ideas. 

Last think about where you want to split the example vs how to guides vs docs. There's also a missing showcase item for the website that would be nice to see https://ratatui.rs/showcase/widgets/

The best bit about this task is tabs is the most in need of a rewrite, so as long as it's clean anything is better than what we have. (There's some observations about code quality at https://discord.com/channels/1070692720437383208/1070692720437383211/1197915965107753020 and reading the conversation after that point.)

Emivvvvv:
I wanted to rewrite the `Tabs` example first before touching the API documents to see if I understand the general idea or not.

Emivvvvv:
About API docs update:

Is something like this enough?

```
/// Also with using enums, more organized code could be achieved. In this example, we used enums to
/// generate titles but they can also be used to decide what to render with a `match` expression.
///
/// ```
/// use strum::{Display, EnumIter, IntoEnumIterator};
/// use ratatui::widgets::Tabs;
///
/// #[derive(Clone, Copy, Display, EnumIter)]
/// enum SelectedTab {
///     Tab0,
///     Tab1,
/// }
///
[...]

PR #855, opened 2024-01-20
docs(examples): update tabs example and tabs.tape

Emivvvvv:
Issue: https://github.com/ratatui-org/ratatui/issues/846, https://github.com/ratatui-org/ratatui/issues/819

This PR adds:

for tabs.rs

- general refactoring on code
- subjectively better looking front
- add tailwind colors

for tabs.tape

Here is the new output:
tabs

### DIFF
--- a/examples/tabs.rs
+++ b/examples/tabs.rs
@@ -8,16 +8,14 @@ use crossterm::{
 use ratatui::{prelude::*, style::palette::tailwind, widgets::*};
 use strum::{Display, EnumIter, FromRepr, IntoEnumIterator};
 
-const TABS_COUNT: usize = 4;
-
-const PALETTES: [tailwind::Palette; TABS_COUNT] = [
+const PALETTES: &[tailwind::Palette] = &[
     tailwind::BLUE,
     tailwind::EMERALD,
     tailwind::INDIGO,
     tailwind::RED,
 ];
 
-const BORDER_TYPES: [BorderType; TABS_COUNT] = [
+const BORDER_TYPES: &[BorderType] = &[
     BorderType::Rounded,
     BorderType::Plain,
     BorderType::Double,
@@ -47,6 +45,8 @@ impl SelectedTab {
         let next_index = current_index.saturating_add(1);
         Self::from_repr(next_index).unwrap_or(*self)
     }
+
+    /// Return enum name as a styled `Line` with two spaces both left and right.
     fn tab_title(value: SelectedTab) -> Line<'static> {
         let text = format!("  {value}  ");
         let palette = &PALETTES[value as usize];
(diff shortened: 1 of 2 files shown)


examples/tabs.rs
```diff
@@ -5,31 +5,72 @@ use crossterm::{
...
+        Self::from_repr(previous_index).unwrap_or(*self)
+    }
+
+    /// Get the next tab, if there is no next tab return the current tab.
+    fn next(&self) -> Self {
+        let current_index = *self as usize;
+        let next_index = current_index.saturating_add(1);
+        Self::from_repr(next_index).unwrap_or(*self)
+    }
+    fn tab_title(value: SelectedTab) -> Line<'static> {
```
orhun, line 50: Let's also document this function.

examples/tabs.rs
```diff
@@ -5,31 +5,72 @@ use crossterm::{
...
-use ratatui::{prelude::*, widgets::*};
+use ratatui::{prelude::*, style::palette::tailwind, widgets::*};
+use strum::{Display, EnumIter, FromRepr, IntoEnumIterator};
 
-struct App<'a> {
-    pub titles: Vec<&'a str>,
-    pub index: usize,
+const TABS_COUNT: usize = 4;
+
+const PALETTES: [tailwind::Palette; TABS_COUNT] = [
```
orhun, line 13: Can be simplified as:

```suggestion
const PALETTES: &[tailwind::Palette] = &[
```

orhun (changes requested):
Nice, had some comments.

examples/tabs.rs
Emivvvvv: How about now?

orhun:
There might be ways to further simplify this - let's wait for comments from other devs.

examples/tabs.rs
```diff
@@ -5,31 +5,72 @@ use crossterm::{
...
+        let next_index = current_index.saturating_add(1);
+        Self::from_repr(next_index).unwrap_or(*self)
+    }
+
+    /// Return enum name as a styled `Line` with two spaces both left and right.
+    fn tab_title(value: SelectedTab) -> Line<'static> {
+        let text = format!("  {value}  ");
+        let palette = &PALETTES[value as usize];
+        text.fg(tailwind::SLATE.c200).bg(palette.c900).into()
+    }
```
Valentin271, line 54: This would make sense as `impl From<SelectedTab> for Line`.

examples/tabs.rs
```diff
@@ -80,30 +121,44 @@ fn run_app<B: Backend>(terminal: &mut Terminal<B>, mut app: App) -> io::Result<(
...
+fn render_tabs(f: &mut Frame, app: &App, area: Rect) {
+    let block = Block::new()
+        .title("Tabs Example".bold())
+        .title("Use h l or ◄ ► to change tab")
+        .title_alignment(Alignment::Center)
+        .padding(Padding::top(1)); // padding to separate tabs from block title.
+
+    let selected_tab_index = app.selected_tab as usize;
+    let titles = SelectedTab::iter().map(SelectedTab::tab_title);
+    let tabs = Tabs::new(titles)
```
Valentin271, line 141: With the earlier impl you could very probably just give `Tabs::new(SelectedTab::iter())` or something similar.

Valentin271 (changes requested):

examples/tabs.rs
Emivvvvv: Thanks!

orhun (approved):

Valentin271 (approved):
Thanks!

examples/tabs.rs
joshka: Just a quick rationale for why I moved away from that implementation in the flex / constraints example. Naming it with "to_title" made the method describe the intended usage of the Line we're returning (Line being an implementation detail, but title being the core concept).

Issue #853, opened 2024-01-20
Add Alignment convenience methods to Line, Text and Paragraph (and perhaps Span).


joshka:
## Problem
Creating a centered or right aligned line / text / paragraph is currently a little verbose:

```rust
Line::raw("foo").alignment(Alignment::Center);
Text::raw("foo").alignment(Alignment::Center);
Paragraph::new("foo").alignment(Alignment::Center);
```

Aligning a span means converting to a line first:
```rust
Line::from(Span::raw("foo")).alignment(Alignment::Center);
```

## Solution
Add several setter methods to Line, Text and Paragraph for each alignment:
- `fn left_aligned(self) -> Self`
- `fn centered(self) -> Self`
- `fn right_aligned(self) -> Self`

```rust
Line::raw("foo").left_aligned();
Text::raw("foo").centered();
Paragraph::new("foo").right_aligned();
```

Add the same methods on span which will convert the span into a Line:
- `fn left_aligned(self) -> Line<'a>`
- `fn centered(self) -> Line<'a>`
- `fn right_aligned(self) -> Line<'a>`

## Implementation

- [ ] simple unit tests for each method (e.g. `let line = Line::default().centered(); assert_eq!(line.alignment, Alignment::Center)`
- [ ] docs for each mentioning this as a convenience shortcut
- [ ] docs for methods on each struct
- [ ] replace example code that calls `.alignment(Alignment::*)` with new calls
- [ ] replace code in website examples

This should be a non-breaking change that could be released in a point release rather than a major release.

## Alternatives
This could be implemented as an extension trait like Stylize, but I think it would be easier to understand the implementation if these methods were just repeated.

## Additional context
Marked as a Good First Issue for newcomers. Please read the [Contribution Guidelines](https://github.com/ratatui-org/ratatui/blob/main/CONTRIBUTING.md) for some help getting started, and don't hesitate to reach out for more questions on this here or on the the [feature discussion](https://discord.com/channels/1070692720437383208/1072879985762762812) discord channel.

Eeelco:
I would like to work on this issue, could you assign it to me please?

PR #856, opened 2024-01-20
feat(Line): add alignment convenience functions

Eeelco:
Added convenience functions left_aligned(), centered() and right_aligned() plus unit tests. Updated example code.

### DIFF
--- a/src/text/line.rs
+++ b/src/text/line.rs
@@ -198,6 +198,8 @@ impl<'a> Line<'a> {
     /// Sets the target alignment for this line of text.
     ///
     /// Defaults to: [`None`], meaning the alignment is determined by the rendering widget.
+    /// Setting the alignment of a Line generally overrides the alignment of its
+    /// parent Text or Widget.
     ///
     /// # Examples
     ///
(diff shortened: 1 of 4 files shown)


Eeelco:
Decided to split the fixes for #853 into multiple PRs to keep the PR small

src/text/line.rs
```diff
@@ -218,6 +218,57 @@ impl<'a> Line<'a> {
...
+    ///
+    /// Convenience shortcut for `Line::alignment(Alignment::Left)`.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// # use ratatui::prelude::*;
+    /// let mut line = Line::from("Hi, what's up?");
+    /// assert_eq!(None, line.alignment);
+    /// assert_eq!(Some(Alignment::Left), line.left_aligned().alignment);
```
joshka, line 231: This variable doesn't need to be mutable.

I generally prefer to avoid testing preconditions in tests like this (but I see that came from the alignment impl). Here I think we could reasonably get away with avoiding any assertions without too much loss of obviousness - what do you think?

```suggestion
    /// # use ratatui::prelude::*;
    /// let line = Line::from("Hi, what's up?").left_aligned();
```

src/text/line.rs
```diff
@@ -218,6 +218,57 @@ impl<'a> Line<'a> {
         }
     }
 
+    /// Left-aligns this line of text.
+    ///
+    /// Convenience shortcut for `Line::alignment(Alignment::Left)`.
```
joshka, line 223: Can you please add something like the following: "Setting the alignment of a Line generally overrides any alignment set on the Text or Widget that contains the Line." and also to the alignment setter. And perhaps something similar to noting the default? There's probably a way to communicate this more concisely in a single sentence too.

I say generally, because third party custom widgets can render however they want, but our widgets will try to adhere to this when possible.

joshka (approved):
Comments are all nitpicks that will improve things very slightly  - This can be merged with or without the suggested changes.

src/text/line.rs
Eeelco: How about this?

> /// Setting the alignment of a Line generally overrides the alignment of its 
    /// parent `Text` or `Widget`, with the default alignment being inherited from the parent.

src/text/line.rs
Eeelco: Sounds good, will change it also for the other two functions!

src/text/line.rs
Valentin271: > How about this?

Yeah that works.

Valentin271 (approved):
Looks all good!

PR #860, opened 2024-01-21
docs(examples): update list example and list.tape

Emivvvvv:
Issue: https://github.com/ratatui-org/ratatui/issues/846

This PR adds:

for list.rs
- subjectively better-looking front
- change list example to a todo list example
- status of a TODO can be changed, further info can be seen under the list.

for list.tape

Here is the new output:
list


Valentin271:
I'll take a look at the code a little bit later, just wanted to say that the gif looks awesome!

Emivvvvv:
> I'll take a look at the code a little bit later, just wanted to say that the gif looks awesome!

Thanks a lot! I'm not sure should we remove the events list or not. It is irrelevant with the TODO list but kinda useful to have in an example.

examples/list.rs
```diff
@@ -69,77 +77,53 @@ impl<T> StatefulList<T> {
...
-                ("Event23", "INFO"),
-                ("Event24", "WARNING"),
-                ("Event25", "INFO"),
-                ("Event26", "INFO"),
-            ],
+            events: Self::create_random_events()
+        }
+    }
+
+    /// Changes the satis of the item list
```
orhun, line 99:
```suggestion
    /// Changes the status of the item list
```

examples/list.rs
```diff
@@ -202,80 +188,159 @@ fn run_app<B: Backend>(
...
-        .map(|&(event, level)| {
-            // Colorcode the level depending on its type
-            let s = match level {
-                "CRITICAL" => Style::default().fg(Color::Red),
-                "ERROR" => Style::default().fg(Color::Magenta),
-                "WARNING" => Style::default().fg(Color::Yellow),
-                "INFO" => Style::default().fg(Color::Blue),
+        .map(|level| {
+            // sets s to the tailwind color depending on its level
+            let s = match level.as_str() {
```
orhun, line 320: I would rather remove this comment and rename `s` to something more explanatory.

orhun (commented):
Good stuff!

examples/list.rs
Emivvvvv: Oh yeah, I must have missed it. Thanks!

joshka:
I think go ahead and remove the events list. A single good example is worth more than multiple of the same

examples/list.rs
```diff
@@ -69,77 +77,53 @@ impl<T> StatefulList<T> {
 /// Check the event handling at the bottom to see how to change the state on incoming events.
 /// Check the drawing logic for items on how to specify the highlighting style for selected items.
 struct App<'a> {
-    items: StatefulList<(&'a str, usize)>,
-    events: Vec<(&'a str, &'a str)>,
+    items: StatefulList<(&'a str, &'a str, bool)>,
```
joshka, line 80: How about making this a list of TodoItems?

examples/list.rs
```diff
@@ -202,80 +188,159 @@ fn run_app<B: Backend>(
...
-            for _ in 0..i.1 {
-                lines.push(
-                    "Lorem ipsum dolor sit amet, consectetur adipiscing elit."
-                        .italic()
-                        .into(),
-                );
-            }
-            ListItem::new(lines).style(Style::default().fg(Color::Black).bg(Color::White))
+        .enumerate()
+        .map(|(i, (todo, _info, status))| {
```
joshka, line 241: This might be good to extract to a method (which could make sense putting in the item itself)

examples/list.rs
```diff
@@ -202,80 +188,159 @@ fn run_app<B: Backend>(
...
-        .highlight_symbol(">> ");
+        .highlight_symbol(">");
 
     // We can now render the item list
-    f.render_stateful_widget(items, item_list_area, &mut app.items.state);
+    f.render_stateful_widget(items, inner_area, &mut app.items.state);
+
+    let info = if let Some(i) = app.items.state.selected() {
+        match app.items.items[i].2 {
+            true => "✓ DONE: ".to_string() + app.items.items[i].1,
```
joshka, line 276: This would work well as a enum instead of a bool. It would give you a place to hang this conversion off (impl Display perhaps, though that wouldn't give you any ability to add color or style, perhaps to_line()?)

examples/list.rs
```diff
@@ -202,80 +188,159 @@ fn run_app<B: Backend>(
...
+        .title("Use j k or ↓ ↑ to move, h or ← to unselect, l → to change status.")
+        .title_alignment(Alignment::Center);
+    f.render_widget(header, header_area);
+
+    render_todo(f, app, &[upper_item_list_area, lower_item_list_area]);
+
+    render_events_list(f, app, event_list_area)
+}
+
+// Renders todo list and info part
```
joshka, line 214: It might be worth splitting this to two methods. One to render the list and the other to render the todo details.

examples/list.rs
```diff
@@ -202,80 +188,159 @@ fn run_app<B: Backend>(
...
+fn render_todo(f: &mut Frame, app: &mut App, area: &[Rect; 2]) {
+    // We create two blocks, one is for the header (outer) and the other is for list (inner).
```
joshka, line 216: This seems a bit confusing as to the need for this (reading on my phone though, so perhaps it's more obvious with greater context)

joshka (commented):
Looking good.

joshka:
I think we should also enable the option that always allocates space for the selection to avoid the display jumping.

examples/list.rs
Emivvvvv: I'm going to separate the todo and info as two functions. Then it should be clearer.

examples/list.rs
Emivvvvv: We need two different strings for todo list and todo info so impl Display won't work imo. But I agree with the opinion of using an enum.

examples/list.rs
```diff
@@ -1,24 +1,70 @@
...
+            Status::Completed => (
+                Style::default().fg(COMPLETED_TEXT_COLOR).bg(bg_color),
+                " ✓ ".to_string() + &self.todo,
+            ),
+            Status::Todo => (
+                Style::default().fg(TEXT_COLOR).bg(bg_color),
+                " ☐ ".to_string() + &self.todo,
+            ),
+        };
+        ListItem::new(Line::from(todo_str)).style(style)
```
joshka, line 44: Does something like this read a bit clearer?

```suggestion
        let line = match self.status {
            TodoState::Todo =>
                Line::styled(format!(" ☐ {}", self.todo, (TEXT_COLOR, bg_color))),
            TodoState::Done =>
                Line::styled(format!(" ✓ {}", self.todo), (COMPLETED_TEXT_COLOR, bg_color)),
        };
        ListItem::new(line)
```

Or does the style really need to be on the list item to make this look good?

Emivvvvv:
This is how the example looks after removing events list

list

examples/list.rs
```diff
@@ -1,24 +1,70 @@
...
+        let mut result_vec = vec![];
+        for item in vec {
+            result_vec.push(TodoItem {
+                todo: item.0.to_string(),
+                info: item.1.to_string(),
+                status: item.2,
+            })
+        }
+        result_vec
+    }
```
joshka, line 57: It wouldn't hurt to move this to the StatefulList and just accept the vec there (it only actually has to be &[])


```rust
impl From<(&str, &str, bool)> for TodoItem {
    fn from((todo, item, status): (&str, &str, bool)) -> Self {
        Self { todo, item, status }
    }
}

impl StatefulList
    fn with_items(items: &[(&str, &str, bool)>]) -> Self {
        Self {
            items: items.map(TodoItem::from).collect()
            state: ListState::default(),
            last_selected: None,
        }
    }
```

examples/list.rs
```diff
@@ -1,24 +1,70 @@
...
 
 use crossterm::{
     event::{self, DisableMouseCapture, EnableMouseCapture, Event, KeyCode, KeyEventKind},
     execute,
     terminal::{disable_raw_mode, enable_raw_mode, EnterAlternateScreen, LeaveAlternateScreen},
 };
-use ratatui::{prelude::*, widgets::*};
+use ratatui::{prelude::*, style::palette::tailwind, widgets::*};
+
+const TODO_HEADER_BG: Color = tailwind::BLUE.c950;
```
joshka, line 10: using consts for these are good, but perhaps it's worth moving them closer to the impl Widget block? Not 100% sure on this - thoughts?

joshka (commented):
A bunch of comments all about refactoring this to the fullest and using the opportunity to set some really neat standards. Nothing in my comments is an error in this PR - just ideas for making this really neat and readable.

examples/list.rs
```diff
@@ -1,24 +1,65 @@
...
+struct StatefulList {
     state: ListState,
-    items: Vec<T>,
+    items: Vec<TodoItem>,
     last_selected: Option<usize>,
 }
 
-impl<T> StatefulList<T> {
-    fn with_items(items: Vec<T>) -> StatefulList<T> {
+impl StatefulList {
```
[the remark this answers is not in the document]
Emivvvvv: > If we add first / last methods and keys that would help too.

I like this idea

> What do you think about making there always be a selection?

I think it is good to have this in the example. This might be worth having for some real-life implementations. However, this makes the example harder to read and understand. If simplicity is more important for these examples I think this selection thing can go, else I can try to simplify them.

examples/list.rs
Emivvvvv: I like the idea of putting const at the top but ofc I'm open to suggestions. I think putting const closer to impl Widget block would make that area look crowded.

Valentin271: I like having consts at the top too, so its like easy-to-change parameters.

kdheepak:
I think some of the contrast can be improved (the greens on darker colors particularly) but that's probably just me (I have color issues).

kdheepak:
Great job improving all the examples! Excited for all the PRs!

Emivvvvv:
This is the final GIF output.

list

The only thing that is not addressed is removing the unselecting option. It is cool to have tbh but simplicity is something important. I'm ready to make the needed changes depending on your opinion on this.

Emivvvvv:
> I think some of the contrast can be improved (the greens on darker colors particularly) but that's probably just me (I have color issues).

Here are some different greens.

c400:
Screenshot 2024-01-22 at 20 33 16

c500:
Screenshot 2024-01-22 at 20 33 02

c600 (currently used):
Screenshot 2024-01-22 at 20 33 28

c700:
Screenshot 2024-01-22 at 20 33 52

I personally liked the c600 most. An improvement could be also changing the selected text to green if it's completed. Which one do you find best among these?

Emivvvvv:
> Great job improving all the examples! Excited for all the PRs!

Thanks! First time contributing to open open-source. This is a big win for me because now I'm able to talk with and learn more from experienced developers. Happy to hear that you're also happy with my work :)

Valentin271 (commented):
> The only thing that is not addressed is removing the unselecting option.

I personally think it's cool, I like a little bit of interactivity.

The only thing I don't really like is this weird line because of line colors alternating (see my comment):
image

---

btw I think you will need to rebase this.

examples/list.rs
```diff
@@ -62,220 +318,30 @@ impl<T> StatefulList<T> {
...
+            0 => NORMAL_ROW_COLOR,
+            _ => ALT_ROW_COLOR,
+        };
+        let line = match self.status {
+            Status::Todo => Line::styled(format!(" ☐ {}", self.todo), (TEXT_COLOR, bg_color)),
+            Status::Completed => Line::styled(
+                format!(" ✓ {}", self.todo),
+                (COMPLETED_TEXT_COLOR, bg_color),
+            ),
+        };
```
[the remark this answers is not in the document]
Emivvvvv: It didn't work. 

I also tried this but also didn't work.

```
let _ = inner_area.rows().enumerate().map(|(i, line)| {
            let bg_color = match i % 2 {
                0 => NORMAL_ROW_COLOR,
                _ => ALT_ROW_COLOR,
            };
            buf.set_style(line, bg_color);
        });
```

maybe I used inner_area(Rect used for list) or buf.set_style() incorrectly. Any other ideas?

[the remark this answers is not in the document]
Valentin271: Maybe I expressed myself wrong. You are trying to style the `Line` struct _inside_ `ListItem`, this line is rendered with an offset because of the selection. But you can style `ListItem` _directly_ like below to style the entire item.

`Line` is a _part_ of `ListItem`, whereas `ListItem` represent an entire entry in the list.

Hope that was clear enough, here is the code I was thinking of originally (it works):

```patch
diff --git a/examples/list.rs b/examples/list.rs
index 2d3ad14..e3037b7 100644
--- a/examples/list.rs
+++ b/examples/list.rs
@@ -325,14 +325,14 @@ impl TodoItem<'_> {
             _ => ALT_ROW_COLOR,
         };
         let line = match self.status {
-            Status::Todo => Line::styled(format!(" ☐ {}", self.todo), (TEXT_COLOR, bg_color)),
+            Status::Todo => Line::styled(format!(" ☐ {}", self.todo), TEXT_COLOR),
             Status::Completed => Line::styled(
                 format!(" ✓ {}", self.todo),
                 (COMPLETED_TEXT_COLOR, bg_color),
             ),
         };
 
-        ListItem::new(line)
+        ListItem::new(line).bg(bg_color)
     }
 }
 ```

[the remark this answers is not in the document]
Emivvvvv: Interesting... I tried this at the beginning but apparently, I made some mistakes so it didn't work.

But thanks!

Emivvvvv:
list

Thank you @Valentin271!

Valentin271 (approved):

kdheepak:
> Here are some different greens.

Thanks for experimenting! `c400` is the highest contrast and easiest to read for me, but `c500` also seems good to me (hard to say exactly without seeing it in while using it in the terminal / seeing the colors change etc). `c600` seems a little too washed out for me.

I did put these numbers into a contrast checker and was experimenting with a few values: e.g. https://coolors.co/contrast-checker/16a34a-0f172a

Here's `c400`:

image

And `c600`:

image

The website says this at the bottom:

> Regarding colors, the standard defines two levels of contrast ratio: AA (minimum contrast) and AAA (enhanced contrast). The level AA requires a contrast ratio of at least 4.5:1 for normal text and 3:1 for large text (at least 18pt) or bold text. The level AAA requires a contrast ratio of at least 7:1 for normal text and 4.5:1 for large text or bold text.

It seems like the colors you have are good, it's just a question of how much better it can be in terms of contrast. Especially in the terminal with small text, if we wanted to optimize for readability, I'd recommend a contrast ratio of 7:1 or higher.

Emivvvvv:
> It seems like the colors you have are good, it's just a question of how much better it can be in terms of contrast. Especially in the terminal with small text, if we wanted to optimize for readability, I'd recommend a contrast ratio of 7:1 or higher.

I like darker colors in general so that might be the reason why I liked the c600 more. But let's meet in the middle and go with c500 then 👍🏻

orhun (approved):

Emivvvvv:
> Hmm @Emivvvvv I kind of messed up the rebase, could you push your branch again?

I force pushed and that messed up the whole PR I guess 💀

Valentin271:
Do you still have the changes on your computer? If so just pushing should be fine. Otherwise I'm really sorry that's my bad ... :/

Emivvvvv:
Should I open a new PR?

Emivvvvv:
> Do you still have the changes on your computer? If so just pushing should be fine. Otherwise I'm really sorry that's my bad ... :/

Yes I have the changes. I can open a new PR

Emivvvvv:
> Do you still have the changes on your computer? If so just pushing should be fine. Otherwise I'm really sorry that's my bad ... :/

I force pushed from the beginning so I think that messed up this PR. The changes are on the branch but nothing happening here

Valentin271:
Probably because its closed, open a new one its fine (I can't reopen it)

Emivvvvv:
Here is the new PR: https://github.com/ratatui-org/ratatui/pull/864

PR #861, opened 2024-01-21
docs(examples): impl Widget for App, SelectedTab

Emivvvvv:
Issue: https://github.com/ratatui-org/ratatui/issues/819

This PR adds:

impl `Widget` for `App`, `SelectedTab` in tabs example.

### DIFF
--- a/examples/tabs.rs
+++ b/examples/tabs.rs
@@ -13,9 +13,9 @@
 //! [examples]: https://github.com/ratatui-org/ratatui/blob/main/examples
 //! [examples readme]: https://github.com/ratatui-org/ratatui/blob/main/examples/README.md
 
-use std::{error::Error, io, io::stdout};
+use std::{io, io::stdout};
 
-use color_eyre::config::HookBuilder;
+use color_eyre::{config::HookBuilder, Result};
 use crossterm::{
     event::{self, Event, KeyCode, KeyEventKind},
     terminal::{disable_raw_mode, enable_raw_mode, EnterAlternateScreen, LeaveAlternateScreen},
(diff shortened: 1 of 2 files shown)


examples/tabs.rs
joshka, line 1: What about making the tabs here actually have some sort of meaningful content instead of just rendering a different border.

examples/tabs.rs
```diff
@@ -31,6 +31,45 @@ enum SelectedTab {
     Tab3,
 }
 
+impl Widget for SelectedTab {
```
joshka, line 34: Move this below the Widget impl for App - to make this read top to bottom more easily.

examples/tabs.rs
```diff
@@ -61,6 +100,16 @@ struct App {
     pub selected_tab: SelectedTab,
 }
 
+impl Widget for &App {
```
joshka, line 103: Move this below impl App (and split it in two at line 128)

examples/tabs.rs
```diff
@@ -75,6 +124,37 @@ impl App {
     pub fn previous(&mut self) {
         self.selected_tab = self.selected_tab.previous();
     }
+
+    fn render_tabs(&self, area: Rect, buf: &mut Buffer) {
```
joshka, line 128: Split App into two impl sections, this part should be below `impl Widget for App {}`

examples/tabs.rs
```diff
@@ -75,6 +124,37 @@ impl App {
...
+        let block = Block::new()
```
joshka, line 129: Can this be a paragraph?

examples/tabs.rs
```diff
@@ -75,6 +124,37 @@ impl App {
...
+        self.selected_tab.render(area, buf);
+    }
+
+    fn draw(&self, terminal: &mut Terminal<impl Backend>) -> io::Result<()> {
+        terminal.draw(|f| f.render_widget(self, f.size()))?;
+        Ok(())
+    }
 }
 
 fn main() -> Result<(), Box<dyn Error>> {
```
joshka, line 160: This should be up top, and consider using the app style from constrains.rs

joshka (commented):
Mostly comments on rearranging this for a more natural reading order that helps readers not have to read code above the reading position in order to understand code below that.

examples/tabs.rs
Emivvvvv: I wanted to leave things as simple as possible, that's why I used that approach. But I'll find a way to put some "Simple" widgets.

joshka: Perhaps just a different Paragraph on each render would be fine, but I think calling back into that inner block method seems like it could be a bit confusing (effectively the current code says to render each tab differently, but actually they're all the same)

joshka:
I tweaked a bunch of things:
- changed the border to a custom one that joins against the tab
- moved the title to the same line as the tab
- made the footer a single line
- muted the colors a little
- rearranged the order of the code for better reading order
- moved the From<SelectedTab> for Line to SelectedTab::title
- moved the colors to SelectedTab::palette()
- derived default on App and removed new()
- rebased on the current main and brought in Line::centered()
- added an AppState enum and moved event handling to App::handle_events()
- added quit instructions to footer and remove h/l (these still work, just aren't advertised)

  Made with VHS

Emivvvvv:
Looks really good! I loved the idea of making tabs connected to the border. 

But I'm not sure about tabs locations because they look not centered nor left aligned. I'm not a fan of that positioning. And we could center the text also. I don't know who it calls but centered horizontally and vertically. Not something significant but it may improve the look.

Note: The reason I'm not active for a couple of days is that I'm EXTREMELY busy with the final days of my semester and I'll be turning back to my home country (I was doing an Erasmus program) so I'm setting up for that. I hope I'll become active with issues and PRs again soon. I just wanted to inform you because of my two open PRs.

joshka:
>Looks really good! I loved the idea of making tabs connected to the border.

Glad you like it. I'm a fan of trying to find ways to help push apps towards using fewer box drawing borders. I think it starts with demonstrating nicer looking TUIs and continues into playing with palettes, and default layouts that avoid them.

>But I'm not sure about tabs locations because they look not centered nor left aligned. I'm not a fan of that positioning. And we could center the text also. I don't know who it calls but centered horizontally and vertically. Not something significant but it may improve the look. I'll slap two examples up to test it.

I can see that perspective. What about putting the title all the way over on the right, or the tabs on the right? If either of those don't work, then I'll move the title back up above things. My rationale on compressing it on the same line is to help gently demonstrate that apps shouldn't throw away vertical space when they can help it.

>Note: The reason I'm not active for a couple of days is that I'm EXTREMELY busy with the final days of my semester and I'll be turning back to my home country (I was doing an Erasmus program) so I'm setting up for that. I hope I'll become active with issues and PRs again soon. I just wanted to inform you because of my two open PRs.

Not a problem in the slightest - I assume that everyone has their own stuff going on and github open source is something that is async. Good luck with your studies. We sometimes update PRs like this to get them over the line like this.

joshka:
I like both of these:
  Made with VHS
  Made with VHS

joshka (approved):

PR #862, opened 2024-01-22
feat(Text): add alignment convenience functions

Eeelco:
Added convenience functions left_aligned(), centered() and right_aligned() plus unit tests. No example code updated since none seems to be using Text::alignment().

### DIFF
--- a/src/text/text.rs
+++ b/src/text/text.rs
@@ -231,6 +231,8 @@ impl<'a> Text<'a> {
     /// Sets the alignment for this text.
     ///
     /// Defaults to: [`None`], meaning the alignment is determined by the rendering widget.
+    /// Setting the alignment of a Text generally overrides the alignment of its
+    /// parent Widget.
     ///
     /// Alignment can be set individually on each line to override this text's alignment.
     ///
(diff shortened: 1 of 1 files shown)


Eeelco:
PR 2/4 for #853

src/text/text.rs
```diff
@@ -211,6 +211,8 @@ impl<'a> Text<'a> {
     /// Sets the alignment for this text.
     ///
     /// Defaults to: [`None`], meaning the alignment is determined by the rendering widget.
+    /// Setting the alignment of a Text generally overrides the alignment of its
+    /// parent Widget.
     ///
     /// Alignment can be set individually on each line to override this text's alignment.
```
Valentin271, line 217: This line could also be reported on other shortcut methods.

Valentin271 (approved):
Looks good, my comment is optional.

joshka:
You'll probably need to rebase this.

orhun (approved):

src/text/text.rs
Eeelco: I also added this to the new convenience functions, or which functions do you mean?

Eeelco: Aah I see, thanks for the clarification! Added it to the other methods

Valentin271:
I think you pushed modifications from your other PR (for line) because you made the changes on your main branch.
That's what I deduce looking at your main branch.

I recommend you update your fork - using the button in the screenshot below - and then rebase this branch. You should have a small conflict in `examples/list.rs`

image

Eeelco:
> I think you pushed modifications from your other PR (for line) because you made the changes on your main branch. That's what I deduce looking at your main branch.
> 
> I recommend you update your fork - using the button in the screenshot below - and then rebase this branch. You should have a small conflict in `examples/list.rs`
> 
> image

I tried to rebase properly but only made it worse, so I just re-created the branch and force pushed, hope that's ok

Valentin271:
> I tried to rebase properly but only made it worse, so I just re-created the branch and force pushed, hope that's ok

Yep perfectly fine, thanks!

PR #864, opened 2024-01-23
docs(examples): update list example and list.tape

Emivvvvv:
Issue: https://github.com/ratatui-org/ratatui/issues/846

This PR adds:

for list.rs

- subjectively better-looking front
- change list example to a todo list example
- status of a TODO can be changed, further info can be seen under the list.

for list.tape

Here is the new output:

list

### DIFF
--- a/examples/list.rs
+++ b/examples/list.rs
@@ -1,8 +1,4 @@
-use std::{
-    error::Error,
-    io,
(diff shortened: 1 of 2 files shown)


Valentin271:
I rebased on main, this time correctly

PR #865, opened 2024-01-24
refactor(colors_rgb): impl widget on mutable refs

joshka:
This commit refactors the colors_rgb example to implement the Widget
trait on mutable references to the app and its sub-widgets. This allows
the app to update its state while it is being rendered.

Additionally the main and run functions are refactored to be similar to
the other recent examples. This uses a pattern where the App struct has
a `run` method that takes a terminal as an argument, and the main
function is in control of initizing and restoring the terminal and
installing the error hooks.

### DIFF
--- a/examples/colors_rgb.rs
+++ b/examples/colors_rgb.rs
@@ -1,12 +1,22 @@
 /// This example shows the full range of RGB colors that can be displayed in the terminal.
 ///
 /// Requires a terminal that supports 24-bit color (true color) and unicode.
+///
+/// This example also demonstrates how implementing the Widget trait on a mutable reference
+/// allows the widget to update its state while it is being rendered. This allows the fps
+/// widget to update the fps calculation and the colors widget to update a cached version of
+/// the colors to render instead of recalculating them every frame.
+///
+/// This is an alternative to using the StatefulWidget trait and a separate state struct. It is
+/// useful when the state is only used by the widget and doesn't need to be shared with other
+/// widgets.
 use std::{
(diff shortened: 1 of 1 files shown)


orhun (approved):
Nice!

Also just tested and it works.

examples/colors_rgb.rs
```diff
@@ -141,75 +150,113 @@ impl Default for Fps {
...
             for (yi, y) in (area.top()..area.bottom()).enumerate() {
+                // render a half block character for each row of pixels with the foreground color
+                // set to the color of the pixel and the background color set to the color of the
+                // pixel below it
                 let fg = colors[yi * 2][xi];
                 let bg = colors[yi * 2 + 1][xi];
                 buf.get_mut(x, y).set_char('▀').set_fg(fg).set_bg(bg);
             }
         }
+        self.frame_count += 1;
```
Valentin271, line 205: We could use the `frame_count` in frame, that would make an example usage.

Valentin271 (approved):
Other than my comment looks good

examples/colors_rgb.rs
joshka: The problem is that it's not available from the widget.render method as you don't have access to frame there. So it would have to be pushed down to each widget as part of the main draw call, which makes this a little messier than I'd like. Not sure if there's a simple solution to that that I've missed.

Valentin271: Oh right, that's fine then

PR #873, opened 2024-01-25
feat(Span): add alignment functions

### LINKED ISSUE #853: Add Alignment convenience methods to Line, Text and Paragraph (and perhaps Span).

Eeelco:
Implemented functions that convert Span into a left-/center-/right-aligned Line. Implemented unit tests.

Closes #853

### DIFF
--- a/src/text/span.rs
+++ b/src/text/span.rs
@@ -275,6 +275,46 @@ impl<'a> Span<'a> {
             .filter(|g| *g != "\n")
             .map(move |g| StyledGrapheme { symbol: g, style })
     }
+
+    /// Converts this Span into a left-aligned [`Line`]
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// # use ratatui::prelude::*;
+    /// let span = Span::styled("Test Content", Style::new().green().italic());
+    /// assert_eq!(span.left_aligned().alignment, Some(Alignment::Left));
+    /// ```
+    #[must_use = "method moves the value of self and returns the modified value"]
+    pub fn left_aligned(self) -> Line<'a> {
+        Line::from(self).left_aligned()
+    }
+
+    /// Converts this Span into a center-aligned [`Line`]
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// # use ratatui::prelude::*;
+    /// let span = Span::styled("Test Content", Style::new().green().italic());
+    /// assert_eq!(span.centered().alignment, Some(Alignment::Center));
+    /// ```
+    pub fn centered(self) -> Line<'a> {
+        Line::from(self).centered()
+    }
+
+    /// Converts this Span into a right-aligned [`Line`]
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// # use ratatui::prelude::*;
+    /// let span = Span::styled("Test Content", Style::new().green().italic());
+    /// assert_eq!(span.right_aligned().alignment, Some(Alignment::Right));
+    /// ```
+    pub fn right_aligned(self) -> Line<'a> {
+        Line::from(self).right_aligned()
+    }
 }
 
 impl<'a, T> From<T> for Span<'a>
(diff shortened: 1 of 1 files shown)


Valentin271:
I've just discovered the issue states this. However I'm not sure we want an implicit conversion like that. Spans can't be aligned and having alignment methods on them feels rather confusing.

Do you think we really need this @joshka?

joshka:
> I've just discovered the issue states this. However I'm not sure we want an implicit conversion like that. Spans can't be aligned and having alignment methods on them feels rather confusing.
> 
> Do you think we really need this @joshka?

The main thing that this opens up is the ability to do the following

```rust
"foobar".red().centered().render(area, buf);
```

I think that's a pretty convenient syntax. What do you think?

src/text/span.rs
```diff
@@ -275,6 +275,46 @@ impl<'a> Span<'a> {
...
     }
+
+    /// Converts this Span into a left-aligned [`Line`]
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// # use ratatui::prelude::*;
+    /// let span = Span::styled("Test Content", Style::new().green().italic());
+    /// assert_eq!(span.left_aligned().alignment, Some(Alignment::Left));
```
joshka, line 286: Consider the following perhaps as an example of how to use this effectively:

```suggestion
    /// assert_eq!(
    ///     "Test Content".green().italic().centered(),
    ///     Line::raw("Test Content")
    ///         .style((Color::Green, Modifier::Italic))
    ///         .alignment(Alignment::Center),
    /// );
```

joshka (commented):
Consider changing the name based on the advice in https://rust-lang.github.io/api-guidelines/naming.html#c-conv if you think it would be sensible. Perhaps `to_centered_line`?

Valentin271:
> ```rust
> "foobar".red().centered().render(area, buf);
> ```

Yeah that's pretty neat. I'm not against merging this as is but following the Rust conventions is even better.

src/text/span.rs
Eeelco: Hm, looks like this will produce different results: 
`Line { spans: [Span { content: "Test Content", style: Style { fg: Some(Green), bg: None, underline_color: None, add_modifier: ITALIC, sub_modifier: NONE } }], style: Style { fg: None, bg: None, underline_color: None, add_modifier: NONE, sub_modifier: NONE }, alignment: Some(Center) }`
and
`Line { spans: [Span { content: "Test Content", style: Style { fg: None, bg: None, underline_color: None, add_modifier: NONE, sub_modifier: NONE } }], style: Style { fg: Some(Green), bg: None, underline_color: None, add_modifier: ITALIC, sub_modifier: NONE }, alignment: Some(Center) }`
Does this matter? If not, I would just write an example without any assertions

joshka: Yeah my bad, we recently changed line so that it had its own style. A styled span that is wrapped in an unstyled line is different to a styled line. I don't mind leaving the assertions off from this. I think it's pretty obvious what the result is based on the context.

joshka (approved):

PR #874, opened 2024-01-26
chore(CONTRIBUTING): remove part about squashing commits

Valentin271:
Removes the part about squashing commits from the CONTRIBUTING file.

We no longer require that because github squashes commits when merging. This will cleanup the CONTRIBUTING file a bit which is already quite dense.


orhun (approved):
lesgo

Issue #843, opened 2024-01-18
Deprecate `StretchLast` and choose new default for `Flex`


kdheepak:
## Problem

`Flex::StretchLast` is an implementation detail and ill-defined. There are many situations were this does not behave the way we could expect. It can also be unstable, i.e. for the same input you can get different results different times you run it.

## Solution

Choose one of the following as the default. 

- `Flex::Stretch`, OR 
- `Flex::Start`

`Flex::Stretch` has the advantage that it will not break any existing UI code. It is also a better default because it is more versatile. Our existing constraint names work better for `Flex::Stretch`. For example, `Min(3)` will never be more than `3` in `Flex::Start` making it useless compared to `Length(3)` or other similar `STRONG`  constraints. In `Flex::Stretch`, `Min(3)` will perform as the name suggests, i.e. that segment will be anywhere between `3` and a value that that fills remaining excess space.

The argument for choosing `Flex::Start` is that the default in CSS flex is left justified.

PR #881, opened 2024-01-28
feat(layout)!: Change `Flex::default()` 

kdheepak:
This PR makes a number of simplifications to the layout and constraint features that were added after v0.25.0.

For users upgrading from v0.25.0, the net effect of this PR (along with the other PRs) is the following:

- New `Flex` modes have been added.
  - `Flex::Start` (new default)
  - `Flex::Center`
  - `Flex::End`
  - `Flex::SpaceAround`
  - `Flex::SpaceBetween`
  - `Flex::Legacy` (old default)
- `Min(v)` grows to allocate excess space in all `Flex` modes instead of shrinking (except in `Flex::Legacy` where it retains old behavior). 
- `Fill(1)` grows to allocate excess space, growing equally with `Min(v)`.

---

The following contains a summary of the changes in this PR and the motivation behind them.

**`Flex`**

- Removes `Flex::Stretch`
- Renames `Flex::StretchLast` to `Flex::Legacy`

**`Constraint`**

- Removes `Fixed`
- Makes `Min(v)` grow as much as possible everywhere (except `Flex::Legacy` where it retains the old behavior)
- Makes `Min(v)` grow equally as `Fill(1)` while respecting `Min` lower bounds. When `Fill` and `Min` are used together, they both fill excess space equally.

Allowing `Min(v)` to grow still allows users to build the same layouts as before with `Flex::Start` with no breaking changes to the behavior.

This PR also removes the unstable feature `SegmentSize`.

This is a breaking change to the behavior of constraints. If users want old behavior, they can use `Flex::Legacy`.

```rust
Layout::vertical([Length(25), Length(25)]).flex(Flex::Legacy)
```

Users that have constraint that exceed the available space will probably not see any difference or see an improvement in their layouts. Any layout with `Min` will be identical in `Flex::Start` and `Flex::Legacy` so any layout with `Min` will not be breaking.

Previouslt, `Table` used `EvenDistribution` internally by default, but with that gone the default is now `Flex::Start`. This changes the behavior of `Table` (for the better in most cases). The only way for users to get exactly the same as the old behavior is to change their constraints. I imagine most users will be happier out of the box with the new Table default.

Resolves https://github.com/ratatui-org/ratatui/issues/843

Thanks to @joshka for the direction

## Background

This is a follow up to the following PRs:

- `Fixed` and `Fill`: https://github.com/ratatui-org/ratatui/pull/783 https://github.com/ratatui-org/ratatui/pull/880
- `Flex`: https://github.com/ratatui-org/ratatui/pull/804
- `Refactor`: https://github.com/ratatui-org/ratatui/pull/788

`Flex::Stretch` was introduced as a stable alternative to `Flex::StretchLast`, as well as a backward compatible way to make `SegmentSize::EvenDistribution` work. 
`Constraint::Fixed` was introduced to fix column spacing.

## Motivation

Typically, there's two ways a layout's constraints are not met:

- When there's too much space

e.g. there's a 100px wide space but a user is only requesting for a total of 75:

image
image

- When there's not enough space

e.g. there's a 50px wide space but a user is requesting for a total of 75:

image

In `Stretch` flex modes, constraints were usually violated in both cases.

This caused instability, and needed constraints like `Fixed` to ensure that some lengths were prioritized over others (e.g. we typically want table column spacing should take precedence over column constraints provided by users).

In `Flex::Start`, `Flex::Center`, `Flex::End`, `Flex::SpaceBetween`, `Flex::SpaceAround`, when there's excess space constraints are not violated:

https://github.com/ratatui-org/ratatui/assets/1813121/56582ce3-53b0-4b43-9689-afbf15a615cc

When there's no enough space constraints are still violated:

image

But the priority order of constraints allows users to control the layout as they see fit.

**TLDR**: `Flex::Stretch` broke more constraints more often than other layouts and with `Min` growing, it has marginal utility over other `Flex` modes, and we can remove it.

---

Another change in this PR is the behavior of `Min`. Now `Min` grows in all `Flex` layouts (except `Flex::Legacy`). 
This allows `Flex::Stretch` like behavior with `Min(0)` instead. e.g. the following two behave identically for any available space:

```diff
- Layout::horizontal([Length(25), Length(25), Length(25)]).flex(Flex::Stretch)
+ Layout::horizontal([Min(0), Min(0), Min(0)])
```

Previously, `Min(v)` would always try to collapse to the value `v` in every `Flex` mode, but because of `Stretch` semantics it would grow to fill remaining space. This behavior made people use `Min(0)` as a space filler. e.g.

```rust
let [header, main] = area.split(&Layout::vertical([Length(3), Min(0)]);
```

This PR makes this "space filler" growing behavior the default in all `Flex` layouts (except `Flex::Legacy` for backwards compatibility reasons), i.e. `Min(v)` does not collapse to the value `v` but instead tries to grow to the area size with low priority.

Current `main` branch:

image

image

This PR:

`Flex::Legacy` vs `Flex::Start`

image
image

`Min(v)` has another behavior which is that is limits the lower bound on the segment size with the user provided value. Previously, breaking a `Min(v)` constraint was more "costly" than `Length`, `Percentage`, `Ratio` etc, so people also used it like this:

```rust
let [main, input] = area.split(&Layout::vertical([Percentage(100), Min(3)]);
```

This PR retains that behavior too:

image

The mechanics by which this behavior is achieved is by ordering the weights for the constraints appropriately.

**TLDR**:  This effectively means that in any layout with `Min`, there shouldn't be any breaking change that a user experiences between `Flex::Legacy` (old default) and `Flex::Start` (new default). 

Additionally, in this PR, `Min` growing is weighted equally to `Fill` growing, which means you can get desirable behavior like this:

https://github.com/ratatui-org/ratatui/assets/1813121/f9c3d14d-08fa-4ac9-9daf-f8791d8b5f11

Since we are removing `Flex::Stretch`, we are also removing `SegmentSize::EvenDistribution`. Since `SegmentSize` is unstable, this PR removes `SegmentSize` entirely.

This PR also removes `Fixed` since `Fixed` was only necessary to have something higher priority in `Stretch` layouts. Without `Stretch` we don't need it anymore.
With the refactor in https://github.com/ratatui-org/ratatui/pull/788 spacers are used instead of `Fixed` for column spacing, and these spacers have a high priority. This makes `Fixed` unnecessary.


joshka (approved):

src/layout/layout.rs
```diff
@@ -40,25 +40,21 @@ thread_local! {
...
-/// avoid this behavior, add an unused `Constraint::Min(0)` as the last constraint. There is also an
-/// unstable API to prefer equal chunks if other constraints are all satisfied, see [`SegmentSize`]
-/// for more info.
+/// many of the constraints in order of their priorities.
 ///
 /// When the layout is computed, the result is cached in a thread-local cache, so that subsequent
-/// calls with the same parameters are faster. The cache is a simple HashMap, and grows
-/// indefinitely. (See <https://github.com/ratatui-org/ratatui/issues/402> for more information)
+/// calls with the same parameters are faster. The cache is a LruCache, and the size of the cache
+/// can be configured.
```
Valentin271, line 57: Perhaps link to the relevant method

Valentin271 (approved):
Looks good, had a small comment but I'm find merging as is.

We might want (in a later PR) to rework some doc examples to not use `Legacy`.

src/layout/layout.rs
kdheepak: Done!

PR #888, opened 2024-01-29
feat: Change rounding to make tests stable

kdheepak:
This fixes some unstable tests


orhun (approved):

src/layout/layout.rs
```diff
@@ -932,56 +936,63 @@ mod strengths {
...
     ///  <= x =>
     /// └       ┘
-    pub const SPACE_GROW: f64 = WEAK / 10.0;
+    pub const SPACE_GROW: f64 = WEAK * 10.0;
+
+    /// The strength to apply to growing the size of all segments equally.
+    ///
+    /// ┌       ┐
+    ///  <= x =>
+    /// └       ┘
```
Valentin271, line 994: Is this the right diagram?

Valentin271 (approved):

kdheepak:
This isn't an issue with the constraint weights. It seems to be an issue with rounding. 

Sometimes it does this:

```
[src/layout/layout.rs:806] i = 1
[src/layout/layout.rs:807] changes.get(&element.start) = Some(
    37.5,
)
[src/layout/layout.rs:808] changes.get(&element.end) = Some(
    62.5,
)
[src/layout/layout.rs:810] size = 25
```

and other times it does this:

```
[src/layout/layout.rs:806] i = 1
[src/layout/layout.rs:807] changes.get(&element.start) = Some(
    37.49999999999999,
)
[src/layout/layout.rs:808] changes.get(&element.end) = Some(
    62.5,
)
[src/layout/layout.rs:810] size = 26
```

kdheepak:
I think is this is a better way to round values:

```
            let start = (start * 100.0).round() as u16 / 100;
            let end = (end * 100.0).round() as u16 / 100;
```

src/layout/layout.rs
kdheepak: Resolved.

kdheepak:
I changed it to use the double precision rounding method. This commit has all the tests that I had to change to make it pass: https://github.com/ratatui-org/ratatui/pull/888/commits/1996db2ed2734a067b42476b71c14c9d61fa1c83

All of the changes are one off differences. 

To make it more clear what users would experience, I added some tests for `Flex::Start` and `Flex::SpaceBetween` in this commit: https://github.com/ratatui-org/ratatui/pull/888/commits/94bce308c3dc05bdcd215cf578dae27cbc7108cd

joshka:
> I think this might be a better way to round values:
> 
> 
> 
> ```
> 
>             let start = (start * 100.0).round() as u16 / 100;
> 
>             let end = (end * 100.0).round() as u16 / 100;
> 
> ```

That sounds like it would work

There's a recent blog post I saw about servo where they ran into similar issues and switched to fixed point instead of floating point to address it. I think this effectively does the same thing, but I wonder if there would be any difference in the calculations doing the multiply on everything as the first step, and rounding to ints then dividing at the end. (Merely curiosity here)

https://servo.org/blog/2024/01/26/two-months-in-servo/

kdheepak:
I just tried this and I think this is better?

```
            let start = ((start * 100.0).round() / 100.0).round() as u16;
            let end = ((end * 100.0).round() / 100.0).round() as u16;
```

`37.4999` becomes `3749.99` gets rounded to `3750` and then that becomes `37.5` and rounded to `38`.

kdheepak:
Do you mean multiplying all user input by `100.0` before running the solver and then dividing by a `100.0` at the end?

kdheepak:
I did it "properly" in this commit:

https://github.com/ratatui-org/ratatui/pull/888/commits/9e7c8624af81695b203b9ff16be433ea4dabc4e1

Which ends up causes some panics:

```
failures:

---- layout::layout::tests::split::min_max::case_4_min_max_priority stdout ----
thread 'layout::layout::tests::split::min_max::case_4_min_max_priority' panicked at src/layout/layout.rs:888:49:
attempt to multiply with overflow

---- layout::layout::tests::split::min_max::case_3_min_max_priority stdout ----
thread 'layout::layout::tests::split::min_max::case_3_min_max_priority' panicked at src/layout/layout.rs:888:49:
attempt to multiply with overflow

```

kdheepak:
There's two ways to do it

1. Round once

```
let start = (start * 100.0).round() as u16 / 100;
let end = (end * 100.0).round() as u16 / 100;
```

2. Round twice

```
let start = ((start * 100.0).round() / 100.0).round() as u16;
let end = ((end * 100.0).round() / 100.0).round() as u16;
```

The "round twice" makes the tests behave as they used to, but the "round once" works the same way was the scaling up the input and scaling down the output, which makes me think the "round once" approach is better?

The most recent commit has the scaling approach with the panics resolved by scaling after converting to f64.

src/layout/layout.rs
```diff
@@ -581,16 +592,15 @@ impl Layout {
         let spacing = self.spacing;
         let constraints = &self.constraints;
 
-        let area_size = Element::from((*variables.first().unwrap(), *variables.last().unwrap()));
-        configure_area(&mut solver, area_size, area_start, area_end)?;
-        configure_variable_constraints(&mut solver, &variables, area_size)?;
+        let area_size = area_end - area_start;
+        configure_variable_constraints(&mut solver, &variables, area_start, area_end)?;
```
joshka, line 596: Can this still use the area approach and just make that [start*100,end*100]?
What makes this need to change to use floats here?

src/layout/layout.rs
```diff
@@ -1460,14 +1464,14 @@ mod tests {
...
-            #[case(Rect::new(0, 0, 1, 1), &[NINETY, FULL], "a")]
-            #[case(Rect::new(0, 0, 1, 1), &[NINETY, DOUBLE], "a")]
+            #[case(Rect::new(0, 0, 1, 1), &[HALF, ZERO], "b")]
+            #[case(Rect::new(0, 0, 1, 1), &[HALF, HALF], "b")]
+            #[case(Rect::new(0, 0, 1, 1), &[HALF, FULL], "b")]
+            #[case(Rect::new(0, 0, 1, 1), &[HALF, DOUBLE], "b")]
+            #[case(Rect::new(0, 0, 1, 1), &[NINETY, ZERO], "b")]
+            #[case(Rect::new(0, 0, 1, 1), &[NINETY, HALF], "b")]
+            #[case(Rect::new(0, 0, 1, 1), &[NINETY, FULL], "b")]
+            #[case(Rect::new(0, 0, 1, 1), &[NINETY, DOUBLE], "b")]
```
joshka, line 1474: - 0/100 is 100% wrong
- 50/50 is debatable. I it should allocate the space as 0.5, 0.5 then round left to right giving the first item the space)
- 50/100, 50/200 are probably better 
- 90/0, 90/50 are wrong for the same reason that 0/100 is

src/layout/layout.rs
```diff
@@ -1482,31 +1486,131 @@ mod tests {
...
+            #[case(Rect::new(0, 0, 2, 1), &[QUARTER, ZERO], "bb")]
+            #[case(Rect::new(0, 0, 2, 1), &[QUARTER, QUARTER], "bb")]
+            #[case(Rect::new(0, 0, 2, 1), &[QUARTER, HALF], "bb")]
+            #[case(Rect::new(0, 0, 2, 1), &[QUARTER, FULL], "bb")]
+            #[case(Rect::new(0, 0, 2, 1), &[QUARTER, DOUBLE], "bb")]
+            #[case(Rect::new(0, 0, 2, 1), &[THIRD, ZERO], "bb")]
+            #[case(Rect::new(0, 0, 2, 1), &[THIRD, QUARTER], "bb")]
+            #[case(Rect::new(0, 0, 2, 1), &[THIRD, HALF], "bb")]
+            #[case(Rect::new(0, 0, 2, 1), &[THIRD, FULL], "bb")]
+            #[case(Rect::new(0, 0, 2, 1), &[THIRD, DOUBLE], "bb")]
```
joshka, line 1498: - 25/0 is wrong should be "aa"
- 25/25 is wrong should be "ab"
- 25/50 is wrong should be "ab"
- ...
- 33/0 is wrong
didn't look at all of these, but there's enough wrong rounding here to make this a bad result.
Not sure if it's the weight change or the 100x change that causes this - perhaps split the PR in two and make that more obvious.

As a pre-refactor on this it might be worth changing the constants to P000, P050, P100 etc. to make this more obviously broken. It could be worth also changing these to take constraints, width, result, description as params.

joshka (commented):
Generally this changes too many test results and doesn't describe why. That makes it difficult to review. An initial look at just the first tests suggests that the changes are wrong (some 100% so, some more debatable).

I'd recommend improving the tests to make them easier to read a bit and then basing a set of smaller changes that fixes one issue at a time in a way that can be easily seen to be correct.

src/layout/layout.rs
kdheepak: These are all changes because of the rounding, not weights. I'll split the weights into a separate PR to make it clear.

I don't think we should pay too much attention to these tests unfortunately. All of these are over or under constrained tests, in a 2 pixel area. Even being 1 pixel off because of rounding gives results that look "wrong".

kdheepak: If I use this following approach these tests don't break:

```rust
let start = ((start * 100.0).round() / 100.0).round() as u16;
let end = ((end * 100.0).round() / 100.0).round() as u16;
```

kdheepak: Here's a commit with no tests changed but all tests passing with the weights and area_size changes reverted: https://github.com/ratatui-org/ratatui/pull/888/commits/fff20fef53192907304002ec0e2153683a33a2d1

src/layout/layout.rs
kdheepak: It's adding an extra variable to the solver and the optimization program, which when debugging I wanted to make sure wasn't the cause of any instability, so I removed it. I do think it is cleaner that way but reverted the change now.

kdheepak:
I have reverted to this behavior:

```
let start = ((start * 100.0).round() / 100.0).round() as u16;
let end = ((end * 100.0).round() / 100.0).round() as u16;
```

i.e. the round twice approach in this comment: https://github.com/ratatui-org/ratatui/pull/888#issuecomment-1915870132

joshka (approved):
LGTM

PR #890, opened 2024-01-30
refactor: make layout tests a bit easier to understand

joshka:


kdheepak (approved):
LGTM. You can even combine `ratio_spacebetween` and `ratio_start` into `ratio` if you want

joshka:
> LGTM. You can even combine `ratio_spacebetween` and `ratio_start` into `ratio` if you want

I like these separate as it breaks up the weight of the tests a bit making it easier to focus on one area..

Issue #334, opened 2023-07-19
Axis ticks label positions is not scaled well


thoongnv:
## Problem
I feel like the X/ Y Axis label positions are not scaled/ aligned well. So I'm using this chart example (Chart-2) with my data:
https://github.com/tui-rs-revival/ratatui/blob/v0.22.0/examples/chart.rs

```
let data: Vec<(f64, f64)> = vec![
    (0.0, 19799459.0),
    (1.0, 21077099.0),
    (2.0, 20541902.0),
    (3.0, 18909404.0),
    (4.0, 18645169.0),
    (5.0, 21282285.0),
    (6.0, 22234309.0),
    (7.0, 20949579.0),
    (8.0, 20605463.0),
    (9.0, 16719816.0),
    (10.0, 18607126.0),
    (11.0, 18926021.0),
    (12.0, 19928185.0),
    (13.0, 19101790.0),
    (14.0, 20660632.0),
    (15.0, 21492383.0),
    (16.0, 21660903.0),
    (17.0, 16454098.0),
    (18.0, 17100739.0),
    (19.0, 17338056.0),
    (20.0, 17587407.0),
    (21.0, 18099830.0),
    (22.0, 18694563.0),
    (23.0, 19575447.0),
    (24.0, 19339535.0),
    (25.0, 20159173.0),
    (26.0, 22084681.0),
    (27.0, 25031109.0),
    (28.0, 27390012.0),
    (29.0, 30009448.0),
    (30.0, 32878528.0),
    (31.0, 34278063.0),
    (32.0, 34248061.0),
    (33.0, 35727267.0),
    (34.0, 33514861.0),
    (35.0, 31185488.0),
    (36.0, 30585935.0),
    (37.0, 32593481.0),
    (38.0, 33444410.0),
    (39.0, 32279025.0),
    (40.0, 32610465.0),
    (41.0, 35403858.0),
    (42.0, 36644960.0),
    (43.0, 34029229.0),
    (44.0, 30687033.0),
    (45.0, 32105288.0),
    (46.0, 33825097.0),
    (47.0, 34940276.0),
    (48.0, 35969156.0),
    (49.0, 39090908.0),
    (50.0, 39496700.0),
    (51.0, 41588700.0),
    (52.0, 42880681.0),
    (53.0, 41811780.0),
    (54.0, 40707691.0),
    (55.0, 40931338.0),
    (56.0, 41699624.0),
    (57.0, 45766022.0),
[...]

PR #891, opened 2024-01-30
feat: Make `Chart` generic over its data

mo8it:
Currently, the only way to provide data to a `Chart` is by providing a slice of points `&[(f64, f64)]`. But this is very bad for live charts since it means that you need to use bloody `remove(0)` or `drain` to remove old points from a `Vec` before pushing new ones. The problem is that `remove(0)` and `drain` have a time complexity of `O(n)` where `n` is the length of vector.

The optimal data structure for the data of a live chart would be a ring buffer like [`LocalRb`](https://docs.rs/ringbuf/0.3.3/ringbuf/ring_buffer/struct.LocalRb.html) from the `ringbuf` crate. By making the chart generic over its data, is able to use a ring buffer now. The data is just an iterator over points. Pushing an item to a ring buffer has the time complexity `O(1)`.

To demonstrate this performance problem, I created a benchmark which you can run with `cargo bench --bench chart`.

The difference isn't huge because the rendering itself is much more expensive. But the difference is bigger for bigger `n` and bigger steps (5 in the benchmark).

This is a breaking change, but most users don't need to change anything. They don't have to use a ring buffer. The `data()` method accepts anything that can be converted into an iterator over `(f64, f64)` which includes `&[(f64, f64)]`.
Users who create `Points` directly need to add `.iter()` for `coords`.

I could polish comments, docs and examples, but I wanted to see what you think of it before proceeding since I didn't create an issue first.

I can remove the usage of `ringbuf` from the examples if you want to keep them as simple as possible.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -58,7 +58,6 @@ palette = "0.7.3"
 pretty_assertions = "1.4.0"
 rand = "0.8.5"
 rand_chacha = "0.3.1"
-ringbuf = "0.3.3"
 rstest = "0.18.2"
 serde_json = "1.0.109"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -58,6 +58,7 @@ palette = "0.7.3"
 pretty_assertions = "1.4.0"
 rand = "0.8.5"
 rand_chacha = "0.3.1"
+ringbuf = "0.3.3"
 rstest = "0.18.2"
 serde_json = "1.0.109"
 
@@ -119,6 +120,10 @@ name = "barchart"
 harness = false
 
 [[bench]]
+name = "chart"
+harness = false
+
+[[bench]]
 name = "block"
 harness = false
 
@@ -136,7 +141,6 @@ harness = false
 [[bench]]
 name = "sparkline"
 harness = false
-
 
 [[example]]
 name = "barchart"
--- a/benches/chart.rs
+++ b/benches/chart.rs
@@ -4,7 +4,6 @@ use ratatui::{
     layout::Rect,
     widgets::{Chart, Dataset, Widget},
 };
-use ringbuf::{LocalRb, Rb};
 
 /// Benchmark for rendering a chart.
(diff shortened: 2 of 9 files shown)


mo8it:
I would like to get feedback first before bothering with signing commits 😅

joshka:
> Currently, the only way to provide data to a Chart is by providing a slice of points &[(f64, f64)]. But this is very bad for live charts since it means that you need to use bloody remove(0) or drain to remove old points from a Vec before pushing new ones. The problem is that remove(0) and drain have a time complexity of O(n) where n is the length of vector.

Not discounting the solution - it might be the right one, I'd have to read more deeply to assess this, but I want to check that I'm not missing something first. Given that Chart already accepts a slice, are you able to provide a different slice of your data vec instead of changing the Vec? How does the internal O(1) actions inside the vector compare to the problem here (is your data just much larger than what's displayed?0

Perhaps what you mean is that a performant data structure for your data doesn't support passing a slice. It looks like the ringbuff will spit out two slices not one from https://docs.rs/ringbuf/0.3.3/ringbuf/ring_buffer/trait.Rb.html#method.as_slices. Is using two datasets on Chart a solution that would work here?

Just as a heads up if you are using the chart widget there's a weird bug that we haven't fixed with where the axis labels print. https://github.com/ratatui-org/ratatui/issues/334.

In the next release, we make it possible to render widgets by reference, which could mean that you don't need to call the functions that recreate the data for the chart so often - only when the data changes. Would this have an impact on any perf issues related to this change?

I'll take a bit more look at the details of the PR soon.

joshka:
On my machine the ringbuf benchmark is consistently the slowest. Can you explain the benchmarks here a bit.

image

Regardless, 50 micro seconds = 20,000 fps. As a performance optimization this doesn't seem to have any reasonable real world impact if those are realistic data sizes, so I'd like to suggest that we reframe this to a narrower scope of whether the API is ergonomic for your use case. Can you touch on that a bit more perhaps?

PR #903, opened 2024-02-01
feat: add WidgetRef and StatefulWidgetRef traits

joshka:
The Widget trait consumes self, which makes it impossible to use in a
boxed context. Previously we implemented the Widget trait for &T, but
this was not enough to render a boxed widget. We now have a new trait
called `WidgetRef` that allows rendering a widget by reference. This
trait is useful when you want to store a reference to one or more
widgets and render them later. Additionaly this makes it possible to
render boxed widgets where the type is not known at compile time (e.g.
in a composite layout with multiple panes of different types).

This change also adds a new trait called `StatefulWidgetRef` which is
the stateful equivalent of `WidgetRef`.

Both new traits are gated behind the `unstable-widget-ref` feature flag
as we may change the exact name / approach a little on this based on
further discussion.

Blanket implementation of `Widget` for `&W` where `W` implements
`WidgetRef` and `StatefulWidget` for `&W` where `W` implements
`StatefulWidgetRef` is provided. This allows you to render a widget by
reference and a stateful widget by reference.

A blanket implementation of `WidgetRef` for `Option<W>` where `W`
implements `WidgetRef` is provided. This makes it easier to render
child widgets that are optional without the boilerplate of unwrapping
the option. Previously several widgets implemented this manually. This
commits expands the pattern to apply to all widgets.

```rust
struct Parent {
    child: Option<Child>,
}

impl WidgetRef for Parent {
        self.child.render_ref(area, buf);
    }
}
```

```rust
let widgets: Vec<Box<dyn WidgetRef>> = vec![Box::new(Greeting), Box::new(Farewell)];
for widget in widgets {
    widget.render_ref(buf.area, &mut buf);
}
assert_eq!(buf, Buffer::with_lines(["Hello        Goodbye"]));
```

### DIFF
--- a/src/terminal/frame.rs
+++ b/src/terminal/frame.rs
@@ -91,6 +91,7 @@ impl Frame<'_> {
     /// let area = Rect::new(0, 0, 5, 5);
     /// frame.render_widget_ref(block, area);
     /// ```
+    #[stability::unstable(feature = "widget-ref")]
     pub fn render_widget_ref<W: WidgetRef>(&mut self, widget: W, area: Rect) {
         widget.render_ref(area, self.buffer);
     }
@@ -117,6 +118,7 @@ impl Frame<'_> {
     /// ```
     ///
     /// [`Layout`]: crate::layout::Layout
+    #[stability::unstable(feature = "widget-ref")]
     pub fn render_stateful_widget<W>(&mut self, widget: W, area: Rect, state: &mut W::State)
     where
         W: StatefulWidget,
--- a/src/terminal/frame.rs
+++ b/src/terminal/frame.rs
@@ -1,4 +1,4 @@
-use crate::prelude::*;
+use crate::{prelude::*, widgets::WidgetRef};
 
 /// A consistent view into the terminal state for rendering a single frame.
 ///
@@ -75,6 +75,26 @@ impl Frame<'_> {
         widget.render(area, self.buffer);
     }
 
+    /// Render a [`WidgetRef`] to the current buffer using [`WidgetRef::render_ref`].
+    ///
+    /// Usually the area argument is the size of the current frame or a sub-area of the current
+    /// frame (which can be obtained using [`Layout`] to split the total area).
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// # use ratatui::{backend::TestBackend, prelude::*, widgets::Block};
+    /// # let backend = TestBackend::new(5, 5);
+    /// # let mut terminal = Terminal::new(backend).unwrap();
+    /// # let mut frame = terminal.get_frame();
+    /// let block = Block::default();
+    /// let area = Rect::new(0, 0, 5, 5);
+    /// frame.render_widget_ref(block, area);
+    /// ```
+    pub fn render_widget_ref<W: WidgetRef>(&mut self, widget: W, area: Rect) {
+        widget.render_ref(area, self.buffer);
+    }
+
     /// Render a [`StatefulWidget`] to the current buffer using [`StatefulWidget::render`].
     ///
     /// Usually the area argument is the size of the current frame or a sub-area of the current
@@ -83,7 +103,7 @@ impl Frame<'_> {
     /// The last argument should be an instance of the [`StatefulWidget::State`] associated to the
     /// given [`StatefulWidget`].
     ///
-    /// # Examples
+    /// # Example
     ///
     /// ```rust
     /// # use ratatui::{backend::TestBackend, prelude::*, widgets::*};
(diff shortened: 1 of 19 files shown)


joshka:
An observant person might notice that we moved away from needing a new trait for this in https://github.com/ratatui-org/ratatui/pull/833. Unfortunately there doesn't seem to be a way to call a method on a trait that consumes the value from within a box. Therefore we need these traits to make this possible.

src/widgets.rs
```diff
@@ -227,3 +216,323 @@ pub trait StatefulWidget {
...
+///
+/// This trait is useful when you want to store a reference to a widget and render it later. It also
+/// allows you to render boxed widgets.
+///
+/// Boxed widgets allow you to store widgets with a type that is not known at compile time. This is
+/// useful when you want to store a collection of widgets with different types. You can then iterate
+/// over the collection and render each widget.
+///
+/// This trait was introduced in Ratatui 0.26.0 and is implemented for all the internal widgets.
+/// Implemetors should prefer to implement this over the `Widget` trait and add an implementation of
```
kdheepak, line 230:
```suggestion
/// Implementors should prefer to implement this over the `Widget` trait and add an implementation of
```

src/widgets.rs
```diff
@@ -227,3 +216,323 @@ pub trait StatefulWidget {
...
+/// // a collection of widgets with different types
+/// let widgets: Vec<Box<dyn WidgetRef>> = vec![Box::new(greeting), Box::new(farewell)];
+/// for widget in widgets {
+///     widget.render_ref(area, buf);
+/// }
+/// # }
+/// ```
+pub trait WidgetRef {
+    fn render_ref(&self, area: Rect, buf: &mut Buffer);
+}
```
kdheepak, line 291: Can we make `WidgetRef` `pub(crate)` only for a version or `unstable`? It'll help understand how this is going to be used, at least for me. 

The docstring also recommends implements use this instead of `Widget`, and I'm worried I have not explored all the facets of this implementation. I just don't understand the implications of this enough to have lots of confidence in this approach.

But I can tell it'll be useful for `Box` and potentially widget lists etc. So if other team members are on board with this, I'm too!

kdheepak: Can you expand on why this doesn't work?

```rust
impl Widget for Box<Block<'_>> {
        self.as_ref().render(area, buf);
    }
}
```

I added this test:

```rust
    #[test]
    fn box_block() {
        let mut buffer = Buffer::empty(Rect::new(0, 0, 4, 2));
        let block = Box::new(
            Block::default()
                .title("test")
                .title_position(Position::Bottom),
        );
        block.render(buffer.area, &mut buffer);
        assert_buffer_eq!(buffer, Buffer::with_lines(vec!["    ", "test"]));
    }
```

And I was able to compile and run it:

image

joshka: > Can we make WidgetRef pub(crate) only for a version or unstable? It'll help understand how this is going to be used, at least for me.

Yep that will work. I added this to the release highlights PR, but I can remove it.

> Can you expand on why this doesn't work?

```rust
impl Widget for Box<Block<'_>> {
        self.as_ref().render(area, buf);
    }
}
```

Sure, lets start with the following example:

```rust
struct WidgetInABox;

impl Widget for WidgetInABox {
        Line::raw("Hello").render(area, buf);
    }
}

impl Widget for &WidgetInABox {
        Line::raw("Hello").render(area, buf);
    }
}

impl Widget for Box<WidgetInABox> {
        self.as_ref().render(area, buf);
    }
}
```

Remove `impl Widget for &WidgetInABox` and the `self.as_ref().render(area, buf);` line does not compile:

```rust
error[E0507]: cannot move out of a shared reference
   --> src/widgets.rs:555:13
    |
555 |             self.as_ref().render(area, buf);
    |             ^^^^^^^^^^^^^ ----------------- value moved due to this method call
    |             |
    |             move occurs because value has type `WidgetInABox`, which does not implement the `Copy` trait
    |
note: `widgets::Widget::render` takes ownership of the receiver `self`, which moves value
   --> src/widgets.rs:101:15
    |
101 |     fn render(self, area: Rect, buf: &mut Buffer);
    |               ^^^^
```
Ok, so we know that that render call has to be on &WidgetInABox not &Widget.
Adding it back in compiles fine, and the following test passes.
```rust
#[rstest]
fn widget_box_render(mut buf: Buffer) {
    let widget = Box::new(WidgetInABox);
    widget.render(buf.area, &mut buf);
    assert_eq!(buf, Buffer::with_lines(["Hello               "]));
}
```
Adding it back in compiles fine, but it's ultimately useless because this doesn't compli
But that's not actually what we want. Widget there is `Box<WidgetInABox>` we want `Box<dyn Widget>` so that we can insert any widget. The following test fails to compile:
```rust
#[rstest]
fn widget_box_dyn_render(mut buf: Buffer) {
    let widget: Box<dyn Widget> = Box::new(WidgetInABox);
    widget.render(buf.area, &mut buf);
    assert_eq!(buf, Buffer::with_lines(["Hello               "]));
}
```
```rust
error[E0161]: cannot move a value of type `dyn widgets::Widget`
   --> src/widgets.rs:569:9
    |
569 |         widget.render(buf.area, &mut buf);
    |         ^^^^^^ the size of `dyn widgets::Widget` cannot be statically determined
```
The only way that I've managed to come up with something workable is to add a new trait that accepts &self (hence WidgetRef). My level of understanding if definitely expert level on this, so there might be something I'm missing. I meant to post on the user's forum to see if there was some unknown approach, but had other things to do last night. Getting to it now.

joshka: Rust lang post:
https://users.rust-lang.org/t/calling-a-method-with-a-self-parameter-on-box-dyn-trait/106220

Rust playground simplification of the problem:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=93f750557d860ea6a10c5c714e7e81c2

joshka: Also posted it in our discord:
https://discord.com/channels/1070692720437383208/1070692720437383211/1202850480162607124

And the Rust Programming Langudage Discord:
https://discord.com/channels/273534239310479360/1202852180999151667

joshka: > Can we make WidgetRef pub(crate) only for a version or unstable? It'll help understand how this is going to be used, at least for me.

I've marked it as unstable - at first I was thinking about pushing back, but I think it makes sense to do that. I'll update the PR title to clarify this point.

src/terminal/frame.rs
```diff
@@ -97,13 +118,42 @@ impl Frame<'_> {
     /// ```
     ///
     /// [`Layout`]: crate::layout::Layout
+    #[stability::unstable(feature = "widget-ref")]
```
kdheepak, line 121: This should go on the `render_stateful_widget_ref` method instead, right?

joshka: yep

kdheepak (approved):

joshka:
There are two solutions in the rust forum for this which are worth following up on. As this is marked unstable, it's not a blocker to merge as-is though and fix this after the 0.26 release.
- https://users.rust-lang.org/t/calling-a-method-with-a-self-parameter-on-box-dyn-trait/106220

PR #904, opened 2024-02-02
feat: replace Rect::split with Layout::areas and spacers

joshka:
In a recent commit we added Rec::split, but this feels more ergonomic as
Layout::areas. We also add Layout::spacers to get the spacers between
the areas.

### DIFF
--- a/examples/barchart.rs
+++ b/examples/barchart.rs
@@ -150,10 +150,10 @@ fn run_app<B: Backend>(
 }
 
 fn ui(frame: &mut Frame, app: &App) {
-    let vertical = Layout::vertical([Constraint::Ratio(1, 3), Constraint::Ratio(2, 3)]);
-    let horizontal = Layout::horizontal([Constraint::Percentage(50), Constraint::Percentage(50)]);
-    let [top, bottom] = frame.size().split(&vertical);
-    let [left, right] = bottom.split(&horizontal);
+    let vertical = &Layout::vertical([Constraint::Ratio(1, 3), Constraint::Ratio(2, 3)]);
+    let horizontal = &Layout::horizontal([Constraint::Percentage(50), Constraint::Percentage(50)]);
+    let [top, bottom] = vertical.areas(frame.size());
+    let [left, right] = horizontal.areas(bottom);
 
     let barchart = BarChart::default()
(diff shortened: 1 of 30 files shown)


examples/barchart.rs
```diff
@@ -150,10 +150,10 @@ fn run_app<B: Backend>(
 }
 
 fn ui(frame: &mut Frame, app: &App) {
-    let vertical = Layout::vertical([Constraint::Ratio(1, 3), Constraint::Ratio(2, 3)]);
-    let horizontal = Layout::horizontal([Constraint::Percentage(50), Constraint::Percentage(50)]);
-    let [top, bottom] = frame.size().split(&vertical);
-    let [left, right] = bottom.split(&horizontal);
+    let vertical = &Layout::vertical([Constraint::Ratio(1, 3), Constraint::Ratio(2, 3)]);
+    let horizontal = &Layout::horizontal([Constraint::Percentage(50), Constraint::Percentage(50)]);
```
kdheepak, line 154: do these _need_ a reference?

src/layout/layout.rs
```diff
@@ -420,6 +420,73 @@ impl Layout {
...
+    /// let [top, main] = layout.areas(area);
+    ///
+    /// // or explicitly specify the number of constraints:
+    /// let areas = layout.areas::<2>(area);
+    /// # }
+    pub fn areas<const N: usize>(&self, area: Rect) -> [Rect; N] {
+        self.split(area)
+            .to_vec()
+            .try_into()
+            .expect("invalid number of rects")
```
kdheepak, line 451: Maybe like this for consistency with `.spacers()`?

```suggestion
        let (areas, _) = self.split_with_spacers(area);
        areas
            .to_vec()
            .try_into()
            .expect("invalid number of rects")
```

kdheepak (approved):
LGTM!

kdheepak:
I like how it looks now more!

examples/barchart.rs
joshka: will fix

src/layout/layout.rs
joshka: makes sense

PR #912, opened 2024-02-03
feat(Constraint): implement minmax

### LINKED ISSUE #899: Add `Constraint::MinMax(u16, u16)`
## Problem

With the new constraints, `Min(u16)` tries to grow as much as possible and `Max(u16)` tries to grow to the size the user provided. This makes a `MinMax(u16, u16)` constraint viable where it can limit the allocated space between the minimum and maximum values provided by the user.

The implementation is fairly straightforward and it behaves intuitively as it should, i.e. a combination of `Min` growing but limited to `Max` instead of all excess area.

This is useful for responsive layouts, and particularly useful for table columns where the constraints can be used to calculate table column widths internally.

## Alternatives

Instead of adding more features to the existing constraints we should work on taffy instead.

Eeelco:
Implemented MinMax Constraint. Added UTs for new functionality and updated examples.

Closes #899

### DIFF
--- a/src/layout/layout.rs
+++ b/src/layout/layout.rs
@@ -720,6 +720,16 @@ fn configure_constraints(
 ) -> Result<(), AddConstraintError> {
     for (&constraint, &element) in constraints.iter().zip(segments.iter()) {
         match constraint {
+            Constraint::MinMax(min, max) => {
+                solver.add_constraint(element.has_min_size(min, MIN_SIZE_GE))?;
+                solver.add_constraint(element.has_max_size(max, MAX_SIZE_LE))?;
+                if flex.is_legacy() {
+                    solver.add_constraint(element.has_int_size(min, MIN_SIZE_EQ))?;
+                } else {
+                    solver.add_constraint(element.has_size(area, FILL_GROW))?;
+                }
+                solver.add_constraint(element.has_int_size(max, MAX_SIZE_EQ))?;
+            }
             Constraint::Max(max) => {
                 solver.add_constraint(element.has_max_size(max, MAX_SIZE_LE))?;
                 solver.add_constraint(element.has_int_size(max, MAX_SIZE_EQ))?;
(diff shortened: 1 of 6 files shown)


src/layout/layout.rs
```diff
@@ -720,6 +720,16 @@ fn configure_constraints(
 ) -> Result<(), AddConstraintError> {
     for (&constraint, &element) in constraints.iter().zip(segments.iter()) {
         match constraint {
+            Constraint::MinMax(min, max) => {
+                solver.add_constraint(element.has_min_size(min, MIN_SIZE_GE))?;
+                solver.add_constraint(element.has_max_size(max, MAX_SIZE_LE))?;
+                if flex.is_legacy() {
+                    solver.add_constraint(element.has_int_size(min, MIN_SIZE_EQ))?;
```
kdheepak, line 727: imo we don't need to special case this one because `MinMax` didn't exist before, so it there's no need to implement it the way `Min` behaves in `Flex::Legacy`.

Additionally, these two constraints will have the opposite effect, i.e. try to make the size `min` versus try to make the size `max`. Meaning only one of them can succeed. I'm not sure if that's what we want.

```
solver.add_constraint(element.has_int_size(min, MIN_SIZE_EQ))?;
solver.add_constraint(element.has_int_size(max, MAX_SIZE_EQ))?;
```

With the inclusion of this constraint, I believe it is always going to favor making the size `max`:

```
solver.add_constraint(element.has_size(area, FILL_GROW))?;
```

I _think_ we want `MinMax` to behave this way, so we should probably drop this line:

```
solver.add_constraint(element.has_int_size(min, MIN_SIZE_EQ))?;
```

fwiw, I'm not asking you to make these changes right now, just commenting for posterity. I think we want to use the flex modes and experiment with them to see what gaps they are and see how they can be filled. Because it might be possible we need to add more than just a `MinMax` constraint.

kdheepak:
Thanks for working on this PR! Seems like you were able to get the hang of the implementation pretty easily which is great!

For now, I'm going to make it as a draft PR (not because of your implementation btw but) because we probably want to experiment with the flex layouts to see what gaps they are and how they can be filled with new constraints. `MinMax` might be just one of the new constraints needed and it would be nice to bundle these changes together. Does that make sense?

PR #917, opened 2024-02-04
fix(canvas): compute circle coordinates with Pyth thm

SOF3:
The original algorithm may cause holes when the circle is large

discussed in https://github.com/SOF3/plotters-ratatui-backend/pull/2/files#r1461295810

The test case changes by one pixel, but the output was very approximate anyway

### DIFF
--- a/src/widgets/canvas/circle.rs
+++ b/src/widgets/canvas/circle.rs
@@ -1,3 +1,5 @@
+use std::{convert, mem};
+
 use crate::{
     style::Color,
     widgets::canvas::{Painter, Shape},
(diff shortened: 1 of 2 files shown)


src/widgets/canvas/circle.rs
```diff
@@ -18,12 +20,44 @@ pub struct Circle {
 
 impl Shape for Circle {
     fn draw(&self, painter: &mut Painter<'_, '_>) {
-        for angle in 0..360 {
-            let radians = f64::from(angle).to_radians();
```
kdheepak, line 22: I like the idea of increasing the resolution for larger circles. Can you elaborate on why something like this wouldn't work?

```suggestion
        let scaling_factor = 100
        for angle in 0..360*scaling_factor {
            let radians = (f64::from(angle) / scaling_factor as f64).to_radians()
```

joshka (changes requested):
Ideally for a circle, every point is either one across or one up (or both) from the previous point. The circles we end up drawing here tend to be too boxy as they're filling multiple points per grid point.

It needs to work entirely in terms of grid points, which I think this change does, but it needs to go one step further:

The algorithm you're looking for for this is this one - https://en.wikipedia.org/wiki/Midpoint_circle_algorithm (you might have heard people refer to Bresenhams algorithm).

Instead of iterating over all the points in the area, it's most performant (and leads to good looking results) to iterate over the points that are in the circle. Start at the top point, iterate while moving either to the right or downwards or both only until you hit 45º and then mirror all those points on the x/y axis.

I suspect this should be an concise but understandable ~25 LoC (which is a large difference from 159 LoC here).

SOF3:
The algorithm I am using here is, for each horizontal line (with the same `y`), identify the pixels that the circle intersects with, and color those points only. I have not looked into how different it is from the midpoint circle algorithm in terms of performance and accuracy, but the core implementation is just the 40 lines in circle.rs

The other 100 lines of change are related to the translation between world and screen coordinates, which is in the canvas.rs `Painter` impls. I believe we would face a similar problem with midpoint circle algorithm since it needs to work with the screen coordinates rather than the world coordinates, which is not very well-supported in the original `Painter` API.

SOF3:
The algorithm I implemented computes the range of coordinates to be plotted on each line in constant time (except the use of `sqrt`), which should not cause any performance issues. As for the part about mirroring the rasterization for 8 times, I believe (I haven't read about the implementation details of your proposed algorithm yet) that it is inappropriate for the case here since the center coordinates are not necessarily perfectly aligned with the screen coordinates, i.e. the result drawn by the ideal algorithm is not necessarily symmetric on any plane, and thus should not be mirrored directly.

src/widgets/canvas/circle.rs
SOF3: How do you suggest to come up with an ideal scaling factor? Even with a scaling factor that equals the number of pixels we need to draw, it is still possible that the circle line just intersects with one of the pixels at the corner, which can still be skipped.

SOF3:
Another possible motivation for computing the range of X pixels on each horizontal line is that the implementation in #920 would be simpler &mdash; non-filled circle draws `-dx_end..-dx_start, dx_start..dx_end`, while filled circle draws `-dx_end..dx_end` directly, such that the same algorithm can be largely reused.

src/widgets/canvas/circle.rs
kdheepak: fwiw I haven’t thought about this problem deeply, was just curious why the scaling factor approach wouldn’t work. The only solution I have to your question is that we can make it user configurable.

SOF3: yeah, well I just don't know what value to put there and leaving it to the user sounds weird

joshka:

joshka:

src/widgets/canvas/circle.rs
```diff
@@ -43,27 +134,315 @@ mod tests {
...
+            " █               █  ",
+            "█                 █ ",
+            "█                 █ ",
+            " █               █  ",
+            "  █             █   ",
+            "   ██         ██    ",
+            "     █████████      ",
+        ]);
+        assert_eq!(buffer, expected);
+    }
```
joshka, line 229: currently fails (draws an actual circle half the width).

PR #919, opened 2024-02-04
feat(canvas)!: add fill property to Rectangle

SOF3:
BREAKING CHANGE: `Rectangle` now has a `fill` field. Set as `false` for
original behavior.

Carries changes from #917


joshka:

PR #920, opened 2024-02-04
feat(canvas)!: add fill property to Circle

SOF3:
BREAKING CHANGE: `Circle` now has a fill field. Set as false for original behavior.

Includes changes from https://github.com/ratatui-org/ratatui/pull/917 to reuse the step_points util


joshka:

PR #924, opened 2024-02-05
fix(list): highlight symbol when using a  multi-bytes char

mrjackwills:
ratatui v0.26.0 brought a regression in the List widget, in which the highlight symbol width was incorrectly calculated - specifically when the highlight symbol was a multi-char character, e.g. `▶`.

I have also added a single additional test to check the output when such a multi-byte char is used for the highlight symbol.

### DIFF
--- a/tests/widgets_list.rs
+++ b/tests/widgets_list.rs
@@ -53,6 +53,38 @@ fn widgets_list_should_highlight_the_selected_item() {
 }
 
 #[test]
+fn widgets_list_should_highlight_the_selected_item_wide_symbol() {
+    let backend = TestBackend::new(10, 3);
+    let mut terminal = Terminal::new(backend).unwrap();
+    let mut state = ListState::default();
+
+    let wide_symbol = "▶  ";
+
+    state.select(Some(1));
+    terminal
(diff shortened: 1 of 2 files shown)


tests/widgets_list.rs
```diff
@@ -52,6 +52,38 @@ fn widgets_list_should_highlight_the_selected_item() {
...
+            f.render_stateful_widget(list, size, &mut state);
+        })
+        .unwrap();
+
+    let mut expected = Buffer::with_lines(vec!["   Item 1 ", "▶  Item 2 ", "   Item 3 "]);
+
+    for x in 0..10 {
+        expected.get_mut(x, 1).set_bg(Color::Yellow);
+    }
+    terminal.backend().assert_buffer(&expected);
```
kdheepak, line 84: For context, this is what this test would have done before:

```plain
Buffers are not equal
Expected:
"   Item 1 "
"▶  Item 2 "
"   Item 3 "

Got:
"     Item "
"▶    Item "
"     Item "
```

kdheepak (approved):

kdheepak:
Nice find!

PR #927, opened 2024-02-06
refactor(rect): move iters to module and add docs

joshka:


src/layout/rect.rs
```diff
@@ -26,56 +27,17 @@ pub struct Rect {
...
-        Some(column)
-    }
+/// See [`Rect::offset`]
+#[derive(Debug, Default, Clone, Copy)]
+pub struct Offset {
+    /// How much to move on the X axis
+    pub x: i32,
+    /// How much to move on the Y axis
+    pub y: i32,
 }
```
joshka, line 41: this is small enough not to really make sense to put in a separate module (I think)

kdheepak (approved):

PR #928, opened 2024-02-06
feat(rect): add Rect::positions iterator

joshka:
Useful for performing some action on all the cells in a particular area. E.g.,

```rust
fn render(area: Rect, buf: &mut Buffer) {
   for position in area.positions() {
        buf.get_mut(position.x, position.y).set_symbol("x");
    }
}
```

### DIFF
--- a/src/layout/rect.rs
+++ b/src/layout/rect.rs
@@ -265,6 +265,24 @@ impl Rect {
         Columns::new(self)
     }
 
+    /// An iterator over the positions within the `Rect`.
+    ///
+    /// The positions are returned in a row-major order (left-to-right, top-to-bottom).
+    ///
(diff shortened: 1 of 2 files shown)


kdheepak (approved):

kdheepak:
Is there value in iterating vertically and horizontally? I'm not sure what to call it but a `positions_row_major()` vs `positions_column_major()`?

joshka:
> Is there value in iterating vertically and horizontally? I'm not sure what to call it but the feature equivalent of `positions_row_major()` vs `positions_column_major()`?

Probably not, mainly because we're usually dealing with text, but `area.columns().flat_map(Area::positions)` gets you that in a one liner if needed.

PR #930, opened 2024-02-06
fix(line): render Line::styled correctly inside a paragraph

m4rch3n1ng:
rendering a `Paragraph` with a `Line::styled` ignores the style, but using styled spans with a `Line::default().spans(vec![Span::styled])` works just fine.

the issue seems to be, that the paragraph tries to render the `styled_graphemes` of the `Span`s instead of the `Line` directly

<details>
<summary>repro of the problem</summary>

```rs
fn ui(frame: &mut Frame) {
	let l1 = Line::styled("should be bold", Style::default().bold());
	let l2 = Line::styled("should be colorful", Style::default().cyan());
	let l3 = Line::styled("should be dim", Style::default().dim());

	let lp = Line::default();

	let s1 = vec![Span::styled("but it works with spans", Style::default().cyan())];
	let ls = Line::default().spans(s1);

	let block = Block::default()
		.borders(Borders::ALL)
		.padding(Padding::new(4, 4, 2, 2));
	let paragraph = Paragraph::new(vec![l1, l2, l3, lp, ls]).block(block);

	frame.render_widget(paragraph, frame.size());
}
```

</details>

### DIFF
--- a/src/widgets/paragraph.rs
+++ b/src/widgets/paragraph.rs
@@ -340,9 +340,7 @@ impl Paragraph<'_> {
         }
 
         let styled = self.text.iter().map(|line| {
-            let graphemes = line
-                .iter()
-                .flat_map(|span| span.styled_graphemes(self.style));
+            let graphemes = line.styled_graphemes(self.style);
             let alignment = line.alignment.unwrap_or(self.alignment);
             (graphemes, alignment)
         });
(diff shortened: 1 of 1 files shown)


kdheepak:
Thanks for making a PR! Can you add a test to make sure this fixes it?

m4rch3n1ng:
done, idk what's going on with codecov though

joshka (approved):
I think at some point we'll change this to style the entire line and not just the area covered by the spans, but this is fine as a bug fix.

PR #933, opened 2024-02-07
fix(block): fix crash on empty right aligned title

joshka:
- Simplified implementation of the rendering for block.
- Introduces a subtle rendering change where centered titles that are
  odd in length will now be rendered one character to the left compared
  to before. This aligns with other places that we render centered text
  and is a more consistent behavior. See
  https://github.com/ratatui-org/ratatui/pull/807#discussion_r1455645954
  for another example of this.

Fixes: 929

fix: tests

### DIFF
--- a/src/widgets/block.rs
+++ b/src/widgets/block.rs
@@ -554,7 +554,7 @@ impl Block<'_> {
     }
 
     fn render_title_position(&self, position: Position, area: Rect, buf: &mut Buffer) {
-        // Note: the order in which these functions are called defines the overlapping behavior
+        // NOTE: the order in which these functions are called defines the overlapping behavior
         self.render_right_titles(position, area, buf);
         self.render_center_titles(position, area, buf);
         self.render_left_titles(position, area, buf);
--- a/src/widgets/block.rs
+++ b/src/widgets/block.rs
@@ -5,6 +5,7 @@
 //! In its simplest form, a `Block` is a [border](Borders) around another widget. It can have a
 //! [title](Block::title) and [padding](Block::padding).
 
+use itertools::Itertools;
 use strum::{Display, EnumString};
 
 use crate::{prelude::*, symbols::border, widgets::Borders};
@@ -147,7 +148,6 @@ pub struct Block<'a> {
     titles_alignment: Alignment,
     /// The default position of the titles that don't have one
     titles_position: Position,
-
     /// Visible borders
     borders: Borders,
     /// Border style
@@ -525,9 +525,11 @@ impl Widget for Block<'_> {
 
 impl WidgetRef for Block<'_> {
     fn render_ref(&self, area: Rect, buf: &mut Buffer) {
+        let area = area.intersection(buf.area);
         if area.is_empty() {
             return;
         }
+        buf.set_style(area, self.style);
         self.render_borders(area, buf);
         self.render_titles(area, buf);
     }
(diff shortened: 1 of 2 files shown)


tests/widgets_block.rs
```diff
@@ -304,283 +198,379 @@ fn widgets_block_title_alignment() {
...
+        ]),
     );
 
     // title center with no left border
     test_case(
         Alignment::Center,
         Borders::TOP | Borders::RIGHT | Borders::BOTTOM,
-        Buffer::with_lines(vec![" ────Title───┐ ", " ────────────┘ "]),
+        Buffer::with_lines(vec![
+            " ───Title────┐ ",
```
joshka, line 293: This is the centering change:
```diff
-" ────Title───┐ "
+" ───Title────┐ "
```

tests/widgets_block.rs
```diff
@@ -589,125 +579,185 @@ fn widgets_block_multiple_titles() {
...
     );
 
     // title center with no left border
     test_case(
         Title::from("foo").alignment(Alignment::Center),
         Title::from("bar").alignment(Alignment::Center),
         Borders::TOP | Borders::RIGHT | Borders::BOTTOM,
-        Buffer::with_lines(vec![" ───foo─bar──┐ ", " ────────────┘ "]),
+        Buffer::with_lines(vec![
+            " ──foo─bar───┐ ",
```
joshka, line 674:
```diff
-" ───foo─bar──┐ "
+" ──foo─bar───┐ "
```

tests/widgets_block.rs
```diff
@@ -304,283 +198,379 @@ fn widgets_block_title_alignment() {
...
 
     // title center with no left border
     test_case(
         Alignment::Center,
         Borders::TOP | Borders::RIGHT | Borders::BOTTOM,
-        Buffer::with_lines(vec![" ────────────┐ ", " ────Title───┘ "]),
+        Buffer::with_lines(vec![
+            " ────────────┐ ",
+            "             │ ",
+            " ───Title────┘ ",
```
joshka, line 479:
```diff
-" ────Title───┘ "
+" ───Title────┘ "
```

joshka (commented):
Marked the 3 meaningful changes in the centering behavior.
All other test changes were to make the tests more readable and were checked before and after as being correct.

kdheepak:
LGTM! Let's mark it as a breaking change though.

kdheepak (approved):

src/widgets/block.rs
```diff
@@ -525,194 +525,238 @@ impl Widget for Block<'_> {
 
 impl WidgetRef for Block<'_> {
     fn render_ref(&self, area: Rect, buf: &mut Buffer) {
+        let area = area.intersection(buf.area);
```
kdheepak, line 528: Just curious, why is this needed?

src/widgets/block.rs
```diff
@@ -525,194 +525,238 @@ impl Widget for Block<'_> {
...
 
-        // Sides
-        if self.borders.intersects(Borders::LEFT) {
+    fn render_titles(&self, area: Rect, buf: &mut Buffer) {
+        self.render_title_position(Position::Top, area, buf);
+        self.render_title_position(Position::Bottom, area, buf);
+    }
+
+    fn render_title_position(&self, position: Position, area: Rect, buf: &mut Buffer) {
+        // Note: the order in which these functions are called defines the overlapping behavior
```
kdheepak, line 557: More standard convention for notes?

```suggestion
        // NOTE: the order in which these functions are called defines the overlapping behavior
```

I have an editor plugin that highlights when written like this.

src/widgets/block.rs
```diff
@@ -525,194 +525,238 @@ impl Widget for Block<'_> {
...
-        // Note: the order in which these functions are called define the overlapping behavior
-        self.render_right_titles(position, area, buf);
-        self.render_center_titles(position, area, buf);
-        self.render_left_titles(position, area, buf);
+    /// An iterator over the titles that match the position and alignment
+    fn filtered_titles(
+        &self,
+        position: Position,
+        alignment: Alignment,
+    ) -> impl DoubleEndedIterator<Item = &Title> {
```
kdheepak, line 736: Does the return type get converted to a `DoubleEndedIterator`?

src/widgets/block.rs
joshka: Prevents a panics for blocks out of the buffer

src/widgets/block.rs
joshka: Makes sense

src/widgets/block.rs
joshka: Needed for the .rev call in the right aligned titles. If it compiles then the return value implements the trait directly (no conversion necessary)

Valentin271 (approved):
I like the render refactor in this, way easier to read.

I wouldn't mark this as breaking change, not feeling strongly about it though, up to you.

kdheepak:
If a user has tests for title, their tests will break with no changes to their code. I think it'd be nice to let people know that the behavior has changed, for the better.

Valentin271:
If we decide it's breaking we should wait after the point release then.

kdheepak:
Hmm. Maybe it is not worth waiting that long? Sorry for going back on this, I didn't consider the release timeline. I think if we make it prominent in the changelog, I'm fine with calling it "not a breaking change" but a "bug fix" instead, and getting it in in this release.

Valentin271:
I don't feel strongly either way. This is technically breaking, but the old behavior was wrong so ... This so small no one will notice it (except unit tests). Might be better to include the fix as soon as possible to correct wrong behavior.

joshka:
I don't think rendering changes should be marked as breaking either from a perspective of recording in the the breaking changes doc or causing a major release. Every change we make changes the way that things render in some way.

PR #934, opened 2024-02-07
fix: use `to_string` to serialize Color

SleepySwords:
Since deserialize now uses `FromStr` to deserialize color, serializing `Color` RGB values, as well as index values, would produce an output that would no longer be able to be deserialized without causing an error.

Color::Rgb will now be serialized as the hex representation of their value.
For example, with serde_json, `Color::Rgb(255, 0, 255)` would be serialized as `"#FF00FF"` rather than `{"Rgb": [255, 0, 255]}`.

Color::Indexed will now be serialized as just the string of the index.
For example, with serde_json, `Color::Indexed(10)` would be serialized as `"10"` rather than `{"Indexed": 10}`.

Other color variants remain the same.


kdheepak:
I think you don't need this in Ratatui. This works for me to serialize and deserialize `Color`.

```rust
#[serde_as]
#[derive(Debug, Deserialize, Serialize)]
pub struct Config {
    #[serde_as(as = "DisplayFromStr")]
    pub background_color: ratatui::style::Color,
}
```

If I export an instance of this struct to toml, I get something like this:

```toml
[config]
background_color = "#111827"
```

Valentin271:
But it does make sense that a Ratatui color should be able to make a round trip. Plus `#ff00ff` makes much more sense to me (except in RON).

kdheepak:
For context, the code block I shared above does round trip, I'm able to read and write to toml and I am able to read `#ff00ff` strings and write out `#ff00ff` strings.

Valentin271:
> For context, the code block I shared above does round trip, I'm able to read and write to toml and I am able to read #ff00ff strings and write out #ff00ff strings.

One missing thing from my comment was probably "by default". Of course you could make it do that, but why not make it the default.

kdheepak:
Good point. One reason I can think of is that people may want to serialize into `hsl(0, 100%, 50%)` instead of `#ff0000`. I'm not sure if that would work if we provide a default (I'm fairly new to serde, so there may be some obvious way to do this that I don't know about).

joshka:
> Good point. One reason I can think of is that people may want to serialize into `hsl(0, 100%, 50%)` instead of `#ff0000`. I'm not sure if that would work if we provide a default (I'm fairly new to serde, so there may be some obvious way to do this that I don't know about).

My guess is that this would never be a default serialization format, only a deserialization. I could be wrong though, the only way I can think of to be able to make serialization work would be to keep the colors a `Color::Hsl` instead of converting to `Color::Rgb` during serialization. At that point, I'd advise the developer of such a problem to use a wrapper type or serialization helper that implements this and does the conversion for their app. This is because the only color schemes that consoles understand are indexed or RGB - not HSL.

orhun:
Yes, with this change we're serializing as hex instead of `Rgb { .. }` which looks like the correct behavior.

e.g. without this change the test fails like so:

```
thread 'style::color::tests::serialize_then_deserialize' panicked at src/style/color.rs:585:9:
  left: "{\"Rgb\":[255,0,255]}"
 right: "\"#FF00FF\""
```

orhun (approved):

joshka (changes requested):
How does this change existing serialized colors?
1. Can you write tests for the existing serialization format?
If it makes the previous serialization format fail, then this is a breaking change, which we should:
2.
a) work out whether we need to provide a serialization method that understands both formats (it's possible that we broke things when switching to using the `from_str` method - I'm not sure how many apps use serialization, it's likely a small amount but growing)
b) document the breaking change as such in the breaking changes doc

3. Can you please update the first comment in the PR to summarize the changes in behavior that users would notice (e.g. write a sucint message for the changelog).
(Something like Color::Indexed is now serialized as ..., Color::Rgb ...)

joshka:
(also, thanks for submitting your first PR to Ratatui!)

kdheepak:
I like hex as a default, I'm just not familiar enough with serde to know if supporting other formats would be possible after adding this default. For example, Base16 color schemes use `ff00ff` instead of `#ff00ff`, e.g.: 

- https://github.com/Defman21/base16-github-scheme/blob/master/github.yaml
- https://github.com/edunfelt/base16-rose-pine-scheme/blob/main/rose-pine.yaml

And I think it doesn't make sense to implement `from_str("ff00ff")`, right? How would one make a serialization and deserialization for these files into a struct containing `Color`s?

SleepySwords:
This test does fail

```rust
#[cfg(feature = "serde")]
#[test]
fn deserialize_with_previous_format() -> Result<(), serde_json::Error> {
    assert_eq!(Color::White, serde_json::from_str::<Color>("\"White\"")?);
    assert_eq!(
        Color::Rgb(255, 0, 255),
        serde_json::from_str::<Color>("{\"Rgb\":[255,0,255]}")?
    );
    assert_eq!(
        Color::Indexed(10),
        serde_json::from_str::<Color>("{\"Indexed\":10}")?
    );
    Ok(())
}
```

SleepySwords:
I have made the deserializer accept the existing serialized format. Are we still considering this a breaking change, since we changed how things are serialized?

src/style/color.rs
```diff
@@ -141,14 +140,44 @@ impl Color {
...
 
+#[cfg(feature = "serde")]
+impl serde::Serialize for Color {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        serializer.serialize_str(&self.to_string())
+    }
+}
```
joshka, line 151: Would it be worth adding docs to this to highlight that the serialization format is the same as the Display format?
This (or the deserialization code) would be a place where it's worth noting the compatibility with previous formats.

src/style/color.rs
```diff
@@ -141,14 +140,44 @@ impl Color {
...
     where
         D: serde::Deserializer<'de>,
     {
-        let s = String::deserialize(deserializer)?;
-        FromStr::from_str(&s).map_err(serde::de::Error::custom)
+        #[derive(serde::Deserialize)]
+        enum ColorWrapper {
+            Rgb(u8, u8, u8),
+            Indexed(u8),
+        }
```
joshka, line 163: It's worth documenting the intentions of this code both for end users and future maintainers who might wonder about the complexity of this method compared to deriving Deserialize directly.

E.g.:
```
/// Colors are serialized with ... because .... Previously they serialized as .... The deserialization implementation supports both formats for backwards compatibility. 
```

joshka (commented):
A couple of small things to add regarding comments, but otherwise good to go.

joshka:
> Are we still considering this a breaking change, since we changed how things are serialized?

Nope - this might break people with tests for exact serialization to strings, but that's really everything. I don't consider that a breaking change worth worrying about.

joshka (changes requested):
Generally approving the code, still a couple of minor things to get it over the line.

src/style/color.rs
```diff
@@ -141,14 +140,53 @@ impl Color {
...
+            String(String),
+            ColorWrapper(ColorWrapper),
+        }
+
+        let multi_type = Helper::deserialize(deserializer)
+            .map_err(|err| serde::de::Error::custom(format!("Failed to parse Colors: {err}")))?;
+        match multi_type {
+            Helper::String(s) => FromStr::from_str(&s).map_err(serde::de::Error::custom),
+            Helper::ColorWrapper(color_wrapper) => match color_wrapper {
+                ColorWrapper::Rgb(r, g, b) => Ok(Self::Rgb(r, g, b)),
```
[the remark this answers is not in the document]
EdJoPaTo: For reference, that lint rule ([`clippy::min_ident_chars`](http://rust-lang.github.io/rust-clippy/stable/index.html#/min_ident_chars)) is not part of the pedantic group, but it's still a good thing to use more readable variable names :innocent: 
Might be worth it to enable this lint in the future too.

src/style/color.rs
```diff
@@ -141,14 +140,53 @@ impl Color {
...
+    where
+        S: serde::Serializer,
+    {
+        serializer.serialize_str(&self.to_string())
+    }
+}
+
 #[cfg(feature = "serde")]
 impl<'de> serde::Deserialize<'de> for Color {
     fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
```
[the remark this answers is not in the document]
EdJoPaTo: I'm not sure doc comments work well on trait implemented methods. Should be checked.

Also, this seems like a good place for `assert_eq!(Color::Foo.to_string(), "bar")` examples in the doc comment as they are both readable and show how this is used.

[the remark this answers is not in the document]
joshka: > I'm not sure doc comments work well on trait implemented methods. Should be checked.

Doc comments on the implementation generally override the trait method's doc comments when rendered. This is usually useful for adding information about the implementation that is distinct from the trait.

When I first encountered this I was unsure too and had to check too.

There are lots of examples of this in the std library both at the trait doc comment level (on `impl Trait for Struct`) and at the method level. E.g. https://doc.rust-lang.org/std/string/struct.String.html#impl-Add%3C%26str%3E-for-String or https://doc.rust-lang.org/std/string/struct.String.html#method.from-10

SleepySwords:
Would it be fine if the value being deserialized is not a string or one of the old formats (for example passing an empty array), it would create this error `Failed to parse Colors: data did not match any variant of untagged enum Helper`?

src/style/color.rs
```diff
@@ -153,6 +153,43 @@ impl serde::Serialize for Color {
...
+    ///
+    /// ```
+    /// use ratatui::prelude::*;
+    ///
+    /// #[derive(serde::Deserialize)]
+    /// struct Theme {
+    ///     color: Color,
+    /// }
+    ///
+    /// let theme: Theme = serde_json::from_str::<Theme>("{\"color\": \"bright-white\"}").unwrap();
```
joshka, line 174: - Use a raw string to make this easier to read `r#"{ "color": "bright-white" }"#
- Avoid using unwrap by adding wrapping the example in a method that returns Result (this can be hidden from from the docs using `#`):
- Remove any unnecessary turbo fish / casts...

E.g.:

```rust
/// # fn get_theme() -> Result<(), ParseError> {
/// let theme: Theme = serde_json::from_str(r#"{ "color": "bright-white" }"#)?;
/// assert...
/// # Ok(())
/// # }
```

src/style/color.rs
```diff
@@ -153,6 +153,43 @@ impl serde::Serialize for Color {
...
+    /// assert_eq!(theme.color, Color::White);
+    ///
+    /// let theme: Theme = serde_json::from_str::<Theme>("{\"color\": \"#00FF00\"}").unwrap();
+    /// assert_eq!(theme.color, Color::Rgb(0, 255, 0));
+    ///
+    /// let theme: Theme = serde_json::from_str::<Theme>("{\"color\": \"42\"}").unwrap();
+    /// assert_eq!(theme.color, Color::Indexed(42));
+    ///
+    /// let theme: Result<Theme, _> = serde_json::from_str::<Theme>("{\"color\": \"invalid\"}");
+    /// assert!(theme.is_err());
```
joshka, line 184: I'd prefer assert_eq!(theme, Err(...)) here if that makes sense.

src/style/color.rs
```diff
@@ -153,6 +153,43 @@ impl serde::Serialize for Color {
 
 #[cfg(feature = "serde")]
 impl<'de> serde::Deserialize<'de> for Color {
+    /// This is used to deserialize a value into Color via serde.
+    ///
+    /// This implementation uses the `FromStr` trait to deserialize strings, so named colours, RGB,
+    /// and indexed values are able to be deserialized. In addition, values that were produced by
+    /// the the older serialization implementation of Color are also able to be deserialized.
```
joshka, line 160: Add some info about what the older serialization impl was. E.g. "Prior to 0.2x.y, Ratatui serialized Color as ..., but now it ... Both formats are able to be deserialized ..."

src/style/color.rs
```diff
@@ -141,14 +140,90 @@ impl Color {
...
+            Rgb(u8, u8, u8),
+            Indexed(u8),
+        }
+
+        #[derive(serde::Deserialize)]
+        #[serde(untagged)]
+        enum Helper {
+            String(String),
+            ColorWrapper(ColorWrapper),
+        }
```
joshka, line 216: > Would it be fine if the value being deserialized is not a string or one of the old formats (for example passing an empty array), it would create this error Failed to parse Colors: data did not match any variant of untagged enum Helper? Or should I rename Helper to Color?

If this ends up appearing in error messages, then it's worth thinking about how best to make the failure obvious to the user. "Helper" is purely related to the Color deserialization. Perhaps this should be ColorOrString or ColorVersionedSerialization, DeserializationFormat or something similar to those names (I'm sure you could think of a better name than those options). Perhaps even name the variants V1 / V2 or Version1 / Version2 to capture the intent rather than the implementation detail?

src/style/color.rs
EdJoPaTo: Probably annoying to achieve as the error is some serde_json error

SleepySwords: Yea the best I can come up with is 
```rust
let theme: Result<Theme, _> = serde_json::from_str(r#"{"color": "invalid"}"#);
assert!(theme.is_err());
if let Err(err) = theme {
    assert_eq!(
        err.to_string(),
        "Failed to parse Colors at line 1 column 20"
    )
}
```
Is this fine, should we leave it, or is there a better way to implement this?

joshka: Perhaps:
```rust
let err = serde_json::from_str::<Theme>(r#"{"color": "invalid"}"#).unwrap_err();
assert!(err.is_data());
assert_eq!(err.to_string(), "Failed to parse Colors");
```

joshka (approved):

joshka:
Thanks for fixing up the requested changes (and apologies in the delay in response). Merging this.

PR #935, opened 2024-02-07
feat(Barchart): reduced verbosity of barchart creation

### LINKED ISSUE #683: Improve Barchart creation for developers
## Problem
BarChart's construction is overly verbose. To construct anything other than the default value / text you need to do:

```rust
let data = BarGroup::default().bars(&[
        Bar::default()
            .label("Red".into())
            .value(2)
            .style(Style::new().red()),
        Bar::default()
            .label("Green".into())
            .value(7)
            .style(Style::new().green()),
        Bar::default()
            .label("Blue".into())
            .value(11)
            .style(Style::new().blue()),
    ]);
```

## Solution

- Bar should have a new() method that accepts a label and value `Bar::new("Red", 2)`
- Bar::label and the new `new` method should accept `Into<Line>` so we can write `Bar::default().label("foo")
`- Bar should implement styled so that style shorthands work `Bar::new("Red", 2").red()`
- BarGroup should have a new method that accepts I: IntoIterator, with I::Item = Into<Bar> so that `Bargroup::new([Bar::new("R", 2").red(), Bar::new("B", 3).blue()])` works
- BarGroup::label should accept Into<Line>
- BarGroup::bars should accept I:IntoIterator, with I::Item : Into<Bar>
- BarChart should have a new function that accepts bars (into iterator item = into bar) (unsure if this should also take direction - probably worth doing it as it highlights that direction is available better than just defaulting it)

+ Docs and unit tests for all of the above

wolfedme:
BREAKING_CHANGE: Due to introduction of IntoInterator in BarGroup, the bars parameter does not accept references anymore

Wasn't able to get it fully working without breaking changes, open for suggestions if that is an issue.

Resolves #683

### DIFF
--- a/benches/barchart.rs
+++ b/benches/barchart.rs
@@ -16,12 +16,12 @@ pub fn barchart(c: &mut Criterion) {
         let data: Vec<Bar> = (0..data_count)
             .map(|i| {
                 Bar::default()
-                    .label(format!("B{i}").into())
+                    .label(format!("B{i}"))
                     .value(rng.gen_range(0..data_count))
             })
             .collect();
 
-        let bargroup = BarGroup::default().bars(&data);
+        let bargroup = BarGroup::default().bars(data);
 
         // Render a basic barchart
         group.bench_with_input(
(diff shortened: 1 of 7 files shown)


wolfedme:
Close to resign commits...

PR #936, opened 2024-02-07
feat(barchart): reduce barchart creation verbosity

wolfedme:
BREAKING_CHANGE: Due to introduction of IntoInterator in BarGroup, the bars parameter does not accept references anymore

Wasn't able to get it fully working without breaking changes, open for suggestions if that is an issue.

Resolves https://github.com/ratatui-org/ratatui/issues/683

### DIFF
--- a/benches/barchart.rs
+++ b/benches/barchart.rs
@@ -16,12 +16,12 @@ pub fn barchart(c: &mut Criterion) {
         let data: Vec<Bar> = (0..data_count)
             .map(|i| {
                 Bar::default()
-                    .label(format!("B{i}").into())
+                    .label(format!("B{i}"))
                     .value(rng.gen_range(0..data_count))
             })
             .collect();
 
-        let bargroup = BarGroup::default().bars(&data);
+        let bargroup = BarGroup::default().bars(data);
 
         // Render a basic barchart
         group.bench_with_input(
--- a/src/widgets/barchart.rs
+++ b/src/widgets/barchart.rs
@@ -53,7 +53,7 @@ pub use bar_group::BarGroup;
 ///     .value_style(Style::new().red().bold())
 ///     .label_style(Style::new().white())
 ///     .data(&[("B0", 0), ("B1", 2), ("B2", 4), ("B3", 3)])
-///     .data(BarGroup::default().bars(&[Bar::default().value(10), Bar::default().value(20)]))
+///     .data(BarGroup::default().bars([Bar::default().value(10), Bar::default().value(20)]))
 ///     .max(4);
 /// ```
 #[derive(Debug, Clone, Eq, PartialEq, Hash)]
@@ -105,6 +105,31 @@ impl<'a> Default for BarChart<'a> {
 }
 
 impl<'a> BarChart<'a> {
+    /// Creates a new [`BarChart`] widget with the given bars.
+    ///
+    /// The `bars` parameter accepts an iterator containing items of type [`Bar`].
+    /// The `direction` parameter takes an enum value of [`Direction`].
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// # use ratatui::{prelude::*, widgets::*};
+    /// BarChart::new(
+    ///     [Bar::new("label", 10), Bar::new("label", 10)],
+    ///     Direction::Vertical,
+    /// );
+    /// ```
+    pub fn new<I>(bars: I, direction: Direction) -> Self
+    where
+        I: IntoIterator<Item = Bar<'a>>,
+    {
+        Self {
+            data: vec![BarGroup::new(bars)],
+            direction,
+            ..Default::default()
+        }
+    }
+
     /// Add group of bars to the BarChart
     ///
     /// # Examples
@@ -116,7 +141,7 @@ impl<'a> BarChart<'a> {
     /// # use ratatui::{prelude::*, widgets::*};
     /// BarChart::default()
     ///     .data(&[("B0", 0), ("B1", 2), ("B2", 4), ("B3", 3)])
-    ///     .data(BarGroup::default().bars(&[Bar::default().value(10), Bar::default().value(20)]));
+    ///     .data(BarGroup::default().bars([Bar::default().value(10), Bar::default().value(20)]));
     /// ```
     pub fn data(mut self, data: impl Into<BarGroup<'a>>) -> BarChart<'a> {
         let group: BarGroup = data.into();
@@ -825,9 +850,11 @@ mod tests {
     #[test]
     fn style() {
         let mut buffer = Buffer::empty(Rect::new(0, 0, 15, 3));
-        let widget = BarChart::default()
-            .data(&[("foo", 1), ("bar", 2)])
-            .style(Style::new().red());
+        let widget = BarChart::new(
+            [Bar::new("foo", 1), Bar::new("bar", 2)],
+            Direction::Vertical,
+        )
+        .red();
         widget.render(buffer.area, &mut buffer);
         let mut expected = Buffer::with_lines(vec![
             "  █            ",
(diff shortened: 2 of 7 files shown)


src/widgets/barchart/bar.rs
```diff
@@ -41,13 +41,35 @@ pub struct Bar<'a> {
...
+
     /// Set the value of this bar.
     ///
     /// The value will be displayed inside the bar.
     ///
     /// # See also
     ///
-    /// [`Bar::value_style`] to style the value.  
+    /// [`Bar::value_style`] to style the value.
     /// [`Bar::text_value`] to set the displayed value.
```
joshka, line 73: These were added to put the items on separate lines, but I dislike the end of line spaces as much as your formatter does. Can you please change these to a list using `-` (and below)

src/widgets/barchart/bar.rs
```diff
@@ -58,13 +80,16 @@ impl<'a> Bar<'a> {
...
-    /// display the label **under** the bar.  
+    /// display the label **under** the bar.
     /// For [`Horizontal`](crate::layout::Direction::Horizontal) bars,
-    /// display the label **in** the bar.  
+    /// display the label **in** the bar.
     /// See [`BarChart::direction`](crate::widgets::BarChart::direction) to set the direction.
     #[must_use = "method moves the value of self and returns the modified value"]
-    pub fn label(mut self, label: Line<'a>) -> Bar<'a> {
-        self.label = Some(label);
+    pub fn label<T>(mut self, label: T) -> Bar<'a>
```
joshka, line 88: Add a doc comment about conversion.

src/widgets/barchart/bar.rs
joshka, line 1: I think we currently only have `From` impls in `BarGroup`, are there missing ones that might make some of this more ergonomic? E.g. `From<(&'a str, u64)> for Bar` perhaps?

src/widgets/barchart/bar_group.rs
```diff
@@ -21,17 +21,44 @@ pub struct BarGroup<'a> {
 }
 
 impl<'a> BarGroup<'a> {
+    /// Creates a new [`BarGroup`] widget with the given bars and no label.
```
joshka, line 24: What's the rationale for preferring no label here?

src/widgets/barchart/bar_group.rs
```diff
@@ -21,17 +21,44 @@ pub struct BarGroup<'a> {
...
     }
 
     /// Set the bars of the group to be shown
     #[must_use = "method moves the value of self and returns the modified value"]
-    pub fn bars(mut self, bars: &[Bar<'a>]) -> BarGroup<'a> {
-        self.bars = bars.to_vec();
+    pub fn bars<I, T>(mut self, bars: I) -> BarGroup<'a>
+    where
+        I: IntoIterator<Item = T>,
+        T: Into<Bar<'a>>,
```
joshka, line 59:
```suggestion
    pub fn bars<Iter>(mut self, bars: Iter) -> BarGroup<'a>
    where
        Iter: IntoIterator
        Iter::Item: Into<Bar<'a>>,
```

src/widgets/barchart/bar_group.rs
```diff
@@ -21,17 +21,44 @@ pub struct BarGroup<'a> {
...
+    {
+        self.label = Some(label.into());
         self
     }
 
     /// Set the bars of the group to be shown
     #[must_use = "method moves the value of self and returns the modified value"]
-    pub fn bars(mut self, bars: &[Bar<'a>]) -> BarGroup<'a> {
-        self.bars = bars.to_vec();
+    pub fn bars<I, T>(mut self, bars: I) -> BarGroup<'a>
```
joshka, line 56: Add a note about conversion (applies generally to the new methods, won't repeat, but do a quick check for this)

joshka (commented):
Thanks - I like these changes. It's worth taking a look to understand what the impact on apps will be for the breaking changes here. I'm concerned that if they are large, then this becomes a barrier to upgrading between versions and leads to devs being annoyed about breaking changes.

For breaking changes we like to write these up in the BREAKING-CHANGES.md doc, and in the changelog (you can use the PR body to write up a message that will appear in the changelog).

As we've just released a major version 6 days ago, I'm going to suggest that we hold off introducing merging any breaking changes too soon (this will allow us to release a point release with non-breaking changes more easily), so bear in mind that this might not be released for a month or more.

The .into() part of this is easy as we just tell users to remove the unnecessary code, but the & part seems like it would cause pain (lots of changes in apps that use charts). Perhaps write some tests of the behavior to show which parts break. You might find that this leads to a novel solution that avoids breaking this part. (Focus on unit tests over integ tests btw).

wolfedme:
@joshka Really appreciating the suggestions and agreeing with all of them. I am completely fine for this not being released for a month+, as it might also take a week or two until I find the time to address every suggestion :)

src/widgets/barchart/bar.rs
EdJoPaTo, line 73: Suggestion to https://github.com/ratatui-org/ratatui/pull/936/files#r1482786822
```suggestion
    /// - [`Bar::value_style`] to style the value.
    /// - [`Bar::text_value`] to set the displayed value.
```

src/widgets/barchart/bar_group.rs
EdJoPaTo: Probably irrelevant with #995

src/widgets/barchart/bar.rs
EdJoPaTo: Probably irrelevant with #995

src/widgets/barchart.rs
```diff
@@ -105,6 +105,31 @@ impl<'a> Default for BarChart<'a> {
...
+    /// # Examples
+    ///
+    /// ```rust
+    /// # use ratatui::{prelude::*, widgets::*};
+    /// BarChart::new(
+    ///     [Bar::new("label", 10), Bar::new("label", 10)],
+    ///     Direction::Vertical,
+    /// );
+    /// ```
+    pub fn new<I>(bars: I, direction: Direction) -> Self
```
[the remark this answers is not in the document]
EdJoPaTo: Haven't looked whether these are already there, but `From<Bar> for BarGroup` and `From<Vec<Bar>> for BarGroup` should exist then.

src/widgets/barchart/bar.rs
```diff
@@ -41,13 +41,35 @@ pub struct Bar<'a> {
...
+    /// Bar::new("label", 10);
+    /// ```
+    pub fn new<L>(label: L, value: u64) -> Self
+    where
+        L: Into<Line<'a>>,
+    {
+        Self {
+            value,
+            label: Some(label.into()),
+            ..Default::default()
```
EdJoPaTo, line 62: I'd prefer being explicit here.
```suggestion
            style: Style::new(),
            value_style: Style::new(),
            text_value: None,
```

benches/barchart.rs
```diff
@@ -16,12 +16,12 @@ pub fn barchart(c: &mut Criterion) {
         let data: Vec<Bar> = (0..data_count)
             .map(|i| {
                 Bar::default()
-                    .label(format!("B{i}").into())
+                    .label(format!("B{i}"))
                     .value(rng.gen_range(0..data_count))
```
EdJoPaTo, line 20: All the examples should use the `::new` function instead of default in my opinion.

EdJoPaTo (changes requested):

EdJoPaTo:
Ha, great, GitHub applies some of my comments to both answers to @joshka and as new review comments… 😒

orhun:
Hey, this was partly implemented in other PRs (e.g. #1453) so we need to rebase this and apply some clean ups. Feel free to open up a new one with these changes if you are still interested. Closing for now.

Issue #738, opened 2024-01-05
Remove `block::Title` and `block::Position`, add `Block::top_title()` and `Block::bottom_title()` for ergonomics


joshka:
## Description
`block::Title` and `block::Position` sit weirdly in the API. They can't be included in the prelude as the names would clash pretty annoyingly with many other type names in common use in apps, but also Title does not actually need to exist.

Title has 3 fields - `content (Line)`, `alignment` and `position`. But `Line` already has an `alignment` field, so the title's alignment is superfluous. `Position` is a property of where the title is rendered, and so is a property of the `Block`, not the title itself. So if you remove the superfluous fields, that leaves `Title` just being a `Line`.

We should remove `Title` and `Position` from the API entirely.

## Suggested fixes
Transition to having just:
```rust
let block = Block::default()
    .title(Line::from("foo")) // keep for backwards compatibility - equivalent top_title
    .top_title(Line::from("foo").alignment(Alignment::Center))
    .bottom_title(Line::from("bar").alignment(Aligment::Right))
```

To do this nicely, we'd need to:
- add `top_title<T: Into<Line>>(line: T)` and similarly `bottom_title(...)`
- mark `Title as deprecated`
- mark `Title::position()` deprecated and point at `Block::top_title()` / `Block::bottom::title`
- mark `Position` as deprecated`
- change Block::titles to store top_titles and bottom_titles separately
- remove `title_on_bottom` (already deprecated)
- deprecate and then remove `Block::title_position`
- fix up the rendering methods

It's possible that we could add a `From<Title>` for Line implementation that drops the position and make existing code still work except for things with bottom titles for a while too.

It's a good idea to split this up into a few PRs:
- [ ] add the `top_` and `bottom_` fields and setters, update the rendering to use these instead of the current titles field (non breaking change)
- [ ] Implement `From<Title>` for Line 
- [ ] Mark all deprecations and write up a transition guide for the change in BREAKING changes
- [x] Remove existing deprecated setter
- [ ] Change `Block::title` to accept `Into<Line>` instead of `Into<Title>`
- [ ] in a later release remove the deprecated stuff.

The external impact of this is 77 files, so fairly small right now:
https://github.com/search?q=ratatui+AND+block+AND+%22Title%3A%3Afrom%22+AND+%28NOT+repo%3Aratatui-org%2Fratatui%29+AND+%28NOT+is%3Afork%29&type=code

kdheepak:
I like this idea! 

[...]

PR #940, opened 2024-02-09
feat: add Block::title_top and Block::title_top_bottom

joshka:
This adds the ability to add titles to the top and bottom of a block
without having to use the `Title` struct (which will be removed in a
future release - likely v0.28.0).

Fixes a subtle bug if the title was created from a right aligned Line
and was also right aligned. The title would be rendered one cell too far
to the right.

```rust
Block::bordered()
    .title_top(Line::raw("A").left_aligned())
    .title_top(Line::raw("B").centered())
    .title_top(Line::raw("C").right_aligned())
    .title_bottom(Line::raw("D").left_aligned())
    .title_bottom(Line::raw("E").centered())
    .title_bottom(Line::raw("F").right_aligned())
// renders
"┌A─────B─────C┐",
"│             │",
"└D─────E─────F┘",
```

Addresses part of https://github.com/ratatui-org/ratatui/issues/738

### DIFF
--- a/src/widgets/block.rs
+++ b/src/widgets/block.rs
@@ -22,6 +22,20 @@ pub use title::{Position, Title};
 /// [`Title`] using [`Block::title`]. It can also be [styled](Block::style) and
 /// [padded](Block::padding).
 ///
+/// You can call the title methods multiple times to add multiple titles. Each title will be
+/// rendered with a single space separating titles that are in the same position or alignment. When
+/// both centered and non-centered titles are rendered, the centered space is calculated based on
+/// the full width of the block, rather than the leftover width.
+///
+/// Titles are not rendered in the corners of the block unless there is no border on that edge.  
+/// If the block is too small and multiple titles overlap, the border may get cut off at a corner.
+///
+/// ```plain
+/// ┌With at least a left border───
+///
+/// Without left border───
+/// ```
+///
 /// # Examples
 ///
 /// ```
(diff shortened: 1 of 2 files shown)


kdheepak:
When should we mark `block::Title` as deprecated?

kdheepak (approved):

joshka:
It's a bit complicated. We would have to make everything that calls it work right, but probably before the next major release.

orhun (approved):

src/widgets/block.rs
```diff
@@ -228,6 +242,56 @@ impl<'a> Block<'a> {
...
+    /// Block::bordered()
+    ///     .title_top("Left1") // By default in the top left corner
+    ///     .title_top(Line::from("Left2").left_aligned())
+    ///     .title_top(Line::from("Right").right_aligned())
+    ///     .title_top(Line::from("Center").centered());
+    ///
+    /// // Renders
+    /// // ┌Left1─Left2───Center─────────Right┐
+    /// ```
+    pub fn title_top<T: Into<Line<'a>>>(mut self, title: T) -> Self {
```
Valentin271, line 264: `#[must_use]` are missing.

Valentin271 (dismissed):
> ```
> "┌A─────B──────C",
> "│             │",
> "└D─────E──────F",
> ```

Isn't it weird that C and F and all the way to the right (over the corners) but A and D keeps the corners? I naturally would expect this code to render all four corners the same way.

joshka:
> Isn't it weird that C and F and all the way to the right (over the corners) but A and D keeps the corners? I naturally would expect this code to render all four corners the same way.

This was a bug and I forgot to update the commit message

src/widgets/block.rs
joshka: fixed: debb696a...a0bd2b2c

PR #949, opened 2024-02-12
fix(List): saturating_sub to fix highlight_symbol overflow

mrjackwills:
An overflow (pedantically an underflow) can occur if the highlight_symbol is a multi-byte char, and area is reduced to a size less than that char length.

I am not sure of the best way to create a test for this, I only discovered it after resizing my ratatui window to as small as possible.

There *might* be a number of other potential overflows in ratatui, but I am not well versed enough in the source to know if they could ever actually occur. Basically anywhere that uses a `Rect` that calculates width by a subtraction, `x-y`, instead of using `x.saturating_sub(y)`

### DIFF
--- a/src/widgets/list.rs
+++ b/src/widgets/list.rs
@@ -973,7 +973,7 @@ impl StatefulWidgetRef for List<'_> {
                 let highlight_symbol_width = self.highlight_symbol.unwrap_or("").width() as u16;
                 Rect {
                     x: row_area.x + highlight_symbol_width,
-                    width: row_area.width - highlight_symbol_width,
+                    width: row_area.width.saturating_sub(highlight_symbol_width),
                     ..row_area
                 }
             } else {
@@ -1046,7 +1046,7 @@ mod tests {
     use std::borrow::Cow;
 
     use pretty_assertions::assert_eq;
-    use rstest::rstest;
+    use rstest::{fixture, rstest};
 
     use super::*;
     use crate::{assert_buffer_eq, widgets::Borders};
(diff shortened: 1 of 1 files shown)


joshka:
> I am not sure of the best way to create a test for this, I only discovered it after resizing my ratatui window to as small as possible.

Probably it would work to create a test that makes the highlight symbol wider than the area that you're rendering the list to

> There might be a number of other potential overflows in ratatui, but I am not well versed enough in the source to know if they could ever actually occur. Basically anywhere that uses a Rect that calculates width by a subtraction, x-y, instead of using x.saturating_sub(y)

Yup - there's definitely a few of these. We've gradually been adding bits and pieces to make it easier to do some of these calculations in ways that avoid overflows (e.g. using `Rect::rows()` / `Rect::columns()` for iterating instead of adding values to heights, or using `Rect::clamp()` to ensure that a rect cannot be moved outside a another one etc.)

mrjackwills:
That makes sense, I thought I had tried to write a test to do that, but I just realised that I had already implemented the `saturating_sub`, so of course it didn't overflow - even though I was looking for an overflow.

I'll update the PR with an overflow test, as I said there are some other places that maybe could do with a `saturating_sub`, but should I use a different PR for that?

joshka:
> That makes sense, I thought I had tried to write a test to do that, but I just realised that I had already implemented the saturating_sub, so of course it didn't overflow - hence my test was "failing" by passing.

I find a good way to handle something like that is always to show the output of the failing test before the change - regardless of the fix. Tests often have more value in their failing state than their passing state (the failing state helps you understand whether the test adequately explains the reason for failure, but the success state only shows that something succeeded).

> as I said there are some other places that maybe could do with a saturating_sub, but should I use a different PR for that?

Sure. Another PR for these would be good.

As a side note, I wonder if there's a simpler way to write the list render implementation in a way that makes it easier to understand at a glance how it works. The underlying problem that might have made it easy to sneak that bug in was that understanding the list rendering requires understanding every line of code rather than being able to understand smaller extracted portions.

tests/widgets_list.rs
joshka, line 1: Hey, I know that there's a bunch of tests in this file, but generally, let's try to prioritize unit tests instead of integration tests (i.e. tests in `list.rs` under mod tests instead of `tests/widgets_list.rs`).

For this problem the key elements of a test are:
- a list, with a highlight symbol
- a buffer
- render the list to the buffer in an area that is narrower than the highlight symbol
- check that the output matches the expected value
- this bug has no overlap with styling
- this bug has no overlap with anything related to terminal
- this bug has no overlap with anything related to the number of items in the list

In some of the more recent tests we've written, we've been taking advantage of using rstest. A test that hits those points and checks around the edge of this bug might look something like:

```rust
#[fixture]
fn single_line_buf() {
    Buffer::empty(Rect::new(0, 0, 10, 1))
}

/// Regression test for a bug where highlight symbol being greater than width caused a panic due to
/// subtraction with underflow.
///
/// See [#949](https://github.com/ratatui-org/ratatui/pull/949) for details
#[rstest]
#[case::under(">>>>", "Item1", ">>>>Item1 ")
#[case::exact(">>>>>", "Item1", ">>>>>Item1")
#[case::overflow(">>>>>>", "Item1", ">>>>>>Item")
fn highlight_symbol_overflow(#[case] highlight_symbol: &str, #[case] item, #[case] expected: &str, mut single_line_buf: Buffer) {
    let list = ...;
    list.render(single_line_buf.area, &mut single_line_buf);
    assert_buffer_eq!(single_line_bug, Buffer::with_lines(vec![expected]));
}
```

joshka (approved):
I'd prefer to see a test approach that is simpler and makes it more obvious what's being tested (suggested in a comment), but this is not blocking.

Valentin271 (approved):
Thanks for this. I agree with joshka, a unit test seems feasible here (but not blocking).

joshka (approved):

tests/widgets_list.rs
joshka: I pushed a change for this

PR #77, opened 2023-02-24
chore: Integrate `committed` for checking conventional commits

orhun:
See: https://github.com/crate-ci/committed
Configuration options: https://github.com/crate-ci/committed/blob/master/docs/reference.md
Closes #50

PR #950, opened 2024-02-12
chore: remove conventional commit check for PR

Valentin271:
This removes conventional commit check for PRs.

Since we use the PR title and description this is useless. It fails a lot of time and we ignore it.

IMPORTANT NOTE: This does **not** mean Ratatui abandons conventional commits. This only relates to commits in PRs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,13 +29,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        if: github.event_name != 'pull_request'
         uses: actions/checkout@v4
-      - name: Checkout
-        if: github.event_name == 'pull_request'
-        uses: actions/checkout@v4
-        with:
-          ref: ${{ github.event.pull_request.head.sha }}
       - name: Install Rust nightly
         uses: dtolnay/rust-toolchain@nightly
         with:
@@ -48,11 +42,6 @@ jobs:
         run: cargo make lint-format
       - name: Check documentation
         run: cargo make lint-docs
-      - name: Check conventional commits
-        uses: crate-ci/committed@master
-        with:
-          args: "-vv"
-          commits: HEAD
       - name: Check typos
         uses: crate-ci/typos@master
       - name: Lint dependencies


joshka:
Seems reasonable - how about also modifying the PR check a bit to help guide PR messages? Something like.

```
Thank you for opening this pull request! Here are some guidelines for writing a good PR message:

- We use the PR message as the default commit message that will be committed to the repository
- We automatically generate the [changelog](...link to changelog) using commits, so please use
  [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) to make this easy:
  - ... examples / specific parts of note
- Writing the perfect PR message is about finding a balance of communicating the problem and
  solution as well as not being too verbose.
- Put yourself in the mindset of the future consumer of this library and what information they need
  to know about this change (it might be very little or nothing beyond the title).
- It's rare that a PR message needs to explain how the change was achieved, and much more
  important to explain the impact on the user, and the rationale for any interesting choices
- If there's info that's useful for reviewing, but not useful for the changelog, move it to followup comment
- If there are changes made to the PR during review, make sure to also update the first PR comment
  (to make it easy to merge)
```

Perhaps this response should be on every PR?

Valentin271:
> - We use the PR message as the default commit message that will be committed to the repository
> - We automatically generate the [changelog](...link to changelog) using commits, so please use
>   [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) to make this easy:
>   - ... examples / specific parts of note

I think this specifically is a good message to put on PRs with non conventional commits titles.

The whole paragraph would make more sense to me in `CONTRIBUTING.md` or even better: in the PR template (perhaps as some sort of checklist?)

I think a comment with this on every PR is too much, but I'd be keen on putting it for every first-contributor PR.

joshka:
> EDIT: Also this is the matter of another PR imo

yep

I want another maintainer's opinion on this change before merging it btw.

joshka (approved):
Check with another maintainer before merging this.

kdheepak (approved):

.github/workflows/ci.yml
```diff
@@ -48,11 +48,6 @@ jobs:
         run: cargo make lint-format
       - name: Check documentation
         run: cargo make lint-docs
-      - name: Check conventional commits
```
orhun, line 51: We can also remove the conditional checkout above. (`github.event_name != 'pull_request'`)

See #77

orhun (approved):
Sad to see this go but sounds reasonable 👍🏼

Valentin271:
> Sad to see this go

Every commit on main is made through a PR, internal PR commits don't matter.

What would be great though is to check if the PR title matches the conventional commit style.

orhun:
yeah, I think we already do that in `check-pr.yml`

PR #951, opened 2024-02-13
perf(layout): const functions

EdJoPaTo:
The layout can often contain things that are already known at compile time which speeds up things on runtime. Especially as render is called a lot.

With 0.26.0 the `Rect::contains` method is now accepting a Position. While `contains` is `const`, `Position::new` is not which triggered the creation of this PR. I switched to `Position { x, y }`, but the compiler should be able to do the same in `const` with `::new()`.

I added the lint to the layout module, but I don't really know what other plans of this project are in regard to configuring lints. I could also just remove the lint there, but mistakes like the current one will happen again as the code uses a lot of `const` already. Do you have other thoughts on how to configure the lints best?

I observed the [missing_const_for_fn](https://rust-lang.github.io/rust-clippy/master/index.html#/missing_const_for_fn) clippy lint in other areas of the library too, but it's mostly pointless there. Some Widgets and WidgetStates could get `::new()` implementations which are `const` which would also allow for optimizations there.

Personally I also think that more `clippy::pedantic` and `clippy::nursery` lints could keep the code quality of ratatui consistent. Like in the current case many things are already `const`, but they are forgotten in other places as no lints are ensuring that.

### DIFF
--- a/src/layout.rs
+++ b/src/layout.rs
@@ -1,3 +1,5 @@
+#![warn(clippy::missing_const_for_fn)]
+
 mod alignment;
 mod constraint;
 mod corner;
--- a/src/layout/position.rs
+++ b/src/layout/position.rs
@@ -38,7 +38,7 @@ pub struct Position {
 
 impl Position {
     /// Create a new position
-    pub fn new(x: u16, y: u16) -> Self {
+    pub const fn new(x: u16, y: u16) -> Self {
         Position { x, y }
     }
 }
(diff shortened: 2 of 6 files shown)


joshka:
> Personally I also think that more clippy::pedantic and clippy::nursery lints could keep the code quality of ratatui consistent. Like in the current case many things are already const, but they are forgotten in other places as no lints are ensuring that.

+1 on these - the issue is that turning them on makes a lot of noise initially, so it's a bunch of work drilling down on the lint errors. Once it's in though things are really nice.

joshka (approved):

PR #952, opened 2024-02-13
feat: impl Widget for `&str` and `String`

kdheepak:
Currently, `f.render_widget("hello world".bold(), area)` works but `f.render_widget("hello world", area)` doesn't. This PR changes that my implementing `Widget` for `&str` and `String`. This makes it easier to render strings with no styles as widgets.

Example usage:

```rust
terminal.draw(|f| f.render_widget("Hello World!", f.size()))?;
```

### DIFF
--- a/src/widgets.rs
+++ b/src/widgets.rs
@@ -417,6 +417,30 @@ pub trait StatefulWidgetRef {
 //     }
 // }
 
+impl Widget for &str {
+    fn render(self, area: Rect, buf: &mut Buffer) {
+        self.render_ref(area, buf);
+    }
+}
+
+impl WidgetRef for &str {
+    fn render_ref(&self, area: Rect, buf: &mut Buffer) {
+        buf.set_string(area.x, area.y, self, crate::style::Style::default())
+    }
+}
+
+impl Widget for String {
+    fn render(self, area: Rect, buf: &mut Buffer) {
+        self.render_ref(area, buf);
+    }
+}
+
+impl WidgetRef for String {
+    fn render_ref(&self, area: Rect, buf: &mut Buffer) {
+        buf.set_string(area.x, area.y, self, crate::style::Style::default())
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use rstest::{fixture, rstest};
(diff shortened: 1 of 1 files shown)


joshka:
+ Tests and docs

joshka (commented):
Tests and docs needed.

src/widgets.rs
```diff
@@ -561,4 +606,42 @@ mod tests {
...
+
+    #[rstest]
+    fn str_render(mut buf: Buffer) {
+        let widget = "hello world";
+        widget.render_ref(buf.area, &mut buf);
+        assert_eq!(buf, Buffer::with_lines(["hello world         "]));
+
+        let widget = "hello world";
+        widget.render(buf.area, &mut buf);
+        assert_eq!(buf, Buffer::with_lines(["hello world         "]));
```
joshka, line 618: Rendering to the same buffer instead of a new one is a bad way to test as the first rendering already updates the buffer with the expected output. Split these into 4 tests or reconstruct the buffer on each branch of the test. That might be easier to read as just the following. It's probably also worth writing this as `"hello world".render_ref(...)` to be canonical with the intended usage.

```rust
let buf = &mut single_line_buffer();
"hello world!".render(buf.area, buf);
assert_eq!(buf, Buffer::with_lines(["hello world!        "]), "str.render()");

let buf = &mut single_line_buffer();
String::new("hello world!").render(buf.area, buf);
assert_eq!(buf, Buffer::with_lines(["hello world!        "]), "String.render()");
```

joshka (commented):
Need to add some sort of comment to the Widget trait or widget module for this.

src/widgets.rs
```diff
@@ -561,4 +607,78 @@ mod tests {
...
+
+    #[rstest]
+    fn str_render() {
+        let mut buf = single_line_buffer();
+        "hello world".render(buf.area, &mut buf);
+        assert_eq!(
+            buf,
+            Buffer::with_lines(["hello world         "]),
+            "str.render()"
+        );
```
joshka, line 623: nit - with this automatically formatting to 5 lines for the assertion, this would probably be better with actual methods (I think you end up with 1 less line).

joshka (approved):

orhun (approved):
LGTM

src/widgets.rs
kdheepak: can you clarify what you mean by "with actual methods"?

joshka: https://github.com/ratatui-org/ratatui/pull/952/commits/d430470d0c8fe8f26f19f60c2b1058c608d9307d

PR #958, opened 2024-02-17
feat(list): Add Scroll Padding to Lists

### LINKED ISSUE #955: Scroll Padding for lists
## Problem

I would like to be able to keep the next item in a list visible when scrolling through a list bigger than the viewable area to make large lists of items more understandable and pleasant to browse through. 

## Solution

Ideally the item before and after the currently selected item would remain visible while scrolling either up or down.
Or as many items before and after as the api user selects. 

We can store a padding value in the ListState struct and have that passed through to the bounds checking that the List performs. 
If we add serde(default) to the parameter in ListState we can avoid breaking changes from serialization. 

Using this new feature would be the same as existing code for manually setting the offset, so any of the following:
```
*state.padding_mut() = 1;
state = state.with_padding(1);
state = ListState::default().with_padding(1);
```
By default padding would be disabled and would have no impact on existing users without any action on their end. 

I've developed an [existing fork](https://github.com/CameronBarnes/ratatui_padding) that has this feature working and tested. I've so far been using the name "padding" but I think "scroll_padding" would be much less ambiguous and confusing. I'm open to suggestions. 

I've attached gif(s) of current and desired behavior bellow.

## Alternatives

Alternatively, I could just handle this myself in the api user side of the code. But it would be very difficult to handle items of inconsistent sizes without reimplementing much of the existing bounds checking code in List rendering. 
Additionally doing it that way could leave that user side code vulnerable to otherwise non-breaking changes that adjust how lists are rendered. Like the [existing PR](https://github.com/ratatui-org/ratatui/pull/150#issue-1684948724) for adding what I'd probably describe as list item padding. 

## Additional context

Current behavior:
no_padding
Desired behavior:
scroll_padding

CameronBarnes:
This PR introduces scroll padding, which allows the api user to request that a certain number of ListItems be kept visible above and below the currently selected item while scrolling. 

It does this by adjusting the 'index_to_display' value in the List::get_item_bounds function. Reducing the provided padding value if the padding would push the selected item out of the renderable area or if there's not enough room for the requested amount of padding, and correctly handling ListItems of inconsistent sizes. 

I have stored the 'padding' value in the ListState object. Using serde(default) on the parameter to avoid creating a breaking change in relation to user deserialization of existing data. 
I believe the ListState object is the correct place to put this, as if we're not rendering a stateful list then we dont need this value. 

'Padding' almost certainly isnt the correct name for this value, I think 'scroll_padding' is a better option and communicates the behavior to users better. But I'm open to suggestions. I havent yet changed the code to reflect the clearer 'scroll_padding' name. 

Using this new feature would be the same as existing code for manually setting the offset, so any of the following:
```
*state.padding_mut() = 1;
state = state.with_padding(1);
state = ListState::default().with_padding(1);
```
By default the 'padding' value is 0 and would have no impact on existing users without any action on their end. 

I've written unit tests that cover the original unmodified behavior, the new intended behavior, and several troublesome edge cases that came up while I was developing this but was able to resolve. 
Most of the code in this PR is either new test cases, or specifically for handling those edge cases.

Demo of the new behavior(visible on the left side of the gif):
scroll_padding

Comparison of modified and unmodified behavior is shown in the issue this PR was created for.
This PR resolves #955

### DIFF
--- a/src/widgets/list.rs
+++ b/src/widgets/list.rs
@@ -1087,6 +1087,7 @@ where
 mod tests {
     use std::borrow::Cow;
 
+    use pretty_assertions::assert_eq;
     use rstest::rstest;
 
     use super::*;
@@ -2181,8 +2182,10 @@ mod tests {
         let mut terminal = Terminal::new(backend).unwrap();
         let mut state = ListState::default();
 
+        let padding_value = 3;
+
         *state.offset_mut() = 2;
-        *state.padding_mut() = 3;
+        *state.padding_mut() = padding_value;
         state.select(Some(4));
 
         let items = vec![
@@ -2213,18 +2216,20 @@ mod tests {
             })
             .unwrap();
 
-        assert_eq!(offset_after_render, state.offset())
+        // Offset after rendering twice should remain the same as after once
+        assert_eq!(offset_after_render, state.offset());
+        // Padding value shouldnt be modified
+        assert_eq!(state.padding(), padding_value);
     }
 
     #[test]
     fn test_padding_inconsistent_item_sizes() {
         let backend = backend::TestBackend::new(10, 3);
         let mut terminal = Terminal::new(backend).unwrap();
-        let mut state = ListState::default();
-
-        *state.offset_mut() = 0;
-        *state.padding_mut() = 1;
-        state.select(Some(3));
+        let mut state = ListState::default()
+            .with_offset(0)
+            .with_padding(1)
+            .with_selected(Some(3));
 
         let items = vec![
             ListItem::new("Item 0"),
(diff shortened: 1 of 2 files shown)


CameronBarnes:
> I need more time to review the render because of its complexity but I think it could/should be delegated to a specific sub-method.

I'm not sure I understand what you mean by "delegated to a specific sub-method." Currently the logic exists entirely in the get_items_bounds function, as it needs access to all the same data and would nessecarily need to be updated if that function was updated, like in the case of adding list item padding where there's empty spaces between list items. 

> Also I think the padding field should be in `List` and not `ListState`. This is directly related to the render rather than state.

My thought process for having it stored in ListState rather than the List was that the value wasnt needed unless an item was selected. But I suppose selection highlighting works similarly. 
I'd be happy to make that change if you'd like. 

Additionally, I'd like some input on 'scroll_padding' as a prefered alternate name. I havent yet changed it in code as I wanted to get input from others so that it only had to be changed once. 

Thanks for your feedback.

Valentin271:
> I'm not sure I understand what you mean by "delegated to a specific sub-method." Currently the logic exists entirely in the get_items_bounds function, as it needs access to all the same data and would nessecarily need to be updated if that function was updated, like in the case of adding list item padding where there's empty spaces between list items.

I didn't pay enough attention to the modification and didn't see it was updating outside variables. I'm just trying to reduce the length and complexity of the method which is now 128 lines long.

> My thought process for having it stored in `ListState` rather than the `List` was that the value wasnt needed unless an item was selected. But I suppose selection highlighting works similarly.

That's a good point. But as highlighting works like that and others widget too, I prefer having it it `List`. It's more a visual parameter.

> I'd like some input on 'scroll_padding'

Right I forgot about that, thanks for the reminder. I do think it is a better name for this feature.

joshka:
> I'm just trying to reduce the length and complexity of the method which is now 128 lines long.

+1 to this. Making long methods more complex is definitely something to avoid.
It may be necessary to apply some simplifying refactoring either before this change or possibly after.

CameronBarnes:
Great news, it looks like two of the three checks for dealing with scroll_padding edge cases were unnessecary as the third check covers the cases of the previous two. This has been a consistent pattern in dealing with some of the tricky edge cases for this feature.
This has notably reduced the amount of code and complexity added. 

Additionally, the added code no longer has any side effects. So it could be refactored out to a separate method, though personally I think doing so might make the code harder to understand for someone viewing this later, rather than easier. But that's just my opinion. I'll move it to a separate function if requested.

Valentin271 (approved):
This looks good to me. No need for a separate function now that it's been simplified.

joshka:
> This looks good to me. No need for a separate function now that it's been simplified.

I prefer smaller methods generally as they tend to be easier to understand and modify when maintaining things later, so I'd like to see this new functionality extracted from the function if it doesn't lead to a silly parameter list. My rationale is that when methods have large blocks of code in them, those blocks have access and can modify any mutable variable of the function. By extracting this logic, it's easier to reason about the functionality, as the input is the parameters and the output is the result.

CameronBarnes:
I've moved the code to its own function. Though in this case I think this just hides implementation details rather than actually making it easier to understand the bounds checking code.

joshka (approved):
Sounds good. I'll slap a few opinionated cleanups in another PR

PR #959, opened 2024-02-18
fix: scrollbar thumb not visible on long lists

ThomasMiz:
## The problem
When displaying somewhat-long lists, the `Scrollbar` widget might not display a single thumb character, and only the track will be visible. Here are some GIFs showing the issue:
before_vertical
before_horizontal

The code to reproduce this issue was based off the scrollbar example and can be found in [this pastebin](https://pastebin.com/x4DzthcF).

## This PR
I fixed this issue by converting the logic in the `part_lengths` function to use integer calculations instead of floating-point (the only reason for the use of floating point seemed to be the rounding after division, which can be done in integers by replacing `(a / b).round() as usize` with `(a + b/2) / b`) and then doing a bit of clamping. The result of the previous test is as shown in these GIFs:
after_vertical
after_horizontal

### DIFF
--- a/src/widgets/scrollbar.rs
+++ b/src/widgets/scrollbar.rs
@@ -498,38 +498,32 @@ impl Scrollbar<'_> {
     ///
     /// This method returns the length of the start, thumb, and end as a tuple.
     fn part_lengths(&self, area: Rect, state: &mut ScrollbarState) -> (usize, usize, usize) {
-        const fn rounded_divide(a: usize, b: usize) -> usize {
-            (a + (b / 2)) / b
-        }
+        let track_length = self.track_length_excluding_arrow_heads(area) as f64;
+        let viewport_length = self.viewport_length(state, area) as f64;
 
-        let track_len = self.track_length_excluding_arrow_heads(area) as usize;
-        let viewport_len = self.viewport_length(state, area) as usize;
+        // Ensure that the position of the thumb is within the bounds of the content taking into
+        // account the content and viewport length. When the last line of the content is at the top
+        // of the viewport, the thumb should be at the bottom of the track.
+        let max_position = state.content_length.saturating_sub(1) as f64;
+        let start_position = (state.position as f64).clamp(0.0, max_position);
+        let max_viewport_position = max_position + viewport_length;
+        let end_position = start_position + viewport_length;
 
-        // Clamp the position to show at least one line of the content
-        // Note: state.content_length is != 0 because otherwise render() returns immediately
-        let position = state.position.min(state.content_length - 1);
+        // Calculate the start and end positions of the thumb. The size will be proportional to the
+        // viewport length compared to the total amount of possible visible rows.
+        let thumb_start = start_position * track_length / max_viewport_position;
+        let thumb_end = end_position * track_length / max_viewport_position;
 
-        // vscode style scrolling behavior (allow scrolling past end of content)
-        let scrollable_content_len = state.content_length + viewport_len - 1;
+        // Make sure that the thumb is at least 1 cell long by ensuring that the start of the thumb
+        // is less than the track_len. We use the positions instead of the sizes and use nearest
+        // integer instead of floor / ceil to avoid problems caused by rounding errors.
+        let thumb_start = thumb_start.round().clamp(0.0, track_length - 1.0) as usize;
+        let thumb_end = thumb_end.round().clamp(0.0, track_length) as usize;
 
-        // Calculate the thumb start (inclusive) and end (exclusive) positions, with rounding
-        let thumb_start = rounded_divide(position * track_len, scrollable_content_len);
-        let thumb_end = rounded_divide(
-            (position + viewport_len) * track_len,
-            scrollable_content_len,
-        );
+        let thumb_length = thumb_end.saturating_sub(thumb_start).max(1);
+        let track_end_length = (track_length as usize).saturating_sub(thumb_start + thumb_length);
 
-        // Ensure there is at least one thumb character by ensuring `thumb_start` has at least one
-        // space after it, and then ensuring `thumb_end` is at least one space after `thumb_start`
-        let thumb_start = thumb_start.min(track_len.saturating_sub(1));
-        let thumb_end = thumb_end.max(thumb_start + 1);
-
-        // Calculate the length of the tracks and thumb
-        let track_start_len = thumb_start;
-        let thumb_len = thumb_end - thumb_start;
-        let track_end_len = track_len - thumb_start - thumb_len;
-
-        (track_start_len, thumb_len, track_end_len)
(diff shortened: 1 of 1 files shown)


ThomasMiz:
A small thing to note about the pastebin code that should probably be fixed in the examples are lines 55-57:
```rust
if key.kind == KeyEventKind::Release {
    continue;
}
```

Otherwise, on Windows terminals, all events get registered twice

Valentin271:
Yep can confirm I reproduce. Though important to note this happens in large lists **and** somewhat small terminal sizes. My fullscreen terminal at `191x46` works fine but `80x27` does not.

ThomasMiz:
> Yep can confirm I reproduce. Though important to note this happens in large lists **and** somewhat small terminal sizes. My fullscreen terminal at `191x46` works fine but `80x27` does not.

My use case, the reason I found this, is because I have a list of up to 64k log events
image

src/widgets/scrollbar.rs
```diff
@@ -498,25 +498,30 @@ impl Scrollbar<'_> {
...
+            (position * track_len + scrollable_content_len / 2) / scrollable_content_len;
+        let thumb_start = thumb_start.min(track_len.saturating_sub(1));
+        let thumb_end = ((position + viewport_len) * track_len + scrollable_content_len / 2)
+            / scrollable_content_len;
+        let thumb_end = thumb_end.max(thumb_start + 1);
+
+        // Calculate the length of the tracks and thumb
+        let track_start_len = thumb_start;
+        let thumb_len = thumb_end - thumb_start;
+        let track_end_len = track_len - thumb_start - thumb_len;
```
Valentin271, line 524: This can be `thumb_end` and will be a little easier to understand imo.
```suggestion
        let track_end_len = track_len - thumb_end;
```

src/widgets/scrollbar.rs
```diff
@@ -498,25 +498,30 @@ impl Scrollbar<'_> {
...
-        let thumb_end = (position + viewport_len) * track_len / scrollable_content_len;
-
-        // We round just the positions (instead of floor / ceil), and then calculate the sizes from
-        // those positions. Rounding the sizes instead causes subtle off by 1 errors.
-        let track_start_len = thumb_start.round() as usize;
-        let thumb_end = thumb_end.round() as usize;
-
-        let thumb_len = thumb_end.saturating_sub(track_start_len);
-        let track_end_len = track_len as usize - track_start_len - thumb_len;
+        let scrollable_content_len = content_length + viewport_len - 1;
```
Valentin271, line 510: I think we can replace this with `state.content_length` and get rid of the `content_length` variable.
Or use `content_length` on line 507, as you wish as long as its consistent :).
```suggestion
        let scrollable_content_len = state.content_length + viewport_len - 1;
```

ThomasMiz: Sure, the only reason I kept the variable was because it was there before hah

ThomasMiz: Applied as of [03a4836](https://github.com/ratatui-org/ratatui/pull/959/commits/03a4836371ce5fbda226a743e055d908e68725f1)

Valentin271: Yeah I saw that after commenting. That's why I deleted my other comment, but this one was still worth cleaning imo :)

joshka (changes requested):
I also just hit this bug in an app I'm writing.

I'd like to see one or more tests for this that show the problem (make sure they fail on the previous code before the fix is applied)

src/widgets/scrollbar.rs
```diff
@@ -498,25 +498,29 @@ impl Scrollbar<'_> {
...
-        let thumb_end = thumb_end.round() as usize;
-
-        let thumb_len = thumb_end.saturating_sub(track_start_len);
-        let track_end_len = track_len as usize - track_start_len - thumb_len;
+        let scrollable_content_len = state.content_length + viewport_len - 1;
+
+        // Calculate the thumb start (inclusive) and end (exclusive) positions, with rounding
+        // and ensuring there is at least one visible thumb character.
+        let thumb_start =
+            (position * track_len + scrollable_content_len / 2) / scrollable_content_len;
```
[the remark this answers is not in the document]
ThomasMiz: > Does adding .max(1) to line 518 to clamp the length to at least 1 solve this problem more clearly, or is there an edge case that this misses?

That won't be enough because we also need to ensure `thumb_start` is not right at the end of the track. Also, another reason why I converted to integer arithmetic is because in my experience this is less prone to weirdness.

> This code seems a bit "magical" in the sense that while it does exactly what is needed, it doesn't explain why this is necessary or how it benefits the problem. This makes it difficult to maintain in the future.

I've made the code more readable as of commit [1af36d9](https://github.com/ratatui-org/ratatui/pull/959/commits/1af36d9957ea8c389e3349615221f4d02679c408)

src/widgets/scrollbar.rs
```diff
@@ -1053,4 +1064,33 @@ mod tests {
         scrollbar_no_arrows.render(buffer.area, &mut buffer, &mut state);
         assert_eq!(buffer, Buffer::with_lines(vec![expected]), "{description}");
     }
+
+    #[rstest]
```
joshka, line 1068:
```suggestion
    /// Fixes <https://github.com/ratatui-org/ratatui/pull/959> which was a bug that would not
    /// render a thumb when the viewport was very small in comparison to the content length.
    #[rstest]
```

joshka (approved):
@Valentin271 can you take another quick look over this please / @kdheepak ?

src/widgets/scrollbar.rs
```diff
@@ -498,25 +498,36 @@ impl Scrollbar<'_> {
...
+        let thumb_end = thumb_end.max(thumb_start + 1);
 
-        let thumb_len = thumb_end.saturating_sub(track_start_len);
-        let track_end_len = track_len as usize - track_start_len - thumb_len;
+        // Calculate the length of the tracks and thumb
+        let track_start_len = thumb_start;
+        let thumb_len = thumb_end - thumb_start;
+        let track_end_len = track_len - thumb_start - thumb_len;
 
         (track_start_len, thumb_len, track_end_len)
```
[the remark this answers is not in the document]
joshka: I hear you on the floats - we've seen that a few times before - particularly in the layout, but I'm convinced that this code expresses the intent of the calculations much clearer than the int version as the rounding is explicit rather than. That is unless there's a specific case where the calculation breaks in ways that make this a problem. Here I think there's unlikely to be that (though this is probably the best way to be proven wrong by someone :)).

Valentin271 (approved):
That's fine by me, I don't mind keeping floats and the code is much easier to understand now :+1:

orhun (approved):
LGTM as well

PR #963, opened 2024-02-23
perf(scrollbar): const creation

EdJoPaTo:
While doing other things in the Scrollbar widget I used a bunch of lints to improve things. A bunch of `const fn` allow for more performance and `Default` now uses the `const` new implementations.

For simplicity reasons I created `Scrollbar::new_with_symbols` for a cleaner `new` without calling `orientation()` on new.

### DIFF
--- a/src/widgets/scrollbar.rs
+++ b/src/widgets/scrollbar.rs
@@ -153,21 +153,32 @@ impl<'a> Default for Scrollbar<'a> {
 }
 
 impl<'a> Scrollbar<'a> {
-    /// Creates a new scrollbar with the given position.
+    /// Creates a new scrollbar with the given orientation.
     ///
-    /// Most of the time you'll want [`ScrollbarOrientation::VerticalLeft`] or
+    /// Most of the time you'll want [`ScrollbarOrientation::VerticalRight`] or
     /// [`ScrollbarOrientation::HorizontalBottom`]. See [`ScrollbarOrientation`] for more options.
     #[must_use = "creates the Scrollbar"]
     pub const fn new(orientation: ScrollbarOrientation) -> Self {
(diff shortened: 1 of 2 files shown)


orhun (approved):

Valentin271 (approved):
This looks good, thank you!

src/widgets/scrollbar.rs
```diff
@@ -315,22 +331,23 @@ impl<'a> Scrollbar<'a> {
     /// └─────────── begin
     /// ```
     ///
-    /// Only sets begin_symbol, end_symbol and track_symbol if they already contain a value.
+    /// Only sets `begin_symbol`, `end_symbol` and `track_symbol` if they already contain a value.
     /// If they were set to `None` explicitly, this function will respect that choice. Use their
     /// respective setters to change their value.
     ///
     /// This is a fluent setter method which must be chained or used as it consumes self
+    #[allow(clippy::needless_pass_by_value)] // Breaking change
```
joshka, line 339: Hey, this is a lint in the pedantic list, which doesn't trigger unless you enable pedantic or this lint itself. Was there some other rationale for adding this line?

https://rust-lang.github.io/rust-clippy/master/#/needless_pass_by_value

EdJoPaTo: This is one of the many pedantic lints (probably except one or two) that should be enabled for all libraries in my opinion. I ran over this file with pedantic, nursery and a bunch of restricted lints to improve it. Afterwards I removed the `#![warn(…)]` block in the top of the file, but it looks like I kept the individual allow for this one.

joshka: Ah got it - makes sense - thanks for the attention to detail :)

Issue #174, opened 2023-05-11
Design of Scrollable Widgets - Request for Comment


joshka:
## Problem

There are a few widgets that need scrolling behavior. It would be wasteful to implement this on each widget rather than doing it properly once. There is existing scroll behavior in Table, Paragraph and List. There are a few issues / PRs related to scrolling currently in the queue

- https://github.com/tui-rs-revival/ratatui/issues/173 
- https://github.com/tui-rs-revival/ratatui/issues/139
- https://github.com/tui-rs-revival/ratatui/issues/136
- https://github.com/tui-rs-revival/ratatui/issues/135
- https://github.com/tui-rs-revival/ratatui/pull/7
- https://github.com/tui-rs-revival/ratatui/pull/4
- https://github.com/tui-rs-revival/ratatui/pull/12
- https://github.com/tui-rs-revival/ratatui/pull/8
- https://github.com/tui-rs-revival/ratatui/pull/150
- https://github.com/fdehau/tui-rs/issues/495
Elsewhere:
- https://github.com/rhysd/tui-textarea/pull/13 has some discussion on scroll behavior for tui-textarea (specifically [this comment](https://github.com/rhysd/tui-textarea/pull/13#issuecomment-1501042147))
- https://github.com/swanandx/bnomial-cli/issues/1
- https://github.com/Rigellute/spotify-tui/issues/229

### Current Behavior

`Paragraph`:
- does not impl StatefulWidget
- implements scrolling as an x,y offset (`Paragraph::default().scroll((1,2))`)
- renders vertical scroll, by ignoring lines before the scroll offset and after the render area
- renders horizontal scroll by chopping the left part of the line off
- has no scroll bar

`Table`:
- impl StatefulWidget with TableState: `{ selected, offset }`
- implements scrolling as a single vertical offset to the row that will be at the top of the scroll view
- overrides the offset to ensure that the selected row is visible
- each row has a height field (longer cells are truncated)
- each row can have a bottom margin
- the table can have a header which is an always visible row
- has no scroll bar
- renders only complete rows

`List`
- impl StatefulWidget with Liststate: `{ selected, offset }`
- implements scrolling as a single vertical offset to the item that will be at the top of the scroll view
- overrides the offset to ensure that the selected item is visible
- each item has a calculated height based on the text content
- has a PR for padding and truncation
- has a PR for implementing widgets as items (which can be variable height)
- has no scroll bar
- renders only complete rows (though there is a PR for truncation

[...]

PR #965, opened 2024-02-23
feat(scrollbar): overscroll and hide_when_not_scrollable

EdJoPaTo:
Personally I don't like the scrollbar to reflect the overscroll. This adds a setting to turn this off. (personal opinion: overscroll shouldn't be the default)

Also, there is often content without the need for scrolling (Like a List with fewer items than its height allows). I also added a setting to disable rendering in such a case.

There are no test cases for this yet as I would prefer to discuss this first before taking even more effort into this.

Workaround until there is a proper solution (like this PR attempts): `state.saturating_sub(viewport_content_length)` both hides the scrollbar without scroll and hides the overscroll.

~~This PR will have merge conflicts with #963, and I will fix them after it's merged.~~ Done

According to the code comment this change will definitely be a breaking change:

https://github.com/ratatui-org/ratatui/blob/d0067c8815d5244d319934d58a9366c8ad36b3e5/src/widgets/scrollbar.rs#L131

### DIFF
--- a/src/widgets/scrollbar.rs
+++ b/src/widgets/scrollbar.rs
@@ -70,6 +70,7 @@ use crate::{prelude::*, symbols::scrollbar::*};
 #[derive(Debug, Clone, Eq, PartialEq, Hash)]
 pub struct Scrollbar<'a> {
     orientation: ScrollbarOrientation,
+    include_overscroll: bool,
     thumb_style: Style,
     thumb_symbol: &'a str,
     track_style: Style,
@@ -180,6 +181,7 @@ impl<'a> Scrollbar<'a> {
     const fn new_with_symbols(orientation: ScrollbarOrientation, symbols: &Set) -> Self {
         Self {
             orientation,
+            include_overscroll: true,
             thumb_symbol: symbols.thumb,
             thumb_style: Style::new(),
             track_symbol: Some(symbols.track),
@@ -224,6 +226,22 @@ impl<'a> Scrollbar<'a> {
     ) -> Self {
         self.orientation = orientation;
         self.symbols(symbols)
+    }
+
+    /// Defines whether the scrollbar should visualize the overscroll or not.
+    ///
+    /// When this is true (default) the scrollbar will show that it can scroll past the end of the
+    /// content until only the last element is in view.
+    /// When this is false the scrollbar will end at the last element in view.
+    ///
+    /// Keep in mind that this does not change the scrolling behaviour and only the visualization of
+    /// the scrollbar.
+    ///
+    /// This is a fluent setter method which must be chained or used as it consumes self
+    #[must_use = "method moves the value of self and returns the modified value"]
+    pub const fn include_overscroll(mut self, overscroll: bool) -> Self {
+        self.include_overscroll = overscroll;
+        self
     }
 
     /// Sets the symbol that represents the thumb of the scrollbar.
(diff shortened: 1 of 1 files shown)


DeflateAwning:
I support the development and encorporation of this feature!

DeflateAwning:
What's the status of this feature? Is there any way I can help out. Would love to get it merged

EdJoPaTo:
> There are no test cases for this yet as I would prefer to discuss this first before taking even more effort into this.

Basically… is this the way we want to go forward?

DeflateAwning:
Hmm this doesn't seem to work for me when I try to use this in [the project I'm working on (in the `test-overscroll-prevention` branch's latest commit)](https://github.com/DeflateAwning/baud-boss/tree/test-overscroll-prevention). It still allows me to scroll past the limits vertically.

I think this is tied to the lack of "communication" between the `ScrollBar` (and its viewport size) to the `ScrollBarState`.

EdJoPaTo:
> Hmm this doesn't seem to work for me when I try to use this in [the project I'm working on (in the `test-overscroll-prevention` branch's latest commit)](https://github.com/DeflateAwning/baud-boss/tree/test-overscroll-prevention). It still allows me to scroll past the limits vertically.

The Scrollbar does not change the scrolling behaviour in any way. It only represents the scroll visually by displaying a scrollbar.

The overscroll behaviour is currently represented on the scrollbar which is what this PR allows to configure. It simply allows the scrollbar to show the end is reached. Whatever the widget itself does.

DeflateAwning:
Hmm I'm not sure I follow entirely, but that might be okay. 

Can you give an example of a before-and-after of what this PR actually does?

joshka:
> The overscroll behaviour is currently represented on the scrollbar which is what this PR allows to configure. It simply allows the scrollbar to show the end is reached. Whatever the widget itself does.

Oh yeah, your right. My brain is still thinking that the scroll position is actually controlling the scroll of the data of whatever element is involved and not just the display of the bar.

EdJoPaTo:
> Hmm I'm not sure I follow entirely

Every widget by itself defines how scrolling works and `widgets::Scrollbar` defines how the scrollbar looks.

This in itself describes the scroll mess-up of ratatui currently well: There is no common ground. Widgets define the scrolling behavior differently (mqttui mainly uses 3 different widgets and has workarounds to get them behave somewhat similarly) and the `widgets::Scrollbar` tries to visualize what different widgets define differently.

> Do you think it would be worth splitting the hide feature and overscroll functionality into separate PRs - the hide functionality seems pretty simple and easy to merge without much discussion, but the overscroll seems more complex and has a few things I'd like to consider.

Hiding the scrollbar when there is nothing to scroll is very pointless with the current overscroll behavior as its basic idea is to be always able to scroll when there is content.

https://github.com/ratatui-org/ratatui/blob/6326d6e516499a5df60be25c2a54c31aca67e296/src/widgets/scrollbar.rs#L649-L652

The only option not to scroll with the current overscroll behavior is when the content is 0 or 1 lines high. Then there is no last line which could be scrolled to. Yes, we could split this but the hide method on its own is pointless then.

> For both, I'd suggest introducing enums that clarify the behavior rather than bools.

I like thinking about it, but I'll postpone this because of my other points.

> For overscroll, I think it's worth considering the behavior of preventing overscrolling, and not just the display.

Here we are at widgets behaviour / scrollbar visualization point again (my first point). Widgets define how scrolling works and the scrollbar only tries to visualize it. The current `widgets::Scrollbar` should not manipulate other widgets. That would be horrible design.

Personal opinion: We need to have some shared idea of scrollable widgets (#174). Then we can ensure the widgets have a shared scrolling behavior. Only then we can visualize the scrolling behavior of the individual widgets consistently.

Therefore, I think this PR is currently pointless until #174 advances. I would close it.

DeflateAwning:
After revisiting this with a better understanding of what this change does (and doesn't do), I'd love to see this merged!

@EdJoPaTo, would you like help to implement the minor changes suggested? Or are you still not feeling great about it?

EdJoPaTo:
It would improve the visual situation a little, but it would not solve this mix up and confusion around it (probably even increase it). There is a workaround (described in my first post) which I use for my applications.

I think my suggested `ScrollBehavior` of #174 could actually be a good step forward for this. Guess it should be tested with a single widget. Once that's done it could be used in more places.

DeflateAwning:
I think that improving the visual situation is an important first step.

I came up with what I assume is a similar work-around: not changing the scroll value when it reaches the limit (based on the viewport size, content length, etc.)

PR #969, opened 2024-02-25
docs(text): update Text and Line docs

joshka:


joshka:
image

src/text/line.rs
```diff
@@ -10,35 +10,37 @@ use crate::prelude::*;
...
 /// # Setter Methods
 ///
 /// These methods are fluent setters. They return a `Line` with the property set.
 ///
 /// - [`Line::spans`] sets the content of the line.
 /// - [`Line::style`] sets the style of the line.
 /// - [`Line::alignment`] sets the alignment of the line.
+/// - [`Line::left_aligned`] sets the alignment of the line to [`Alignment::Left`].
+/// - [`Line::centered`] sets the alignment of the line to [`Alignment::Center`].
+/// - [`Line::right_aligned`] sets the alignment of the line to [`Alignment::Right`].
```
EdJoPaTo, line 37: Is it really worth it to have these in the doc comment? `cargo doc` / <docs.rs> lists them relatively obviously and will always be up-to-date as it's generated from the source. This is basically duplicated documentation.

Conversions and iteration methods could be displayed more nicely by the docs, so I see the point of including them.

joshka: The idea / rationale behind this is to provide a list of methods upfront that helps newer rust users find what they need, and groups the methods in a way that makes it easier to understand the general ideas of the types. This is more difficult to see from the auto generated summary.

Valentin271: It's added work that I don't like, but I must say I usually have a hard time finding what I'm looking for in the long list of methods (in ratatui or other libs) without the text search (which suppose I know what I'm looking for).
Plus we can list other methods not associated with the type like here https://github.com/ratatui-org/ratatui/pull/969/files#diff-d8aa1f145d6f63ab98ad22cb5d47fc3cd79b4a4fb3b7f0c407c3eecc5c19dafdR25

Valentin271 (approved):
LGTM

PR #970, opened 2024-02-25
feat(backend) : Add Bevy Backend

gold-silver-copper:
This pull request adds support for using the rust game engine 'Bevy' as a Backend for Ratatui. This allows users of Bevy to make a game UI with ratatui. It also allows for users of Ratatui to compile their programs to webassembly.

Things this pull request does:

- Add Bevy as an optional dependency and sets all the proper feature cfg flags
- Add once-cell as a dev-dependency (I couldn't figure out how to do the demo without it because of some borrow checker issues, I welcome a solution without it)
- Adds a bevy option for the demo, run it with `cargo run --example demo --features=bevy --no-default-features --release`
- Adds a new example 'hello_bevy.rs' which is a barebones example showing how to render sprites from bevy as well as use ratatui for UI, run it with `cargo run --example hello_bevy --features=bevy --release`
- Slightly changes the demo to demonstrate more style options (I just did this to help keep track while implementing kind of unnecessary)
-  Derives 'Component' on ratatui 'Terminal' so that a terminal Entity can be spawned for Bevy (behind bevy feature flag)
- Adds a whole new BevyBackend for ratatui!

https://github.com/ratatui-org/ratatui/assets/157451540/c59f021f-b87a-4f0c-9049-9afdbe2beee4


joshka:
Copied from the discord:

Hey, first up, this is pretty damn cool, and I can see it was a pretty decent amount of work to do.
However, I'm going to push back gently against this being part of the core ratatui library.

The primary reason is that adding bevy as a new backend adds the need to maintain the new feature code for the ratatui maintainers, and means that they need experience making choices / design decisions that are appropriate and fit with the bevy way of doing things. This seems like a high bar to meet, and an extra burden. Case in point is the 750 lines of code in the backend that are super bevy specific - most of which I'd have no clue about whether is correct or not (or where they fit in a bevy application).

Taking on a new dependency like this also subjects Ratatui to the perils of transitive semver concerns.

Backends are notoriously under-tested areas of the ratatui source code, which makes them fragile to changes generally. (See https://app.codecov.io/gh/ratatui-org/ratatui)

My recommendation (and this is a common one I've made for various widget ideas as well)  is to extract the code into an external crate that uses Ratatui. The only part I see that would prevent this from the PR is the derive `Component` for terminal that you probably can't do externally, but I think that sort of problem is usually solved by an additional layer of code (e.g. your own `TerminalComponent` struct that has a `Terminal` field).

The biggest benefit that you'll get from this approach is the ability to release this on your own schedule to fix bugs / create new functionality etc. I have a few crates that use `release-plz` workflows for releasing and it's super convenient to be able to merge an automatically created release PR with the changelog / and cargo.toml version bump to kick off a release.

gold-silver-copper:
Thank you for your suggestions, I will create a bevy_ratatui crate soon and make that the interface between the two libraries. TUI is so fun !

gold-silver-copper:
https://github.com/gold-silver-copper/bevy_ratatui/

PR #976, opened 2024-02-28
refactor(non-src): apply pedantic lints

EdJoPaTo:
Fixes many not yet enabled lints (mostly pedantic) on everything that is not the lib (examples, benchs, tests). Therefore, this is not containing anything that can be a breaking change.

I do not think this breaks something, but I might have overlooked something.

Lints are not enabled as that should be the job of #974. I created this as a separate PR as its mostly independent and would only clutter up the diff of #974 even more.

Also see https://github.com/ratatui-org/ratatui/pull/974#discussion_r1506458743

### DIFF
--- a/examples/canvas.rs
+++ b/examples/canvas.rs
@@ -13,6 +13,8 @@
 //! [examples]: https://github.com/ratatui-org/ratatui/blob/main/examples
 //! [examples readme]: https://github.com/ratatui-org/ratatui/blob/main/examples/README.md
 
+#![allow(clippy::wildcard_imports)]
+
 use std::{
     io::{self, stdout, Stdout},
     time::{Duration, Instant},
@@ -44,8 +46,8 @@ struct App {
 }
 
 impl App {
-    fn new() -> App {
-        App {
+    fn new() -> Self {
+        Self {
             x: 0.0,
             y: 0.0,
             ball: Circle {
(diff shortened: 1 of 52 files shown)


examples/canvas.rs
```diff
@@ -160,8 +162,12 @@ impl App {
...
 
     fn boxes_canvas(&self, area: Rect) -> impl Widget {
-        let (left, right, bottom, top) =
-            (0.0, area.width as f64, 0.0, area.height as f64 * 2.0 - 4.0);
+        let (left, right, bottom, top) = (
+            0.0,
+            f64::from(area.width),
+            0.0,
+            f64::from(area.height).mul_add(2.0, -4.0),
+        );
```
joshka, line 170:
```suggestion
        let left = 0.0;
        let right = f64::from(area.width);
        let bottom = 0.0;
        let top = f64::from(area.height).mul_add(2.0, -4.0);
```

joshka (commented):
LGTM - one small suggestion, but otherwise good to merge.

examples/canvas.rs
EdJoPaTo: After seeing the suggestion this change is really obvious :D

joshka (approved):

joshka:
Thanks - I've merged this :)

PR #977, opened 2024-03-03
perf(borders): allow border!() in const

EdJoPaTo:
This allows more compiler optimizations when the macro is used.

Interestingly `TOP | BOTTOM` is bitwise-or and not `const` (as it's a trait…). `insert` is also not `const`, `union` is.

Also moves tests from test folder to the file itself as I overlooked them initially. Borders was together in a file with Block back when it was introduced.

Also see #11 and #832


joshka (approved):
LGTM, thanks for the PR

kdheepak (approved):

Issue #978, opened 2024-03-04
Remove Default when useless on its own


EdJoPaTo:
## Description

Some structs implement Default. When using said default its pointless on its own. It requires other methods which can easily lead to mistakes.

What is a `Terminal::default()`? (Chart)`Dataset::default()` without data.

## To Reproduce

Just using default without something else is pointless:

```rust
let datasets = vec![
    Dataset::default(),
    Dataset::default(),
];
let x_axis = Axis::default();
let y_axis = Axis::default();
// 
let chart = Chart::new(datasets)
    .x_axis(x_axis)
    .y_axis(y_axis);

// lets render... nothing?
let chart = Chart::default();
```

## Expected behavior

Default should work on its own without further modifications.
Rust is well suited to create only valid data structures. The builder pattern is a great example for this preventing invalid data structures on `build()`.

There should be a new method accepting the required data instead of manipulating a default.

```diff
-Dataset::default().data(data);
+Dataset::new(data);
```

In this case the `data` method should also be removed to have a single way of creating the struct. Especially with the single-use, short-lived structs there is no point in keeping the setters rather than confusion.

When structs contain other structs that have a Default implementation which does not make any sense on its own then `Option::None` is a way better fit instead of some pointless default.

Migration suggestion:
- create new method with required arguments, deprecate the setter methods which are required, pointing towards the new method.
- after some time remove said setters and Default implementation.

## Defaults which seem useless

An easy indicator: When the documentation of the struct never uses the `default()` on its own.

Some are debatable like `Style` (its very convenient to use `Style::default()` over `None`) and especially widgets are straight forward.

- [ ] `BarChart`
  - [ ] `BarGroup`
  - [ ] `Bar`
- [ ] `Block` (default is useless but with titles and borders probably the simplest way to keep it that way?)
  - [ ] `Title`
- [ ] `Buffer`
  - [ ] `Cell`
- [ ] `CalendarEventStore` (unsure)
- [ ] `Canvas`
  - [ ] `Label`
  - [ ] `Layer`
  - [ ] `BrailleGrid`
  - [ ] `CharGrid`
  - [ ] `HalfBlockGrid`
  - [ ] `Circle`
  - [ ] `Line`
  - [ ] `Map` (drawing with `Color::Reset` me be intended but very unlikely)
  - [ ] `Rectangle`
- [ ] `Chart`
  - [ ] `Axis`
  - [ ] `ChartLayout`
  - [ ] `Dataset`
[...]

PR #980, opened 2024-03-04
refactor: remove pointless default on internal structs

EdJoPaTo:
See #978

Also remove other derives. They are unused and just slow down compilation.

### DIFF
--- a/src/widgets/chart.rs
+++ b/src/widgets/chart.rs
@@ -400,7 +400,6 @@ impl<'a> Dataset<'a> {
 
 /// A container that holds all the infos about where to display each elements of the chart (axis,
 /// labels, legend, ...).
-#[derive(Debug, Default, Clone, Eq, PartialEq, Hash)]
 struct ChartLayout {
     /// Location of the title of the x axis
     title_x: Option<(u16, u16)>,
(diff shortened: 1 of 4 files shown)


src/widgets/chart.rs
```diff
@@ -691,50 +690,58 @@ impl<'a> Chart<'a> {
...
+            axis_y = Some(x);
             x += 1;
         }
 
-        if x < area.right() && y > 1 {
-            layout.graph_area = Rect::new(x, area.top(), area.right() - x, y - area.top() + 1);
+        let graph_width = area.right().saturating_sub(x);
+        let graph_height = y.saturating_sub(area.top()).saturating_add(1);
+        if graph_width == 0 || graph_height == 0 {
+            return None;
```
Valentin271, line 724: Could this be moved to the top of the method? To trigger the return earlier if the condition is met.

Valentin271 (approved):
Hey, this looks good to me, thanks for this.

Had a little comment but it's non-blocking.

src/widgets/chart.rs
EdJoPaTo: At the top is already a check for total area == 0. I am not even sure this one is actually needed as the lines above should be skipped when there isn't enough space.

It might be another idea to just assert in this position as it should never be == 0 with the check above.

Valentin271: Nevermind we use `x` and `y` which can be modified just above.

> It might be another idea to just assert in this position as it should never be == 0 with the check above

It could be 0 because of the substraction.

The check at the top is actually the one that could be removed because this is in the render. But it's in another method so might be best to leave it.

EdJoPaTo: > Nevermind we use `x` and `y` which can be modified just above.

Yes, but the whole idea of that method is to not do it when there is not enough space.

> The check at the top is actually the one that could be removed[…]

It's basically the early check you asked for? 🤔

EdJoPaTo: I switched the check / early return into debug assertions. They should never trigger as the code above should be correct. Debug assertions should be enough as people will test with debug enabled first. I don't really see the point of enabling them on release builds too.

Also, even without this != 0 check the Chart renders just fine without any panic on empty areas. It's just an early abort to save resources.

Valentin271: > It's basically the early check you asked for?

Kind of. I just took a more in depth look at the code and saw the `render` method does the same check too.

> I switched the check / early return into debug assertions

That's a good idea. I think we're good with this version of the code.

Valentin271 (approved):

orhun (approved):

EdJoPaTo:
It might be a good idea to split the canvas changes from the rest of this PR as it has also changes to the code behavior?

Valentin271:
Surely I'm forgetting/misunderstanding something... Canvas doesn't have a change in behavior ..?

EdJoPaTo:
> Canvas doesn't have a change in behavior ..?

The checks and math logic has changed. That might have changed something. It's not much so it could be fine to mix it.

joshka:
> The checks and math logic has changed. That might have changed something. It's not much so it could be fine to mix it.

I think you might be referring to the Chart (not canvas)
If it didn't break the tests (and [those lines are covered by tests](https://app.codecov.io/gh/ratatui-org/ratatui/blob/main/src%2Fwidgets%2Fchart.rs)) then I'm not bothered too much.

PR #981, opened 2024-03-04
fix(termion): Add fallback for failing cursor detection

erak:
With `TermionBackend`, when the cursor position on `stdout` is being read from another thread then the one
reading events from `stdin`, cursor detection will fail. See: https://gitlab.redox-os.org/redox-os/termion/-/issues/173

The workaround is to always store the last known cursor position and use it as fallback.

### DIFF
--- a/ratatui-termion/src/lib.rs
+++ b/ratatui-termion/src/lib.rs
@@ -80,6 +80,7 @@ where
     W: Write,
 {
     writer: W,
+    last_known_cursor_position: Option<Position>,
 }
 
 impl<W> TermionBackend<W>
@@ -103,7 +104,10 @@ where
     /// let backend = TermionBackend::new(stdout());
     /// ```
     pub const fn new(writer: W) -> Self {
-        Self { writer }
+        Self {
+            writer,
+            last_known_cursor_position: None,
+        }
     }
 
     /// Gets the writer.
@@ -176,14 +180,26 @@ where
         self.writer.flush()
     }
 
+    /// Get the current cursor position.
+    ///
+    /// If termion fails to get the cursor position, it will return the last known cursor position,
+    /// if available. If there is no known cursor position, it will return an error.
     fn get_cursor_position(&mut self) -> io::Result<Position> {
-        termion::cursor::DetectCursorPos::cursor_pos(&mut self.writer)
-            .map(|(x, y)| Position { x: x - 1, y: y - 1 })
+        let position = termion::cursor::DetectCursorPos::cursor_pos(&mut self.writer)
+            .map(|(x, y)| Position::new(x - 1, y - 1))
+            .or_else(|e| self.last_known_cursor_position.ok_or_else(|| e))?;
+        self.last_known_cursor_position = Some(position);
+        Ok(position)
     }
 
+    /// Set the cursor position.
+    ///
+    /// Setting the cursor position will update the last known cursor position.
     fn set_cursor_position<P: Into<Position>>(&mut self, position: P) -> io::Result<()> {
-        let Position { x, y } = position.into();
+        let position = position.into();
+        let Position { x, y } = position;
         write!(self.writer, "{}", termion::cursor::Goto(x + 1, y + 1))?;
+        self.last_known_cursor_position = Some(position);
         self.writer.flush()
     }
 
--- a/ratatui-termion/src/lib.rs
+++ b/ratatui-termion/src/lib.rs
@@ -187,7 +187,7 @@ where
     fn get_cursor_position(&mut self) -> io::Result<Position> {
         let position = termion::cursor::DetectCursorPos::cursor_pos(&mut self.writer)
             .map(|(x, y)| Position::new(x - 1, y - 1))
-            .or_else(|e| self.last_known_cursor_position.ok_or_else(|| e))?;
+            .or_else(|e| self.last_known_cursor_position.ok_or(e))?;
         self.last_known_cursor_position = Some(position);
         Ok(position)
     }
--- a/ratatui/tests/backend_termion.rs
+++ b/ratatui/tests/backend_termion.rs
@@ -61,3 +61,58 @@ fn backend_termion_should_only_write_diffs() -> Result<(), Box<dyn std::error::E
 
     Ok(())
 }
+
+/// Retrieving the cursor from an in-memory buffer will fail in
+/// `termion::cursor::DetectCursorPos::cursor_pos()`. If there is no known last cursor, retrieving
+/// the cursor will ulimately fail.
+///
+/// This simulates a usage of `termion` where retrieving the cursor fails in a multi-threaded
+/// environment: [](https://gitlab.redox-os.org/redox-os/termion/-/issues/173)
+#[cfg(feature = "termion")]
+#[test]
+fn backend_termion_should_fail_on_missing_last_known_cursor(
+) -> Result<(), Box<dyn std::error::Error>> {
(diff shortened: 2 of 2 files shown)


Cargo.toml
```diff
@@ -100,7 +100,7 @@ use_self = "warn"
 ## By default, we enable the crossterm backend as this is a reasonable choice for most applications
 ## as it is supported on Linux/Mac/Windows systems. We also enable the `underline-color` feature
 ## which allows you to set the underline color of text.
-default = ["crossterm", "underline-color"]
+default = ["termion", "underline-color"]
```
EdJoPaTo, line 103: This was probably changed while working on it and mistakenly added to the commit?

src/backend/termion.rs
```diff
@@ -137,11 +141,20 @@ where
...
-        termion::cursor::DetectCursorPos::cursor_pos(&mut self.writer).map(|(x, y)| (x - 1, y - 1))
+        match termion::cursor::DetectCursorPos::cursor_pos(&mut self.writer) {
+            Ok((x, y)) => {
+                self.cursor = Some((x - 1, y - 1));
+                Ok((x - 1, y - 1))
+            }
+            Err(_) => self
+                .cursor
+                .ok_or_else(|| io::Error::other("No last known cursor position")),
+        }
```
EdJoPaTo, line 152: Personally I would keep the map logic and do the match after that. Then there is no duplicate x-1,y-1 logic needed and its only in one place.

src/backend/termion.rs
```diff
@@ -137,11 +141,20 @@ where
...
     fn get_cursor(&mut self) -> io::Result<(u16, u16)> {
-        termion::cursor::DetectCursorPos::cursor_pos(&mut self.writer).map(|(x, y)| (x - 1, y - 1))
+        match termion::cursor::DetectCursorPos::cursor_pos(&mut self.writer) {
+            Ok((x, y)) => {
+                self.cursor = Some((x - 1, y - 1));
+                Ok((x - 1, y - 1))
+            }
+            Err(_) => self
+                .cursor
+                .ok_or_else(|| io::Error::other("No last known cursor position")),
```
EdJoPaTo, line 151: Can a cursor position of 0,0 just be assumed as a fallback to the fallback? Or could that have issues?

EdJoPaTo (changes requested):

Cargo.toml
erak: Oh, nice catch. Fixed.

src/backend/termion.rs
erak: I think that could work. I discovered this issue with auto-resizing an inline viewport. Can't tell if that would have other side effects then the frame being redrawn at 0,0.

src/backend/termion.rs
erak: Kept the match structure as it was, but introduced a local variable that holds the mapped cursor.

erak:
I'd need some advice on the unit testing. So currently, both tests that we're added to this PR do fail on terminal 
construction with `Inappropriate ioctl for device`:

```rust
let backend = TermionBackend::new(&mut stdout);
```

I could change the construction to use terminal options, which would work. But then, retrieving the cursor position seems to block and the test never finishes.

```rust
let backend = TermionBackend::new(&mut stdout);
let area = Rect::new(0, 0, 3, 1);
let mut terminal = Terminal::with_options(
    backend,
    TerminalOptions {
        viewport: Viewport::Fixed(area),
     },
)?;
```

kdheepak:
Are the tests failing locally for you as well, or is it just failing on GitHub Actions?

erak:
> Are the tests failing locally for you as well, or is it just failing on GitHub Actions?

They fail with `cargo nextest run <args>`, but they pass with `cargo test <args>` :thinking:

src/backend/termion.rs
```diff
@@ -67,6 +67,7 @@ where
     W: Write,
 {
     writer: W,
+    cursor: Option<(u16, u16)>,
```
EdJoPaTo, line 70: This could use `ratatui::layout::Position`

EdJoPaTo (commented):
Running the tests normally still works fine, nextest still fails. Not sure if they could be ignored just for nextest. Shouldn't be the reason this PR gets stuck…

src/backend/termion.rs
joshka: This would means the underlying error will never be detectable. I think I'd prefer the error to just pass through here. Does that make sense?

I.e. if error and no fallback based on the previously set cursor, then just let the original error pass through

src/backend/termion.rs
```diff
@@ -137,11 +141,21 @@ where
     }
 
     fn get_cursor(&mut self) -> io::Result<(u16, u16)> {
```
joshka, line 143: Add doc comments to document the fallback behavior here

src/backend/termion.rs
```diff
@@ -137,11 +141,21 @@ where
...
+                self.cursor = Some(cursor);
+                Ok(cursor)
+            }
+            Err(_) => self
+                .cursor
+                .ok_or_else(|| io::Error::other("No last known cursor position")),
+        }
     }
 
     fn set_cursor(&mut self, x: u16, y: u16) -> io::Result<()> {
```
joshka, line 156: And perhaps doc comment here ("A cursor position set here will be used as a fallback...")

joshka (commented):
Generally ok with the idea, some comments about the implementation / behavior

joshka (approved):
Merged and rebased and made to use the Position struct instead of a bare tuple.

joshka:
There is some CI integration test fun to still resolve here.

I think we should aim to get this in 0.29.0, at a medium priority.

joshka:
I changed our makefile to use cargo test instead of nextest for the backend tests. This fixes the errors locally and probably on CI.

orhun:
One little test failure then this should be good to go.

joshka:
> One little test failure then this should be good to go.

It's a windows / termion / CI whack-a-mole problem. I don't think I have the headspace to dig into before we drop 0.29.0.
If we merge it as-is this will break for any contributors that use windows, which I'd prefer not to do. That means that we should defer merging this to the following release IMO.

joshka:
In https://gitlab.redox-os.org/redox-os/termion/-/merge_requests/200 for a MR I just added that fixes the internal panics that calling this on a different thread that the input thread causes. These should now instead return errors.

However, the reality of this problem is that calling the cursor location function on a different thread / async is a bad idea. The internal timeout on this call in termion is 100ms, the behavior can succeed, fail due to both threads pulling parts of the input off the stream, timeout, and I wouldn't be surprised to see an succeeding by incorrect value be returned from time to time (if the reads are interleaved). This seems like the sort of thing which cannot be fixed without wrapping the code which reads from the tty device in a lock.

Because of this, I don't think the right behavior of this method call is to fallback to a previous known position.

When there is a thread reading from the tty the following code would be an undefined result. It may timeout, return (0, 0), return (5, 0), or fail due to an invalid response.

```rust
backend.set_cursor_position(Position { x: 0, y: 0 });
backend.write(b"Hello");
assert_eq!(backend.get_cursor_position().unwrap(), Position { x: 5, y: 0 });
```

joshka (commented):
Marking as not approved

joshka (changes requested):
Marking as not approved

joshka:
@erak I'm closing this PR, but if I've done so in error and you disagree with my assessment of the problem, just let us know and we can reopen and continue this PR.

Issue #990, opened 2024-03-20
Paragraph wrapped in a block can not be styled as expected


PierreQuebriac:
## Description
I was following the tutorial on the JSON editor: [https://ratatui.rs/tutorials/json-editor/](https://ratatui.rs/tutorials/json-editor/) and noticed a dissimilarity of result after compilation on the  [title](https://ratatui.rs/tutorials/json-editor/ui-main/#the-title) step.

The tutorial offer to write this:
```rs
    let title_block = Block::default()
        .borders(Borders::ALL)
        .style(Style::default());

    let title = Paragraph::new(Text::styled(
        "Create New Json",
        Style::default().fg(Color::Green),
    ))
    .block(title_block);

    f.render_widget(title, chunks[0]);
```

And I was expecting to get a box with white borders surrounding a green text saying "Create new JSON" and after compilation I got a box with white borders and a **white** text.

## To Reproduce
The easiest way to reproduce was to download the raratui website app [here](https://github.com/ratatui-org/ratatui-website/tree/main/code/ratatui-json-editor-app) and run `cargo run`

## Expected behavior
I was expected that a code like this
```rs
let title = Paragraph::new(Text::styled(
        "Create New Json",
        Style::default().fg(Color::Green),
    ))
    .block(
        Block::default()
            .borders(Borders::ALL)
            .style(Style::default()),
    );
```

would provide a white border box with a green text inside.

## Screenshots
![issue_screenshot](https://github.com/ratatui-org/ratatui/assets/72198491/230d579b-9820-4b80-8c25-adc3c3b1a523)

## Environment
- OS: Linux
- Terminal Emulator: xterm
- Font: Monospace
- Crate version: 0.26.1
- Backend: crossterm

## Additional context
I noticed that in version 0.25.0 this bug does not happen

Thanks for helping me !

orhun:
Changing `Text` to `Line` fixes this:

```diff
diff --git a/code/ratatui-json-editor-app/src/ui.rs b/code/ratatui-json-editor-app/src/ui.rs
index edccbaa..8b47d4d 100644
--- a/code/ratatui-json-editor-app/src/ui.rs
+++ b/code/ratatui-json-editor-app/src/ui.rs
@@ -29,7 +29,7 @@ pub fn ui(f: &mut Frame, app: &App) {
         .borders(Borders::ALL)
         .style(Style::default());
 
-    let title = Paragraph::new(Text::styled(
[...]

PR #992, opened 2024-03-21
fix: ensure that paragraph correctly renders styled text

joshka:
Paragraph was ignoring the new style field on Text added in 0.26.0

Fixes: https://github.com/ratatui-org/ratatui/issues/990

### DIFF
--- a/src/widgets/paragraph.rs
+++ b/src/widgets/paragraph.rs
@@ -403,7 +403,6 @@ impl<'a> Styled for Paragraph<'a> {
 
 #[cfg(test)]
 mod test {
-
     use super::*;
     use crate::{
         backend::TestBackend,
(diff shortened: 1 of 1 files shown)


src/widgets/paragraph.rs
```diff
@@ -402,6 +403,7 @@ impl<'a> Styled for Paragraph<'a> {
 
 #[cfg(test)]
 mod test {
+
```
EdJoPaTo, line 406: Why is `clippy fmt` happy about this? :eyes: 
I would have expected that fmt removes this empty line at the start of a block.

joshka: Ha, I initially wrote a test using rstest for this and vscode inserted an import automatically. Decided not to use it and missed that there was an added newline. It can be removed.

orhun (approved):

src/widgets/paragraph.rs
EdJoPaTo: Best suggestion ever I guess? :ghost: 

```suggestion
```

PR #993, opened 2024-03-22
refactor(span): rename to_aligned_line into into_aligned_line

EdJoPaTo:
This deprecates the old methods hinting the new name.

With the Rust method naming conventions this method is an into method consuming the Span. Therefore, it's more consistent to name it into instead of to.

### DIFF
--- a/examples/constraint-explorer.rs
+++ b/examples/constraint-explorer.rs
@@ -282,7 +282,7 @@ impl App {
 
     fn header() -> impl Widget {
         let text = "Constraint Explorer";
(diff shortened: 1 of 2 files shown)


joshka (approved):
LGTM - will merge with a minor addition to the commit message (naming the specific methods changed to make the changelog searchable)

EdJoPaTo:
I updated the PR text too as the commit links there and people might look there for context. (I think it’s the easier way to update the PR message and then merge with the updated message)

PR #998, opened 2024-03-25
feat(text): add push methods for text and line

joshka:
Adds the following methods to the `Text` and `Line` structs:
- Text::push_line
- Text::push_span
- Line::push_span

This allows for adding lines and spans to a text object without having
to call methods on the fields directly, which is usefult for incremental
construction of text objects.

### DIFF
--- a/src/text/text.rs
+++ b/src/text/text.rs
@@ -441,6 +441,20 @@ impl<'a> Text<'a> {
     pub fn iter_mut(&mut self) -> std::slice::IterMut<Line<'a>> {
         self.lines.iter_mut()
     }
+
+    /// Adds a line to the text.
+    pub fn push_line(&mut self, line: Line<'a>) {
+        self.lines.push(line);
+    }
+
+    /// Adds a span to the last line of the text.
+    pub fn push_span(&mut self, span: Span<'a>) {
+        if let Some(last) = self.lines.last_mut() {
+            last.push_span(span);
+        } else {
+            self.lines.push(Line::from(span));
+        }
+    }
 }
 
 impl<'a> IntoIterator for Text<'a> {
@@ -854,6 +868,61 @@ mod tests {
         assert_eq!(Text::default().italic().style, Modifier::ITALIC.into());
     }
 
+    #[test]
+    fn left_aligned() {
+        let text = Text::from("Hello, world!").left_aligned();
+        assert_eq!(text.alignment, Some(Alignment::Left));
+    }
+
+    #[test]
+    fn centered() {
+        let text = Text::from("Hello, world!").centered();
+        assert_eq!(text.alignment, Some(Alignment::Center));
+    }
+
+    #[test]
+    fn right_aligned() {
+        let text = Text::from("Hello, world!").right_aligned();
+        assert_eq!(text.alignment, Some(Alignment::Right));
+    }
+
+    #[test]
+    fn push_line() {
+        let mut text = Text::from("Hello, world!");
+        text.push_line(Line::from("How are you?"));
+        assert_eq!(
+            text.lines,
+            vec![Line::from("Hello, world!"), Line::from("How are you?")]
+        );
+    }
+
+    #[test]
+    fn push_line_empty() {
+        let mut text = Text::default();
+        text.push_line(Line::from("Hello, world!"));
+        assert_eq!(text.lines, vec![Line::from("Hello, world!")]);
+    }
+
+    #[test]
+    fn push_span() {
+        let mut text = Text::from("Hello, world!");
+        text.push_span(Span::raw("How are you?"));
+        assert_eq!(
+            text.lines,
+            vec![Line::from(vec![
+                Span::raw("Hello, world!"),
+                Span::raw("How are you?")
+            ])],
+        );
+    }
+
+    #[test]
+    fn push_span_empty() {
+        let mut text = Text::default();
+        text.push_span(Span::raw("Hello, world!"));
+        assert_eq!(text.lines, vec![Line::from(Span::raw("Hello, world!"))],);
+    }
+
     mod widget {
         use super::*;
         use crate::assert_buffer_eq;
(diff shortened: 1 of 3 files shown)


joshka:
The underlying use case that prompted this change is the following code that defines a `JsonWidget`. This code is significantly simplified from the equivalent that calls into the fields, as there are problems borrowing the fields mutably multiple times.

```rust
pub struct JsonWidget {
    json: Value,
}

const PUNCTUATION_STYLE: Style = Style::new().add_modifier(Modifier::BOLD).fg(Color::White);
const KEY_STYLE: Style = Style::new().add_modifier(Modifier::BOLD).fg(Color::Blue);
const STRING_STYLE: Style = Style::new().fg(Color::Green);
const NUMBER_STYLE: Style = Style::new().fg(Color::Yellow);
const BOOLEAN_STYLE: Style = Style::new().fg(Color::Cyan);
const NULL_STYLE: Style = Style::new().add_modifier(Modifier::DIM);

impl Widget for &JsonWidget {
        let mut text = Text::raw("");
        push_value(&self.json, &mut text, 0);
        text.render(area, buf);
    }
}

fn push_value(value: &Value, text: &mut Text, indent: usize) {
    match value {
        Value::Null => push_null(text),
        Value::Bool(b) => push_bool(text, *b),
        Value::Number(num) => push_number(text, num),
        Value::String(s) => push_string(text, s),
        Value::Array(arr) => push_array(text, arr, indent),
        Value::Object(map) => push_object(text, map, indent),
    };
}

fn push_null(text: &mut Text) {
    let null = Span::styled("null", NULL_STYLE);
    text.push_span(null);
}

fn push_bool(text: &mut Text, b: bool) {
    let bool = Span::styled(b.to_string(), BOOLEAN_STYLE);
    text.push_span(bool);
}

fn push_number(text: &mut Text, num: &Number) {
    let num = Span::styled(format!("{}", num), NUMBER_STYLE);
    text.push_span(num);
}

fn push_string(text: &mut Text, s: &str) {
    let string = Span::styled(format!("\"{}\"", s), STRING_STYLE);
    text.push_span(string);
}

fn push_array(text: &mut Text, arr: &[Value], indent: usize) {
    let open_bracket = Span::styled("[", PUNCTUATION_STYLE);
    let close_bracket = Span::styled("]", PUNCTUATION_STYLE);
    let comma = Span::styled(", ", PUNCTUATION_STYLE);

    text.push_span(open_bracket);
    for (i, value) in arr.iter().enumerate() {
        if i > 0 {
            text.push_span(comma.clone());
        }
        text.push_line(" ".repeat(indent + 2).into());
        push_value(value, text, indent + 2);
    }
    if !arr.is_empty() {
        text.push_line(" ".repeat(indent).into());
    }
    text.push_span(close_bracket)
}

fn push_object(text: &mut Text, map: &Map<String, Value>, indent: usize) {
    let open_brace = Span::styled("{", PUNCTUATION_STYLE);
    let close_brace = Span::styled("}", PUNCTUATION_STYLE);
    let comma = Span::styled(", ", PUNCTUATION_STYLE);
    let colon = Span::styled(": ", PUNCTUATION_STYLE);

    text.push_span(open_brace);
    for (i, (key, value)) in map.iter().enumerate() {
        let key = Span::styled(format!("\"{}\"", key), KEY_STYLE);
        if i > 0 {
            text.push_span(comma.clone());
        }
        text.push_line(" ".repeat(indent + 2).into());
        text.push_span(key);
        text.push_span(colon.clone());
        push_value(value, text, indent + 2);
    }
    if !map.is_empty() {
        text.push_line(" ".repeat(indent).into());
    }
    text.push_span(close_brace)
}
```

image

orhun (approved):
Looks convenient to me

joshka:
I'll wait on a second review for some feedback on this one as it's a change to the APIs which could do with some input. I think these are the most likely the right names for this, but I was wrong about the `to_xxx_aligned()` method names recently.

EdJoPaTo (approved):
Personally I try to create something only once I have all the parts for it together. So in this case I would have started with a `Vec<Line>` which I only put into the Text as a last step. But especially with the `push_span` these are more helpful in this case.

---

My mqttui displays JSON within my tree widget and a combination of both might be neat to have. (See [payload_view.rs](https://github.com/EdJoPaTo/mqttui/blob/494eece57eb2235bf410d1b25273c6f93dd8663f/src/interactive/details/payload_view.rs#L91-L129), [json.rs](https://github.com/EdJoPaTo/mqttui/blob/494eece57eb2235bf410d1b25273c6f93dd8663f/src/payload/json.rs) and [json_selector.rs](https://github.com/EdJoPaTo/mqttui/blob/494eece57eb2235bf410d1b25273c6f93dd8663f/src/payload/json_selector.rs)). Now I'm curious about improving the output of that to look more like yours (while maintaining the tree like navigation) 🙃

src/text/text.rs
```diff
@@ -441,6 +441,20 @@ impl<'a> Text<'a> {
...
         self.lines.iter_mut()
     }
+
+    /// Adds a line to the text.
+    pub fn push_line(&mut self, line: Line<'a>) {
+        self.lines.push(line);
+    }
+
+    /// Adds a span to the last line of the text.
+    pub fn push_span(&mut self, span: Span<'a>) {
```
EdJoPaTo, line 451: What about `Into<Line/Span>`?

```suggestion
    pub fn push_span<S: Into<Span<'a>>(&mut self, span: S) {
```

Then you could just add unstyled String for example.

joshka: Good idea. I'll make that change later tonight

joshka:
> Personally I try to create something only once I have all the parts for it together. So in this case I would have started with a Vec<Line> which I only put into the Text as a last step. But especially with the push_span these are more helpful in this case.

Yeah, that's reasonable I guess and for making things editable it might be the better approach for the JSON editor tutorial anyway especially because the selection needs to track which spans are editable and which are not. For that I'm trying to simplify the code needed to create an editor as much as possible. The idea there being more about showing how to make an app interactive than about making it pretty (but making it pretty is an important secondary goal).

I would probably use a tree widget in a real app for this sort of thing, but generally try to avoid third party things for tutorial code.

PR #1003, opened 2024-03-28
feat(layout): accept (x, y) tuple for Rect::offset

joshka:
This allows callers to call `Rect::offset((x, y))`` instead of the more
verbose `Rect::offset(Offset { x, y })`.

### DIFF
--- a/src/layout/rect.rs
+++ b/src/layout/rect.rs
@@ -40,6 +40,16 @@ pub struct Offset {
     pub y: i32,
 }
 
+impl<X: Into<i32>, Y: Into<i32>> From<(X, Y)> for Offset {
+    /// Creates a new `Offset` from a tuple of (x, y).
+    fn from((x, y): (X, Y)) -> Self {
+        Self {
+            x: x.into(),
+            y: y.into(),
+        }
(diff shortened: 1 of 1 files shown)


EdJoPaTo:
Personally I like being more explicit with things in order to ensure what is meant is actually happening. Autocompletion exists and pairs tend to create mistakes. I would either remove the Offset type and use `Rect::offset(x: u42, y: u42)` or require the offset to be used. Pairs only have types, no names so it might contain something. A simple struct (Which is already there: `Offset`) seems to be the better option in my opinion.

joshka:
> Personally I like being more explicit with things in order to ensure what is meant is actually happening. Autocompletion exists and pairs tend to create mistakes. I would either remove the Offset type and use `Rect::offset(x: u42, y: u42)` or require the offset to be used. Pairs only have types, no names so it might contain something. A simple struct (Which is already there: `Offset`) seems to be the better option in my opinion.

Thanks - this is the sort of push back that I was wondering whether would be there. I'm sort of two minds about it. It seemed neater to have this conversion, but I'm not certain.

PR #1007, opened 2024-03-30
refactor(buffer): deprecate assert_buffer_eq! in favor of assert_eq!

EdJoPaTo:
- Simplify `assert_buffer_eq!` logic.
- Deprecate `assert_buffer_eq!`.
- Introduce `TestBackend::assert_buffer_lines`.

Also simplify many tests involving buffer comparisons.

For the deprecation, just use `assert_eq` instead of `assert_buffer_eq`:

```diff
-assert_buffer_eq!(actual, expected);
+assert_eq!(actual, expected);
```

---

I noticed `assert_buffer_eq!` creating no test coverage reports and looked into this macro. First I simplified it. Then I noticed a bunch of `assert_eq!(buffer, …)` and other indirect usages of this macro (like `TestBackend::assert_buffer`).

The good thing here is that it's mainly used in tests so not many changes to the library code.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,16 +30,15 @@ cassowary = "0.3"
 compact_str = "0.7.1"
 crossterm = { version = "0.27", optional = true }
 document-features = { version = "0.2.7", optional = true }
-indoc = "2.0"
 itertools = "0.12"
 lru = "0.12.0"
 paste = "1.0.2"
 serde = { version = "1", optional = true, features = ["derive"] }
 stability = "0.2.0"
 strum = { version = "0.26", features = ["derive"] }
-time = { version = "0.3.11", optional = true, features = ["local-offset"] }
 termion = { version = "3.0", optional = true }
 termwiz = { version = "0.22.0", optional = true }
+time = { version = "0.3.11", optional = true, features = ["local-offset"] }
 unicode-segmentation = "1.10"
 unicode-truncate = "1"
 unicode-width = "0.1"
@@ -56,6 +55,7 @@ criterion = { version = "0.5.1", features = ["html_reports"] }
 derive_builder = "0.20.0"
 fakeit = "1.1"
 font8x8 = "0.3.1"
+indoc = "2"
 palette = "0.7.3"
 pretty_assertions = "1.4.0"
 rand = "0.8.5"
--- a/src/backend/test.rs
+++ b/src/backend/test.rs
@@ -9,7 +9,6 @@ use std::{
 use unicode_width::UnicodeWidthStr;
 
 use crate::{
-    assert_buffer_eq,
     backend::{Backend, ClearType, WindowSize},
     buffer::{Buffer, Cell},
     layout::{Rect, Size},
@@ -30,7 +29,7 @@ use crate::{
 ///
 /// let mut backend = TestBackend::new(10, 2);
 /// backend.clear()?;
-/// backend.assert_buffer(&Buffer::with_lines(vec!["          "; 2]));
+/// backend.assert_buffer_lines(["          "; 2]);
 /// # std::io::Result::Ok(())
 /// ```
 #[derive(Debug, Clone, Eq, PartialEq, Hash)]
@@ -97,11 +96,33 @@ impl TestBackend {
     }
 
     /// Asserts that the `TestBackend`'s buffer is equal to the expected buffer.
-    /// If the buffers are not equal, a panic occurs with a detailed error message
-    /// showing the differences between the expected and actual buffers.
+    ///
+    /// This is a shortcut for `assert_eq!(self.buffer(), &expected)`.
+    ///
+    /// # Panics
+    /// When they are not equal, a panic occurs with a detailed error message showing the
+    /// differences between the expected and actual buffers.
+    #[allow(deprecated)]
     #[track_caller]
     pub fn assert_buffer(&self, expected: &Buffer) {
-        assert_buffer_eq!(&self.buffer, expected);
+        // TODO: use assert_eq!()
+        crate::assert_buffer_eq!(&self.buffer, expected);
+    }
+
+    /// Asserts that the `TestBackend`'s buffer is equal to the expected lines.
+    ///
+    /// This is a shortcut for `assert_eq!(self.buffer(), &Buffer::with_lines(expected))`.
+    ///
+    /// # Panics
+    /// When they are not equal, a panic occurs with a detailed error message showing the
+    /// differences between the expected and actual buffers.
+    #[track_caller]
+    pub fn assert_buffer_lines<'line, Lines>(&self, expected: Lines)
+    where
+        Lines: IntoIterator,
+        Lines::Item: Into<crate::text::Line<'line>>,
+    {
+        self.assert_buffer(&Buffer::with_lines(expected));
     }
 }
 
@@ -246,7 +267,7 @@ mod tests {
             TestBackend {
                 width: 10,
                 height: 2,
-                buffer: Buffer::with_lines(vec!["          "; 2]),
+                buffer: Buffer::with_lines(["          "; 2]),
                 cursor: false,
                 pos: (0, 0),
             }
@@ -254,14 +275,14 @@ mod tests {
     }
     #[test]
     fn test_buffer_view() {
-        let buffer = Buffer::with_lines(vec!["aaaa"; 2]);
+        let buffer = Buffer::with_lines(["aaaa"; 2]);
         assert_eq!(buffer_view(&buffer), "\"aaaa\"\n\"aaaa\"\n");
     }
 
     #[test]
     fn buffer_view_with_overwrites() {
         let multi_byte_char = "👨‍👩‍👧‍👦"; // renders 8 wide
-        let buffer = Buffer::with_lines(vec![multi_byte_char]);
+        let buffer = Buffer::with_lines([multi_byte_char]);
         assert_eq!(
             buffer_view(&buffer),
             format!(
@@ -274,29 +295,27 @@ mod tests {
     #[test]
     fn buffer() {
         let backend = TestBackend::new(10, 2);
-        assert_eq!(backend.buffer(), &Buffer::with_lines(vec!["          "; 2]));
+        backend.assert_buffer_lines(["          "; 2]);
     }
 
     #[test]
     fn resize() {
         let mut backend = TestBackend::new(10, 2);
         backend.resize(5, 5);
-        assert_eq!(backend.buffer(), &Buffer::with_lines(vec!["     "; 5]));
+        backend.assert_buffer_lines(["     "; 5]);
     }
 
     #[test]
     fn assert_buffer() {
         let backend = TestBackend::new(10, 2);
-        let buffer = Buffer::with_lines(vec!["          "; 2]);
-        backend.assert_buffer(&buffer);
+        backend.assert_buffer_lines(["          "; 2]);
     }
 
     #[test]
     #[should_panic = "buffer contents not equal"]
     fn assert_buffer_panics() {
         let backend = TestBackend::new(10, 2);
-        let buffer = Buffer::with_lines(vec!["aaaaaaaaaa"; 2]);
-        backend.assert_buffer(&buffer);
+        backend.assert_buffer_lines(["aaaaaaaaaa"; 2]);
     }
 
     #[test]
@@ -312,7 +331,7 @@ mod tests {
         cell.set_symbol("a");
         backend.draw([(0, 0, &cell)].into_iter()).unwrap();
         backend.draw([(0, 1, &cell)].into_iter()).unwrap();
-        backend.assert_buffer(&Buffer::with_lines(vec!["a         "; 2]));
+        backend.assert_buffer_lines(["a         "; 2]);
     }
 
     #[test]
@@ -344,24 +363,19 @@ mod tests {
 
     #[test]
     fn clear() {
-        let mut backend = TestBackend::new(10, 4);
+        let mut backend = TestBackend::new(4, 2);
         let mut cell = Cell::default();
         cell.set_symbol("a");
         backend.draw([(0, 0, &cell)].into_iter()).unwrap();
         backend.draw([(0, 1, &cell)].into_iter()).unwrap();
         backend.clear().unwrap();
-        backend.assert_buffer(&Buffer::with_lines(vec![
-            "          ",
-            "          ",
-            "          ",
-            "          ",
-        ]));
+        backend.assert_buffer_lines(["    ", "    "]);
     }
 
     #[test]
     fn clear_region_all() {
         let mut backend = TestBackend::new(10, 5);
-        backend.buffer = Buffer::with_lines(vec![
+        backend.buffer = Buffer::with_lines([
             "aaaaaaaaaa",
             "aaaaaaaaaa",
             "aaaaaaaaaa",
@@ -370,19 +384,19 @@ mod tests {
         ]);
 
         backend.clear_region(ClearType::All).unwrap();
-        backend.assert_buffer(&Buffer::with_lines(vec![
+        backend.assert_buffer_lines([
             "          ",
             "          ",
             "          ",
             "          ",
             "          ",
-        ]));
+        ]);
     }
 
     #[test]
     fn clear_region_after_cursor() {
         let mut backend = TestBackend::new(10, 5);
-        backend.buffer = Buffer::with_lines(vec![
+        backend.buffer = Buffer::with_lines([
             "aaaaaaaaaa",
             "aaaaaaaaaa",
             "aaaaaaaaaa",
@@ -392,19 +406,19 @@ mod tests {
 
         backend.set_cursor(3, 2).unwrap();
         backend.clear_region(ClearType::AfterCursor).unwrap();
-        backend.assert_buffer(&Buffer::with_lines(vec![
+        backend.assert_buffer_lines([
             "aaaaaaaaaa",
             "aaaaaaaaaa",
             "aaaa      ",
             "          ",
             "          ",
-        ]));
+        ]);
     }
 
     #[test]
     fn clear_region_before_cursor() {
         let mut backend = TestBackend::new(10, 5);
-        backend.buffer = Buffer::with_lines(vec![
+        backend.buffer = Buffer::with_lines([
             "aaaaaaaaaa",
             "aaaaaaaaaa",
             "aaaaaaaaaa",
@@ -414,19 +428,19 @@ mod tests {
 
         backend.set_cursor(5, 3).unwrap();
         backend.clear_region(ClearType::BeforeCursor).unwrap();
-        backend.assert_buffer(&Buffer::with_lines(vec![
+        backend.assert_buffer_lines([
             "          ",
             "          ",
             "          ",
             "     aaaaa",
             "aaaaaaaaaa",
-        ]));
+        ]);
     }
 
     #[test]
     fn clear_region_current_line() {
         let mut backend = TestBackend::new(10, 5);
-        backend.buffer = Buffer::with_lines(vec![
+        backend.buffer = Buffer::with_lines([
             "aaaaaaaaaa",
             "aaaaaaaaaa",
             "aaaaaaaaaa",
@@ -436,19 +450,19 @@ mod tests {
 
         backend.set_cursor(3, 1).unwrap();
         backend.clear_region(ClearType::CurrentLine).unwrap();
-        backend.assert_buffer(&Buffer::with_lines(vec![
+        backend.assert_buffer_lines([
             "aaaaaaaaaa",
             "          ",
             "aaaaaaaaaa",
             "aaaaaaaaaa",
             "aaaaaaaaaa",
-        ]));
+        ]);
     }
 
     #[test]
     fn clear_region_until_new_line() {
         let mut backend = TestBackend::new(10, 5);
-        backend.buffer = Buffer::with_lines(vec![
+        backend.buffer = Buffer::with_lines([
             "aaaaaaaaaa",
             "aaaaaaaaaa",
             "aaaaaaaaaa",
@@ -458,19 +472,19 @@ mod tests {
 
         backend.set_cursor(3, 0).unwrap();
         backend.clear_region(ClearType::UntilNewLine).unwrap();
-        backend.assert_buffer(&Buffer::with_lines(vec![
+        backend.assert_buffer_lines([
             "aaa       ",
             "aaaaaaaaaa",
             "aaaaaaaaaa",
             "aaaaaaaaaa",
             "aaaaaaaaaa",
-        ]));
+        ]);
     }
 
     #[test]
     fn append_lines_not_at_last_line() {
         let mut backend = TestBackend::new(10, 5);
-        backend.buffer = Buffer::with_lines(vec![
+        backend.buffer = Buffer::with_lines([
             "aaaaaaaaaa",
             "bbbbbbbbbb",
             "cccccccccc",
@@ -496,19 +510,19 @@ mod tests {
         assert_eq!(backend.get_cursor().unwrap(), (4, 4));
 
         // As such the buffer should remain unchanged
-        backend.assert_buffer(&Buffer::with_lines(vec![
+        backend.assert_buffer_lines([
             "aaaaaaaaaa",
             "bbbbbbbbbb",
             "cccccccccc",
             "dddddddddd",
             "eeeeeeeeee",
-        ]));
+        ]);
     }
 
     #[test]
     fn append_lines_at_last_line() {
         let mut backend = TestBackend::new(10, 5);
-        backend.buffer = Buffer::with_lines(vec![
+        backend.buffer = Buffer::with_lines([
             "aaaaaaaaaa",
             "bbbbbbbbbb",
             "cccccccccc",
@@ -522,7 +536,7 @@ mod tests {
 
         backend.append_lines(1).unwrap();
 
-        backend.buffer = Buffer::with_lines(vec![
+        backend.buffer = Buffer::with_lines([
             "bbbbbbbbbb",
             "cccccccccc",
             "dddddddddd",
@@ -538,7 +552,7 @@ mod tests {
     #[test]
     fn append_multiple_lines_not_at_last_line() {
         let mut backend = TestBackend::new(10, 5);
-        backend.buffer = Buffer::with_lines(vec![
+        backend.buffer = Buffer::with_lines([
             "aaaaaaaaaa",
             "bbbbbbbbbb",
             "cccccccccc",
@@ -555,19 +569,19 @@ mod tests {
         assert_eq!(backend.get_cursor().unwrap(), (1, 4));
 
         // As such the buffer should remain unchanged
-        backend.assert_buffer(&Buffer::with_lines(vec![
+        backend.assert_buffer_lines([
             "aaaaaaaaaa",
             "bbbbbbbbbb",
             "cccccccccc",
             "dddddddddd",
             "eeeeeeeeee",
-        ]));
+        ]);
     }
 
     #[test]
     fn append_multiple_lines_past_last_line() {
         let mut backend = TestBackend::new(10, 5);
-        backend.buffer = Buffer::with_lines(vec![
+        backend.buffer = Buffer::with_lines([
             "aaaaaaaaaa",
             "bbbbbbbbbb",
             "cccccccccc",
@@ -580,19 +594,19 @@ mod tests {
         backend.append_lines(3).unwrap();
         assert_eq!(backend.get_cursor().unwrap(), (1, 4));
 
-        backend.assert_buffer(&Buffer::with_lines(vec![
+        backend.assert_buffer_lines([
             "cccccccccc",
             "dddddddddd",
             "eeeeeeeeee",
             "          ",
             "          ",
-        ]));
+        ]);
     }
 
     #[test]
     fn append_multiple_lines_where_cursor_at_end_appends_height_lines() {
         let mut backend = TestBackend::new(10, 5);
-        backend.buffer = Buffer::with_lines(vec![
+        backend.buffer = Buffer::with_lines([
             "aaaaaaaaaa",
             "bbbbbbbbbb",
             "cccccccccc",
@@ -605,19 +619,19 @@ mod tests {
         backend.append_lines(5).unwrap();
         assert_eq!(backend.get_cursor().unwrap(), (1, 4));
 
-        backend.assert_buffer(&Buffer::with_lines(vec![
+        backend.assert_buffer_lines([
             "          ",
             "          ",
             "          ",
             "          ",
             "          ",
-        ]));
+        ]);
     }
 
     #[test]
     fn append_multiple_lines_where_cursor_appends_height_lines() {
         let mut backend = TestBackend::new(10, 5);
-        backend.buffer = Buffer::with_lines(vec![
+        backend.buffer = Buffer::with_lines([
             "aaaaaaaaaa",
             "bbbbbbbbbb",
             "cccccccccc",
@@ -630,13 +644,13 @@ mod tests {
         backend.append_lines(5).unwrap();
         assert_eq!(backend.get_cursor().unwrap(), (1, 4));
 
-        backend.assert_buffer(&Buffer::with_lines(vec![
+        backend.assert_buffer_lines([
             "bbbbbbbbbb",
             "cccccccccc",
             "dddddddddd",
             "eeeeeeeeee",
             "          ",
-        ]));
+        ]);
     }
 
     #[test]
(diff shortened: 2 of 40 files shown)


EdJoPaTo:
> [Codecov](https://app.codecov.io/gh/ratatui-org/ratatui/pull/1007) Report

+4.53% coverage as the macro isn't creating unreached code anymore. :sparkles:

EdJoPaTo:
#849 will not be possible with this change.

joshka (approved):
LGTM

- [ ] I'd suggest fixing the "needless pass by value" lints rather than allowing them, but this is a small thing in comparison to the larger change.

I wonder how possible it would be to accept Into<Buffer> for the expected and then make it possible to just pass a bare Iterator / IntoIterator of Into<Line> values.

E.g.
```rust
buf.assert_eq(&Buffer::with_lines(vec!["  foo", " bar "]));
```
Becomes:
```rust
buf.assert_eq(["  foo", " bar "]);
// or perhaps it's meaningful or necessary to disambiguate:
buf.assert_lines([...]);
```

orhun (approved):

EdJoPaTo:
> I wonder how possible it would be to accept Into for the expected and then make it possible to just pass a bare Iterator / IntoIterator of Into values.

I am not sure if I like that idea. The idea of `assert_eq` is to make sure it's exactly what you expect it should be. `Into` can easily end up in mistakes as it's similar but not exact. I think this assertion should focus on being precise rather than quick to write. And when its really needed `.into()` isn't that far away either.

src/buffer/assert.rs
```diff
@@ -1,57 +1,36 @@
...
+            })
+            .collect::<Vec<String>>()
+            .join("\n");
+        assert!(
+            nice_diff.is_empty(),
+            "buffer contents not equal\nexpected: {expected:?}\nactual:   {actual:?}\ndiff:\n{nice_diff}",
+        );
+        // shouldn't get here, but this guards against future behavior
+        // that changes equality but not area or content
+        assert_eq!(actual, expected, "buffers not equal");
```
EdJoPaTo, line 29: Should the macro just use the assert_eq? I started with simplifying this but now its completely moved to the assert_eq function anyway so prevent code duplication until this deprecated macro is removed?

```suggestion
        actual.assert_eq!(expected);
```

EdJoPaTo:
The CI is making code up that is gone with this PR :ghost: 

https://shouldiblamecaching.com/

https://github.com/ratatui-org/ratatui/actions/runs/8501473392/job/23284430017#step:6:21

src/buffer/assert.rs
joshka: The entire rationale for this macro existing is built around two ideas:
1. When a test fails, the information reported should be enough to understand exactly how the buffers differ
2. When writing new tests, the output of the assertion should make it easy to write an intentionally failing assertion (`assert_buffer_eq!(actual, Buffer::empty(actual.area))`) and directly copy the result into the test once confirmed to be correct (with the exact number of spaces, newlines and quotes to make this seemless).

So long as both of those items are true, how the macro / function call works is mostly irrelevant.

It's possible that the first idea could be met by providing a better `Debug` representation that makes each line be displayed in a way that makes the style information easy to read and also fully copy-paste friendly. E.g. something like the following:
```rust
[
    Line::raw("A"),
    Line::styled("B", Color::Blue),
    Line::from([Span::raw("C"), Span::styled("D", Color::Red]),
    Line::styled("E", (Color::Black, Color::White, Modifier::BOLD | Modfier::Italic)),
].into()
```

If the debug output was that good, then we could just use `assert_eq!` throughout our tests instead of having a specific macro / function call.

This makes me think that perhaps it's worth making this function pub(crate) for now and having a stretch to get to the good debug output like that.

joshka:
> > I wonder how possible it would be to accept Into for the expected and then make it possible to just pass a bare Iterator / IntoIterator of Into values.
> 
> I am not sure if I like that idea. The idea of `assert_eq` is to make sure it's exactly what you expect it should be. `Into` can easily end up in mistakes as it's similar but not exact. I think this assertion should focus on being precise rather than quick to write. And when its really needed `.into()` isn't that far away either.

My rationale on this is that we have a huge amount of tests that call that `Buffer::with_lines()` method. The method really only exists for testing, and effectively is a point of repetition that could make the whole testing approach smaller and neater.

A well written test IMHO is one that can be grokked as correct just by looking at it. Such tests are an exercise in removing all details that are unnecessary to prove the desired assertion. If there's a large scale repetition like this, then it's really worth considering how things look when it's removed.

joshka:
P.s. I didn't realize the lint I asked you to fix would be affect such a large amount of code - my apologies.

src/buffer/assert.rs
EdJoPaTo: > 2\. When writing new tests, the output of the assertion should make it easy to write an intentionally failing assertion (`assert_buffer_eq!(actual, Buffer::empty(actual.area))`) and directly copy the result into the test once confirmed to be correct (with the exact number of spaces, newlines and quotes to make this seemless).

This is done by the custom Debug implementation already. The assert_buffer_eq macro doesnt do anything special for that case.

> If the debug output was that good, then we could just use `assert_eq!` throughout our tests instead of having a specific macro / function call.

I like that idea! Sadly `Buffer::diff` is used in the assertion to highlight the differences better. We can not achieve something like this with just the Debug implementation as it only knows one of the two buffers. Also, it would just notice the two buffers being different and then print the whole Debug output of the two buffers which isn't as easy to compare.

> How does the output look if you replace this with `actual.assert_eq`? If it's sane, then yep go for it.

The code is exactly the same, its duplicated. The output is also exactly the same. It's also the same compared to the main branch output.

EdJoPaTo:
I like that we are discussing alternatives to improve the usage of this!

---

Writing `Buffer::assert_lines` would be fairly simple:

```rs
impl Buffer {
    /// Asserts that this `Buffer` equals the expected lines
    pub fn assert_lines<'l, I>(&self, expected: I)
    where
        I: IntoIterator,
        I::Item: Into<Line<'l>>,
    {
        self.assert_eq(&Self::with_lines(expected));
    }
}
```

The usage is very similar:

```diff
-one.assert_eq(&Buffer::with_lines(["11", "11", "22", "22"]));
+one.assert_lines(["11", "11", "22", "22"]);
```

When the expected Buffer has more complicated styles this currently is written as `let mut expected = Buffer::with_lines(…); expected.set_style(…);`. These would need to use `.assert_eq` instead of `.assert_lines` which could create confusion as this differs. Also, many tests are using `Buffer::empty` or `Buffer::with_lines` in other places too.
I think there should be one way of achieving this instead of multiple ways that might create more confusion. Especially as this shortcut is only saving 18 characters. Often its on its own line anyway.

`assert_eq<B: Into<Buffer>>(&self, expected: B)` would also require the expected Buffer to be moved which is probably fine given that this is test code without performance relevance.
But I still think `Into` makes this code less clear as it does things without the user noticing it. Using `.into()` would be more clear.

Also side note: Into can sneak in stuff like clone() which is not obvious from looking at the code. For example https://github.com/EdJoPaTo/tui-rs-tree-widget/commit/b4d0c302b0fabc491286ba9a9da85ccbc6cb38d7 took a reference and into cloned it while the caller dropped the value later resulting in a superfluous clone on every render which was invisible from looking at the code. Therefore, I like to be more explicit instead of using into in order to not introduce hidden performance relevant mistakes.

---

Your `assert_eq!(lines()` example seems more annoying to write than the current implementation. Looking at my example above it would need to look like this:

```diff
-one.assert_eq(&Buffer::with_lines(["11", "11", "22", "22"]));
+assert_eq!(one.lines(), [
+    "11".into(),
+    "11".into(),
+    "22".into(),
+    "22".into(),
+]);
```

And this still does not highlight the exact diffs like the current implementation does with `Buffer::diff`.
Which got me thinking if `Buffer::diff` would be a good way of doing this (`assert_eq!(actual.diff(expected), []);`), but the output is less useful in that case as it does not display the Buffer Debug output. Also, it does not notice different area sizes.

---

I think the only real alternative is the usage of `assert_eq!()`. As a downside we would lose the diff output:

```diff
 ---- buffer::assert::tests::assert_buffer_eq_panics_on_unequal_area stdout ----
 thread 'buffer::assert::tests::assert_buffer_eq_panics_on_unequal_area' panicked at src/buffer/assert.rs:70:9:
-buffer areas not equal
-expected:  Buffer {
+assertion `left == right` failed
+  left: Buffer {
     area: Rect { x: 0, y: 0, width: 6, height: 1 },
     content: [
         "      ",
     ],
     styles: [
         x: 0, y: 0, fg: Reset, bg: Reset, underline: Reset, modifier: NONE,
     ]
 }
-actual:    Buffer {
+ right: Buffer {
     area: Rect { x: 0, y: 0, width: 5, height: 1 },
     content: [
         "     ",
     ],
     styles: [
         x: 0, y: 0, fg: Reset, bg: Reset, underline: Reset, modifier: NONE,
     ]
 }

 ---- buffer::assert::tests::assert_buffer_eq_panics_on_unequal_style stdout ----
 thread 'buffer::assert::tests::assert_buffer_eq_panics_on_unequal_style' panicked at src/buffer/assert.rs:78:9:
-buffer contents not equal
-expected: Buffer {
+assertion `left == right` failed
+  left: Buffer {
     area: Rect { x: 0, y: 0, width: 5, height: 1 },
     content: [
         "     ",
     ],
     styles: [
         x: 0, y: 0, fg: Red, bg: Reset, underline: Reset, modifier: NONE,
         x: 1, y: 0, fg: Reset, bg: Reset, underline: Reset, modifier: NONE,
     ]
 }
-actual: Buffer {
+ right: Buffer {
     area: Rect { x: 0, y: 0, width: 5, height: 1 },
     content: [
         "     ",
     ],
     styles: [
         x: 0, y: 0, fg: Reset, bg: Reset, underline: Reset, modifier: NONE,
     ]
 }
-diff:
-0: at (0, 0)
-  expected: Cell { symbol: " ", fg: Red, bg: Reset, underline_color: Reset, modifier: NONE, skip: false }
-  actual:   Cell { symbol: " ", fg: Reset, bg: Reset, underline_color: Reset, modifier: NONE, skip: false }
```

This is fine for smaller Buffer comparisons but gets annoying for bigger diffs. For example the Map widget tests have hundreds of Cells. Finding a single difference in there is annoying.

---

So… I would stick with the `Buffer::assert_eq` function. `assert_eq!()` would also do the job but is less informative.

EdJoPaTo:
As there hasn't been a better idea so far… What's the status on this? I would like to see this merged in some way to use the new syntax in other places. This also fixes #1043 as a bonus.

joshka:
It feels odd to me that test code is hanging off the Buffer.

I think this might feel more appropriate as a standalone function to be imported rather than a method in the Buffer implementation. I haven't ever seen other places where the test code ends up as a method, and a function to be imported seems like a reasonable compromise. Thoughts?

To be clear, I'm suggesting:
```rust
fn assert_eq(left: Buffer, right: Buffer) {}
```
or assert_buffer_eq. Either seems fine, assert_eq is shorter, which makes tests easier to read, but could be mistaken for assert_eq! easily.

Another approach to keeping the assert_eq out of the public could be 
```rust
trait AssertEq {
   fn assert_eq(&self, other: &Self);
}
impl AssertEq for Buffer { ...}
```

I really like that this removes 600 LoC. That's a bunch less to think about.

I'd really like to simplify the assertions more - perhaps `fn assert_buffer_lines()` would be appropriate - no magic, just a standard shortcut for what is currently (with this change) `buf.assert_eq(&Buffer::with_lines([...])`

By keeping this out of the buffer and as a separate function, it makes it obvious where other similar functions like that would end up (rather than bloating `Buffer`'s namespace).

src/widgets/barchart.rs
```diff
@@ -700,15 +685,12 @@ mod tests {
             .data(&[("foo", 1), ("bar", 2)])
             .bar_style(Style::new().red());
         widget.render(buffer.area, &mut buffer);
-        let mut expected = Buffer::with_lines(vec![
-            "  █            ",
-            "1 2            ",
-            "f b            ",
-        ]);
+        let mut expected =
+            Buffer::with_lines(["  █            ", "1 2            ", "f b            "]);
```
joshka, line 689: The buffer width in a bunch of tests was chosen so that rustfmt can wrap nicely while maintaining a 2D view of what the buffer looks like visually. If that breaks here, then the buffer should be wider / taller / use a macro to ensure that this doesn't wrap.

(The intent behind this is that every rendering test looks visually correct at a glance)

src/widgets/block.rs
```diff
@@ -1367,15 +1345,11 @@ mod tests {
             .border_style(Style::new().yellow())
             .render(buffer.area, &mut buffer);
 
-        let mut expected_buffer = Buffer::with_lines(vec![
-            "┌test─────────┐",
-            "│             │",
-            "└─────────────┘",
-        ]);
+        let mut expected_buffer =
+            Buffer::with_lines(["┌test─────────┐", "│             │", "└─────────────┘"]);
```
joshka, line 1349: Broken formatting here

src/widgets/paragraph.rs
```diff
@@ -537,23 +526,19 @@ mod test {
         for paragraph in paragraphs {
             test_case(
                 paragraph,
-                Buffer::with_lines(vec![
-                    "┌Title─────────┐",
-                    "│Hello, worlds!│",
-                    "└──────────────┘",
-                ]),
+                &Buffer::with_lines(["┌Title─────────┐", "│Hello, worlds!│", "└──────────────┘"]),
```
joshka, line 529: formatting breaks

tests/stylize.rs
```diff
@@ -73,12 +72,8 @@ fn block_can_be_stylized() -> io::Result<()> {
...
     })?;
 
-    let mut expected = Buffer::with_lines(vec![
-        "┌Title─┐   ",
-        "│      │   ",
-        "└──────┘   ",
-        "           ",
-    ]);
+    let mut expected =
+        Buffer::with_lines(["┌Title─┐   ", "│      │   ", "└──────┘   ", "           "]);
```
joshka, line 76: format break

tests/widgets_block.rs
```diff
@@ -208,182 +205,121 @@ fn widgets_block_title_alignment() {
...
     // title top-left with all borders
     test_case(
         Alignment::Left,
         Borders::ALL,
-        Buffer::with_lines(vec![
-            " ┌Title──────┐ ",
-            " │           │ ",
-            " └───────────┘ ",
-        ]),
+        &Buffer::with_lines([" ┌Title──────┐ ", " │           │ ", " └───────────┘ "]),
```
joshka, line 216: format break (I'm not going to mark more of these, perhaps run a regex over the diff to find them all.

EdJoPaTo:
> It feels odd to me that test code is hanging off the Buffer.
> 
> I think this might feel more appropriate as a standalone function to be imported rather than a method in the Buffer implementation. I haven't ever seen other places where the test code ends up as a method, and a function to be imported seems like a reasonable compromise. Thoughts?

`Buffer::assert_eq(actual, expected)` is possible with this too.

I think having this existing on the `impl Buffer` is the perfect place for it as it makes it more clear where its relevant: on the Buffer.

It could be explicitly not on self, but I'm not sure why as its kinda clear to me what is happening in that case as it is used on a buffer.

> bloating `Buffer`'s namespace

Like prelude is bloating the users namespace which results in uncertainty where something is coming from and what is used? :fire: 

> Another approach to keeping the assert_eq out of the public could be [trait]

I don't think this should be kept out of public code as people either don't use Buffer directly (just passing it around where its needed) or use it when creating widgets, and it's really helpful for testing your code works so it should be discoverable. Maybe add to the doc comment that this is helpful for test cases and add an example for that in there?

> I'd really like to simplify the assertions more - perhaps `fn assert_buffer_lines()` would be appropriate - no magic, just a standard shortcut for what is currently (with this change) `buf.assert_eq(&Buffer::with_lines([...])`

https://github.com/ratatui-org/ratatui/pull/1007#issuecomment-2029975480

joshka:
> `Buffer::assert_eq(actual, expected)` is possible with this too.

I think this idea works great. It's super clear what it's doing, while still being similar enough to the existing `assert_eq!` macro to drop it in without much effort when the macro doesn't  hit the spot. If you're also ok with it, let's settle on that.

> Like prelude is bloating the users namespace which results in uncertainty where something is coming from and what is used? 🔥

This only bloats namespaces for users that opt in to using the prelude. I'd assume your apps / widgets would be marked `#[safe_from_prelude_bloat_today]` [;P](https://knowyourmeme.com/memes/marked-safe-from-x)

> https://github.com/ratatui-org/ratatui/pull/1007#issuecomment-2029975480

Perhaps a followup PR for that change is in order - let's get this over the line and I'll slap one together.

EdJoPaTo:
> > [#1007 (comment)](https://github.com/ratatui-org/ratatui/pull/1007#issuecomment-2029975480)
> 
> Perhaps a followup PR for that change is in order - let's get this over the line and I'll slap one together.

Just to make sure I understand you correctly there. I preferred not to add this as this isn’t much shorter and less clear what is happening. Your follow up PR still wants to add it?

EdJoPaTo:
> This only bloats namespaces for users that opt in to using the prelude.

Sadly I also code inside of ratatui which makes a lot of use of it… 👻

EdJoPaTo:
> > `Buffer::assert_eq(actual, expected)` is possible with this too.
> 
> I think this idea works great.

This is already possible. So should I change the usage of it?

Keeping the &self allows for using that on the buffer directly too.
Maybe add this to the doc comment as an example? On the other hand this diverges the usage of it. Having only one way to use it would be better…

joshka:
> Sadly I also code inside of ratatui which makes a lot of use of it… 👻

I wonder if our differing perspectives on the point of prelude come from which languages we predominantly used prior to Rust. My primary work related technologies were Java and C# for the last 20 years or so, with a bunch of different languages thrown in on top (Javascript, Ruby, etc.)?

> Just to make sure I understand you correctly there. I preferred not to add this as this isn’t much shorter and less clear what is happening. Your follow up PR still wants to add it?

It's probably not a high priority item. It's a maybe sometime thing.

EdJoPaTo:
Looking at the diff of this… not sure I like it as it's less concise afterwards. Basically wants me to have the TestBackend::assert_buffer back because of it being more concise. Not sure what to think of it.

https://github.com/ratatui-org/ratatui/blob/baedc39494ea70292b1d247934420a20d0544b7e/src/backend/test.rs#L103

(I know the formatting is broken, I want get the usage of this right first.)

> > `Buffer::assert_eq(actual, expected)` is possible with this too.
> 
> I think this idea works great. It's super clear what it's doing, while still being similar enough to the existing `assert_eq!` macro to drop it in without much effort when the macro doesn't hit the spot. If you're also ok with it, let's settle on that.

Also notice that I haven't touched the implementation of this yet. Which means that `assert_eq!()` can easily replaced by `Buffer::assert_eq` too. So being able to change that easily would be still the case with allowing to use `self.assert_eq`.

> I really like that this removes 600 LoC. That's a bunch less to think about.

With fixing all the formattings this is likely gone then again 🙃

joshka:
I think this likely points to the idea that assert_eq should just be a method in the buffer module. From an ergonomics perspective, I think having a test that does:
```rust
use ratatui::buffer::assert_eq;

#[test]
fn some_test() {
    let buf = Buffer::empty(Rect::new(0, 0, 10, 3));
    assert_eq(
        buf,
        Buffer::with_lines([
            "          ",
            "          ",
        ]);
}

// or

use ratatui::buffer::assert_lines_eq

#[test]
fn some_test_with_lines() {
    let buf = Buffer::empty(Rect::new(0, 0, 10, 3));
    assert_lines_eq(
        buf,
        [
            "          ",
            "          ",
        ]);
}
```

Both seem pretty reasonable. That said, `Buffer::empty` probably should just be `Buffer::new`, and with_lines could easily be replaced with a `FromIter` implementation.

EdJoPaTo:
> I think this likely points to the idea that assert_eq should just be a method in the buffer module.

That wouldn’t change much. It would be `Buffer::assert_eq` vs the imported `assert_buffer_eq`.

The big benefit in brevity came from `buffer.assert_eq(expected)`.

Having `Buffer::empty` and `with_lines` is pretty clear to me what it does. IntoIter would result in less clear code. Also Buffer is only created internally or for tests…

Also, repeating myself, assert_lines would end up with more annoying style testing so hiding the creation completely would also be counter productive.

kdheepak:
fwiw, I like the idea of putting the `assert_eq` in a trait:

```
trait AssertEq {
   fn assert_eq(&self, other: &Self);
}
impl AssertEq for Buffer { ...}
```

Maybe the trait can also be implemented for other types like `Span` or `Line`?

joshka:
> The big benefit in brevity came from `buffer.assert_eq(expected)`.

Yeah, you're right about that. the buffer.assert_eq syntax is better.

> fwiw, I like the idea of putting the `assert_eq` in a trait:

I'm in agreement with @kdheepak on this one. It just feels odd to me to have testing related methods on types. Do you think making this a trait would be a reasonable compromise?

> Maybe the trait can also be implemented in the future for other types like `Span` or `Line`?

I'd be inclined to make this `ratatui::buffer::AssertEq` for now.

Just a reminder to ground this idea, the reason `assert_buffer_eq` exists is purely because the diff between buffers is difficult to visually inspect when there's an error, and the output is difficult to copy and paste from to update a broken test in a TDD approach. At steady state, this is entirely unnecessary, it's only when developing a new test or fixing a broken one when this matters. If the debug format was a bit less verbose, and we used something like [pretty_assertions](https://crates.io/crates/pretty-assertions), then the problem might be mostly solved.

EdJoPaTo:
> Just a reminder to ground this idea, the reason `assert_buffer_eq` exists is purely because the diff between buffers is difficult to visually inspect when there's an error, and the output is difficult to copy and paste from to update a broken test in a TDD approach. At steady state, this is entirely unnecessary, it's only when developing a new test or fixing a broken one when this matters.

I think this is an argument against using a more generic trait and going with a simple implementation. This shouldn't lead people to creating more things like this. This is a workaround around the `assert_eq!()` being hard to understand for Buffers. Ideally it does not exist for buffer nor other things so there shouldn't be a generic.

Going with `pretty_assertions::assert_eq` might also be an alternative approach interesting to look into. `assert_eq!` is hard to read when colours are involved. When its only the content that is different it's relatively easy to spot. Especially when test cases are small representations and don't use larger buffers to make their point.
Trying to get it look the same in source code might also work against getting smaller diffs… Not sure what I think of that yet.

EdJoPaTo:
funny realisation: `self.assert_lines(lines)` is even shorter and would require bigger example buffers to ensure the formatting in the sources stays similar to rendering…

EdJoPaTo:
Thought about adding `Buffer::assert_eq(&self, other: AsRef<Self>)` but that only removes one & per line. its not that significant and it results in less clear code that the input is only taken by reference. Also it requires a dummy `impl AsRef<Self> for Buffer` taking up 5 lines… don't think it's really worth it in comparison to just one `&`.

joshka:
> Thought about adding `Buffer::assert_eq(&self, other: AsRef<Self>)` but that only removes one & per line. its not that significant and it results in less clear code that the input is only taken by reference. Also it requires a dummy `impl AsRef<Self> for Buffer` taking up 5 lines… don't think it's really worth it in comparison to just one `&`.

As it's >180 refs, I think that might be worth including. Is there any downside to that impl?

> `assert_eq!` is hard to read when colours are involved. When its only the content that is different it's relatively easy to spot

Interestingly if you do give pretty_assertions an ansi string, it will render it as colors (though it will overwrite the colors with its red/green diff colors):

```rust
use pretty_assertions::assert_eq;
#[test]
fn to_string() {
    let mut buf = AnsiStringBuffer::new(5, 2);
    buf.render_ref(&Line::styled("Hello", Color::Blue), Rect::new(0, 0, 5, 1));
    buf.render_ref(&Line::styled("World", Color::Green), Rect::new(0, 1, 5, 1));
    assert_eq!(
        buf.to_string(),
        "\u{1b}[34;49mHello\n\u{1b}[32;49mMars \u{1b}[0m"
    );
}
```
image

Experimenting with a style diff:

image

The main hassle with this output it doesn't easily lend itself to being copied back into the test.

When writing test code that interacts with the buffer (widgets etc.), my usual (Test after Dev) flow is:

1. write a broken test that asserts that the buffer is an empty buffer
2. look at the output of the test and confirm that it looks correct
3. copy the output of the test directly into the test

This leads to a pretty fast and easy approach to going from nothing to tested. And the effectiveness of this is proven out by the increase in the quantity and quality of unit tests in Ratatui even as we've added a significant amount of code.

For interest sake, I ran coverage (`cargo llvm-cov --open --all-targets`) on v0.19.0 (tui-rs) and main.

Before we started improving this:
image

Current situation:
image
image

Summary, we added ~10k LoC, and increased the test coverage by ~10k LoC.
This number is probably higher as the coverage bump we get from this change is pretty big.

EdJoPaTo:
> When writing test code that interacts with the buffer (widgets etc.), my usual (Test after Dev) flow is:
> 
>     1. write a broken test that asserts that the buffer is an empty buffer
>     2. look at the output of the test and confirm that it looks correct
>     3. copy the output of the test directly into the test
> 
> This leads to a pretty fast and easy approach to going from nothing to tested. And the effectiveness of this is proven out by the increase in the quantity and quality of unit tests in Ratatui even as we've added a significant amount of code.

This also works with the native `assert_eq!()` as the output is from Debug, not from the `assert_buffer_eq`.

So maybe the best approach is to remove the dedicated buffer assert stuff all together?

Adding a `#[rustfmt::skip]` exactly at the wrapping thing prevents it from happening while the rest of the code is still formatted:

```rs
#[rustfmt::skip]
    " a",
    " b",
    " c",
]);

let expected = Buffer::with_lines([" a", " b", " c"]);
```

This way the buffer content can get way smaller resulting in smaller Debug outputs and therefore easier to spot differences for test cases. Then the need for a special macro for this is even less important.

If that still requires a diff as it's still complicated to spot the difference then a (temporary?) `dbg!(actual.diff(expected))` is not that far away.

That would remove the somewhat weird custom `assert_eq!()` implementation for the buffer making use of standard Rust while addressing the pain point of bigger diffs due to having bigger than necessary buffers in code due to code formatting prevention.

joshka:
To make that work, the debug format also needs to be much less verbose (but in a way that makes it easy to identify differences).

For the most part, the visual inspection approach applies to multi-line output in the buffers. I'm not opposed to `rustfmt::skip`, I used the approach of longer lines before I knew that existed.

> If that still requires a diff as it's still complicated to spot the difference then a (temporary?) dbg!(actual.diff(expected)) is not that far away.

Makes sense I guess.

Perhaps give it a try and see how it looks with just `assert_eq!`

src/backend/test.rs
```diff
@@ -97,11 +96,33 @@ impl TestBackend {
...
+    ///
+    /// # Panics
+    /// When they are not equal, a panic occurs with a detailed error message showing the
+    /// differences between the expected and actual buffers.
+    #[allow(deprecated)]
     #[track_caller]
     pub fn assert_buffer(&self, expected: &Buffer) {
-        assert_buffer_eq!(&self.buffer, expected);
+        // TODO: use assert_eq!()
+        crate::assert_buffer_eq!(&self.buffer, expected);
```
joshka, line 109: Why the TODO for this?

src/buffer/assert.rs
```diff
@@ -1,57 +1,44 @@
 /// Assert that two buffers are equal by comparing their areas and content.
 ///
-/// On panic, displays the areas or the content and a diff of the contents.
+/// # Panics
+/// When the buffers differ this method panics and displays the differences similar to
+/// `assert_eq!()`.
+#[deprecated = "use actual.assert_eq(&expected)"]
```
joshka, line 6:
```suggestion
#[deprecated = "use assert_eq!"]
```

src/text/masked.rs
```diff
@@ -46,14 +46,14 @@ impl<'a> Masked<'a> {
 impl fmt::Debug for Masked<'_> {
     /// Debug representation of a masked string is the underlying string
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.write_str(&self.inner)
+        fmt::Display::fmt(&self.inner, f)
```
joshka, line 49: What makes this better / necessary? It seems more complex for no obvious reason (and below).

joshka (approved):
Great work in wrangling all the tests into a really consistent form and making them into rstests when appropriate. I went through the entire PR with a pretty detailed view, and only found the comments previously raised to get answers to / fixes.

I think this is ready to merge, but I'm 50/50 on whether to wait til after 0.26.3 to merge (so we can get out the unicode line bug fix).
Pros: (less carrying cost of large PR, less future PR conflicts)
Cons: (potential breakage so close to releasing a patch release)

Your call on whether to merge this.

src/text/masked.rs
EdJoPaTo: This is a relic from before #1087 was created / merged. Basically: passing the object to its inner Debug/Display implementation.

`format!("{:>5}", "bla")` will be rendered as `"  bla"` when the padding stuff is correctly used. `write_str` ignores all this stuff and just renders `"bla"`. `f.pad(&self.inner)` would take into account more of this but it's simpler to just use the implementation of the inner one to have exactly the features it has.

src/backend/test.rs
EdJoPaTo: Changing it breaks tests as they expect panics with the current assert_buffer_eq!() output. We could just adapt the expected output?

src/buffer/assert.rs
EdJoPaTo: Good catch!
```suggestion
#[deprecated = "use assert_eq!(&actual, &expected)"]
```

src/text/masked.rs
joshka: I think this probably needs tests so someone doesn't simplify it. It's also Worth pulling out of this PR, which is only refactoring changes, into another one that does actual changes.

src/backend/test.rs
joshka: I think the impact of breaking those might be in user code that uses the macro, so let's do that as a follow up instead.

EdJoPaTo:
> I think this is ready to merge, but I'm 50/50 on whether to wait til after 0.26.3 to merge (so we can get out the unicode line bug fix).
>
> Pros: (less carrying cost of large PR, less future PR conflicts)

Future PR conflicts is not really an argument as stuff isn't branched of release tags, they are branched of from the default branch.

> Cons: (potential breakage so close to releasing a patch release)

This changes mostly tests. Only the macro itself is a non-test change. And it fixes #1043.

I could see why removing the deprecation might be a good idea until after the patch release, but it's really just a deprecation. `#[allow(deprecation)]` is an easy workaround. Nothing is panicking on runtime without any heads up.

> Your call on whether to merge this.

Therefore, I would go for merging this (once the open points are all resolved).

src/text/masked.rs
EdJoPaTo: Removed from this PR

Cargo.toml
```diff
@@ -30,16 +30,15 @@ cassowary = "0.3"
...
 itertools = "0.12"
 lru = "0.12.0"
 paste = "1.0.2"
 serde = { version = "1", optional = true, features = ["derive"] }
 stability = "0.2.0"
 strum = { version = "0.26", features = ["derive"] }
-time = { version = "0.3.11", optional = true, features = ["local-offset"] }
 termion = { version = "3.0", optional = true }
 termwiz = { version = "0.22.0", optional = true }
+time = { version = "0.3", optional = true, features = ["local-offset"] }
```
joshka, line 41: https://docs.rs/time/0.3.9/time/struct.Date.html#method.replace_day was introduced in 0.3.9 and is used in the calendar widget. Not sure if there's a specific need for something from the 0.3.11 version, but the downgrading that dep is out of scope for this PR.

`cargo +nightly update -Zdirect-minimal-versions && cargo check --all-features --all-targets` suggests we actually need:
```
itertools 0.12.1
serde 1.0.166
```

EdJoPaTo: The main change was the sorting. Personally I don't like seeing the fix version here as it doesn't change anything for the version resolution. (`3` and `3.0.1` will use `3.4.5`). Either the one from the lock file or the latest one is used.

With Rust edition 2024 taking MSRV into account it might actually be useful but before that it isn't.

And as a lot of crates interpret SemVer differently it's a mess anyway… `0.3` indicates it's not stable yet anyway so that's another mess by itself…

This PR is not about the version ranges so maybe just add the pointless bugfix version to skip this discussion?

joshka: 8aa7c6c9e

EdJoPaTo:
Any improvements for the first message (above the line) as a commit message? When not, this should be merged.

src/buffer/assert.rs
EdJoPaTo: Settled on assert_eq and a good Debug impl.

joshka:
> Any improvements for the first message (above the line) as a commit message? When not, this should be merged.

LGTM

Issue #931, opened 2024-02-06
Better `Margin` features with `Rect`


kdheepak:
## Problem

I want to be able to write `area.inner(Margin::left(3))`.

But `Margin` has just two fields, `horizontal` and `vertical` instead of `left`, `right`, `top`, `bottom`, and these two fields have been public

EdJoPaTo:
[`Padding`](https://docs.rs/ratatui/0.27.0/ratatui/widgets/block/struct.Padding.html) could be merged with `Margin`

PR #1008, opened 2024-03-30
perf(rect)!: `Rect::inner` takes `Margin` directly instead of reference

EdJoPaTo:
BREAKING CHANGE: Margin needs to be passed without reference now.

```diff
-let area = area.inner(&Margin {
+let area = area.inner(Margin {
     vertical: 0,
     horizontal: 2,
 });
```


joshka (changes requested):
LGTM as a breaking change
- [x] Update the PR message to be changelog focused (e.g. tell the user how they need to change their apps.) Something like "Rect::inner() now takes Margin instead of &Margin"
- [ ] Add the info to the breaking change doc
- [x] Add the breaking change section of the commit (https://www.conventionalcommits.org/)
- [x] Make sure there's a small example of how to fix the change in both places (this is overkill for such a small change, but doing it consistently makes it easy for users upgrading to search for this sort of thing easily)

Hold off merging until 0.27.0 release is ready to go after 0.26.2

joshka (approved):
Changes look good. We can merge this when we go to 0.27.0

I think the next version will probably be 0.26.2 and then have all breaking changes batched up in 0.27.0

joshka:
> Speaking of `0.26.2`, let's make that happen soon.

https://github.com/ratatui-org/ratatui/issues/1025

joshka:
BTW, I expect this will probably break a lot of projects (520 results):

https://github.com/search?q=ratatui+%22.inner%28%22+AND+%28NOT+org%3Aratatui-org%29+AND+%28NOT+is%3Afork%29&type=code

kdheepak:
I'd like for this change to be made because it'll make it consistent with a future feature like: https://github.com/ratatui-org/ratatui/issues/931

joshka:
> I'd like for this change to be made because it'll make it consistent with a future feature like: #931

Sounds like we should break both things at the same time to avoid double breaks.

joshka (approved):

joshka:
Can you please update the PR message to fixup the notes about Rect that are no longer part of this PR?

PR #1010, opened 2024-03-30
perf(scrollbar)!: pass symbol set by reference

EdJoPaTo:
A symbol set contains `&'static str`. It itself only needs to be some reference as whatever is in it has to be `'static` anyway, and only the inner values are used after the method ends, and the set is out of scope again.

BREAKING CHANGE: `Scrollbar::symbols` used to move the value and now only needs a reference.

```diff
 Scrollbar::new(ScrollbarOrientation::VerticalLeft)
-    .symbols(ratatui::symbols::scrollbar::VERTICAL)
+    .symbols(&ratatui::symbols::scrollbar::VERTICAL)
```

### DIFF
--- a/examples/scrollbar.rs
+++ b/examples/scrollbar.rs
@@ -182,7 +182,7 @@ fn ui(f: &mut Frame, app: &mut App) {
     f.render_widget(paragraph, chunks[2]);
     f.render_stateful_widget(
         Scrollbar::new(ScrollbarOrientation::VerticalLeft)
-            .symbols(scrollbar::VERTICAL)
+            .symbols(&scrollbar::VERTICAL)
             .begin_symbol(None)
             .track_symbol(None)
             .end_symbol(None),
--- a/src/widgets/scrollbar.rs
+++ b/src/widgets/scrollbar.rs
@@ -68,15 +68,15 @@ use crate::{prelude::*, symbols::scrollbar::*};
 /// # }
 /// ```
 #[derive(Debug, Clone, Eq, PartialEq, Hash)]
-pub struct Scrollbar<'a> {
+pub struct Scrollbar<'symbols> {
     orientation: ScrollbarOrientation,
     thumb_style: Style,
-    thumb_symbol: &'a str,
+    thumb_symbol: &'symbols str,
     track_style: Style,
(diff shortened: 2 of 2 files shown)


orhun:
I'm curious about the performance aspect of this, do we have any benchmarks in place that can be used to get an idea of that?

EdJoPaTo:
> I'm curious about the performance aspect of this, do we have any benchmarks in place that can be used to get an idea of that?

I don't think this one is really significant as these are very short `str` which are likely set on compile. LLVM is very likely optimizing this away anyway.
Other things like many items are far more significant. A scrollbar is only created a few times per render, not hundreds or thousands.

So I think a change like this will be insignificant in benchmarks.

clippy notices this as something that can be improved, and I would like to enable the lints for public facing interfaces (→ breaking lints) to catch more important mistakes in the future.

So tldr: unlikely to be significant but something an automated lint finds.

joshka (approved):
LGTM.
Before we merge this, can you please:
- [x] add a small before / after example to the BREAKING CHANGES part of the PR message
- [ ] Add an entry in the BREAKING_CHANGES.md doc
- [x] I added a ! to the PR title (see https://www.conventionalcommits.org/en/v1.0.0/#specification)

Let's also consider releasing the current source as a non-breaking 0.26.2 before we merge this as it make this 0.27.0. (The only breaking change currently sitting in main right now is an MSRV bump which we're not generally considering as breaking).

orhun (approved):

EdJoPaTo:
As I think the current widgets::Scrollbar should exist in the first place (#174) and the main goal of this is to enable the breaking change lints (#988) I would just go with an ignore comment. As pointed out earlier this change is pretty much pointless for performance and a breaking change for something that needs major refactoring in the first place is also somewhat pointless.

I would close this PR and keep the allow lint rule in #988.

joshka:
> I would close this PR and keep the allow lint rule in #988.

Seems reasonable. It might be worth revisiting if there's some reason to build an app where the symbols for scrollbars are more configurable, but even then, the size of the strings involved are small enough that calling clone() on a symbol set is unlikely to be a bottleneck most apps.

PR #1019, opened 2024-04-04
feat: Create `is_at_bottom()` function

### LINKED ISSUE #1018: Add `ScrollbarState.is_at_start()`, `ScrollbarState.is_at_end()` methods
## Problem
I want to have logic that's something like:

```rust
let was_at_bottom = sb_state.is_at_bottom()

// ... a bunch of code that changes the amount of content in the paragraph controlled by the scrollbar

if was_at_bottom {
    sb_state.last()
}
```

The idea is that I want the scrollbar section to always be at its bottom, but to also be able to be scrolled up without immediately jumping back to the bottom.

## Solution
I think it might be as simple as adding a new function in `src/widgets/scrollbar.rs` with the following content:

```rust
pub fn is_at_bottom(&self) -> bool {
      self.position == self.content_length.saturating_sub(1)
  }
```

## Alternatives
* Expose the private `ScrollbarState` struct attribute `ScrollbarState.content_length`
* Also, maybe expose the private struct attribute `ScrollbarState.position`

DeflateAwning:
Fixes #1018 very crudely. Seems to work in my project though. Super open to feedback overall though!

### DIFF
--- a/src/widgets/scrollbar.rs
+++ b/src/widgets/scrollbar.rs
@@ -472,6 +472,11 @@ impl ScrollbarState {
             }
         }
     }
+
+    /// Returns `true` if the scrollbar is at the bottom.
+    pub fn is_at_bottom(&self) -> bool {
+        self.position == self.content_length.saturating_sub(1)
+    }
 }
 
 impl<'a> StatefulWidget for Scrollbar<'a> {


orhun:
Would be nice to add tests for this as well.

joshka:
Seems reasonable. What about changing this to be is_at_beginning / is_at_end to match with the begin/end terms for the symbols?

DeflateAwning:
I agree with the rename (and thought of it too, just as I clicked to open the PR). Will discuss merit in the issue.

DeflateAwning:
Closing this one entirely; will open a new one.

joshka:
Hey there. As a quick aside for future PRs, it's often better to keep PRs open and adjust them when necessary by rebasing / force pushing changes as capturing the flow of comments in a single place helps reduce the quantity of places devs have to go to get the complete story on a PR.

PR #1024, opened 2024-04-10
Add `is_at_start` and `is_at_end` methods

### LINKED ISSUE #1018: Add `ScrollbarState.is_at_start()`, `ScrollbarState.is_at_end()` methods

DeflateAwning:
Fixes #1018.

This is still a work in progress. Looking for help/feedback with the tests. 

For this to work reliably and as expected, I believe that the `Scroll` object needs to "write back" to this `ScrollbarState` object, so that this `ScrollbarState` understands its viewport size. This communication is at the heart of most of the scrollbar deficiencies, as far as I can tell.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,7 +20,7 @@ description to help the reviewer (follow the provided template). Make sure to hi
 which may need additional attention or you are uncertain about. Any idea with a large scale impact
 on the crate or its users should ideally be discussed in a "Feature Request" issue beforehand.
 
-### Keep PRs small, intentional and focused
+### Keep PRs small, intentional, and focused
 
 Try to do one pull request per change. The time taken to review a PR grows exponential with the size
 of the change. Small focused PRs will generally be much more faster to review. PRs that include both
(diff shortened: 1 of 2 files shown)


EdJoPaTo:
I think this could be closed for the reasons stated in #1018? There should be focus on improving the design, therefore #174.

DeflateAwning:
Hmm it's been a very long time since I've thought of this library, and I ended up putting my project that was going to use it on hold until it was a bit more feature-full. 

I'm happy if you think the best option is to close this, with a better alternative available (or soon-to-be-available) in the future.

EdJoPaTo:
> with a better alternative available (or soon-to-be-available)

Currently, there isn't.

joshka:
Marked as a draft for now. We can come back and investigate this at some later point if still needed.

PR #1026, opened 2024-04-11
chore(README): badges now are in the terminal theme

TadoTheMiner:
The badges in the readme were all the default theme. Giving them prettier colors that match the terminal gif is better. I've used the colors from the VHS repo.
```json
{
    "name": "Aardvark Blue",
    "black": "#191919",
    "red": "#aa342e",
    "green": "#4b8c0f",
    "yellow": "#dbba00",
    "blue": "#1370d3",
    "magenta": "#c43ac3",
    "cyan": "#008eb0",
    "white": "#bebebe",
    "brightBlack": "#454545",
    "brightRed": "#f05b50",
    "brightGreen": "#95dc55",
    "brightYellow": "#ffe763",
    "brightBlue": "#60a4ec",
    "brightMagenta": "#e26be2",
    "brightCyan": "#60b6cb",
    "brightWhite": "#f7f7f7",
    "background": "#102040",
    "foreground": "#dddddd",
    "cursor": "#007acc",
    "selection": "#bfdbfe",
    "meta": {
      "isDark": true,
      "credits": null
    }
  },
``` 
grafik
The dependencies badge couldn't unfortunately be themed

### DIFF
--- a/README.md
+++ b/README.md
@@ -329,7 +329,7 @@ Running this example produces the following output:
 [GitHub Sponsors]: https://github.com/sponsors/ratatui-org
 [Crate Badge]: https://img.shields.io/crates/v/ratatui?logo=rust&style=flat-square&color=F05B50&logoColor=F05B50&labelColor=102040
 [License Badge]: https://img.shields.io/crates/l/ratatui?style=flat-square&color=1370D3&labelColor=102040
-[CI Badge]: https://img.shields.io/github/actions/workflow/status/ratatui-org/ratatui/ci.yml?style=flat-square&logo=github&color=4B8C0F&labelColor=102040
+[CI Badge]: https://img.shields.io/github/actions/workflow/status/ratatui-org/ratatui/ci.yml?style=flat-square&logo=github&labelColor=102040
 [CI Workflow]: https://github.com/ratatui-org/ratatui/actions/workflows/ci.yml
 [Codecov Badge]:
     https://img.shields.io/codecov/c/github/ratatui-org/ratatui?logo=codecov&style=flat-square&token=BAQ8SOKEST&color=C43AC3&labelColor=102040&logoColor=C43AC3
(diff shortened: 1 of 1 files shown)


EdJoPaTo:
I think passing / failing should still be green / red and not some custom Color. When the successful Color is closer to red it’s harder to see at a glance.

TadoTheMiner:
Now the ci and docs have the default color
grafik

joshka:
The left side should probably be the dark grey color (light black), so the badges stand out from the background.

If there's several different shades of the same color (like with the greens / reds), that seems like it would look a bit worse compared to the existing situation where the colors are consistent but not the same as the terminal.

This is a good idea, but I think it might be blocked by constraints.

orhun:
I like this idea, and I follow it for my projects as well. Color coordinating stuff is nice.

But as other maintainers said, not sure if we can make much customizations due to the fact that some of the colors are the indicator of the status.

I still like the purple though :)

orhun (approved):

TadoTheMiner:
You're right. The grey looks much like the default one, so we rather keep the background.  We can atleast keep the blue and magenta. I think the rust logos should have the red as the failing red for consistency. Also I think it would be better if we had for the badge style because on larger fonts, the dependencies is misrendered.

TadoTheMiner:
Btw is there a reason we don't use  &link in the badges?

joshka:
> You're right. The grey looks much like the default one, so we rather keep the background. We can atleast keep the blue and magenta. I think the rust logos should have the red as the failing red for consistency.

Yeah - the blue and magenta change seems like it would be nice

> Also I think it would be better if we had for the badge style because on larger fonts, the dependencies is misrendered.

The badges get rendered in a bit of a constrained area in https://crates.io/crates/ratatui so the current size works better there. I think it might be difficult to find something that works well at all font sizes unfortunately, though perhaps changes "dependencies" to "deps.rs" would work.

> Btw is there a reason we don't use &link in the badges?

It's not something I've seen before and had to look it up (https://shields.io/badges):

> Specify what clicking on the left/right of a badge should do. Note that this only works when integrating your badge in an `<object>` HTML tag, but not an `` tag or a markup language.

These get rendered as img tags, so the &link approach won't work.

TadoTheMiner:
Do you know what red the failing build produces?

joshka:
> Do you know what red the failing build produces?

No, my code is always perfect and never breaks the build ;P

I jest, but I don't know off the top of my head what the exact color is ;)

joshka (approved):
LGTM

PR #1040, opened 2024-04-16
perf(table): const TableState::new

EdJoPaTo:
Personally I think there should be either `TableState::default()` or `TableState::new()` but not both (#978). Here we have both, so at least allow for `const fn new`.

### DIFF
--- a/src/widgets/table/table_state.rs
+++ b/src/widgets/table/table_state.rs
@@ -60,8 +60,11 @@ impl TableState {
     /// # use ratatui::{prelude::*, widgets::*};
     /// let state = TableState::new();
     /// ```
-    pub fn new() -> Self {
-        Self::default()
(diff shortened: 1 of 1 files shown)


orhun (approved):

joshka:
> Personally I think there should be either `TableState::default()` or `TableState::new()` but not both (#978). Here we have both, so at least allow for `const fn new`.

The rationale behind having both new and default is from https://rust-lang.github.io/api-guidelines/interoperability.html#c-common-traits

> Note that it is common and expected for types to implement both Default and an empty new constructor. new is the constructor convention in Rust, and users expect it to exist, so if it is reasonable for the basic constructor to take no arguments, then it should, even if it is functionally identical to default.

joshka (approved):

PR #1041, opened 2024-04-16
perf(block): use Block::bordered

EdJoPaTo:
`Block::bordered()` is shorter than `Block::new().borders(Borders::ALL)`, requires one less import (`Borders`) and in case `Block::default()` was used before can even be `const`.

This mostly changes examples, documentation, and tests.
As I used a lot of regular expressions I could have overlooked something when checking all the changes myself. Another set of eyes might find something I missed.

### DIFF
--- a/src/widgets/block.rs
+++ b/src/widgets/block.rs
@@ -41,12 +41,12 @@ pub use title::{Position, Title};
 /// ```
 /// use ratatui::{prelude::*, widgets::*};
 ///
-/// Block::default()
-///     .title("Block")
+/// Block::new()
+///     .border_type(BorderType::Rounded)
 ///     .borders(Borders::LEFT | Borders::RIGHT)
 ///     .border_style(Style::default().fg(Color::White))
-///     .border_type(BorderType::Rounded)
-///     .style(Style::default().bg(Color::Black));
+///     .style(Style::default().bg(Color::Black))
+///     .title("Block");
 /// ```
 ///
 /// You may also use multiple titles like in the following:
@@ -56,7 +56,7 @@ pub use title::{Position, Title};
 ///     widgets::{block::*, *},
 /// };
 ///
-/// Block::default()
+/// Block::new()
 ///     .title("Title 1")
 ///     .title(Title::from("Title 2").position(Position::Bottom));
 /// ```
@@ -224,7 +224,7 @@ impl<'a> Block<'a> {
     ///     widgets::{block::*, *},
     /// };
     ///
-    /// Block::default()
+    /// Block::new()
     ///     .title("Title") // By default in the top left corner
     ///     .title(Title::from("Left").alignment(Alignment::Left)) // also on the left
     ///     .title(Title::from("Right").alignment(Alignment::Right))
@@ -330,12 +330,12 @@ impl<'a> Block<'a> {
     ///     widgets::{block::*, *},
     /// };
     ///
-    /// Block::default()
+    /// Block::new()
+    ///     .title_alignment(Alignment::Center)
     ///     // This title won't be aligned in the center
     ///     .title(Title::from("right").alignment(Alignment::Right))
     ///     .title("foo")
-    ///     .title("bar")
-    ///     .title_alignment(Alignment::Center);
+    ///     .title("bar");
     /// ```
     #[must_use = "method moves the value of self and returns the modified value"]
     pub const fn title_alignment(mut self, alignment: Alignment) -> Self {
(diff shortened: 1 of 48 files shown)


orhun:
There seems to be some usages left in `README.md`, chart example and other places. I can fix it if you want 🐻

EdJoPaTo:
README, good call! I only checked *.rs files to prevent CHANGELOG.md mishaps :innocent: 

And… `Borders::all()`… havnt thought of that while searching the repo.

I think I got them all now?

orhun (approved):
LGTM

src/widgets/block.rs
```diff
@@ -1134,64 +1120,19 @@ mod tests {
...
             // .title_style(_DEFAULT_STYLE)     // no longer const
             .title_alignment(Alignment::Left)
             .title_position(Position::Top)
-            .borders(Borders::ALL)
             .padding(_DEFAULT_PADDING);
     }
 
-    /// This test ensures that we have some coverage on the [`Style::from()`] implementations
-    #[test]
-    fn block_style() {
```
joshka, line 1150: This test is an intentional duplication of the tests in style, acting as capturing the intent of the style implementation. It exists because it's rare we'd call to `Style::from()` manually, and much more common to see the `.into()` call that this exercises.

While it's true that in effect this tests that the blanket implementation of `Into` is implemented on all the items that have a `From`, that's the effect that we want.

Put more simply, we want to ensure that a user can configure various widgets using styles that are automatically converted like this. That, not the `Style::from()` call, is the important thing to test.

This could probably do with a bit more of an expanded doc comment that helps show the rationale for this. Perhaps:

/// This test intentionally duplicates the tests of `Style::from()` implementations. It does so to test from the perspective of the common use case (passing various values to a method that accepts `Into<Style>`. It captures the intent and documents the valid values better than the corresponding Style unit tests.

tests/widgets_block.rs
```diff
@@ -15,9 +15,8 @@ use ratatui::{
 fn widgets_block_renders() {
     let backend = TestBackend::new(10, 10);
     let mut terminal = Terminal::new(backend).unwrap();
-    let block = Block::default()
-        .title(Span::styled("Title", Style::default().fg(Color::LightBlue)))
-        .borders(Borders::ALL);
+    let block: Block<'_> =
```
joshka, line 18: Why is the type required here?

joshka (changes requested):
LGTM other than my comments.

src/widgets/block.rs
EdJoPaTo: Then its still a test for Style and should be with the style implementation and style tests?

tests/widgets_block.rs
EdJoPaTo: good catch!

src/widgets/block.rs
EdJoPaTo: When someone needs to read tests to understand / find / use this, then its probably bad documentation or bad design.

`Into` creates hard to understand code. What exactly is happening? (To me personally it also created performance issues in the past without being obvious as the actual implementation is hidden a few steps away and did things I wasn't aware of.)

I can assume that your example is foreground, background, modifier, but it's not clearly stated resulting in guessing when reading source code. Yes it might be slightly shorter, but we have modern editors with autocompletion. `Style::new().fg().bg().modifier()` or `Block::new().fg().bg().modifier()` isn't that much longer but so much easier to read. So personally I try to avoid Into in the first place. (Into is one of the biggest things I don't like about the ratatui code design. But as long as I don't need to use it extensively…)

Tests are there to make sure it works. Examples are there to display how something could be done. Docs to highlight specific usage of some type. Maybe move this to the examples or docs then?

joshka: I'd really like to keep these tests, but I don't mind moving them into Style instead of block or into examples / docs - it's hard to know what the right place for this is (hence defaulting to just putting it in the block tests).

The biggest reason behind keeping this is that it tests the usage of the shortcuts from the perspective of a user - i.e. what the applications that people write will see. If these tests break, then we know that there are customer apps that will be broken. Testing From says the same thing, but when those break we know that customer code that is different from the tested code breaks. I prefer a direct first order effect over a second order one when possible.

Regarding the use of `Into`, `some_value<T: Into<SomeValue>>(some_value: T)` approach is something that a bunch of the derive builder crates tend to do, and this is just a manual version of that.
- https://docs.rs/derive_builder/latest/derive_builder/#generic-setters
- https://idanarye.github.io/rust-typed-builder/typed_builder/derive.TypedBuilder.html#customization-with-attributes
- https://crates.io/crates/derive_setters
- etc.

I think there's probably a few places where we probably should have used conversion implementations rather than Into in a builder parameter (specifically in constructors). There's also some places where we should fix up the From Vec stuff to make this supper FromIter instead (in the Text related types), but that's a breaking change I've been avoiding.

joshka (approved):

joshka:
Thanks for keeping the block tests in. LGTM

Issue #1011, opened 2024-04-01
Add methods that return results / options for fallible operations


joshka:
## Problem
There are a variety of places, usually in calls to Buffer, where things can fail.
These methods should generally have alternatives that fail with an Err result, or return None instead of panicking.

## Solution
1. Exhaustively determine all the places where we can fail. Pay particular attention to the Buffer indexing methods
2. Determine for each whether it's reasonable to change the existing method by breaking compatibility or whether it's better to do so in an additive way (e.g. adding `try_xxx() -> Result`
3. Determine an Error type that we use
4. Implement the changes
5. Transition the example code and docs to use the new code where possible
6. Where things can safely ignore the failures, do so (e.g. rendering a widget outside of the buffer should generally just be ignored, and perhaps logged, instead of causing a panic)

## Additional context
Specifically Buffer::get() is a place that often causes panics.

EdJoPaTo:
The currently allowed lint clippy::missing_panics_doc catches a bunch of these cases and requires to write a panic documentation. Then we just need to seek documentation including this panic section.

Also, other lints detect arithmetic or array access out of bounds that can panic. A lot of this should be solvable by enforcing more lints.

joshka:
Good idea - added to the solution.

EdJoPaTo:
`Buffer::index_of` could get an alternative `Buffer::index_of_opt` returning `Option<usize>` instead. Same with `Buffer::get(_mut)` → `Buffer::get(_mut)_opt` returning `Option<&Cell>`.
The methods potentially panicing should be deprecated in favor of the `_opt` methods.

See #1047 which does something similar despite `Rect::intersect` not panicing.

joshka:
I was thinking about how to best migrate some of the methods that have poor adherence to the general rust API standards like this. An alternate approach is to fix the methods properly (i.e. the `get()` method returns `Option<&Cell>`), but put the old behavior behind a feature flag to make it easy for anyone to upgrade. What about something like:
```rust
#[cfg(not(feature = "legacy-buffer-methods"))]
fn get<T: Into<Position>>(position: T) -> Option<&Cell> {
   ...
}

#[cfg(feature = "legacy-buffer-methods")]
fn get(x: u16, y: u16) -> &Cell {
  ...
}
```

EdJoPaTo:
[...]

PR #1046, opened 2024-04-17
feat(buffer): track_caller for index_of

EdJoPaTo:
The caller put in the wrong x/y -> the caller is the cause.

In the future get and get_mut should return Option but thats a bigger change.

### DIFF
--- a/src/buffer/buffer.rs
+++ b/src/buffer/buffer.rs
@@ -97,12 +97,14 @@ impl Buffer {
     }
 
     /// Returns a reference to Cell at the given coordinates
+    #[track_caller]
     pub fn get(&self, x: u16, y: u16) -> &Cell {
         let i = self.index_of(x, y);
         &self.content[i]
     }
 
     /// Returns a mutable reference to Cell at the given coordinates
+    #[track_caller]
     pub fn get_mut(&mut self, x: u16, y: u16) -> &mut Cell {
         let i = self.index_of(x, y);
         &mut self.content[i]
@@ -134,6 +136,7 @@ impl Buffer {
     /// // starts at (200, 100).
     /// buffer.index_of(0, 0); // Panics
     /// ```
+    #[track_caller]
     pub fn index_of(&self, x: u16, y: u16) -> usize {
         debug_assert!(
             x >= self.area.left()


kdheepak:
> In the future get and get_mut should return Option but thats a bigger change.

I agree with this change. Can we open an issue for this to track it?

EdJoPaTo:
> > In the future get and get_mut should return Option but thats a bigger change.
> 
> I agree with this change. Can we open an issue for this to track it?

#1011 is less specific to Buffer but includes this.

joshka (changes requested):
I think these might give some messages that might not make contextual sense, but I'm not sure. Perhaps a few unit tests for this would help clarify. I'd want to see them from the perspective of a direct call (i.e. what panic message will a widget see calling `get(x, y).symbol("x")`), as well as a call to methods in the buffer that call these  methods e.g. `buf.set_style(area, Color::Blue)`.

EdJoPaTo:
`get` looks like this then:

```diff
--- a/tests/terminal.rs
+++ b/tests/terminal.rs
@@ -38,7 +38,7 @@ fn terminal_draw_returns_the_completed_frame() -> Result<(), Box<dyn Error>> {
         let paragraph = Paragraph::new("Test");
         f.render_widget(paragraph, f.size());
     })?;
-    assert_eq!(frame.buffer.get(0, 0).symbol(), "T");
+    assert_eq!(frame.buffer.get(30_000, 0).symbol(), "T");
     assert_eq!(frame.area, Rect::new(0, 0, 10, 10));
```

```plaintext
---- terminal_draw_returns_the_completed_frame stdout ----
thread 'terminal_draw_returns_the_completed_frame' panicked at tests/terminal.rs:41:29:
Trying to access position outside the buffer: x=30000, y=0, area=Rect { x: 0, y: 0, width: 10, height: 10 }
```

Note that the test output panicked at the diff line and not at the assertion inside the `index_of` method (which is used by `get`).

`buf.set_style` is not able to create this kind of error as it uses intersect to ensure the input is in range. But `set_stringn` uses `index_of` which can fail here:

```diff
--- a/src/widgets/list.rs
+++ b/src/widgets/list.rs
@@ -992,7 +992,7 @@ impl StatefulWidgetRef for List<'_> {
                 };
                 if selection_spacing {
                     buf.set_stringn(
-                        x,
+                        x + 30_000,
                         y + j as u16,
                         symbol,
                         list_area.width as usize,
```

```plaintext
---- widgets::list::tests::test_padding_offset_pushback_break stdout ----
thread 'widgets::list::tests::test_padding_offset_pushback_break' panicked at src/buffer/buffer.rs:219:30:
Trying to access position outside the buffer: x=30000, y=0, area=Rect { x: 0, y: 0, width: 10, height: 4 }
```

The panic happened inside the `set_stringn` method at the call of `index_of`.

joshka:
Added / adjusted tests.

joshka (approved):

joshka:
> (Also probably won’t work on `cargo test --release`)

Good catch - I didn't think of that.

I'll remove and merge this without the added tests.

PR #1047, opened 2024-04-17
feat(rect): intersect_opt

EdJoPaTo:
While trying to figure out #1001 I tried to understand why `max_x` is calculated so overcomplicated.

https://github.com/ratatui-org/ratatui/blob/bef5bcf750375a78b11ae06f217091b2463e842f/src/text/span.rs#L366-L373

`max_x` should be `intersected.right()`. No other logic required. `assert_eq!()` noticed differences which surprised me. Then I found out: intersect returns an empty area but the offset stays. This means in these cases `right()` is the offset at some wonky position without an area.
The simple solution to this is to not render anything when `intersected.is_empty()`. In some places of the `intersect()` usage this was done, in some it was not. This means it's a bug introduced by mistake (#997 in this case). Just fixing this would introduce this bug again at some point in the future.

The better way is to return whether there is an actual intersection. When there is no intersection its unlikely someone wants to use this `Rect`. So return `None` instead.
The user of `intersect()` then has to actively think about what to do with this case. In case of rendering to an area: abort the render early. No space to render into anyway. So this might also be a performance gain.

While having this implemented wrong I had a bunch of failing tests which sparked #1046 and the `saturating_sub` in `Block`.

The name of the new method is inspired by `chrono` which moves to `_opt` methods as the existing ones can panic. (Maybe `Buffer::index_of` should also introduce `Buffer::index_of_opt`, same for `Buffer::get(_mut)`)

### DIFF
--- a/examples/flex.rs
+++ b/examples/flex.rs
@@ -338,10 +338,11 @@ impl App {
         }
 
         if scrollbar_needed {
-            let area = area.intersection(buf.area);
-            let mut state = ScrollbarState::new(max_scroll_offset() as usize)
-                .position(self.scroll_offset as usize);
-            Scrollbar::new(ScrollbarOrientation::VerticalRight).render(area, buf, &mut state);
+            if let Some(area) = area.intersection_opt(buf.area) {
+                let mut state = ScrollbarState::new(max_scroll_offset() as usize)
+                    .position(self.scroll_offset as usize);
+                Scrollbar::new(ScrollbarOrientation::VerticalRight).render(area, buf, &mut state);
+            }
         }
         scrollbar_needed
     }
--- a/src/buffer/buffer.rs
+++ b/src/buffer/buffer.rs
@@ -273,7 +273,9 @@ impl Buffer {
     /// your own type that implements [`Into<Style>`]).
     pub fn set_style<S: Into<Style>>(&mut self, area: Rect, style: S) {
         let style = style.into();
-        let area = self.area.intersection(area);
+        let Some(area) = self.area.intersection_opt(area) else {
+            return;
+        };
         for y in area.top()..area.bottom() {
             for x in area.left()..area.right() {
                 self.get_mut(x, y).set_style(style);
--- a/src/layout/rect.rs
+++ b/src/layout/rect.rs
@@ -152,15 +152,15 @@ impl Rect {
     /// Returns a new `Rect` that contains both the current one and the given one.
     #[must_use = "method returns the modified value"]
     pub fn union(self, other: Self) -> Self {
-        let x1 = min(self.x, other.x);
-        let y1 = min(self.y, other.y);
-        let x2 = max(self.right(), other.right());
-        let y2 = max(self.bottom(), other.bottom());
+        let left = min(self.x, other.x);
+        let top = min(self.y, other.y);
+        let right = max(self.right(), other.right());
+        let bottom = max(self.bottom(), other.bottom());
         Self {
-            x: x1,
-            y: y1,
-            width: x2.saturating_sub(x1),
-            height: y2.saturating_sub(y1),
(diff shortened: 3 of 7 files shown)


joshka:
> max_x should be intersected.right(). No other logic required

yep. The implementation here was mostly copied from `Buffer::set_stringn` with modifications. This should have been simplified.

Taking a quick look at the current usage of intersection in the wild, I think we probably shouldn't deprecate this, and perhaps shouldn't have the _opt version at all as a .is_empty() check seems like it's good enough. Making users have to specifically handle the empty case differently than just treating it the same as if they'd been passed an empty area is annoying and is inconsistent with other places that return empty rects (e.g. layout::split()) where calling `is_empty()` would be idiomatic.

e.g. 
```rust
  // area can be empty already
  
```

EdJoPaTo:
> as a .is_empty() check seems like it's good enough. Making users have to specifically handle the empty case differently than just treating it the same as if they'd been passed an empty area is annoying and is inconsistent with other places that return empty rects (e.g. layout::split()) where calling `is_empty()` would be idiomatic.

Rust has the concept of always valid states and to think about the different cases. When something exists its valid. This is especially noticeable on the Result enum as a return type. You have to care about the differences in the return type. What is what you want to have? Handle it? Ignore it? The language helps you to think about these cases and don't forget about it.

In this case it was forgotten that `intersection` has these differences. If you don't care, its one `unwrap_or_default()` away. But when you care the compiler and language will help you in thinking about all the cases correctly.

Not fixing this would result in another mistake like this. Maybe in this repo, maybe in another one.

joshka:
An empty Rect is a valid state. Just because `Option` exists doesn't necessarily mean that it's the right tool for the job in all cases where there's a reasonable way of representing zero in the type. e.g. `1-1` is not `None`, `vec[]` is not `None` either.

> Not fixing this would result in another mistake like this. Maybe in this repo, maybe in another one.

Can you clarify what mistake you mean here? I'm probably misunderstanding something. #997 was a change that added an intersection call. All the code following the call just becomes noops if the area is empty, and doesn't need any checks to see that. It could check if there was any perf impact, but it's unlikely to be more than a few nanosecs and not worth bothering with.

EdJoPaTo:
> Can you clarify what mistake you mean here?

The following code interacted on a Rect with an empty area but with an unexpected offset. The offset requires additional complexity (in this case `min`, which was already there before the fix of #997).

joshka:
I might be still misunderstanding the need for this a bit.

> The following code

Which specific code? There's no link aside from the link to #997.

#997 uses `intersection` as a means to prevent out of bounds writes. The bug is that it wasn't called, not that there was any problem with calling `intersection` and returning an empty area. That said, the Span::render_ref method could probably be simplified to use two iterators working together, one on the area, and one on the graphemes instead of using math.

I (generally) would never do any of the following:
- check for empty `Vec`s when iterating, because I trust that iterating over an empty `Vec` is a noop. For working with `Rect`s we can iterate correctly with a noop using `rows()` and `cols()`.
- return `Option<Vec>` from any code, because an empty `Vec` is idiomatic. An empty `Rect` is idiomatically the same as an empty Vec.
- expect that a call to find the intersection of two values would return `None` instead of an empty value. Example: https://doc.rust-lang.org/std/collections/hash_set/struct.Intersection.html returns an iterator over the values. It does not return an optional iterator that returns None if there is no intersection of values.

For the most part, the changes to where area.intersection is called to use `intersection_opt()` look better as a guard clause which checks `Rect::is_empty()`. We usually have omitted this as most code will effectively noop on empty areas. We could probably add these guards if benchmarks show these are on the hot path, but I don't think most would be.

TL;DR, I don't think this specific change improves the code, as it provides a something that is already working fine, in a non-idiomatic way.

PR #1049, opened 2024-04-18
fix: integrate render Line/Span into Buffer methods

EdJoPaTo:
Results in same behavior of rendering for Span/Line.render vs Buffer.set_span/line.

Also adds Buffer::_opt methods without enforcing them yet. (Creates a lot of churn)

Related: #1046 #1011 #1007 #1001 (especially https://github.com/ratatui-org/ratatui/pull/1001#issuecomment-2062581438)

While this touches a lot of core logic I am confident that this does not change the intended behavior. The tests are still green, and only one test was adapted to merge with `Line::render_ref` behavior.

Thought about splitting this up into multiple PR but the new `Buffer::set_stringn` implementation uses and requires `index_of_opt` in order to not be a breaking change. Could probably work around that in some way when of interest.

### DIFF
--- a/src/buffer/buffer.rs
+++ b/src/buffer/buffer.rs
@@ -1,8 +1,3 @@
-use std::{
-    cmp::min,
-    fmt::{Debug, Formatter, Result},
-};
-
 use unicode_segmentation::UnicodeSegmentation;
 use unicode_width::UnicodeWidthStr;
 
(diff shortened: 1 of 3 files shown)


joshka (commented):
My intent by copying the implementation details into the widgets instead of the buffer was to eventually deprecate the buffer methods. The idea behind that was to push the idea of composing widgets from smaller parts.

https://github.com/ratatui-org/ratatui/pull/709#issuecomment-1869494715 was the original discussion (not much more than what I wrote above though).

It's also worth checking if there is any untested behavior in the existing implementations of the buffer methods - I vaguely recall changing the way the background was handled for the line impl (by making it cover the entire line rather than just the spans).

I think the _opt changes are good, but there's a note about a possible alternative approach in the issue that talks more about this that's worth considering. It might be worth splitting that part out.

TL;DR: I think hold off on this PR for now unless there's a strong reason to go ahead.

EdJoPaTo:
> My intent by copying the implementation details into the widgets instead of the buffer was to eventually deprecate the buffer methods. The idea behind that was to push the idea of composing widgets from smaller parts.
> 
> [#709 (comment)](https://github.com/ratatui-org/ratatui/pull/709#issuecomment-1869494715) was the original discussion (not much more than what I wrote above though).
> 

Currently both are diverged in features and functionality which is… not helpful.

The Buffer.set_line/span methods have return values which are used. At first I removed them from the Buffer but as I noticed them being used I moved the logic to the Buffer.
Whatever we do, we need to prevent multiple places of usage.

I can split this up and include the set_string method which improves over the current one.

joshka:
> Currently both are diverged in features and functionality which is… not helpful.

You're right it's not helpful in the short term, but the implementations in Line and Span are more correct for longer term and I think that's where we should focus the implementation (replacement rather than trying to make these have the same behavior). We can't easily change behavior of the functions in Buffer without breaking the world, but we can introduce new, better, divergent functionality in the Line and Span Widget implementations.

EdJoPaTo:
Superseded by #1082 and #1083.

PR #1053, opened 2024-04-19
build: add clippy::cargo lint

joshka:
Followup to https://github.com/ratatui-org/ratatui/pull/1035 and
https://github.com/ratatui-org/ratatui/discussions/1034

It's reasonable to enable this and deal with breakage by fixing any
specific issues that arise.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,7 +64,9 @@ serde_json = "1.0.109"
 
 [lints.rust]
 unsafe_code = "forbid"
+
 [lints.clippy]
+cargo = { level = "warn", priority = -1 }
 pedantic = { level = "warn", priority = -1 }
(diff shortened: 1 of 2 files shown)


EdJoPaTo:
Without a `Cargo.lock` there will come up breaking CI at random. I am not sure how well this will behave for people providing PRs. I use `git push --no-verify` already because it's annoying, but then we might have a randomly occurring thing preventing someone from pushing without the `--no-verify`?
That will likely be annoying to contributors. When this lint is wanted it should be done in some other way. Maybe as its own CI job independent of the rest.

joshka:
I'm fine with breaking CI. We fix things like this pretty quickly.

orhun (approved):

Issue #1028, opened 2024-04-12
Support styling as a hyperlink


mo8it:
See https://github.com/crossterm-rs/crossterm/issues/602 and https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda

joshka:
I want this too. I suspect OSC 8 will just work if https://github.com/ratatui-org/ratatui/issues/902 is fixed.
A super high level on that issue is that the calculation of how many cells to skip needs to take into account that ANSI escape codes might mean that we write many symbols to the same cell.

A full solution to this would be to add a hyperlink widget that has properties for the text and hyperlink so app code can just do:
```rust
Hyperlink::new("Google", "https://google.com).render(area, buf);
```

Alternatively, I think as phrased you might be looking to add the hyperlink as another part of the `Style` struct, which would allow it to be added to any span/line/text without having to break the flow. Thoughts?

mo8it:
Thanks for your quick reply 🤗

A widget would be too limiting. I think that this should be part of a `Paragraph` since such a hyperlink can be part of a longer text and might be broken/wrapped into multiple lines (so not only in one `Line`).

joshka:
To be clear, a Hyperlink Widget as proposed above would accept anything that supports `Into<Text>`. I'd like eventually to move the wrapping into `Text` / `Line` / `Span` instead of wrapping being implemented in `Paragraph` only.

joshka:
![Demo](https://vhs.charm.sh/vhs-2c5cvJzN7KTKR7U3Jbk71J.gif)

```rust
/// A hyperlink widget that renders a hyperlink in the terminal using [OSC 8].
///
/// [OSC 8]: https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda
struct Hyperlink<'content> {
    text: Text<'content>,
    url: String,
}

impl<'content> Hyperlink<'content> {
    fn new(text: impl Into<Text<'content>>, url: impl Into<String>) -> Self {
        Self {
            text: text.into(),
            url: url.into(),
        }
    }
}

impl WidgetRef for Hyperlink<'_> {
    fn render_ref(&self, area: Rect, buffer: &mut Buffer) {
        self.text.render_ref(area, buffer);
        let width = (area.width as usize).min(self.text.width());

        // this is a hacky workaround for https://github.com/ratatui-org/ratatui/issues/902, a bug
[...]

PR #1063, opened 2024-04-25
docs(examples): add hyperlink example

joshka:
Demo

Proof of concept for: https://github.com/ratatui-org/ratatui/issues/1028


joshka:
ping @ratatui-org/maintainers

orhun:
Tested and works:

rec_20240526T205504

The 2-char hack felt a bit weird at first but I can't really think of something better that we can do here.

joshka:
> The 2-char hack felt a bit weird at first but I can't really think of something better that we can do here.

Fixing the parsing of ANSI inside the terminal / backend is the right fix. When that happens, this becomes a simpler example, but for now it documents the workaround.

orhun (approved):
This is good to have with its current state I think.

Issue #1056, opened 2024-04-22
Enable lint `clippy::string_slice`


EdJoPaTo:
## Problem

Slicing of strings can panic on multi width characters. See for example #1032

## Solution

Enable the following lint to find all cases where this might happen (now and in the future).

https://rust-lang.github.io/rust-clippy/master/index.html#string_slice

This lint does not catch #1032 for example. But it catches other places. It is probably a good idea to suggest false negatives like #1032 to clippy to be integrated in this lint.

```bash
$ cargo clippy -- -W clippy::string_slice
warning: indexing into a string may panic if the index is within a UTF-8 character
   --> src/style/color.rs:237:49
    |
237 | ...                   u8::from_str_radix(&s[1..3], 16),
    |                                           ^^^^^^^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#string_slice
    = note: requested on the command line with `-W clippy::string-slice`

warning: indexing into a string may panic if the index is within a UTF-8 character
   --> src/style/color.rs:238:49
    |
238 | ...                   u8::from_str_radix(&s[3..5], 16),
    |                                           ^^^^^^^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#string_slice

warning: indexing into a string may panic if the index is within a UTF-8 character
   --> src/style/color.rs:239:49
    |
239 | ...                   u8::from_str_radix(&s[5..7], 16),
    |                                           ^^^^^^^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#string_slice

warning: indexing into a string may panic if the index is within a UTF-8 character
   --> src/widgets/reflow.rs:338:6
    |
338 |     &src[start..]
    |      ^^^^^^^^^^^^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#string_slice

warning: `ratatui` (lib) generated 4 warnings
    Finished dev [unoptimized + debuginfo] target(s) in 6.70s
```

joshka:
Sounds like a good plan. Just to clarify, this entails actually fixing those places identified by the lint, right?

EdJoPaTo:
Enabling the lint and fixing the places yeah. But also finding cases where this lint is not noticing this bad behavior to report this and improve the lint (for us and everyone else).

[...]

PR #1064, opened 2024-04-25
chore: add clippy::string_slice

joshka:
Fixes: https://github.com/ratatui-org/ratatui/issues/1056

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -99,6 +99,7 @@ string_slice = "warn"
 string_to_string = "warn"
 unnecessary_self_imports = "warn"
 use_self = "warn"
+string_slice = "warn"
 
 [features]
 #! The crate provides a set of optional features that can be enabled in your `cargo.toml` file.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -99,7 +99,6 @@ string_slice = "warn"
 string_to_string = "warn"
 unnecessary_self_imports = "warn"
 use_self = "warn"
-string_slice = "warn"
 
 [features]
 #! The crate provides a set of optional features that can be enabled in your `cargo.toml` file.
--- a/src/style/color.rs
+++ b/src/style/color.rs
@@ -313,17 +313,20 @@ impl FromStr for Color {
                 _ => {
                     if let Ok(index) = s.parse::<u8>() {
                         Self::Indexed(index)
-                    } else if let (Ok(r), Ok(g), Ok(b)) = {
-                        if !s.starts_with('#') || s.len() != 7 {
-                            return Err(ParseColorError);
-                        }
-                        (
-                            u8::from_str_radix(&s.chars().skip(1).take(2).collect::<String>(), 16),
-                            u8::from_str_radix(&s.chars().skip(3).take(2).collect::<String>(), 16),
-                            u8::from_str_radix(&s.chars().skip(5).take(2).collect::<String>(), 16),
-                        )
-                    } {
-                        Self::Rgb(r, g, b)
+                    } else if s.starts_with('#') && s.len() == 7 {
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -96,6 +96,7 @@ rest_pat_in_fully_bound_structs = "warn"
 string_lit_chars_any = "warn"
 string_to_string = "warn"
 use_self = "warn"
+string_slice = "warn"
```
EdJoPaTo, line 99: Sort alphabetically

src/style/color.rs
```diff
@@ -321,9 +321,9 @@ impl FromStr for Color {
                             return Err(ParseColorError);
                         }
                         (
-                            u8::from_str_radix(&s[1..3], 16),
-                            u8::from_str_radix(&s[3..5], 16),
-                            u8::from_str_radix(&s[5..7], 16),
+                            u8::from_str_radix(&s.chars().skip(1).take(2).collect::<String>(), 16),
+                            u8::from_str_radix(&s.chars().skip(3).take(2).collect::<String>(), 16),
+                            u8::from_str_radix(&s.chars().skip(5).take(2).collect::<String>(), 16),
```
EdJoPaTo, line 326: This looks like a performance nightmare. It should use the same iterator. And collect String does heap allocations which could be prevented by only using references. 

Maybe something like `s.get(range).ok()?`? When it doesn’t work it’s not a hex Color.

EdJoPaTo (changes requested):

src/style/color.rs
joshka: Makes sense - will fix.

src/style/color.rs
```diff
@@ -316,17 +316,20 @@ impl FromStr for Color {
...
+                            .get(1..3)
+                            .and_then(|v| u8::from_str_radix(v, 16).ok())
+                            .ok_or(ParseColorError)?;
+                        let green = s
+                            .get(3..5)
+                            .and_then(|v| u8::from_str_radix(v, 16).ok())
+                            .ok_or(ParseColorError)?;
+                        let blue = s
+                            .get(5..7)
+                            .and_then(|v| u8::from_str_radix(v, 16).ok())
```
EdJoPaTo, line 330: I would prefer non single letter variable names. Maybe `str`, maybe `blue`?

joshka:
btw @EdJoPaTo I think you were right about that multiple dependencies lint.

```
❯ cargo tree -i windows-targets@0.48.5 --target all
windows-targets v0.48.5
├── parking_lot_core v0.9.9
│   └── parking_lot v0.12.1
│       └── crossterm v0.27.0
│           └── ratatui v0.26.2 (/Users/joshka/local/ratatui)
└── windows-sys v0.48.0
    └── mio v0.8.10
        ├── crossterm v0.27.0 (*)
        └── signal-hook-mio v0.2.3
            └── crossterm v0.27.0 (*)

❯ cargo tree -i windows-targets@0.52.5 --target all
windows-targets v0.52.5
└── windows-sys v0.52.0
    ├── console v0.15.8
    │   └── better-panic v0.3.0
    │       [dev-dependencies]
    │       └── ratatui v0.26.2 (/Users/joshka/local/ratatui)
    ├── is-terminal v0.4.12
    │   └── criterion v0.5.1
    │       [dev-dependencies]
    │       └── ratatui v0.26.2 (/Users/joshka/local/ratatui)
    └── winapi-util v0.1.8
        ├── same-file v1.0.6
        │   └── walkdir v2.5.0
        │       └── criterion v0.5.1 (*)
        └── walkdir v2.5.0 (*)
```

joshka:
I think this was probably done well enough elsewhen.

Issue #1045, opened 2024-04-17
Easier debugging


kdheepak:
## Problem

Sometimes I want to print a widget that I'm developing to the terminal to see whether it is being displayed the way I expect it to. Currently, I have to set up a TUI with key handlers etc, run the TUI, exit the TUI.

I would like to instead just print the widget to the terminal.

`TestBackend` has a debug print which almost does what I want but it doesn't display any colors. 

## Solution

This works for me but I'm wondering if there's a better way to do this?

```rust
#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn render_default() -> color_eyre::Result<()> {
        let mut buf = Vec::new();
        let backend = CrosstermBackend::new(&mut buf);
        let mut terminal = Terminal::with_options(
            backend,
            TerminalOptions {
                viewport: Viewport::Inline(1),
            },
        )?;
        terminal
            .draw(|f| f.render_widget("hello world", f.size()))?;
        drop(terminal);
        let view = String::from_utf8(buf)?;
        println!("{view}");
        Ok(())
    }
}
```

Ideally `TestBackend` would have a way to print with styles.

kdheepak:
We should probably update this function to handle styles as well:

https://github.com/ratatui-org/ratatui/blob/326a461f9a345ba853d57afefc8d77ba0b0b5a14/src/backend/test.rs#L46-L73

And make the `TestBackend` configurable such that it either strips ansi styles or not when `Display` is invoked.

joshka:
Creating a new TestBuffer type for this would probably work ok.

Imagine:

```rust
let w = MyWidget::new(...);
let mut buf = TestBuffer::new(10, 3);
assert_eq!(
  buf.render_string(w)),
  "..."
);
```

Add props for ansi / plain.
Add a method for `Vec<String>` to make it easier to compare.

Maybe even `TestBuffer::new(w, 10, 3).to_string()` would work.

Even just a method would be fine perhaps but it would be nice to be able to configure this a little.

Internally it would allocate the right size buffer, render and pull out the string value.

imsnif:
Hey, I have a hacked together function that receives a ratatui `Buffer` and returns stringified ANSI (with styles and all). I unfortunately do not have the capacity to do a proper PR for this, but will it help if I paste it here?

kdheepak:
Yes absolutely! I'm happy to make it a PR in the future too (crediting you of course)!

imsnif:
[...]

PR #1065, opened 2024-04-26
feat: add WidgetExt trait for debugging widgets

joshka:
Importing the `WidgetExt` trait allows users to easily get a string
representation of a widget with ANSI escape sequences for the terminal.
This is useful for debugging and testing widgets.

```rust
use ratatui::{prelude::*, widgets::widget_ext::WidgetExt};

fn main() {
    let greeting = Text::from(vec![
        Line::styled("Hello", Color::Blue),
        Line::styled("World ", Color::Green),
    ]);
    println!("{}", greeting.to_ansi_string(5, 2));
}
```

Fixes: https://github.com/ratatui-org/ratatui/issues/1045


src/widgets/ansi_string_buffer.rs
```diff
@@ -0,0 +1,171 @@
...
+                    last_style = Some(style);
+                }
+                f.write_str(cell.symbol())?;
+            }
+        }
+        f.write_str("\u{1b}[0m")
+    }
+}
+
+fn write_cell_style(cell: &buffer::Cell, f: &mut Formatter<'_>) -> fmt::Result {
```
kdheepak, line 54: it _feels_ more natural to me to have the order of these arguments reversed:

```suggestion
fn write_cell_style(f: &mut Formatter<'_>, cell: &buffer::Cell) -> fmt::Result {
```

Thoughts?

src/widgets/widget_ext.rs
```diff
@@ -0,0 +1,50 @@
...
+                Line::styled("World ", Color::Green),
+            ]);
+            text.render(area, buf);
+        }
+    }
+
+    #[test]
+    fn widget_ext_to_ansi_string() {
+        let ansi_string = Greeting.to_ansi_string(5, 2);
+        println!("{ansi_string}");
```
kdheepak, line 44: Maybe remove the `println!` in the tests before merging?

src/widgets/ansi_string_buffer.rs
joshka: Agreed!. Done

src/widgets/widget_ext.rs
joshka: Yep, I put these in there before adding the example.

joshka:
Made AnsiStringBuffer less private so this works too:

```
use ratatui::{prelude::*, widgets::ansi_string_buffer::AnsiStringBuffer};

fn main() {
    let mut buf = AnsiStringBuffer::new(5, 2);
    buf.render_ref(&Line::styled("Hello", Color::Blue), Rect::new(0, 0, 5, 1));
    buf.render_ref(&Line::styled("World", Color::Green), Rect::new(0, 1, 5, 1));
    println!("{buf}");
}
```

joshka:
Things I'm not 100% sure of with this:
- Naming
  - WidgetExt / WidgetAnsiExt
  - AnsiStringBuffer / ?
  - to_ansi_string() ?
- Use of WidgetRef vs Widget
  - render_ref / render
  - I can't make a blanket impl for `Widget` if there's a blanket impl for `WidgetRef` as the implementations clash - this could suggest that there's a deeper issue about how these traits work.
- Putting this in a new type (AnsiStringBuffer) vs adding to the existing buffer
- Any obvious idiomatic approaches using conversion traits
- Whether to just use a lib for the ANSI stuff
- Underline color is missing (I skipped because it was a little more complex than I'd like)

src/widgets/widget_ext.rs
```diff
@@ -0,0 +1,50 @@
...
+
+use super::ansi_string_buffer::AnsiStringBuffer;
+use crate::prelude::*;
+
+/// An extension trait for widgets that provides methods that are useful for debugging.
+#[stability::unstable(
+    feature = "widget-ext",
+    issue = "https://github.com/ratatui-org/ratatui/issues/1045"
+)]
+pub trait WidgetExt {
```
kdheepak, line 12: I agree with your comment about the naming. In the off chance we want to use more extensions, this should be named `ToAnsiStringWidgetExt` or something to that effect.

EdJoPaTo: Either this trait should allow for more things on the same trait (in the future), then the name is fine, or it should be specific to this.
I also dislike the Ext ending here.
`AnsiStringPrintable` would be a trait name that represents what it does.

joshka: I think Ext is *probably* the right suffix for this pattern if it's not a more generic trait. https://rust-lang.github.io/rfcs/0445-extension-trait-conventions.html. It's used in tokio / futures streams (StreamExt), tracing-subscriber (a bunch of times https://docs.rs/tracing-subscriber/latest/tracing_subscriber/?search=ext) and various other crates use similar naming. I mostly wasn't sure if adding something before Ext made sense.

> The proposed convention is, first of all, to (1) prefer adding default methods to existing traits or (2) prefer generically useful traits to extension traits whenever feasible.

Adding default methods doesn't seem like the right idea as converting to ANSI doesn't seem like it's something everyone would need on a widget, and it's something I'd prefer not to confuse newer users with if possible (hence adding as an extension trait / generically useful trait seems right). Adding a default method would however allow each widget to define their own ANSI representation (though I can't think of a useful need for this).

I'm not sure that given that the method needs a certain size to render, whether this trait has a generic usefulness. I don't see that there's really any need for this trait other than printing out widgets (I could be wrong there).

EdJoPaTo: Other objects not being Widgets could implement that trait too. Buffer for example.

kdheepak: I was also thinking `Ext` was the right suffix since I had seen it in several other places, but maybe it is more commonly used when extending external traits or standard library traits? I like `AnsiStringPrintable` too. 

If it has just one method `.to_ansi_string`, why not `trait ToAnsiString {}`?

joshka: > Other objects not being Widgets could implement that trait too. Buffer for example.

The bit I struggle with is working out what the purpose of passing in the width and height would be if implemented on Buffer? Would `trait RenderToAnsiString { fn render_to_ansi_string() }` make sense?

src/widgets.rs
```diff
@@ -21,6 +21,7 @@
 //! - [`Tabs`]: displays a tab bar and allows selection.
 //!
 //! [`Canvas`]: crate::widgets::canvas::Canvas
+pub mod ansi_string_buffer;
```
joshka, line 24: I want to mark this as unstable and the widget module and re-export the trait / struct, but this needs https://github.com/sagebind/stability/pull/11 (enable unstable attribute on use statements).

jkelleyrtp:
Just FYI the logic here has issues with subsequent unicode characters. 
It seems the correct approach is to break the lines into grapheme clusters and then assemble the Line from Cells manually.

I was using the code here in the dioxus CLI, but we had bugs like https://github.com/DioxusLabs/dioxus/issues/4409.

For anyone just wanting to dump arbitrary ansi output from running commands, here's a more correct version:

```rust
/// A buffer that can be rendered to and then dumped as raw ansi codes
///
/// This is taken from a PR on the ratatui repo (<https://github.com/ratatui/ratatui/pull/1065>) and
/// modified to be more appropriate for our use case.
pub fn ansi_string_to_line(line: Line) -> String {
    let graphemes = line.styled_graphemes(Style::default());
    let mut out = String::with_capacity(line.spans.first().map_or(0, |s| s.content.len() * 2));
    let mut last_style = None;
    for grapheme in graphemes {
        let mut cell = Cell::default();
        cell.set_symbol(grapheme.symbol);
        cell.set_style(grapheme.style);

        let style = (cell.fg, cell.bg, cell.modifier);
        if last_style.is_none() || last_style != Some(style) {
            _ = write_cell_style(&mut out, cell.modifier, cell.fg, cell.bg);
            last_style = Some(style);
        }

        _ = out.write_str(cell.symbol());
    }

    _ = out.write_str("\u{1b}[0m");

    out
}

fn write_cell_style(f: &mut impl Write, modifier: Modifier, fg: Color, bg: Color) -> fmt::Result {
    f.write_str("\u{1b}[")?;

    // Write the modifier codes
    if modifier.contains(Modifier::BOLD) {
        f.write_str("1;")?;
    }
    if modifier.contains(Modifier::DIM) {
        f.write_str("2;")?;
    }
    if modifier.contains(Modifier::ITALIC) {
        f.write_str("3;")?;
    }
    if modifier.contains(Modifier::UNDERLINED) {
        f.write_str("4;")?;
    }
    if modifier.contains(Modifier::SLOW_BLINK) {
        f.write_str("5;")?;
    }
    if modifier.contains(Modifier::RAPID_BLINK) {
        f.write_str("6;")?;
    }
    if modifier.contains(Modifier::REVERSED) {
        f.write_str("7;")?;
    }
    if modifier.contains(Modifier::HIDDEN) {
        f.write_str("8;")?;
    }
    if modifier.contains(Modifier::CROSSED_OUT) {
        f.write_str("9;")?;
    }

    // Write the foreground
    f.write_str(match fg {
        Color::Reset => "39",
        Color::Black => "30",
        Color::Red => "31",
        Color::Green => "32",
        Color::Yellow => "33",
        Color::Blue => "34",
        Color::Magenta => "35",
        Color::Cyan => "36",
        Color::Gray => "37",
        Color::DarkGray => "90",
        Color::LightRed => "91",
        Color::LightGreen => "92",
        Color::LightYellow => "93",
        Color::LightBlue => "94",
        Color::LightMagenta => "95",
        Color::LightCyan => "96",
        Color::White => "97",
        _ => "",
    })?;
    if let Color::Rgb(red, green, blue) = fg {
        f.write_fmt(format_args!("38;2;{red};{green};{blue}"))?;
    }
    if let Color::Indexed(i) = fg {
        f.write_fmt(format_args!("38;5;{i}"))?;
    }
    f.write_str(";")?;

    // Write the background
    f.write_str(match bg {
        Color::Reset => "49",
        Color::Black => "40",
        Color::Red => "41",
        Color::Green => "42",
        Color::Yellow => "43",
        Color::Blue => "44",
        Color::Magenta => "45",
        Color::Cyan => "46",
        Color::Gray => "47",
        Color::DarkGray => "100",
        Color::LightRed => "101",
        Color::LightGreen => "102",
        Color::LightYellow => "103",
        Color::LightBlue => "104",
        Color::LightMagenta => "105",
        Color::LightCyan => "106",
        Color::White => "107",
        _ => "",
    })?;
    if let Color::Rgb(red, green, blue) = bg {
        f.write_fmt(format_args!("48;2;{red};{green};{blue}"))?;
    }
    if let Color::Indexed(i) = bg {
        f.write_fmt(format_args!("48;5;{i}"))?;
    }

    f.write_str("m")
}
```

If you have multi-line text, simply break the incoming String into lines and then write line by line. As an example, this code prints rendered cargo log messages with some chrono formatting:
```rust

    fn tracemsg_to_ansi_string(log: TraceMsg) -> Vec<String> {
        use ansi_to_tui::IntoText;
        use chrono::Timelike;

        let rendered = match log.content {
            TraceContent::Cargo(msg) => msg.rendered.unwrap_or_default(),
            TraceContent::Text(text) => text,
        };

        let mut lines = vec![];

        for (idx, raw_line) in rendered.lines().enumerate() {
            let line_as_text = raw_line.into_text().unwrap();
            let is_pretending_to_be_frame = !raw_line.is_empty()
                && raw_line
                    .chars()
                    .all(|c| c == '=' || c == '-' || c == ' ' || c == '─');

            for (subline_idx, mut line) in line_as_text.lines.into_iter().enumerate() {
                if idx == 0 && subline_idx == 0 {
                    let mut formatted_line = Line::default();

                    formatted_line.push_span(
                        Span::raw(format!(
                            "{:02}:{:02}:{:02} ",
                            log.timestamp.hour(),
                            log.timestamp.minute(),
                            log.timestamp.second()
                        ))
                        .dark_gray(),
                    );

                    formatted_line.push_span(
                        Span::raw(format!(
                            "[{src}] {padding}",
                            src = log.source,
                            padding =
                                " ".repeat(3usize.saturating_sub(log.source.to_string().len()))
                        ))
                        .style(match log.source {
                            TraceSrc::App(_platform) => match log.level {
                                Level::ERROR => Style::new().red(),
                                Level::WARN => Style::new().yellow(),
                                Level::INFO => Style::new().magenta(),
                                Level::DEBUG => Style::new().magenta(),
                                Level::TRACE => Style::new().magenta(),
                            },
                            TraceSrc::Dev => match log.level {
                                Level::ERROR => Style::new().red(),
                                Level::WARN => Style::new().yellow(),
                                Level::INFO => Style::new().blue(),
                                Level::DEBUG => Style::new().blue(),
                                Level::TRACE => Style::new().blue(),
                            },
                            TraceSrc::Cargo => Style::new().yellow(),
                            TraceSrc::Build => Style::new().blue(),
                            TraceSrc::Bundle => Style::new().blue(),
                            TraceSrc::Unknown => Style::new().gray(),
                        }),
                    );

                    for span in line.spans {
                        formatted_line.push_span(span);
                    }

                    line = formatted_line;
                }

                if is_pretending_to_be_frame {
                    line = line.dark_gray();
                }

                // Create the ansi -> raw string line with a width of either the viewport width or the max width
                lines.push(ansi_string_to_line(line));
            }
        }

        lines
    }
```

FWIW I'd love to see this method in ratatui itself.

joshka:
> FWIW I'd love to see this method in ratatui itself.

I think it would be reasonable to try to get this into 0.30 (releasing soon). Would you perhaps be interested in taking the reins on getting it done? If you are, I'd suggest taking this branch, rebasing on main, then adding your changes on top as commits to show the difference.

src/widgets.rs
joshka: (we moved to our own fork of the crate instability since then, so this is available now)

Issue #1062, opened 2024-04-25
Constraints example does not handle keys correctly on Windows


psobolik:
## Description

The Constraints example program does not work correctly on Windows. Key events are duplicated. 

## To Reproduce

Run the example from a Window terminal. Press any of the navigation keys -- up, down, left, right, h,  j, k or l.

## Expected behavior

The next or previous tab should be selected, or the content should be scrolled up or down one line.

## Environment

- OS: Windows
- Terminal Emulator: Windows Terminal or Cmd
- Font: N/A
- Crate version: N/A
- Backend: Crossterm

## Additional context

The problem is that the `handle_events` function does not apply the necessary `key.kind == KeyEventKind::Press` check as documented here: https://github.com/ratatui-org/ratatui/issues/347

joshka:
Hey there - I wonder if you'd like to submit a PR to fix this? It's probably just a single line fix similar to the other examples where we do the check.

PR #1066, opened 2024-04-26
fix(examples): fix key handling in constraints

psobolik:
Add check for `KeyEventKind::Press` to constraints example's event handler to eliminate double keys
on Windows.

Fixes: #1062

### DIFF
--- a/examples/constraints.rs
+++ b/examples/constraints.rs
@@ -108,18 +108,19 @@ impl App {
 
     fn handle_events(&mut self) -> Result<()> {
         if let Event::Key(key) = event::read()? {
-            if key.kind == KeyEventKind::Press {
-                use KeyCode::*;
-                match key.code {
-                    Char('q') | Esc => self.quit(),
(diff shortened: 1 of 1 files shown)


orhun (approved):

examples/constraints.rs
```diff
@@ -108,16 +108,18 @@ impl App {
...
-            match key.code {
-                Char('q') | Esc => self.quit(),
-                Char('l') | Right => self.next(),
-                Char('h') | Left => self.previous(),
-                Char('j') | Down => self.down(),
-                Char('k') | Up => self.up(),
-                Char('g') | Home => self.top(),
-                Char('G') | End => self.bottom(),
-                _ => (),
+            if key.kind == KeyEventKind::Press {
```
joshka, line 111: Generally try to use guard clauses over adding extra indentation to a block. E.g.:

```rust
if key.kind != KeyEventKind::Press {
    return Ok();
}
```

joshka (approved):

PR #1067, opened 2024-04-26
docs(github): update code owners

orhun:
Removes the team members that are not able to review PRs recently (with their approval ofc)


kdheepak:
I would not be opposed to listing out past maintainers in a separate section.

joshka (approved):

joshka:
> I would not be opposed to listing out past maintainers in a separate section.

I don't think that's possible in the code owners file (which is really just about who ends up getting review requests).

Issue #1068, opened 2024-04-26
User Input example doesn't work with UTF-8 characters


OkieOth:
## Description
If I am running the user_input example, then the application paniced when I am using e.g. 'ä' or other UTF-8 characters as input.

## To Reproduce
Run `cargo run --example=user_input --features=crossterm`
Enter in the input field multiple times an 'ä' (or similar)

## Expected behavior
Application should somehow not crash

## Environment
- OS: Linux
- Terminal Emulator: xterm
- Font: Intone Mone (Nerdfonts)

## Solution
The old example doesn't handle the use vector of String by chars. Instead the operations were based on bytes.

I will provide a PR that fixes that issue.

OkieOth:
PR is there ... be kind is my first one x-D

PR #1069, opened 2024-04-26
fix(examples): changed user_input example to work with multi-byte unicode chars

OkieOth:
This is the proposed solution for issue #1068. It solves the bug in the user_input example with multi-byte UTF-8 characters as input.


joshka (approved):

joshka:
I changed this a bit, to instead calculate the byte index and use that to insert. There's still a bug with displaying double width characters however.

PR #1070, opened 2024-04-27
fix: fix a small clippy lint

joshka:
error: binding's name is too similar to existing binding


joshka:
Ugh - closing - this was in my local main branch not the remote.

PR #1071, opened 2024-04-27
chore: fixup cargo lint for windows targets

joshka:
Crossterm brings in multiple versions of the same dep


a-kenji (approved):

a-kenji:
Was just about to send the same PR.

orhun (approved):

PR #1072, opened 2024-04-27
perf(crossterm): Speed up combined fg and bg color changes by up to 20%

joshka:
Use Crossterm SetColors instead of SetForegroundColor and
SetBackgroundColor.

In https://github.com/crossterm-rs/crossterm/pull/879 I changed the
SetColors command to write both colors at once with a single write
instead of multiple writes that more bytes. This led to a 15-25% fps
increase when testing the colors_rgb example on iTerm2 on an M2 Macbook
Pro.


orhun (approved):
very nice

PR #1074, opened 2024-04-27
fix(reflow): handle ZWSP

kxxt:
Handle ZWSP as a possible wrapping point.

See https://github.com/ratatui-org/ratatui/issues/293#issuecomment-2080824898 for more details.

### DIFF
--- a/src/widgets/reflow.rs
+++ b/src/widgets/reflow.rs
@@ -6,6 +6,7 @@ use unicode_width::UnicodeWidthStr;
 use crate::{layout::Alignment, text::StyledGrapheme};
 
 const NBSP: &str = "\u{00a0}";
+const ZWSP: &str = "\u{200b}";
 
 /// A state machine to pack styled symbols into lines.
 /// Cannot implement it as Iterator since it yields slices of the internal buffer (need streaming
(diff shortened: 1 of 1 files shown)


src/widgets/reflow.rs
```diff
@@ -6,6 +6,7 @@ use unicode_width::UnicodeWidthStr;
 use crate::{layout::Alignment, text::StyledGrapheme};
 
 const NBSP: &str = "\u{00a0}";
+const ZWSP: char = '\u{200b}';
```
orhun, line 9: Is there a reason of this being a `char` whereas `NBSP` is `&str`?

kxxt: > Is there a reason of this being a `char` whereas `NBSP` is `&str`?

It's because ZWSP is compared with another char while NBSP is compared with the symbol itself. I am not familiar with Graphemes, maybe I could use symbol == ZWSP here?

kxxt: > I am not familiar with Graphemes, maybe I could use symbol == ZWSP here?

Well, it seems that's the case:

```
>> :dep  unicode-segmentation
>> use unicode_segmentation::UnicodeSegmentation;
>> "\u{200b}\u{200b} \u{00a0}\u{200b} \u{00a0}".graphemes(false).collect::<Vec<_>>()
["\u{200b}", "\u{200b}", " ", "\u{a0}", "\u{200b}", " ", "\u{a0}"]
```

I updated it to use &str like NBSP.

orhun (approved):
LGTM, needs additional review/approval

joshka (approved):

PR #1078, opened 2024-05-01
docs(examples): Simplify scrollbar example

### LINKED ISSUE #975: Simplify scrollbar example code
the `ui` function triggers a lint that the method is too long and causes truncaction. - see #974 for info

marianomarciello:
This PR simplifies the scrollbar example.
Divide the function ui into multiple functions such that clippy does not throw the `clippy::too_many_lines` error anymore.
Handle correctly the cast from usize to u16.

fix #975

### DIFF
--- a/src/buffer/buffer.rs
+++ b/src/buffer/buffer.rs
@@ -303,15 +303,17 @@ impl Buffer {
     /// buffer.pos_of(100); // Panics
     /// ```
     #[must_use]
-    pub fn pos_of(&self, i: usize) -> (u16, u16) {
+    pub fn pos_of(&self, index: usize) -> (u16, u16) {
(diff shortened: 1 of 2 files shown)


EdJoPaTo:
Looks like no one ever took a longer look into this… Sorry for that!
Also, it has merge conflicts that don't look too friendly…

With the change in its current form I would have kept the ui function at the top to simplify the diff for review.

I think it's still a good idea to split this example up and the used approach seemed good. Maybe it's simpler to start new from the current main rather than solving these merge conflicts? Are you still interested in continuing with this @marianomarciello ?

joshka:
+1 to rewrite and make it pretty on this one

PR #1081, opened 2024-05-02
chore(github): use the GitHub organization team as codeowners

EdJoPaTo:
This is mainly an idea on how to improve this as the Team can be managed by organization admins in the GitHub GUI.
Also, we get mentioned as team rather than individually which helps with organizing notifications.

Past maintainers could end up as a past maintainer team in the organization.

Thought about creating an issue about this first but why not discuss this directly on the change.


orhun:
One thing that I don't like much is that this makes things less explicit. e.g. commits like 078e97e won't appear in the changelog.

> we get mentioned as team rather than individually which helps with organizing notifications

Doesn't it work as same as before? Can you explain how this helps organizing the notifications?

> Past maintainers could end up as a past maintainer team in the organization.

Does that mean we need to create another team for past maintainers? Or how does it work?

kdheepak:
> One thing that I don't like much is that this makes things less explicit. e.g. commits like https://github.com/ratatui-org/ratatui/commit/078e97e4ff65c02afa7c884914ecd38a6e959b58 won't appear in the changelog.

Do we need a separate file to track past and present maintainers? That information is useful for a potential user evaluating this package. In some projects I've worked on we've had a `MAINTAINERS.md` or `CONTRIBUTORS.md` file for that.

orhun:
Maybe it feels like a boomer thing to do but I think it's nice to have a file (or a place) to track the organizational changes both from security and transparency standpoint. 

`MAINTAINERS.md` (or something similar) is a good idea, but currently we already have that functionality with `CODEOWNERS`. I'm okay with having `MAINTAINERS.md` and using the team name in `CODEOWNERS` though.

EdJoPaTo:
> Doesn't it work as same as before? Can you explain how this helps organizing the notifications?

It might be the same because both would be a review request… I thought it's a person vs team mention but it isn't. 😒

orhun:
Yeah I see. Then I don't really see a huge benefit from this change - what the others think? @ratatui-org/maintainers

joshka:
I think I'm with @EdJoPaTo on this one. Use the github team to manage users. I basically never check my github email notifications (as they're just a searchable backup of the site data for me), but I can empathize with those that do.

I suspect the headers for review requests will change to the team, which would make filtering easier, but I don't actually know for sure. Also, the list of reviewers in the top right will change to just the team.
```
X-GitHub-Sender: EdJoPaTo
X-GitHub-Recipient: joshka
X-GitHub-Reason: review_requested
```

I don't necessary think we need to worry about keeping maintainer history. Slap it in a markdown file somewhere (readme / contributors / credits / maintainers / thanks / something).

I'd like to see what happens with this, so slapping approve on it. I don't think there was a hard objection to this, and it's reversible, so if you feel like it makes sense to try go for it @EdJoPaTo

joshka (approved):

joshka:
https://github.com/orgs/ratatui-org/teams/previous-maintainers/ created https://github.com/orgs/ratatui-org/teams/maintainers/ updated

orhun:
I would prefer to have `MAINTAINERS.md` if we decide to go ahead with this.

orhun (changes requested):
I would prefer to have MAINTAINERS.md if we decide to go ahead with this.

EdJoPaTo:
As there seems to be not that much interest in it (silent for weeks) I'll just close this.

EdJoPaTo:
I think this will also improve direct review requests. You can filter for review requests on your team or you as a person directly which might be neat.

joshka (approved):

orhun (approved):

PR #1082, opened 2024-05-02
perf(line): render line without clone

EdJoPaTo:
Due to the truncation the non truncation path used to_owned which cloned the line.

Calling an inner function on both cases resolves this as the temporary truncated line lives long enough for that function to end.

---

I have not benchmarked this. This is part of the improvements I did with #1049.

### DIFF
--- a/src/text/line.rs
+++ b/src/text/line.rs
@@ -584,34 +584,39 @@ impl Widget for Line<'_> {
 
 impl WidgetRef for Line<'_> {
     fn render_ref(&self, area: Rect, buf: &mut Buffer) {
+        fn render_spans(spans: &[Span], area: Rect, buf: &mut Buffer, mut x: u16) {
+            for span in spans {
+                let span_width = span.width() as u16;
+                let span_area = Rect {
+                    x,
+                    width: span_width.min(area.right().saturating_sub(x)),
+                    ..area
+                };
+                span.render(span_area, buf);
+                x = x.saturating_add(span_width);
+                if x >= area.right() {
+                    break;
+                }
+            }
+        }
+
         let area = area.intersection(buf.area);
         buf.set_style(area, self.style);
-        let width = self.width() as u16;
+
+        let line_width = self.width() as u16;
         let mut x = area.left();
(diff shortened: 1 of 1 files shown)


src/text/line.rs
```diff
@@ -584,34 +584,39 @@ impl Widget for Line<'_> {
...
-                ..area
-            };
-            span.render(span_area, buf);
-            x = x.saturating_add(span_width);
-            if x >= area.right() {
-                break;
-            }
-        }
     }
 }
```
joshka, line 621:
```suggestion
impl WidgetRef for Line<'_> {
    fn render_ref(&self, area: Rect, buf: &mut Buffer) {
        let area = area.intersection(buf.area);
        buf.set_style(area, self.style);

        let line_width = self.width() as u16;
        if line_width <= area.width {
            let offset = match self.alignment {
                Some(Alignment::Center) => (area.width.saturating_sub(line_width)) / 2,
                Some(Alignment::Right) => area.width.saturating_sub(line_width),
                Some(Alignment::Left) | None => 0,
            };
            render_spans(&self.spans, area, buf, area.x + offset);
        } else {
            let line = self.truncated(area.width);
            render_spans(&line.spans, area, buf, area.x);
        };
    }
}

fn render_spans(spans: &[Span], area: Rect, buf: &mut Buffer, mut x: u16) {
    for span in spans {
        let span_width = span.width() as u16;
        let available_width = area.right().saturating_sub(x);
        let width = span_width.min(available_width);
        let span_area = Rect { x, width, ..area };
        span.render(span_area, buf);
        x = x.saturating_add(span_width);
        if x >= area.right() {
            break;
        }
    }
}
```
Changes / rationale:
- x < y reads more obvious than x > y in code, so the if statement if flipped.
- area.x + offset < buf.area.right() < u16::max, so saturating add isn't necessary
- render_spans moved outside of the method so the reading order is top to bottom
- gave names to the sub calculations in render spans to make them a bit more obvious

EdJoPaTo:
Superseded by #1089

PR #1083, opened 2024-05-02
refactor(buffer): simplify set_stringn logic

EdJoPaTo:
Simplifies the logic. Instead of trying to figure out the `max_offset` in a seemingly complicated way keep the `remaining_width` and count down as long as there are characters left. This is way easier to understand for me what is going on.

---

Part of the improvements I did with #1049. As this is without `_opt` methods this can still panic (same as current code on wrong input).

### DIFF
--- a/src/buffer/buffer.rs
+++ b/src/buffer/buffer.rs
@@ -195,7 +195,7 @@ impl Buffer {
     /// Use [`Buffer::set_string`] when the maximum amount of characters can be printed.
     pub fn set_stringn<T, S>(
         &mut self,
-        x: u16,
+        mut x: u16,
         y: u16,
         string: T,
         max_width: usize,
@@ -205,7 +205,6 @@ impl Buffer {
         T: AsRef<str>,
         S: Into<Style>,
     {
-        let mut index = self.index_of(x, y);
         let max_width = max_width.try_into().unwrap_or(u16::MAX);
         let mut remaining_width = self.area.right().saturating_sub(x).min(max_width);
         let graphemes = UnicodeSegmentation::graphemes(string.as_ref(), true)
@@ -216,25 +215,17 @@ impl Buffer {
                 Some((symbol, width))
             });
         let style = style.into();
-        let mut x_offset = x;
         for (symbol, width) in graphemes {
-            self.content
-                .get_mut(index)
-                .expect("Index only exists when inside the buffer area")
-                .set_symbol(symbol)
-                .set_style(style);
+            self.get_mut(x, y).set_symbol(symbol).set_style(style);
+            let next_symbol = x + width;
+            x += 1;
             // Reset following cells if multi-width (they would be hidden by the grapheme),
-            let index_after_symbol = index + width as usize;
-            for i in (index + 1)..index_after_symbol {
-                self.content
-                    .get_mut(i)
-                    .expect("Index only exists when inside the buffer area")
-                    .reset();
+            while x < next_symbol {
+                self.get_mut(x, y).reset();
+                x += 1;
             }
-            index = index_after_symbol;
-            x_offset += width;
         }
-        (x_offset, y)
+        (x, y)
     }
 
     /// Print a line, starting at the position (x, y)
--- a/src/buffer/buffer.rs
+++ b/src/buffer/buffer.rs
@@ -1,8 +1,3 @@
-use std::{
-    cmp::min,
-    fmt::{Debug, Formatter, Result},
-};
-
 use unicode_segmentation::UnicodeSegmentation;
 use unicode_width::UnicodeWidthStr;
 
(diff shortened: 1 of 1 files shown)


src/buffer/buffer.rs
```diff
@@ -186,65 +181,65 @@ impl Buffer {
     }
 
     /// Print a string, starting at the position (x, y)
-    ///
-    /// `style` accepts any type that is convertible to [`Style`] (e.g. [`Style`], [`Color`], or
-    /// your own type that implements [`Into<Style>`]).
```
joshka, line 191: The removed comment is pretty standard language we use across the library. It's focused on newer users that might not be as savvy about how the `Into` idiom works as a prompt to find out more. While definitely not necessary from a more advanced user perspective, I'd prefer to keep this.

src/buffer/buffer.rs
```diff
@@ -186,65 +181,65 @@ impl Buffer {
...
+            .map_while(|(symbol, width)| {
+                remaining_width = remaining_width.checked_sub(width)?;
+                Some((symbol, width))
+            });
+        let style = style.into();
+        let mut x_offset = x;
+        for (symbol, width) in graphemes {
+            self.content
+                .get_mut(index)
+                .expect("Index only exists when inside the buffer area")
```
joshka, line 223: This message seems a bit vague to me. From the caller of this method's perspective there is no index parameter, so this would make no sense to the user.

src/buffer/buffer.rs
```diff
@@ -186,65 +181,65 @@ impl Buffer {
...
+        let graphemes = UnicodeSegmentation::graphemes(string.as_ref(), true)
+            .map(|symbol| (symbol, symbol.width() as u16))
+            .filter(|(_symbol, width)| *width > 0)
+            .map_while(|(symbol, width)| {
+                remaining_width = remaining_width.checked_sub(width)?;
+                Some((symbol, width))
+            });
+        let style = style.into();
+        let mut x_offset = x;
+        for (symbol, width) in graphemes {
```
joshka, line 220: This whole block can be simplified further and `index` can be removed (change x to `mut x` in the method params).

```rust
        for (symbol, width) in graphemes {
            self.get_mut(x, y).set_symbol(symbol).set_style(style);
            let next_symbol = x + width;
            x += 1;
            // Reset following cells if multi-width (they would be hidden by the grapheme),
            while x < next_symbol {
                self.get_mut(x, y).reset();
                x += 1;
            }
        }
        (x, y)
```

src/buffer/buffer.rs
```diff
@@ -186,65 +181,65 @@ impl Buffer {
...
-        self.set_stringn(x, y, string, usize::MAX, style.into());
+        self.set_stringn(x, y, string, usize::MAX, style);
     }
 
     /// Print at most the first n characters of a string if enough space is available
-    /// until the end of the line
+    /// until the end of the line.
     ///
-    /// `style` accepts any type that is convertible to [`Style`] (e.g. [`Style`], [`Color`], or
-    /// your own type that implements [`Into<Style>`]).
```
joshka, line 204: Same

src/buffer/buffer.rs
```diff
@@ -385,15 +379,15 @@ impl Buffer {
...
+impl std::fmt::Debug for Buffer {
     /// Writes a debug representation of the buffer to the given formatter.
     ///
     /// The format is like a pretty printed struct, with the following fields:
     /// * `area`: displayed as `Rect { x: 1, y: 2, width: 3, height: 4 }`
     /// * `content`: displayed as a list of strings representing the content of the buffer
     /// * `styles`: displayed as a list of: `{ x: 1, y: 2, fg: Color::Red, bg: Color::Blue,
     ///   modifier: Modifier::BOLD }` only showing a value when there is a change in style.
-    fn fmt(&self, f: &mut Formatter<'_>) -> Result {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
```
joshka, line 390: I've seen this change in other PRs, but was wasn't sure what the purpose of it was. Can you clarify this?
I'd generally by practice try to avoid writing a bare Result like this to avoid confusion, so I can understand it there, but `Debug` and `Formatter` seem benign enough.

joshka (commented):
Looks good. A couple of comments

src/buffer/buffer.rs
EdJoPaTo:
```suggestion
                .unwrap() // Index only exists when inside the buffer area
```

src/buffer/buffer.rs
EdJoPaTo: Mainly: get rid of the `use Result` which at best confuses and at worst introduces subtle bugs in other parts of the code without noticing. I also like to see what exactly is implemented for the local struct and as its most often only one time it's imported to be used once. So less clear to read and adds import which is used only once.

src/buffer/buffer.rs
EdJoPaTo: Users are unlikely playing around with the buffer directly. Only people implementing widgets are at the buffer.

Also: I dislike Into because it's less clear what happens. This only adds to it. Requiring a comment a dozen times all over the place to clarify something hard to understand hints a code smell that should be improved.

src/buffer/buffer.rs
joshka: Sounds fine.

src/buffer/buffer.rs
joshka: This goes away with the simplification, and you get errors about (x, y) instead coming from the get_mut(x, y).
I had an error in the simplification when I first wrote it, where the `x+=1` came before the `next_symbol = x+width` which highlighted this.

src/buffer/buffer.rs
EdJoPaTo: The usage of index here is likely a performance improvement. But that code is >8 years old. Using `get_mut(x, y)` needs to calculate the index again every call, but this removes the need to know that the index +1 is x +1.

Probably doesn't matter performance wise… unwrap vs. calculation… But easier to read and less assumptions about the index.

src/buffer/buffer.rs
joshka: I understand your concern over the use of `Into`, but it's something we're definitely keeping. Requiring code to construct a full Style in order to set colors or modifiers is overly verbose.

> Only people implementing widgets

I've generally been pushing the idea that implementing widgets is a thing that almost every developer should do in most apps (`impl Widget for &App`). E.g. see https://github.com/ratatui-org/ratatui/pull/1079. This is in contrast to writing functions that call methods on Frame, which tends to produce more convoluted apps.

Ideally in the future, this method should not generally be used. People should just use Span / Line / Text instead of this for greater flexibility.

src/buffer/buffer.rs
EdJoPaTo: "goes away" → silently panic via `debug_assert` now :ghost:

EdJoPaTo: x/y should be correct for the first call (this is the same behavior of set_stringn before this) and following calls are ensured to be correct by remaining width. The `expect` was mainly a "this should never fail when the above code is correct" message.

In the long run #1011 is a good idea, but this method is unlikely to be problematic after the first call. And as no one seems to care about this panicing on bad x/y input it's not that important to fix this quickly.

src/buffer/buffer.rs
joshka: Resolving for now as I don't feel strongly enough about this specific use to block it.

joshka (approved):

src/buffer/buffer.rs
joshka: Did this more uniformly in #1087

PR #1084, opened 2024-05-03
feat(buffer): add Buffer::cell, cell_mut and index implementations

joshka:
Code which previously called `buf.get(x, y)` or `buf.get_mut(x, y)`
should now use index operators, or be transitioned to `buff.cell()` or
`buf.cell_mut()` for safe access that avoids panics by returning
`Option<&Cell>` and `Option<&mut Cell>`.

The new methods accept `Into<Position>` instead of `x` and `y`
coordinates, which makes them more ergonomic to use.

```rust
let mut buffer = Buffer::empty(Rect::new(0, 0, 10, 10));

let cell = buf[(0, 0)];
let cell = buf[Position::new(0, 0)];

let symbol = buf.cell((0, 0)).map(|cell| cell.symbol());
let symbol = buf.cell(Position::new(0, 0)).map(|cell| cell.symbol());

buf[(0, 0)].set_symbol("🐀");
buf[Position::new(0, 0)].set_symbol("🐀");

buf.cell_mut((0, 0)).map(|cell| cell.set_symbol("🐀"));
buf.cell_mut(Position::new(0, 0)).map(|cell| cell.set_symbol("🐀"));
```

The existing `get()` and `get_mut()` methods are marked as deprecated.
These are fairly widely used and we will leave these methods around on
the buffer for a longer time than our normal deprecation approach (2
major release)

Addresses part of: https://github.com/ratatui-org/ratatui/issues/1011


src/buffer/buffer.rs
```diff
@@ -370,6 +449,52 @@ impl Buffer {
     }
 }
 
+impl<P: Into<Position>> Index<P> for Buffer {
```
EdJoPaTo, line 452: Increase readability by clarifying what exactly is implemented here inline. When first reading this I thought there is yet another ratatui trait. (same for `IndexMut`)

```suggestion
impl<P: Into<Position>> std::ops::Index<P> for Buffer {
```

src/buffer/buffer.rs
```diff
@@ -124,23 +196,30 @@ impl Buffer {
...
-        ((y - self.area.y) * self.area.width + (x - self.area.x)) as usize
+        self.index_of_opt(x, y).unwrap_or_else(|| {
+            panic!(
+                "index outside of buffer: the area is {area:?} but index is ({x}, {y})",
+                area = self.area,
+            )
+        })
+    }
+
+    const fn index_of_opt(&self, x: u16, y: u16) -> Option<usize> {
```
EdJoPaTo, line 214: doc comment

src/buffer/buffer.rs
```diff
@@ -124,23 +196,30 @@ impl Buffer {
...
     /// # use ratatui::prelude::*;
-    /// let rect = Rect::new(200, 100, 10, 10);
-    /// let buffer = Buffer::empty(rect);
+    /// let buffer = Buffer::empty(Rect::new(200, 100, 10, 10));
     /// // Top coordinate is outside of the buffer in global coordinate space, as the Buffer's area
     /// // starts at (200, 100).
     /// buffer.index_of(0, 0); // Panics
     /// ```
     #[track_caller]
     pub fn index_of(&self, x: u16, y: u16) -> usize {
```
EdJoPaTo, line 205: Given the low usage of `index_of` internally and externally I would consider changing it into `index_of_opt` directly.

https://github.com/search?q=ratatui+%22.index_of%28%22+AND+%28NOT+org%3Aratatui-org%29+AND+%28NOT+is%3Afork%29&type=code

src/buffer/buffer.rs
```diff
@@ -1,7 +1,9 @@
+use std::ops::{Index, IndexMut};
+
 use unicode_segmentation::UnicodeSegmentation;
 use unicode_width::UnicodeWidthStr;
 
-use crate::{buffer::Cell, prelude::*};
+use crate::{buffer::Cell, layout::Position, prelude::*};
```
EdJoPaTo, line 6: While I still dislike prelude… either `layout::Position` should be part of prelude or prelude shouldn't be used.

```suggestion
use crate::{
    buffer::Cell,
    layout::{Position, Rect},
    style::Style,
    text::{Line, Span},
};
```

With this it's also easily visible that something seems to use Line / Span here and hints your thoughts about removing set_line/span.

EdJoPaTo (changes requested):
Is there a reason why this was not done with `pos_of` too?

#1049: https://github.com/ratatui-org/ratatui/blob/30a6fb2e1335d408547f04d0c1a0f527d578e647/src/buffer/buffer.rs#L241-L250

src/buffer/buffer.rs
joshka: Yeah, I actually had a change as part of this which added `Position` to prelude, but I wasn't entirely sure about it. The rationale in the past for not putting Position and Size in the prelude was they are names that are a little too common. Making that change is potentially breaking, so It belongs outside of this PR.

I'm not sure that I follow why making Line/Span visible matters.

src/buffer/buffer.rs
joshka: I'd prefer to make the `content` field and the methods that deal directly with indexes (`index_of`, `pos_of`) private as it may allow us to migrate to more efficient storage approaches than we currently use. For now, I'd prefer to keep this as an additive only change and do the larger things separately after some discussion.

EdJoPaTo: > I'd prefer to make the `content` field and the methods that deal directly with indexes (`index_of`, `pos_of`) private as it may allow us to migrate to more efficient storage approaches than we currently use.

The doc comment should tell the users then to not use this and to look at get_opt instead. Otherwise this will get lost in some PR comments and will not be read in the future by users of index_of.

src/buffer/buffer.rs
EdJoPaTo: > I'm not sure that I follow why making Line/Span visible matters.

https://github.com/ratatui-org/ratatui/pull/1049#pullrequestreview-2012084663

It basically highlights the dependency graph more clearly. Low level stuff shouldn’t use high level stuff and create circles which is the rationale behind moving set_line/span out of buffer.

joshka: BTW, Position is a type name also used in the block title, which we want to remove soon, so this is not a hypothetical issue. Adding Position to the prelude breaks our own code right now. :D

joshka: (also in case you didn't see the email I sent, https://users.rust-lang.org/t/to-use-prelude-or-to-not-to-use-prelude-that-is-the-question/110855 is a topic I posted to get some more broader perspective on prelude use).

src/buffer/buffer.rs
```diff
@@ -91,19 +91,90 @@ impl Buffer {
...
+    ///     buffer.get_mut_opt(Position::new(0, 0)),
+    ///     Some(&mut Cell::default())
+    /// );
+    /// assert_eq!(buffer.get_mut_opt(Position::new(10, 10)), None);
+    /// ```
+    pub fn get_mut_opt<P: Into<Position>>(&mut self, pos: P) -> Option<&mut Cell> {
+        let pos = pos.into();
+        let index = self.index_of_opt(pos.x, pos.y)?;
+        self.content.get_mut(index)
+    }
```
kdheepak, line 176: Should this function name be `get_mut_opt` or `get_opt_mut`? 

It returns `Option<&mut Cell>` and not `&mut Option<Cell>`, so reading left to right would return in `get_opt_mut`. 

https://github.com/search?q=get_opt_mut+language%3Arust&type=code has 143 results
https://github.com/search?q=get_mut_opt+language%3Arust&type=code has 19 results

joshka: I think probably reading inside out might make more sense - I'd anticipate something that is suffixed `_mut` to return `&mut something` generally. I'm not 100% sure though. I wouldn't necessarily worry about quantity of results here and instead be more concerned more about being consistent with a reasonably obvious rule.

kdheepak: I don't know about the reading inside out? How do you read `Arc<Mutex<T>`? The only part I read inside out in Rust is "mutable reference"; I think everything else I tend to read left to right.

btw, if we introduce `RenderContext`, would we have `RenderContext::get` and `RenderContext::get_mut` return `Option<T>`?

joshka: Rust has settled on using suffixes to modify the type of thing returned, not prefixes
E.g. https://rust-lang.github.io/rfcs/0199-ownership-variants.html

`_mut` is a modifier to the thing that came before.
- If `get` returns a `Cell`, then `get_mut` returns `&mut Cell`
- If `get_opt` returns an `Option<Cell>` then get_opt_mut returns `&mut Option<Cell>`

More generally when you apply a suffix to an item, you do so in the order from right to left, so IMHO:
- `get_opt_mut` should return &mut<Option<T>>
- `get_mut_opt` should return Option<&mut T>

Regardless, `cell` and `cell_mut` are a couple of better names (or `get_cell`/`get_cell_mut`) that would avoid this altogether.

src/buffer/buffer.rs
joshka: Resolving - will handle this across the board rather than individually.

src/buffer/buffer.rs
joshka: > The doc comment should tell the users then to not use this and to look at get_opt instead. Otherwise this will get lost in some PR comments and will not be read in the future by users of index_of.

Out of scope for this PR.

kdheepak:
This looks good to me but maybe we can get more reviewers on this? 

PR title + description should be updated.

EdJoPaTo (commented):
I like the usage of the Index trait. It is still panicking like before but should have the same performance.

joshka:
I actually intended to put the deprecation of the existing methods in https://github.com/ratatui-org/ratatui/pull/1123, but accidentally pushed the commit to this PR. Happy to do it either way (though it would be easier at this point to just keep it in this PR and update the main PR comment.

src/buffer/buffer.rs
```diff
@@ -93,19 +93,91 @@ impl Buffer {
...
     pub fn get_mut(&mut self, x: u16, y: u16) -> &mut Cell {
         let i = self.index_of(x, y);
         &mut self.content[i]
     }
 
+    /// Returns a reference to Cell at the given coordinates.
+    ///
+    /// Returns `None` if the index is out of bounds.
+    ///
+    /// Note that unlike `get`, this method accepts a `Position` instead of `x` and `y` coordinates.
```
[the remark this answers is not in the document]
joshka: Replacing this with some better docs in the next commit.

src/buffer/buffer.rs
```diff
@@ -126,23 +197,35 @@ impl Buffer {
...
+            )
+        })
+    }
+
+    /// Returns the index in the `Vec<Cell>` for the given global (x, y) coordinates.
+    ///
+    /// Returns `None` if the given coordinates are outside of the Buffer's area.
+    ///
+    /// Note that this is private because of <https://github.com/ratatui-org/ratatui/issues/1122>
+    const fn index_of_opt(&self, x: u16, y: u16) -> Option<usize> {
```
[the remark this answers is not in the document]
joshka: Done

src/buffer/buffer.rs
joshka: I changed this back to just Index as the trait has to be in scope for links to a trait defined method in a doc comment to resolve correctly.

src/buffer/buffer.rs
joshka: Ended up being in scope

src/buffer/buffer.rs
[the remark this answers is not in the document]
joshka: 7e0b6e47

src/buffer/buffer.rs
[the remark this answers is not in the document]
EdJoPaTo: `area` and `content` also look like must_use candidates… Guess we should get that lint enabled for the full repo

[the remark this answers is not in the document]
joshka: area and content are not relevant to this PR.

src/buffer/buffer.rs
```diff
@@ -126,23 +209,37 @@ impl Buffer {
...
+
+    /// Returns the index in the `Vec<Cell>` for the given global (x, y) coordinates.
+    ///
+    /// Returns `None` if the given coordinates are outside of the Buffer's area.
+    ///
+    /// Note that this is private because of <https://github.com/ratatui-org/ratatui/issues/1122>
+    #[must_use]
+    const fn index_of_opt(&self, x: u16, y: u16) -> Option<usize> {
+        let area = self.area;
+        if x < area.left() || x >= area.right() || y < area.top() || y >= area.bottom() {
```
EdJoPaTo, line 236: Maybe even accept Position as an argument like the new cell(_mut) method has?

```suggestion
        if !area.contains(Position {x, y}) {
```

src/buffer/buffer.rs
```diff
@@ -23,22 +26,22 @@ use crate::{buffer::Cell, prelude::*};
 ///     width: 10,
 ///     height: 5,
 /// });
-/// buf.get_mut(0, 2).set_symbol("x");
-/// assert_eq!(buf.get(0, 2).symbol(), "x");
+/// buf[(0, 2)].set_symbol("x");
+/// assert_eq!(buf[(0, 2)].symbol(), "x");
```
EdJoPaTo, line 30: Maybe add an example of cell(_mut) here to highlight both ways of accessing cells? Probably with `?` in that case.

EdJoPaTo (approved):

src/buffer/buffer.rs
joshka: d2321a458

src/buffer/buffer.rs
joshka: d2321a458

PR #1087, opened 2024-05-05
style: use std::fmt instead of importing Debug and Display

joshka:
Based on a [discussion](https://github.com/ratatui-org/ratatui/pull/1083#discussion_r1588466060) with @EdJoPaTo on a PR.

This is a small universal style change to avoid making this change a part of other PRs.

### DIFF
--- a/src/backend/test.rs
+++ b/src/backend/test.rs
@@ -2,7 +2,7 @@
 //! It is used in the integration tests to verify the correctness of the library.
 
 use std::{
-    fmt::{Display, Write},
+    fmt::{self, Write},
     io,
 };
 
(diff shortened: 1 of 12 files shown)


joshka:
Lint error is fixed in https://github.com/ratatui-org/ratatui/pull/1088

kdheepak (approved):

EdJoPaTo:
I would have went with `std::fmt::Debug` as it only affects two likes to use `std::fmt` instead of `fmt` and is a bit more concise on what exactly was impl here.

also side note: merging quickly prevents more feedback. Especially when PR are open for weeks normally.

joshka:
> I would have went with `std::fmt::Debug` as it only affects two likes to use `std::fmt` instead of `fmt` and is a bit more concise on what exactly was impl here.

`use std::fmt` comes directly from the docs, which I'd treat as idiomatic (as well as being aligned with what I see as subjectively tasteful).
- https://doc.rust-lang.org/std/fmt/trait.Display.html#examples
- https://doc.rust-lang.org/std/fmt/trait.Debug.html#examples
- https://doc.rust-lang.org/std/fmt/index.html#fmtdisplay-vs-fmtdebug

> also side note: merging quickly prevents more feedback. Especially when PR are open for weeks normally.

Agreed. Apologies. I should have waited. Ideally PRs shouldn't be hitting weeks, but I understand that's been happening a bit more than usual lately. To some extent all the other maintainers have had various travel plans / other stuff going on in their life.

Issue #75, opened 2023-02-21
unicode-width and emojis


markus-bauer:
I'm not a user of helix or tui. But I found this helix issue and thought I'd mention it here:
https://github.com/helix-editor/helix/issues/6012

Helix uses a fork of a subset of tui. Notably the core of Buffer is basically the same (AFAIK).

I'm pretty sure that this bug is shared by everything using the tui-style Buffer.
AFAIK, the reason is that unicode-width doesn't (or can't?) report the column width for emojis. See readme and issues:
https://github.com/unicode-rs/unicode-width

I won't have anything to contribute here, or any personal interest, really.
Perhaps you should get into contact with the helix team for this, as this effects both.

parasyte:
`unicode-width` is not the right tool for computing column widths, since it depends on the font. `tui` won't have access to information about the font, since that is the terminal emulator's job.

To complicate matters further, not every terminal emulator handles emojis the same way. For instance, Windows Terminal is decent for the most part, but renders the "woman scientist" emoji from the `unicode-width` readme using 5 columns. This is of course a bug in Windows Terminal. It should only be 2 columns.

`unicode-segmentation` does the right thing when breaking strings into grapheme clusters. You just need a way to determine whether your terminal emulator will draw the grapheme cluster as 1 column or 2 (or 5, lol).

mindoodoo:
Has this resulted in a bug or particular issue during your use ?

LeoRiether:
> Has this resulted in a bug or particular issue during your use ?

It has for me. Because `unicode-width` returns 1, but kitty (the terminal) displays some emojis with width 2, the layout can be broken in some places. For example, if we render a block with a heart emoji on the title, the top-right corner is rendered at the wrong spot.

![heart-titled block](https://user-images.githubusercontent.com/8211902/234155141-1235182d-9d9d-43c6-9c24-6a03d699aca0.png)

joshka:
Thanks for this info - can you link to your fork?

I noticed a while back that Wez commented about their approach to this at https://github.com/unicode-rs/unicode-width/issues/4#issuecomment-549906181:




> I don't really know much about this space, but here's my attempt at dealing with this in a terminal emulator.

```rust
/// Returns the number of cells visually occupied by a sequence
/// of graphemes
pub fn unicode_column_width(s: &str) -> usize {
    use unicode_segmentation::UnicodeSegmentation;
    s.graphemes(true).map(grapheme_column_width).sum()
}

[...]

PR #1089, opened 2024-05-06
fix: unicode truncation bug

joshka:
- Rewrote the line / span rendering code to take into account how multi-byte / wide emoji characters are truncated when rendering into areas that cannot accommodate them in the available space
- Added comprehensive coverage over the edge cases
- Adds a benchmark to ensure perf

Fixes: https://github.com/ratatui-org/ratatui/issues/1032

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -164,6 +164,10 @@ name = "block"
 harness = false
 
 [[bench]]
+name = "line"
+harness = false
+
+[[bench]]
 name = "list"
 harness = false
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,6 +43,7 @@ document-features = { version = "0.2.7", optional = true }
 lru = "0.12.0"
 stability = "0.2.0"
 compact_str = "0.7.1"
+unicode-truncate = "1"
 
 [dev-dependencies]
 anyhow = "1.0.71"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,25 +25,24 @@ rust-version = "1.74.0"
 [badges]
 
 [dependencies]
-crossterm = { version = "0.27", optional = true }
-termion = { version = "3.0", optional = true }
-termwiz = { version = "0.22.0", optional = true }
-
-serde = { version = "1", optional = true, features = ["derive"] }
 bitflags = "2.3"
 cassowary = "0.3"
+compact_str = "0.7.1"
+crossterm = { version = "0.27", optional = true }
+document-features = { version = "0.2.7", optional = true }
 indoc = "2.0"
 itertools = "0.12"
+lru = "0.12.0"
 paste = "1.0.2"
+serde = { version = "1", optional = true, features = ["derive"] }
+stability = "0.2.0"
 strum = { version = "0.26", features = ["derive"] }
 time = { version = "0.3.11", optional = true, features = ["local-offset"] }
+termion = { version = "3.0", optional = true }
+termwiz = { version = "0.22.0", optional = true }
 unicode-segmentation = "1.10"
-unicode-width = "0.1"
-document-features = { version = "0.2.7", optional = true }
-lru = "0.12.0"
-stability = "0.2.0"
-compact_str = "0.7.1"
 unicode-truncate = "1"
+unicode-width = "0.1"
 
 [dev-dependencies]
 anyhow = "1.0.71"
--- a/src/text/line.rs
+++ b/src/text/line.rs
@@ -586,64 +586,70 @@ impl WidgetRef for Line<'_> {
 
 /// Renders all the spans of the line that should be visible. Visibility is determined by the area
 /// and the (remaining) `span_skip_width`.
-///
-/// Algorithm:
-/// 1. For each span check whether it is (fully) visible. Skip or render it (partially).
-/// 2. Indent `area` to account for the width of the rendered span
-/// 3. Update the remaining `span_skip_width`. This will be 0 after the first fully rendered span.
-/// 4. Repeat
 #[allow(clippy::cast_possible_truncation, clippy::arithmetic_side_effects)] // Ensured by debug_assert or explained in comment
 fn render_spans(spans: &[Span], mut area: Rect, buf: &mut Buffer, mut span_skip_width: usize) {
-    for span in spans {
-        let span_width = span.width();
-        // Ignore spans that are completely before the offset. Decrement `span_skip_width` by the
-        // span width until we find a span that is partially or completely visible.
-        if span_skip_width >= span_width {
-            span_skip_width -= span_width;
-            continue;
-        }
-        let span_only_partially_visible = span_skip_width > 0;
-        if span_only_partially_visible {
-            // The first visible span might only be partially visible. Truncate the start and render
-            // it starting at span_skip_width. All subsequent visible spans are rendered in full (as
-            // available space permits).
-            // As the end is truncated by the area width we only need to truncate the start.
-            debug_assert!(span_width > span_skip_width, "span_skip_width seems wrong");
-            let available_width = span_width - span_skip_width;
-            let (content, actual_width) = span.content.unicode_truncate_start(available_width);
-
-            // if the truncation has truncated an initial grapheme, then we need to start rendering
-            // from a position that takes that into account by indenting the start of the area
-            let offset_of_initial_grapheme = available_width - actual_width;
-
-            // as u16 is safe as a single grapheme is kinda short.
-            // Some emojis are 20, but not nearly close to u16::MAX
-            area = area.indent_x(offset_of_initial_grapheme as u16);
-            if area.is_empty() {
-                break;
+    let spans = spans
+        .iter()
+        .map(|span| (span, span.width()))
+        // Determine whether a span is visible or not. Filter non visible ones out.
+        .filter_map(|(span, span_width)| {
+            // Ignore spans that are completely before the offset. Decrement `span_skip_width` by
+            // the span width until we find a span that is partially or completely visible.
+            if span_skip_width >= span_width {
+                span_skip_width -= span_width;
+                return None;
             }
 
-            let visible = Span::styled(content, span.style);
-            visible.render_ref(area, buf);
-
-            debug_assert!(
-                u16::try_from(actual_width).is_ok(),
-                "span_skip_width seems wrong"
-            );
-            area = area.indent_x(actual_width as u16);
+            // This applies the skip from the start but not the end.
+            // The end will be trimmed by the actual render into the buffer.
+            // This means that the available_width might be > u16::MAX.
+            let available_width = span_width.saturating_sub(span_skip_width);
 
             span_skip_width = 0; // ensure that the next span is rendered in full
-        } else {
-            // render the whole span (until there is no area left)
-            if area.is_empty() {
-                break;
-            }
-            span.render_ref(area, buf);
 
-            // The span might be longer than u16. But the area only calculated up to u16
-            let span_width = u16::try_from(span_width).unwrap_or(u16::MAX);
-            area = area.indent_x(span_width);
+            Some((span, span_width, available_width))
+        })
+        // Handle truncation when needed
+        .map(|(span, span_width, available_width)| {
+            if span_width > available_width {
+                // Partially visible
+
+                // As the end is truncated by the area width we only need to truncate the start.
+                let (content, actual_width) = span.content.unicode_truncate_start(available_width);
+                // if the truncation has truncated an initial grapheme, then we need to start rendering
+                // from a position that takes that into account by indenting the start of the area
+                let offset_of_initial_grapheme = available_width - actual_width;
+
+                debug_assert!(
+                    u16::try_from(offset_of_initial_grapheme).is_ok(),
+                    "single graphemes are kinda short. Some Emojis reach 22 but not nearly close to u16::MAX"
+                );
+                let offset_of_initial_grapheme = offset_of_initial_grapheme as u16;
+
+                (
+                    content,
+                    span.style,
+                    actual_width,
+                    offset_of_initial_grapheme,
+                )
+            } else {
+                // Fully visible
+                (&*span.content, span.style, span_width, 0)
+            }
+        });
+
+    // Render the spans one after each other
+    for (content, style, span_width, offset) in spans {
+        area = area.indent_x(offset);
+        if area.is_empty() {
+            break;
         }
+
+        Span::styled(content, style).render_ref(area, buf);
+
+        // The span might be longer than u16. But the area only calculates up to u16
+        let render_width = u16::try_from(span_width).unwrap_or(u16::MAX);
+        area = area.indent_x(render_width);
     }
 }
 
--- a/src/text/line.rs
+++ b/src/text/line.rs
@@ -604,8 +604,8 @@ fn render_spans(
             width: area.width.saturating_sub(buf_x_offset),
             ..area
         };
-        let is_span_partially_visible = span_offset > 0;
-        if is_span_partially_visible {
+        let span_is_only_partially_visible = span_offset > 0;
+        if span_is_only_partially_visible {
             // render the partially visible span starting at the offset -> right side
             let (_left, right) = span.split_at_width(span_offset as usize);
             right.render_ref(area, buf);
--- a/src/text/line.rs
+++ b/src/text/line.rs
@@ -602,18 +602,19 @@ fn render_spans(spans: &[Span], mut area: Rect, buf: &mut Buffer, mut span_offse
         if span_offset > 0 {
             // the first span is only partially visible, so truncate the start and render it
             // starting at span offset
-            let render_width = span_width - span_offset;
-            let (content, display_width) =
-                span.content.unicode_truncate_start(render_width as usize);
-            let display_width = display_width as u16;
+            let available_width = span_width - span_offset;
+            let (content, actual_width) = span
+                .content
+                .unicode_truncate_start(available_width as usize);
+            let actual_width = actual_width as u16;
 
             // if the truncation has truncated an initial grapheme, then we need to start rendering
             // from a position that takes that into account by indenting the start of the area
-            area = area.indent_x(render_width - display_width);
+            area = area.indent_x(available_width - actual_width);
             let right = Span::styled(content, span.style);
             right.render_ref(area, buf);
 
-            area = area.indent_x(display_width);
+            area = area.indent_x(actual_width);
             span_offset = 0; // ensure that the next span is rendered in full
         } else {
             // render the whole span
--- a/src/text/line.rs
+++ b/src/text/line.rs
@@ -1,9 +1,10 @@
 #![deny(missing_docs)]
 use std::borrow::Cow;
 
+use unicode_truncate::UnicodeTruncateStr;
+
 use super::StyledGrapheme;
 use crate::prelude::*;
-use unicode_truncate::UnicodeTruncateStr;
 
 /// A line of text, consisting of one or more [`Span`]s.
 ///
@@ -940,7 +941,6 @@ mod tests {
         use unicode_width::UnicodeWidthStr;
 
         use self::buffer::Cell;
-
         use super::*;
         use crate::assert_buffer_eq;
         const BLUE: Style = Style::new().fg(Color::Blue);
--- a/src/text/line.rs
+++ b/src/text/line.rs
@@ -3,6 +3,7 @@ use std::borrow::Cow;
 
 use super::StyledGrapheme;
 use crate::prelude::*;
+use unicode_truncate::UnicodeTruncateStr;
 
 /// A line of text, consisting of one or more [`Span`]s.
 ///
@@ -560,14 +561,15 @@ impl WidgetRef for Line<'_> {
                 Some(Alignment::Right) => area.width.saturating_sub(line_width),
                 Some(Alignment::Left) | None => 0,
             };
-            render_spans(&self.spans, area, buf, offset, 0);
+            let area = indent(area, offset);
+            render_spans(&self.spans, area, buf, 0);
         } else {
             let offset = match self.alignment {
                 Some(Alignment::Center) => (line_width.saturating_sub(area.width)) / 2,
                 Some(Alignment::Right) => line_width.saturating_sub(area.width),
                 Some(Alignment::Left) | None => 0,
             };
-            render_spans(&self.spans, area, buf, 0, offset);
+            render_spans(&self.spans, area, buf, offset);
         };
     }
 }
@@ -582,16 +584,10 @@ impl WidgetRef for Line<'_> {
 ///    1. If the span is completely before the `span_offset`, skip it
 ///    2. If the span is partially before the `span_offset`, render the right side of the span
 ///    3. If the span is completely after the `span_offset`, render the whole span
-/// 2. Update `buf_x_offset` to account for the width of the rendered span
+/// 2. Indent `area` to account for the width of the rendered span
 /// 3. Update `span_offset` to 0 after the first span is rendered in full
 /// 4. Repeat
-fn render_spans(
-    spans: &[Span],
-    area: Rect,
-    buf: &mut Buffer,
-    mut buf_x_offset: u16,
-    mut span_offset: u16,
-) {
+fn render_spans(spans: &[Span], mut area: Rect, buf: &mut Buffer, mut span_offset: u16) {
     for span in spans {
         let span_width = span.width() as u16;
         // ignore spans that are completely before the offset
@@ -599,42 +595,42 @@ fn render_spans(
             span_offset -= span_width;
             continue;
         }
-        let area = Rect {
-            x: area.x.saturating_add(buf_x_offset),
-            width: area.width.saturating_sub(buf_x_offset),
-            ..area
-        };
         if area.is_empty() {
             break;
         }
-        let span_is_only_partially_visible = span_offset > 0;
-        if span_is_only_partially_visible {
-            // render the partially visible span starting at the offset -> right side
-            use unicode_truncate::UnicodeTruncateStr;
+        if span_offset > 0 {
+            // the first span is only partially visible, so truncate the start and render it
+            // starting at span offset
+            let render_width = span_width - span_offset;
+            let (content, display_width) =
+                span.content.unicode_truncate_start(render_width as usize);
+            let display_width = display_width as u16;
 
-            let limit = span_width - span_offset;
-            let (content, width) = span.content.unicode_truncate_start(usize::from(limit));
-            let width = width as u16;
-
-            let x = limit - width;
-
-            // Empty the cells before as there is a multi width grapheme not rendered before. But as
-            // it is not rendered in the first place (its truncated away) dont reset, set " ".
-            for x in area.x..x {
-                buf.get_mut(x, area.y).set_style(span.style).set_symbol(" ");
-            }
-
-            let span_area = Rect { x, width, ..area };
+            // if the truncation has truncated an initial grapheme, then we need to start rendering
+            // from a position that takes that into account by indenting the start of the area
+            area = indent(area, render_width - display_width);
             let right = Span::styled(content, span.style);
-            right.render_ref(span_area, buf);
+            right.render_ref(area, buf);
 
-            buf_x_offset += limit;
+            area = indent(area, display_width);
             span_offset = 0; // ensure that the next span is rendered in full
         } else {
             // render the whole span
             span.render_ref(area, buf);
-            buf_x_offset += span_width;
+            area = indent(area, span_width);
         }
+    }
+}
+
+/// indents a given `area`'s x value by the given `x_offset`, returning a new `Rect`
+///
+/// TODO: this should probably be move into a pubic method on `Rect` as it's a common operation
+/// that's used in a few places
+const fn indent(area: Rect, x_offset: u16) -> Rect {
+    Rect {
+        x: area.x.saturating_add(x_offset),
+        width: area.width.saturating_sub(x_offset),
+        ..area
     }
 }
 
@@ -1150,9 +1146,9 @@ mod tests {
 
         /// Ensures the rendering also works away from the 0x0 position.
         #[rstest]
-        #[case::left_6(Alignment::Left, 6, "a🦀bc ")]
+        #[case::left_6(Alignment::Left, 6, "a🦀bcX")]
         #[case::center_6(Alignment::Center, 6, "🦀bc🦀")]
-        #[case::right_6(Alignment::Right, 6, " bc🦀d")]
+        #[case::right_6(Alignment::Right, 6, "Xbc🦀d")]
         fn render_truncates_away_from_0x0(
             #[case] alignment: Alignment,
             #[case] buf_width: u16,
@@ -1175,7 +1171,7 @@ mod tests {
         #[rstest]
         #[case::right_4(4, "c🦀d")]
         #[case::right_5(5, "bc🦀d")]
-        #[case::right_6(6, " bc🦀d")]
+        #[case::right_6(6, "Xbc🦀d")]
         #[case::right_7(7, "🦀bc🦀d")]
         #[case::right_8(8, "a🦀bc🦀d")]
         fn render_right_aligned_multi_span(#[case] buf_width: u16, #[case] expected: &str) {
--- a/src/text/line.rs
+++ b/src/text/line.rs
@@ -1,4 +1,5 @@
 #![deny(missing_docs)]
+#![warn(clippy::cast_possible_truncation)]
 use std::borrow::Cow;
 
 use unicode_truncate::UnicodeTruncateStr;
@@ -555,23 +556,27 @@ impl WidgetRef for Line<'_> {
     fn render_ref(&self, area: Rect, buf: &mut Buffer) {
         let area = area.intersection(buf.area);
         buf.set_style(area, self.style);
-        let line_width = self.width() as u16;
-        if line_width <= area.width {
+
+        let area_width = usize::from(area.width);
+        let line_width = self.width();
+        #[allow(clippy::cast_possible_truncation)] // explained in comment
+        if line_width <= area_width {
             // there is enough space to render the whole line, but we may need to indent it
             let indent_width = match self.alignment {
-                Some(Alignment::Center) => (area.width.saturating_sub(line_width)) / 2,
-                Some(Alignment::Right) => area.width.saturating_sub(line_width),
+                Some(Alignment::Center) => (area_width.saturating_sub(line_width)) / 2,
+                Some(Alignment::Right) => area_width.saturating_sub(line_width),
                 Some(Alignment::Left) | None => 0,
             };
-            let area = area.indent_x(indent_width);
+            // cast as u16 is safe as area_width is u16
+            let area = area.indent_x(indent_width as u16);
             render_spans(&self.spans, area, buf, 0);
         } else {
             // there is not enough space to render the whole line, so we need to truncate it
             // and only render the visible part. We only care about truncating the left side of the
             // line, as the right side will be truncated by the area width
             let skip_width = match self.alignment {
-                Some(Alignment::Center) => (line_width.saturating_sub(area.width)) / 2,
-                Some(Alignment::Right) => line_width.saturating_sub(area.width),
+                Some(Alignment::Center) => (line_width.saturating_sub(area_width)) / 2,
+                Some(Alignment::Right) => line_width.saturating_sub(area_width),
                 Some(Alignment::Left) | None => 0,
             };
             render_spans(&self.spans, area, buf, skip_width);
@@ -594,9 +599,10 @@ impl WidgetRef for Line<'_> {
 /// 2. Indent `area` to account for the width of the rendered span
 /// 3. Update `span_skip_width` to 0 after the first span is rendered in full
 /// 4. Repeat
-fn render_spans(spans: &[Span], mut area: Rect, buf: &mut Buffer, mut span_skip_width: u16) {
+#[allow(clippy::cast_possible_truncation)] // Ensured by debug_assert or explained in comment
+fn render_spans(spans: &[Span], mut area: Rect, buf: &mut Buffer, mut span_skip_width: usize) {
     for span in spans {
-        let span_width = span.width() as u16;
+        let span_width = span.width();
         // ignore spans that are completely before the offset by decrementing `span_skip_width` by
         // the span width in each iteration until we find a span that is partially or completely
         // visible
@@ -607,24 +613,32 @@ fn render_spans(spans: &[Span], mut area: Rect, buf: &mut Buffer, mut span_skip_
         if span_skip_width > 0 {
             // the first visible span is only partially visible, so truncate the start and render it
             // starting at span_skip_width, all subsequent visible spans are rendered in full (as
-            // available space permits). We only need to truncate the start of the first visible span
-            // because the rest of the line will be truncated automatically by the area width
+            // available space permits). We only need to truncate the start of the first visible
+            // span because the rest of the line will be truncated automatically by the
+            // area width
             let available_width = span_width - span_skip_width;
-            let (content, actual_width) = span
-                .content
-                .unicode_truncate_start(available_width as usize);
-            let actual_width = actual_width as u16;
+            let (content, actual_width) = span.content.unicode_truncate_start(available_width);
 
             // if the truncation has truncated an initial grapheme, then we need to start rendering
             // from a position that takes that into account by indenting the start of the area
-            area = area.indent_x(available_width - actual_width);
+            let offset_of_initial_grapheme = available_width - actual_width;
+
+            // as u16 is safe as a single grapheme is kinda short.
+            // Some emojis are 20, but not nearly close to u16::MAX
+            area = area.indent_x(offset_of_initial_grapheme as u16);
             if area.is_empty() {
                 break;
             }
+
             let right = Span::styled(content, span.style);
             right.render_ref(area, buf);
 
-            area = area.indent_x(actual_width);
+            debug_assert!(
+                u16::try_from(actual_width).is_ok(),
+                "span_skip_width seems wrong"
+            );
+            area = area.indent_x(actual_width as u16);
+
             span_skip_width = 0; // ensure that the next span is rendered in full
         } else {
             // render the whole span
@@ -632,7 +646,12 @@ fn render_spans(spans: &[Span], mut area: Rect, buf: &mut Buffer, mut span_skip_
                 break;
             }
             span.render_ref(area, buf);
-            area = area.indent_x(span_width);
+
+            debug_assert!(
+                u16::try_from(span_width).is_ok(),
+                "span_skip_width should be > 0 when the span_width is too long"
+            );
+            area = area.indent_x(span_width as u16);
         }
     }
 }
@@ -942,9 +961,9 @@ mod tests {
         use unicode_segmentation::UnicodeSegmentation;
         use unicode_width::UnicodeWidthStr;
 
-        use self::buffer::Cell;
         use super::*;
-        use crate::assert_buffer_eq;
+        use crate::{assert_buffer_eq, buffer::Cell};
+
         const BLUE: Style = Style::new().fg(Color::Blue);
         const GREEN: Style = Style::new().fg(Color::Green);
         const ITALIC: Style = Style::new().add_modifier(Modifier::ITALIC);
@@ -1191,11 +1210,11 @@ mod tests {
         /// middle of the emoji. This test documents just the emoji part of the test.
         #[test]
         fn flag_emoji() {
-            let s = "🇺🇸1234";
-            assert_eq!(s.len(), 12); // flag is 4 bytes
-            assert_eq!(s.chars().count(), 6); // flag is 2 chars
-            assert_eq!(s.graphemes(true).count(), 5); // flag is 1 grapheme
-            assert_eq!(s.width(), 6); // flag is 2 display width
+            let str = "🇺🇸1234";
+            assert_eq!(str.len(), 12); // flag is 4 bytes
+            assert_eq!(str.chars().count(), 6); // flag is 2 chars
+            assert_eq!(str.graphemes(true).count(), 5); // flag is 1 grapheme
+            assert_eq!(str.width(), 6); // flag is 2 display width
         }
 
         /// Part of a regression test for <https://github.com/ratatui-org/ratatui/issues/1032> which
@@ -1213,6 +1232,50 @@ mod tests {
             let mut buf = Buffer::empty(Rect::new(0, 0, buf_width, 1));
             line.render_ref(buf.area, &mut buf);
             assert_buffer_eq!(buf, Buffer::with_lines([expected]));
+        }
+
+        // Buffer width is `u16`. A line can be longer.
+        #[test]
+        fn render_truncates_very_long_line_of_many_spans() {
+            let part = "this is some content with a somewhat long width to be repeated over and over again to create horribly long Line over u16::MAX";
+            let min_width = usize::from(u16::MAX).saturating_add(1);
+
+            // width == len as only ASCII is used here
+            let factor = min_width.div_ceil(part.len());
+
+            let line = Line::from(vec![Span::raw(part); factor]).right_aligned();
+
+            dbg!(line.width());
+            assert!(line.width() >= min_width);
+
+            let mut buf = Buffer::empty(Rect::new(0, 0, 32, 1));
+            line.render_ref(buf.area, &mut buf);
+            assert_buffer_eq!(
+                buf,
+                Buffer::with_lines(["horribly long Line over u16::MAX"])
+            );
+        }
+
+        // Buffer width is `u16`. A single span inside a line can be longer.
+        #[test]
+        fn render_truncates_very_long_single_span_line() {
+            let part = "this is some content with a somewhat long width to be repeated over and over again to create horribly long Line over u16::MAX";
+            let min_width = usize::from(u16::MAX).saturating_add(1);
+
+            // width == len as only ASCII is used here
+            let factor = min_width.div_ceil(part.len());
+
+            let line = Line::from(vec![Span::raw(part.repeat(factor))]).right_aligned();
+
+            dbg!(line.width());
+            assert!(line.width() >= min_width);
+
+            let mut buf = Buffer::empty(Rect::new(0, 0, 32, 1));
+            line.render_ref(buf.area, &mut buf);
+            assert_buffer_eq!(
+                buf,
+                Buffer::with_lines(["horribly long Line over u16::MAX"])
+            );
         }
     }
 
--- a/src/text/line.rs
+++ b/src/text/line.rs
@@ -453,39 +453,6 @@ impl<'a> Line<'a> {
         self.spans.iter_mut()
     }
 
-    /// Returns a line that's truncated corresponding to it's alignment and result width
-    #[must_use = "method returns the modified value"]
-    fn truncated(&'a self, result_width: u16) -> Self {
-        let mut truncated_line = Line::default();
-        let width = self.width() as u16;
-        let mut offset = match self.alignment {
-            Some(Alignment::Center) => (width.saturating_sub(result_width)) / 2,
-            Some(Alignment::Right) => width.saturating_sub(result_width),
-            _ => 0,
-        };
-        let mut x = 0;
-        for span in &self.spans {
-            let span_width = span.width() as u16;
-            if offset >= span_width {
-                offset -= span_width;
-                continue;
-            }
-            let mut new_span = span.clone();
-            let new_span_width = span_width - offset;
-            if x + new_span_width > result_width {
-                let span_end = (result_width - x + offset) as usize;
-                new_span.content = Cow::from(&span.content[offset as usize..span_end]);
-                truncated_line.spans.push(new_span);
-                break;
-            }
-
-            new_span.content = Cow::from(&span.content[offset as usize..]);
-            truncated_line.spans.push(new_span);
-            x += new_span_width;
-            offset = 0;
-        }
-        truncated_line
-    }
     /// Adds a span to the line.
     ///
     /// `span` can be any type that is convertible into a `Span`. For example, you can pass a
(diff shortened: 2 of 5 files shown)


joshka:
I'd suggest that for this PR it's probably worth:
- fixing up any obvious simplifications / problems I've missed
- disabling the broken tests (unless there's a really obvious test fix for the right alignment one)
- merging this and making a 0.26.3 release pretty soon.
- If there's an easy and obvious way that `unicode-truncate` helps simplify this, then we should consider that

Note this stems from https://github.com/ratatui-org/ratatui/pull/1082 credit goes to @EdJoPaTo for making it easier to see a way to make a fix for this.

Side note: there's also a [truncate](https://crates.io/crates/truncate) crate, which does various unicode split_at things, but it uses `unicode-segmentation` instead of `unicode-width` for calculating positions - there's some subtle incompatibilities with the approaches. The ideas / methods might be useful to consider for unicode-truncate however.

EdJoPaTo:
> Side note: there's also a [truncate](https://crates.io/crates/truncate) crate, which does various unicode split_at things, but it uses `unicode-segmentation` instead of `unicode-width` for calculating positions - there's some subtle incompatibilities with the approaches. The ideas / methods might be useful to consider for unicode-truncate however.

I also suggested unicode-segmentation for unicode-truncation as it currently rips apart stuff that belongs together like Emoji combinations (Flags for example): https://github.com/Aetf/unicode-truncate/pull/11

joshka:
> > Side note: there's also a [truncate](https://crates.io/crates/truncate) crate, which does various unicode split_at things, but it uses `unicode-segmentation` instead of `unicode-width` for calculating positions - there's some subtle incompatibilities with the approaches. The ideas / methods might be useful to consider for unicode-truncate however.
> 
> I also suggested unicode-segmentation for unicode-truncation as it currently rips apart stuff that belongs together like Emoji combinations (Flags for example): [Aetf/unicode-truncate#11](https://github.com/Aetf/unicode-truncate/pull/11)

There's a lengthy debate on to use / not use segmentation in https://github.com/ratatui-org/ratatui/issues/75 (ratatui's longest open issue). I don't know which side of that debate has more merit, and it's likely that using unicode-segmentation within the truncation crate would break things in other ways for many terminals / use cases.

src/text/span.rs
```diff
@@ -332,6 +332,37 @@ impl<'a> Span<'a> {
...
+    /// span will contain the remaining characters.
+    pub(crate) fn split_at_width(&'a self, width: usize) -> (Self, Self) {
+        // There's at least two ways to do this. I chose the second option as it iterates over the
+        // string once at the cost of always allocating two strings regardless of the split point.
+        // 1. iterate the chars to find the split point and then split the spans.
+        // 2. iterate the chars and build the spans at the same time.
+        let mut left = String::new();
+        let mut right = String::new();
+        let mut left_width = 0;
+        for c in self.content.chars() {
```
kdheepak, line 348: I _think_ this will cause issues with flag emojis

```rust
    let s = "🇺🇸1234";
    dbg!(s);
    dbg!(s.len());
    dbg!(s.chars().count());
    dbg!(s.width());
    dbg!(s.graphemes(true).count());
```

gives me

```
[src/main.rs:43:5] s = "🇺🇸1234"
[src/main.rs:44:5] s.len() = 12
[src/main.rs:45:5] s.chars().count() = 6
[src/main.rs:46:5] s.width() = 6
[src/main.rs:47:5] s.graphemes(true).count() = 5
```

I haven't thought about it extensively but can we do this instead? 

```suggestion 
        for s in self.content.graphemes(true) {
            for c in s.chars() {
```
      
```

Comment (kdheepak) [reply src/text/span.rs:348]: Maybe it is fine? It'll get split like this:

```
[src/main.rs:50:9] c = '🇺'
[src/main.rs:50:9] c = '🇸'
```

So it still won't panic.

Comment (kdheepak) [reply src/text/span.rs:348]: Oops, this is the same conversation that is happening on the main thread. Ignore these comments.

**Comment by joshka** [comment]
> There's an **EXTREMELY** lengthy debate on to use / not use segmentation in #75 (ratatui's longest open issue). I don't know which side of that debate has more merit, and it's likely that using unicode-segmentation within the truncation crate would break things in other ways for many terminals / use cases.

Also, I forgot that we do use segmentation when doing grapheme splitting, so maybe that is the right abstraction. I'm unsure how exactly to test an edge case on this one without delving a vast amount more into unicode than I want to right now.

File: src/text/span.rs (hunk 1 above)
Comment (kdheepak) [reply src/text/span.rs:348]: Can you try `1234🇺🇸`?

Comment (joshka) [reply src/text/span.rs:348]: No difference
```rust
#[case::flag_1(1, "1")]
        #[case::flag_2(2, "12")]
        #[case::flag_3(3, "123")]
        #[case::flag_4(4, "1234")]
        #[case::flag_5(5, "1234 ")]
        #[case::flag_6(6, "1234🇺🇸")]
        #[case::flag_7(7, "1234🇺🇸 ")]
        #[case::flag_8(8, "1234🇺🇸  ")]
        #[case::flag_9(9, "1234🇺🇸   ")]
```

joshka:
I've updated the tests so that the document the current behavior even if it's wrong (as noted in the first post).
This is better than panicking, so I'd suggest merge, release, and keep a bug open to fix the more minor issue.

src/text/line.rs
```diff
@@ -586,31 +553,55 @@ impl WidgetRef for Line<'_> {
...
-            }
+    }
+}
+
+fn render_spans(
+    spans: &[Span],
+    area: Rect,
+    buf: &mut Buffer,
+    mut x_offset: u16,
+    mut span_offset: u16,
```
kdheepak, line 580: Is `x_offset` the byte offset and `span_offset` the offset in chars?

src/text/line.rs
```diff
@@ -586,31 +553,55 @@ impl WidgetRef for Line<'_> {
...
+        }
+        let area = Rect {
+            x: area.x.saturating_add(x_offset),
+            width: area.width.saturating_sub(x_offset),
+            ..area
+        };
+        let is_span_partially_visible = span_offset > 0;
+        if is_span_partially_visible {
+            // only render the right side of the partially visible span
+            let (_left, right) = span.split_at_width(span_offset as usize);
```
kdheepak, line 597: I don't quite understand why we only render the right side when a `Span` is partially visible?

kdheepak:
The code is very readable and merging this even with bugs is good to me! I like that there are tests that capture the behavior. 

I don't quite understand why this works (haven't had a lot of time to go through it in detail). Maybe the PR description can have a high level description of why this works? 

Unrelated to this PR but it would nice to collect links to resources on the developer guide sector of website about Unicode that for uninitiated potential contributors to read.

src/text/line.rs
joshka: The other way around - likely a better name would be `buffer_x_offset` perhaps?

src/text/line.rs
joshka: vec![Span::raw("Hello"), Span::raw("World")]
width 4, alignment centered -> "lo" (right side of the first span that is past the offset) then "Wo"
width 7, alignment right -> same

src/text/line.rs
joshka: Rect needs an easy way to increase the x value and decrease the width the same amount, but I'm not sure of a good name for it.

Having that method would make for loops that have to be area constrained, but that write across horizontally easier to write, and the x_offset parameter would then go away and be already part of the area passed into the function.

Something like 
```rust
area.indent_x(span_width);
```

joshka:
> I don't quite understand why this works (haven't had a lot of time to go through it in detail). Maybe the PR description can have a high level description of why this works?

Added an algorithm comment to help in the code

src/text/line.rs
```diff
@@ -586,31 +553,68 @@ impl WidgetRef for Line<'_> {
...
+            span_offset -= span_width;
+            continue;
+        }
+        let area = Rect {
+            x: area.x.saturating_add(buf_x_offset),
+            width: area.width.saturating_sub(buf_x_offset),
+            ..area
+        };
+        let is_span_partially_visible = span_offset > 0;
+        if is_span_partially_visible {
```
EdJoPaTo, line 608: When first reading this I thought "oh, the span is visible, so it should be >=?"
Inverting the `if` might clarify this: `fully_visible = span_offset == 0` → else only partial.

src/text/line.rs
```diff
@@ -586,31 +553,68 @@ impl WidgetRef for Line<'_> {
...
+            // only render the right side of the partially visible span
```
EdJoPaTo, line 609:
```suggestion
            // render the partially visible span starting at the offset -> right side
```

src/text/span.rs
```diff
@@ -332,6 +332,37 @@ impl<'a> Span<'a> {
     pub fn to_right_aligned_line(self) -> Line<'a> {
         self.into_right_aligned_line()
     }
+
+    /// Splits the span into two at the given width.
+    ///
+    /// The first span will contain the characters that fit within the given width, and the second
+    /// span will contain the remaining characters.
+    pub(crate) fn split_at_width(&'a self, width: usize) -> (Self, Self) {
```
EdJoPaTo, line 340: I think this might be a good overgeneralization example. We only need the right side here, but a bunch of stuff is done for the left side too.
This will be called every time a Line or Text is rendered and should not do stuff that's not important here. Maybe its optimized away by the compiler again, but why overcomplicate the code which is not needed in the first place?

src/text/span.rs
```diff
@@ -332,6 +332,37 @@ impl<'a> Span<'a> {
...
+        // There's at least two ways to do this. I chose the second option as it iterates over the
+        // string once at the cost of always allocating two strings regardless of the split point.
+        // 1. iterate the chars to find the split point and then split the spans.
+        // 2. iterate the chars and build the spans at the same time.
+        let mut left = String::new();
+        let mut right = String::new();
```
EdJoPaTo, line 346: Using iterators or `with_capacity` will improve performance. `push` in a loop will create a bunch of allocs which can be prevented with knowing accurate or best effort width. Iterators also allow for a similar optimization without needing to write that.

EdJoPaTo (changes requested):

src/text/line.rs
joshka: I prefer this direction for clarity as it any partial rendered span comes before all the fully rendered ones. Top to bottom reading order helps convey the order of spans.

src/text/span.rs
joshka: Mainly because there's a bug that's affecting users. This is a known good state that passes (most) tests from which we can definitely optimize / rewrite. Happy for this to be ripped out entirely if there's a simpler more obvious way to approach this. My constraints were working code, and it looks like it does the right thing.

src/text/span.rs
joshka: If there's an obvious fix that can be made which keeps this simple and obvious, and passes the tests please make it. (Especially if you can show there's a perf impact that is fixed with a quick benchmark).

src/text/line.rs
EdJoPaTo: Then improve the variable name to make this more clear
```suggestion
        let span_is_only_partially_visible = span_offset > 0;
        if span_is_only_partially_visible {
```

joshka: Actually, I'll change this on second thoughts.

joshka: Actually - I don't like the clippy lint that occurs here (or the allow that it necessitates) - it requires flipping the if / else blocks.

joshka: Suggestion applied.

EdJoPaTo:
Added heavily inspired by unicode-truncate code to find the index in order to get the starting index. This allows for taking a reference rather than cloning into an ever-more-allocating String.

I think this would be a great addition for unicode-truncate especially with all the related test cases specific for this.

Some test cases expect `" c"` with a space in the beginning where the multi width Unicode was.
This requires padding → Owned data → has a bigger performance impact. But it's something for step 2, ideally even solved within unicode-truncate too.

(Mentioning @Aetf here as they might be interested in this discussion)

Did not remove the split_at method which obviously annoys the CI.

Benchmark after the change. (Results on a Raspberry Pi 3 are similar while taking ~10 times the time)

```plaintext
line_render/0           time:   [82.312 ns 82.321 ns 82.329 ns]
                        change: [-13.630% -13.444% -13.279%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 148 outliers among 1000 measurements (14.80%)
  100 (10.00%) low severe
  3 (0.30%) low mild
  16 (1.60%) high mild
  29 (2.90%) high severe
line_render/3           time:   [179.75 ns 179.88 ns 180.00 ns]
                        change: [-26.027% -25.958% -25.891%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 4 outliers among 1000 measurements (0.40%)
  3 (0.30%) high mild
  1 (0.10%) high severe
line_render/4           time:   [202.37 ns 202.48 ns 202.60 ns]
                        change: [-24.577% -24.503% -24.432%] (p = 0.00 < 0.05)
                        Performance has improved.
line_render/6           time:   [262.62 ns 262.75 ns 262.88 ns]
                        change: [-20.892% -20.719% -20.522%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 5 outliers among 1000 measurements (0.50%)
  3 (0.30%) high mild
  2 (0.20%) high severe
line_render/7           time:   [281.92 ns 282.05 ns 282.20 ns]
                        change: [-20.378% -20.302% -20.212%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 25 outliers among 1000 measurements (2.50%)
  19 (1.90%) high mild
  6 (0.60%) high severe
line_render/10          time:   [359.13 ns 359.72 ns 360.49 ns]
                        change: [-13.931% -13.808% -13.652%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 32 outliers among 1000 measurements (3.20%)
  20 (2.00%) low mild
  5 (0.50%) high mild
  7 (0.70%) high severe
line_render/42          time:   [518.82 ns 518.92 ns 519.03 ns]
                        change: [-1.9234% -1.8597% -1.7990%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 6 outliers among 1000 measurements (0.60%)
  1 (0.10%) low mild
  1 (0.10%) high mild
  4 (0.40%) high severe
```

EdJoPaTo:
> Some test cases expect `" c"` with a space in the beginning where the multi width Unicode was.
> This requires padding → Owned data → has a bigger performance impact. But it's something for step 2, ideally even solved within unicode-truncate too.

Oh… it actually doesn't in our use case as we can just increase the `x += n` and continue render further right (`Buffer::set_stringn` does something similar).

src/text/span.rs
```diff
@@ -359,7 +360,29 @@ impl<'a> Span<'a> {
...
+            .map(|(byte_index, char)| (byte_index, char.width().unwrap_or(0)))
+            // Dont include zero width characters at the beginning
+            .filter(|(_, char_width)| *char_width > 0)
+            .scan(0, |sum: &mut usize, (byte_index, char_width)| {
+                *sum = sum.checked_add(char_width)?;
+                Some((byte_index, *sum))
+            })
+            .find(|(_, current_width)| *current_width > skip_width)
+            .map_or_else(|| self.content.len(), |(index, _current_width)| index);
+        // unwrap is safe as it comes from char_indicies
```
joshka, line 379:
```suggestion
        // unwrap is safe as it comes from char_indices
```

src/text/span.rs
joshka: https://github.com/ratatui-org/ratatui/pull/1089/commits/d64f785e8705f7da37f9d392746a914086ea7093

src/text/span.rs
joshka: https://github.com/ratatui-org/ratatui/pull/1089/commits/d64f785e8705f7da37f9d392746a914086ea7093

src/text/span.rs
```diff
@@ -359,7 +360,29 @@ impl<'a> Span<'a> {
         )
     }
 
+    /// Skips the content until the given `skip_width` is skipped to start after the `skip_width`
+    /// while honoring unicode width of characters.
+    #[must_use]
+    pub(crate) fn skip_unicode_width(&'a self, skip_width: usize) -> Self {
```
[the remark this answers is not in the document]
EdJoPaTo: No it cant as the important part missing is the resulting width and the difference between the wanted and the actual width of the trimmed result.
Solved that with b4c2930ea67a1ad0a2258f0248c38202cab16f17

src/text/line.rs
```diff
@@ -604,12 +604,31 @@ fn render_spans(
...
+
+            let limit = span_width - span_offset;
+            let (content, width) = span.content.unicode_truncate_start(usize::from(limit));
+            let width = width as u16;
+
+            let x = limit - width;
+
+            // Empty the cells before as there is a multi width grapheme not rendered before. But as
+            // it is not rendered in the first place (its truncated away) dont reset, set " ".
+            for x in area.x..x {
```
joshka, line 623: Does this work for area.x > 0? (should it be `area.x..area.x+x` perhaps?)

src/text/line.rs
```diff
@@ -1097,16 +1115,14 @@ mod tests {
...
         #[case::center_8_3(8, 3, "🦀c")]
         #[case::center_8_4(8, 4, "🦀cd")]
         #[case::center_8_5(8, 5, "b🦀cd")]
         #[case::center_9_0(9, 0, "")]
         #[case::center_9_1(9, 1, "c")]
-        /// FIXME should be " c" https://github.com/ratatui-org/ratatui/pull/1089#issue-2280234173
-        #[case::center_9_2(9, 2, "🦀")] // should be right side of "🦀c"
-        #[case::center_9_3(9, 3, "🦀c")]
+        #[case::center_9_2(9, 2, " c")] // right side of "🦀c"
+        #[case::center_9_3(9, 3, " cd")]
```
joshka, line 1125: Good catch!

joshka (commented):
I think there might be some general assumed area.x == 0 in here due to code that uses the span_offset. This works for the tests if the line is rendered at x=0 but i think it might break when that's not the case (for the partial rendered span part of the line at least).

src/text/span.rs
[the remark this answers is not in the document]
joshka: nice!

src/text/span.rs
```diff
@@ -359,7 +360,29 @@ impl<'a> Span<'a> {
...
+        let byte_index = self
+            .content
+            .char_indices()
+            .map(|(byte_index, char)| (byte_index, char.width().unwrap_or(0)))
+            // Dont include zero width characters at the beginning
+            .filter(|(_, char_width)| *char_width > 0)
```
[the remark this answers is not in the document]
joshka: I'm ok with breaking it for now, but it was worth noting this in case we do see this actually cause problems.

src/text/line.rs
EdJoPaTo: Wasn't me, was `cargo test` :ghost:

EdJoPaTo:
> I think there might be some general assumed area.x == 0 in here due to code that uses the span_offset. This works for the tests if the line is rendered at x=0 but i think it might break when that's not the case (for the partial rendered span part of the line at least).

Yeah, this fails currently: 0a55e7618d2422377fe5b3704a81bbdb06a26c00

src/text/line.rs
```diff
@@ -1148,18 +1148,38 @@ mod tests {
             assert_buffer_eq!(buf, Buffer::with_lines([expected]));
         }
 
+        /// Ensures the rendering also works away from the 0x0 position.
+        #[rstest]
+        #[case::left_6(Alignment::Left, 6, "a🦀bc ")]
+        #[case::center_6(Alignment::Center, 6, "🦀bc🦀")]
+        #[case::right_6(Alignment::Right, 6, " bc🦀d")]
```
joshka, line 1155: It's probably a little easier to immediately understand these tests by adding some of the padding around (e.g. "XXXXa🦀bc XXXX") (and the assertion becomes simpler too)

src/text/line.rs
```diff
@@ -586,31 +553,87 @@ impl WidgetRef for Line<'_> {
...
+            let limit = span_width - span_offset;
+            let (content, width) = span.content.unicode_truncate_start(usize::from(limit));
+            let width = width as u16;
+
+            let x = limit - width;
+
+            // Empty the cells before as there is a multi width grapheme not rendered before. But as
+            // it is not rendered in the first place (its truncated away) dont reset, set " ".
+            for x in area.x..x {
+                buf.get_mut(x, area.y).set_style(span.style).set_symbol(" ");
```
joshka, line 624: Thinking a bit more about this, I don't think rendering spaces is necessarily the right thing for truncated wide emojis for two reasons:

1. The underlying character of the cell might be a better choice. Consider an example where there is a block, with a border, and we render a title that is "🦀crab" right aligned (assume "/-\" represent the borders)
- 8 => `"/--🦀crab\"`
- 7 => `"/-🦀crab\"`
- 6 => `"/🦀crab\"`
- 5 => `"/ crab\"` wouldn't `/-crab\` be nicer here?
- 4 => `"/crab\"`
- 3 => `"/rab\"`

2. The truncated emoji might be at the end of the first span (left aligned example
- 8 => `"/crab🦀--\"`
- 7 => `"/crab🦀-\"`
- 6 => `"/crab🦀\"`
- 5 => `"/ crab\"` wouldn't `/crab-\` be nicer here?
- 4 => `"/crab\"`
- 3 => `"/cra\"`

EdJoPaTo: We can omit this code and add it later again when there seems to be a real need for this. The important part is aligning it to the right when the truncated width is shorter than the space (due to the emoji that doesn't fit)

src/text/line.rs
EdJoPaTo: I don't think the assertion becomes simpler as the buffer is wider than the truncation width. Truncate to 6 characters and put it somewhere into a 10 width Buffer introduces additional logic.

Creating a Buffer starting at 0x0 and render into 2x2 would likely be simpler, but the diff would be annoying. For this x is only relevant so y could be left at 0.

Side question… why is the Buffer not always based on 0x0 and only takes a size? where is the offset relevant?

joshka: Simpler because the assertion is just the full contents of the buffer. xx123456xx. This makes it a direct comparison against a constant value rather than a comparison with a mutated expected buffer.

Buffer doesn't start at zero for Fixed viewports which can be anywhere on the screen with the same coordinates as screen coordinates.

src/text/line.rs
joshka: I fixed this up in the latest push.

joshka:
All tests passing (and corrected so that the edge cases seem like they're the right result)
E.g. right aligned text that does not have enough space to render the first emoji will not overwrite any first character(s) in that position.

I think this is probably good to go now

src/text/line.rs
```diff
@@ -586,35 +555,86 @@ impl WidgetRef for Line<'_> {
...
+/// TODO: this should probably be move into a pubic method on `Rect` as it's a common operation
+/// that's used in a few places
+const fn indent(area: Rect, x_offset: u16) -> Rect {
+    Rect {
+        x: area.x.saturating_add(x_offset),
+        width: area.width.saturating_sub(x_offset),
+        ..area
+    }
+}
+
```
kdheepak, line 637: Should we add functions called `shift_x` and `shift_y` that take signed integers? This is functionality that I've wanted as well.

joshka: Yes - my two problems with introducing this are:
- what to name it so that the name obviously evokes that we're increasing x, while decreasing the width
- should it be able to work on all sides (left, right, top, bottom)
- should it be a single method or multiple
- counting things points and not updating the number of points I'm making

kdheepak: It might be possible to express it as a single method:

```rust
rect.shift(Corner::TopLeft, (-1, 5))
```

Alternatives:

- `rect.move(::Corner, (x, y))`
- `rect.drag(::Corner, (x, y))` (I'm picturing click and drag the corner, although this might not be intuitive)

joshka: I don't like either name
- shift implies moving without changing the size
- move is a reserved keyword
- drag is a bad name because it implies an event that we're not actually handliing

"Indent" is the best I have for now, or perhaps some other verb that means to modify / change the size in some way.

joshka: Either way - not something to fix in this PR - this method is private and useful here. We can iterate without breaking on what the public version of that looks like.

EdJoPaTo: Rust has a [`AddAssign`](https://doc.rust-lang.org/std/ops/trait.AddAssign.html) trait which has the `add_assign` method. (or `sub_assign`, `mul_assign` and so on)
Doesnt seem to improve the naming…

`indent_x` would be a good name for it for my taste. Not sure whether there needs to be indent from the right. `indent_y` might be neat. But yeah, not part of this PR to stabilize this.

src/text/line.rs
```diff
@@ -586,35 +555,86 @@ impl WidgetRef for Line<'_> {
...
+            area = indent(area, span_width);
         }
     }
 }
 
+/// indents a given `area`'s x value by the given `x_offset`, returning a new `Rect`
+///
+/// TODO: this should probably be move into a pubic method on `Rect` as it's a common operation
+/// that's used in a few places
+const fn indent(area: Rect, x_offset: u16) -> Rect {
```
EdJoPaTo, line 630: While this shouldn't be public / stabilized yet for the reasons in the other comment I would still move this to the Rect and add `#[must_use]` to ensure its not accidentally misused.

In the rect file it would be `pub(crate)` but it seems way more sorted than having this somewhere in the line file.

src/text/line.rs
```diff
@@ -586,35 +555,86 @@ impl WidgetRef for Line<'_> {
...
+        if area.is_empty() {
+            break;
+        }
+        if span_offset > 0 {
+            // the first span is only partially visible, so truncate the start and render it
+            // starting at span offset
+            let render_width = span_width - span_offset;
+            let (content, display_width) =
+                span.content.unicode_truncate_start(render_width as usize);
+            let display_width = display_width as u16;
```
EdJoPaTo, line 608: The width naming is horrible. I have no clue what the variables mean without reading the code what is assigned and when.

Before this change they were named `limit` and `width`. I still like `limit` as its way more clear what it means. `width` is definitely debatable.

Other ideas:
- `render_width` → `target_render_width` or `available_(for_span_)width`, maybe also the name of the input of the function: `max_width`
- `display_width` → `actual_width` or `actually_rendered_width`

src/text/line.rs
```diff
@@ -586,35 +555,86 @@ impl WidgetRef for Line<'_> {
...
+        }
+        if area.is_empty() {
+            break;
+        }
+        if span_offset > 0 {
+            // the first span is only partially visible, so truncate the start and render it
+            // starting at span offset
+            let render_width = span_width - span_offset;
+            let (content, display_width) =
+                span.content.unicode_truncate_start(render_width as usize);
```
EdJoPaTo, line 607: `as` can silently fail / apply different meaning on refactorings. Using `usize::from` ensures that there is actually an implementation for that change. `u64 as usize` works on 64 bit machines but truncates on 32 bit ones which won't be noticed with `as`.

Using an approach that is sure to work even when it might never happen in this case is better code style in my opinion. Don't train yourself into using the pitfall normally.

```suggestion
                span.content.unicode_truncate_start(usize::from(render_width));
```

src/text/line.rs
```diff
@@ -586,35 +555,86 @@ impl WidgetRef for Line<'_> {
...
+        if span_offset >= span_width {
+            span_offset -= span_width;
+            continue;
+        }
+        if area.is_empty() {
+            break;
+        }
+        if span_offset > 0 {
+            // the first span is only partially visible, so truncate the start and render it
+            // starting at span offset
```
EdJoPaTo, line 604: It's not necessarily the first of them all. It's the first rendered one. The ones before might be skipped due to `offset >= span_width`. So the wording might be misleading here.

```suggestion
            // render the partially visible span, so truncate the start and render it starting at span offset. As this one is partially visible all the following ones are fully visible.
```

(`cargo fmt` the line width)

src/text/line.rs
```diff
@@ -586,35 +555,86 @@ impl WidgetRef for Line<'_> {
...
+            let render_width = span_width - span_offset;
+            let (content, display_width) =
+                span.content.unicode_truncate_start(render_width as usize);
+            let display_width = display_width as u16;
+
+            // if the truncation has truncated an initial grapheme, then we need to start rendering
+            // from a position that takes that into account by indenting the start of the area
+            area = indent(area, render_width - display_width);
+            let right = Span::styled(content, span.style);
```
EdJoPaTo, line 613: As the indentation math moved down here the check of emptiness should move down here too
```suggestion
            area = indent(area, render_width - display_width);
            if area.is_empty() {
                break;
            }
            let right = Span::styled(content, span.style);
```

src/text/line.rs
```diff
@@ -586,35 +555,86 @@ impl WidgetRef for Line<'_> {
...
+    for span in spans {
+        let span_width = span.width() as u16;
+        // ignore spans that are completely before the offset
+        if span_offset >= span_width {
+            span_offset -= span_width;
+            continue;
+        }
+        if area.is_empty() {
+            break;
+        }
```
EdJoPaTo, line 601: As the area math moved down move the check down too. (review comment before this one)

```suggestion
```

EdJoPaTo (changes requested):

src/text/line.rs
joshka: done in a28718b19

src/text/line.rs
joshka: Fixed in bc32964e5

src/text/line.rs
joshka: `u16 as usize` can never fail. This is consistent with the rest of ratatui. I the refactoring pain, but we have that already throughout. usize::from adds no value here.

src/text/line.rs
joshka: bd8f8cf17

src/text/line.rs
joshka: 7c6037bcf

src/text/line.rs
joshka: bd8f8cf17

src/text/line.rs
joshka: bd8f8cf17

joshka:
There's another possible simplification (for future consideration on this) that teases apart the span selection and span rendering. Something like (pseudo code):

```
fn render_ref(...) {
  for span in self.visible_spans(...) {
    span.render_ref(area, buf)
    let area = area.indent(span.width())
    if area.is_empty() { break }
  }
}

fn visible_spans(...) -> Iterator<&Span> {
  skip non-visible spans
  chain!(truncated first span, rest of spans)
}
```

I experimented briefly with this, but I think it's only possible if visible_spans returns owned Spans instead of refs due to the truncation - it's worth another try (in a future PR).

src/text/line.rs
```diff
@@ -586,31 +555,75 @@ impl WidgetRef for Line<'_> {
...
             let offset = match self.alignment {
-                Some(Alignment::Center) => (area.width.saturating_sub(width)) / 2,
-                Some(Alignment::Right) => area.width.saturating_sub(width),
+                Some(Alignment::Center) => (line_width.saturating_sub(area.width)) / 2,
+                Some(Alignment::Right) => line_width.saturating_sub(area.width),
                 Some(Alignment::Left) | None => 0,
             };
-            x = x.saturating_add(offset);
-            self.to_owned()
+            render_spans(&self.spans, area, buf, offset);
```
kdheepak, line 573: When `line_width` is less than `area.width`, offset is only used to "indent" the area rect, but when `line_width` is greater than `area.width`, offset is the "span offset".

I think it may make sense to break this down into two parts? 

```rust
let line_width = self.width() as u16;
let has_enough_room = line_width <= area.width;
let offset = match (has_enough_room, alignment) {
    (true, Some(Alignment::Center)) => (area.width.saturating_sub(line_width)) / 2,
    (true, Some(Alignment::Right)) => area.width.saturating_sub(line_width),
    (false, Some(Alignment::Center)) => (line_width.saturating_sub(area.width)) / 2,
    (false, Some(Alignment::Right)) => line_width.saturating_sub(area.width),
    (_, Some(Alignment::Left) | None) => 0,
}
if has_enough_room {
    // we have enough room to render the entire line
    // shift the x coordinate of the rect and render all the spans 
    // starting from the 0th offset into the new inner area
    let inner_area = area.indent_x(offset);
    render_spans(&self.spans, inner_area, buf, 0)
} else {
    // we don't have enough room, and have to truncate the spans based on the starting offset
    render_spans(&self.spans, area, buf, offset)
}
```

EdJoPaTo: I thought about this, but I think the if else is more clear for this here? Putting both into match only creates more branch points… So I think it's both easier to understand in the if and probably more performant? But the latter is pure speculation as LLVM optimized so much stuff.

src/text/line.rs
```diff
@@ -586,31 +555,75 @@ impl WidgetRef for Line<'_> {
...
+            area = area.indent_x(available_width - actual_width);
+            if area.is_empty() {
+                break;
+            }
+            let right = Span::styled(content, span.style);
+            right.render_ref(area, buf);
+
+            area = area.indent_x(actual_width);
+            span_offset = 0; // ensure that the next span is rendered in full
+        } else {
```
kdheepak, line 620: Having `span_offset` be `mut` made it hard (for me personally) to understand but I think I get what is happening here now.

Here's some of my comments.

1. the name of the variable is `span_offset` but it really is some kind of width. I think of an `offset` as a point on a line, but here in every calculation you can see the other terms are all "width". fwiw, whenever I read `span_offset` I parsed it as "available_total_span_area_for_rendering".

2. the meat of this function is whether or not to render a partial span. After reading through it, I pictured the code as 3 parts. 
    
a) Skip everything before `span_offset`

b) 
- If the current span is partially renderable, truncate the span
- If the current span is fully renderable, render it

c) Render all the remaining spans

Does it make sense to move the second part into a separate function like `render_maybe_partial_span`?

Also, (I don't know it is possible) I was wondering if it were possible to present the code like so:

```rust
    let mut spans_iter = spans.iter().skip_while(|span| {
       // skip till span_offset
    });

    if let Some(maybe_partial_span) = spans_iter.next() {
       render_maybe_partial_span(...)
    }

    for span in spans_iter {
        // render all the remaining
    }
```

It would have been easier to track the 3 parts for me that way.

All that said, the code + functionality looks good to me overall and happy to merge it to fix this issue.

EdJoPaTo: > 1. the name of the variable is `span_offset` but it really is some kind of width. I think of an "offset" as a point on a line, but here in every calculation you can see the other terms are all "width". fwiw, while reading this whenever I came across `span_offset`, I parsed it as `available_total_span_area_for_rendering`.

The offset is the starting point within the line. So `total_span_width - actually_possible_to_display_here = offset`. The offset is definitely not a width. Its the about to be skipped to end up at the width that can be displayed.

---

> Also, (I don't know it is possible) I was wondering if it were possible to present the code like so:

The first iter might be possible relatively easy.

The second and third are kinda hard as the second is involving a condition. Partially or not.

I read the current code kinda like that already. First part to skip everything until we are actually at our offset.

What might simplify this code logic is to reverse the spans and go from right to left. That will be writable, until it isnt. Then abort. The last one might be partial.
But that results in interesting logic to get the correct x. The current code increases the x while the reversed one would reduce it.

I would think of this as an improvement in a later PR. It might be more performance efficient for many spans but I dont know. But we have the tests for many of these cases now so improving this code in the future will be way easier.

src/text/line.rs
```diff
@@ -586,31 +555,75 @@ impl WidgetRef for Line<'_> {
...
+/// because that would have made the above code just a little uglier
+///
+/// Algorithm:
+/// 1. For each span in the line:
+///    1. If the span is completely before the `span_offset`, skip it
+///    2. If the span is partially before the `span_offset`, render the right side of the span
+///    3. If the span is completely after the `span_offset`, render the whole span
+/// 2. Indent `area` to account for the width of the rendered span
+/// 3. Update `span_offset` to 0 after the first span is rendered in full
+/// 4. Repeat
```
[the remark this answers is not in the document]
EdJoPaTo: Thats fixed and tested now: https://github.com/ratatui-org/ratatui/blob/bd8f8cf17d5346182ed2c68e84acae7bfc9f42b1/src/text/line.rs#L1075

Cargo.toml
```diff
@@ -43,6 +43,7 @@ document-features = { version = "0.2.7", optional = true }
 lru = "0.12.0"
 stability = "0.2.0"
 compact_str = "0.7.1"
+unicode-truncate = "1"
```
EdJoPaTo, line 46: We should probably sort these or something… currently its somewhat of a mess, historically layered or something…

src/text/span.rs
```diff
@@ -332,6 +332,12 @@ impl<'a> Span<'a> {
     pub fn to_right_aligned_line(self) -> Line<'a> {
         self.into_right_aligned_line()
     }
+
+    /// Returns `true` if the span is empty.
+    #[must_use]
+    pub fn is_empty(&self) -> bool {
```
EdJoPaTo, line 338: Is this actually used? I think this was used on a previous iteration and could be removed again / introduced in its own PR.

src/text/line.rs
```diff
@@ -586,31 +555,75 @@ impl WidgetRef for Line<'_> {
...
+            continue;
+        }
+        if span_offset > 0 {
+            // the first visible span is only partially visible, so truncate the start and render it
+            // starting at span offset, all subsequent visible spans are rendered in full (as
+            // available space permits)
+            let available_width = span_width - span_offset;
+            let (content, actual_width) = span
+                .content
+                .unicode_truncate_start(available_width as usize);
```
kdheepak, line 606: Why do you use `.unicode_truncate_start` here? The docs say:

> Truncates a string to be at most width in terms of display width by removing the start characters.

Shouldn't we use `.unicode_truncate` and `.unicode_truncate_start` depending on the `Line`'s alignment?

src/text/line.rs
kdheepak: > The offset is the starting point within the line.

Having the "starting point within the line" change in the algorithm (i.e. be `mut`) made it difficult to read for me. I get that it is "the point that represents the summation of widths that are to be skipped". I guess I was wondering if there was a better way to name it?

src/text/line.rs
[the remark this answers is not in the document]
kdheepak: Ah, I missed that. PR description need to be updated?

src/text/line.rs
EdJoPaTo: The right truncation at end end happens when the area ends all by itself. That’s way Alignment::Left starts at 0 and at some point area.is_empty().

Center alignment is similar. Find the starting point and render until area.is_empty().

Basically: truncation on the right side happens due to the limited area. Truncation on the left is calculated and then done by the truncation crate.

src/text/line.rs
joshka: Added comments and modified variable names to indent_width and skip_width in 1d15ef6b

src/text/line.rs
joshka: addressed.

src/text/line.rs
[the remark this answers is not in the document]
joshka: Updated the PR comment to clarify this.

src/text/line.rs
joshka: It's kinda a "width" of the spans to skip... I think `span_skip_width` might make sense as a name.

I can see how offset could a bad name for this, as offset often carries an implication that something is being moved, but here it's referring to how much to move an implicit skip cursor in the spans.

```rust
    let mut spans_iter = spans.iter().skip_while(|span| {
       // skip till span_offset
    });
```
I actually attempted this at one point, but it's difficult to satisfy rust's borrowing rules. To skip the spans, you need to keep a variable that is updated in the closure, but you can't use that variable after the closure as it's moved within the closure. You need that remaining width in order to know the amount to truncate the first span by.

```
error[E0503]: cannot use `span_skip_width` because it was mutably borrowed
   --> src/text/line.rs:606:27
    |
598 |     let spans = spans.iter().skip_while(|span| {
    |                                         ------ `span_skip_width` is borrowed here
599 |         if span_skip_width >= span.width() as u16 {
600 |             span_skip_width -= span.width() as u16;
    |             --------------- borrow occurs due to use of `span_skip_width` in closure
...
606 |     let span_skip_width = dbg!(span_skip_width);
    |                           ^^^^^^^^^^^^^^^^^^^^^ use of borrowed `span_skip_width`
607 |     for span in spans {
    |                 ----- borrow later used here
    |
    = note: this error originates in the macro `dbg` (in Nightly builds, run with -Z macro-backtrace for more info)
```

> The first iter might be possible relatively easy.

Famous last words of a Rust programmer

We could write this first part as a while loop, but it ends up looking pretty much like the for loop. You have to special case the span that you got from `span.next()` when there is span is partially visible, so to do it outside of the loop means either putting the span back into the iterator (i.e. using peekable to not pull each element out, or using [put_back](https://docs.rs/itertools/latest/itertools/fn.put_back.html)), or storing it in a option variable at the end of the loop and checking it. Either way, the added complexity of that code is bad. A better factoring of this logic is to instead do the simplification mentioned in https://github.com/ratatui-org/ratatui/pull/1089#issuecomment-2102340388 (make an method that returns an iterator over the visible spans, including the first partial span).

Rename and comments added in 2fa465c2

Cargo.toml
joshka: 5cc0072f

src/text/span.rs
joshka: Revoved in 65842dd1

src/text/line.rs
joshka: Clarified with a comment in c08cd0c4

joshka: clarified with a comment in c08cd0c4

src/text/span.rs
joshka: removed in 65842dd15

Cargo.toml
joshka: Fixed in 5cc0072f6

src/text/line.rs
```diff
@@ -586,31 +555,84 @@ impl WidgetRef for Line<'_> {
...
+        // visible
+        if span_skip_width >= span_width {
+            span_skip_width -= span_width;
+            continue;
+        }
+        if span_skip_width > 0 {
+            // the first visible span is only partially visible, so truncate the start and render it
+            // starting at span_skip_width, all subsequent visible spans are rendered in full (as
+            // available space permits). We only need to truncate the start of the first visible span
+            // because the rest of the line will be truncated automatically by the area width
```
EdJoPaTo, line 611: This is explained multiple times now.

EdJoPaTo (commented):
I really like the amount of added tests. As this has to deal with Unicode I still think there could be way more test cases due to the complexity of Unicode but it improves the situation a lot.

I want to take a look at the documentation comments but that’s easier on a bigger screen so that’s something for tomorrow me. I just noticed something being duplicate and having a lot of comments which might be able to be boiled down especially with better variable names to not need as much text (when possible).

src/text/line.rs
joshka: I generally agree that good variable names are better than comments and in my earlier years I would have probably died on the hill proclaiming an opinion that "comments are unnecessary if you write good code". Here though there's lots of ways to misinterpret the meaning of each of the variables due to the inherent complexity of what we need to do. The comments in this change serve as an apology of sorts for not being able to make the algorithm perfectly clear with just variable names. The increased quantity and verbosity of comments in an area like this also highlights the amount of intent that has gone into choosing to do things a particular way, and that needs to be understood before any refactoring / changes are made to this code.

You're right that the area truncation is explained above in the render_ref function, and that is a duplicate of this comment, but given that @kdheepak asked about this in the PR, it's reasonable to assume that someone of lesser knowledge about Ratatui who wants to edit this code in the future might have the same questions (and may not look at the calling code for answers). So I think explaining this twice doesn't do too much harm here.

kdheepak (approved):
I like that there are lots of comments and documentation, and the code + variable names + tests should be enough for someone in the future to refactor if required.

joshka:
I'll wait on @EdJoPaTo's approval to merge this

src/text/line.rs
EdJoPaTo: Im not against the comments explaining what is going on. Just that there is duplicated stuff now and in this case the one above is more clear to me.
There shouldn’t be duplicate explanations for the same thing.

when stuff is refactored the documentations also needs to be adapted and personally I would adapt the first thing I see describing the now changed behaviour. Another spot might be completely missed. Reviews should find something like this but it might be out of view for the changes so it might be missed.

so my general approach would be to split into good variable names and function names, then comments and doc comments. The function comment should explain the general idea, the comments on code should explain specifics why things are done the way they are done.

comments on tests explain why they are relevant to be tested and shouldn’t explain the behaviour of the function so they are out of scope in order to understand the function.

yes, @kdheepak had questions, so others might be too, but I would try to prefer pointing to the relevant documentation rather than duplicating the same comment.

(will look into it on a bigger screen)

EdJoPaTo: What triggered me here was the "rest" of the line instead of the right of the line. Then I found the same comment with "right" above.

EdJoPaTo:
The logic is somewhat horribly complex as a lot of stuff isn't intuitive… The widths are usize and not u16 as the truncation of the end is implicitly done. That resulted in misunderstandings in what should happen. I don't think we can fully explain the code by comments. We need many tests to ensure it actually does what it is expected to do. Lines/Spans longer than u16 for example is a whole new mess which I haven't thought about before.

Also, I refactored the code to iterator logic. I'm fine with reverting it. Benchmarks suggest it's similar in performance. But it might be even harder to follow what is going on?

EdJoPaTo (approved):
It somewhat worries me that new test cases pop up and break stuff even when we thought we are done and only need documentation stuff. But it's way better than the current state of random panics.

The approach of debug_asserts is somewhat I like here as it ensures that assumptions are actually true and documents these assertions in the process. And they only panic on debug, so production code will only be displaying stuff wrong, not panicking. Also, we have many test cases and users of ratatui will provide more when they run their stuff as dev builds.

So… documentation won't be ever perfect, and I kinda expect new cases that break the current code as Unicode is Unicode. So whoever wants to improve this should have many test cases available. Which is what we achieved here for sure.

So I'm in for merging this in order to have a better state than before. We won't reach a perfect state anyway.

src/text/line.rs
```diff
@@ -586,32 +556,100 @@ impl WidgetRef for Line<'_> {
...
+                );
+                let offset_of_initial_grapheme = offset_of_initial_grapheme as u16;
+
+                (
+                    content,
+                    span.style,
+                    actual_width,
+                    offset_of_initial_grapheme,
+                )
+            } else {
```
kdheepak, line 635: This looks a lot better to me! Nice find!

The `if` block still executes just once, right? If so, can we add a comment.

EdJoPaTo: Should be. (I don't think we can include a debug_assertion for that one in a neat and simple way?)

But that depends on the `available_width` math from above which is further away. That should be stated in that comment then too.

kdheepak: Maybe you can add `.enumerate()` before the `.map()` and `debug_assert!(i == 0)`? LLVM might optimize it away if it sees it is not being used anyway.

joshka:
Some final tweaks:

It's important to look the absolute magnitude of a perf gain, and not just the relative amount. On my M2 mac, removing this costs ~30ns. This which means for an app running at an absolutely absurd 1000FPS, the app would now run at 999.97 FPS. This isn't worth keeping the code to optimize.

- Remove perf fast path for left aligned lines as the absolute perf diff
  is negligible (30ns on an M2 Mac)
- move the visible spans calculation to a method that returns an
  iterator over the spans that are visible
- replace truncation clippy lints with converions that clamp to u16::MAX
- replace arithmetic_side_effects clippy lint with saturating_sub
- remove debug_asserts
- rewrite comments in the precise imperative mood ("when this, then that", rather than "if this then it's necessary / we only need to")

src/text/line.rs
joshka: This is unlikely to ever be a hotspot

joshka:
I'd like to get this released (perhaps Monday if @orhun is up to it).

Direct link to the current file as of the last commit: https://github.com/ratatui-org/ratatui/blob/43118c5494f5218332319a41eb742a51138d7e3c/src/text/line.rs#L555

The actual complexity of the implementation is pretty much aligned with the inherent complexity of the problem, and it's documented well enough.

EdJoPaTo:
> I'm not too worried - most text that's rendered in the context of a tui app isn't > u16::MAX, and when it is the resultant bug is easy to fix by pre-trimming the text instead of doing this in the rendering.

I am not worried about the actual bug that came up. I am more worried that 3 people looked over it for days and missed yet another bug.

> I replaced these with infallible code.

What I liked about it were the assumptions actually tested. We wrote code with assumptions which are now hidden behind infallible code. But with all the tests, fine by me.

> Remove perf fast path for left aligned lines as the absolute perf diff is negligible (30ns on an M2 Mac)

I am not at all worried about way overpowered hardware. I know that `mqttui` is used on Raspberry Pi, personally I regularly use `mqttui` on Raspberry Pi 1.

Line is one of the core constructs used by a lot of things. Relatively small improvements will impact a lot of code depending on it. So it's way more significant to improve the performance here. And also the `Alignment::Left` is the default for most things so it has the most impact.

src/text/line.rs
```diff
@@ -585,36 +555,96 @@ impl Widget for Line<'_> {
...
+            // when rendering the span to the buffer.
+            let available_width = span_width.saturating_sub(span_skip_width);
+            span_skip_width = 0; // ensure the next span is rendered in full
+            Some((span, span_width, available_width))
+        })
+        .map(|(span, span_width, available_width)| {
+            if span_width <= available_width {
+                // Span is fully visible. Clone here is fast as the underlying content is `Cow`.
+                return (span.clone(), span_width, 0u16);
+            }
```
EdJoPaTo, line 631: @joshka You were against this as it changes the order execution and the code.

Also return then else case could be just the else and omit the `return`.

src/text/line.rs
```diff
@@ -585,36 +555,96 @@ impl Widget for Line<'_> {
...
+        let span_width = u16::try_from(span_width).unwrap_or(u16::MAX);
+        area = area.indent_x(span_width);
     }
 }
 
+/// Returns an iterator over the spans that are visible
+///
+/// Visibility is determined by the spans that are partially or completely visible taking into
+/// account just the amount of width to skip from the start of the `Line`.
+fn visible_spans<'a>(
```
EdJoPaTo, line 605: I don't think moving this into its own method is very useful. Its very specific to the render_span method.

Also, this was a pitfall for me in the past, this is not visibility. This is only skipping from the left. The right is still untouched and will only be trucated at the actual render. Visible would mean that everything can be u16, but that is not the case here. This should not be named `visible` as a function name is what is read first. It gives a very wrong impression about what this function is and for what it might be used in other cases.

Also, the doc comment about `partially or completely visible` is somewhat fine for code inside `render_spans` but very wrong for its own method with its own assumptions.

When this method is named maybe `skip_spans_by_width` it would be better, but it should not be named visible. Personally I would prefer it inside `render_spans` again as its very specific to the logic of `render_spans`. Otherwise state in the doc comments that this is very dependent on the assumptions of `Widget::render_ref` and move `render_spans` back into `render_ref`.

```suggestion
/// Returns an iterator over the spans that skips are visible
///
/// Visibility is determined by the spans that are partially or completely visible taking into
/// account just the amount of width to skip from the start of the `Line`.
fn skip_spans_by_width<'a>(
```

src/text/line.rs
```diff
@@ -585,36 +555,96 @@ impl Widget for Line<'_> {
 impl WidgetRef for Line<'_> {
     fn render_ref(&self, area: Rect, buf: &mut Buffer) {
         let area = area.intersection(buf.area);
+        let line_width = self.width();
+        if area.is_empty() || line_width == 0 {
+            return;
+        }
```
EdJoPaTo, line 561: `area.is_empty` is `O(1)` while `.width()` is `O(content.len())`. In other words: The written logic here still uses the expensive `line_width` even when the area is empty.
LLVM does not notice this, and I would prefer to explicitly write it like it was before.

Benchmark on Raspberry Pi 2:

```plaintext
line_render/Left/0      time:   [532.39 ns 532.45 ns 532.52 ns]
                        change: [+681.68% +683.74% +684.94%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 62 outliers among 1000 measurements (6.20%)
  30 (3.00%) high mild
  32 (3.20%) high severe
```

EdJoPaTo (changes requested):

EdJoPaTo:
To pick even more on the performance fast path of `Alignment::Left` on slower devices…

Let's imagine a table rendered on a `100x100` terminal. Assume 5 columns and 90 rows. The table cell used `Text`, so many `Line`s. More cells vs. more lines per cells are probably irrelevant.

`90 * 5` means 450 visible Cells to be rendered. The worst case time in the benchmark below is the Buffer wider than the Line width with `9.7 µs`. Let's assume `10 µs` for ease of calculation.
`10 µs * 450 = 4.5 ms`. 60 FPS are `16.66 ms`. So we took 1/4 of that for the rendering of lines alone.

And this is a Raspberry Pi 2 benchmark, I use `mqttui` on Raspberry Pi 1 regularly.

Even an improvement of 12% is really good here already. The benchmark goes to a regression of 36%. So this is definitely significant. For the default of left alignment which is the most often used one.

Benchmark on Raspberry Pi 2:

```plaintext
line_render/Left/3      time:   [2.7422 µs 2.7427 µs 2.7433 µs]
                        change: [+34.945% +36.396% +37.329%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 68 outliers among 1000 measurements (6.80%)
  48 (4.80%) high mild
  20 (2.00%) high severe
line_render/Left/4      time:   [2.9284 µs 2.9290 µs 2.9297 µs]
                        change: [+33.260% +33.549% +33.885%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 23 outliers among 1000 measurements (2.30%)
  13 (1.30%) high mild
  10 (1.00%) high severe
line_render/Left/6      time:   [4.7153 µs 4.7158 µs 4.7164 µs]
                        change: [+23.409% +24.391% +24.993%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 63 outliers among 1000 measurements (6.30%)
  30 (3.00%) high mild
  33 (3.30%) high severe
line_render/Left/7      time:   [4.9508 µs 4.9513 µs 4.9518 µs]
                        change: [+23.962% +24.193% +24.427%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 44 outliers among 1000 measurements (4.40%)
  24 (2.40%) high mild
  20 (2.00%) high severe
line_render/Left/10     time:   [7.0004 µs 7.0016 µs 7.0030 µs]
                        change: [+16.254% +16.648% +16.980%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 46 outliers among 1000 measurements (4.60%)
  21 (2.10%) high mild
  25 (2.50%) high severe
line_render/Left/42     time:   [9.7193 µs 9.7207 µs 9.7224 µs]
                        change: [+12.279% +12.559% +12.887%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 131 outliers among 1000 measurements (13.10%)
  57 (5.70%) low mild
  35 (3.50%) high mild
  39 (3.90%) high severe
```

Benchmark of Raspberry Pi 4: regression of 14.3 to 38.7%. Worst case takes `~2.03 µs`, so ~5 times faster than Pi 2.

I will not benchmark this on Pi 1 as running the benchmarks on Pi 4 take ~10 min, on Pi 2 ~40 min. Pi 1 is single core in comparison to Pi 2 with 4 cores.

src/text/line.rs
joshka: Fixed in 10815513

src/text/line.rs
joshka: My heuristic for writing readable code is that a simple guard clause is usually a simplification worth pursuing when it removes a layer of nesting like this (even at the expense of the order that lines will be encountered like this). If this was more than 1 line like this I'd still want it to be handled as an `if { } else { }`

More specifically: layers of nesting and the else statements here cost more to maintain. This is due to the cognitive burden when reading of carrying both branches in your head for the entire block of code.

joshka:
> 10 µs * 450 = 4.5 ms. 60 FPS are 16.66 ms

Let's look at the absolute magnitude of the change though: (10 * .125) = 1.25us per call

1.25*450 = 0.563ms, so each frame now takes (16.667+0.563) = 17.23ms = 58 FPS. That's measurable sure, but rarely directly noticeable. Assuming a factor of 4 on the Pi 1, 18.92ms = 53 FPS.

One thing I did note when removing the fast path was that the slow paths got slightly faster (1-4%), so the tradeoff is one that doesn't necessarily have a definitively good answer.

For now, I'd like to keep it simplified, and keep this PR about fixing the bug and not trying to eek out all the performance of this. You've convinced me that there are some parts of this that matter to you, so let's continue the discussion about perf in an issue / forum topic if it's more a discussion.

> Pi 1 is single core in comparison to Pi 2 with 4 cores.

This raises an interesting point. I wonder (not for this PR), what parts of our rendering pipeline might be able to be done more in parallel.

Issue #201, opened 2023-05-29
Line: Allow `Line`s to be individually truncated/wrapped


TimerErTim:
## Idea
Similar to #149 , one could also individually wrap different `Line`s inside the same paragraph. These `Line`s would default to the `LineComposer` used for the paragraph but can override that to use their own.

### Use cases

The uses cases would be the same as #149. People could use this to have differently formatted lines (truncated, word wrapped, character wrapped, etc.) inside the same paragraph with some specific scrolling and styling. 

## Feedback

This is just an idea and probably not worth the effort if there is no demand by the user base. So we would like to have some feedback from you as both the project maintainers/contributors and users. Is this feature even desired?

kxxt:
I need this feature so I tried to implement it in a WIP PR: #1091. Early feedback is welcome, especially on how individual `Line`s' wrap interact with the Paragraph's wrap.

joshka:
(As mentioned in the PR). We're currently rewriting line rendering in #1091 to fix #1032. Implementing this also has to take into account #293. We have a constraint that this should not rely on the reflow module for an implementation (as the reflow module is painful to maintain / modify further).

joshka:
Closing out for now to reduce the backlog to issues that we have time to work on.
Feel free to reopen if there's more to add on this.

(Mostly this would be done as part of #293), so doesn't need to be tracked individually.

PR #1091, opened 2024-05-09
feat: allow `Line`s to be wrapped individually wrapped

kxxt:
I am implementing #201 because I need this feature.

This PR is still a WIP as I haven't added tests. Early feedback is always welcome!

How the individual line wrapping should interact with the paragraph's wrapping is a big headache. For now this PR only handles wrapping for lines without considering the interaction with paragraphs, which means that when rendering paragraph, `Line's' wrap is ignored.

It seems I also need to update `Text` to make it work in the `List`.

### DIFF
--- a/src/text.rs
+++ b/src/text.rs
@@ -47,7 +47,7 @@
 mod grapheme;
 pub use grapheme::StyledGrapheme;
 
-mod line;
+pub(crate) mod line;
 pub use line::Line;
 
 mod masked;
--- a/src/text/line.rs
+++ b/src/text/line.rs
@@ -1,8 +1,28 @@
 #![deny(missing_docs)]
 use std::{borrow::Cow, fmt};
 
+use unicode_width::UnicodeWidthStr;
+
 use super::StyledGrapheme;
-use crate::prelude::*;
+use crate::{
+    prelude::*,
+    widgets::{
+        reflow::{LineComposer, WordWrapper, WrappedLine},
+        Wrap,
+    },
+};
+
+pub(crate) const fn get_line_offset(
+    line_width: u16,
+    text_area_width: u16,
+    alignment: Alignment,
+) -> u16 {
+    match alignment {
+        Alignment::Center => (text_area_width / 2).saturating_sub(line_width / 2),
+        Alignment::Right => text_area_width.saturating_sub(line_width),
+        Alignment::Left => 0,
+    }
+}
 
 /// A line of text, consisting of one or more [`Span`]s.
 ///
@@ -154,6 +174,9 @@ pub struct Line<'a> {
 
     /// The alignment of this line of text.
     pub alignment: Option<Alignment>,
+
+    /// How to wrap the text
+    pub wrap: Option<Wrap>,
 }
 
 impl<'a> Line<'a> {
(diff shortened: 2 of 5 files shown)


joshka:
As a heads up, you might want to digest https://github.com/ratatui-org/ratatui/issues/293 for context before tackling just a part of this.

We're effectively rewriting `Line::render_ref` in #1089 to fix the unicode truncation bug #1032, so you're aiming at a moving target right now and will have to rewrite your changes with that in mind.

My biggest concern about introducing a change like this is that it's important to do it in a way that doesn't break users of the new API many times as the new functionality is introduced. This means gating things behind unstable flags and if there are breaking changes, doing them in a way that gives many several major versions worth of compiler deprecation warnings.

Ideally, we'd implement something fully new here, and fully deprecate the reflow module. That likely means a new WrapOptions type rather than using the existing type. It also means that we shouldn't make the new functionality dependent in any way on the reflow module.

Lastly, taking a brief look at the implementation suggests that it adds significant complexity to the render method (29 Loc -> 64 LoC). From an implementation perspective, it's important that these particular methods are simple, obvious, maintainable, with easily reasoned about structure and performance. The number of levels of indentation is usually a good indicator of this. The PR's structure changes things to the following nesting:

```rust
fn ...
    if ...
        while ...
            for ...
                if ...
  else
      if ...
      else ...
      for ...
```

I wouldn't review this PR without simplifying it significantly. Finding the right simplifications on this particular method is a pretty hard thing to do even for us core devs. Take a read of the various iterations of #1089 for some insight on this.

kxxt:
@joshka Thanks for your feedback! As I dig deeper I find that it's much more complex to make it work because I also need to modify ListItem and Text. Given that I need this feature now and it will take weeks(or even monthes) for me to finish this PR and get it reviewed and merged. I think it makes more sense for me to create a tailored widget that suites my own need rather than getting it merged into ratatui. So I am closing this PR for now.

joshka:
> I think it makes more sense for me to create a tailored widget

I'm not sure what the underlying problem you're solving is, but reading between the lines (no pun intended), it sounds like it's probably that you have a non wrappable part and a wrappable part in a list. Perhaps using https://crates.io/crates/tui-widget-list and having a small widget that has a non-wrappable line and a wrapped paragraph (or pre-wrapped text using textwrap) could be an approach that works for you.

kxxt:
> > I think it makes more sense for me to create a tailored widget
> 
> I'm not sure what the underlying problem you're solving is, but reading between the lines (no pun intended), it sounds like it's probably that you have a non wrappable part and a wrappable part in a list. Perhaps using https://crates.io/crates/tui-widget-list and having a small widget that has a non-wrappable line and a wrapped paragraph (or pre-wrapped text using textwrap) could be an approach that works for you.

Basically I am building something similar to tui-widget-list but not the same. The list I am building might have items taller than the view port so I want it scrolls pixel by pixel instead of item by item. Initially I want to implement it by creating a wrapper component of the built-in list with a scrollview but that would need ListItem to implement text wrapping. Now I guess I should put a Vec of Paragraphs inside a scrollview and manually manage the selection. Pre-wrapping the text isn't really a choice for me because I need to keep the styles, which means I need a solution at the StyledGraphemes level.

joshka:
>The list I am building might have items taller than the view port so I want it scrolls pixel by pixel instead of item by item

Makes sense - I think we'd probably also want that behavior in the main list eventually, but it's a difficult one to get right without breaking all existing code. And it has a lot of overlap with the concepts in https://github.com/ratatui-org/ratatui/issues/174 (another deep problem to solve).

PR #1093, opened 2024-05-11
chore(changelog): put commit id on the same line as the message

joshka:


joshka:
image

kdheepak:
Is that not what it does already? 

image

What's the difference?

joshka:
> What's the difference?

GitHub releases use GFM for displaying release notes. GfM interprets the previous format as a hard break instead of a soft break.

image

orhun (approved):

PR #1094, opened 2024-05-11
fix: termwiz underline color test

joshka:
Fixes code that doesn't compile in the termwiz tests when underline-color feature is enabled.


kdheepak (approved):

joshka:
Ugh - I blindly believed the message that stylize wasn't used -- fixing

PR #1097, opened 2024-05-12
feat: use inner Display implementation

EdJoPaTo:
This allows for `format!("{:>42})` formatting.
See https://doc.rust-lang.org/std/fmt/struct.Formatter.html#method.pad

---

I am not very attached to this. It feels more correct to reuse the inner Display implementation but we can also drop this. Main reason for this PR was to move it out of #1007

### DIFF
--- a/src/text/masked.rs
+++ b/src/text/masked.rs
@@ -45,16 +45,16 @@ impl<'a> Masked<'a> {
 
 impl fmt::Debug for Masked<'_> {
     /// Debug representation of a masked string is the underlying string
-    fn fmt(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         // note that calling display instead of Debug here is intentional
-        fmt::Display::fmt(&self.inner, formatter)
+        fmt::Display::fmt(&self.inner, f)
     }
 }
 
 impl fmt::Display for Masked<'_> {
     /// Display representation of a masked string is the masked string
-    fn fmt(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
-        fmt::Display::fmt(&self.value(), formatter)
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        fmt::Display::fmt(&self.value(), f)
     }
 }
 
--- a/src/text/masked.rs
+++ b/src/text/masked.rs
@@ -46,14 +46,15 @@ impl<'a> Masked<'a> {
 impl fmt::Debug for Masked<'_> {
     /// Debug representation of a masked string is the underlying string
     fn fmt(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
-        formatter.pad(&self.inner)
+        // note that calling display instead of Debug here is intentional
+        fmt::Display::fmt(&self.inner, formatter)
     }
 }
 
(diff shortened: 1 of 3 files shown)


joshka:
Added tests, and switched to use formatter.pad instead of the calling display on the inner value.
nit: This is somewhere where I'd generally feel fine using a single letter parameter name as that's idiomatic throughout the entire std lib whenever a debug / display impl is written.

joshka (approved):

src/text/masked.rs
```diff
@@ -45,15 +45,15 @@ impl<'a> Masked<'a> {
 
 impl fmt::Debug for Masked<'_> {
     /// Debug representation of a masked string is the underlying string
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.write_str(&self.inner)
+    fn fmt(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
+        formatter.pad(&self.inner)
```
EdJoPaTo, line 49: Why implement Display / Formatter stuff ourselves when we can just use the existing Display implementation?

joshka: because we know what the contents contain and can skip the extra layer of indirection. Happy to revert.

EdJoPaTo: But now we need to care how to implement `Display for str` well. The point of this PR is to not care about the implementation and use the Rust made implementation for the given thing we want to Display.

joshka: 402c5b45a

EdJoPaTo:
> nit: This is somewhere where I'd generally feel fine using a single letter parameter name as that's idiomatic throughout the entire std lib whenever a debug / display impl is written.

Personally I use `fmt` but so far I also used `core::fmt::…` so I'm not sure how much the similar names confuses in such a case. I have `clippy::min_ident_chars` enabled which is annoyed at the single character here so I change it regularly. We can use `f` in ratatui, I disable that lint in ratatui anyway as its very annoying here.

joshka:
> > nit: This is somewhere where I'd generally feel fine using a single letter parameter name as that's idiomatic throughout the entire std lib whenever a debug / display impl is written.
> 
> Personally I use `fmt` but so far I also used `core::fmt::…` so I'm not sure how much the similar names confuses in such a case. I have `clippy::min_ident_chars` enabled which is annoyed at the single character here so I change it regularly. We can use `f` in ratatui, I disable that lint in ratatui anyway as its very annoying here.

It's not worth changing this back, noting it as a not worth changing for future PRs.

joshka (approved):

orhun (approved):

EdJoPaTo:
> as that's idiomatic throughout the entire std lib

They are also not that aligned with more speaking variable / function names and like to abbreviate way more often than I like. I remember being annoyed by the lack of clear names in the std lib when I started with Rust as that was more confusing than necessary in my opinion. After all, most people use language servers with completion.

joshka:
> > as that's idiomatic throughout the entire std lib
> 
> They are also not that aligned with more speaking variable / function names and like to abbreviate way more often than I like. I remember being annoyed by the lack of clear names in the std lib when I started with Rust as that was more confusing than necessary in my opinion. After all, most people use language servers with completion.

I'm also a "nvr abbrv." opinion holder generally, so I see your point on this. I always rename f to frame whenever I see it in ratatui code, because it poorly communicates what the value is for readers. When faced with choosing idiom vs opinion, I tend towards accepting the former most of the time.

PR #1098, opened 2024-05-12
fix(buffer): fix Debug panic and fix formatting of overridden parts

EdJoPaTo:
Fix panic in `Debug for Buffer` when `width == 0`.
Also corrects the output when symbols are overridden.

---

Let's do some final testing for #1007… test runner aborts on panic in panic handler. Nice. :ghost:

### DIFF
--- a/src/buffer/buffer.rs
+++ b/src/buffer/buffer.rs
@@ -419,12 +419,13 @@ impl fmt::Debug for Buffer {
                     }
                 }
             }
+            f.write_str("\",")?;
             if !overwritten.is_empty() {
                 f.write_fmt(format_args!(
-                    "// hidden by multi-width symbols: {overwritten:?}"
+                    " // hidden by multi-width symbols: {overwritten:?}"
                 ))?;
             }
-            f.write_str("\",\n")?;
+            f.write_str("\n")?;
         }
         f.write_str("    ],\n    styles: [\n")?;
         for s in styles {
@@ -468,6 +469,24 @@ mod tests {
         let expected = "Buffer {
     area: Rect { x: 0, y: 0, width: 0, height: 0 }
 }";
+        assert_eq!(result, expected);
+    }
+
+    #[cfg(feature = "underline-color")]
+    #[test]
+    fn debug_grapheme_override() {
+        let buffer = Buffer::with_lines(["a🦀b"]);
+        let result = format!("{buffer:?}");
+        println!("{result}");
+        let expected = r#"Buffer {
+    area: Rect { x: 0, y: 0, width: 4, height: 1 },
+    content: [
+        "a🦀b", // hidden by multi-width symbols: [(2, " ")]
+    ],
+    styles: [
+        x: 0, y: 0, fg: Reset, bg: Reset, underline: Reset, modifier: NONE,
+    ]
+}"#;
         assert_eq!(result, expected);
     }
 
--- a/src/buffer/buffer.rs
+++ b/src/buffer/buffer.rs
@@ -381,14 +381,13 @@ impl fmt::Debug for Buffer {
     /// * `styles`: displayed as a list of: `{ x: 1, y: 2, fg: Color::Red, bg: Color::Blue,
     ///   modifier: Modifier::BOLD }` only showing a value when there is a change in style.
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_fmt(format_args!("Buffer {{\n    area: {:?}", &self.area))?;
+
         if self.area.is_empty() {
-            return f.write_fmt(format_args!("Buffer {{\n    area: {:?}\n}}", self.area));
+            return f.write_str("\n}");
         }
 
-        f.write_fmt(format_args!(
-            "Buffer {{\n    area: {:?},\n    content: [\n",
-            &self.area
-        ))?;
+        f.write_str(",\n    content: [\n")?;
         let mut last_style = None;
         let mut styles = vec![];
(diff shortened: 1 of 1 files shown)


src/buffer/buffer.rs
```diff
@@ -457,10 +462,39 @@ mod tests {
...
-        buf.set_string(0, 0, "Hello World!", Style::default());
-        buf.set_string(
+    fn debug_empty_buffer() {
+        let buffer = Buffer::empty(Rect::ZERO);
+        let result = format!("{buffer:?}");
+        println!("{result}");
+        let expected = "Buffer {
+    area: Rect { x: 0, y: 0, width: 0, height: 0 }
+}";
+        assert_eq!(result, expected);
```
EdJoPaTo, line 472: While refactoring the tests I used a formatting like the following. Maybe that's a bit more readable?
Using an additional macro (indoc) feels like overkill…

```suggestion
        let expected = "
Buffer {
    area: Rect { x: 0, y: 0, width: 0, height: 0 }
}";
        assert_eq!(result, expected.trim());
```

src/buffer/buffer.rs
```diff
@@ -457,10 +462,39 @@ mod tests {
...
+    }
+
+    #[cfg(feature = "underline-color")]
```
EdJoPaTo, line 475: Codecov seems to run without this feature. I mainly used it as it's irrelevant for this change to not write two versions of the same thing.

src/buffer/buffer.rs
```diff
@@ -381,6 +381,10 @@ impl fmt::Debug for Buffer {
     /// * `styles`: displayed as a list of: `{ x: 1, y: 2, fg: Color::Red, bg: Color::Blue,
     ///   modifier: Modifier::BOLD }` only showing a value when there is a change in style.
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        if self.area.is_empty() {
+            return f.write_fmt(format_args!("Buffer {{\n    area: {:?}\n}}", self.area));
```
joshka, line 385: How about writing this on one line for compactness of representation?

src/buffer/buffer.rs
joshka: I don't generally like seeing code in the left margin like this, which is why indoc works well (handling the indentation correctly. If you want to avoid indoc, perhaps use `\n` here instead (that's also what comes out of any test failure, so there's a convenience factor there).

src/buffer/buffer.rs
joshka: I'm fine with no coverage on the underline-color feature. It's rarely used feature. This might be better:
```suggestion
    #[cfg(not(feature = "underline-color"))]
```

src/buffer/buffer.rs
```diff
@@ -457,10 +462,39 @@ mod tests {
...
+        let buffer = Buffer::with_lines(["a🦀b"]);
+        let result = format!("{buffer:?}");
+        println!("{result}");
+        let expected = r#"Buffer {
+    area: Rect { x: 0, y: 0, width: 4, height: 1 },
+    content: [
+        "a🦀b", // hidden by multi-width symbols: [(2, " ")]
+    ],
+    styles: [
+        x: 0, y: 0, fg: Reset, bg: Reset, underline: Reset, modifier: NONE,
```
joshka, line 487:
```suggestion
        x: 0, y: 0, fg: Reset, bg: Reset, modifier: NONE,
```

src/buffer/buffer.rs
```diff
@@ -469,42 +503,34 @@ mod tests {
...
-                        \"G'day World!\",
-                    ],
-                    styles: [
-                        x: 0, y: 0, fg: Reset, bg: Reset, underline: Reset, modifier: NONE,
-                        x: 0, y: 1, fg: Green, bg: Yellow, underline: Reset, modifier: BOLD,
-                    ]
-                }"
-            )
-        );
+        let expected = r#"Buffer {
```
joshka, line 509: I prefer the indoc version as the indentation is more obvious and it doesn't cause the alignment of the code to look wonky.

joshka (approved):
Non blocking suggestions. Strong opinions weakly held here.

Can the commit message title be written in a more imperative form to describe the change in a way that makes more sense for the changelog? ("no" is ok here)

src/buffer/buffer.rs
EdJoPaTo: Then it's indentation differs from the one with content resulting in less easy visual comparison by the two.

src/buffer/buffer.rs
EdJoPaTo: It's enabled by default and having this test being active with default features seems smart

src/buffer/buffer.rs
joshka: In what circumstances would this matter?

EdJoPaTo: `assert_eq!(Buffer::empty(…), actual)`

joshka: In that case, `actual` is still displayed correctly which is all that matters, the test writing becomes change the Buffer::empty to Buffer::with_lines(<output copied from debug view of actual>), if actual is actually empty, then it doesn't really matter that much.

src/buffer/buffer.rs
joshka: Got it - it should be enabled in the coverage. Looking at the comment in the makefile and the PR where the feature flag was introduced, it seems this might have just been missed. PR incoming to fix that. No change necessary here.

joshka (approved):

PR #1099, opened 2024-05-12
feat: Make Stylize's `.bg(color)` generic

kdheepak:
This PR makes `.bg(color)` generic accepting anything that can be converted into `Color`; similar to the `.fg(color)` method on the same trait

### DIFF
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -10,6 +10,8 @@ GitHub with a [breaking change] label.
 
 This is a quick summary of the sections below:
 
+- [v0.26.1](#v0261)
+  - `Stylize::bg()` now accepts `Into<Color>`
 - [v0.26.0](#v0260)
   - `Flex::Start` is the new default flex mode for `Layout`
   - `patch_style` & `reset_style` now consume and return `Self`
@@ -47,6 +49,14 @@ This is a quick summary of the sections below:
   - MSRV is now 1.63.0
   - `List` no longer ignores empty strings
 
+## [v0.26.1](https://github.com/ratatui-org/ratatui/releases/tag/v0.26.1)
+
+### `Stylize::bg()` now accepts `Into<Color>` ([#1099])
+
+[#1099]: https://github.com/ratatui-org/ratatui/pull/1099
+
+Previously, `Stylize::bg()` accepted `Color` but now accepts `Into<Color>`. This allows more flexible types from calling scopes, though it can break some type inference in the calling scope.
+
 ## [v0.26.0](https://github.com/ratatui-org/ratatui/releases/tag/v0.26.0)
 
 ### `Flex::Start` is the new default flex mode for `Layout` ([#881])
@@ -74,7 +84,7 @@ existing layouts with `Flex::Start`. However, to get old behavior, use `Flex::Le
 
 [#774]: https://github.com/ratatui-org/ratatui/pull/774
 
-Previously, `Table::new()` accepted `IntoIterator<Item=Row<'a>>`.  The argument change to
+Previously, `Table::new()` accepted `IntoIterator<Item=Row<'a>>`. The argument change to
 `IntoIterator<Item: Into<Row<'a>>>`, This allows more flexible types from calling scopes, though it
 can some break type inference in the calling scope for empty containers.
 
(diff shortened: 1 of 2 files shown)


src/style/stylize.rs
```diff
@@ -137,7 +137,7 @@ macro_rules! modifier {
 /// ```
 pub trait Stylize<'a, T>: Sized {
     #[must_use = "`bg` returns the modified style without modifying the original"]
-    fn bg(self, color: Color) -> T;
+    fn bg<S: Into<Color>>(self, color: S) -> T;
```
EdJoPaTo, line 140: S was probably a short for Style some time ago…
```suggestion
    fn bg<C: Into<Color>>(self, color: C) -> T;
```

joshka (approved):
Does this need any updated tests?

EdJoPaTo (commented):
As every impl needs to change this is a breaking change

src/style/stylize.rs
```diff
@@ -179,8 +179,8 @@ impl<'a, T, U> Stylize<'a, T> for U
 where
     U: Styled<Item = T>,
```
EdJoPaTo, line 180: Ah… maybe this trait needs to be protected so it cant be implemented outside of ratatui… then it wouldnt be a breaking change to change this trait?

src/style/stylize.rs
kdheepak: Resolved in https://github.com/ratatui-org/ratatui/pull/1099/commits/090fb90075cd50d6a2f1935f6506cdbd60077155

src/style/stylize.rs
joshka: I'm fine with the breaking change - it's pretty unlikely anyone has implemented Stylize manually on a widget (it's difficult to write a github search query to prove this however).

kdheepak:
> Does this need any updated tests?

There weren't any tests for this. I think there's no implementations for `From` objects into `Color` in Ratatui. There's `crossterm::style::Color` -> `ratatui::style::Color` but that test will only work for the crossterm backend.

> As every impl needs to change this is a breaking change

This is strictly breaking but in practice I don't think it'll break code because `Stylize` is fairly new and people may have not had the chance or reason to implement it. In the past we've merged similar "strictly breaking but not breaking in practice" changes in patch releases. Thoughts?

Open to suggestions on making this protected.

src/style/stylize.rs
kdheepak: Arguably this is a mistake in our current code and this PR fixes it. 

And if someone has implemented `Stylize` they would have noticed the inconsistency between `fg()` and `bg()` and (hopefully) reported it. That gives me a lot of confidence that this will not break anyone's code.

joshka: yup - it could still be worth adding to the breaking changes doc, mainly because it's easier to just say all breaking changes are documented rather than all breaking changes that we think matter are.

joshka:
> Open to suggestions on making this protected.

No opinions - I know that I hate seeing traits in other libs that I'd like to be able to implement even if they later change them in the lib, but can't because they're not exposed, so I'd lean towards just keeping it as is, and dealing with breaking changes in the usual manners (communicating when we can, being mindful of the changes, etc.)

src/style/stylize.rs
kdheepak: I've added a section for v0.26.1 into the `BREAKING-CHANGES.md` file.

BREAKING-CHANGES.md
```diff
@@ -47,6 +49,14 @@ This is a quick summary of the sections below:
   - MSRV is now 1.63.0
   - `List` no longer ignores empty strings
 
+## [v0.26.1](https://github.com/ratatui-org/ratatui/releases/tag/v0.26.1)
```
joshka, line 52: Is this the correct version for this - likely 'unreleased' or 0.26.3

joshka:
Also - 0.26.3 was slated to be a bugfix / no breaking changes release tomorrow, to make it easy for broken apps to just be fixed without any changes required. (version spec 0.26.2 will actually use 0.26.3 automatically when resolving, but it won't use 0.27.0)

BREAKING-CHANGES.md
kdheepak: Ugh my bad.

kdheepak:
Should I revert this PR until 0.27.0?

PR #1100, opened 2024-05-12
build: add underline-color to all features flag in makefile

joshka:
Ensures that we upload coverage details for the underline-color tests (at the expense of possible removing some coverage where the feature is not enabled). Given `underline-color` is a default feature this is sensible.

Also fixes up a small build warning that pops up when not building with the underline-color feature

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -7,11 +7,12 @@ skip_core_tasks = true
 # all features except the backend ones
 ALL_FEATURES = "all-widgets,macros,serde"
 
+[env.ALL_FEATURES_FLAG]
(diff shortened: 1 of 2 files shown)


EdJoPaTo:
`cargo hack --keep-going --feature-powerset --no-dev-deps check` might be a good idea…

joshka:
> `cargo hack --keep-going --feature-powerset --no-dev-deps check` might be a good idea…

I agree, it's outside the scope of this change however. Let's add that sometime.

orhun (approved):

orhun:
Agree with adding `cargo-hack` to the CI. Might be a good first issue even.

PR #1101, opened 2024-05-12
chore: Fix version number in BREAKING-CHANGES.md

kdheepak:
Follow up to https://github.com/ratatui-org/ratatui/pull/1099#discussion_r1597679587


joshka:
Looking at this, it feels odd to see a breaking change in a 0.x.y release. How about we revert the change instead?

PR #1104, opened 2024-05-13
feat: Add selection of multiple marks for table widget

kdheepak:
This PR allows users to display a mark next to multiple rows. The marked rows are stored in the table.

e.g. here's a table with 2 rows marked and the first row selected (i.e. where the cursor currently is located).

```
• Cell Cell     // e.g. unmarked and selected row
⦾ Cell Cell    // e.g. marked row
  Cell Cell    // e.g. unmarked row
  Cell Cell    // e.g. unmarked row
⦾ Cell Cell    // e.g. marked row
  Cell Cell    // e.g. unmarked row    
```

If we move the cursor down 1, the marked row and the currently selected row overlap:

```
  Cell Cell     // e.g. unmarked row
⦿ Cell Cell    // e.g. marked and selected row
  Cell Cell    // e.g. unmarked row
  Cell Cell    // e.g. unmarked row
⦾ Cell Cell    // e.g. marked row
  Cell Cell    // e.g. unmarked row    
```

The index for all marked rows can be retrieved from `TableState::marked()`. This PR adds other methods to the `TableState` struct to manipulate the marks.

### DIFF
--- a/src/widgets/table/table.rs
+++ b/src/widgets/table/table.rs
@@ -210,8 +210,11 @@ pub struct Table<'a> {
     /// Style used to render the selected row
     highlight_style: Style,
 
-    /// Symbol in front of the selected rom
+    /// Symbol in front of the selected row
     highlight_symbol: Text<'a>,
+
+    /// Symbol in front of the marked row
+    mark_symbol: Text<'a>,
 
     /// Decides when to allocate spacing for the row selection
     highlight_spacing: HighlightSpacing,
@@ -232,6 +235,7 @@ impl<'a> Default for Table<'a> {
             style: Style::new(),
             highlight_style: Style::new(),
             highlight_symbol: Text::default(),
+            mark_symbol: Text::default(),
             highlight_spacing: HighlightSpacing::default(),
             flex: Flex::Start,
         }
@@ -503,6 +507,24 @@ impl<'a> Table<'a> {
         self
     }
 
+    /// Set the symbol to be displayed in front of the marked row
+    ///
+    /// This is a fluent setter method which must be chained or used as it consumes self
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// # use ratatui::{prelude::*, widgets::*};
+    /// # let rows = [Row::new(vec!["Cell1", "Cell2"])];
+    /// # let widths = [Constraint::Length(5), Constraint::Length(5)];
+    /// let table = Table::new(rows, widths).mark_symbol("\u{2714}");
+    /// ```
+    #[must_use = "method moves the value of self and returns the modified value"]
+    pub fn mark_symbol<T: Into<Text<'a>>>(mut self, mark_symbol: T) -> Self {
+        self.mark_symbol = mark_symbol.into();
+        self
+    }
+
     /// Set when to show the highlight spacing
     ///
     /// The highlight spacing is the spacing that is allocated for the selection symbol column (if
@@ -616,6 +638,7 @@ impl StatefulWidgetRef for Table<'_> {
             state,
             selection_width,
             &self.highlight_symbol,
+            &self.mark_symbol,
             &columns_widths,
         );
 
@@ -672,6 +695,7 @@ impl Table<'_> {
         state: &mut TableState,
         selection_width: u16,
         highlight_symbol: &Text<'_>,
+        marked_symbol: &Text<'_>,
         columns_widths: &[(u16, u16)],
     ) {
         if self.rows.is_empty() {
@@ -698,6 +722,7 @@ impl Table<'_> {
             );
             buf.set_style(row_area, row.style);
 
+            let is_marked = state.marked().contains(&(i + state.offset));
             let is_selected = state.selected().is_some_and(|index| index == i);
             if selection_width > 0 && is_selected {
                 let selection_area = Rect {
@@ -706,6 +731,15 @@ impl Table<'_> {
                 };
                 buf.set_style(selection_area, row.style);
                 highlight_symbol.clone().render(selection_area, buf);
+            } else if selection_width > 0 && is_marked {
+                let marked_area = Rect {
+                    width: selection_width,
+                    ..row_area
+                };
+                buf.set_style(marked_area, row.style);
+                marked_symbol.clone().render(marked_area, buf);
+            } else {
+                // we don't care about this
             };
             for ((x, width), cell) in columns_widths.iter().zip(row.cells.iter()) {
                 cell.render(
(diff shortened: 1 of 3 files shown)


kdheepak:
To keep this proof of concept PR small, I didn't address the following features:

- [ ] What should happen visually when selected row overlaps with a marked row? In `taskwarrior-tui`, I allow the user to choose. That requires an additional separate field `mark_and_highlight_symbol`. 
- [ ] Should we allow users to set a symbol for unmarked rows? In `taskwarrior-tui`, I allow the user to choose a `unmark_symbol` as well. This allows users to make rows with ballot boxes as the mark and unmark symbols.

```
☐ Unmarked Row 1
☑ Marked Row 1
☑ Marked Row 2
☐ Unmarked Row 1
```

- [ ] Currently in this PR, if a user wants to toggle between showing all marked rows or not, they'll have to store the state of the marked rows, drain the elements, and then later set them back again. In `taskwarrior-tui` I have `TableMode::SingleSelection` and `TableMode::MultipleSelection` that just changes the UI without changing the state.

joshka:
> * `TableMode::SingleSelection` and `TableMode::MultipleSelection`

Perhaps move `Selection` to the name of the enum? (I wouldn't be surprised if there is a pedantic lint for that).

I suspect this should be an enum named `ratatui::widgets::table::Selection` or something similar.

kdheepak:
I'm tempted to leave the `TableSelectionMode` out of this PR for a few reasons. I think it'll be possible to add it in a future PR if the need arises.

EdJoPaTo:
Selection is something done normally with the cursor. Which row is selected is moved up / down by keyboard / mouse.
Maybe split the naming more clearly, mark and cursor? At the position of the cursor a mark can be added/removed?

Maybe I am mainly tipped of by the PR title, haven't really looked into the implementation yet.

joshka:
Generally most widgets / controls in other frameworks will have a flag that is set to enable multi select. This is not generally enabled by default. It feels a bit odd to me for that to be different here. Does that sway you?

EdJoPaTo:
I am also fine with cursor and selection. Currently, the selection is the cursor. Mixing these feels weird to me.

> Generally most widgets / controls in other frameworks will have a flag that is set to enable multi select.

They are likely mouse based?

kdheepak:
Maybe the name of the PR should be "multi-mark" instead? We use the terminology "selection" to mean current cursor position. In this PR, you can "mark" the current cursor selection and the state will remember the `index`, and you can repeatedly "mark" multiple rows one after another.

e.g. in `taskwarrior-tui` I use `v` to toggle the mark on a current row. Here's an example where I have marked multiple rows:

image

joshka:
I think (could be worng) Selection and Highlight are the two concept names I'd use for what you're calling Mark and Selection. I'm not 100% sure of this though.

kdheepak:
We unfortunately already have `TableState::select`, `TableState::selected` and `TableState::with_selected`, so it'd be a breaking change to do it that way. 

But you are right, we use `highlight_symbol`, `highlight_spacing` and `HighlightSpacing` in `Table`. 

I'm open to suggestions on how to proceed.

joshka:
> We unfortunately already have `TableState::select`, `TableState::selected` and `TableState::with_selected`, so it'd be a breaking change to do it that way.

Yeah, I think this is a difficult one to get right - I wonder if the table would be worth moving out of Ratatui to get some higher velocity / experimental breaking changes like this and some of the other things (multi-select, built-in scrolling, column / cell selection, borders, property name changes etc.)?

kdheepak:
I see the appeal but I'm not sure I want to go down that route. I'd guess an external package containing a `Table` wouldn't see much use unless we removed `Table` from Ratatui. vc

src/widgets/table/table.rs
```diff
@@ -210,9 +210,18 @@ pub struct Table<'a> {
...
+    /// Symbol in front of the selected row
     highlight_symbol: Text<'a>,
 
+    /// Symbol in front of the marked row
+    mark_symbol: Text<'a>,
+
+    /// Symbol in front of the unmarked row
+    unmark_symbol: Text<'a>,
+
+    /// Symbol in front of the marked row
```
kdheepak, line 222:
```suggestion
    /// Symbol in front of the marked and selected row
```

src/widgets/table/table.rs
```diff
@@ -503,6 +515,60 @@ impl<'a> Table<'a> {
...
+    /// # use ratatui::{prelude::*, widgets::*};
+    /// # let rows = [Row::new(vec!["Cell1", "Cell2"])];
+    /// # let widths = [Constraint::Length(5), Constraint::Length(5)];
+    /// let table = Table::new(rows, widths).unmark_symbol(" ");
+    /// ```
+    #[must_use = "method moves the value of self and returns the modified value"]
+    pub fn unmark_symbol<T: Into<Text<'a>>>(mut self, unmark_symbol: T) -> Self {
+        self.unmark_symbol = unmark_symbol.into();
+        self
+    }
```
kdheepak, line 552: Is there a better name for this? `nomark_symbol`? `blank_symbol`?

joshka:
> I see the appeal but I'm not sure I want to go down that route. I'd guess an external package containing a `Table` wouldn't see much use unless we removed `Table` from `ratatui`.

We can feature flag the new table in or something like that. I guess it would only really cause problems for situations where another library uses the table widget as part of its rendering. That seems pretty niche however (perhaps tui-logger or similar crates might be affected).

Either way, given that the table has many things that it doesn't do, I'd really suggest that if we want to make it really nice getting it out of Ratatui and into a separate crate is the right thing to do. This would allow faster iteration, more frequent breaking changes / unstable ideas, and doing the right thing with respect to the existing downsides of the table without having to worry about existing users (e.g. rendering tables with large numbers of rows with reasonable perf)

EdJoPaTo:
What's the state on this? Is there some progress in other crates? Should this be closed as progress is happening elsewhere? Was this just abandoned? Should it be closed?

joshka:
Swinging back around to this, I wonder whether this should not be implemented. Thinking about how I'd implement this without this feature built into the table, I'd:
- add a column to the table
- keep track of which rows are marked in my app's data model

E.g. (pseudo-code)
```rust

struct Item {
    marker: Marker,
    col_1: String,
    col_2: String,
}

enum Marker {
    Todo,
    InProgress,
    Complete,
}

impl fmt::Display for Marker { ... }

impl From<&Item> for Row {
    fn from(value: Item) -> Self {
        Row::from([value.marker.to_string(), value.col_1, value.col_2])
    }
}

impl Item {
    fn toggle_marker(&mut self) { ... }
}

// then to use this:

let data = [
    Item::new(Marker::Todo, "foo", "bar"),
    Item::new(Marker::Todo, "baz", "quux"),
];

let rows = data.iter().map(Row::from);
let table = ...

match event::read()? {
	Event::Key(key) => match key.code {
		KeyCode::Space => data[table_state.selected].toggle_marker();
	},
}
```

This seems much easier and also more powerful than building this into the table. Is there an ergonomics issue with this that doesn't quite work?

This would make a good recipe on the website perhaps.

joshka:
@kdheepak can you add some more comments about the alternative approach I mentioned above?

joshka:
The Table experiment now lives at https://github.com/ratatui/ratatui-table, with focused successor tracking for this proposal at https://github.com/ratatui/ratatui-table/issues/11.

@kdheepak, would you be interested in recreating this as a focused PR against the extracted 0.1.0 baseline? I am leaving this PR open until a successor exists or you prefer to close it.

PR #1105, opened 2024-05-13
perf(layout): avoid allocating memory when using split ergonomic utils

tranzystorekk:
Noticed we create an intermediate vec in `Layout::areas` and `Layout::spacers` but there's no need for one.

### DIFF
--- a/src/layout/layout.rs
+++ b/src/layout/layout.rs
@@ -448,7 +448,7 @@ impl Layout {
     /// # }
     pub fn areas<const N: usize>(&self, area: Rect) -> [Rect; N] {
         let (areas, _) = self.split_with_spacers(area);
-        areas.to_vec().try_into().expect("invalid number of rects")
+        areas.as_ref().try_into().expect("invalid number of rects")
     }
 
     /// Split the rect into a number of sub-rects according to the given [`Layout`] and return just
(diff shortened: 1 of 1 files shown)


src/layout/layout.rs
```diff
@@ -448,7 +448,7 @@ impl Layout {
     /// # }
     pub fn areas<const N: usize>(&self, area: Rect) -> [Rect; N] {
         let (areas, _) = self.split_with_spacers(area);
-        areas.to_vec().try_into().expect("invalid number of rects")
+        areas.as_ref().try_into().expect("invalid number of rects")
```
EdJoPaTo, line 451: I am curious what the types are before and after the `try_into` and which actual implementation is used here to look into it.
Sometimes into implementations use clone or stuff like that internally, so maybe this has the same outcome?

EdJoPaTo (commented):
I like seeing improvements approaches like this!

src/layout/layout.rs
tranzystorekk: AFAIU `as_ref` produces a `&[Rect]` slice, then `try_into` trivially casts it to `&[Rect; N]` and then copies that to the result array.

kdheepak (approved):

src/layout/layout.rs
EdJoPaTo: Finding the actual implementations of (Try)Into stuff is horrible…

Looks like this still copies the `Rect`s, but that's way better than having a `Vec` on heap. :sparkles:

EdJoPaTo (approved):

PR #1112, opened 2024-05-15
fix(buffer): add try methods

TadoTheMiner:
This PR partially resolves issue #1011


.envrc
joshka, line 1: remove

flake.lock
joshka, line 1: remove

src/buffer/buffer.rs
```diff
@@ -93,19 +94,37 @@ impl Buffer {
     }
 
     /// Returns a reference to Cell at the given coordinates
+    /// # Panics
+    /// Panics if the coordinates are out of bounds
```
joshka, line 98:
```suggestion
    ///
    /// # Panics
    /// 
    /// Panics if the coordinates are out of bounds
```

src/buffer/buffer.rs
```diff
@@ -93,19 +94,37 @@ impl Buffer {
...
     #[track_caller]
     pub fn get_mut(&mut self, x: u16, y: u16) -> &mut Cell {
         let i = self.index_of(x, y);
         &mut self.content[i]
     }
 
+    /// Returns an option containing a mutable reference to Cell at the given coordinates, or
+    /// `None`, if the coordinates are out of bounds
+    pub fn try_get_mut(&mut self, x: u16, y: u16) -> Option<&mut Cell> {
+        let i = self.index_of(x, y);
```
joshka, line 124: This will panic

src/buffer/buffer.rs
```diff
@@ -134,17 +153,31 @@ impl Buffer {
...
+    ///
+    /// ```
+    /// # use ratatui::prelude::*;
+    /// let rect = Rect::new(200, 100, 10, 10);
+    /// let buffer = Buffer::empty(rect);
+    /// // Global coordinates to the top corner of this buffer's area
+    /// assert_eq!(buffer.try_index_of(200, 100), Some(0));
+    /// ```
+    #[must_use]
+    pub fn try_index_of(&self, x: u16, y: u16) -> Option<usize> {
```
joshka, line 176: I'd prefer not to add more cruft. The fact that the underlying storage of the cells is a Vec is an implementation detail, not something that should generally be relied on by callers. `index_of` should be pub(crate) as should anything in buffer that deals directly with the idea of the index. That's a bigger change than this however. But, adding another public method makes that more annoying.

src/buffer/buffer.rs
```diff
@@ -171,38 +204,102 @@ impl Buffer {
...
         (
-            self.area.x + (i as u16) % self.area.width,
-            self.area.y + (i as u16) / self.area.width,
+            area.x + (i as u16) % area.width,
+            area.y + (i as u16) / area.width,
         )
     }
 
     /// Print a string, starting at the position (x, y)
+    /// # Panics
```
joshka, line 240:
```suggestion
    ///
    /// # Panics
```

and elsewhere

src/buffer/buffer.rs
```diff
@@ -171,38 +204,102 @@ impl Buffer {
...
+        self.try_set_stringn(x, y, string, usize::MAX, style.into())
+            .unwrap_or_else(|| Self::panic_on_wrong_position(x, y, self.area));
+    }
+
+    /// Print a string, starting at the position (x, y)
+    ///
+    /// `style` accepts any type that is convertible to [`Style`] (e.g. [`Style`], [`Color`], or
+    /// your own type that implements [`Into<Style>`]).
+    #[must_use]
+    pub fn try_set_string<T, S>(&mut self, x: u16, y: u16, string: T, style: S) -> Option<()>
```
joshka, line 257: I'd like to eventually deprecate these methods and use Line/Span render instead.

src/buffer/buffer.rs
```diff
@@ -93,19 +94,37 @@ impl Buffer {
...
+    /// Panics if the coordinates are out of bounds
     #[track_caller]
     pub fn get(&self, x: u16, y: u16) -> &Cell {
         let i = self.index_of(x, y);
         &self.content[i]
     }
 
+    /// Returns an option containing a reference to Cell at the given coordinates, or `None`, if the
+    /// coordinates are out of bounds
+    pub fn try_get(&self, x: u16, y: u16) -> Option<&Cell> {
```
joshka, line 107: Should be called get_opt as `try_` indicates a method will return `Result`. We briefly discussed naming in the underlying issue.

joshka (commented):
There's an existing PR that covers much the same as this already:
https://github.com/ratatui-org/ratatui/pull/1084

Issue #1122, opened 2024-05-21
Make methods / fields on buffer that deal with the underlying content index private.


joshka:
I'd prefer to make the `content` field and the methods that deal directly with indexes (`index_of`, `pos_of`) private as it may allow us to migrate to more efficient storage approaches than we currently use. For now, I'd prefer to keep this as an additive only change and do the larger things separately after some discussion.

_Originally posted by @joshka in https://github.com/ratatui-org/ratatui/pull/1084#discussion_r1588893283_

pharrison31415:
What steps need to be taken before these methods can be deprecated?

orhun:
I believe the first thing to check would be searching if anyone out there depending on these methods currently to see the impact of this. (just a GitHub search maybe)

MusicalNinjaDad:
I was about to use them to obtain a sub-section of the buffer (for testing).

It would be nicely ergonomic to have something like an `impl<R,P> Index<R> for Buffer where R: RangeBounds<P>, P: Into<Position>`, which would remove this use case for `index_of`

PR #1123, opened 2024-05-21
chore: deprecate Buffer::get and get_mut

joshka:
Fixes: https://github.com/ratatui-org/ratatui/issues/1122


EdJoPaTo:
Probably superseded by #1084?

joshka:
> Probably superseded by #1084?

Yeah, unless we want to split the two for easier review.

EdJoPaTo:
Its only a `#[deprecated]` more, so not a huge difference on reviewing. The bigger question… do we want to deprecate directly? As this PR exists it seems to be something that was considered not to do at first?

Personally I think the Index trait is an exact replacement of the current get methods and the cell methods provide the new ones so it's only better than before with exact replacements. It would be better to include the deprecation to have a consistent usage.

joshka:
This is not about whether the deprecation happens, it's about whether it happens in the same PR as the other changes.

The reason for splitting this PR is for review purposes. Large PRs are more difficult to review, and cause problems when there are merges, more features changed, rebases, etc. The cognitive load required to review, and the amount of time necessary to dedicate to just opening a larger PR is usually much larger. Smaller amounts of work are generally easier to get over the line in a reasonable time, and this time scales in a non-linear manner with the PR size.

joshka:
Merged in #1084

PR #1124, opened 2024-05-22
refactor(cell): must_use and simplify style()

EdJoPaTo:

### DIFF
--- a/src/buffer/cell.rs
+++ b/src/buffer/cell.rs
@@ -35,6 +35,7 @@ pub struct Cell {
 
 impl Cell {
     /// Gets the symbol of the cell.
+    #[must_use]
     pub fn symbol(&self) -> &str {
         self.symbol.as_str()
     }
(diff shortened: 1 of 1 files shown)


src/buffer/cell.rs
joshka, line 1: It's a shame the fields on Cell are pub, which prevents us from just storing a style variable directly instead of by components. I wonder if this would be worth addressing (breaking change - not this PR)?

joshka (approved):

src/buffer/cell.rs
EdJoPaTo: On simpler structs without much logic on their own I like having stuff pub and being able to just use them directly.
Having Style on here would make Cell even simpler. But style allows for adding / removing modifiers while a Cell has one single merged modifier. Guess that's not bad on something that is persistent over multiple renders and exists many times.

Guess it has its pros and cons, but I would tend towards keeping it the way it is as a single modifier value makes more sense here.

joshka: https://rust-lang.github.io/api-guidelines/future-proofing.html#c-struct-private

> Public fields are most appropriate for struct types in the C spirit: compound, passive data structures. Otherwise, consider providing getter/setter methods and hiding fields instead.

We're on the cusp with Cell.

> But style allows for adding / removing modifiers while a Cell has one single merged modifier

It might be worth benchmarking this and understanding if there's any perf benefit (I suspect it's probably ~none).

Anyway - not blocking - feel free to merge this.

PR #1125, opened 2024-05-22
style: enable more rustfmt settings

EdJoPaTo:
Not sure whether I like the reordering of the trait types…


rustfmt.toml
```diff
@@ -1,6 +1,11 @@
 # configuration for https://rust-lang.github.io/rustfmt/
+comment_width = 100
+format_code_in_doc_comments = true
+format_macro_matchers=true
 group_imports = "StdExternalCrate"
 imports_granularity = "Crate"
+normalize_doc_attributes=true
+reorder_impl_items=true
+unstable_features=true
```
joshka, line 9: I just read a bit more about the [unstable_features](https://rust-lang.github.io/rustfmt/?version=v1.6.0&search=#unstable_features) setting. Sounds like this makes it possible to run the unstable rustfmt from the stable compiler!!! But, it's required to pass this at the command line for it to work (https://github.com/rust-lang/rustfmt/issues/3387#issuecomment-1867606088 / https://github.com/rust-lang/rustfmt/issues/5511#issuecomment-1867600019 have the details).

I think perhaps add this to the makefile and ci.yml to make this part of the formatting / checks that happen there and we can drop one of the needs for nightly. Docs lint still requires it in order to scrape the examples. I wonder if we actually need that in CI, or if we could just compile all with stable?

src/text/line.rs
```diff
@@ -475,26 +475,26 @@ impl<'a> Line<'a> {
 }
 
 impl<'a> IntoIterator for Line<'a> {
-    type Item = Span<'a>;
     type IntoIter = std::vec::IntoIter<Span<'a>>;
```
joshka, line 478: The IntoIter types can use Self::Item for a small simplification. Not a blocker though.
```suggestion
    type IntoIter = std::vec::IntoIter<Self::Item>;
```

I'm not too worried about the type order here - it's odd (because it changes the order from the definition), but it's not something that would make anyone unable to understand the code.

joshka (approved):

src/text/line.rs
EdJoPaTo: It's a bit funny given that the formatter isn't applied on the std lib there :upside_down_face:

joshka (approved):

rustfmt.toml
joshka: Did you want to do the CI / makefile changes before merging? Happy either way.

EdJoPaTo: I am thinking about removing the `unstable_feature` from the config file… Given that they are unstable its kind of correct to prefer nightly for this.

I am also curious about these things getting stablized. Some of them are also done by rust-analyzer so nothing completely experimental.

joshka (approved):

rustfmt.toml
EdJoPaTo: I still think keeping unstable stuff on nightly seems the better thing to do. I wonder why this even exists as its either stable or not… this is not, but it's on the stable rust version… somewhat weird.

PR #1126, opened 2024-05-22
build(bench): improve benchmark consistency

EdJoPaTo:
Codegen units are optimized on their own. Per default bench / release have 16 codegen units. What ends up in a codeget unit is rather random and can influence a benchmark result as a code change can move stuff into a different codegen unit → prevent / allow LLVM optimizations unrelated to the actual change.

More details: https://doc.rust-lang.org/cargo/reference/profiles.html

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -152,6 +152,11 @@ all-features = true
 cargo-args = ["-Zunstable-options", "-Zrustdoc-scrape-examples"]
 rustdoc-args = ["--cfg", "docsrs"]
 
+# Improve benchmark consistency
+[profile.bench]
+codegen-units = 1
+lto = true
+
 [[bench]]
 name = "barchart"
 harness = false


joshka (approved):
Can you link to a doc for this to make it easier to understand why this improves the consistency for a later reader of the cargo.toml?

EdJoPaTo:
> Can you link to a doc for this to make it easier to understand why this improves the consistency for a later reader of the cargo.toml?

I updated the PR description with a link to the profile options. But a reader probably needs to read more into codegen units and lto so not sure whether there is a great link to add here.
Personally I would check for the Commit / PR that added it to find more details… But that isnt perfect either.

joshka:
> > Can you link to a doc for this to make it easier to understand why this improves the consistency for a later reader of the cargo.toml?
> 
> I updated the PR description with a link to the profile options. But a reader probably needs to read more into codegen units and lto so not sure whether there is a great link to add here. Personally I would check for the Commit / PR that added it to find more details… But that isnt perfect either.

Generally I like to have non-obvious things explained right in front of my face and not have to guess where to look for docs. A comment in a commit won't survive reformatting that hits the line being commented. Is there some extra doc you can link to that documents the impact of codegen-units / lto on the benchmark tooling? The best I've found is perhaps https://nnethercote.github.io/perf-book/build-configuration.html#maximizing-runtime-speed

EdJoPaTo:
> A comment in a commit won't survive reformatting that hits the line being commented.

`git blame` is my friend ;)

> Is there some extra doc you can link to that documents the impact of codegen-units / lto on the benchmark tooling? The best I've found is […]

I would prefer to stick to official Rust documentation when possible as it will be kept more likely up to date / available. The profile stuff isn't that bad as it continues to link further and is exactly what is changed/configured here. It's not perfect but a starting point?

PR #1127, opened 2024-05-22
refactor(cell): reset instead of applying default

EdJoPaTo:
Using reset is clearer to me what actually happens. On the other case a struct is created to override the old one completely which basically does the same in a less clear way.

### DIFF
--- a/examples/demo2/destroy.rs
+++ b/examples/demo2/destroy.rs
@@ -1,6 +1,6 @@
 use rand::Rng;
 use rand_chacha::rand_core::SeedableRng;
-use ratatui::{buffer::Cell, layout::Flex, prelude::*};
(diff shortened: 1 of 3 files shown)


joshka (approved):
As an alternative, what about adding a Buffer::reset_range() method - something like:
```rust
buffer.reset_range(Position(x, y)..Position(x1, y1));
```

Probably unnecessary over-engineering (this PR is approved without that change)

EdJoPaTo:
> As an alternative, what about adding a Buffer::reset_range() method - something like:
> 
> ```rust
> buffer.reset_range(Position(x, y)..Position(x1, y1));
> ```
> 
> Probably unnecessary over-engineering (this PR is approved without that change)

Guess that would be neat for #1122. And probably out of scope for this PR.

PR #1130, opened 2024-05-23
refactor: must_use

EdJoPaTo:
Add `#[must_use]` where appropriate.

While this could be breaking code it should only break when the code is wrong before as the thing marked as `#[must_use]` must be used to be useful.


joshka:
If it's weird and leads to surprising results, then it's probably not worth complicating things. Clever code is rarely worth it. Can you show the results and explain them?

src/layout/layout.rs
```diff
@@ -217,6 +220,7 @@ impl Layout {
     /// * is called before [`Layout::split()`] otherwise, the cache size is
     ///   [`Self::DEFAULT_CACHE_SIZE`].
     /// * is called for the first time, subsequent calls do not modify the cache size.
+    #[allow(clippy::must_use_candidate)] // TODO: should return Result?
     pub fn init_cache(cache_size: usize) -> bool {
```
EdJoPaTo, line 224: This return value isn't always used… And somehow I think it should be handled as it's basically a `Result::is_ok -> bool`.

Personally I think this is a bug that should be fixed in its own PR?

joshka (changes requested):
I started to review this, but stopped about half way through because this seems like spam to have it on every method that returns a value and doesn't modify state.

Instead, I'd like to suggest the applying the following metric for choosing whether to add must_use:

Not using this value would be a surprise because the method looks like it should modify state.

This idea is supported by https://users.rust-lang.org/t/liberal-application-of-must-use/85173/20, https://std-dev-guide.rust-lang.org/policy/must-use.html, https://internals.rust-lang.org/t/use-of-must-use-in-the-standard-library/15425

Thus constructors (new, styled, horizontal, etc.), types (Palette, ...), calculations (width), etc. should not be marked must_use as while not using them is an error, it's not one that would generally confuse someone in a way that matters.

For pretty much everywhere in Ratatui which is not a builder lite method (i.e. something that take self, and returns Self), this is just extraneous noise that seems to add very little value and makes the code worse to maintain. There's a reason why this is in the pedantic lints. We should not enable this lint more generally on Ratatui IMO.

src/buffer/buffer.rs
```diff
@@ -559,7 +567,8 @@ mod tests {
         let buf = Buffer::empty(rect);
 
         // There are a total of 100 cells; zero-indexed means that 100 would be the 101st cell.
-        buf.pos_of(100);
+        let result = buf.pos_of(100);
+        dbg!(result);
```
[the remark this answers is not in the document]
EdJoPaTo: https://rust-lang.github.io/rust-clippy/master/index.html#/let_underscore_untyped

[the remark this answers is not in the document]
joshka: That's in clippy:restriction and contains the following:

> The _ = <expr> is not properly supported by some tools (e.g. IntelliJ) and may seem odd to many developers. This lint also partially overlaps with the other let_underscore_* lints.

The syntax seems odd enough to me that I had to look up what the effect of this was.
This is never a bug (or one that just doesn't matter).

src/style/stylize.rs
```diff
@@ -14,12 +14,14 @@ pub trait Styled {
...
 
     /// Returns the style of the object.
+    #[must_use]
     fn style(&self) -> Style;
 
     /// Sets the style of the object.
     ///
     /// `style` accepts any type that is convertible to [`Style`] (e.g. [`Style`], [`Color`], or
     /// your own type that implements [`Into<Style>`]).
+    #[must_use = "https://ratatui.rs/concepts/builder-lite-pattern/"]
```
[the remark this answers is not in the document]
EdJoPaTo: I think it is: `Something::set_style` moves something like done on the builder-lite pattern.

The only exception is the implementation of &str which is an exception to the normal implementation of this trait.

EdJoPaTo:
> Thus constructors (new, styled, horizontal, etc.), types (Palette, ...), calculations (width), etc. should not be marked must_use as while not using them is an error, it's not one that would generally confuse someone in a way that matters.
> 
> For pretty much everywhere in Ratatui which is not a builder lite method (i.e. something that take self, and returns Self), this is just extraneous noise that seems to add very little value and makes the code worse to maintain.

Personally, I like having a compiler that can actually help me find mistakes especially on refactorings. When the return value isn't used, the method needs to have a side effect, or it's either wrong or a waste of performance. While I can do something like that manually the compiler can be way better at finding stuff like that when we allow it to do so.

Rust's philosophy is to be explicit about what happens. Result needs to be explicitly handled which is one of the core differences to other languages. I like this philosophy and would like libraries to embrace that too. (But we seem to have different opinions on that, see #1047)

joshka:
> Personally, I like having a compiler that can actually help me find mistakes especially on refactorings. When the return value isn't used, the method needs to have a side effect, or it's either wrong or a waste of performance. While I can do something like that manually the compiler can be way better at finding stuff like that when we allow it to do so.
> 
> Rust's philosophy is to be explicit about what happens. Result needs to be explicitly handled which is one of the core differences to other languages. I like this philosophy and would like libraries to embrace that too. (But we seem to have different opinions on that, see #1047)

I'm 70/30 on not adding extra stuff to maintain that's unnecessary and noisy. For me the maintenance burden of being corrected and having to think about must_use, and being annoyed by future lint errors that pick this up, extra code to use values that you don't care about, and just the extra lines of code to read etc. outweighs the benefit to end users on things which are almost always an obvious error.

Let's take an extremely simplified argument and walk back:
```rust
fn add(a: usize, b: usize) -> usize {
   a + b;
}
```
Is this an error? Sure. Do I need a must_use on the addition operator to work that out? No, because my function returns `()` instead of usize. In contrast, adding a must use on `saturating_add()` is useful because it's easy to mistakenly assume when reading that this is a mutating method.

To help try and understand why this is necessary I'd like to ask what does a failing line of code that hits the `must_use` error (for one of the non-builder methods) look like. E.g. `Line::width()`, `Text::styled()`, `List::new()`, `TableState::selected()`? What's the scenario / code that this helps make better for each of these?

EdJoPaTo:
> for one of the non-builder methods

Following your logic: it’s hard to misuse the builder methods as they move the input (except for Copy, like Style). So according to that must_use could be removed in almost all places where it’s currently used in this project.

> not adding extra stuff to maintain that's unnecessary and noisy

I don’t see a maintenance burden here. In regards to noisy: Rust is designed to be explicit. In other words noisy. That’s one of its core principles. And I appreciate that about Rust. It results in explicit code. I am explicitly handling the cases like errors. The compiler helps me that I do it correctly by being noisy.

> extra code to use values that you don't care about

When that’s the case then the `#[must_use]` is wrong there and should be removed. The cases where this came up most of the time were it tests which tested about panics (=side effects) and not the actual correct usage of the method.

> extra lines of code to read etc. outweighs the benefit to end users on things which are almost always an obvious error

It’s a single line resulting in the compiler checking that the user uses a method in the correct way. Users (myself included) are not perfect and „almost always obvious“ will always result in mistakes. Like ignoring the that an intersections might not intersect. Or ignoring whatever the return value of init_cache means.

Humans do make mistakes. Like not noticing that a string slicing might panic. Multiple humans failed to notice this „obvious“ mistake while a lint (string_slice is also restriction btw) notices that in a deterministic way.

I don’t want to bet on humans noticing almost always obvious things when the compiler can make sure the obvious thing is done correctly.

> Let's take an extremely simplified argument and walk back:
> 
> ```rust
> fn add(a: usize, b: usize) -> usize {
>    a + b;
> }
> ```
> 
> Is this an error? Sure. Do I need a must_use on the addition operator to work that out? No, because my function returns `()` instead of usize.

This example has nothing to do with `#[must_use]` as it doesn’t care about the return value. Your example is inside the method trying to make sure the return value is actually given. the following code would be fine without must_use. It’s obvious to a human that it’s wrong but the compiler could warn that it’s fishy.

```rust
let mut foo = 42;
foo *= 2;
add(foo, 1337);
println!("{foo}");
```

Now let’s assume a more complex code base where I refactor stuff and don’t read every doc comment of every method I use to ensure I don’t overlook the obvious.

tldr: I think must_use helps the compiler to ensure correct code. When the code is correct without using the result, then it‘s not must_use. I use the lints for quite some time and add must_use in more places than the lint suggests (like on private methods). I personally don’t notice a maintenance burden on my projects. But I do notice Rust being a compiler / linter that helps me write (and refactor) correct code.

joshka:
> > for one of the non-builder methods
> 
> Following your logic: it’s hard to misuse the builder methods as they move the input (except for Copy, like Style). So according to that must_use could be removed in almost all places where it’s currently used in this project.

Sounds good to me...

> > not adding extra stuff to maintain that's unnecessary and noisy
> 
> I don’t see a maintenance burden here. In regards to noisy: Rust is designed to be explicit. In other words noisy. That’s one of its core principles. And I appreciate that about Rust. It results in explicit code. I am explicitly handling the cases like errors. The compiler helps me that I do it correctly by being noisy.

When I push a PR and get a failure from CI because of something I didn't think about like a clippy lint that doesn't really add value, that costs extra time in adding another commit, repushing. That's the maintenance burden I'd prefer to avoid.

Case in point: db2b1a44a Neither of the Clippy issues fixed would benefit anyone by the amount of time that it cost to fix them.

> > extra code to use values that you don't care about
> 
> When that’s the case then the `#[must_use]` is wrong there and should be removed. The cases where this came up most of the time were it tests which tested about panics (=side effects) and not the actual correct usage of the method.
> 
> > extra lines of code to read etc. outweighs the benefit to end users on things which are almost always an obvious error
> 
> It’s a single line resulting in the compiler checking that the user uses a method in the correct way. Users (myself included) are not perfect and „almost always obvious“ will always result in mistakes. Like ignoring the that an intersections might not intersect. Or ignoring whatever the return value of init_cache means.

It's 180 extra lines across 51 files in this commit and all the future errors.

> Humans do make mistakes. Like not noticing that a string slicing might panic. Multiple humans failed to notice this „obvious“ mistake while a lint (string_slice is also restriction btw) notices that in a deterministic way.
> 
> I don’t want to bet on humans noticing almost always obvious things when the compiler can make sure the obvious thing is done correctly.

This is not "correct" it's "pedantically verbosely exact"

> > Let's take an extremely simplified argument and walk back:
> > ```rust
> > fn add(a: usize, b: usize) -> usize {
> >    a + b;
> > }
> > ```
> >   
> > Is this an error? Sure. Do I need a must_use on the addition operator to work that out? No, because my function returns `()` instead of usize.
> 
> This example has nothing to do with `#[must_use]` as it doesn’t care about the return value. Your example is inside the method trying to make sure the return value is actually given. the following code would be fine without must_use. It’s obvious to a human that it’s wrong but the compiler could warn that it’s fishy.

> ```rust
> let mut foo = 42;
> foo *= 2;
> add(foo, 1337);
> println!("{foo}");
> ```
> 
> Now let’s assume a more complex code base where I refactor stuff and don’t read every doc comment of every method I use to ensure I don’t overlook the obvious.
> 
> tldr: I think must_use helps the compiler to ensure correct code. When the code is correct without using the result, then it‘s not must_use. I use the lints for quite some time and add must_use in more places than the lint suggests (like on private methods). I personally don’t notice a maintenance burden on my projects. But I do notice Rust being a compiler / linter that helps me write (and refactor) correct code.

Just doing the review of this is a maintenance burden... It distracts from higher value tasks that would be more beneficial than chasing pedantic / restrictive lints.

EdJoPaTo:
> When I push a PR and get a failure from CI because of something I didn't think about like a clippy lint that doesn't really add value, that costs extra time in adding another commit, repushing.

This describes what I commonly had when adding more lints to my default lints (like starting to use clippy, later pedantic lints, …). I wasn't used to that way of doing things. The compiler / linter suggested me what to do, and it got normal for me. Now I do these things directly without thinking about it much. Its especially helpful when the feedback is quick (= rust-analyzer in your IDE).

src/text/span.rs
```diff
@@ -284,13 +287,14 @@ impl<'a> Span<'a> {
     /// # use ratatui::prelude::*;
     /// let line = "Test Content".green().italic().into_left_aligned_line();
     /// ```
-    #[must_use = "method moves the value of self and returns the modified value"]
+    #[must_use = "https://ratatui.rs/concepts/builder-lite-pattern/"]
     pub fn into_left_aligned_line(self) -> Line<'a> {
```
joshka, line 291: I think this one should just be must_use

src/text/span.rs
```diff
@@ -284,13 +287,14 @@ impl<'a> Span<'a> {
...
         Line::from(self).left_aligned()
     }
 
     #[allow(clippy::wrong_self_convention)]
     #[deprecated = "use into_left_aligned_line"]
+    #[must_use = "https://ratatui.rs/concepts/builder-lite-pattern/"]
     pub fn to_left_aligned_line(self) -> Line<'a> {
```
joshka, line 298: Just must_use

src/text/span.rs
```diff
@@ -303,13 +307,14 @@ impl<'a> Span<'a> {
     /// # use ratatui::prelude::*;
     /// let line = "Test Content".green().italic().into_centered_line();
     /// ```
-    #[must_use = "method moves the value of self and returns the modified value"]
+    #[must_use = "https://ratatui.rs/concepts/builder-lite-pattern/"]
     pub fn into_centered_line(self) -> Line<'a> {
```
joshka, line 311: Just must_use

src/text/span.rs
```diff
@@ -303,13 +307,14 @@ impl<'a> Span<'a> {
...
         Line::from(self).centered()
     }
 
     #[allow(clippy::wrong_self_convention)]
     #[deprecated = "use into_centered_line"]
+    #[must_use = "https://ratatui.rs/concepts/builder-lite-pattern/"]
     pub fn to_centered_line(self) -> Line<'a> {
```
joshka, line 318: Just must_use

src/text/span.rs
```diff
@@ -322,13 +327,14 @@ impl<'a> Span<'a> {
     /// # use ratatui::prelude::*;
     /// let line = "Test Content".green().italic().into_right_aligned_line();
     /// ```
-    #[must_use = "method moves the value of self and returns the modified value"]
+    #[must_use = "https://ratatui.rs/concepts/builder-lite-pattern/"]
     pub fn into_right_aligned_line(self) -> Line<'a> {
```
joshka, line 331: Just must_use

src/text/span.rs
```diff
@@ -322,13 +327,14 @@ impl<'a> Span<'a> {
...
         Line::from(self).right_aligned()
     }
 
     #[allow(clippy::wrong_self_convention)]
     #[deprecated = "use into_right_aligned_line"]
+    #[must_use = "https://ratatui.rs/concepts/builder-lite-pattern/"]
     pub fn to_right_aligned_line(self) -> Line<'a> {
```
joshka, line 338: Just must_use

src/widgets/gauge.rs
```diff
@@ -460,19 +461,22 @@ mod tests {
     #[test]
     #[should_panic = "Percentage should be between 0 and 100 inclusively"]
     fn gauge_invalid_percentage() {
-        let _ = Gauge::default().percent(110);
+        let gauge = Gauge::default().percent(110);
+        dbg!(gauge);
```
joshka, line 465: I'd prefer the `let _` approach on these (generally) for reasons previously discussed. Unless I'm missing a rationale for doing this?

src/widgets/gauge.rs
```diff
@@ -460,19 +461,22 @@ mod tests {
...
     }
 
     #[test]
     #[should_panic = "Ratio should be between 0 and 1 inclusively"]
     fn gauge_invalid_ratio_upper_bound() {
-        let _ = Gauge::default().ratio(1.1);
+        let gauge = Gauge::default().ratio(1.1);
+        dbg!(gauge);
```
joshka, line 472: let _

src/widgets/gauge.rs
```diff
@@ -460,19 +461,22 @@ mod tests {
...
     }
 
     #[test]
     #[should_panic = "Ratio should be between 0 and 1 inclusively"]
     fn gauge_invalid_ratio_lower_bound() {
-        let _ = Gauge::default().ratio(-0.5);
+        let gauge = Gauge::default().ratio(-0.5);
+        dbg!(gauge);
```
joshka, line 479: let _

src/widgets/table/table.rs
```diff
@@ -971,7 +976,8 @@ mod tests {
     #[test]
     #[should_panic = "Percentages should be between 0 and 100 inclusively"]
     fn table_invalid_percentages() {
-        let _ = Table::default().widths([Constraint::Percentage(110)]);
+        let table = Table::default().widths([Constraint::Percentage(110)]);
+        dbg!(table);
```
joshka, line 980: let _

joshka (commented):
I spent a bit more time thinking about this overnight, and I think your arguments about correctness land better than mine about maintenance burden. As such I reviewed the rest of the code from where I'd left off. I'm not sure if it's possible to reopen this by undeleting the branch. Let's work out a way to move ahead with this if you're still keen to have it in.

I'd prefer the `let _` approach in the tests over `let x = ...; dbg(x);`, but the code doesn't really get executed unless the test is failing, so this is a minor concern if there's a rationale for this that I'm missing.

I'm not sure that the link makes sense on the `into` / `to` methods - perhaps just make those without the link?

PR #1139, opened 2024-05-25
fix(color): change parse error message to single color

EdJoPaTo:
Found while enabling <https://rust-lang.github.io/rust-clippy/master/index.html#assertions_on_result_states>

### DIFF
--- a/src/style/color.rs
+++ b/src/style/color.rs
@@ -187,7 +187,7 @@ impl<'de> serde::Deserialize<'de> for Color {
     /// assert!(err.is_data());
     /// assert_eq!(
     ///     err.to_string(),
-    ///     "Failed to parse Colors at line 1 column 20"
+    ///     "Failed to parse Color at line 1 column 20"
     /// );
     ///
     /// // Deserializing from the previous serialization implementation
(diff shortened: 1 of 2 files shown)


kdheepak (approved):

src/style/color.rs
```diff
@@ -225,7 +222,7 @@ impl<'de> serde::Deserialize<'de> for Color {
         }
 
         let multi_type = ColorFormat::deserialize(deserializer)
-            .map_err(|err| serde::de::Error::custom(format!("Failed to parse Colors: {err}")))?;
+            .map_err(|err| serde::de::Error::custom(format!("Failed to parse Color: {err}")))?;
```
joshka, line 225: I don't think this can actually fail as deserializing to ColorFormat::V2 accepts any string. There's no test for this and I can't work out how to write one that triggered this message. The message in the tests is actually coming from the FromStr call, not this line).

src/style/color.rs
```diff
@@ -654,14 +651,20 @@ mod tests {
 
     #[cfg(feature = "serde")]
     #[test]
-    fn deserialize_error() {
+    #[should_panic = "Failed to parse Color"]
+    fn deserialize_unknown_string_errors() {
         let color: Result<_, serde::de::value::Error> =
             Color::deserialize("invalid".into_deserializer());
-        assert!(color.is_err());
+        color.unwrap();
```
joshka, line 658: Can these be expressed as an `assert_eq!` call against `<some instance of the Error trait>.custom("failed...")` (possibly https://docs.rs/serde/1.0.202/serde/de/value/struct.Error.html ?)

joshka (changes requested):
These messages should be lowercase. Is there a better error we can return than custom - like invalid value?

EdJoPaTo:
> These messages should be lowercase.

Thought about that too, but it's about the struct `Color`.

joshka:
> > These messages should be lowercase.
> 
> Thought about that too, but it's about the struct `Color`.

Failed -> failed

EdJoPaTo:
My main intention was to enable this lint and fix the mistake of outputting colors instead of color. As this seems to be more complicated about stuff I don't use in the first place (= don't care much about it) I'll just close this.

PR #1143, opened 2024-05-25
feat(cell): add EMPTY and (const) new method

EdJoPaTo:
This simplifies calls to `Buffer::filled` in tests.

### DIFF
--- a/src/buffer/cell.rs
+++ b/src/buffer/cell.rs
@@ -34,6 +34,39 @@ pub struct Cell {
 }
 
 impl Cell {
+    /// An empty `Cell`
+    pub const EMPTY: Self = Self::new_inline(" ");
+
+    /// Creates a new `Cell` with the given symbol.
+    pub fn new(symbol: &str) -> Self {
+        Self {
+            symbol: CompactString::new(symbol),
+            fg: Color::Reset,
+            bg: Color::Reset,
+            #[cfg(feature = "underline-color")]
+            underline_color: Color::Reset,
+            modifier: Modifier::empty(),
+            skip: false,
+        }
+    }
+
+    /// Creates a new `Cell` with the given symbol.
+    ///
+    /// This works at compile time and puts the symbol onto the stack. Fails to build when the
+    /// symbol doesnt fit onto the stack and requires to be placed on the heap. Use [`Self::new`] in
+    /// that case. See [`CompactString::new_inline`] for more details on this.
+    pub const fn new_inline(symbol: &str) -> Self {
+        Self {
+            symbol: CompactString::new_inline(symbol),
+            fg: Color::Reset,
+            bg: Color::Reset,
+            #[cfg(feature = "underline-color")]
+            underline_color: Color::Reset,
+            modifier: Modifier::empty(),
+            skip: false,
+        }
+    }
+
     /// Gets the symbol of the cell.
     #[must_use]
     pub fn symbol(&self) -> &str {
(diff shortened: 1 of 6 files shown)


EdJoPaTo:
A (breaking) follow up to this will be the following performance improvement (removes needless one needless):

```diff
     /// Returns a Buffer with all cells initialized with the attributes of the given Cell
     #[must_use]
-    pub fn filled(area: Rect, cell: &Cell) -> Self {
+    pub fn filled(area: Rect, cell: Cell) -> Self {
         let size = area.area() as usize;
-        let content = vec![cell.clone(); size];
+        let content = vec![cell; size];
         Self { area, content }
     }
```

As this would result in a bunch of conflicts (and is breaking) I'll wait with that until this one is sorted out.

src/buffer/cell.rs
```diff
@@ -34,6 +34,39 @@ pub struct Cell {
...
+            skip: false,
+        }
+    }
+
+    /// Creates a new `Cell` with the given symbol.
+    ///
+    /// This works at compile time and puts the symbol onto the stack. Fails to build when the
+    /// symbol doesnt fit onto the stack and requires to be placed on the heap. Use [`Self::new`] in
+    /// that case. See [`CompactString::new_inline`] for more details on this.
+    pub const fn new_inline(symbol: &str) -> Self {
```
joshka, line 58: Would it be the worst thing to make this the default (i.e. `new()`) and just not support creating new Cells with long symbols that can't be stored on the stack without using `Cell::default().set_symbol()`? I suspect that there is rarely if ever a case where the cell can't be stored on the stack (long ANSI strings spring to mind as a counterpoint).

joshka (approved):

joshka:
> A (breaking) follow up to this will be the following performance improvement (removes one needless clone / drop):
> 
> ```diff
>      /// Returns a Buffer with all cells initialized with the attributes of the given Cell
>      #[must_use]
> -    pub fn filled(area: Rect, cell: &Cell) -> Self {
> +    pub fn filled(area: Rect, cell: Cell) -> Self {
>          let size = area.area() as usize;
> -        let content = vec![cell.clone(); size];
> +        let content = vec![cell; size];
>          Self { area, content }
>      }
> ```

+1 to this - looks like a good idea. It's breaking but in a function that's only really useful for testing, so that's not too bad.
https://github.com/search?q=lang%3Arust+ratatui+%22Buffer%3A%3Afilled%22+AND+%28NOT+org%3Aratatui-org%29+AND+%28NOT+is%3Afork%29&type=code

src/buffer/cell.rs
EdJoPaTo: Especially as its only the creation. Normally set_symbol is used on the existing ones on the existing buffer.
So yeah, can be the standard new method.

joshka (approved):
Another idea that springs from this - doing a quick search suggests perhaps a `Cell::styled()` constructor could be useful
```rust
// instead of:
buf.get_mut(x,y)
  .set_symbol("x")
  .set_style(style);
// what about:
buf[(x,y)] = Cell::styled("x", style);
```
I'm not sure what impact this would have on allocations (likely many) / perf (likely little). Doesn't have to be part of this PR at all, but I'm curious to hear your thoughts on this.

EdJoPaTo:
> Another idea that springs from this - doing a quick search suggests perhaps a `Cell::styled()` constructor could be useful
> 
> ```rust
> // instead of:
> buf.get_mut(x,y)
>   .set_symbol("x")
>   .set_style(style);
> // what about:
> buf[(x,y)] = Cell::styled("x", style);
> ```
> 
> I'm not sure what impact this would have on allocations (likely many) / perf (likely little). Doesn't have to be part of this PR at all, but I'm curious to hear your thoughts on this.

`set_styled` could be a neat idea for simpler code. For performance it would be the same as the two methods `set_symbol` / `set_style`.
Using a constructor and overriding it seems weird, this should stick with `&mut`.

The constructor in this PR is mainly simplifying tests.

PR #1147, opened 2024-05-25
docs(examples): add widget implementation example

joshka:
This new example documents the various ways to implement widgets in
Ratatui. It demonstrates how to implement the `Widget` trait on a type,
a reference, and a mutable reference. It also shows how to use the
`WidgetRef` trait to render boxed widgets.

### DIFF
--- a/examples/widget_impl.rs
+++ b/examples/widget_impl.rs
@@ -204,7 +204,7 @@ impl Widget for &BoxedSquares {
 
 /// `RedSquare` and `BlueSquare` are widgets that render a red and blue square, respectively. They
 /// implement the `WidgetRef` trait instead of the `Widget` trait, which which allows them to be
-/// rendered as boxed widgets. It's not possible to use Widget for this as a dynamic refernce to a
+/// rendered as boxed widgets. It's not possible to use Widget for this as a dynamic reference to a
 /// widget cannot generally be moved out of the box.
 impl WidgetRef for RedSquare {
     fn render_ref(&self, area: Rect, buf: &mut Buffer) {
--- a/examples/widget_impl.rs
+++ b/examples/widget_impl.rs
@@ -1,0 +1,343 @@
+//! # [Ratatui] Widgets implementation examples
+//!
+//! This example demonstrates various ways to implement widget traits in Ratatui on a type, a
+//! reference, and a mutable reference. It also shows how to use the `WidgetRef` trait to render
+//! boxed widgets.
+//!
+//! The latest version of this example is available in the [examples] folder in the repository.
+//!
+//! Please note that the examples are designed to be run against the `main` branch of the Github
+//! repository. This means that you may not be able to compile with the latest release version on
+//! crates.io, or the one that you have installed locally.
+//!
+//! See the [examples readme] for more information on finding examples that match the version of the
+//! library you are using.
+//!
+//! [Ratatui]: https://github.com/ratatui-org/ratatui
+//! [examples]: https://github.com/ratatui-org/ratatui/blob/main/examples
+//! [examples readme]: https://github.com/ratatui-org/ratatui/blob/main/examples/README.md
+use std::{
+    iter,
+    time::{Duration, Instant},
+};
+
+use color_eyre::Result;
+use common::Terminal;
+use crossterm::event::{self, Event, KeyCode};
+use ratatui::{
+    buffer::Buffer,
+    layout::{Constraint, Layout, Rect},
+    style::{Color, Style},
+    text::Span,
+    widgets::{Widget, WidgetRef},
+};
+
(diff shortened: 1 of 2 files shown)


examples/widget_impl.rs
```diff
@@ -0,0 +1,342 @@
...
+//! [Ratatui]: https://github.com/ratatui-org/ratatui
+//! [examples]: https://github.com/ratatui-org/ratatui/blob/main/examples
+//! [examples readme]: https://github.com/ratatui-org/ratatui/blob/main/examples/README.md
+use std::{
+    iter,
+    time::{Duration, Instant},
+};
+
+use color_eyre::Result;
+use common::Terminal;
```
EdJoPaTo, line 25:
```suggestion
use self::common::Terminal;
```

examples/widget_impl.rs
```diff
@@ -0,0 +1,342 @@
...
+    app.run(tui)?;
+    common::restore_terminal()?;
+    Ok(())
+}
+
+#[derive(Default)]
+struct App {
+    should_quit: bool,
+    timer: Timer,
+    #[cfg(feature = "unstable-widget-ref")]
```
EdJoPaTo, line 47: This is already a required feature for this example.

```suggestion
```

(there are multiple cases of this in this PR)

examples/widget_impl.rs
```diff
@@ -0,0 +1,342 @@
...
+    start: Instant,
+}
+
+impl Default for Timer {
+    fn default() -> Self {
+        Self {
+            start: Instant::now(),
+        }
+    }
+}
```
EdJoPaTo, line 154: I don't like defaults that do stuff or have side effects.

```suggestion
impl Timer {
    fn start() -> Self {
        Self {
            start: Instant::now(),
        }
    }
}
```

examples/widget_impl.rs
```diff
@@ -0,0 +1,342 @@
...
+
+/// A widget that renders a blue square.
+struct BlueSquare;
+
+/// This implements the `Widget` trait on a reference to the type. It contains a list of boxed
+/// widgets that implement the `WidgetRef` trait. This is useful for widgets that contain a list of
+/// other widgets that can be different types.
+impl Widget for &BoxedSquares {
+    fn render(self, area: Rect, buf: &mut Buffer) {
+        let constraints = iter::repeat(Constraint::Length(4)).take(self.squares.len());
```
EdJoPaTo, line 197: Not sure that works, but it looks way less intimidating.

```suggestion
        let constraints = vec![Constraint::Length(4); self.squares.len()];
```

EdJoPaTo (commented):
Quickly scrolled over it. Still against color_eyre. Also, not sure whether an example using an unstable feature is a great idea…

examples/widget_impl.rs
joshka: This is an intentional prophylactic against the regular amount "Why doesn't this compile" type questions from newer rust users.

There are multiple paths to get to seeing the examples. Newer users may not be aware of where this is set in the cargo.toml. They may be reading this on github without the context, or have copied the example code into their own projects.

examples/widget_impl.rs
joshka: Instant::now() doesn't have a side effect - it's a pure function which returns the current time.
This enables App::default(), which would otherwise need to be impl App::new().
This also isn't likely something that would trip anyone up in any way.

joshka:
I've heard and understood your concern about color_eyre.

The examples are intentionally structured in a way that helps address common scenarios that people run into in real world building Ratatui apps. I've seen evidence (from reading a significant amount of user application code on GitHub and discord) that users often grab code from the example that most matches their particular need rather than looking at all the relevant examples. Having the necessary error handling in every example helps ensure that regardless of where they grab their code from, they're getting a reasonable basis for building their app.

This is not something I'm willing to be convinced otherwise on as the benefit to the Ratatui ecosystem is much larger than the added complexity. The only thing that would make me rethink this is if there was code that was part of Ratatui that basically did the same things. Regardless, that concern is orthogonal, and until we have something that works for that, the right approach is to harmonize the various examples with boilerplate that's easily replaced by calls to that eventual support module.

kdheepak:
I agree about using `color_eyre` (or `anyhow`) for examples, it's something I suspect a lot of users appreciate (I know I do) and easy to remove for those that know how to do so.

orhun:
What's the status on this? Should we get it over the finish line?

joshka:
I believe this is ready to merge - assuming that we want to document these.
It could do with a VHS tape, but we likely need to go through and make sure all the examples are all up to date at some point, and I wouldn't block on that for this as the concept is less about the display and more about the mechanism.

orhun (approved):

PR #1148, opened 2024-05-26
perf(buffer)!: filled moves the cell to be filled

EdJoPaTo:
This removes a clone from inside the method allowing the user to do it explicitly when needing.

BREAKING CHANGE: `Buffer::filled` moves the cell instead of taking a reference.

```diff
-Buffer::filled(area, &Cell::new("X"));
+Buffer::filled(area, Cell::new("X"));
```

---

https://github.com/ratatui-org/ratatui/pull/1143#issuecomment-2131247181

### DIFF
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -12,6 +12,7 @@ This is a quick summary of the sections below:
 
 - [Unreleased](#unreleased)
   - `Rect::inner` takes `Margin` directly instead of reference
+  - `Buffer::fill` takes `Cell` directly instead of reference
   - `Stylize::bg()` now accepts `Into<Color>`
   - Removed deprecated `List::start_corner`
 - [v0.26.0](#v0260)
@@ -65,6 +66,17 @@ This is a quick summary of the sections below:
      vertical: 0,
      horizontal: 2,
  });
+```
+
+### `Buffer::fill` takes `Cell` directly instead of reference ([#1148])
+
+[#1148]: https://github.com/ratatui-org/ratatui/pull/1148
+
+`Buffer::fill` moves the `Cell` instead of taking a reference.
+
+```diff
+-Buffer::fill(area, &Cell::new("X"));
++Buffer::fill(area, Cell::new("X"));
 ```
 
 ### `Stylize::bg()` now accepts `Into<Color>` ([#1103])
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -12,7 +12,7 @@ This is a quick summary of the sections below:
 
 - [Unreleased](#unreleased)
   - `Rect::inner` takes `Margin` directly instead of reference
-  - `Buffer::fill` takes `Cell` directly instead of reference
+  - `Buffer::filled` takes `Cell` directly instead of reference
   - `Stylize::bg()` now accepts `Into<Color>`
   - Removed deprecated `List::start_corner`
 - [v0.26.0](#v0260)
@@ -68,15 +68,15 @@ This is a quick summary of the sections below:
  });
 ```
 
-### `Buffer::fill` takes `Cell` directly instead of reference ([#1148])
+### `Buffer::filled` takes `Cell` directly instead of reference ([#1148])
 
 [#1148]: https://github.com/ratatui-org/ratatui/pull/1148
 
-`Buffer::fill` moves the `Cell` instead of taking a reference.
+`Buffer::filled` moves the `Cell` instead of taking a reference.
 
 ```diff
--Buffer::fill(area, &Cell::new("X"));
-+Buffer::fill(area, Cell::new("X"));
+-Buffer::filled(area, &Cell::new("X"));
++Buffer::filled(area, Cell::new("X"));
 ```
 
 ### `Stylize::bg()` now accepts `Into<Color>` ([#1103])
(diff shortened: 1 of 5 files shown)


orhun:
Q: Isn't taking a reference better in terms of performance? (no copy & clone?)

EdJoPaTo:
Depends on the case. Moving is always the simplest solution with the best performance. (= moving the ownership, the data stays exactly the same, nothing needs to be done at runtime with the data). When the data is only read it depends on the size of the data (u16 for example is smaller than a pointer on a 64-bit system → its more efficient to copy the u16 instead of using a reference).
When the data is used (written to or whatever) and also required after then method then it needs to be cloned at some point. So either take a reference and clone it inside the method or clone it outside the method.
Cloning it outside the method allows the user to make that choice depending on their use case. Cloning it inside the method forces the user with an additional clone / drop even when it might not be required like in this case.

`Buffer::filled(area, &Cell::new())` resulted basically in this pseudocode:

```
let cell = Cell::new();
Buffer::filled(area, &cell)
drop(cell);
```

In this case we can save the drop and one clone as we can move in the cell directly.

This is not very significant on a Cell as a cell is rather small. With stuff like `String` this can get rather expensive.

Basically… when the method takes a reference and clones it directly, then it could just take the value directly and allow the caller to decide whether cloning or moving is the better choice.

Fun fact: `Into` can hide situations like this resulting in additional clones.
```rust
fn something<S: Into<String>>(input: S) { … }

let huge_string: String = …;
something(&huge_string); // into silently clones in this case
drop(huge_string); // rust silently does this at the end of the context
```
Changing the reference to a move can be a performance improvement (and did for me in the past) which is why I avoid `Into` which is not a pure **in**to and only a to. Especially as .into() might change to a different implementation silently on a refactoring, and it's quite annoying to find the actually used implementation making sure it doesn't do something performance critical.

BREAKING-CHANGES.md
```diff
@@ -12,6 +12,7 @@ This is a quick summary of the sections below:
 
 - [Unreleased](#unreleased)
   - `Rect::inner` takes `Margin` directly instead of reference
+  - `Buffer::fill` takes `Cell` directly instead of reference
```
joshka, line 15:
```suggestion
  - `Buffer::filled` takes `Cell` directly instead of reference
```

BREAKING-CHANGES.md
```diff
@@ -67,6 +68,17 @@ This is a quick summary of the sections below:
  });
 ```
 
+### `Buffer::fill` takes `Cell` directly instead of reference ([#1148])
```
joshka, line 71:
```suggestion
### `Buffer::filled` takes `Cell` directly instead of reference ([#1148])
```

BREAKING-CHANGES.md
```diff
@@ -67,6 +68,17 @@ This is a quick summary of the sections below:
...
+
+[#1148]: https://github.com/ratatui-org/ratatui/pull/1148
+
+`Buffer::fill` moves the `Cell` instead of taking a reference.
```
joshka, line 75:
```suggestion
`Buffer::filled` moves the `Cell` instead of taking a reference.
```

BREAKING-CHANGES.md
```diff
@@ -67,6 +68,17 @@ This is a quick summary of the sections below:
...
+
+```diff
+-Buffer::fill(area, &Cell::new("X"));
++Buffer::fill(area, Cell::new("X"));
```
joshka, line 79:
```suggestion
-Buffer::filled(area, &Cell::new("X"));
+Buffer::filled(area, Cell::new("X"));
```

joshka (changes requested):
Typos: fill -> filled in the change and the commit message / PR

It would be nice to have a succinct rationale added for this to explain why this change makes sense. Not a massive issue as this is not a widely used function.

Probably off-topic to this PR, any idea what the effect of calling `Buffer::filled` with a double width string should be / actually is?

EdJoPaTo:
> Probably off-topic to this PR, any idea what the effect of calling `Buffer::filled` with a double width string should be / actually is?

Likely gibberish, but that's hard to detect as cells with width > 1 are possible and allowed (emojis, …). The ones after the width > 1 cells are skipped I think?

It might be an idea to `Buffer::filled(area, symbol)` directly and set Cell private. The only use case of filled is probably testing.

joshka:
Yeah, I think it might be ok, AAAA AA -> A A A . Every second cell is hidden, but will display correctly regardless I think.

src/buffer/buffer.rs
```diff
@@ -893,25 +893,29 @@ mod tests {
     #[rstest]
     #[case(false, true, [false, false, true, true, true, true])]
     #[case(true, false, [true, true, false, false, false, false])]
-    fn merge_skip(#[case] one: bool, #[case] two: bool, #[case] expected: [bool; 6]) {
-        let mut one = Buffer::filled(
-            Rect {
+    fn merge_skip(#[case] skip_one: bool, #[case] skip_two: bool, #[case] expected: [bool; 6]) {
```
joshka, line 896: Wow, this test is profoundly opaque.

joshka (approved):

src/buffer/buffer.rs
EdJoPaTo: Looking at git blame… They were introduced with #215, so at least that's some context why that stuff exists.

Issue #1141, opened 2024-05-25
`clippy::same_name_method` with Stylize


EdJoPaTo:
## Problem

Stylize introduces style and set_style. They are also implemented manually too.
[`clippy::same_name_method`](https://rust-lang.github.io/rust-clippy/master/index.html#same_name_method) notices this.

## Solution

Probably remove the manually implemented methods and only use the ones implemented on Stylize.
This will be a breaking change as the trait needs to be imported then.

joshka:
To move this ahead, I'd suggest thinking through the impact of this change on downstream code. It would be useful to consider the specific magnitude (how much will be affected) and effort involved in updating the code (i.e. specifically what would the compiler do, what find and replace would help, etc.). I've assumed above that these are likely too large to consider, and this might be correct, but perhaps there's an incremental approach that I'm missing here.

It might also be worth thinking about possible alternatives to this. Something that I've been thinking about is whether it's worth implementing the stylize methods directly on widgets using macros rather than as a blanket trait implementation. There are likely downsides to this approach too.

EdJoPaTo:
Macros would be neat as that can result in `const` methods then. As styles are done many times per render that could be a performance benefit on runtime, but comes at a compile time cost of these macros and their own complexity of writing / maintaining them. I don't think that its worth the effort here.

rust-analyzer should suggest importing traits for used methods so writing them newly shouldn't be a big problem for people using rust-analyzer. Migrating from direct impl to trait seems like a rather simple change, just add the import. Noting that in the breaking changes section should be easy to adapt to. Also, the Rust error output suggests importing the trait too if I remember correctly. Should be checked.

joshka:
> I'd be surprised if the performance gains / losses mentioned are meaningful in any reasonable use case.

Confirmed there is no meaningful difference:

- 0.32s vs 0.31s compile time in a synthetic test of macro vs no macro.
- Single digit pico-seconds difference between benchmarks of trait impl, manual const function, macro generated const function.

PR #1149, opened 2024-05-27
chore(prelude)!: add / remove items

joshka:
Based on a suggestion from @burntsushi on Reddit, this PR removes the
items from the prelude that don't form a coherent common vocabulary and
adds the missing items that do.

> I think that in order for a prelude to be successful, it needs to form
> a coherent common vocabulary.

<https://www.reddit.com/r/rust/comments/1cle18j/comment/l2uuuh7/>

BREAKING CHANGE:
The following items have been removed from the prelude:
- `style::Styled` - this trait is useful for widgets that want to
  support the Stylize trait, but it adds complexity as widgets have two
  `style` methods and a `set_style` method.
- `symbols::Marker` - this item is used by code that needs to draw to
  the `Canvas` widget, but it's not a common item that would be used by
  most users of the library.
- `terminal::{CompletedFrame, TerminalOptions, Viewport}` - these items
  are rarely used by code that needs to interact with the terminal, and
  they're generally only ever used once in any app.

The following items have been added to the prelude:
- `layout::{Position, Size}` - these items are used by code that needs
  to interact with the layout system. These are newer items that were
  added in the last few releases, which should be used more liberally.

### DIFF
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -53,6 +53,53 @@ This is a quick summary of the sections below:
 
 ## Unreleased
 
+### Uncommon items removed from the prelude ([#1149])
+
+The following items have been removed from the prelude:
+
+- `style::Styled` - this trait is useful for widgets that want to
+  support the Stylize trait, but it adds complexity as widgets have two
+  `style` methods and a `set_style` method.
+- `symbols::Marker` - this item is used by code that needs to draw to
+  the `Canvas` widget, but it's not a common item that would be used by
+  most users of the library.
+- `terminal::{CompletedFrame, TerminalOptions, Viewport}` - these items
+  are rarely used by code that needs to interact with the terminal, and
+  they're generally only ever used once in any app.
+
+The following items have been added to the prelude:
+
+- `layout::{Position, Size}` - these items are used by code that needs
+  to interact with the layout system. These are newer items that were
+  added in the last few releases, which should be used more liberally.
+  This may cause conflicts for types defined elsewhere with a similar
+  name.
+
+To update your app:
+
+```diff
+// if your app uses Styled::style() or Styled::set_style():
+-use ratatui::prelude::*;
++use ratatui::{prelude::*, style::Styled};
+
+// if your app uses symbols::Marker:
+-use ratatui::prelude::*;
++use ratatui::{prelude::*, symbols::Marker}
+
+// if your app uses terminal::{CompletedFrame, TerminalOptions, Viewport}
+-use ratatui::prelude::*;
++use ratatui::{prelude::*, terminal::{CompletedFrame, TerminalOptions, Viewport}};
+
+// to disambiguate existing types named Position or Size:
+- use some_crate::{Position, Size};
+- let size: Size = ...;
+- let position: Position = ...;
++ let size: some_crate::Size = ...;
++ let position: some_crate::Position = ...;
+```
+
+[#1149]: https://github.com/ratatui-org/ratatui/pull/1149
+
 ### `Rect::inner` takes `Margin` directly instead of reference ([#1008])
 
 [#1008]: https://github.com/ratatui-org/ratatui/pull/1008
@@ -181,8 +228,6 @@ The following example shows how to migrate for `Line`, but the same applies for 
 ```
 
 ### Remove deprecated `Block::title_on_bottom` ([#757])
-
-[#757]: https://github.com/ratatui-org/ratatui/pull/757
 
 `Block::title_on_bottom` was deprecated in v0.22. Use `Block::title` and `Title::position` instead.
 
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -12,6 +12,7 @@ This is a quick summary of the sections below:
 
 - [Unreleased](#unreleased)
   - `Rect::inner` takes `Margin` directly instead of reference
+  - `Buffer::filled` takes `Cell` directly instead of reference
   - `Stylize::bg()` now accepts `Into<Color>`
   - Removed deprecated `List::start_corner`
 - [v0.26.0](#v0260)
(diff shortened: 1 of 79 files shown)


EdJoPaTo:
> * `style::Styled` - this trait is useful for widgets that want to
>       support the Stylize trait, but it adds complexity as widgets have two
>       `style` methods and a `set_style` method.

#1141: The duplicate methods could also be removed to reduce confusion. Only keeping the trait methods seems like a good thing to me.

BREAKING-CHANGES.md
```diff
@@ -182,8 +229,6 @@ The following example shows how to migrate for `Line`, but the same applies for
 
 ### Remove deprecated `Block::title_on_bottom` ([#757])
 
-[#757]: https://github.com/ratatui-org/ratatui/pull/757
```
EdJoPaTo, line 185: Why remove the link? Even when its duplicated somewhere its directly there to be easily clicked so I would keep it here.

Otherwise: put the full URL into the title?

src/lib.rs
```diff
@@ -336,6 +336,6 @@ pub mod text;
 pub mod widgets;
 
 #[doc(inline)]
-pub use self::terminal::{CompletedFrame, Frame, Terminal, TerminalOptions, Viewport};
+pub use crate::terminal::{CompletedFrame, Frame, Terminal, TerminalOptions, Viewport};
```
EdJoPaTo, line 339: This also works. Could be less or more confusing, depending on who you ask.

```suggestion
pub use terminal::{CompletedFrame, Frame, Terminal, TerminalOptions, Viewport};
```

src/widgets/canvas/line.rs
```diff
@@ -1,5 +1,5 @@
 use crate::{
-    style::Color,
+    prelude::*,
```
EdJoPaTo, line 2: Seems like a step back? It was more specific and now its less specific? And only seemed to use a single item anyway?

EdJoPaTo:
I think Styled is the only thing where I can understand the prelude. It's a common building block to style something. Like `ToString` is part of the std prelude.

(But even that is easily autocompleted by rust-analyzer)

Personally I would look what is used across most files in a ratatui project. Backend isn't part of it, that's a one time thing in a single file → shouldn't be part of the common language.

Layouting, styling, text stuff (Span, Line, …) are something that are very common. And as they are common it's easier to know, they are from ratatui.

Looking at the remaining stuff… why is Masked in there? It's not a thing being used in most files. `symbols` also seem somewhat weird. Having whole modules in there in general seems weird too.

BREAKING-CHANGES.md
joshka: I'm not sure - likely some sort of auto formatting and I didn't notice. Will fix.

I avoid links in the title for shorter lines which are easier to avoid wrapping with in this context. Generally in prose refs rather than inline links lead to more readable markdown. Markdownlint detects missing links. Vscode removes unlinked refs, and highlights links as an error if they are missing corresponding refs. There's likely similar tooling for other editors.

src/lib.rs
joshka: I think I'd like to normalize this to something, but it's lower priority. The self seemed odd in a few places though. I don't have strong feelings about it though.

joshka:
Backend is a trait, which means anything that uses any of the backends needs it. Masked should be removed. Let's hold the discussion of prelude use within the library a bit perhaps? Having modules in the code was an approach that simplified a lot of code when that was added. I actively don't want to make a call on that in this PR which is about removing items from the common vocabulary.

kdheepak:
regarding Stylize trait: 

I think I can be convinced it is not part of the vocabulary but it is certainly one of the more useful parts of the prelude and I would like to not remove it. 

How about putting it in`use ratatui::prelude::traits::*` along with all other traits? If that happens that will be the only prelude that end up in the final version of all my projects.

EdJoPaTo:
> How about putting it in`use ratatui::prelude::traits::*` along with all other traits? If that happens that will be the only prelude that ends up in the final version of all my projects.

Having multiple preludes seems even worse to me.

[`clippy::wildcard_imports`](https://rust-lang.github.io/rust-clippy/master/index.html#/wildcard_imports) explicitly ignores wildcard imports from modules containing the name prelude, so `ratatui::trait_prelude::*` would be a better fit to work together with this lint.
(I think wildcards are similarly annoying for readability and avoid them like I avoid preludes, especially as rust-analyzer can automatically add stuff anyway.)

joshka:
> Having multiple preludes seems even worse to me.

Same. As does naming this anything else other than just prelude. This is a hard push back on the idea of changing the name of the prelude from me in any way.

joshka:
> regarding Stylize trait:
> 
> I think I can be convinced it is not part of the vocabulary but it is certainly one of the more useful parts of the prelude and I would like to not remove it outright.

Stylize is not and should not be removed from the prelude as the methods it adds everywhere **are** intentionally part of the vocabulary of ratatui (.bold(), .green(), etc.)

This PR removes Styled (which allows Widgets to support Stylize through a blanket implementation), not Stylize.

kdheepak:
My bad! Ignore my comment.

src/widgets/canvas/line.rs
joshka: 1051d9db1

src/lib.rs
joshka: 1051d9db1

BREAKING-CHANGES.md
joshka: The removal of this was due to this being a duplicate incorrect link (the other one should have been 759, not 757). Markdown lint removed this as having two link definitions that point at links is potentially ambiguous.

Fixed in 1051d9db1

src/text/grapheme.rs
```diff
@@ -1,4 +1,4 @@
-use crate::prelude::*;
+use crate::{prelude::*, style::Styled};
```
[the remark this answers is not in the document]
joshka: My preference is opposite this. Use an import even when it's a single import, because this leads to generally more readable and succinct code when it's used.  When code is dominated with information about where the items come from, the module path makes it more difficult to quickly scan the words.

src/lib.rs
```diff
@@ -336,17 +348,4 @@ pub mod terminal;
 pub mod text;
 pub mod widgets;
```
EdJoPaTo, line 350: removing this empty line should sort the prelude module into the other modules

EdJoPaTo (approved):

joshka:
@EdJoPaTo any ideas on how to repro the test failures? They're succeeding locally for me.

joshka:
Merging this despite the broken tests to see if there's something that resolves in the github caching.

EdJoPaTo:
They fail locally for me too.

Caches can be removed and then the CI rerun to ensure it's not related to caching.

PR #1151, opened 2024-05-28
feat: re-export backends from the ratatui crate

joshka:
`crossterm`, `termion`, and `termwiz` can now be accessed as
`ratatui::{crossterm, termion, termwiz}` respectively. This makes it
possible to just add the Ratatui crate as a dependency and use the
backend of choice without having to add the backend crates as
dependencies.

To update existing code, replace all instances of `crossterm::` with
`ratatui::crossterm::`, `termion::` with `ratatui::termion::`, and
`termwiz::` with `ratatui::termwiz::`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,10 +43,10 @@ Ratatui was forked from the [tui-rs] crate in 2023 in order to continue its deve
 
 ## Installation
 
-Add `ratatui` and `crossterm` as dependencies to your cargo.toml:
+Add `ratatui` as a dependency to your cargo.toml:
 
 ```shell
-cargo add ratatui crossterm
+cargo add ratatui
 ```
 
 Ratatui uses [Crossterm] by default as it works on most platforms. See the [Installation]
@@ -110,7 +110,8 @@ module] and the [Backends] section of the [Ratatui Website] for more info.
 
 The drawing logic is delegated to a closure that takes a [`Frame`] instance as argument. The
 [`Frame`] provides the size of the area to draw to and allows the app to render any [`Widget`]
-using the provided [`render_widget`] method. See the [Widgets] section of the [Ratatui Website]
+using the provided [`render_widget`] method. After this closure returns, a diff is performed and
+only the changes are drawn to the terminal. See the [Widgets] section of the [Ratatui Website]
 for more info.
 
 ### Handling events
@@ -125,12 +126,17 @@ Website] for more info. For example, if you are using [Crossterm], you can use t
 ```rust
 use std::io::{self, stdout};
 
-use crossterm::{
-    event::{self, Event, KeyCode},
-    terminal::{disable_raw_mode, enable_raw_mode, EnterAlternateScreen, LeaveAlternateScreen},
-    ExecutableCommand,
+use ratatui::{
+    crossterm::{
+        event::{self, Event, KeyCode},
+        terminal::{
+            disable_raw_mode, enable_raw_mode, EnterAlternateScreen, LeaveAlternateScreen,
+        },
+        ExecutableCommand,
+    },
+    prelude::*,
+    widgets::*,
 };
-use ratatui::{prelude::*, widgets::*};
 
 fn main() -> io::Result<()> {
     enable_raw_mode()?;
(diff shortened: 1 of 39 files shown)


kdheepak (approved):

README.md
```diff
@@ -331,17 +330,14 @@ Running this example produces the following output:
 [License Badge]: https://img.shields.io/crates/l/ratatui?style=flat-square&color=1370D3
 [CI Badge]: https://img.shields.io/github/actions/workflow/status/ratatui-org/ratatui/ci.yml?style=flat-square&logo=github
 [CI Workflow]: https://github.com/ratatui-org/ratatui/actions/workflows/ci.yml
-[Codecov Badge]:
-    https://img.shields.io/codecov/c/github/ratatui-org/ratatui?logo=codecov&style=flat-square&token=BAQ8SOKEST&color=C43AC3&logoColor=C43AC3
+[Codecov Badge]:https://img.shields.io/codecov/c/github/ratatui-org/ratatui?logo=codecov&style=flat-square&token=BAQ8SOKEST&color=C43AC3&logoColor=C43AC3
```
EdJoPaTo, line 333:
```suggestion
[Codecov Badge]: https://img.shields.io/codecov/c/github/ratatui-org/ratatui?logo=codecov&style=flat-square&token=BAQ8SOKEST&color=C43AC3&logoColor=C43AC3
```

README.md
```diff
@@ -331,17 +330,14 @@ Running this example produces the following output:
...
 [Codecov]: https://app.codecov.io/gh/ratatui-org/ratatui
 [Deps.rs Badge]: https://deps.rs/repo/github/ratatui-org/ratatui/status.svg?style=flat-square
 [Deps.rs]: https://deps.rs/repo/github/ratatui-org/ratatui
-[Discord Badge]:
-    https://img.shields.io/discord/1070692720437383208?label=discord&logo=discord&style=flat-square&color=1370D3&logoColor=1370D3
+[Discord Badge]:https://img.shields.io/discord/1070692720437383208?label=discord&logo=discord&style=flat-square&color=1370D3&logoColor=1370D3
```
EdJoPaTo, line 337:
```suggestion
[Discord Badge]: https://img.shields.io/discord/1070692720437383208?label=discord&logo=discord&style=flat-square&color=1370D3&logoColor=1370D3
```

EdJoPaTo (approved):

README.md
joshka: The change goes via lib.rs -> readme for this. Fixed

README.md
joshka: fixed

README.md
EdJoPaTo: Then maybe include the docs via macro to prevent the duplication?

joshka: That section of the readme ids generated by Cargo-rdme as it allows writing and running code in the code file. The opposite (writing code in a markdown file) is worse.

src/lib.rs
```diff
@@ -339,3 +340,13 @@ pub mod widgets;
 pub use self::terminal::{CompletedFrame, Frame, Terminal, TerminalOptions, Viewport};
 
 pub mod prelude;
+
+/// re-export the `crossterm` crate so that users don't have to add it as a dependency
```
orhun, line 344: Maybe reword this like the following for making it more user facing rather than a 3rd person comment:

> Re-export of the [crossterm] crate

And also making this a link would be useful too

orhun (approved):
This is all we needed 💯

src/lib.rs
EdJoPaTo: It's not avoiding the import, it's simplifying the Cargo.toml setup as the version reexported will always be the correct one. Not sure how to write that in a relatively short way.

And yeah, a link is pretty simple to add here, but on the other hand its just below… so both variants seem fine to me.

joshka: Neither docs.rs or rust-analyzer show this comment.
- docs.rs shows `crossterm` in the re-exports section, without a comment
- hovering over `ratatui::crossterm` shows the docs for crossterm, not this comment

There is a `#[doc(inline)]` attribute that can be applied, which moves re-exports into where the item would be if it were part of the current module, and uses this comment. E.g. this would move to the modules section, but that's not really appropriate as `crossterm` isn't a submodule of Ratatui, it's much more useful to call it out as what it is - a re-export.

You could also ask why this is a `///` if it doesn't appear in rustdoc, but that seems appropriate here because `///` is generally seen as the right comment to write against an item (rather than `//`).

Issue #1150, opened 2024-05-27
Move away from using Prelude / glob imports


joshka:
## Problem
Based on much feedback from @EdJoPaTo across multiple PRs, as well as seeking external feedback on users.rust-lang.org, reddit, our forum and discord servers, it's time to fix up how we use the prelude.

- https://users.rust-lang.org/t/to-use-prelude-or-to-not-to-use-prelude-that-is-the-question/110855
- https://www.reddit.com/r/rust/comments/1cle18j/to_prelude_or_to_not_to_prelude/
- https://discord.com/channels/1070692720437383208/1070692720437383211/1236238741001994340
- https://forum.ratatui.rs/t/help-us-understand-prelude-usage/33

The overwhelming response gave me the impression at least in code that is used for teaching / communicating that the prelude causes more harm than good. This was counter to my prior understanding of this.

## Solution
- [ ] Summarize the responses more fully (perhaps write a blog post about this)
- [x] Fixup the common vocabulary (https://github.com/ratatui-org/ratatui/pull/1149)
- [x] Remove prelude from `examples/*.rs` (https://github.com/ratatui-org/ratatui/pull/1152)
- [ ] Remove the prelude from visible portions of examples in rustdocs
- [ ] Remove the prelude from tutorial code on the website

joshka:
Added https://github.com/ratatui/ratatui-website/issues/750 for the website side of this

PR #1152, opened 2024-05-28
refactor: expand glob imports

joshka:
Consensus is that explicit imports make it easier to understand the
example code. This commit removes the prelude import from all examples
and replaces it with the necessary imports, and expands other glob
imports (widget::*, Constraint::*, KeyCode::*, etc.) everywhere else.
Prelude glob imports not in examples are not covered by this PR.

See https://github.com/ratatui-org/ratatui/issues/1150 for more details.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -96,6 +96,7 @@ rest_pat_in_fully_bound_structs = "warn"
 string_lit_chars_any = "warn"
 string_slice = "warn"
 string_to_string = "warn"
+unnecessary_self_imports = "warn"
 use_self = "warn"
 
 [features]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -74,7 +74,6 @@ missing_errors_doc = "allow"
 missing_panics_doc = "allow"
 module_name_repetitions = "allow"
 must_use_candidate = "allow"
-wildcard_imports = "allow"
 
 # nursery or restricted
 as_underscore = "warn"
@@ -98,6 +97,7 @@ string_slice = "warn"
 string_to_string = "warn"
 unnecessary_self_imports = "warn"
 use_self = "warn"
+wildcard_imports = "deny"
 
 [features]
 #! The crate provides a set of optional features that can be enabled in your `cargo.toml` file.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -97,7 +97,6 @@ string_slice = "warn"
 string_to_string = "warn"
 unnecessary_self_imports = "warn"
 use_self = "warn"
-wildcard_imports = "deny"
 
 [features]
 #! The crate provides a set of optional features that can be enabled in your `cargo.toml` file.
--- a/examples/constraint-explorer.rs
+++ b/examples/constraint-explorer.rs
@@ -13,8 +13,6 @@
 //! [examples]: https://github.com/ratatui-org/ratatui/blob/main/examples
 //! [examples readme]: https://github.com/ratatui-org/ratatui/blob/main/examples/README.md
 
-#![allow(clippy::enum_glob_use, clippy::wildcard_imports)]
-
 use std::io::{self, stdout};
 
 use color_eyre::{config::HookBuilder, Result};
--- a/examples/constraint-explorer.rs
+++ b/examples/constraint-explorer.rs
@@ -25,11 +25,20 @@ use crossterm::{
 };
 use itertools::Itertools;
 use ratatui::{
-    layout::{Constraint::*, Flex},
-    prelude::*,
-    style::palette::tailwind::*,
-    symbols::line,
-    widgets::{Block, Paragraph, Wrap},
+    backend::{Backend, CrosstermBackend},
+    buffer::Buffer,
+    layout::{
+        Constraint::{self, Fill, Length, Max, Min, Percentage, Ratio},
+        Flex, Layout, Rect,
+    },
+    style::{
+        palette::tailwind::{BLUE, SKY, SLATE, STONE},
+        Color, Style, Stylize,
+    },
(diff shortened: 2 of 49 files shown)


kdheepak (approved):

examples/demo2/main.rs
```diff
@@ -30,16 +30,16 @@ mod tabs;
...
 
-pub use app::*;
+pub use app::App;
 use color_eyre::Result;
-pub use colors::*;
-pub use term::*;
-pub use theme::*;
+pub use colors::{color_from_oklab, RgbSwatch};
+pub use term::{init, restore};
+pub use theme::THEME;
```
EdJoPaTo, line 37: `cargo +nightly fmt` should split these because of them being modules / external… So something seems fishy here.

examples/demo2/tabs/weather.rs
```diff
@@ -1,8 +1,14 @@
 use itertools::Itertools;
 use palette::Okhsv;
 use ratatui::{
-    prelude::*,
-    widgets::{calendar::CalendarEventStore, *},
+    buffer::Buffer,
+    layout::{Constraint, Direction, Layout, Margin, Rect},
+    style::{Color, Style, Stylize},
+    symbols::{self},
```
EdJoPaTo, line 7: https://rust-lang.github.io/rust-clippy/master/index.html#/unnecessary_self_imports
```suggestion
    symbols,
```

EdJoPaTo (changes requested):

examples/demo/ui.rs
```diff
@@ -1,7 +1,16 @@
+use ratatui::widgets::canvas;
 #[allow(clippy::wildcard_imports)]
```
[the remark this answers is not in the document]
joshka: removed

examples/constraint-explorer.rs
```diff
@@ -123,7 +132,7 @@ impl App {
     }
 
     fn handle_events(&mut self) -> Result<()> {
-        use KeyCode::*;
+        use KeyCode::{Char, Down, Esc, Left, Right, Up};
```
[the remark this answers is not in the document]
joshka: I don't mind this - it's good enough and I think it probably suffers the same edge case that can make people wonder what `Char` is. The others are a bit more obvious, but `Char` having the same name as `char` seems enough to disambiguate like this to me.

examples/demo2/tabs/weather.rs
joshka: fixed

examples/demo2/main.rs
joshka: (changed this to `pub use crate::{...}` and removed the unnecessary exports)

examples/demo2/tabs/weather.rs
EdJoPaTo: How about adding that lint? This one is fixed now, the lint will ensure it being fixed in the future too.

The described edge case if this lint sounds like fishy code to me that should be renamed so when this lint results in a naming conflict, that should be fixed rather than keeping `::{self}`

examples/demo2/main.rs
EdJoPaTo: I think self:: would be clearer as it's an example. Yeah, an example is a crate too, but it's less obvious. Reading it from self seems clearer to me.

examples/constraint-explorer.rs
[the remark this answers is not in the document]
EdJoPaTo: The I'd prefer to inline this when this results in the same amount of lines → `match … { KeyCode::Esc =>  }`
Then its even more explicit what it is and also the way its normally written / auto generated by rust-analyzer.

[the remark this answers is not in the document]
joshka: done

examples/demo2/main.rs
joshka: done

examples/demo2/tabs/weather.rs
joshka: done

examples/constraint-explorer.rs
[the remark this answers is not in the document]
joshka: I've left the constraint / layout / flex examples with imports for `Constraint::{...}` because the quantity of usages of Constraint there is overwhelming and obvious.

EdJoPaTo:
Given that there are 4 warnings remaining when enabling (= removing the allow) of wildcard_imports (one in the examples), should it just be enabled in this PR too?

https://github.com/ratatui-org/ratatui/blob/74a32afbaef8851f9462b27094d88d518e56addf/Cargo.toml#L77

joshka:
> wildcard_imports = "allow"

b82ccc28e

Cargo.toml
```diff
@@ -96,7 +95,9 @@ rest_pat_in_fully_bound_structs = "warn"
 string_lit_chars_any = "warn"
 string_slice = "warn"
 string_to_string = "warn"
+unnecessary_self_imports = "warn"
 use_self = "warn"
+wildcard_imports = "deny"
```
EdJoPaTo, line 100: It’s part of pedantic, so already warn. Why deny it?

EdJoPaTo (approved):
This helps a lot with reading code without an IDE :sparkles:

Maybe remove the "all" from the title as prelude::* is still used in the lib itself. Also, this is a "refactor:" as it adapts the source code while maintaining the behavior? :thinking:

joshka:
> This helps a lot with reading code without an IDE ✨
> 
> Maybe remove the "all" from the title as prelude::* is still used in the lib itself. Also, this is a "refactor:" as it adapts the source code while maintaining the behavior? 🤔

done and done. Also fixed a doc lint

Issue #1157, opened 2024-05-31
Consistent examples in ratatui


kdheepak:
## Problem

On discord, we mentioned there was an issue with examples being inconsistent. I wanted to open an issue to track it.

Some of my thoughts are below. I think we need a few different kinds of examples:

1. Examples of features used in a widget 
2. Examples of how to build more complicated applications on top of widgets / patterns one can follow etc
3. Examples of how to structure applications

As an example, compare `examples/paragraph.rs` and `examples/list.rs`. IMO,

- `paragraph.rs` falls under the category of showing features of the `Paragraph` widget
- `list.rs` falls into the category of showing how to build a simple TODO application on top of the `List` widget while showcasing the `List` and `ListState` functionality.

There's a complexity difference between these two categories.

Suggestions:

Perhaps the `list.rs` example works better as a tutorial than an example? 

Examples that fall into category 1 should make it extremely easy for anyone to experiment with different features offered by said widget in Ratatui. To debug a problem on discord, I tried to start with the `list.rs` example but found it was too complicated to experiment with, and it was faster to write the following from scratch:

```rust
use crossterm::{
    event::{self, Event, KeyCode, KeyEventKind},
    execute,
    terminal::{disable_raw_mode, enable_raw_mode, EnterAlternateScreen, LeaveAlternateScreen},
};
use ratatui::{backend::CrosstermBackend, Terminal};

fn main() -> Result<(), color_eyre::eyre::Error> {
    let original_hook = std::panic::take_hook();

    std::panic::set_hook(Box::new(move |panic| {
        disable_raw_mode().unwrap();
        let mut terminal = Terminal::new(CrosstermBackend::new(std::io::stdout())).unwrap();
        execute!(terminal.backend_mut(), LeaveAlternateScreen).unwrap();
        original_hook(panic);
    }));

    let mut terminal = Terminal::new(CrosstermBackend::new(std::io::stdout()))?;
    enable_raw_mode()?;
    execute!(terminal.backend_mut(), EnterAlternateScreen)?;

    loop {
        let mut list_state = ratatui::widgets::ListState::default();

        list_state.select(Some(1));

        terminal.draw(|frame| {
            let area = frame.size();

            let items = ["Item 1", "Item 2", "Item 3"]
                .into_iter()
                .map(|s| ratatui::text::Text::from(s).centered());

            let list = ratatui::widgets::List::new(items)
                .highlight_symbol(" > ")
                .highlight_style(
[...]

PR #1159, opened 2024-05-31
feat(list)!: add list navigation methods (first, last, previous, next)

joshka:
Also cleans up the list example significantly (see also
<https://github.com/ratatui-org/ratatui/issues/1157>)
    
Fixes: <https://github.com/ratatui-org/ratatui/pull/1159>
    
BREAKING CHANGE: The `List` widget now clamps the selected index to the
bounds of the list when navigating with `first`, `last`, `previous`, and
`next`, as well as when setting the index directly with `select`.

### DIFF
--- a/examples/list.rs
+++ b/examples/list.rs
@@ -13,19 +13,19 @@
 //! [examples]: https://github.com/ratatui-org/ratatui/blob/main/examples
 //! [examples readme]: https://github.com/ratatui-org/ratatui/blob/main/examples/README.md
 
-use std::{error::Error, io, io::stdout};
+use std::{error::Error, io};
 
-use color_eyre::config::HookBuilder;
+use crossterm::event::KeyEvent;
 use ratatui::{
-    backend::{Backend, CrosstermBackend},
+    backend::Backend,
     buffer::Buffer,
-    crossterm::{
-        event::{self, Event, KeyCode, KeyEventKind},
-        terminal::{disable_raw_mode, enable_raw_mode, EnterAlternateScreen, LeaveAlternateScreen},
-        ExecutableCommand,
+    crossterm::event::{self, Event, KeyCode, KeyEventKind},
+    layout::{Constraint, Layout, Rect},
+    style::{
+        palette::tailwind::{BLUE, GREEN, SLATE},
+        Color, Modifier, Style, Stylize,
     },
-    layout::{Alignment, Constraint, Layout, Rect},
-    style::{palette::tailwind, Color, Modifier, Style, Stylize},
+    symbols,
     terminal::Terminal,
     text::Line,
     widgets::{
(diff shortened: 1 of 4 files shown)


src/symbols.rs
```diff
@@ -470,6 +470,28 @@ pub mod border {
...
+    pub const EMPTY: Set = Set {
+        top_left: " ",
+        top_right: " ",
+        bottom_left: " ",
+        bottom_right: " ",
+        vertical_left: " ",
+        vertical_right: " ",
+        horizontal_top: " ",
+        horizontal_bottom: " ",
+    };
```
joshka, line 494: This likely needs to go as a different PR.

I needed it because a title on a block without borders set doesn't apply the border style to the line that the title is on. This seems odd, but adding an empty / solid border set makes simple work of that. (I started with solid and then flipped to empty and didn't delete the code.)

src/widgets/list.rs
```diff
@@ -1301,7 +1472,7 @@ mod tests {
             &single_item,
             Some(1),
             [
-                "  Item 0  ",
+                ">>Item 0  ",
```
joshka, line 1475: This is a change that's likely more correct than the existing approach, but is probably a breaking change and should be noted as such.

src/widgets/list.rs
```diff
@@ -156,6 +156,72 @@ impl ListState {
...
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// # use ratatui::{prelude::*, widgets::*};
+    /// let mut state = ListState::default();
+    /// state.select_next();
+    /// ```
+    pub fn select_next(&mut self) {
+        let next = self.selected.map_or(0, |i| i.saturating_add(1));
```
joshka, line 173: It might be a good idea to add support for wrap around here and on prev, but for now I left that out of the PR.

examples/list.rs
joshka, line 1: This is mostly a rewrite

- Terminal related methods moved to the bottom
- pretty much everything rearranged in a more natural reading order (main, structs, App::default, List initialization, App logic, rendering logic
- Simplified the rendering a bunch
- code reduction of ~35 lines (64 if you don't count the small methods to navigate the todo list - these could be inline, but I think it generally makes sense to extract them)

src/widgets/list.rs
a-kenji: Agreed, in my lists I do the same by default. This feels much more natural. If consumers don't implement this manually - for me personally this always leads to confusing interactions.

src/symbols.rs
joshka: https://github.com/ratatui-org/ratatui/pull/1182

src/widgets/list.rs
EdJoPaTo: On my tree widgets I found wrap around to be a bit more confusing. Hitting keys with nothing selected selects something here. Up ends at the top. For someone to move to the extremes, Home and End seem way better.

joshka: There are a variety of use cases where this is reasonable. What I meant was that this should be something that can be enabled by an app rather than something that needs to be part of this PR.

joshka:
Rebased on main, added breaking changes docs, updated commit message and PR comment to note breaking change.

orhun (approved):

errwip:
I don't know if this is the right place to comment, but the clamping is incorect. It clamps to `items.len()` instead of `items.len() - 1`
It doesn't cause any problems for just displaying and navigating the list, BUT if you couple some other part of your code with `list_state.selected()`, you will get index out of bound errors.

My current workaround:
```
let next = self.list_state.selected().map_or(0, |i| i+1);
self.list_state.select(Some(next.min(self.items.len()-1)));
```

joshka:
Great find - mind submitting a PR?

PR #1162, opened 2024-06-03
feat(layout): impl Display for Position and Size

joshka:


orhun (approved):

EdJoPaTo (approved):
I would use `13x37` for a Position too but its probably not that relevant.

As a downside… changing the Display implementation later on is probably a breaking change.

joshka:
> I would use `13x37` for a Position too but its probably not that relevant.
> 
> As a downside… changing the Display implementation later on is probably a breaking change.

Bikeshed comment, but hard disagree. (x, y) is pretty standard for things which are coordinates.

Issue #1160, opened 2024-05-31
index out of bounds: the len is 17526 but the index is 17526


janstarke:
## Description
I display a lot of long lines in a table, and in one specific line I receive the following stacktrace:

```
thread 'main' panicked at /home/jst/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ratatui-0.26.3/src/text/span.rs:380:17:
index out of bounds: the len is 17526 but the index is 17526
stack backtrace:
   0:     0x5d67dafc6ad2 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::hffecb437d922f988
   1:     0x5d67dafea28c - core::fmt::write::hd9a8d7d029f9ea1a
   2:     0x5d67dafc401f - std::io::Write::write_fmt::h0e1226b2b8d973fe
   3:     0x5d67dafc68a4 - std::sys_common::backtrace::print::he907f6ad7eee41cb
   4:     0x5d67dafc83cb - std::panicking::default_hook::{{closure}}::h3926193b61c9ca9b
   5:     0x5d67dafc8123 - std::panicking::default_hook::h25ba2457dea68e65
   6:     0x5d67dafc8a00 - std::panicking::rust_panic_with_hook::h0ad14d90dcf5224f
   7:     0x5d67dafc8742 - std::panicking::begin_panic_handler::{{closure}}::h4a1838a06f542647
   8:     0x5d67dafc6fa6 - std::sys_common::backtrace::__rust_end_short_backtrace::h77cc4dc3567ca904
   9:     0x5d67dafc8474 - rust_begin_unwind
  10:     0x5d67daefc3a5 - core::panicking::panic_fmt::h940d4fd01a4b4fd1
  11:     0x5d67daefc582 - core::panicking::panic_bounds_check::h9bb22f08a42e1ac8
  12:     0x5d67daf3b459 - <ratatui::text::span::Span as ratatui::widgets::WidgetRef>::render_ref::h7d1b172c055f1900
  13:     0x5d67daf3ba85 - ratatui::text::line::render_spans::hd7e85f2853a83213
  14:     0x5d67daf352ac - <ratatui::text::text::Text as ratatui::widgets::WidgetRef>::render_ref::h9c358aaf6d0b008b
  15:     0x5d67daf35007 - <ratatui::text::text::Text as ratatui::widgets::Widget>::render::hc0996a9e0a6c6e20
[...]

PR #1165, opened 2024-06-04
fix(span): ensure that zero-width characters are rendered correctly

joshka:
Zero-width characters at the start of a span are now rendered with the
subsequent character in the same cell. Zero-width characters in the
middle of a span are rendered at the end of the previous cell.

Fixes: https://github.com/ratatui-org/ratatui/issues/1160

### DIFF
--- a/src/buffer/cell.rs
+++ b/src/buffer/cell.rs
@@ -67,6 +67,14 @@ impl Cell {
         self
     }
 
+    /// Appends a symbol to the cell.
+    ///
+    /// This is particularly useful for adding zero-width characters to the cell.
+    pub fn append_symbol(&mut self, symbol: &str) -> &mut Self {
+        self.symbol.push_str(symbol);
+        self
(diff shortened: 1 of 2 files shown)


src/buffer/cell.rs
```diff
@@ -67,6 +67,14 @@ impl Cell {
         self
     }
 
+    /// Appends a symbol to the cell.
+    ///
+    /// This is particularly useful for adding zero-width characters to the cell.
+    pub fn append_symbol(&mut self, symbol: &str) -> &mut Self {
```
EdJoPaTo, line 73: I'm not sure whether I like this complexity being added to this low level building block…

joshka: I can see that point. What about making this `pub(crate)`?

joshka:
Let's get this one merged soon

orhun (approved):

PR #1166, opened 2024-06-04
chore(ci): integrate cargo-semver-checks

orhun:
> [`cargo-semver-checks`](https://github.com/obi1kenobi/cargo-semver-checks): Lint your crate API changes for semver violations.

It is intended to be run _before_ publishing the crate but I don't feel like that is aligned with our workflow since we are already tracking the breaking changes in a separate document. That is why I integrated it into `check-pr.yml` where it is being run for PRs.

I had some experiments with `cargo-semver-checks` before and I think it is pretty useful. I have an open PR where I also improved the workflow a bit: https://github.com/obi1kenobi/cargo-semver-checks-action/pull/65 - in the future we will be able to add comments to the PRs about semver violations.

Let me know what you think!

Inviting @obi1kenobi to the discussion 🐻


.github/workflows/check-pr.yml
```diff
@@ -84,3 +84,15 @@ jobs:
...
+
+  check-semver:
+    name: Check semver
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout the repository
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.head.ref }}
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
```
EdJoPaTo, line 96: Isn't the default of checkout to use the head? Could this be removed with the same outcome → simpler config?

EdJoPaTo (commented):
I like to see cargo-semver-check being integrated!

joshka (approved):
I wonder if it might also be possible to make this ensure that a breaking PR has a !, check that there are changes in the breaking changes doc, and add a breaking change label.

+1 to remove the unnecessary checkout config

obi1kenobi:
👋 Hi all! `cargo-semver-checks` maintainer here, thanks for checking it out.

@orhun is exactly right that the _current_ action is best when used immediately before release, and that a better workflow we want to support in the future is running on each PR. We ran into some security concerns around being able to post comments and set labels safely after scanning an arbitrary PR, which e.g. might contain a malicious `build.rs` script. We obviously don't want a malicious PR to be able to steal a project's GitHub API key and take over.

This is a problem that I'm confident has a solution. I'm currently in the middle of a fundraising push aimed at securing funding to solve this problem and [add several other missing pieces of functionality](https://predr.ag/blog/four-challenges-cargo-semver-checks-has-yet-to-tackle/). Wish me luck! 🤞

Please ping me if you run into any issues and I'd be happy to help! If you find the tool helpful, I also love hearing about that too 😁 And if you are able to support its development, [support on GitHub Sponsors goes a long way](https://github.com/sponsors/obi1kenobi) 🙏

orhun:
> I wonder if it might also be possible to make this ensure that a breaking PR has a !, check that there are changes in the breaking changes doc, and add a breaking change label.

Good idea! I think we can do that in a follow-up PR.

> Hi all! 

Hey @obi1kenobi! 👋🏼 

> I'm currently in the middle of a fundraising push

Best of luck! Very excited to see the future of `cargo-semver-checks`.

EdJoPaTo:
> We ran into some security concerns around being able to post comments and set labels safely after scanning an arbitrary PR, which e.g. might contain a malicious `build.rs` script. We obviously don't want a malicious PR to be able to steal a project's GitHub API key and take over.

Personal opinion: GitHub Actions should only run after approving it as every PR can add things to the `.github/workflows/…` or whatever. (sadly I can not set that automated for all my repos: https://github.com/orgs/community/discussions/35808)

obi1kenobi:
I agree with the sentiment! I use two things in my own repos for this:
- A CODEOWNERS rule that ensures I personally must review any changes to `.github/`: https://github.com/obi1kenobi/cargo-semver-checks/blob/main/CODEOWNERS#L4-L9
- All PRs run only on the `pull_request` trigger and not on `pull_request_target`, which means PR changes to anything inside `.github/` cannot see _my repo's_ secrets etc. — only ones in their own fork instead.
I've found this works well enough for me that I only require approvals for GitHub Actions runs for first-time contributors.

Unfortunately for our desired `cargo-semver-checks` workflow, GitHub currently doesn't allow leaving PR comments or setting labels from a `pull_request` trigger, only from `pull_request_target`. The easiest way I know of working around this is having a piece of infra that _I maintain_ that runs a GitHub App which you authorize to comment on and add labels to your PRs. But obviously, I can't afford to pay out of pocket for thousands of people and companies' use of `cargo-semver-checks`. Hence the funding drive! 🤞

EdJoPaTo:
Can it read the PR title and check whether a `!:` exists or not? Or read the labels?

When it exists, and it seems breaking, then the run can be fine. When it does not exist, and it isn't breaking, then it can be fine too. Only when this misaligns the run could fail in order to correct that title.

obi1kenobi:
The cargo-semver-checks action _by itself_ cannot do that at the moment. But I'd be happy to work with you and @orhun to put something like that together, and once it works well, I'd love to highlight it as a recommended workflow for all users of cargo-semver-checks!

.github/workflows/check-pr.yml
joshka: Ok, getting back to this change - I'd suggest documenting what's happening here a bit better. It's difficult to assume knowledge of what each property here represents for someone reading this, and so a comment would help explain both what this is doing and why it's necessary here.

joshka (approved):
Generally approving, with some docs needed.

joshka:
> Personal opinion: GitHub Actions should only run after approving it as every PR can add things to the `.github/workflows/…` or whatever. (sadly I can not set that automated for all my repos: https://github.com/orgs/community/discussions/35808)

I was under the impression that Github runs the workflows from the main branch, and not the PR itself. You can see that in action in this PR as the check-semver check isn't actually run https://github.com/ratatui-org/ratatui/actions/runs/10186995364. It's possibly I'm misunderstanding this however.

.github/workflows/check-pr.yml
orhun: Added in fdad2c9e16bcc5f473c7e0d13f34079a8fb99ebc

orhun:
Ready to go.

> You can see that in action in this PR as the check-semver check isn't actually run

I thought that's because the PR hasn't landed in `main` yet 🤔

.github/workflows/check-pr.yml
EdJoPaTo: It's still "what is happening here"… Can we simplify that in some way? Maybe by using its own workflow file and `on: pull_request` or something like that which does exactly what we need already requiring less confusing configuration? Not sure whether that's possible?

EdJoPaTo:
Currently, this will be annoyed at every PR after a single breaking change PR is merged into main. So we need to be extra careful not to merge breaking changes until directly before a breaking release, so this can actually be useful in its current state.

.github/workflows/check-semver.yml
```diff
@@ -0,0 +1,23 @@
...
+    steps:
+      - name: Checkout the repository
+        uses: actions/checkout@v4
+        with:
+          # Check out the merge commit as default.
+          # Reference: actions/checkout#checkout-pull-request-head-commit-instead-of-merge-commit
+          ref: ${{ github.event.pull_request.head.ref }}
+          # Check out from the source (forked repository) rather than the destination.
+          # This helps ensure we are working with the correct source code.
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
```
obi1kenobi, line 21: With the `pull_request` target, I think you probably don't need these anymore. The arguments here set the values that are already set by default, I believe.
```suggestion
```

obi1kenobi (commented):
I recommended to @orhun that the cargo-semver-checks run be moved out of the workflow with the `pull_request_target` trigger and into a separate `pull_request`-triggered workflow.

The concern is this:
- `cargo-semver-checks` needs to compile your crates, which means that `build.rs` scripts in your crates and their dependencies execute.
- With the `pull_request_target` trigger, the `GITHUB_TOKEN` passed to workflows is _your repo's_ (`https://github.com/ratatui-org/ratatui`) and not the PR's repo (`https://github.com/someones-fork/ratatui`).
- An attacker that adds a malicious dependency could use a `build.rs` script to steal the `ratatui-org/ratatui` repo's `GITHUB_TOKEN` and abuse it at will.

.github/workflows/check-semver.yml
```diff
@@ -0,0 +1,23 @@
+name: Check Semver
+
+on:
+  pull_request:
+    branches:
+      - main
```
EdJoPaTo, line 6: Is there a benefit of also running that on branch main? Probably not as its only relevant when doing the push then. So I think it's currently the only relevant target? Making sure a breaking PR is getting detected?

orhun: I think it will be sufficient to run it for PRs for now. There is https://github.com/obi1kenobi/cargo-semver-checks-action/pull/65 which will make things easier in the future.

.github/workflows/check-semver.yml
orhun: c0a911b3e92570731e3e7ff632b9dc470ed684f0

EdJoPaTo (approved):

joshka (approved):

obi1kenobi (approved):
This looks good to me. The only other thing I might recommend is allowing this job to fail without blocking CI for now, to allow for breaking changes to be merged without _immediately_ releasing a new major version.

Like @orhun pointed out, creating a variant of the `cargo-semver-checks` GitHub Action that is more PR-friendly instead of the current release-optimized one is a near-term TODO item for me.

orhun:
> The only other thing I might recommend is allowing this job to fail without blocking CI for now, 

The last time I checked there wasn't an option for it: https://github.com/actions/runner/issues/2347 (well, there is but AFAIK it is not shown as "failed" in the UI)

Do you know any workarounds for it?

obi1kenobi:
Oh I just meant not marking it as [a required check](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/managing-a-branch-protection-rule) in a branch protection rule.

I think you'd still want it to look "failed" in the UI so that you know to inspect it when it first breaks.

orhun:
Got it, we all good then!

EdJoPaTo:
For a PR `--baseline-rev=upstream/main` should be useful.

When the PR has a label or `!` in the title it could use `--release-type=major` but that's something that should not be part of the action itself, only some input of release-type.

PR #1167, opened 2024-06-04
feat(text): support constructing `Line` and `Text` from `usize`

orhun:
Now you can create `Line` and `Text` from numbers like so:

```rust
let line = Line::from(42);
let text = Text::from(666);
```

(I was doing little testing for my TUI app and saw that this isn't supported - then I was like WHA and decided to make it happen :tm:)


orhun:
I also looked into using something more generic (like `Num` trait from [`num-traits`](https://docs.rs/num-traits/)) but felt like an overkill at the end.

joshka:
This doesn't seem like something that would be expected idiomatically in rust.
- `String::from(num)` doesn't exist
- `Cow<str>::from(num)` also doesn't

I'd expect to generally use `format!` for this rather than adding conversions from random things to strings.

Perhaps this could be implemented as traits: `ToSpan` / `ToLine` / `ToText` if we're going for something more aligned with existing rust idioms, and then those would be implemented on num etc.

EdJoPaTo:
What about creating a Span from every Display implementation?

joshka:
> What about creating a Span from every Display implementation?

I like the idea - I'd be concerned about whether this is too magical for newer users. Any ideas about whether there could be other downsides of this?

orhun:
> This doesn't seem like something that would be expected idiomatically in rust.

Yeah, I see your point & makes sense. I'm happy to take the `ToLine` approach here.

> What about creating a Span from every Display implementation?

Can you elaborate this? I didn't quite get what this would look like.

joshka:
> Can you elaborate this? I didn't quite get what this would look like.

perhaps either:
```rust
impl<T:Display> From<T> for Span { ... }
// or
impl <T: Display> ToSpan for T { ... }
```

I suspect the first approach might clash with other existing / future generic implementations, so it's something to be careful with. It's a reasonable idea that's worth exploring, but I'd want to see how it pans out in real use cases to evaluate whether it really works well.

EdJoPaTo:
Maybe a more explicit `Span::from_display<T: Display>(input: T)` might be helpful for readable and understandable code.

joshka:
> Maybe a more explicit `Span::from_display<T: Display>(input: T)` might be helpful for readable and understandable code.

The ToSpan trait idea is more in line with ToString in std.

https://doc.rust-lang.org/std/string/struct.String.html#impl-ToString-for-T

```rust
impl<T: fmt::Display + ?Sized> ToString for T { ... }
```

So (copying wording etc. from std too):

```rust
/// A trait for converting a value to a [`Span`].
///
/// This trait is automatically implemented for any type that implements the [`Display`] trait. As
/// such, `ToSpan` shouln't be implemented directly: [`Display`] should be implemented instead, and
/// you get the `ToSpan` implementation for free.
///
/// [`Display`]: std::fmt::Display
pub trait ToSpan {
    fn to_span(&self) -> Span<'_>;
}

/// # Panics
///
/// In this implementation, the `to_span` method panics if the `Display` implementation returns an
/// error. This indicates an incorrect `Display` implementation since `fmt::Write for String` never
/// returns an error itself.
impl<T: fmt::Display> ToSpan for T {
    fn to_span(&self) -> Span<'_> {
        Span::raw(self.to_string())
    }
}
```

```rust
#[test]
fn test_num() {
    assert_eq!(42.to_span(), Span::raw("42"));
}
```

... and ToLine / ToText follow pretty nicely.

orhun:
That looks good! Implemented in bdb2096c7af75360293f86e13546ed1a631eb382

However, this is not exactly what I wanted to see:

```diff
-let text = Text::from(42);
+let text = Text::from(42.to_span());
```

To take the value directly I'm guessing we need `impl<T: Display> for Text` but that conflicts with the other implementations.

joshka:
> However, this is not exactly what I wanted to see:
> 
> ```diff
> -let text = Text::from(42);
> +let text = Text::from(42.to_span());
> ```

What about adding `ToLine` and `ToText`?

```rust
let text = 42.to_text();
```

kdheepak:
I like the `.to_span()` idea but I don't think we should do `.to_line()` and `.to_text()` because it is non intuitive that 

- `42.to_line()` becomes `[Span::new(42.to_string())]` and 
- `42.to_text()` becomes `[ [Span::new(42.to_string())] ]`

I'd prefer upstreaming the macros from `ratatui-macros` after adding a `.to_span()` method. Then users can write:

```rust
span!(42)
// or
line![42]
// or
text![42]
```

joshka:
> I like the `.to_span()` idea but I don't think we should do `.to_line()` and `.to_text()` because it is non intuitive that
> 
>     * `42.to_line()` becomes `[Span::new(42.to_string())]` and
> 
>     * `42.to_text()` becomes `[ [Span::new(42.to_string())] ]`

They would create a Line and a Text, not arrays. I think these extra traits seem to add some consistency. In particular, if you know that the Display implementation of something is multi-line, then you should expect to be able to convert that to text, not to a span / line.

> I'd prefer upstreaming the macros from `ratatui-macros` after adding a `.to_span()` method. Then users can write:
> 
> ```rust
> span!(42)
> // or
> line![42]
> // or
> text![42]
> ```

This is not necessarily an alternative - wouldn't it make sense to do both?

joshka:
> `to_line` seems pointless, as a Line consists of multiple Spans with different styling. 

As a direct conversion perhaps not, but as an intermediate value where a line is composed, or just as a more obviously explicit call where a line is expected
```rust
let mut line = 42.to_line();
line.push_span("%".red());

let mut text = Text::from("asdf");
text.push_line(42.to_line());
```

I definitely agree that ToSpan and ToText have strong reasons to belong. ToLine is a weaker reason, but still useful (and for consistency this makes sense). There's no obvious downside to this.

Handling newlines should be the same as for Line - they get swallowed afaik (noting that there might be some ways to get newline characters in there).

If there is no to_line then the latter might be written as:
```rust
text.push_line(42.to_span());
// which does something completely different (but not obviously so) to:
text.push_span(42.to_span());
```

EdJoPaTo:
> ```rust
> let mut line = 42.to_line();
> line.push_span("%".red());
> ```

When the length is already known it's also more performant to write it differently:

```rust
let line = Line::from([42.to_span(), "%".red()];
```

`to_line` would only be a convenience thing, but a single span should always work as a line, so `to_line` would be pointless.

---

> If there is no to_line then the latter might be written as: […]

As long as the concept of Span, Line, and Text are clear, this isn't confusing. `push_line` accepts `Into<Line>` so `push_line(42.to_span())` should work as expected.

When the concept isn't clear, `Into` is introducing confusing behavior. So these three thingies should be explained well enough, then I don't see an issue here.

joshka:
> > ```rust
> > let mut line = 42.to_line();
> > line.push_span("%".red());
> > ```
> 
> When the length is already known it's also more performant to write it differently:
> ```rust
> let line = Line::from([42.to_span(), "%".red()];
> ```
> 
> `to_line` would only be a convenience thing, but a single span should always work as a line, so `to_line` would be pointless.
> 
> > If there is no to_line then the latter might be written as: […]
> 
> As long as the concept of Span, Line, and Text are clear, this isn't confusing. `push_line` accepts `Into<Line>` so `push_line(42.to_span())` should work as expected.
> 
> When the concept isn't clear, `Into` is introducing confusing behavior. So these three thingies should be explained well enough, then I don't see an issue here.

Ed, you're straw-manning the argument. The examples are intentionally simplified. Try strong-manning instead and see if you're able to find some examples that do make sense for ToLine to exist. Particularly try taking on the new rust user perspective.

kdheepak:
> They would create a Line and a Text, not arrays. 

My mental model is that we have `Span`, `const Line = Vec<Span>`, `const Text = Vec<Line>`.

> In particular, if you know that the Display implementation of something is multi-line, then you should expect to be able to convert that to text, not to a span / line.

Having something that can convert multiple lines separated by `\n` in the display output has me convinced of `.to_text()`. I'm indifferent towards `.to_line()`. Maybe just for the sake on consistency it should exist?

joshka:
Thinking about this more, I think there's a bit of a problem with using the Display for any of this - what if you'd want to put styling info directly on the type you're dealing with instead of external to the type. Let's take an example like:
```rust
struct MarkdownHeading {
    content: String,
    ...
}

impl fmt::Display for MarkdownHeading {
        write!(f, "{}", content)
    }
}

// but what if I want this to be always styled

impl ToLine for MarkdownHeading {
    fn to_line(&self) -> Line {
        content.blue().bold().into()
    }
}
```

The same could be said for Span and Text easily - if we implement ToXXX for Display, then we prevent the ability for types to implement their own conversion to Span/Line/Text through these methods. Perhaps these traits might be better named ToRawSpan, ToRawLine, ToRawText?

> My mental model is that we have `Span`, `const Line = Vec<Span>`, `const Text = Vec<Line>`.

Ah, I see. Mine is that these are all just `String`s, but with formatting and alignment. The internal representation should be treated as an implementation detail.

Where I really do like this idea is that it provides a more natural way to compose text/line oriented things (like the markdown example, or writing pretty printed JSON). If you have a container of things which implement `trait MyCustomThing: ToLine`, then you can iterate over that for rendering nicely. In https://github.com/joshka/ratatui-json-tutorial/ I was trying to work out ways to convert the json tree into a tree of widgets, but got stuck thinking about how to keep track of where each element widget starts and finishes on a line. In this situation, switching to treating them as being convertible to individual lines / spans seems a logical step.

matta:
> Thinking about this more, I think there's a bit of a problem with using the Display for any of this - what if you'd want to put styling info directly on the type you're dealing with instead of external to the type.

It is probably better to talk about doing this for the `std::string::ToString` trait instead of `std::fmt::Display` trait, since at least in that case it is clear there are zero available formatting options.

But I agree with you about the bit of a problem.

I usually think it is a good practice to define a trait/protocol/interface next to its consumer. That is the case with `std::fmt::Display`: it is defined in `std::fmt` for the purpose of `std::fmt::format!`.

The benefit of this approach is that the two can be designed together. For example, `std::fmt::Display` implementations are given a [Formatter](https://doc.rust-lang.org/std/fmt/struct.Formatter.html), which has all sorts of stuff related to formatting numbers, debug strings, etc., in support formatting options set on the format string. By omitting the format string, and just calling `Display`, flexibility is lost. At the same time, stuff related to rendering into terminals is absent (colors, underline, blink, cursor position, etc.).

It might be convenient to take advantage of a type's `std::fmt::Display` trait for rendering in Ratatui (or other UIs), but that is perhaps done in an opt-in way. A pretty clear, obvious way is to require a call to `format!`, which has the advantage of allowing the above-mentioned `std::fmt` options to be set on the `Formatter` passed to `Display::fmt`.

...and if talking about supporting the `ToString` trait, how about just having application code call `thing.to_string()` instead?

> Where I really do like this idea is that it provides a more natural way to compose text/line oriented things (like the markdown example, or writing pretty printed JSON). If you have a container of things which implement `trait MyCustomThing: ToLine`, then you can iterate over that for rendering nicely. In https://github.com/joshka/ratatui-json-tutorial/ I was trying to work out ways to convert the json tree into a tree of widgets, but got stuck thinking about how to keep track of where each element widget starts and finishes on a line. In this situation, switching to treating them as being convertible to individual lines / spans seems a logical step.

Analogous to `std::fmt` that would be a trait under the `ratatui::` module for Ratatui's purposes. Makes a lot of sense to me.

orhun:
I personally like the idea of having `ToXyz` implementations for `Line`, `Span` and `Text`. I implemented that in my latest commit along with some tests.

I'm happy to rename it to `ToRawXyz`, but I think having this in the first place is a good start.

> what if you'd want to put styling info directly on the type you're dealing with instead of external to the type. 

Can't we export the trait and let the users implement it for their own types? Also, we should think about how common that use-case is too.

matta:
>> what if you'd want to put styling info directly on the type you're dealing with instead of external to the type.
>
> Can't we export the trait and let the users implement it for their own types? Also, we should think about how common that use-case is too.

See https://github.com/rust-lang/rfcs/blob/master/text/0565-show-string-guidelines.md#user-facing-fmtdisplay, which talks about how it is expected that application code will use a different type (most likely a newtype) for each variation of `Display` behavior they want.

joshka:
> >> what if you'd want to put styling info directly on the type you're dealing with instead of external to the type.
> 
> >
> 
> > Can't we export the trait and let the users implement it for their own types? Also, we should think about how common that use-case is too.
> 
> 
> 
> See https://github.com/rust-lang/rfcs/blob/master/text/0565-show-string-guidelines.md#user-facing-fmtdisplay, which talks about how it is expected that application code will use a different type (most likely a newtype) for each variation of `Display` behavior they want.

I think what you're saying is if the widget implements display, and that's not what we want to be rendered, then it's easy enough to do `MyWidget::for_whatever().to_text()` etc., so we shouldn't worry too much about the blanket Display implementation being a blocker.

@orhun we can export the trait, but my point was that you can't implement it twice on the one type, and implementing it as a blanket implementation for every type that implements display would be the first implementation in some cases. I'd guess this is probably rare enough that it shouldn't matter too much though (and the workaround above is fine).

src/text/text.rs
```diff
@@ -557,6 +557,12 @@ impl<'a> From<Vec<Line<'a>>> for Text<'a> {
     }
 }
 
+impl<'a> From<usize> for Text<'a> {
+    fn from(v: usize) -> Self {
+        Self::raw(v.to_string())
+    }
+}
+
```
joshka, line 565: Remove...
```suggestion
```

src/text/text.rs
```diff
@@ -580,6 +586,36 @@ where
...
+            .to_string()
+            .lines()
+            .map(|line| Line::from(line.to_string()))
+            .collect();
+        Text {
+            lines,
+            ..Default::default()
+        }
+    }
+}
```
joshka, line 617: Can this call one of the text constructors instead of duplicating it? Otherwise this may lead to divergent implementations in future.

joshka (approved):

src/text/line.rs
```diff
@@ -646,6 +646,28 @@ fn spans_after_width<'a>(
...
+}
+
+/// # Panics
+///
+/// In this implementation, the `to_line` method panics if the `Display` implementation returns an
+/// error. This indicates an incorrect `Display` implementation since `fmt::Write for String` never
+/// returns an error itself.
+impl<T: fmt::Display> ToLine for T {
+    fn to_line(&self) -> Line<'_> {
+        Line::from(self.to_string())
```
matta, line 667: This is more a question about idioms than anything else, but why not implement `ToLine` for all `string::ToString` rather than all `fmt::Display`? All this does is call to_string() anyway.

Trying to answer my own question:

1. The docs for `ToString` say don't implement `ToString`, implement `Display`. Counterpoint: that doesn't mean API's shouldn't accept `ToString` when all they do is call `to_string()` on the thing.
2. std::fmt can work with no_std, but std::string can't. Counterpoint: not an issue for Ratatui?
3. There are subtle cultural differences in expectation between `Display` and `ToString`. Counterpoint: I haven't found any.  ;-)

src/text/text.rs
orhun: d3f786567fa6de51a02600a03c9b509823caa9aa

src/text/text.rs
orhun: d3f786567fa6de51a02600a03c9b509823caa9aa

src/text/line.rs
orhun: 1st and 3rd answers makes the most sense to me and that's probably what I'd come up with too here.

orhun:
Addressed the review comments and rebased on main. Now the only issue is:

```
warning: method `to_line` is never used
warning: method `to_span` is never used
warning: method `to_text` is never used
```

Should we come up places to use this new mechanism somewhere in the code or just slap `dead_code` on it? What do you think @ratatui-org/maintainers

matta:
> Now the only issue is:
> 
> ```
> warning: method `to_line` is never used
> warning: method `to_span` is never used
> warning: method `to_text` is never used
> ```
> 
> Should we come up places to use this new mechanism somewhere in the code or just slap `dead_code` on it? What do you think @ratatui-org/maintainers

Nothing `pub` should warn about being unused. A `pub` thing counts as being (possibly) used by the users of the crate. 

It took me a while to figure this out, but it turns out that the new traits were not actually pub outside Ratatui. Try the patch below (which causes new warnings about no doc comment for the trait methods):

```diff
diff --git a/src/text.rs b/src/text.rs
index 6262ca6..92aad3a 100644
--- a/src/text.rs
+++ b/src/text.rs
@@ -49,13 +49,16 @@ pub use grapheme::StyledGrapheme;

 mod line;
 pub use line::Line;
+pub use line::ToLine;

 mod masked;
 pub use masked::Masked;

 mod span;
 pub use span::Span;
+pub use span::ToSpan;

 #[allow(clippy::module_inception)]
 mod text;
 pub use text::Text;
+pub use text::ToText;
```

src/text/line.rs
joshka: I suspect that it's rare to see `<T: ToString>`, though I'm not sure which is correct. The thing to note is that if the ToString case is correct on seeing this in the real world, then changing the blanket implementation to be against ToString is a 100% compatible change for all callers, the other direction is not however.

joshka (commented):
Check clippy

orhun:
Thanks for the pointers guys, it should be good to go now!

joshka (approved):

PR #1171, opened 2024-06-06
fix: avoid unicode-width breaking change in tests

joshka:
unicode-width 0.1.13 changed the width of \u{1} from 0 to 1.
Our tests assumed that \u{1} had a width of 0, so this change replaces
the \u{1} character with \u{200B} (zero width space) in the tests.

Upstream issue (closed as won't fix):
https://github.com/unicode-rs/unicode-width/issues/55


orhun (approved):

orhun:
this is kinda annoying lol

PR #1172, opened 2024-06-07
feat(style): add conversions from the palette crate colors

joshka:
This is behind the "palette" feature flag.

```rust
use palette::{LinSrgb, Srgb};
use ratatui::style::Color;

let color = Color::from(Srgb::new(1.0f32, 0.0, 0.0));
let color = Color::from(LinSrgb::new(1.0f32, 0.0, 0.0));
```

### DIFF
--- a/src/style.rs
+++ b/src/style.rs
@@ -73,9 +73,9 @@ use std::fmt;
 use bitflags::bitflags;
 pub use color::{Color, ParseColorError};
 pub use stylize::{Styled, Stylize};
-pub mod palette;
 
 mod color;
+pub mod palette;
 #[cfg(feature = "palette")]
 mod palette_conversion;
 mod stylize;
--- a/src/style.rs
+++ b/src/style.rs
@@ -71,13 +71,14 @@
 use std::fmt;
 
 use bitflags::bitflags;
-
-mod color;
-mod stylize;
-
 pub use color::{Color, ParseColorError};
 pub use stylize::{Styled, Stylize};
 pub mod palette;
+
+mod color;
+#[cfg(feature = "palette")]
+mod palette_conversion;
+mod stylize;
 
 bitflags! {
     /// Modifier changes the way a piece of text is displayed.
--- a/src/style/palette_conversion.rs
+++ b/src/style/palette_conversion.rs
@@ -61,7 +61,7 @@ mod tests {
         const RED: Color = Color::Rgb(255, 0, 0);
         assert_eq!(Color::from(Srgb::new(255u8, 0, 0)), RED);
         assert_eq!(Color::from(Srgb::new(65535u16, 0, 0)), RED);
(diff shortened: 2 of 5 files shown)


src/style.rs
```diff
@@ -71,14 +71,15 @@
...
 use bitflags::bitflags;
-
-mod color;
-mod stylize;
-
 pub use color::{Color, ParseColorError};
 pub use stylize::{Styled, Stylize};
 pub mod palette;
 
+mod color;
```
EdJoPaTo, line 78: Move the mod into the mod block out of the use block. (Then use `cargo +nightly fmt` to order them)

```suggestion
pub use stylize::{Styled, Stylize};

pub mod palette;
mod color;
```

src/style/palette_conversion.rs
```diff
@@ -0,0 +1,83 @@
...
+            Color::from(Srgb::new(0.5f32, 0.5, 0.5)),
+            Color::Rgb(128, 128, 128)
+        );
+    }
+
+    #[test]
+    fn from_lin_srgb() {
+        const RED: Color = Color::Rgb(255, 0, 0);
+        assert_eq!(Color::from(LinSrgb::new(1.0f32, 0., 0.)), RED);
+        assert_eq!(Color::from(LinSrgb::new(1.0f64, 0., 0.)), RED);
```
EdJoPaTo, line 76: Numbers with dots in the end look weird. Clarify it's a floating point `0.0`. (There are more than this)
```suggestion
        assert_eq!(Color::from(LinSrgb::new(1.0f32, 0.0, 0.0)), RED);
        assert_eq!(Color::from(LinSrgb::new(1.0f64, 0.0, 0.0)), RED);
```

EdJoPaTo:
Given that the terminals are reexported, should this be reexported too (when the feature flag is given)?

joshka:
> Given that the terminals are reexported, should this be reexported too (when the feature flag is given)?

There's no public code to export - just trait impls.

EdJoPaTo:
> > Given that the terminals are reexported, should this be reexported too (when the feature flag is given)?
> 
> There's no public code to export - just trait impls.

`Srgb` for example? Currently, this feature needs to be enabled and palette be added as dependency to your crate to use it?

joshka:
> > > Given that the terminals are reexported, should this be reexported too (when the feature flag is given)?
> > 
> > 
> > There's no public code to export - just trait impls.
> 
> `Srgb` for example? Currently, this feature needs to be enabled and palette be added as dependency to your crate to use it?

Ah, gotcha. I'd expect that the only people wanting to turn this on are probably people that are specifically using palette already. Also we have a palette module already (so this could easily cause confusion).

orhun (approved):
I think I saw someone using `palette` colors in one of the [videos](https://ratatui.rs/tutorials/videos/) before. So this will definitely be useful!

EdJoPaTo (approved):

PR #1175, opened 2024-06-10
chore(ci): check GitHub Action workflow files via actionlint

orhun:
https://github.com/rhysd/actionlint

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,6 +46,10 @@ jobs:
         uses: crate-ci/typos@master
       - name: Lint dependencies
         uses: EmbarkStudios/cargo-deny-action@v1
+      - name: Check workflow files
+        uses: docker://rhysd/actionlint:latest
+        with:
+          args: -color
 
   clippy:
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


joshka (approved):

.github/workflows/ci.yml
```diff
@@ -46,6 +46,10 @@ jobs:
         uses: crate-ci/typos@master
       - name: Lint dependencies
         uses: EmbarkStudios/cargo-deny-action@v1
+      - name: Check workflow files
+        uses: docker://rhysd/actionlint:latest
```
EdJoPaTo, line 50: Not sure its really useful to keep this always running. CI is changing not that often. A link to the online check tool might be more relevant.

orhun: I think it's fine, it's not much of a resource heavy thing anyways.

joshka: Perhaps gate it on the path - not sure I like the docker approach to running something like this, but it seems like it takes 0 seconds to run, so not too much of a concern I guess.

orhun: There are other ways of using this as well: <https://github.com/rhysd/actionlint/blob/main/docs/usage.md>

The docker approach looked safer than yolo-downloading a script and running it so I went with that

orhun:
Going to close this one, it's something very trivial.

PR #1177, opened 2024-06-11
fix: render of &str and String doesn't respect area.width

thscharler:
Those two shouldn't write outside the given area I would think.

### DIFF
--- a/src/widgets.rs
+++ b/src/widgets.rs
@@ -52,6 +52,7 @@ pub use self::{
     table::{Cell, HighlightSpacing, Row, Table, TableState},
     tabs::Tabs,
 };
+use crate::style::Style;
 use crate::{buffer::Buffer, layout::Rect};
 
 /// A `Widget` is a type that can be drawn on a [`Buffer`] in a given [`Rect`].
@@ -444,13 +445,7 @@ impl Widget for &str {
 /// [`Rect`].
 impl WidgetRef for &str {
     fn render_ref(&self, area: Rect, buf: &mut Buffer) {
-        buf.set_stringn(
-            area.x,
-            area.y,
-            self,
-            area.width as usize,
-            crate::style::Style::default(),
-        );
+        buf.set_stringn(area.x, area.y, self, area.width as usize, Style::new());
     }
 }
 
@@ -471,13 +466,7 @@ impl Widget for String {
 /// without the need to give up ownership of the underlying text.
 impl WidgetRef for String {
     fn render_ref(&self, area: Rect, buf: &mut Buffer) {
-        buf.set_stringn(
-            area.x,
-            area.y,
-            self,
-            area.width as usize,
-            crate::style::Style::default(),
-        );
+        buf.set_stringn(area.x, area.y, self, area.width as usize, Style::new());
     }
 }
 
--- a/src/widgets.rs
+++ b/src/widgets.rs
@@ -444,7 +444,13 @@ impl Widget for &str {
 /// [`Rect`].
 impl WidgetRef for &str {
     fn render_ref(&self, area: Rect, buf: &mut Buffer) {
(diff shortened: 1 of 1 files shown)


src/widgets.rs
```diff
@@ -444,7 +444,13 @@ impl Widget for &str {
 /// [`Rect`].
 impl WidgetRef for &str {
     fn render_ref(&self, area: Rect, buf: &mut Buffer) {
-        buf.set_string(area.x, area.y, self, crate::style::Style::default());
+        buf.set_stringn(
+            area.x,
+            area.y,
+            self,
+            area.width as usize,
+            crate::style::Style::default(),
```
EdJoPaTo, line 452: I am curious why it's not imported in the file with a use and rather used with a full path here.

(Also I would go with the `const fn` `Style::new()` but thats not important for this PR.)

EdJoPaTo (approved):

src/widgets.rs
thscharler: I just left that as I found it. Can change it if you want.

EdJoPaTo: I am mainly curious why it was that way to begin with… What the intention of that was.

thscharler: Don't think I can help you with that one. 

Changed it, looks nicer now as it's one row again.

matta:
Personally, I'm not a big fan of str implementing Widget. There are many ways to interpret the intent, and it isn't clear what it'll actually do. Because of the way traits show up in rust docs, I'm not sure how to document the behavior.

Should a str Widget behave like a Span with no styling? Or a multi-line thing? What if the Str has newlines? Should the rendering word wrap within the area? It would be good to document the answer.

If a str Widget is basically a Span with no styling it would be nice to find a way to share more of this implementation with `Span`. See @joshka 's commit yesterday which fixes rendering zero-width characters -> d370aa7. Does that bug apply to Str widgets too? If this is a case it might be better to remove the str Widget impl and instead have an `UnstyledSpan` widget. People can then write `draw(UnstyledSpan(my_str))` if they want to.

src/widgets.rs
EdJoPaTo: (I mainly asked that question towards other @ratatui-org/maintainers)

thscharler:
> Personally, I'm not a big fan of str implementing Widget. There are many ways to interpret the intent, and it isn't clear what it'll actually do. Because of the way traits show up in rust docs, I'm not sure how to document the behavior.
> 
> Should a str Widget behave like a Span with no styling? Or a multi-line thing? What if the Str has newlines? Should the rendering word wrap within the area? It would be good to document the answer.

I stumbled upon this implementation, and just tried it. Personally I wouldn't have expected it to handle newlines in any way and form. 

But I agree that the behaviour should be documented. 


> 
> If a str Widget is basically a Span with no styling it would be nice to find a way to share more of this implementation with `Span`. See @joshka 's commit yesterday which fixes rendering zero-width characters -> [d370aa7](https://github.com/ratatui-org/ratatui/commit/d370aa75af99da3e0c41ceb28e2d02ee81cd2538). Does that bug apply to Str widgets too? If this is a case it might be better to remove the str Widget impl and instead have an `UnstyledSpan` widget. People can then write `draw(UnstyledSpan(my_str))` if they want to.

Probably `Buffer::set_string_n()` should use the same algorithm as Span. 

I looked at that one and wondered why `graphemes(true)` doesn't handle that case. As far as I understand <http://www.unicode.org/reports/tr29/#Extend> it shouldn't break at those zero width joiners. Or was it just a case of invalid data?

src/widgets.rs
joshka: Likely just an autocomplete that wasn't noticed.

orhun (approved):
This LGTM, don't see reason why we shouldn't have it.

PR #1180, opened 2024-06-16
feat(backend): Add convenience methods for setting up terminals

joshka:
Adds:
- `CrosstermBackend::stdout` and `CrosstermBackend::stderr` for creating
  backends with `std::io::stdout` and `std::io::stderr` respectively.
- `CrosstermBackend::stdout_with_defaults` and
  `CrosstermBackend::stderr_with_defaults` for creating
  backends with that use the alternate screen and raw mode.
- `CrosstermBackend::with_raw_mode` for enabling raw mode.
- `CrosstermBackend::with_alternate_screen` for switching to the
  alternate screen.
- `CrosstermBackend::with_mouse_capture` for enabling mouse support.
- `CrosstermBackend::reset` for resetting the terminal to its default
  state.
- `Drop` implementation for restoring raw mode, mouse capture, and
  alternate screen on drop.

Most applications should generally use `stdout_with_defaults` to create
a backend with raw mode and the alternate screen, and now longer need
to worry about restoring the terminal to its default state when the
application exits. The reset method can be used to restore the terminal to
its default state if needed (e.g. in a panic handler).

```rust
use ratatui::{backend::CrosstermBackend};

let backend = CrosstermBackend::stdout_with_defaults()?;
// or
let backend = CrosstermBackend::stderr_with_defaults()?;
// or with custom settings
let backend = CrosstermBackend::stdout()
    .with_raw_mode()?
    .with_alternate_screen()?
    .with_mouse_capture()?
```

```rust
use std::io::stderr();

// in a panic / error handler:
CrosstermBackend::reset(stderr())
```

### DIFF
--- a/examples/colors_rgb.rs
+++ b/examples/colors_rgb.rs
@@ -97,7 +97,7 @@ struct ColorsWidget {
 
 fn main() -> Result<()> {
     install_error_hooks()?;
-    let terminal = CrosstermBackend::stdout().into_terminal_with_defaults()?;
+    let terminal = CrosstermBackend::stdout_with_defaults()?.to_terminal()?;
     App::default().run(terminal)?;
     Ok(())
 }
--- a/examples/colors_rgb.rs
+++ b/examples/colors_rgb.rs
@@ -97,7 +97,7 @@ struct ColorsWidget {
 
 fn main() -> Result<()> {
     install_error_hooks()?;
-    let terminal = CrosstermBackend::stdout_with_defaults()?.to_terminal()?;
+    let terminal = CrosstermBackend::stdout_with_defaults()?.into_terminal()?;
     App::default().run(terminal)?;
     Ok(())
 }
--- a/examples/colors_rgb.rs
+++ b/examples/colors_rgb.rs
@@ -37,16 +37,12 @@ use palette::{convert::FromColorUnclamped, Okhsv, Srgb};
 use ratatui::{
     backend::{Backend, CrosstermBackend},
     buffer::Buffer,
-    crossterm::{
-        event::{self, Event, KeyCode, KeyEventKind},
-        terminal::{disable_raw_mode, enable_raw_mode, EnterAlternateScreen, LeaveAlternateScreen},
-        ExecutableCommand,
-    },
+    crossterm::event::{self, Event, KeyCode, KeyEventKind},
     layout::{Constraint, Layout, Rect},
     style::Color,
-    terminal::Terminal,
     text::Text,
     widgets::Widget,
+    Terminal,
 };
 
 #[derive(Debug, Default)]
@@ -101,9 +97,8 @@ struct ColorsWidget {
 
 fn main() -> Result<()> {
     install_error_hooks()?;
-    let terminal = init_terminal()?;
+    let terminal = CrosstermBackend::stdout().into_terminal_with_defaults()?;
     App::default().run(terminal)?;
-    restore_terminal()?;
     Ok(())
 }
 
@@ -272,27 +267,12 @@ fn install_error_hooks() -> Result<()> {
     let panic = panic.into_panic_hook();
     let error = error.into_eyre_hook();
     eyre::set_hook(Box::new(move |e| {
-        let _ = restore_terminal();
+        let _ = CrosstermBackend::reset(stdout());
         error(e)
     }))?;
     panic::set_hook(Box::new(move |info| {
-        let _ = restore_terminal();
+        let _ = CrosstermBackend::reset(stdout());
         panic(info);
     }));
-    Ok(())
-}
-
-fn init_terminal() -> Result<Terminal<impl Backend>> {
-    enable_raw_mode()?;
-    stdout().execute(EnterAlternateScreen)?;
-    let mut terminal = Terminal::new(CrosstermBackend::new(stdout()))?;
-    terminal.clear()?;
-    terminal.hide_cursor()?;
-    Ok(terminal)
-}
-
-fn restore_terminal() -> Result<()> {
-    disable_raw_mode()?;
-    stdout().execute(LeaveAlternateScreen)?;
     Ok(())
 }
--- a/examples/colors_rgb.rs
+++ b/examples/colors_rgb.rs
@@ -97,7 +97,8 @@ struct ColorsWidget {
 
 fn main() -> Result<()> {
     install_error_hooks()?;
-    let terminal = CrosstermBackend::stdout_with_defaults()?.into_terminal()?;
+    let backend = CrosstermBackend::stdout_with_defaults()?;
+    let terminal = Terminal::new(backend)?;
     App::default().run(terminal)?;
     Ok(())
 }
--- a/examples/colors_rgb.rs
+++ b/examples/colors_rgb.rs
@@ -97,7 +97,7 @@ struct ColorsWidget {
 
 fn main() -> Result<()> {
     install_error_hooks()?;
-    let backend = CrosstermBackend::stdout_with_defaults()?;
+    let backend = CrosstermBackend::stdout()?;
     let terminal = Terminal::new(backend)?;
     App::default().run(terminal)?;
     Ok(())
@@ -268,11 +268,11 @@ fn install_error_hooks() -> Result<()> {
     let panic = panic.into_panic_hook();
     let error = error.into_eyre_hook();
     eyre::set_hook(Box::new(move |e| {
-        let _ = CrosstermBackend::reset(stdout());
+        let _ = CrosstermBackend::restore(stdout());
         error(e)
     }))?;
     panic::set_hook(Box::new(move |info| {
-        let _ = CrosstermBackend::reset(stdout());
+        let _ = CrosstermBackend::restore(stdout());
         panic(info);
     }));
     Ok(())
--- a/examples/minimal.rs
+++ b/examples/minimal.rs
@@ -26,7 +26,7 @@ use ratatui::{
 /// [examples]: https://github.com/ratatui-org/ratatui/blob/main/examples
 /// [hello-world]: https://github.com/ratatui-org/ratatui/blob/main/examples/hello_world.rs
 fn main() -> Result<(), Box<dyn std::error::Error>> {
-    let mut terminal = CrosstermBackend::stdout_with_defaults()?.to_terminal()?;
+    let mut terminal = CrosstermBackend::stdout_with_defaults()?.into_terminal()?;
     terminal.clear()?;
     loop {
         terminal.draw(|frame| frame.render_widget(Text::raw("Hello World!"), frame.size()))?;
--- a/examples/minimal.rs
+++ b/examples/minimal.rs
@@ -15,13 +15,8 @@
 
 use ratatui::{
     backend::CrosstermBackend,
-    crossterm::{
-        event::{self, Event, KeyCode, KeyEventKind},
-        execute,
-        terminal::{disable_raw_mode, enable_raw_mode, EnterAlternateScreen, LeaveAlternateScreen},
-    },
+    crossterm::event::{self, Event, KeyCode, KeyEventKind},
     text::Text,
-    Terminal,
 };
 
 /// This is a bare minimum example. There are many approaches to running an application loop, so
@@ -31,9 +26,8 @@ use ratatui::{
 /// [examples]: https://github.com/ratatui-org/ratatui/blob/main/examples
 /// [hello-world]: https://github.com/ratatui-org/ratatui/blob/main/examples/hello_world.rs
 fn main() -> Result<(), Box<dyn std::error::Error>> {
-    let mut terminal = Terminal::new(CrosstermBackend::new(std::io::stdout()))?;
-    enable_raw_mode()?;
-    execute!(terminal.backend_mut(), EnterAlternateScreen)?;
+    let mut terminal = CrosstermBackend::stdout().into_terminal_with_defaults()?;
+    terminal.clear()?;
     loop {
         terminal.draw(|frame| frame.render_widget(Text::raw("Hello World!"), frame.size()))?;
         if let Event::Key(key) = event::read()? {
@@ -42,7 +36,5 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
             }
         }
     }
-    disable_raw_mode()?;
-    execute!(terminal.backend_mut(), LeaveAlternateScreen)?;
     Ok(())
 }
--- a/examples/minimal.rs
+++ b/examples/minimal.rs
@@ -27,7 +27,7 @@ use ratatui::{
 /// [examples]: https://github.com/ratatui-org/ratatui/blob/main/examples
 /// [hello-world]: https://github.com/ratatui-org/ratatui/blob/main/examples/hello_world.rs
 fn main() -> Result<(), Box<dyn std::error::Error>> {
-    let backend = CrosstermBackend::stdout_with_defaults()?;
+    let backend = CrosstermBackend::stdout()?;
     let mut terminal = Terminal::new(backend)?;
     terminal.clear()?;
     loop {
--- a/examples/minimal.rs
+++ b/examples/minimal.rs
@@ -14,9 +14,10 @@
 //! [examples readme]: https://github.com/ratatui-org/ratatui/blob/main/examples/README.md
 
 use ratatui::{
-    backend::{Backend, CrosstermBackend},
+    backend::CrosstermBackend,
     crossterm::event::{self, Event, KeyCode, KeyEventKind},
     text::Text,
+    Terminal,
 };
 
 /// This is a bare minimum example. There are many approaches to running an application loop, so
@@ -26,7 +27,8 @@ use ratatui::{
 /// [examples]: https://github.com/ratatui-org/ratatui/blob/main/examples
 /// [hello-world]: https://github.com/ratatui-org/ratatui/blob/main/examples/hello_world.rs
 fn main() -> Result<(), Box<dyn std::error::Error>> {
-    let mut terminal = CrosstermBackend::stdout_with_defaults()?.into_terminal()?;
+    let backend = CrosstermBackend::stdout_with_defaults()?;
+    let mut terminal = Terminal::new(backend)?;
     terminal.clear()?;
     loop {
         terminal.draw(|frame| frame.render_widget(Text::raw("Hello World!"), frame.size()))?;
--- a/examples/minimal.rs
+++ b/examples/minimal.rs
@@ -14,7 +14,7 @@
 //! [examples readme]: https://github.com/ratatui-org/ratatui/blob/main/examples/README.md
 
 use ratatui::{
-    backend::CrosstermBackend,
+    backend::{Backend, CrosstermBackend},
     crossterm::event::{self, Event, KeyCode, KeyEventKind},
     text::Text,
 };
@@ -26,7 +26,7 @@ use ratatui::{
 /// [examples]: https://github.com/ratatui-org/ratatui/blob/main/examples
 /// [hello-world]: https://github.com/ratatui-org/ratatui/blob/main/examples/hello_world.rs
 fn main() -> Result<(), Box<dyn std::error::Error>> {
-    let mut terminal = CrosstermBackend::stdout().into_terminal_with_defaults()?;
+    let mut terminal = CrosstermBackend::stdout_with_defaults()?.to_terminal()?;
     terminal.clear()?;
     loop {
         terminal.draw(|frame| frame.render_widget(Text::raw("Hello World!"), frame.size()))?;
--- a/src/backend/crossterm.rs
+++ b/src/backend/crossterm.rs
@@ -42,15 +42,15 @@ use crate::{
 /// alternate screen using [`CrosstermBackend::stdout_with_defaults`] or
 /// [`CrosstermBackend::stderr_with_defaults`].
 ///
-/// The `CrosstermBackend` can be converted into a [`Terminal`] instance using
-/// [`CrosstermBackend::to_terminal`].
-///
 /// If the default settings are not desired, the `CrosstermBackend` can be configured using the
 /// `with_*` methods. These methods return an [`io::Result`] containing self so that they can be
 /// chained with other methods. The settings are restored when the `CrosstermBackend` is dropped.
 /// - [`CrosstermBackend::with_raw_mode`] enables raw mode for the terminal.
 /// - [`CrosstermBackend::with_alternate_screen`] switches to the alternate screen.
 /// - [`CrosstermBackend::with_mouse_capture`] enables mouse capture.
+/// - [`CrosstermBackend::with_bracketed_paste`] enables bracketed paste.
+/// - [`CrosstermBackend::with_focus_change`] enables focus change.
+/// - [`CrosstermBackend::with_keyboard_enhancement_flags`] enables keyboard enhancement flags.
 ///
 /// If a backend is configured using the `with_*` methods, the settings are restored when the
 /// `CrosstermBackend` is dropped.
--- a/src/backend/crossterm.rs
+++ b/src/backend/crossterm.rs
@@ -37,7 +37,7 @@ use crate::{
 ///
 /// Convenience methods ([`CrosstermBackend::stdout`] and [`CrosstermBackend::stderr`] are provided
 /// to create a `CrosstermBackend` with [`std::io::stdout`] or [`std::io::stderr`] as the writer
-/// with default settings to enable raw mode and switch to the alternate screen.
+/// with raw mode enabled and the terminal switched to the alternate screen.
 ///
 /// If the default settings are not desired, the `CrosstermBackend` can be configured using the
 /// `with_*` methods. These methods return an [`io::Result`] containing self so that they can be
--- a/src/backend/crossterm.rs
+++ b/src/backend/crossterm.rs
@@ -22,8 +22,7 @@ use crate::{
             ContentStyle, Print, SetAttribute, SetBackgroundColor, SetColors, SetForegroundColor,
         },
         terminal::{
-            self, disable_raw_mode, enable_raw_mode, Clear, EnterAlternateScreen,
-            LeaveAlternateScreen,
+            disable_raw_mode, enable_raw_mode, Clear, EnterAlternateScreen, LeaveAlternateScreen,
         },
     },
     layout::{Rect, Size},
@@ -514,7 +513,7 @@ where
     }
 
     fn size(&self) -> io::Result<Rect> {
-        let (width, height) = terminal::size()?;
+        let (width, height) = crossterm::terminal::size()?;
         Ok(Rect::new(0, 0, width, height))
     }
 
@@ -524,7 +523,7 @@ where
             rows,
             width,
             height,
-        } = terminal::window_size()?;
+        } = crossterm::terminal::window_size()?;
         Ok(WindowSize {
             columns_rows: Size {
                 width: columns,
--- a/src/backend/crossterm.rs
+++ b/src/backend/crossterm.rs
@@ -4,25 +4,32 @@
 //! [Crossterm]: https://crates.io/crates/crossterm
 use std::io::{self, Write};
 
+use crossterm::event::{
+    DisableBracketedPaste, DisableFocusChange, EnableBracketedPaste, EnableFocusChange,
+    KeyboardEnhancementFlags, PopKeyboardEnhancementFlags, PushKeyboardEnhancementFlags,
+};
+
 #[cfg(feature = "underline-color")]
-use crossterm::style::SetUnderlineColor;
+use crate::crossterm::style::SetUnderlineColor;
 
 use crate::{
     backend::{Backend, ClearType, WindowSize},
     buffer::Cell,
     crossterm::{
         cursor::{Hide, MoveTo, Show},
+        event::{DisableMouseCapture, EnableMouseCapture},
         execute, queue,
         style::{
             Attribute as CAttribute, Attributes as CAttributes, Color as CColor, Colors,
             ContentStyle, Print, SetAttribute, SetBackgroundColor, SetColors, SetForegroundColor,
         },
-        terminal::{self, Clear},
+        terminal::{
+            self, disable_raw_mode, enable_raw_mode, Clear, EnterAlternateScreen,
+            LeaveAlternateScreen,
+        },
     },
-    layout::Size,
-    prelude::Rect,
+    layout::{Rect, Size},
     style::{Color, Modifier, Style},
-    Terminal, TerminalOptions,
 };
 
 /// A [`Backend`] implementation that uses [Crossterm] to render to the terminal.
@@ -31,12 +38,21 @@ use crate::{
 /// to send commands to the terminal. It provides methods for drawing content, manipulating the
 /// cursor, and clearing the terminal screen.
 ///
-/// Most applications should enable raw mode and switch to alternate screen mode when using
-/// `CrosstermBackend`. This can be done by calling [`CrosstermBackend::with_raw_mode`] and
-/// [`CrosstermBackend::with_alternate_screen`] before calling [`CrosstermBackend::into_terminal`].
-/// Mouse support can also be enabled with [`CrosstermBackend::with_mouse_support`]. A convenience
-/// method, [`CrosstermBackend::into_terminal_with_defaults`], is provided to enable raw mode and
-/// switch to the alternate screen with default settings.
+/// Convenience methods ([`CrosstermBackend::stdout`] and [`CrosstermBackend::stderr`] are provided
+/// to create a `CrosstermBackend` with [`std::io::stdout`] or [`std::io::stderr`] as the writer.
+/// Additionally, these can be created with default settings to enable raw mode and switch to the
+/// alternate screen using [`CrosstermBackend::stdout_with_defaults`] or
+/// [`CrosstermBackend::stderr_with_defaults`].
+///
+/// The `CrosstermBackend` can be converted into a [`Terminal`] instance using
+/// [`CrosstermBackend::to_terminal`].
+///
+/// If the default settings are not desired, the `CrosstermBackend` can be configured using the
+/// `with_*` methods. These methods return an [`io::Result`] containing self so that they can be
+/// chained with other methods. The settings are restored when the `CrosstermBackend` is dropped.
+/// - [`CrosstermBackend::with_raw_mode`] enables raw mode for the terminal.
+/// - [`CrosstermBackend::with_alternate_screen`] switches to the alternate screen.
+/// - [`CrosstermBackend::with_mouse_capture`] enables mouse capture.
 ///
 /// If a backend is configured using the `with_*` methods, the settings are restored when the
 /// `CrosstermBackend` is dropped.
@@ -44,17 +60,21 @@ use crate::{
 /// # Example
 ///
 /// ```rust,no_run
-/// use ratatui::{backend::CrosstermBackend, Terminal};
+/// use ratatui::backend::{Backend, CrosstermBackend};
+/// use crossterm::event::KeyboardEnhancementFlags;
 ///
-/// let mut terminal = CrosstermBackend::stdout().into_terminal_with_defaults()?;
+/// let mut terminal = CrosstermBackend::stdout_with_defaults()?.to_terminal()?;
 /// // or
-/// let mut terminal = CrosstermBackend::stderr().into_terminal_with_defaults()?;
+/// let mut terminal = CrosstermBackend::stderr_with_defaults()?.to_terminal()?;
 /// // or with custom settings
 /// let mut terminal = CrosstermBackend::stdout()
 ///     .with_raw_mode()?
 ///     .with_alternate_screen()?
-///     .with_mouse_support()?
-///     .into_terminal()?;
+///     .with_mouse_capture()?
+///     .with_bracketed_paste()?
+///     .with_focus_change()?
+///     .with_keyboard_enhancement_flags(KeyboardEnhancementFlags::DISAMBIGUATE_ESCAPE_CODES)?
+///     .to_terminal()?;
 /// # std::io::Result::Ok(())
 /// ```
 ///
@@ -73,6 +93,9 @@ pub struct CrosstermBackend<W: Write> {
     restore_raw_mode_on_drop: bool,
     restore_alternate_screen_on_drop: bool,
     restore_mouse_capture_on_drop: bool,
+    restore_bracketed_paste_on_drop: bool,
+    restore_focus_change_on_drop: bool,
+    restore_keyboard_enhancement_flags_on_drop: bool,
 }
 
 impl<W> CrosstermBackend<W>
@@ -81,12 +104,17 @@ where
 {
     /// Creates a new `CrosstermBackend` with the given writer.
     ///
+    /// Applications will typically use [`CrosstermBackend::stdout`] or [`CrosstermBackend::stderr`]
+    /// to create a `CrosstermBackend` with [`std::io::stdout`] or [`std::io::stderr`] as the
+    /// writer.
+    ///
     /// # Example
     ///
     /// ```rust,no_run
-    /// # use std::io::stdout;
-    /// # use ratatui::prelude::*;
-    /// let backend = CrosstermBackend::new(stdout());
+    /// use std::io;
+    /// use ratatui::backend::CrosstermBackend;
+    ///
+    /// let backend = CrosstermBackend::new(io::stdout());
     /// ```
     pub const fn new(writer: W) -> Self {
         Self {
@@ -94,6 +122,9 @@ where
             restore_raw_mode_on_drop: false,
             restore_alternate_screen_on_drop: false,
             restore_mouse_capture_on_drop: false,
+            restore_bracketed_paste_on_drop: false,
+            restore_focus_change_on_drop: false,
+            restore_keyboard_enhancement_flags_on_drop: false,
         }
     }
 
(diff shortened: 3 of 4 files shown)


joshka:
Note - this is intentionally only implemented for crossterm right now, and it's not carried through to the examples as yet to allow for review of the concept without the bulk.

Note also the comments / previous effort on https://github.com/ratatui-org/ratatui/pull/280

src/backend/crossterm.rs
```diff
@@ -127,6 +119,194 @@ where
...
 
+impl CrosstermBackend<std::io::Stdout> {
+    /// Creates a new `CrosstermBackend` with `std::io::stdout`.
+    ///
+    /// # Example
+    ///
+    /// ```rust,no_run
+    /// use ratatui::prelude::*;
+    ///
+    /// let backend = CrosstermBackend::stdout();
```
EdJoPaTo, line 130:
```suggestion
    /// # use ratatui::prelude::*;
    /// let backend = CrosstermBackend::stdout();
```

src/backend/crossterm.rs
```diff
@@ -127,6 +119,194 @@ where
...
+    /// use ratatui::backend::CrosstermBackend;
+    ///
+    /// let terminal = CrosstermBackend::stdout()
+    ///     .with_raw_mode()?
+    ///     .with_alternate_screen()?
+    ///     .with_mouse_support()?
+    ///     .into_terminal()?;
+    /// # std::io::Result::Ok(())
+    /// ```
+    pub fn into_terminal(self) -> io::Result<Terminal<Self>> {
```
EdJoPaTo, line 167: I dont think I like this approach as it creates two ways of creating a terminal.
Either `Terminal::new(backend)` or `backend.into_terminal()`. The second one needs to be implemented again and again.

I think there should be exactly one way to do something, not multiple ones.

src/backend/crossterm.rs
```diff
@@ -127,6 +119,194 @@ where
...
+    ///
+    /// ```rust,no_run
+    /// use ratatui::backend::CrosstermBackend;
+    ///
+    /// let backend = CrosstermBackend::stdout().with_raw_mode()?;
+    /// # std::io::Result::Ok(())
+    /// ```
+    pub fn with_raw_mode(mut self) -> io::Result<Self> {
+        self.restore_raw_mode_on_drop = true;
+        terminal::enable_raw_mode()?;
```
EdJoPaTo, line 226: Currently, there are `Terminal` (ratatui) and `terminal` (crossterm) close together and not confusing at all. Using an explicit `crossterm::` here clarifies this confusion. 

```suggestion
        crossterm::terminal::enable_raw_mode()?;
```

(There are more places like this)

src/backend/crossterm.rs
```diff
@@ -127,6 +119,194 @@ where
...
+    ///
+    /// ```rust,no_run
+    /// use ratatui::backend::CrosstermBackend;
+    ///
+    /// let backend = CrosstermBackend::stdout().with_alternate_screen()?;
+    /// # std::io::Result::Ok(())
+    /// ```
+    pub fn with_alternate_screen(mut self) -> io::Result<Self> {
+        self.restore_alternate_screen_on_drop = true;
+        execute!(self.writer, terminal::EnterAlternateScreen)?;
```
EdJoPaTo, line 246: (Not only for methods, also constants)
```suggestion
        execute!(self.writer, crossterm::terminal::EnterAlternateScreen)?;
```

src/backend/crossterm.rs
```diff
@@ -127,6 +119,194 @@ where
...
+    ///
+    /// ```rust,no_run
+    /// use ratatui::backend::CrosstermBackend;
+    ///
+    /// let backend = CrosstermBackend::stdout().with_mouse_support()?;
+    /// # std::io::Result::Ok(())
+    /// ```
+    pub fn with_mouse_support(mut self) -> io::Result<Self> {
+        self.restore_mouse_capture_on_drop = true;
+        execute!(self.writer, crossterm::event::EnableMouseCapture)?;
```
EdJoPaTo, line 266: (Funnily, this is already used here)

src/backend/crossterm.rs
```diff
@@ -127,6 +119,194 @@ where
...
+        terminal::disable_raw_mode()?;
+        execute!(
+            writer,
+            terminal::LeaveAlternateScreen,
+            crossterm::event::DisableMouseCapture
+        )
+    }
+}
+
+impl<W: Write> Drop for CrosstermBackend<W> {
```
EdJoPaTo, line 294: Drop implementations are kinda annoying to work with. It also creates another duplicate way of doing things. Reset or drop?
I would go for the explicit reset. It's also quite obvious when the reset wasn't done when stopping the process. Having a reset and for some reason accidentally using the drop implementation on error cases for example might result in different behavior which might also be confusing.

src/backend/crossterm.rs
```diff
@@ -127,6 +119,194 @@ where
...
+    ///
+    /// ```rust,no_run
+    /// use std::io::stdout;
+    ///
+    /// use ratatui::backend::CrosstermBackend;
+    ///
+    /// CrosstermBackend::reset(stdout())?;
+    /// # std::io::Result::Ok(())
+    /// ```
+    pub fn reset(mut writer: W) -> io::Result<()> {
```
EdJoPaTo, line 284: `CrosstermBackend::stdout()` works without a stdout reference. Reset doesnt.
As the reference is required anyway, I think I would remove the `CrosstermBackend::stdout()` (and stderr) as the import is needed for reset in the end anyway.

src/backend/crossterm.rs
```diff
@@ -127,6 +119,194 @@ where
...
+    ///
+    /// # Example
+    ///
+    /// ```rust,no_run
+    /// use ratatui::backend::CrosstermBackend;
+    ///
+    /// let backend = CrosstermBackend::stdout().with_mouse_support()?;
+    /// # std::io::Result::Ok(())
+    /// ```
+    pub fn with_mouse_support(mut self) -> io::Result<Self> {
```
EdJoPaTo, line 264: I think it should be named closer to the name of crossterm
```suggestion
    pub fn with_mouse_event_capture(mut self) -> io::Result<Self> {
```

EdJoPaTo (changes requested):
I like seeing the `with_*` methods and the internal state to allow for a simpler reset. This will simplify a lot of code.

With other stuff I have my problems, noted on them directly for easier, threaded discussions.

src/backend/crossterm.rs
joshka: I'm fine with this. Its purpose is clear and obvious. It makes it easy for the common use case to be a single one liner. Reading through a bunch of examples of into_ vs to_, I think this actually makes more sense as a `to_` method as it's not strictly a conversion. There are many many instances of being able to do the same thing multiple ways in the std lib.

One change I'd make that I think works nicer though is to move the _with_defaults part out of creating a terminal. 
```rust
// So instead of
let terminal = Crossterm::stdout().into_terminal_with_defaults();
// change to
let terminal = Crossterm::stdout().with_default_config().into_terminal();
```

src/backend/crossterm.rs
joshka: I'll actually move these to the imports to avoid this.

src/backend/crossterm.rs
joshka: Also moved `to_terminal` to Backend trait.

src/backend/crossterm.rs
joshka: One possible approach to this might be to add reset to the Backend trait with no parameter and have a version with the parameter on the method which will be called on CrosstermBackend. This overloads the name, but it's reasonable-ish.

src/backend/crossterm.rs
joshka: One possible approach to this might be to add reset to the Backend trait with no parameter and have a version with the parameter on the method which will be called on CrosstermBackend. This overloads the name, but it's reasonable-ish.

src/backend/crossterm.rs
EdJoPaTo: `into` moves, `to` does not move. This moves → into.

Still don't like different ways to do the same.

src/backend/crossterm.rs
EdJoPaTo: Then reset should probably be removed? When users want to do it explicitly they can use `drop(backend)` then.

src/backend/crossterm.rs
EdJoPaTo: This is not symmetrical. Personally I find this confusing. Either there is a shortcut / helper for both, or for none. But having only one for creation and not for ending it seems strange.

src/backend/crossterm.rs
```diff
@@ -127,6 +152,248 @@ where
...
+    ///
+    /// Raw mode and alternate screen are restored when the `CrosstermBackend` is dropped.
+    ///
+    /// # Example
+    ///
+    /// ```rust,no_run
+    /// use ratatui::backend::CrosstermBackend;
+    ///
+    /// let backend = CrosstermBackend::stdout_with_defaults()?;
+    /// # std::io::Result::Ok(())
```
EdJoPaTo, line 180: One shows the import of itself, one does not. Personally I think both shouldn't show the explicit import of itself as the viewer of the docs found said object already. (There are more cases like this)

EdJoPaTo (changes requested):

src/backend/crossterm.rs
joshka: That's literally not possible within a panic handler which is where this is necessary.

src/backend/crossterm.rs
joshka: https://rust-lang.github.io/api-guidelines/naming.html#c-conv

> Conversions prefixed as_ and into_ typically decrease abstraction, either exposing a view into the underlying representation (as) or deconstructing data into its underlying representation (into). Conversions prefixed to_, on the other hand, typically stay at the same level of abstraction but do some work to change from one representation to another.

This feels more like a to_ method to my understanding as it doesn't convert, it just wraps.

src/backend/crossterm.rs
joshka: This method is necessary to make it possible to be called from a panic handler and to avoid having to write the same code everywhere. It doesn't need to keep state. It hangs off the Crossterm struct as a convenience to be findable when users are using code completion in their editors.

What about this do you find confusing?

src/backend/crossterm.rs
```diff
@@ -5,87 +5,98 @@
...
+            PopKeyboardEnhancementFlags, PushKeyboardEnhancementFlags,
+        },
         execute, queue,
         style::{
             Attribute as CAttribute, Attributes as CAttributes, Color as CColor, Colors,
             ContentStyle, Print, SetAttribute, SetBackgroundColor, SetColors, SetForegroundColor,
         },
-        terminal::{self, Clear},
+        terminal::{
+            self, disable_raw_mode, enable_raw_mode, Clear, EnterAlternateScreen,
```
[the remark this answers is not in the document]
joshka: 0c96035c3

src/backend/crossterm.rs
```diff
@@ -94,15 +105,29 @@ where
...
     ///
     /// ```rust,no_run
-    /// # use std::io::stdout;
-    /// # use ratatui::prelude::*;
-    /// let backend = CrosstermBackend::new(stdout());
+    /// use std::io;
+    ///
+    /// use ratatui::backend::CrosstermBackend;
+    ///
+    /// let backend = CrosstermBackend::new(io::stdout());
```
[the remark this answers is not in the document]
joshka: cd525c335

src/backend/crossterm.rs
joshka: cd525c335

src/backend/crossterm.rs
joshka: Comment added in cd525c335

src/backend/crossterm.rs
joshka: Clarification comment added in cd525c335

src/backend/crossterm.rs
joshka: While attempting to convert all the examples to use this, often github CoPilot would attempt to complete `let terminal = CrosstermBackend::stdout_with_defaults()?` with `.into_terminal()?;`. This sort of implies that there's enough statistical evidence that into is a better name (like AI or hate it, often LLM hallucinations tell us what the obvious thing should be even if it's not that thing).

Upshot - I was wrong here - and will change it to into_terminal.

joshka: 408e37ea6

src/backend/crossterm.rs
EdJoPaTo: The following uses the same stdout in both places.
```rust
use std::io::stdout;
let backend = CrosstermBackend::new(stdout());
backend.reset(stdout());
```

 For reset the code already needs to import to stdout. There is not much benefit of having `CrosstermBackend::stdout` when the import is already needed in the first place. And it results in a asymetric design: at one place there is a helper for it, at the other there is not. Either both should have a helper for stdout or neither place should have.

EdJoPaTo: guess I have misread the method header. reset is `CrosstermBackend::reset(writer)` without a reference to self… so the idea is to only use it for the panic handler and nowhere else…? Ok, so pure drop implementation all the time and nothing else.

src/backend/crossterm.rs
EdJoPaTo: So basically, reset is never used except for panic handlers? The single way of cleaning up is the drop implementation.

src/backend/crossterm.rs
EdJoPaTo: Given your cited paragraph, `as_terminal` might be another good naming. `into` also differentiates from `to` by moving into or out of something else. `to` works with reference while `into` moves the ownership.

---

Conversion methods can be fine and great, but ratatui seems to use them way too much in my taste. The Into / From traits are easy foot guns in refactorings, especially when they might use expensive logic. But here I am not against the conversion methods, I am against the multiple ways of doing things.
Currently, there are a lot of test coverage warnings because there are multiple ways of doing the same thing. Tests use one but not the other way. These can easily diverge by accident. In the case of `Buffer::set_span` and `impl Widget for Span` even actively diverge. Which can create subtle bugs as people fine one and use it. (Argument with Rust: Rust decided against the ternary operator as the `if` can do the same.)

Multiple ways of doing the same creates complexity in maintaining the code and in reading code of other people. It can create subtle bugs. Both ways need to be tested, both ways need to be remembered when adapting things.

Boilerplate:

```diff
-Terminal::new(backend);
+backend.into_terminal();
```

This isn't shorter, just using the same method differently. Abstracting a lot of crossterm specific methods as `CrosstermBackend` methods is great. When we need to abstract our own methods away (`Terminal::new`) then there is something fishy. Is `Terminal` needed in the first place then? Could it be abstracted away completely?

Also: you like to search GitHub for code usage. Having exactly a single way of how to do things is way easier to search for.


> * you've previously mentioned introducing code that helps write more concise examples. However I don't think there should be example specific code in Ratatui. I.e. the code used in examples should be the same used in general user's apps
> 
> * this approach simplifies the necessary boilerplate by a significant amount

I think you are mixing up many methods which abstract crossterm related methods which are great and simplify using the `CrosstermBackend` by a lot and the `into_terminal` method which creates a duplicated way of creating a Terminal with the given Backend. I like the simplified usage of crossterm related methods by putting them as with_* methods. The into_terminal method is exactly the same as before, just different but introducing two ways of doing it.


Personally I would like to see either `into_terminal` or `Terminal::new` but not both. Maybe `Terminal::new` is more annoying to use than `into_terminal`, but then it should be marked as deprecated and the new way embraced.
One way should be used in examples, then its also discoverable how to use it.
If it feels like boilerplate, then it should be removed transparently for the user and only be some internal logic.

I agree with this PR in removing a lot of code from examples and real world code and I really look forward to using it, but having `into_terminal` and `Terminal::new` at the same time seem to be a step too much.

src/backend/crossterm.rs
```diff
@@ -5,87 +5,97 @@
...
+#[allow(clippy::struct_excessive_bools)]
 pub struct CrosstermBackend<W: Write> {
     /// The writer used to send commands to the terminal.
     writer: W,
+    restore_raw_mode_on_drop: bool,
+    restore_alternate_screen_on_drop: bool,
+    restore_mouse_capture_on_drop: bool,
+    restore_bracketed_paste_on_drop: bool,
+    restore_focus_change_on_drop: bool,
+    restore_keyboard_enhancement_flags_on_drop: bool,
```
orhun, line 98: Maybe wrap these in a new struct called `BackendConfig` (or `BackendRestoreConfig`)? We might end up having the same or similar configuration values for other backends in the future, so it might be good to do that now.

Also I'm thinking about if exposing these will have any user-facing benefits... But can't think of any right now. Just a thought 😃

EdJoPaTo: Not sure whether it's a great idea to generalize them or not. I am somewhat in favor of not generalizing them to be specific for the given backend implementation.

If there is stuff similar for multiple backends, then its better implemented on common trait.

EdJoPaTo: I think mainly we should get the public interface right, then generalize. As much private as possible, when there is a need for public, open it up.

src/backend/crossterm.rs
joshka: I'll remove this method to get this PR over the line.

src/backend/crossterm.rs
joshka: Yes. Reset should be used in situations where you don't have access to the backend instance. Panic handlers is one such example. I can conceive of some others, but the panic handler is the absolutely necessary part of this.

src/backend/crossterm.rs
joshka: yes

src/backend/crossterm.rs
joshka: > We might ...

YAGNI

src/backend/crossterm.rs
joshka: 9f608f954

examples/colors_rgb.rs
```diff
@@ -37,16 +37,12 @@ use palette::{convert::FromColorUnclamped, Okhsv, Srgb};
...
-        terminal::{disable_raw_mode, enable_raw_mode, EnterAlternateScreen, LeaveAlternateScreen},
-        ExecutableCommand,
-    },
+    crossterm::event::{self, Event, KeyCode, KeyEventKind},
     layout::{Constraint, Layout, Rect},
     style::Color,
-    terminal::Terminal,
     text::Text,
     widgets::Widget,
+    Terminal,
```
EdJoPaTo, line 45: I don't think I like that this kind of change is possible in the first place… would be breaking, but the public export should only exist in a single location. Either it's something global or it needs structuring in modules. Allowing for both seems strange. General thought here: reduce the stuff that is publicly accessible. One of the two places should be gone. (Unrelated to this PR, should be done in its own PR.)

EdJoPaTo (commented):
I really like the simplicity that will come with this and a built-in panic handler. I still want to provoke some thoughts about the current state here. (Also, I want to test this branch to get a feel for its real world usage, which I haven't done yet.)

src/backend/crossterm.rs
```diff
@@ -127,6 +147,245 @@ where
...
+    /// # use ratatui::backend::CrosstermBackend;
+    /// let backend = CrosstermBackend::stdout_with_defaults()?;
+    /// # std::io::Result::Ok(())
+    /// ```
+    pub fn stdout_with_defaults() -> io::Result<Self> {
+        Self::stdout().with_raw_mode()?.with_alternate_screen()
+    }
+}
+
+impl CrosstermBackend<io::Stderr> {
```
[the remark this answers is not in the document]
joshka: Yes. It's used and wanted by users and is a natural thing to have sure to symmetry

examples/colors_rgb.rs
joshka: Yeah agreed., and irrelevant to this PR

examples/colors_rgb.rs
```diff
@@ -272,27 +268,12 @@ fn install_error_hooks() -> Result<()> {
...
     eyre::set_hook(Box::new(move |e| {
-        let _ = restore_terminal();
+        let _ = CrosstermBackend::reset(stdout());
         error(e)
     }))?;
     panic::set_hook(Box::new(move |info| {
-        let _ = restore_terminal();
+        let _ = CrosstermBackend::reset(stdout());
         panic(info);
     }));
```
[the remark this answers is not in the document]
joshka: I don't like the name. Manual doesn't add any value. Bikeshedding here IMO.

src/backend/crossterm.rs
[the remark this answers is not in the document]
EdJoPaTo: What is a common use case for this?

examples/colors_rgb.rs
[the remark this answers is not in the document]
EdJoPaTo: My point is not at all about being replaced in a following PR. I was heavily confused by the existence of this method and I think people should think twice before using it when seeing it. Yeah, there are docs about it that explain it in more details, but I also glanced over that. Personally I think this method should be as hidden as possible as it's not indented to be used normally. It should be `pub` to be used by more advanced approaches, but that's it. Panic handlers and drop should be favored. The name of the method should reflect that more obviously.

src/backend/crossterm.rs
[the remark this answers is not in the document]
joshka: https://github.com/search?q=crosstermbackend+stderr+AND+%28NOT+org%3Aratatui-org%29+AND+%28NOT+is%3Afork%29&type=code

src/backend/crossterm.rs
```diff
@@ -127,6 +147,245 @@ where
...
+    ///
+    /// # Example
+    ///
+    /// ```rust,no_run
+    /// # use ratatui::backend::CrosstermBackend;
+    /// let backend = CrosstermBackend::stdout_with_defaults()?;
+    /// # std::io::Result::Ok(())
+    /// ```
+    pub fn stdout_with_defaults() -> io::Result<Self> {
+        Self::stdout().with_raw_mode()?.with_alternate_screen()
```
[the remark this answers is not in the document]
joshka: The ideal case from my perspective on this is:
CrosstermBackend::stdout() / stderr create a backend that is usable for 95% of the app use cases. It enables raw mode and enters the alternate screen. There doesn't need to be a with_defaults() part, you don't want that, then call CrosstermBackend::new()...

Is that something you'd get behind?

examples/colors_rgb.rs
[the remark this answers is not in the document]
joshka: Restore is a slightly more correct name for this as reset might actually imply doing the same as `tput reset` on a unix devices (this doesn't seem to be surfaced through crossterm, but perhaps should in the future).

Function naming follows from what it does generally more than how it's used. You don't see Write::flush being named manually_flush, and the same concept applies here.

src/backend/crossterm.rs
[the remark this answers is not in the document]
EdJoPaTo: That’s still no real use case why that’s of interest instead of stdout. There is a blogpost comparing the two resulting in stdout being faster. And some other weird usages in there.

Normally stderr is used to not print to stdout as that’s used for piping but everything on the alternate mode doesn’t end up in a pipe?

@orhun has a gpg-tui which uses stderr instead of stdout. Curious why that is?

which might also be fun in the future with a panic handler: the panic handler needs to use the stateless reset method and supply a target. Per default stdout. What happens when stderr is used to start and the panic handler uses stdout?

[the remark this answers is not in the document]
joshka: > That’s still no real use case why that’s of interest instead of stdout. There is a blogpost comparing the two resulting in stdout being faster. And some other weird usages in there.

There are 616 results. That's enough to suggest there's a reasonable use case.

> everything on the alternate mode doesn’t end up in a pipe?

This is (verifiably) incorrect

> which might also be fun in the future with a panic handler: the panic handler needs to use the stateless reset method and supply a target. Per default stdout. What happens when stderr is used to start and the panic handler uses stdout?

Stdout and Stderr don't have separate concepts of the things which are being restored. Choosing one or another is fairly arbitrary, as changes on one affect both Writers. Something that's on my someday investigate list is to look at whether there's a good pattern for choosing the tty instead of stdout/err for the writter. The termwiz backend does this automatically, which makes the output of a Ratatui app not end up in the pipe.

The following example code does exactly what you'd reasonably expect. It restores the terminal to its original state even though the setup was on stdout and the restore was on stderr. You can also check that it pipes (cargo run | wc -l returns various values that align with the number of keys you press).

```rust
use std::io::{stderr, stdout, Write};

use crossterm::{
    cursor::MoveToNextLine,
    event::{
        self, DisableBracketedPaste, DisableFocusChange, DisableMouseCapture, EnableBracketedPaste,
        EnableFocusChange, EnableMouseCapture, Event, KeyboardEnhancementFlags,
        PopKeyboardEnhancementFlags, PushKeyboardEnhancementFlags,
    },
    execute,
};

fn main() -> color_eyre::Result<()> {
    let mut stdout = stdout();
    let mut stderr = stderr();
    enable_raw_mode()?;

    execute!(
        stdout,
        EnterAlternateScreen,
        EnableMouseCapture,
        EnableBracketedPaste,
        EnableFocusChange,
        PushKeyboardEnhancementFlags(KeyboardEnhancementFlags::DISAMBIGUATE_ESCAPE_CODES)
    )?;
    stdout.flush()?;
    write!(stdout, "stdout (alt)")?;
    execute!(stdout, MoveToNextLine(1))?;
    write!(stderr, "stderr (alt)")?;
    execute!(stdout, MoveToNextLine(1))?;

    loop {
        let event = event::read()?;
        write!(stdout, "{:?}", event)?;
        execute!(stdout, MoveToNextLine(1))?;
        if let Event::Key(event) = event {
            if event.code == event::KeyCode::Char('q') {
                break;
            }
        }
        println!()
    }
    execute!(
        stderr,
        LeaveAlternateScreen,
        DisableMouseCapture,
        DisableBracketedPaste,
        DisableFocusChange,
        PopKeyboardEnhancementFlags
    )?;
    stderr.flush()?;
    disable_raw_mode()?;
    write!(stdout, "stdout (alt)")?;
    execute!(stdout, MoveToNextLine(1))?;
    write!(stderr, "stderr (alt)")?;
    execute!(stdout, MoveToNextLine(1))?;
    Ok(())
}
```

src/backend/crossterm.rs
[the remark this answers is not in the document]
joshka: (assuming the 👍 indicates agreement here, 48d6c5a)

EdJoPaTo:
I am curious why crossterm itself doesn't have something like this. I think everyone using crossterm would benefit from something like this. Should this be proposed for crossterm?

joshka:
> I am curious why crossterm itself doesn't have something like this. I think everyone using crossterm would benefit from something like this. Should this be proposed for crossterm?

Define "this".

Regardless it doesn't change whether we'd want this PR to land in the meantime as the development cycle on that side of the fence is pretty slow.

@EdJoPaTo do you have any hard blockers for this PR getting merged? This PR underpins a massive simplification of all the example code in Ratatui as well as customer apps, tutorials, recipes, etc. I'm blocked on completing all of that by this not being merged.

examples/minimal.rs
```diff
@@ -31,9 +27,9 @@ use ratatui::{
 /// [examples]: https://github.com/ratatui-org/ratatui/blob/main/examples
 /// [hello-world]: https://github.com/ratatui-org/ratatui/blob/main/examples/hello_world.rs
 fn main() -> Result<(), Box<dyn std::error::Error>> {
-    let mut terminal = Terminal::new(CrosstermBackend::new(std::io::stdout()))?;
-    enable_raw_mode()?;
-    execute!(terminal.backend_mut(), EnterAlternateScreen)?;
+    let backend = CrosstermBackend::stdout()?;
+    let mut terminal = Terminal::new(backend)?;
+    terminal.clear()?;
```
EdJoPaTo, line 32: Is this clear relevant? I assume it's not and can be removed to be more minimal. Otherwise it should be integrated into Terminal::new to be minimal?

src/backend/crossterm.rs
```diff
@@ -5,87 +5,94 @@
...
-/// The `CrosstermBackend` struct is a wrapper around a writer implementing [`Write`], which is
-/// used to send commands to the terminal. It provides methods for drawing content, manipulating
-/// the cursor, and clearing the terminal screen.
+/// The `CrosstermBackend` struct is a wrapper around a writer implementing [`Write`], which is used
+/// to send commands to the terminal. It provides methods for drawing content, manipulating the
+/// cursor, and clearing the terminal screen.
+///
+/// Convenience methods ([`CrosstermBackend::stdout`] and [`CrosstermBackend::stderr`] are provided
+/// to create a `CrosstermBackend` with [`std::io::stdout`] or [`std::io::stderr`] as the writer
+/// with default settings to enable raw mode and switch to the alternate screen.
```
EdJoPaTo, line 40: The default went away from the method name. It provides the most used ways of using it. The docs should reflect it in my opinion.
```suggestion
/// to create a `CrosstermBackend` with [`std::io::stdout`] or [`std::io::stderr`] as the writer
/// with the often used settings to enable raw mode and switch to the alternate screen.
```

src/backend/crossterm.rs
```diff
@@ -5,87 +5,94 @@
...
 ///
-/// Most applications should not call the methods on `CrosstermBackend` directly, but will instead
-/// use the [`Terminal`] struct, which provides a more ergonomic interface.
+/// If the default settings are not desired, the `CrosstermBackend` can be configured using the
+/// `with_*` methods. These methods return an [`io::Result`] containing self so that they can be
```
EdJoPaTo, line 43: Remove the mentioning of the default. Also, it sounds like it can be reverted while `::stdout().without_raw_mode()` doesn't exist, so phrase it better.

(Will need reformatting for alignment for sure)

```suggestion
/// Additional settings be configured using the
/// `with_*` methods. These methods return an [`io::Result`] containing self so that they can be
```

src/backend/crossterm.rs
```diff
@@ -5,87 +5,94 @@
...
+/// - [`CrosstermBackend::with_keyboard_enhancement_flags`] enables keyboard enhancement flags.
 ///
-/// Usually applications will enable raw mode and switch to alternate screen mode after creating
-/// a `CrosstermBackend`. This is done by calling [`crossterm::terminal::enable_raw_mode`] and
-/// [`crossterm::terminal::EnterAlternateScreen`] (and the corresponding disable/leave functions
-/// when the application exits). This is not done automatically by the backend because it is
-/// possible that the application may want to use the terminal for other purposes (like showing
-/// help text) before entering alternate screen mode.
+/// If a backend is configured using the `with_*` methods, the settings are restored when the
+/// `CrosstermBackend` is dropped.
```
EdJoPaTo, line 53: Maybe this is a bit more clear what is happening?
```suggestion
/// If a backend is configured using the `with_*` methods, the settings are restored back to their default state when the
/// `CrosstermBackend` is dropped.
```

examples/colors_rgb.rs
```diff
@@ -101,9 +97,9 @@ struct ColorsWidget {
 
 fn main() -> Result<()> {
     install_error_hooks()?;
-    let terminal = init_terminal()?;
+    let backend = CrosstermBackend::stdout()?;
+    let terminal = Terminal::new(backend)?;
     App::default().run(terminal)?;
```
EdJoPaTo, line 102: Haven't looked into it, just as an idea: how about merging Terminal and Backend? Putting them into a single line is unfriendly to newcomers but they have a similar goal. Terminal abstracts common things for all the backends. Can we put them on the Backends directly by a trait or something like that? Then something like this might be possible in the future:

```suggestion
    let backend = CrosstermBackend::stdout()?;
    App::default().run(backend)?;
```

Will be breaking, might simplify the usage even further and would be more friendly to beginners. If it seems interesting, get another PR/Issue for this to discuss on this?

src/backend/crossterm.rs
```diff
@@ -127,6 +145,223 @@ where
...
+    /// let backend = CrosstermBackend::stdout()?.with_raw_mode()?;
+    /// # std::io::Result::Ok(())
+    /// ```
+    pub fn with_raw_mode(mut self) -> io::Result<Self> {
+        enable_raw_mode()?;
+        self.restore_raw_mode_on_drop = true;
+        Ok(self)
+    }
+
+    /// Enables raw mode for the terminal and switches to the alternate screen.
```
EdJoPaTo, line 206: raw mode is another method.

```suggestion
    /// Switches to the alternate screen.
```

src/backend/crossterm.rs
```diff
@@ -127,6 +145,223 @@ where
...
+        execute!(self.writer, EnterAlternateScreen)?;
+        self.restore_alternate_screen_on_drop = true;
+        Ok(self)
+    }
+
+    /// Enables mouse capture for the terminal.
+    ///
+    /// Returns an [`io::Result`] containing self so that it can be chained with other methods.
+    ///
+    /// Mouse capture is disabled when the `CrosstermBackend` is dropped.
```
EdJoPaTo, line 229: This comment doesn't exist for all the implementations of `with_*`.

src/backend/crossterm.rs
```diff
@@ -127,6 +145,223 @@ where
...
+    /// let backend = CrosstermBackend::stdout()?.with_mouse_capture()?;
+    /// # std::io::Result::Ok(())
+    /// ```
+    pub fn with_mouse_capture(mut self) -> io::Result<Self> {
+        execute!(self.writer, EnableMouseCapture)?;
+        self.restore_mouse_capture_on_drop = true;
+        Ok(self)
+    }
+
+    /// Enables bracketed paste for the terminal.
```
EdJoPaTo, line 244: What is a bracketed paste? Probably link something on crossterm to read on their description of what it is? Same with other `with_*` methods.

src/backend/crossterm.rs
```diff
@@ -127,6 +145,223 @@ where
...
+    pub fn with_keyboard_enhancement_flags(
+        mut self,
+        flags: KeyboardEnhancementFlags,
+    ) -> io::Result<Self> {
+        execute!(self.writer, PushKeyboardEnhancementFlags(flags))?;
+        self.restore_keyboard_enhancement_flags_on_drop = true;
+        Ok(self)
+    }
+
+    /// Restores the terminal to its default state.
```
EdJoPaTo, line 299: Be a bit more clear that this is for an argument and not `&self` (or similar).
```suggestion
    /// Restores the given terminal to its default state.
```

src/backend/crossterm.rs
```diff
@@ -127,6 +145,223 @@ where
...
+    /// - Disables focus change
+    /// - Pops keyboard enhancement flags
+    ///
+    /// This method is an associated method rather than an instance method to make it possible to
+    /// call without having a `CrosstermBackend` instance. This is often useful in the context of
+    /// error / panic handling.
+    ///
+    /// If you have created a `CrosstermBackend` using the `with_*` methods, the settings are
+    /// restored when the `CrosstermBackend` is dropped, so you do not need to call this method
+    /// manually.
```
EdJoPaTo, line 316: Maybe move this comment up? General description, why might I don't need to read further, what does it in detail.

src/backend/crossterm.rs
```diff
@@ -127,6 +145,223 @@ where
...
+    /// ```
+    pub fn restore(mut writer: W) -> io::Result<()> {
+        disable_raw_mode()?;
+        execute!(
+            writer,
+            LeaveAlternateScreen,
+            DisableMouseCapture,
+            DisableBracketedPaste,
+            DisableFocusChange,
+            PopKeyboardEnhancementFlags
```
EdJoPaTo, line 333: Out of curiosity: Is the order relevant? If not, maybe sort them alphabetically?

src/backend/crossterm.rs
```diff
@@ -127,6 +145,223 @@ where
...
+        }
+        if self.restore_bracketed_paste_on_drop {
+            let _ = execute!(self.writer, DisableBracketedPaste);
+        }
+        if self.restore_focus_change_on_drop {
+            let _ = execute!(self.writer, DisableFocusChange);
+        }
+        if self.restore_keyboard_enhancement_flags_on_drop {
+            let _ = execute!(self.writer, PopKeyboardEnhancementFlags);
+        }
```
EdJoPaTo, line 360: Same curiosity: Is the order relevant?

src/backend/crossterm.rs
```diff
@@ -247,7 +482,7 @@ where
     }
 
     fn size(&self) -> io::Result<Rect> {
```
EdJoPaTo, line 484: breaking change independent of this PR: use Size rather than Rect?

EdJoPaTo (commented):
I still like the simplification that will come with this a lot. Some curiosity / minor nitpicks remain, some docs should improve, then I think this can be merged.

---

@joshka dismissed my review, I think this should be enabled automatically for this repo: dismiss reviews on new commits. Either it's quite fast to review them again or it might have changed something that should be reviewed again. (Maybe also a review of multiple and not just one person might be ideal especially for bigger things.)

joshka:
> @joshka dismissed my review, I think this should be enabled automatically for this repo: dismiss reviews on new commits. Either it's quite fast to review them again or it might have changed something that should be reviewed again. (Maybe also a review of multiple and not just one person might be ideal especially for bigger things.)

That approach causes a long turn around time for things which don't matter. I prefer to work based on trust that we'll do the right thing once given the go ahead to merge.

src/backend/crossterm.rs
joshka: I think this is clear enough and seems like too much of a nitpick to worry about changing.

src/backend/crossterm.rs
joshka: I think this is clear enough

src/backend/crossterm.rs
joshka: nitpick

src/backend/crossterm.rs
joshka: nitpick

src/backend/crossterm.rs
joshka: nitpick

src/backend/crossterm.rs
joshka: nitpick

src/backend/crossterm.rs
joshka: these are arranged in the order of most to least important to not fail.

src/backend/crossterm.rs
joshka: nitpick

src/backend/crossterm.rs
joshka: noted

joshka (commented):
I've noted that a most of these points are nitpicks and my tolerance for handling this amount of not relevant feedback on this PR is rapidly approaching zero. There's nothing that would block this from being merged or that would cause a user not to understand how these methods work. As such resolving them.

joshka (commented):
I've noted that a most of these points are nitpicks and my tolerance for handling this amount of not relevant feedback on this PR is rapidly approaching zero. There's nothing that would block this from being merged or that would cause a user not to understand how these methods work. As such resolving them.

joshka:
@EdJoPaTo see my email about nits etc.

kdheepak:
Most of the PR looks good to me (but I haven't read all the comments on this thread), will do so over the weekend.

src/backend/crossterm.rs
```diff
@@ -127,6 +145,223 @@ where
...
+    ///
+    /// ```rust,no_run
+    /// # use ratatui::backend::CrosstermBackend;
+    /// let backend = CrosstermBackend::stdout();
+    /// ```
+    pub fn stdout() -> io::Result<Self> {
+        Self::new(io::stdout())
+            .with_raw_mode()?
+            .with_alternate_screen()
+    }
```
kdheepak, line 164: The PR description says:

> - `CrosstermBackend::stdout` and `CrosstermBackend::stderr` for creating backends with std::io::stdout and std::io::stderr respectively.
> - `CrosstermBackend::stdout_with_defaults` and `CrosstermBackend::stderr_with_defaults` for creating backends with that use the alternate screen and raw mode.

which doesn't match this implementation. 

I prefer it to work the way the PR description mentions. It makes it much more clearer imo.

kdheepak: Alternatively, `CrosstermBackend::stdout_tui` and `CrosstermBackend::stderr_tui`.

joshka: The idea behind doing some reasonable defaults (raw/alternate screen) on these is that this is necessary for 90% of all apps. Anything else can do `CrosstermBackend::new(stdout()).with_xxx()`...

kdheepak: It's a little confusing imo to find in a Ratatui app's user code two things that are very similar but do different things

- `CrosstermBackend::new(stdout())` for inline
- `CrosstermBackend::stdout()` for a TUI

kdheepak: I like the reasonable defaults in place here, I'm just not convinced that the name conveys those defaults clearly.

examples/minimal.rs
EdJoPaTo: To clearify: My question here was not to cleanup the examples. "just enough to show the real world usage" sparked whether the default usage can be even more simplified. But yeah, this PR is not about the terminal.

src/backend/crossterm.rs
EdJoPaTo: Marked as unresolved. Should be adapted once a better name is figured out (https://github.com/ratatui-org/ratatui/pull/1180#discussion_r1659432037).

src/backend/crossterm.rs
EdJoPaTo: Reading this again: It's also way easier to read and understand in my opinion.

examples/colors_rgb.rs
EdJoPaTo: https://github.com/ratatui-org/ratatui/issues/1251

src/backend/crossterm.rs
EdJoPaTo: Probably linking to the actual used behind the scenes structs in crossterm is both the simplest and most helpful way. Basically adding `[EnterAlternateScreen]` in the docs will automatically link to <https://docs.rs/crossterm/latest/crossterm/terminal/struct.EnterAlternateScreen.html> which allows the user to read further about what it actually is that is enabled.

In case of [bracketed paste, its docs](https://docs.rs/crossterm/latest/crossterm/event/struct.EnableBracketedPaste.html) link to <https://en.wikipedia.org/wiki/Bracketed-paste>. Maybe link there too?

src/backend/crossterm.rs
EdJoPaTo: Still think this would be helpful to differentiate more clearly between the `&self` methods and this static method consuming some other reader.

src/backend/crossterm.rs
EdJoPaTo: Searched the crossterm docs for a bit and didn't find an obvious hint to make sure of some order here. So probably fine to invoke them in any order.

src/backend/crossterm.rs
```diff
@@ -127,6 +145,223 @@ where
...
+        disable_raw_mode()?;
+        execute!(
+            writer,
+            LeaveAlternateScreen,
+            DisableMouseCapture,
+            DisableBracketedPaste,
+            DisableFocusChange,
+            PopKeyboardEnhancementFlags
+        )?;
+        writer.flush()
```
EdJoPaTo, line 335: `execute!` calls flush already. This likely wont change in crossterm as `execute` is `queue` + flush.

```suggestion
        execute!(
            writer,
            LeaveAlternateScreen,
            DisableMouseCapture,
            DisableBracketedPaste,
            DisableFocusChange,
            PopKeyboardEnhancementFlags
        )
```

src/backend/crossterm.rs
```diff
@@ -127,6 +145,223 @@ where
...
+            let _ = execute!(self.writer, DisableBracketedPaste);
+        }
+        if self.restore_focus_change_on_drop {
+            let _ = execute!(self.writer, DisableFocusChange);
+        }
+        if self.restore_keyboard_enhancement_flags_on_drop {
+            let _ = execute!(self.writer, PopKeyboardEnhancementFlags);
+        }
+        let _ = self.writer.flush();
+    }
```
EdJoPaTo, line 362: Thought: `execute!` is `queue!` + flush. This method ends with flush already. It could use `queue!` for all the calls and then flush once in the end.
Even when something fails the errors are ignored and the flush call is definitely executed.

It's probably not that relevant as it's not called that often, but I still think a worthy improvement as queue and execute both exist for exactly this reason.

EdJoPaTo, line 362: Given that `CrosstermBackend::restore` just does all these without checking for boolean flags. Is there a downside to do it that way in all the cases?

This would simplify this to just call `CrosstermBackend::restore(self.writer)` and removes a lot of boolean flags resulting in less complexity or possibility for mistakes.

src/backend/crossterm.rs
EdJoPaTo: I like the idea of `CrosstermBackend::stdout_tui()` with the alternative of `CrossteermBackend::new(stdout()).with_*` is actually quite good. It removes the `CrosstermBackend::stdout()` == `CrosstermBackend::new(stdout())` similarity and is not just `default` which is more confusing as there are different defaults for different use cases.

EdJoPaTo:
As a crossterm version change is included with the upcoming release I would like to reduce the stuff needed from crossterm even further with the 0.28 release. Can we get this into the release? (Without rushing it, I would prefer to wait with the release in order to resolve all the thoughts to make this right)

joshka:
> As a crossterm version change is included with the upcoming release I would like to reduce the stuff needed from crossterm even further with the 0.28 release. Can we get this into the release? (Without rushing it, I would prefer to wait with the release in order to resolve all the thoughts to make this right)

I think this might be worth missing the 0.28.0 release. There's a lot of small nuanced things to go through, and then once this lands it would be nice to revamp **all** the examples to use this, that's a fair bit bigger than I'd like to commit to before releasing 0.28, and I'd really like those two things to go together.

joshka:
Something to consider as a total simplification would be to just have `Terminal::init()` and `Terminal::restore()` do the right thing (crossterm, stdout, raw mode, alternate screen).

joshka:
Closed by #1289

PR #1181, opened 2024-06-16
feat(color-eyre): Add color eyre to the default app terminal setup

joshka:
Apps can now enable color-eyre hooks using the with_color_eyre_hooks
method on CrosstermBackend. This is also added to the default features
in the Cargo.toml file, and the defaults that are applied to terminals
created using CrosstermBackend::into_terminal_with_defaults.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,6 +50,7 @@ unicode-width = "0.1.13"
 anyhow = "1.0.71"
 argh = "0.1.12"
 better-panic = "0.3.0"
+color-eyre = "0.6.2"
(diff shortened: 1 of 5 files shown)


EdJoPaTo:
When setting up a panic hook with an external crate, there should also be the panic hook with std only methods.

joshka:
> When setting up a panic hook with an external crate, there should also be the panic hook with std only methods.

Added

joshka:
Blocked on https://github.com/ratatui-org/ratatui/pull/1180

joshka:
Closed by #1289

PR #1182, opened 2024-06-16
feat(borders): Add FULL and EMPTY border sets

joshka:
`border::FULL` uses a full block symbol, while `border::EMPTY` uses an
empty space. This is useful for when you need to allocate space for the
border and apply the border style to a block without actually drawing a
border. This makes it possible to style the entire title area or a block
rather than just the title content.

```rust
use ratatui::{symbols::border, widgets::Block};
let block = Block::bordered().title("Title").border_set(border::FULL);
let block = Block::bordered().title("Title").border_set(border::EMPTY);
```


joshka:
This was required in order to properly handle titles in the list example. Extracted from #1159

kdheepak (approved):

src/symbols.rs
```diff
@@ -470,6 +470,56 @@ pub mod border {
         horizontal_top: QUADRANT_TOP_HALF,
         horizontal_bottom: QUADRANT_BOTTOM_HALF,
     };
+
+    /// Solid border set
+    ///
+    /// The border is created by using full blocks for all sides.
+    ///
+    /// ```text
```
EdJoPaTo, line 478: Just as an idea… could this be a test case? `let expected = from_lines([…]);`? `assert_eq` and stuff like that could start with `#` to only show the lines in the docs, but the docs are checked then?

Not sure if its obvious enough as only const str is involved here, so maybe over the top to add a test here?

src/symbols.rs
```diff
@@ -470,6 +470,56 @@ pub mod border {
...
+        horizontal_top: block::FULL,
+        horizontal_bottom: block::FULL,
+    };
+
+    /// Empty border set
+    ///
+    /// The border is created by using empty strings for all sides.
+    ///
+    /// This is useful for ensuring that the border style is applied to a border on a block with a
+    /// title without actually drawing a border.
```
EdJoPaTo, line 499: Maybe we should enforce an empty border when a title is defined, but no border is explicitly defined? Sounds like a bug that should be fixed internally rather than adding a workaround for external use?

joshka: It's sort of an edge case (pun intended). I think it probably does need to be treated as a bug, but I'm not 100% sure of that. In the meantime, these are useful in addressing the problem more simply.

src/symbols.rs
joshka: Done in b3353c350

orhun (approved):

src/symbols.rs
EdJoPaTo: It’s now in its own test module and not in the doc comment? Or am I missing something?

my idea was to ensure the doc comment is always correct.

joshka: Doc tests are slow to compile and should be avoided when there's an obvious unit test that can be made. There's very little value to testing these doc comments as they won't change and are correct already.

EdJoPaTo: There is no need to test anything as we obviously do everything correct already :ghost: 

Personal opinion: If there is doc, and it can be verified, then it should be verified. Obvious unit tests aren't visible when using the API or looking at the docs. And they duplicate the example in the docs.
Another personal opinion: When I need to look at tests, the docs are kinda useless.

Yeah, running them is more expensive as they are done in sequence rather than in parallel, but it can be done on CI so not big of a downside in my opinion.

I could argue here that it's only about short strings and there is no need for a test in a first place. But duplicating it and testing it in a different place seems actually worse to me as its further away and prone to mistakes.

joshka: > If there is doc, and it can be verified, then it should be verified

The documentation is clear. It shows how the borders appear. Adding code assertions to these docs makes it less clear.

An easy mistake to accidentally make is to swap the left and right vertical borders - particularly when these are used in the QUADRANT_INSIDE / ONE_EIGHTH_WIDE where the names of the symbols are opposite their location. The unit tests catch these errors and ensure they are not later introduced by some errant future refactoring.

Resolving this as a nitpick as there is no real impact on readability or maintainability of this.

EdJoPaTo: > Adding code assertions to these docs makes it less clear.

> `assert_eq` and stuff like that could start with `#` to only show the lines in the docs

Issue #996, opened 2024-03-24
Rename `render_ref` for `StatefulWidgetRef` to `render_stateful_ref`


kdheepak:
## Problem

This will prevent ambiguity between `WidgetRef` and `StatefulWidgetRef`

https://github.com/ratatui-org/ratatui/blob/8719608bdaf32ba92bdfdd60569cf73f7070a618/src/widgets.rs#L403

It would be nice to disambiguate it for `Widget` and `StatefulWidget` as well but that would be a big breaking change. For `WidgetRef` and `StatefulWidgetRef`, these traits are still unstable at the moment.

joshka:
+1 on this. I think while it's inconsistent with the StatefulWidget, it's the right choice.

One alternative worth considering is naming the traits: `RenderRef` and `RenderRefWithState` and creating two new traits to cover the old widget ones. I'm not sure how I feel about that idea however.

sww1235:
> It would be nice to disambiguate it for Widget and StatefulWidget as well but that would be a big breaking change. For WidgetRef and StatefulWidgetRef, these traits are still unstable at the moment so we can make this change.

The `Widget`/`StatefulWidget` part just bit me when implementing `Widget` as opposed to using one big `layout` function. 
It would be nice to fix this API issue in future.

kdheepak:
Maybe:

- `RenderRef`
- `RenderRefWithState`
- `RenderOwned`
- `RenderOwnedWithState`

? And at some point in the future we can deprecate `Widget` and `StatefulWidget`.

joshka:
> And at some point in the future we can deprecate Widget and StatefulWidget.

The hassle is that doing this breaks everything conceptually (as well as semver). The pay off seems low as it's easy to work around.

EdJoPaTo:
Im not sure that we actually need `RenderRef` and `RenderOwned`. We just need one and implement it either for the `&ListWidget` or for `ListWidget` (`impl Render for &ListWidget`). That way its the same trait and the widget knows whether it requires ownership on rendering.

joshka:
A trait that accepts `&self` is required in order to make boxed traits work.

joshka:
(Removed good first issue while we're discussing this)

kdheepak:
If we are renaming these methods, can we do something like this instead?

```
pub trait RenderRef {
    type State;
    
    fn render_ref_with_state(&self,  area: Rect, buf: &mut Buffer, state: Option<&mut Self::State>);

[...]

PR #1183, opened 2024-06-17
feat(widgets): Rename `StatefulWidget::render` to `render_stateful`

joshka:
This change renames the `StatefulWidget::render` method to
`render_stateful` to avoid conflicts with the `Widget::render` method.

Often both the `Widget` and `StatefulWidget` traits are in scope, and
when calling the `render` method on a `StatefulWidget` the compiler
cannot determine which trait to use. This change resolves that issue by
renaming the `StatefulWidget::render` method to `render_stateful`.

The `StatefulWidget::render` method is still available, but it is
deprecated. A default implementation of `render_stateful` is provided
that calls the deprecated `render` method.

Callers should update their code to use `render_stateful` instead of
`render`.
implementors of the `StatefulWidget` trait should update their
implementations to implement `render_stateful` instead of `render`, and
provide an implementation of `render` that calls `render_stateful`.

This change is non-breaking. The deprecated `render` method will be
removed in a future release of Ratatui (likely 0.29.0).

Addresses part of a problem raised in
<https://github.com/ratatui-org/ratatui/issues/996>

### DIFF
--- a/benches/list.rs
+++ b/benches/list.rs
@@ -63,7 +63,7 @@ fn render_stateful(bencher: &mut Bencher, list: &List, mut state: ListState) {
     bencher.iter_batched(
         || list.to_owned(),
         |bench_list| {
-            StatefulWidget::render(bench_list, buffer.area, &mut buffer, &mut state);
+            bench_list.render_stateful(buffer.area, &mut buffer, &mut state);
         },
         BatchSize::LargeInput,
     );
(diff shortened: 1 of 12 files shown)


kdheepak (approved):
Similar to the other PR, I think we could do `render_with_state()` instead, but I'm fine with `render_stateful()`. Again, worth checking in with other maintainers for a second opinion.

joshka:
> Similar to the other PR, I think we could do `render_with_state()` instead, but I'm fine with `render_stateful()`. Again, worth checking in with other maintainers for a second opinion.

likewise - I'm 55/45 on stateful vs with_state here too.

joshka:
Holding for resolution of discussion at https://forum.ratatui.rs/t/naming-render-traits-methods/68

PR #1184, opened 2024-06-17
fix(widgets)!: rename StatefulWidgetRef::render_ref to `render_stateful_ref`

joshka:
This commit renames the `StatefulWidgetRef::render_ref` method to
`render_stateful_ref`. This helps avoid collisions with the `WidgetRef`
trait's `render_ref` method. This change is breaking and requires
updating all implementations of `StatefulWidgetRef`.

BREAKING CHANGE: `StatefulWidgetRef::render_ref` has been renamed to
`StatefulWidgetRef::render_stateful_ref`.

```diff
 trait StatefulWidgetRef {
     type State;
-    fn render_ref(&self, area: Rect, buf: &mut Buffer, state: &mut Self::State) { }
+    fn render_stateful_ref(&self, area: Rect, buf: &mut Buffer, state: &mut Self::State) { }
 }
```

Partially addresses <https://github.com/ratatui-org/ratatui/issues/996>

### DIFF
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -55,6 +55,20 @@ This is a quick summary of the sections below:
 
 ## Unreleased
 
+### `StatefulWidgetRef::render_ref` renamed to `render_stateful_ref` [#1184]
+
+[#1184]: https://github.com/ratatui-org/ratatui/pull/1184
+
+This change helps avoid collisions with `WidgetRef::render_ref`.
+
+```diff
+ trait StatefulWidgetRef {
+     type State;
+-    fn render_ref(&self, area: Rect, buf: &mut Buffer, state: &mut Self::State) { }
++    fn render_stateful_ref(&self, area: Rect, buf: &mut Buffer, state: &mut Self::State) { }
+ }
+```
+
 ### Prelude items added / removed ([#1149])
 
 The following items have been removed from the prelude:
--- a/src/terminal/frame.rs
+++ b/src/terminal/frame.rs
@@ -129,7 +129,7 @@ impl Frame<'_> {
     }
 
     /// Render a [`StatefulWidgetRef`] to the current buffer using
-    /// [`StatefulWidgetRef::render_ref`].
+    /// [`StatefulWidgetRef::render_stateful_ref`].
     ///
     /// Usually the area argument is the size of the current frame or a sub-area of the current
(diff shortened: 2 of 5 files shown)


kdheepak (approved):

kdheepak:
I think we could do `render_ref_with_state()` instead, but I'm fine with `render_stateful_ref`. Maybe check in with other maintainers before merging?

joshka:
> I think we could do `render_ref_with_state()` instead, but I'm fine with `render_stateful_ref()`. Maybe check in with other maintainers before merging?

I thought about that a little prior to writing this. The main pro for `_stateful` is that it's in the name of the trait already, so it's kinda expected to be in the method name. The main pro for `_with_state` seems is that it's a natural name for what this does. For me this feels like a 55/45 naming decision. I have a small amount of preference for _stateful, but would be convinced if there were good examples of _with_xxx on trait implementations in `std`.

kdheepak:
I have a preference for renaming these new trait names (the ones that are unstable) to `Render` and `RenderWithState` and then changing the names of the trait methods to match accordingly. 

The disadvantage is that it would break from convention with the `Widget` and `StatefulWidget` traits. The advantage is that it is more consistent and clear (imo). 

And I can see a world where everyone moves to using these new traits, and we deprecate the old ones and remove them eventually (a few years later?).

joshka:
> I have a preference for renaming these new trait names (the ones that are unstable) to `Render` and `RenderWithState` and then changing the names of the trait methods to match accordingly.
> 
> The disadvantage is that it would break from convention with the `Widget` and `StatefulWidget` traits. The advantage is that it is more consistent and clear (imo).
> 
> And I can see a world where everyone moves to using these new traits, and we deprecate the old ones and remove them eventually (a few years later?).

Let's move this point to a discussion on the forum.

joshka:
> Let's move this point to a discussion on the forum.

https://forum.ratatui.rs/t/naming-render-traits-methods/68

kdheepak:
I’d like to hold out till we figure out the naming, if that’s okay. Making one breaking change later (v0.28.0?) is better than making one now and potentially again one later.

joshka:
> Probably off-topic to this PR… why move this to a different place? Personally the forum feels like a place for "how to do this" and "showoffs". Stuff that should end in the code should be discussed as issues next to the code.

It is off-topic for this PR. If you ask about the rationale for this on the forum, I'll answer there... ;P

Issue #1179, opened 2024-06-15
TableState resets offset to 0


Cretezy:
## Description

`TableState` is calculating wrong row bounds, causing `offset` to be reset to 0. This means that `offset` is broken on `TableState`.

## To Reproduce

```rust
use crossterm::{
    event::{self, KeyCode, KeyEventKind},
    terminal::{disable_raw_mode, enable_raw_mode, EnterAlternateScreen, LeaveAlternateScreen},
    ExecutableCommand,
};
use ratatui::{
    layout::Constraint,
    prelude::{CrosstermBackend, Stylize, Terminal},
    widgets::{Paragraph, Row, Table, TableState},
};
use std::io::{stdout, Result};

fn main() -> Result<()> {
    stdout().execute(EnterAlternateScreen)?;
    enable_raw_mode()?;
    let mut terminal = Terminal::new(CrosstermBackend::new(stdout()))?;
    terminal.clear()?;

    let mut table_state = TableState::default();

    loop {
        terminal.draw(|frame| {
            let area = frame.size();

            let mut rows = vec![];
            for i in 0..100 {
                rows.push(Row::new([i.to_string(), table_state.offset().to_string()]))
            }

            let table = Table::new(rows, [Constraint::Fill(1), Constraint::Fill(1)]);

            frame.render_stateful_widget(table, area, &mut table_state);
        })?;

        if let event::Event::Key(key) = event::read()? {
            if key.kind == KeyEventKind::Press && key.code == KeyCode::Char('j') {
                *table_state.offset_mut() += 1;
            }
            if key.kind == KeyEventKind::Press && key.code == KeyCode::Char('q') {
                break;
            }
        }
    }

    stdout().execute(LeaveAlternateScreen)?;
    disable_raw_mode()?;
    Ok(())
}
```

## Expected behavior

When pressing `j`, the second column (which is `table_state.offset()`) should increment to 1, then 2, then 3, etc.

Currently, it goes to 1, then gets reset to 0 on the next render.

## Environment

- OS: Arch Linux
- Terminal Emulator: Kitty
- Font: Hack
- Crate version: 0.26.3
- Backend: crossterm 0.27.0

joshka:
@Cretezy can you please check that https://github.com/ratatui-org/ratatui/pull/1187 solves your issue (See https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#the-patch-section for info on overriding deps).

Cretezy:
[...]

PR #1187, opened 2024-06-17
fix(table): ensure render offset without selection properly

joshka:
Fixes: <https://github.com/ratatui-org/ratatui/issues/1179>

### DIFF
--- a/src/widgets/table/table.rs
+++ b/src/widgets/table/table.rs
@@ -20,7 +20,9 @@ use crate::{layout::Flex, prelude::*, style::Styled, widgets::Block};
 /// [`Table`] implements [`Widget`] and so it can be drawn using [`Frame::render_widget`].
 ///
 /// [`Table`] is also a [`StatefulWidget`], which means you can use it with [`TableState`] to allow
-/// the user to scroll through the rows and select one of them.
+/// the user to scroll through the rows and select one of them. When rendering a [`Table`] with a
+/// [`TableState`], the selected row will be highlighted. If the selected row is not visible (based
+/// on the offset), the table will be scrolled to make the selected row visible.
 ///
 /// Note: if the `widths` field is empty, the table will be rendered with equal widths.
 ///
@@ -1007,6 +1009,8 @@ mod tests {
 
     #[cfg(test)]
     mod render {
+        use rstest::rstest;
+
         use super::*;
 
         #[test]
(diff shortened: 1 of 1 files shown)


src/widgets/table/table.rs
```diff
@@ -775,7 +775,7 @@ impl Table<'_> {
             end += 1;
         }
 
-        let selected = selected.unwrap_or(0).min(self.rows.len() - 1);
+        let selected = selected.unwrap_or(start).min(self.rows.len() - 1);
```
EdJoPaTo, line 778: Not sure if I understand the unwrap here. When there is no selection, the following block gets pointless? If selection, do something, else skip ahead? Or am I missing something?

src/widgets/table/table.rs
```diff
@@ -1197,6 +1197,74 @@ mod tests {
...
+            let rows = (1..100).map(|i| Row::new([i.to_string()]));
+            let table = Table::new(rows, [Constraint::Length(2)]);
+            let mut buf = Buffer::empty(Rect::new(0, 0, 2, 5));
+            let mut state = TableState::new().with_offset(50);
+            StatefulWidget::render(table, Rect::new(0, 0, 5, 5), &mut buf, &mut state);
+            assert_eq!(buf, Buffer::with_lines(["51", "52", "53", "54", "55"]));
+        }
+
+        #[test]
+        fn render_with_selection_and_offset() {
```
EdJoPaTo, line 1215: This looks like a good rstest candidate.

src/widgets/table/table.rs
joshka: This is an easy way to skip past the two while blocks (they both evaluate as false). In general, I never check for an empty before a for loop unless there's some behavior that needs to act differently on an empty list. This is similar.

Because this is a bug fix and not a rewrite, the simplest smallest fix that addresses the problem is the right one.

Happy for you to rewrite this if you'd like to put in the time and effort, but that's not the purpose of this PR.

src/widgets/table/table.rs
joshka: Agreed 3482bbca3

joshka: Make that 62320257d (the no selection case can be handled in the same test)

src/widgets/table/table.rs
EdJoPaTo: > the simplest smallest fix that addresses the problem is the right one.

The unwrap_or(start) seems more confusing to me than `if let Some(selected) = selected.map(… min …)`. It somehow magically skips the following loops? Then I need to read and understand the loops to understand what is happening. Reading an if is way simpler here: loops only handled when there is a selection.

I am not about a rewrite, I am for a simple and readable way to fix this. And I don't think the `unwrap_or(start)` is readable.

src/widgets/table/table.rs
```diff
@@ -1197,6 +1201,37 @@ mod tests {
...
+        #[case::selection_before_offset(20, ["20", "21", "22", "23", "24"], 20)]
+        #[case::selection_immediately_before_offset(49, ["49", "50", "51", "52", "53"], 49)]
+        #[case::selection_at_start_of_offset(50, ["50", "51", "52", "53", "54"], 50)]
+        #[case::selection_at_end_of_offset(54, ["50", "51", "52", "53", "54"], 50)]
+        #[case::selection_immediately_after_offset(55, ["51", "52", "53", "54", "55"], 51)]
+        #[case::selection_after_offset(80, ["76", "77", "78", "79", "80"], 76)]
+        fn render_with_selection_and_offset<T: Into<Option<usize>>>(
+            #[case] selected_row: T,
+            #[case] expected_items: [&str; 5],
+            #[case] expected_offset: usize,
```
EdJoPaTo, line 1220: Minor nitpick: I would switch items and offset, so the simple single number is before all the items. Gets easier to read then. First this, then that, the rest is the array.

src/widgets/table/table.rs
```diff
@@ -1197,6 +1201,37 @@ mod tests {
             ]);
             assert_eq!(buf, expected);
         }
+
+
+        /// Note that this includes a regression test for a bug where the table would not render the
+        /// correct rows when there is no selection.
+        /// <https://github.com/ratatui-org/ratatui/issues/1179>
+        #[rstest]
+        #[case::no_selction(None, ["50", "51", "52", "53", "54"], 50)]
```
EdJoPaTo, line 1210:
```suggestion
        #[case::no_selection(None, ["50", "51", "52", "53", "54"], 50)]
```

src/widgets/table/table.rs
EdJoPaTo: The rstest is way easier to read :sparkles:

src/widgets/table/table.rs
joshka: Fair call (Noting that the order of assertions below is intentionally the opposite of this, as the buffer assertion would have more info in it than the offset.)

src/widgets/table/table.rs
joshka: 6eb01268e

src/widgets/table/table.rs
joshka: 6eb01268e

src/widgets/table/table.rs
joshka: 6eb01268e

EdJoPaTo (approved):

Issue #1129, opened 2024-05-23
Italic Borders?


Jaxx497:
## Description
I've found that when displayed in the Windows terminal (via powershell), the borders of a widget will appear italicized.

## To Reproduce
This is the relevant code that will reproduce the error (image below):
```                
Block::new()
    .borders(Borders::ALL)
    .border_type(BorderType::Double)
    .border_style(Style::new().red().on_black())
    .title(Title::from(" Now Playing ").alignment(Alignment::Center))
    .style(Style::new().light_red().italic())
```

## Expected behavior
I'd expect the borders to render normally (see code snippet at bottom for work-around)

## Screenshots
![image](https://github.com/ratatui-org/ratatui/assets/29159748/dc99e4e8-4e8f-4581-b824-e88cf7e13613)

## Environment
- OS: Windows 11
- Terminal Emulator: Windows Terminal
- Font: Tested on multiple fonts including Lucida Console, SauceCodePro Nerd Font, NSimSun...
- Crate version: 0.26.3
- Backend: Crossterm

## Additional context
It's possible to workaround this issue with the following code. I didn't look too far into it, but it seems that the initial implementation is applying the italics to the border characters as well as the title characters, whereas the following implementation limits the italics to the title itself, but this feels far less intuitive to write out. It's worth noting that I did not see any problems with this when using Wezterm on Windows or Kitty on Arch Linux.

```
Block::new()
    .borders(Borders::ALL)
    .border_type(BorderType::Double)
    .border_style(Style::new().red().on_black())
    .title(
        Title::from(Line::from(vec![" Now Playing ".italic()]))
            .alignment(Alignment::Center),
    )
```

joshka:
This is the intended behavior.

https://docs.rs/ratatui/latest/ratatui/widgets/block/struct.Block.html#method.border_style
> If a [Block::style](https://docs.rs/ratatui/latest/ratatui/widgets/block/struct.Block.html#method.style) is defined, border_style will be applied on top of it.

https://docs.rs/ratatui/latest/ratatui/widgets/block/struct.Block.html#method.style
[...]

PR #1190, opened 2024-06-19
docs(block): add docs about style inheritance

joshka:
Fixes: https://github.com/ratatui-org/ratatui/issues/1129

### DIFF
--- a/src/widgets/block.rs
+++ b/src/widgets/block.rs
@@ -36,6 +36,10 @@ pub use title::{Position, Title};
 /// Without left border───
 /// ```
 ///
+/// Styles are applied first to the entire block, then to the borders, and finally to the titles.
+/// If the block is used as a container for another widget, the inner widget can also be styled.
+/// Styles will be inherited by each level unless they are overridden.
+///
 /// # Examples
 ///
 /// ```
@@ -396,16 +400,36 @@ impl<'a> Block<'a> {
         self
     }
 
-    /// Defines the block style.
+    /// Defines the style of the entire block.
     ///
     /// This is the most generic [`Style`] a block can receive, it will be merged with any other
-    /// more specific style. Elements can be styled further with [`Block::title_style`] and
-    /// [`Block::border_style`].
+    /// more specific styles. Elements can be styled further with [`Block::title_style`] and
+    /// [`Block::border_style`], which will be applied on top of this style. If the block is used
+    /// as a container for another widget, the inner widget can also be styled.
     ///
     /// `style` accepts any type that is convertible to [`Style`] (e.g. [`Style`], [`Color`], or
     /// your own type that implements [`Into<Style>`]).
     ///
     /// This will also apply to the widget inside that block, unless the inner widget is styled.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use ratatui::{prelude::*, widgets::*};
+    /// let block = Block::new().style(Style::new().red().on_black());
+    ///
+    /// // To override styles set on the block level, you can style the title, border.
+    /// let block = Block::new()
+    ///     .style(Style::new().red().bold().italic())
+    ///     .border_style(Style::new().not_italic()) // will be red and bold
+    ///     .title_style(Style::new().not_bold()) // will be red and italic
+    ///     .title("Title");
+    ///
+    /// // To style the inner widget, you can style the widget itself.
+    /// let paragraph = Paragraph::new("Content")
+    ///     .block(block)
+    ///     .style(Style::new().white().not_bold()); // will be white, and italic
+    /// ```
     #[must_use = "method moves the value of self and returns the modified value"]
     pub fn style<S: Into<Style>>(mut self, style: S) -> Self {
         self.style = style.into();
--- a/src/widgets/block.rs
+++ b/src/widgets/block.rs
@@ -27,8 +27,8 @@ pub use title::{Position, Title};
 /// both centered and non-centered titles are rendered, the centered space is calculated based on
 /// the full width of the block, rather than the leftover width.
 ///
-/// Titles are not rendered in the corners of the block unless there is no border on that edge.
-/// If the block is too small and multiple titles overlap, the border may get cut off at a corner.
+/// Titles are not rendered in the corners of the block unless there is no border on that edge. If
+/// the block is too small and multiple titles overlap, the border may get cut off at a corner.
 ///
 /// ```plain
 /// ┌With at least a left border───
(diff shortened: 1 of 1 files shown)


src/widgets/block.rs
```diff
@@ -396,16 +400,36 @@ impl<'a> Block<'a> {
...
+    /// ```
+    /// # use ratatui::{prelude::*, widgets::*};
+    /// let block = Block::new().style(Style::new().red().on_black());
+    ///
+    /// // To override styles set on the block level, you can style the title, border.
+    /// let block = Block::new()
+    ///     .style(Style::new().red().bold().italic())
+    ///     .border_style(Style::new().not_italic()) // will be red and bold
+    ///     .title_style(Style::new().not_bold()) // will be red and italic
+    ///     .title("Title");
```
EdJoPaTo, line 426: I think the sentence in from of this should be phrased more clearly, but I don't have a good idea how yet.
Currently, it reads like `border_style` overrides, but actually the `.not_italic()` overrides.

src/widgets/block.rs
```diff
@@ -36,6 +36,10 @@ pub use title::{Position, Title};
 /// Without left border───
 /// ```
 ///
+/// Styles are applied first to the entire block, then to the borders, and finally to the titles.
+/// If the block is used as a container for another widget, the inner widget can also be styled.
+/// Styles will be inherited by each level unless they are overridden.
```
EdJoPaTo, line 41: This is the case here too. `overridden` how? By another style on top? no, by another style with reset or actively setting what to remove (`no_…`).

src/widgets/block.rs
joshka: Funnily, the user that was originally confused about this implementation stated they understood the explanation well. 😀.

The styles are actually overridden as specified, style by style, not element by element.

joshka: Ah I can see how what I wrote is possibly confusing. The way that one style overrides another is by combination of the parts that are specified overwriting those that are previously specified. That's the "overriding" that I meant here.

This is called our more explicitly in [`Style`](https://docs.rs/ratatui/0.26.3/ratatui/style/struct.Style.html):

> Styles represents an incremental change. If you apply the styles S1, S2, S3 to a cell of the terminal buffer, the style of this cell will be the result of the merge of S1, S2 and S3, not just S3.

joshka: Added comment in 1e7b6f47a

src/widgets/block.rs
```diff
@@ -396,16 +400,38 @@ impl<'a> Block<'a> {
...
+    ///
+    /// See [`Style`] for more information on how merging styles works.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use ratatui::{prelude::*, widgets::*};
+    /// let block = Block::new().style(Style::new().red().on_black());
+    ///
+    /// // To override styles set on the block level, you can style the title, border.
```
EdJoPaTo, line 423: The example is way more helpful than the current sentence here. I still think override is a confusing word. unset styles or add partially to override seems better to me, as that doesnt imply the style "below" is ignored then.
Alternatively, remove the sentence in front of this code block completely as the description above explains them to be applied on top of the block style already.

```suggestion
    /// // For border and title you can partially override / unset styles set on the block level.
```

joshka:
ping @ratatui-org/maintainers

src/widgets/block.rs
```diff
@@ -396,16 +400,38 @@ impl<'a> Block<'a> {
...
-    /// [`Block::border_style`].
+    /// more specific styles. Elements can be styled further with [`Block::title_style`] and
+    /// [`Block::border_style`], which will be applied on top of this style. If the block is used
+    /// as a container for another widget, the inner widget can also be styled.
     ///
     /// `style` accepts any type that is convertible to [`Style`] (e.g. [`Style`], [`Color`], or
     /// your own type that implements [`Into<Style>`]).
     ///
-    /// This will also apply to the widget inside that block, unless the inner widget is styled.
+    /// This will also apply to the widget inside that block, the inner widget overrides the style.
```
kdheepak, line 413: At some point I think we should rephrase this? This says "inside that block" and "inner widget" but that isn't how the code works and I don't think we use that language on the website docs.

kdheepak (approved):

src/widgets/block.rs
joshka: Can you suggest alternate wording for this?

kdheepak: The shortest change I can think of is:

```suggestion
    /// This will also apply to the widget that houses the block, in which case the widget's styles will be applied on top of the block's styles.
```

But I'm fine with scoping this PR to not include this kind of change because I think we'd have to read through all the docs with an eye to make them consistent.

kdheepak: Hmm, isn't the original docs and the change in this PR (and my previous comment suggestion) all incorrect?

For a `Paragraph`:

https://github.com/ratatui-org/ratatui/blob/cbfcd34ef5cfaa28f72706af373b7dfe4a648795/src/widgets/paragraph.rs#L332-L334

The code for a `Paragraph` applies the `block` style _after_ applying the paragraph's own style. I always thought it was the other way around. I haven't checked all the widgets for consistency here.

kdheepak: The `block`'s style overlaps or is "painted over" the paragraph's style. In the actual behavior, the block's style "overrides" the paragraph's style. 

I also think "overrride" implies that the block's style replaces the paragraph's style but really the block's style is superimposed on top of the paragraph's style. A paragraph style with background white and a block style with foreground red will result in "red on white" behavior, right?

joshka: This comment refers to widgets that are rendered in the Block::inner rect, not in widgets that have blocks.

> The code for a Paragraph applies the block style after applying the paragraph's own style. I always thought it was the other way around. I haven't checked all the widgets for consistency here.

This is the consistent behavior for all widgets that have a block and widget style. If this was the other way around, then styling borders would be impossible because the styles set on the widget would override the block. I have a vague recollection this is covered in an issue / PR comment from eons ago, but the behavior is correct either way.

> I also think "overrride" implies that the block's style replaces the paragraph's style but really the block's style is superimposed on top of the paragraph's style. A paragraph style with background white and a block style with foreground red will result in "red on white" behavior, right?

Override here means the same as what we're taking about in terms of merging in the style docs:

> Styles represents an incremental change. If you apply the styles S1, S2, S3 to a cell of the terminal buffer, the style of this cell will be the result of the merge of S1, S2 and S3, not just S3.

The problem with the word merge is that it doesn't imply a precedence of the items that are merged, whereas override does imply this.

kdheepak: > This comment refers to widgets that are rendered in the Block::inner rect, not in widgets that have blocks.

My bad, I don't think I caught that it is referring to `Block::inner`. I still don't quite follow what this line is supposed to mean then. The following line was just above it, so I read it as widgets that have a block.

```rust
/// If the block is used as a container for another widget, the inner widget can also be styled.
```

> The problem with the word merge is that it doesn't imply a precedence of the items that are merged, whereas override does imply this.

Maybe "overriding merge" would solve my nit here.

kdheepak: None of my comments are blocking this PR!

joshka: tweaked it a little 05389d5b8

Issue #1111, opened 2024-05-15
Inconsistent behavior between `Line::from` and `Line::raw` 


kdheepak:
## Description

`Line::from` doesn't split the spans on new lines:

```rust
let line = Line::from("hello world\ngoodbye world");
dbg!(line.spans.len()) // 1

let line = Line::raw("hello world\ngoodbye world");
dbg!(line.spans.len()) // 2
```

EdJoPaTo:
The idea of line is to have no new lines. So… not sure which is the more correct behaviour?

kdheepak:
I would expect this to be correct behaviour:

> ```
> let line = Line::raw("hello world\ngoodbye world");
> dbg!(line.spans.len()) // 2
> ```

precisely because of the reason you've stated: `Line` shouldn't have newlines.

joshka:
Two spans makes it possible to use `&strs`, one span would mean that an owned string would have to be constructed from the two lines. I think we currently just filter out the newlines when displaying the spans, so the behavior probably just works alright as-is, but going to two makes sense.

PR #1191, opened 2024-06-19
fix(line): remove newlines when converting strings to Lines

joshka:
`Line::from("a\nb")` now returns a line with two `Span`s instead of 1
Fixes: https://github.com/ratatui-org/ratatui/issues/1111

### DIFF
--- a/src/text/line.rs
+++ b/src/text/line.rs
@@ -12,6 +12,9 @@ use crate::{prelude::*, style::Styled, text::StyledGrapheme};
 /// text. When a [`Line`] is rendered, it is rendered as a single line of text, with each [`Span`]
 /// being rendered in order (left to right).
 ///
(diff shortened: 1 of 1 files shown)


joshka:
Unsure whether to mark this as breaking - thoughts?

kdheepak:
In my opinion this is a bug and should not be considered breaking. I’d be surprised if someone intentionally were relying on this behavior. I think the issue is that if someone were unwittingly relying on this behavior, would they understand it is a bug or be frustrated that their code broke in a minor release? I personally lean toward the former.

kdheepak (approved):

joshka:
Yeah, me too.I'm wary based on a recent similar assertion in unicode-width, where they changed behavior in a point release and which broke our tests. This will go out in 0.27.0, so it's in a release that can break things. I'm just unsure whether to document it as such or if the commit is enough here.

PR #1189, opened 2024-06-18
chore(examples): simplify all examples

joshka:
In <https://github.com/ratatui-org/ratatui/pull/1180> and
<https://github.com/ratatui-org/ratatui/pull/1181>, we introduced some
simpler ways to create a backend and terminal, and to setup color_eyre
to handle errors. This PR updates the examples to use these new methods,
and removes a bunch of unnecessary boilerplate code.

PR #1192, opened 2024-06-19
feat(docs): add tracing example

joshka:
Add an example that demonstrates logging to a file for:
<https://forum.ratatui.rs/t/how-do-you-println-debug-your-tui-programs/66>

```shell
cargo run --example tracing
RUST_LOG=trace cargo run --example=tracing
cat tracing.log
```

Made with VHS

### DIFF
--- a/examples/tracing.rs
+++ b/examples/tracing.rs
@@ -1,0 +1,151 @@
+//! # [Ratatui] Tracing example
+//!
+//! The latest version of this example is available in the [examples] folder in the repository.
+//!
+//! Please note that the examples are designed to be run against the `main` branch of the Github
+//! repository. This means that you may not be able to compile with the latest release version on
+//! crates.io, or the one that you have installed locally.
+//!
+//! See the [examples readme] for more information on finding examples that match the version of the
+//! library you are using.
+//!
+//! [Ratatui]: https://github.com/ratatui-org/ratatui
+//! [examples]: https://github.com/ratatui-org/ratatui/blob/main/examples
+//! [examples readme]: https://github.com/ratatui-org/ratatui/blob/main/examples/README.md
+
+// A simple example demonstrating how to use the [tracing] with Ratatui to log to a file.
+//
+// This example demonstrates how to use the [tracing] crate with Ratatui to log to a file. The
+// example sets up a simple logger that logs to a file named `tracing.log` in the current directory.
+//
+// Run the example with `cargo run --example tracing` and then view the `tracing.log` file to see
+// the logs. To see more logs, you can run the example with `RUST_LOG=tracing=debug cargo run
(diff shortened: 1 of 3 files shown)


EdJoPaTo:
Should probably wait for #1180 or #1189? Otherwise, it will result in more merge conflicts.

examples/tracing.rs
```diff
@@ -0,0 +1,151 @@
...
+// A simple example demonstrating how to use the [tracing] with Ratatui to log to a file.
+//
+// This example demonstrates how to use the [tracing] crate with Ratatui to log to a file. The
+// example sets up a simple logger that logs to a file named `tracing.log` in the current directory.
+//
+// Run the example with `cargo run --example tracing` and then view the `tracing.log` file to see
+// the logs. To see more logs, you can run the example with `RUST_LOG=tracing=debug cargo run
+// --example`
+//
+// [tracing]: https://crate.io/crates/tracing
```
EdJoPaTo, line 25: `tracing` is a crate cargo knows of. It should be able to figure this out by itself.

```suggestion
// A simple example demonstrating how to use the [`tracing`] with Ratatui to log to a file.
//
// This example demonstrates how to use the [`tracing`] crate with Ratatui to log to a file. The
// example sets up a simple logger that logs to a file named `tracing.log` in the current directory.
//
// Run the example with `cargo run --example tracing` and then view the `tracing.log` file to see
// the logs. To see more logs, you can run the example with `RUST_LOG=tracing=debug cargo run
// --example`
```

orhun (approved):
Looks good! I suggest mentioning [`tui-logger`](https://docs.rs/tui-logger) somewhere in the docs as well. It greatly reduces the pain of designing custom widgets to display/filter the tracing logs!

examples/tracing.rs
joshka: In VSCode, clicking a link like this will navigate to the lib.rs file in the crate. This isn't where I'd like that like that interaction to end up. I want users to end up on the crate page, which information that would help someone decide when to use / not use this for their logging purposes.

(there is however a typo `s/crate.io/crates.io/` that I'll fix)

joshka:
> Should probably wait for #1180 or #1189? Otherwise, it will result in more merge conflicts.

I'm fine with fixing up merge conflicts later. I'd prefer that over leaving a large amount of things which are good enough in a pending state.

examples/tracing.rs
joshka: No. The root module of a crate is different to the crate. There's no bug here. They are two separate things. I'd always expect someone clicking a link to ratatui in code to navigate to the ratatui module which is defined in lib.rs. The same way that clicking some::type navigates to where that type is defined in some.rs.

PR #1198, opened 2024-06-24
feat(table): select first, last, etc to table state

robertpsoane:
Add select_previous, select_next, select_first & select_last to TableState

Used equivalent API as in ListState

### DIFF
--- a/src/widgets/table/table.rs
+++ b/src/widgets/table/table.rs
@@ -613,6 +613,14 @@ impl StatefulWidgetRef for Table<'_> {
             return;
         }
 
+        if state.selected.is_some_and(|s| s >= self.rows.len()) {
+            state.select(Some(self.rows.len().saturating_sub(1)));
+        }
+
+        if self.rows.is_empty() {
+            state.select(None);
+        }
+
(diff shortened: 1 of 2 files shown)


orhun (approved):

src/widgets/table/table.rs
```diff
@@ -1016,6 +1024,96 @@ mod tests {
         assert_eq!(table.widths, vec![Constraint::Percentage(100)], "vec ref");
     }
 
+    #[cfg(test)]
+    mod state {
```
joshka, line 1028: this (mostly) belongs in table_state.rs - the exceptions are how the rendering affects the state, which belongs here. (List hasn't been split into multiple files - yet)

joshka (approved):

src/widgets/table/table.rs
robertpsoane: Have moved the first set of tests across - should the latter two stay here?

joshka: The other two are fine. Thanks for fixing this up

joshka (approved):

PR #1203, opened 2024-06-26
chore: replace big_text with hardcoded logo

joshka:
big_text.rs was a copy of the code from tui-big-text and was getting
gradually out of sync with the original crate. It was also rendering
something a bit different than the Ratatui logo. This commit replaces
the big_text.rs file with a much smaller string representation of the
Ratatui logo.

demo2


orhun:
Q: Why not use `tui-big-text` instead of `big_text.rs`? Is it because we want to avoid a dependency?

joshka:
> Q: Why not use `tui-big-text` instead of `big_text.rs`? Is it because we want to avoid a dependency?

tui-big-text depends on ratatui from crates.io. Afaik, it's not possible to compile a version that works with the version compiled from source when compiling examples as the widget traits don't match even though they are the same.

kdheepak:
I think number of pixels that change in a gif affects the size. Maybe decrease the randomness a tad? Alternatively, you can record a `.mov` and that will render on github.

joshka:
> I think number of pixels that change in a gif affects the size. Maybe decrease the randomness a tad? Alternatively, you can record a `.mov` and that will render on github.

Yep, this is one of the parameters I've been tweaking and which also has non-linear behavior
The problem is not how to make this work, it's finding the motivation to sit and repeatedly tweak things until they're just right with each iteration taking ~60s to edit, run, evaluate. "Just right" in this context is a GIF that is
- 15-30s
- <10MB
- animation is subjectively not too fast (to not be too jarring)
- has to end with the text fully visible
- should not start immediately

I don't like MOVs in this context as they tend to break (with a can't render error) on github often for some reason. That said, this might be a valid fall back position.

I'd suggest running VHS locally `vhs examples/vhs/demo2-destroy.tape` and checking whether the output looks good enough (I changed the Ratatui text color to white from Red).

demo2-destroy

Note the other thing I noticed is that there's a bit of extra space on the top and bottom of the frame. Not sure exactly what caused that however. Could be anything from OS update, font installation, ttyd change, vhs change, ?

orhun:
I feel like we can tweak:
- the delay before the animation starts (I feel like it is taking a lot of time)
- speed up the animation a bit maybe
- width & height

For me the VHS output looks like this: https://paste.orhun.dev/NiPi8p.gif

joshka:
Filesize for me is now < 10MB. @orhun can you run this on your machine and check (I haven't yet diagnosed why I'm seeing changed sizes to the border)

orhun:
Yup, perfect: https://p.orhun.dev/lhen9G.gif (9MB)

I think this is ready to go 🚀

orhun (approved):

joshka:
> Yup, perfect: https://p.orhun.dev/lhen9G.gif (9MB)
> 
> I think this is ready to go 🚀

Ha that's also got a bit of weirdness happening around the top and bottom now too (but it looks fun regardless)

lhen9G

joshka:
I installed Jetbrains Mono ([default font for VHS](https://github.com/charmbracelet/vhs/blob/5a207ef92d81b671c9d0f30ddb245434cb49a2f8/vhs.go#L63-L73)) which fixes up the render height issue for me. I must have had that installed but uninstalled it at some point to test something.

Issue #914, opened 2024-02-04
Add better docs for how to use block in block.rs


joshka:
## Problem
Similar to https://github.com/ratatui-org/website/issues/424 which asks this to be added to the website the blocks docs https://docs.rs/ratatui/latest/ratatui/widgets/block/struct.Block.html could add some more details about using it to surround widgets by passing it as a parameter.

The main doc comment needs to have several summary list similar to the other newer widgets that shows the setter methods that are used to configure the widget, the constructors, and those that are used after the widget is created (e.g. inner()), pointing to the appropriate methods lower.

## Solution
Update the docs

PR #1206, opened 2024-06-27
docs(block): update block documentation

leohscl:
Update block documentation with constructor methods and setter methods in the main doc comment Added an example for using it to surround widgets

Solves this: https://github.com/ratatui-org/ratatui/issues/914

### DIFF
--- a/src/widgets/block.rs
+++ b/src/widgets/block.rs
@@ -38,9 +38,6 @@ pub use title::{Position, Title};
 /// # Constructor methods
 ///
 /// - [`Block::new`] creates a new [`Block`] with no border or paddings.
-/// - [`Block::border_style`] Defines the style of the borders.
-/// - [`Block::border_type`] Sets the symbols used to display the border (e.g. single line, double
-///   line, thick or rounded borders).
 /// - [`Block::bordered`] Create a new block with all borders shown.
 ///
 /// # Setter methods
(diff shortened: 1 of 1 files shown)


src/widgets/block.rs
```diff
@@ -35,6 +35,30 @@ pub use title::{Position, Title};
...
+/// - [`Block::new`] creates a new [`Block`] with no border or paddings.
+///
+/// # Setter methods
+///
+/// These methods are fluent setters. They return a new [`Block`] with the specified property set.
+///
+/// - [`Block::border_style`] Defines the style of the borders.
+/// - [`Block::border_type`] Sets the symbols used to display the border (e.g. single line, double
+///   line, thick or rounded borders).
+/// - [`Block::bordered`] Create a new block with all borders shown.
```
joshka, line 49: move thie to the constructor section

joshka (approved):
LGTM - 1 minor change if you have a sec

src/widgets/block.rs
```diff
@@ -38,15 +38,15 @@ pub use title::{Position, Title};
 /// # Constructor methods
 ///
 /// - [`Block::new`] creates a new [`Block`] with no border or paddings.
+/// - [`Block::border_style`] Defines the style of the borders.
+/// - [`Block::border_type`] Sets the symbols used to display the border (e.g. single line, double
+///   line, thick or rounded borders).
```
joshka, line 43: Can you please move these lines back down - my comment was made in error. I meant only the `bordered` function to be move - apologies.

joshka (changes requested):

src/widgets/block.rs
leohscl: Yes of course !

joshka (approved):
Thanks for the PR :)

PR #1208, opened 2024-06-27
chore(deps): use instabilty instead of stabilty

joshka:
https://github.com/ratatui-org/instability


joshka:
Forked stability to instability and applied the PR which enables use statements to be marked as unstable. This means that we can make entire modules unstable (we need this for the AnsiBuffer PR to work properly).

kdheepak (approved):

PR #1209, opened 2024-06-28
feat(terminal): add Terminal::try_draw() method

joshka:
This makes it easier to write fallible rendering methods which can use the `?` operator


src/terminal/terminal.rs
```diff
@@ -234,39 +234,142 @@ where
         Ok(())
     }
 
-    /// Synchronizes terminal size, calls the rendering closure, flushes the current internal state
-    /// and prepares for the next draw call.
+    /// Draws a single frame to the terminal.
     ///
-    /// This is the main entry point for drawing to the terminal.
+    /// Returns a [`CompletedFrame`] if successful, otherwise a [`std::io::Error`].
```
kdheepak, line 239:
```suggestion
    /// Returns a Ok([`CompletedFrame`]) if successful, otherwise a [`std::io::Error`].
```

kdheepak (approved):
Cool idea! LGTM! Curious what other people think.

src/terminal/terminal.rs
joshka:
```rust
    /// Returns [`Result::Ok`] containing a [`CompletedFrame`] if successful, otherwise
    /// [`Result::Err`] containing the [`std::io::Error`] that caused the failure.
```

joshka:
> Cool idea! LGTM! Curious what other people think.

@EdJoPaTo ?

src/terminal/terminal.rs
EdJoPaTo: Result is a basic concept of Rust and this seems a bit too much explanation of that. I would just go with CompletedFrame or Error. Maybe "Returns a Result containing either …".

src/terminal/terminal.rs
```diff
@@ -234,39 +234,143 @@ where
...
+    ///     let value: u8 = "not a number".parse().map_err(io::Error::other)?;
+    ///     let area = frame.size();
+    ///     frame.render_widget(Paragraph::new("Hello World!"), area);
+    ///     frame.set_cursor(0, 0);
+    ///     io::Result::Ok(())
+    /// })?;
+    ///
+    /// // or with a function
+    /// terminal.try_draw(render)?;
+    ///
```
EdJoPaTo, line 355: Either empty lines between both or no empty lines. It took me a moment to realise where render is defined as the comment looked like its somewhere completely else.

src/terminal/terminal.rs
```diff
@@ -234,39 +234,143 @@ where
...
+        })
+    }
+
+    /// Tries to draw a single frame to the terminal.
+    ///
+    /// Returns [`Result::Ok`] containing a [`CompletedFrame`] if successful, otherwise
+    /// [`Result::Err`] containing the [`std::io::Error`] that caused the failure.
+    ///
+    /// This is the equivalent of [`Terminal::draw`] but accepts a function or closure that returns
+    /// a `Result` instead of nothing.
```
EdJoPaTo, line 305: Everything after this repeats the same docs again. Changing a comment in the future might miss the same comment in the other method. I would just remove it as this comment links to the draw method explaining it with more details already.

EdJoPaTo (commented):
This simplifies the rendering but I'm not sure how useful it actually might be. Ideally the same Render Widget or whatever its named trait should exist here too I think? That would reduce the differences. Especially as a context is of interest anyway and Frame would reduce the need for the context too.
And the current trait doesn't allow for a Result currently either? Not sure what the ideal way would be.

src/terminal/terminal.rs
joshka: Are you looking at the rendered docs?

image

src/terminal/terminal.rs
joshka: That's intentional - I want the docs to be useful for those that read them in their LSP without having to navigate to another method.

src/terminal/terminal.rs
joshka: I checked the std lib and it varies in the explanation. This is one of the more verbose versions copied and paraphrased from one of the std lib methods (In string / vec / result).

joshka:
> This simplifies the rendering but I'm not sure how useful it actually might be. Ideally the same Render Widget or whatever its named trait should exist here too I think? That would reduce the differences. Especially as a context is of interest anyway and Frame would reduce the need for the context too.
> And the current trait doesn't allow for a Result currently either? Not sure what the ideal way would be.

I regularly find myself using methods which return results and having to throw away the possible errors inside of the draw method by either crashing or ignoring them. This would make it possible to add more context to the errors (using anyhow / eyre), and have the app more gracefully crash.

This is a first point in making some of these future things more useful. We don't need to solve the entire problem, but this is a piece that is fine and necessary on its own.

src/terminal/terminal.rs
```diff
@@ -234,39 +234,143 @@ where
...
+    /// }
+    /// # io::Result::Ok(())
     /// ```
-    pub fn draw<F>(&mut self, f: F) -> io::Result<CompletedFrame>
+    pub fn draw<F>(&mut self, render: F) -> io::Result<CompletedFrame>
     where
         F: FnOnce(&mut Frame),
+    {
+        self.try_draw(|frame| {
+            render(frame);
```
kdheepak, line 294: Counterintuitively, I would have preferred the name of this function to be `f` instead of `render`. 

When I first saw this function in the PR I was like "huh, I wonder where `render` is defined, I should search the file or look in the imports". If it were named `f` (or `render_callback`?) I would have immediately looked in the function arguments to begin with. 

I'm sharing this comment not because I want change this, but because I thought it was counterintuitive and hence interesting. I think it's one of those "variables with smaller names implies smaller scope" kind of effect.

joshka: Yeah - this was a change I wasn't 100% sure about. render_callback makes more sense. The scope of this is anything that can call this method - so a single letter is a bad name. I think ed would probably agree with that.

src/terminal/terminal.rs
```diff
@@ -234,39 +234,143 @@ where
...
+    ///     let value: u8 = "not a number".parse().map_err(io::Error::other)?;
+    ///     frame.render_widget(Paragraph::new("Hello World!"), frame.size());
+    ///     Ok(())
+    /// }
+    /// # io::Result::Ok(())
+    /// ```
+    pub fn try_draw<F, E>(&mut self, render: F) -> io::Result<CompletedFrame>
+    where
+        F: FnOnce(&mut Frame) -> Result<(), E>,
+        E: Into<io::Error>,
```
kdheepak, line 366: Is `io::Error` the right error here? I'm looking at the variants in [`ErrorKind`](https://doc.rust-lang.org/std/io/enum.ErrorKind.html) and it doesn't feel like a lot of these variants will be used. Is there a CustomError type we should introduce for this? Can we make it `Box<dyn Err>`?

kdheepak: Alternatively, we can introduce this as experimental until we explore this more?

joshka: All of the crossterm errors are io Errors. There is an io::Error::Other that can be used to represent anything that doesn't fit. try_draw and draw shouldn't diverge in their error types, as the complexity is unwarranted and would be unpredictable / weird.

I've thought about someday introducing a proper error type for ratatui, but this is orthogonal to this. Because most apps will use draw (due to inertia), we are no worse off by adding try_draw with io::Error (it locks us into a error type that we're already locked into).

Put another way, draw has to return io::Error, and the things which try_draw call return that too. There's nothing we'd gain by using some other error. Note - the E: Into<io::Error> means that all we're saying is that the error that the callback returns must be convertible into the errors that we need as input and output of the various parts.

PR #1213, opened 2024-06-29
fix: ScrollbarOrientation should probably be Copy too.

thscharler:
This seems to be some oversight, and is just slightly annoying to use.

### DIFF
--- a/src/layout/constraint.rs
+++ b/src/layout/constraint.rs
@@ -310,7 +310,7 @@ impl Constraint {
     /// ```rust
     /// # use ratatui::prelude::*;
     /// # let area = Rect::default();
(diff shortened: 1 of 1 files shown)


joshka (approved):

joshka:
This introduces a couple of clippy errors where this is passed as a reference. Fixing that wil be a breaking change (which we document, etc.) As we've just released 0.27.0, there might be some time until this gets merged. I'd suggest slapping an allow attribute for the lint on the broken code now and come back and fix those later before 0.28.0. Alternatively leave this until closer to 0.28.0 and do both at once.

EdJoPaTo (approved):
Yeah, allow the breaking changes now and fix them in their own breaking change PRs (which should be merged directly before the breaking release)

PR #1214, opened 2024-07-01
chore(changelog): fixup git-cliff formatting

joshka:
- Replaced backticks surrounding body with blockquotes. This formats
  significantly better.
- Added a few postprocessors to remove junk like the PR template,
  horizontal lines and some trailing whitespace

---

Note: there is extra non-automatically generated stuff in the changelog that this would remove - the changes to CHANGELOG.md should not be merged as-is, and this is worth waiting for @orhun to check out.

Compare: https://github.com/ratatui-org/ratatui/blob/5e7fbe8c328050a282a49b8c6ab3f5b33c7cfb4c/CHANGELOG.md
To: https://github.com/ratatui-org/ratatui/blob/main/CHANGELOG.md


joshka:
> It seems to be missing a summary section

Yep - I mentioned this in the PR comment and the reason for the do not merge label.

> Note: there is extra non-automatically generated stuff in the changelog that this would remove - the changes to CHANGELOG.md should not be merged as-is, and this is worth waiting for @orhun to check out.

CHANGELOG.md
orhun, line 1: We can revert the changes in CHANGELOG.md and start using this new format for the upcoming releases. There are already old formats in the changelog since we manually append it each time anyways.

cliff.toml
```diff
@@ -87,6 +85,12 @@ trim = false
 footer = """
 <!-- generated by git-cliff -->
 """
+postprocessors = [
+  # { pattern = '<REPO>', replace = "https://github.com/joshka/focusable" },                                # replace repository URL
```
orhun, line 89: not needed?
```suggestion
```

orhun (approved):

cliff.toml
joshka: woops - bad copy paste

CHANGELOG.md
joshka: Makes sense, though the manual part of this feeds into the issue about releasing more frequently and less manually.

orhun: Yup, commented in #1232 about this.

Also, we can use the following command for appending the unreleased to the changelog without changing the previous entries:

```sh
git cliff -l -p CHANGELOG.md
```

orhun (approved):

joshka:
Thanks! @orhun

PR #1217, opened 2024-07-05
docs: Constraints::from_fills docs correction

EmiOnGit:
Quick fix for the documentation of the `from_fills` method. 
Should be uncontroversial


joshka (approved):
Thanks!

PR #1218, opened 2024-07-06
feat(calendar): add chrono as a backend for calendar widget

InnocentZero:


InnocentZero:
This is weird. Why are the commits unverified? 

image

orhun:
Hmm, for some reason I'm getting the following panic for the `calendar_chrono` example:

```
thread 'main' panicked at src/buffer/buffer.rs:221:18:
                                                      Trying to access position outside the buffer: x=160, y=58, area=Rect { x: 0, y: 0, width: 213, height: 58 }
```

`calendar` example works fine.

Apparently this is a terminal size problem. Maybe it makes sense to add a VHS tape (`calendar_chrono.tape`) for reproducibility.

joshka:
> Apparently this is a terminal size problem. Maybe it makes sense to add a VHS tape (`calendar_chrono.tape`) for reproducibility.

We definitely need to continue working towards making Ratatui panic-free where possible.

orhun:
Yes, of course. I meant creating a VHS tape would be useful to test this out on other platforms under the same circumstances. I'm not sure if the panic is related to my configuration system or something else is going on.

src/widgets/calendar.rs
```diff
@@ -181,24 +237,86 @@ impl<DS: DateStyler> Monthly<'_, DS> {
             y += 1;
         }
     }
+    #[cfg(feature = "widget-calendar-chrono")]
+    fn render_monthly(&self, area: Rect, buf: &mut Buffer) {
```
EdJoPaTo, line 241: Seems like a bunch of duplicated code. Should either use more specific `cfg(feature` or duplicate them completely (which would also allow for `--all-features` again).

src/widgets.rs
```diff
@@ -24,8 +24,12 @@
 mod barchart;
 pub mod block;
 mod borders;
-#[cfg(feature = "widget-calendar")]
+#[cfg(any(feature = "widget-calendar", feature = "widget-calendar-chrono"))]
 pub mod calendar;
+#[cfg(all(feature = "widget-calendar", feature = "widget-calendar-chrono"))]
+compile_error!(
+    "These two feature flags are actually conflicting. You are supposed to only use one of them"
+);
```
EdJoPaTo, line 32: This is horrible for trying to check as much code as possible with `--all-features`. (I know it's already a problem on Windows, but we don't need to make this even harder)

EdJoPaTo (commented):
What's the status of this?

Personally I would be against the current approach. It creates a lot of duplicated code and prevents `--all-features`.

src/widgets.rs
joshka: Yeah, I agree here. I wonder if there's a way we can support both time and chrono in the calendar widget?

I want to soften the language here a bit to make sure that @InnocentZero doesn't get offended by this being called "horrible". I think Ed means this as a shorthand for this causing unintended consequences in a way that would be frustrating when developing. We often will use the `--all-features` flag to make it easy to make changes that build all the code. Taking away that ability makes it more difficult to develop the library.

Issue #1044, opened 2024-04-17
`render` methods should be passed context instead of buffer


kdheepak:
## Problem

Currently, there's no way for the render method to know additional contextual information that would be useful for widgets.

For example, frame count is currently unavailable to the render method.

## Solution

While `WidgetRef` and `StatefulWidgetRef` are unstable, I'd like to propose a `Context` instead of buffer.

```
 pub trait WidgetRef { 
     fn render_ref(&self, area: Rect, ctx: &mut Context); 
 } 
```

`Context` can be defined like so:

```rust
pub struct Context {
    buffer: Buffer
}
```

We can even have `Context` have the similar methods as `Buffer`, and instead of

```rust
            buf.get_mut(area.left(), area.top())
                .set_symbol(border_top_left)
                .set_style(border_style);
```

widget authors would write the following instead:

```rust
            ctx.get_buf_mut(area.left(), area.top())
                .set_symbol(border_top_left)
                .set_style(border_style);
```

The main advantage of this is that we can add features to `Context` in the future without breaking backward compatibility.

For example, `Context` can contain a frame count.

The main motivation for this idea was having a `Palette` in `Context`. For example, if `Context` is defined as this:

```rust
pub struct Context {
    buffer: Buffer,
    frame_count: usize,
    palette: Palette,
}
```

Imagine a `Palette` that is defined like so:

```rust
struct Palette {
    base: Color,
    primary: Color,
    secondary: Color,
    overlay0: Color,
    overlay1: Color,
    overlay2: Color,
    subtext0: Color,
    subtext1: Color,
    surface0: Color,
    surface1: Color,
    surface2: Color,
    text: Color,
}

impl Palette {
    fn catppuccin() -> Self {
        Palette {
            base: Color::from_str("#eff1f5"),
            primary: Color::from_str("#04a5e5"),
            secondary: Color::from_str("#209fb5"),
            overlay0: Color::from_str("#9ca0b0"),
            overlay1: Color::from_str("#8c8fa1"),
            overlay2: Color::from_str("#7c7f93"),
            subtext0: Color::from_str("#6c6f85"),
            subtext1: Color::from_str("#5c5f77"),
            surface0: Color::from_str("#ccd0da"),
            surface1: Color::from_str("#bcc0cc"),
            surface2: Color::from_str("#acb0be"),
            text: Color::from_str("#4c4f69"),
        }
    }
}
```

[...]

PR #1219, opened 2024-07-07
feat: add cursor_position to Buffer

thscharler:
Maybe one way to go with #1044 is to recognize that there is already 
a context struct passed into render, and it's named Buffer. 

I gave this a try and added cursor_pos, and it doesn't fit too badly. 

--

The current state is quite a chore, when writing an application with more then 3 
text-input fields. You have to collect all the possible cursor-positions, check 
against the current focus and set the cursor. and all this in the application code.

### DIFF
--- a/src/buffer/buffer.rs
+++ b/src/buffer/buffer.rs
@@ -48,6 +48,8 @@ pub struct Buffer {
     /// The content of the buffer. The length of this Vec should always be equal to area.width *
     /// area.height
     pub content: Vec<Cell>,
+    /// The cursor shown for this buffer.
+    pub cursor_position: Option<(u16, u16)>,
 }
 
 impl Buffer {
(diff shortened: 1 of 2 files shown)


joshka:
As a brief initial reaction, I'm not certain I like this idea, because it feels like we're just pushing more random stuff into the buffer without a strong plan for how this looks and thinking about where things belong. I need to give this some proper thought, but I wanted to make sure that this was noted for the other devs considering this PR.

thscharler:
Fine with me. 

One more thought why it might be fitting: 
I create a secondary buffer, eg for a viewport and hand it down somewhere to render. Then I relocate the result and copy it in the main buffer. In this case it would be necessary to remap the cursor position too. 

In the end there is a link between the buffer and the cursor, as both directly contribute to what the terminal displays. 

But I agree that this is not a general solution for other context values.

joshka:
I'm going to close this PR out as it's not quite the right fit (at least for now). We might revisit this at some point in the future though as things evolve. It was a good thought provoking idea.

thscharler:
> I'm going to close this PR out as it's not quite the right fit (at least for now). We might revisit this at some point in the future though as things evolve. It was a good thought provoking idea.

Ok for me.

Issue #1014, opened 2024-04-01
Add vhs tapes and images for the new constraint explorer / flex examples


joshka:
## Problem
The examples readme is missing images for the new examples (flex and constraints explorer)
There is a VHS tape for flex, but not for the explorer.

## Solution

- Generate / update the tapes
- Check if there are other missing tapes
- Generate the images (gifs / pngs)
- Add the images to the images branch of this repo
- Update the readme to point at the new images

## Constraints

- Match the sizes and font from the other tapes (adjust height to make sense, but width should be generally consistent)
- Images should generally not be frantic (lots of fast animation). Always leave enough time for people to read any changes. (usually around 2s or so after any significant animation seems to work)
- Images need to be under 5MB to be able to be able to be displayed on github.
- Use a PNG if that conveys all the info, use a GIF if the animation will be useful.

orhun:
Maybe a good first issue?

joshka:
> Maybe a good first issue?

Added some constraints to make it clearer and yep

PR #1223, opened 2024-07-08
docs(examples): add missing images

kibibyt3:
Missing example images added to the images fork to help address #1014.


joshka (approved):
Thanks for doing this.

PR #1225, opened 2024-07-09
docs(examples): add missing examples to README

### LINKED ISSUE #1014: Add vhs tapes and images for the new constraint explorer / flex examples

kibibyt3:
Resolves #1014.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -141,8 +141,6 @@ of the VHS tape.
 
 <https://github.com/ratatui-org/ratatui/assets/381361/485e775a-e0b5-4133-899b-1e8aeb56e774>
 
-![Constraint Explorer][constraint-explorer.gif]
-
 ## Constraint Explorer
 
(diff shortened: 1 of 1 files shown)


joshka (approved):
Thanks again

PR #1226, opened 2024-07-12
fix(buffer): dont render control characters

### LINKED ISSUE #1211: Visual glitch - Null character inside block wrapped component alters borders
## Description
G'day,

This one took me a while to investigate in a project so thought I'd report here in case it helps someone else.
When a block wrapped widget contains text with a null character (eg `\0`), and is styled (e.g colour or bold), it can cause a visual glitch where outter block borders do not align correctly.

## To Reproduce
Reproducible example located here: https://github.com/nick42d/ratatuirepro

## Expected behavior
Expect inner widget contents not to impact outter block - however note that the cause could be a technical limitation of the terminal that I'm not aware of.

## Screenshots
image

## Environment
- OS: Linux
- Terminal Emulator: tried Kitty, Alacritty, Xterm
- Font: JetBrains Nerd Font Mono
- Crate version: 0.27
- Backend: crossterm, also tried termwiz

EdJoPaTo:
fixes #1211

As I have the stuff tested in a local branch, I can also push it for more discussions.

See https://doc.rust-lang.org/std/primitive.char.html#method.is_control

### DIFF
--- a/src/buffer/buffer.rs
+++ b/src/buffer/buffer.rs
@@ -192,7 +192,7 @@ impl Buffer {
     }
 
     /// Print at most the first n characters of a string if enough space is available
-    /// until the end of the line.
+    /// until the end of the line. Skips zero-width graphemes and control characters.
     ///
     /// Use [`Buffer::set_string`] when the maximum amount of characters can be printed.
     pub fn set_stringn<T, S>(
--- a/src/buffer/buffer.rs
+++ b/src/buffer/buffer.rs
@@ -210,6 +210,7 @@ impl Buffer {
         let max_width = max_width.try_into().unwrap_or(u16::MAX);
         let mut remaining_width = self.area.right().saturating_sub(x).min(max_width);
         let graphemes = UnicodeSegmentation::graphemes(string.as_ref(), true)
+            .filter(|symbol| !symbol.contains(|char: char| char.is_control()))
             .map(|symbol| (symbol, symbol.width() as u16))
             .filter(|(_symbol, width)| *width > 0)
             .map_while(|(symbol, width)| {
@@ -930,5 +931,36 @@ mod tests {
         buf.set_string(0, 0, "foo", Style::new().red());
         buf.set_string(0, 1, "bar", Style::new().blue());
         assert_eq!(buf, Buffer::with_lines(["foo".red(), "bar".blue()]));
+    }
+
+    #[test]
+    fn control_sequence_rendered_full() {
+        let text = "I \x1b[0;36mwas\x1b[0m here!";
+
+        let mut buffer = Buffer::filled(Rect::new(0, 0, 25, 3), Cell::new("x"));
+        buffer.set_string(1, 1, text, Style::new());
+
+        let expected = Buffer::with_lines([
+            "xxxxxxxxxxxxxxxxxxxxxxxxx",
+            "xI [0;36mwas[0m here!xxxx",
+            "xxxxxxxxxxxxxxxxxxxxxxxxx",
+        ]);
+        assert_eq!(buffer, expected);
+    }
+
+    #[test]
+    fn control_sequence_rendered_partially() {
+        let text = "I \x1b[0;36mwas\x1b[0m here!";
+
+        let mut buffer = Buffer::filled(Rect::new(0, 0, 11, 3), Cell::new("x"));
+        buffer.set_string(1, 1, text, Style::new());
+
+        #[rustfmt::skip]
+        let expected = Buffer::with_lines([
+            "xxxxxxxxxxx",
+            "xI [0;36mwa",
+            "xxxxxxxxxxx",
+        ]);
+        assert_eq!(buffer, expected);
     }
 }


joshka (approved):
This seems inherently reasonable to me. I don't see any obvious downsides. Are there any that would make you not want to merge this as-is?

Perhaps documenting the behavior somewhere user facing could be a good idea and marking it as a breaking change, but otherwise, seems like a good idea to do this.

EdJoPaTo:
> marking it as a breaking change

I don't think this is a breaking change in its current form. It's a bug fix restoring old behavior. Before `unicode-width` reported these control characters as width 0 which then got filtered out.

https://github.com/ratatui-org/ratatui/blob/935a7187c273e0efc876d094d6247d50e28677a3/src/buffer/buffer.rs#L213-L214

`unicode-width` changed their behavior (whether it's considered breaking or not is debatable as it's as far as I understand according to Unicode specs, so it's only a fix to be spec compliant) and this restores the behavior of before: filtering them out.
So it's basically preventing a breaking change for users of ratatui.

---

> Are there any that would make you not want to merge this as-is?
> 
> Perhaps documenting the behavior somewhere user facing could be a good idea

- This does not include documentation at all and I dont really know how to document something that restores old behaviour.
- Is this the same set of control characters which were width == 0 before? So is it actually restoring the exact old behaviour or is it introducing other bugs now?
- https://github.com/unicode-rs/unicode-width/pull/45 suggests following Unicode to render a fallback glyph instead to notify the user of a missing element here being not rendered. Which would be both the more correct approach according to Unicode and the more helpful one to notice that there are more characters than visible.

I think I would like the last one the most. Rendering some fallback symbol instead. But this PR currently does something different.

joshka:
>Is this the same set of control characters which were width == 0 before? So is it actually restoring the exact old behaviour or is it introducing other bugs now?

I read the spec and the two code paths and they are equivalent. 0x00..0x20 and 0x7F..0xA0 are treated as control characters and are length 1.

>https://github.com/unicode-rs/unicode-width/pull/45 suggests following Unicode to render a fallback glyph instead to notify the user of a missing element here being not rendered. Which would be both the more correct approach according to Unicode and the more helpful one to notice that there are more characters than visible.

>I think I would like the last one the most. Rendering some fallback symbol instead. But this PR currently does something different.

I think that's a good idea also, but I think it would be worth making an application configurable option. So just filtering seems like a good default.

> I don't think this is a breaking change in its current form. It's a bug fix restoring old behavior. Before unicode-width reported these control characters as width 0 which then got filtered out.

👍 

> This does not include documentation at all and I dont really know how to document something that restores old behaviour.

What about a one line mention in `Terminal::draw`'s doc comment?

EdJoPaTo:
> > > [unicode-rs/unicode-width#45](https://github.com/unicode-rs/unicode-width/pull/45) suggests following Unicode to render a fallback glyph instead to notify the user of a missing element here being not rendered. Which would be both the more correct approach according to Unicode and the more helpful one to notice that there are more characters than visible.
> >
> > I think I would like the last one the most. Rendering some fallback symbol instead. But this PR currently does something different.
> 
> I think that's a good idea also, but I think it would be worth making an application configurable option. So just filtering seems like a good default.

Let's go with restoring the old behavior for now: filtering them out completely and release a bug fix release `0.27.1` with this.

In the long run… Configuring this would require a Buffer configuration which seems annoying? Maybe set_stringn should be moved out of the Buffer then? Anyway, that's a longer run question. Maybe create another issue for this to discuss this further…

---

> > This does not include documentation at all and I dont really know how to document something that restores old behaviour.
> 
> What about a one line mention in `Terminal::draw`'s doc comment?

It seems far away from actually drawing something to the terminal but maybe that's a technical view… Any good ideas for such a one line?

joshka:
> It seems far away from actually drawing something to the terminal but maybe that's a technical view… Any good ideas for such a one line?

My bad - I was thinking about the setting being needed on Terminal to control the buffer. Probably on set_string / set_stringn would be better for this. Just "This method filters out (does not output) control characters" or something similar.

EdJoPaTo:
A lot is depending on set_stringn which doesn't document this either, but I think it's fine.

And while at it… set_line and set_span should probably be deprecated (in their own pr) in favor of Line as a widget in order to simplify the Buffer API to its core functionality.

EdJoPaTo:
Anything other needed for a 0.27.1 release?

joshka (approved):

joshka:
> Anything other needed for a 0.27.1 release?

@orhun is away for a couple of weeks. He usually handles all the release activities (there's a bunch of stuff that's not automated), I think it'll be around the 26th or so.

EdJoPaTo:
I'm not sure what the point of the labels accepted or review needed is, as that's something GitHub itself already displays?

Also, why accepted and not just merged then?

---

I think there should be automated quick released features and bug fixes… so versions like ratatui 0.27.7 could easily be the case… Similar to the weekly alpha release currently which seem a bit weird to me as the Cargo.toml could also just point to the main branch. The point of PR reviews is to ensure that it works correctly when merged, so it should also be releasable? Breaking PRs can be merged at the same time and released together.
That way the current bus factor is reduced and merged stuff ends up quicker with the users?

joshka:
I'm using labels mostly as an aid to whether I need to look at an issue again at some point. I didn't merge this one in case there was anything you wanted to do before merge. Feel free to merge if there's nothing pressing. 

Agreed on the quick releases etc. we're not there right now, but are close. The alpha release existed when there was a much larger barrier to releasing and to make it possible for downstream apps to easily publish crates that relied on unreleased functionality. I'm a fan of releasing whenever there's a reasonable amount of stuff to go however. Every few days if needed. When Orhun is back let's discuss making that easier. (Perhaps cut an issue or discussion about it)

EdJoPaTo:
> I'm using labels mostly as an aid to whether I need to look at an issue again at some point.

This is selectable by queries (and the buttons at the top create them for you)
https://github.com/ratatui-org/ratatui/pulls?q=is%3Apr+is%3Aopen+-reviewed-by%3A%40me



> Agreed on the quick releases etc. we're not there right now, but are close. The alpha release was created to address the gap when there was a much longer period between releasing and to make it possible for downstream apps to easily publish crates that relied on unreleased functionality.
> 
> However I'm a fan of releasing whenever there's a reasonable amount of stuff to go however. Every few days if needed. When Orhun is back let's discuss making that easier. (Perhaps cut an issue or discussion about it)

→ #1232

PR #1236, opened 2024-07-22
feat(text): Add `Add` and `AddAssign` implementations for `Line`, `Span`, and `Text`

joshka:
This enables:

```rust
let line = Span::raw("Red").red() + Span::raw("blue").blue();
let line = Line::raw("Red").red() + Span::raw("blue").blue();
let line = Line::raw("Red").red() + Line::raw("Blue").blue();
let text = Line::raw("Red").red() + Line::raw("Blue").blue();
let text = Text::raw("Red").red() + Line::raw("Blue").blue();

let mut line = Line::raw("Red").red();
line += Span::raw("Blue").blue();

let mut text = Text::raw("Red").red();
text += Line::raw("Blue").blue();

line.extend(vec![Span::raw("1"), Span::raw("2"), Span::raw("3")]);
```

### DIFF
--- a/src/text/line.rs
+++ b/src/text/line.rs
@@ -548,6 +548,37 @@ where
     }
 }
 
+/// Adds a `Span` to a `Line`, returning a new `Line` with the `Span` added.
+impl<'a> std::ops::Add<Span<'a>> for Line<'a> {
+    type Output = Self;
+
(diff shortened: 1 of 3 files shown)


kdheepak (approved):
I like it! I am curious what others think though

orhun (approved):
very nice!

PR #1239, opened 2024-07-26
perf(reflow): eliminate most WordWrapper allocations

SUPERCILEX:
On large paragraphs (~1MB), this saves hundreds of thousands of allocations.

TL;DR: reuse as much memory as possible across `next_line` calls. Instead of allocating new buffers each time, allocate the buffers once and clear them before reuse.

### DIFF
--- a/src/widgets/reflow.rs
+++ b/src/widgets/reflow.rs
@@ -1,4 +1,4 @@
-use std::{collections::VecDeque, vec::IntoIter};
+use std::{collections::VecDeque, mem};
 
 use unicode_segmentation::UnicodeSegmentation;
 use unicode_width::UnicodeWidthStr;
@@ -34,11 +34,16 @@ where
     /// The given, unprocessed lines
     input_lines: O,
     max_line_width: u16,
-    wrapped_lines: Option<IntoIter<Vec<StyledGrapheme<'a>>>>,
+    wrapped_lines: VecDeque<Vec<StyledGrapheme<'a>>>,
     current_alignment: Alignment,
     current_line: Vec<StyledGrapheme<'a>>,
     /// Removes the leading whitespace from lines
     trim: bool,
+
+    // These are cached allocations that hold no state across next_line invocations
+    pending_word: Vec<StyledGrapheme<'a>>,
+    pending_whitespace: VecDeque<StyledGrapheme<'a>>,
+    pending_line_pool: Vec<Vec<StyledGrapheme<'a>>>,
 }
 
 impl<'a, O, I> WordWrapper<'a, O, I>
@@ -50,28 +55,29 @@ where
         Self {
             input_lines: lines,
             max_line_width,
-            wrapped_lines: None,
+            wrapped_lines: VecDeque::new(),
             current_alignment: Alignment::Left,
             current_line: vec![],
             trim,
-        }
-    }
 
-    fn next_cached_line(&mut self) -> Option<Vec<StyledGrapheme<'a>>> {
-        self.wrapped_lines.as_mut()?.next()
+            pending_word: Vec::new(),
+            pending_line_pool: Vec::new(),
+            pending_whitespace: VecDeque::new(),
+        }
     }
 
     /// Split an input line (`line_symbols`) into wrapped lines
     /// and cache them to be emitted later
     fn process_input(&mut self, line_symbols: impl IntoIterator<Item = StyledGrapheme<'a>>) {
-        let mut result_lines = vec![];
-        let mut pending_line = vec![];
+        let mut pending_line = self.pending_line_pool.pop().unwrap_or_default();
         let mut line_width = 0;
-        let mut pending_word = vec![];
         let mut word_width = 0;
-        let mut pending_whitespace: VecDeque<StyledGrapheme> = VecDeque::new();
         let mut whitespace_width = 0;
         let mut non_whitespace_previous = false;
+
+        self.pending_word.clear();
+        self.pending_whitespace.clear();
+        pending_line.clear();
 
         for grapheme in line_symbols {
             let is_whitespace = grapheme.is_whitespace();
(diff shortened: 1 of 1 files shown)


src/widgets/reflow.rs
```diff
@@ -189,13 +197,18 @@ where
...
                     .map(|grapheme| grapheme.symbol.width() as u16)
                     .sum();
 
-                self.current_line = line;
+                {
+                    let old = mem::replace(&mut self.current_line, line);
+                    if old.capacity() > 0 {
+                        self.pending_line_pool.push(old);
+                    }
+                }
```
joshka, line 211: Why the block here?

src/widgets/reflow.rs
```diff
@@ -34,11 +34,15 @@ where
...
-    wrapped_lines: Option<IntoIter<Vec<StyledGrapheme<'a>>>>,
     current_alignment: Alignment,
     current_line: Vec<StyledGrapheme<'a>>,
     /// Removes the leading whitespace from lines
     trim: bool,
+
+    wrapped_lines: VecDeque<Vec<StyledGrapheme<'a>>>,
+    pending_word: Vec<StyledGrapheme<'a>>,
+    pending_whitespace: VecDeque<StyledGrapheme<'a>>,
+    pending_line_pool: Vec<Vec<StyledGrapheme<'a>>>,
```
joshka, line 45: Would be helpful to document how each of these is used / and how they fit into the algorithm. Reflow is mostly magic (it has traditionally had some high complexity, which is perhaps lower now than it has been in the past). But I don't pretend to (or even try to) understand it in its current state.

joshka (approved):
Can you add some comments about the algorithm used to make it easier to maintain this. In particular what the change was that helps. Think of this as a way to help future readers understand why this code does what it does. One way to do this is to update the first comment on the PR, which often ends up in the changelog verbatim (and at the very least is the first thing people see when clicking through to find more details on the change).

Also I'd like to see some of the allocation measurements you've done and benchmark comparisons if you can show these.

Pinging @tranzystorekk as they were the last person to touch this and have more context on a review than I do.

General approval for this to go ahead assuming it does what it says on the commit message, but definitely want a second review.

src/widgets/reflow.rs
SUPERCILEX: This is just my coding style. Can remove if you'd like, but I always try to minimally scope variables (in this case `old`).

src/widgets/reflow.rs
SUPERCILEX: `wrapped_lines` usage hasn't changed (the type just went from an iterator to a buffer) and I added a comment to the others indicating they're just caches and don't actually hold any data. I'm mainly pulling vecs that were defined inside `process_input` to the struct so their allocations are reused.

src/widgets/reflow.rs
joshka: It's not a huge deal, but I'd generally interpret the implication of the block surrounding some small amount of code like this is that old requires some sort of special drop behavior. But without a comment to help guide the next person that reads this, the intent of why that is is lost. For something like this perhaps documenting it with a method name would help (and achieve the minimal scope)? E.g. what do you think `self.replace_current_line(line)` reads better for this block?

src/widgets/reflow.rs
joshka: Yeah, my problem is every time I jump into this code it's something that I have to relearn from scratch - off the top of my head I couldn't describe the algorithm in detail except it statefully iterates over the lines and puts them into lines (which is sort of definitional rather than algorithmic so 🤷).

src/widgets/reflow.rs
SUPERCILEX: Sure, `replace_current_line` seems pretty clear. I also renamed the variable `cache` to be clearer.

src/widgets/reflow.rs
SUPERCILEX: Lol, fair enough.

SUPERCILEX:
Benchmarks:
```
paragraph/render_wrap/64
                        time:   [203.71 µs 204.68 µs 205.74 µs]
                        change: [-14.266% -11.921% -9.5791%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 10 outliers among 100 measurements (10.00%)
  3 (3.00%) low mild
  5 (5.00%) high mild
  2 (2.00%) high severe
paragraph/render_wrap_scroll_full/64
                        time:   [202.86 µs 203.71 µs 204.62 µs]
                        change: [-5.6208% -4.4032% -3.2639%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 4 outliers among 100 measurements (4.00%)
  3 (3.00%) high mild
  1 (1.00%) high severe
paragraph/render_wrap/2048
                        time:   [229.06 µs 229.88 µs 230.80 µs]
                        change: [-8.6312% -8.1080% -7.6043%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 4 outliers among 100 measurements (4.00%)
  3 (3.00%) high mild
  1 (1.00%) high severe
paragraph/render_wrap_scroll_full/2048
                        time:   [229.72 µs 230.77 µs 231.92 µs]
                        change: [-7.7813% -7.2111% -6.6529%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 10 outliers among 100 measurements (10.00%)
  7 (7.00%) high mild
  3 (3.00%) high severe
paragraph/render_wrap/65535
                        time:   [1.6269 ms 1.6399 ms 1.6593 ms]
                        change: [+4.9704% +6.1672% +7.6439%] (p = 0.00 < 0.05)
                        Performance has regressed.
  5 (5.00%) high mild
  7 (7.00%) high severe
paragraph/render_wrap_scroll_full/65535
                        time:   [1.4048 ms 1.4114 ms 1.4190 ms]
                        change: [-30.278% -26.925% -23.530%] (p = 0.00 < 0.05)
                        Performance has improved.
  2 (2.00%) high mild
  11 (11.00%) high severe
```

SUPERCILEX:
The semver stuff seems unrelated?

joshka:
> The semver stuff seems unrelated?

Yep - we're experimenting with a new CI tool that tells us when the're breaking changes, but haven't fully tweaked it to understand already breaking changes in main.

joshka (approved):
LGTM - thanks for the changes

PR #1240, opened 2024-07-26
feat: use SmallVec in various APIs

SUPERCILEX:
WARNING: THIS IS A BREAKING CHANGE

I replaced the internal Vec structures of Block, Text, Line, and Layout with SmallVecs. All except for Layout have space for a single stack allocation while Layout has space for 5. One could choose bigger numbers, but I think these address the common use cases.

I expect this change to be somewhat controversial, but I think it's a good idea for both theoretical and data reasons. On the theory side, I expect most ratatui applications will be lists of stuff which means many applications will have a Text with a single Line which has a single Span. In fact, several APIs create singleton Vecs. In the real world, my application (with that 1 MB Paragraph) saves _millions_ of allocations scrolling the paragraph. Obviously I would be better off caching the paragraph for scrolling, but that requires work on the application front whereas this fixes any jank.

### Before
image

### After
image


src/layout/layout.rs
```diff
@@ -109,7 +110,7 @@ thread_local! {
 #[derive(Debug, Default, Clone, Eq, PartialEq, Hash)]
 pub struct Layout {
     direction: Direction,
-    constraints: Vec<Constraint>,
+    constraints: SmallVec<Constraint, 5>,
```
kdheepak, line 113: fwiw, we won't need this change here because `Layout` is cached.

SUPERCILEX: Hmmm, I don't think that's true. Every time you call vertical or horizontal there's an allocation: https://github.com/ratatui-org/ratatui/blob/57dc90736f2daa624ceb6649e90c9cb794cc235d/src/layout/layout.rs#L163

kdheepak: My bad!

SUPERCILEX: No worries! :) Also turns out just accessing the cache clones the layout (which I don't think can be fixed sadly): https://github.com/ratatui-org/ratatui/blob/57dc90736f2daa624ceb6649e90c9cb794cc235d/src/layout/layout.rs#L582

And there's some locking that doesn't make sense since it's a thread-local which I've fixed here: https://github.com/ratatui-org/ratatui/pull/1245

kdheepak: yeah, I just noticed the clone there too.

Thanks for all the PRs!

orhun:
> I replaced the internal Vec structures of Block, Text, Line, and Layout with SmallVecs. 

If you only changed the internal types and if we are not exposing them in any way, isn't this theoretically not a breaking change from the user point of view? Am I missing something?

SUPERCILEX:
> If you only changed the internal types and if we are not exposing them in any way, isn't this theoretically not a breaking change from the user point of view? Am I missing something?

The fields are public (maybe this isn't actually intentional?):
- https://docs.rs/ratatui/latest/ratatui/text/struct.Text.html
- https://docs.rs/ratatui/latest/ratatui/text/struct.Line.html
- Paragraph and Layout's fields are private, so we could actually make that change without breaking backwards compat, though most of the improvements come from the smallvec in Line.

> edit: Also, I'm having trouble following the benchmark results. Can you possibly explain what is the difference between those two screenshots? Less system calls, memory usage?

Ah yeah sorry, I should have cropped the screenshots. Ignore everything except the orange memory usage bar: every spike is me scrolling one line in a ~1MB paragraph. The old version is so slow that you can actually see the memory allocations happen in real-time (hence the triangle). In the new version, all those little rectangles are also me scrolling but instead of the slow growth is just a few big allocations upfront. Also the top-left of the memory usage bar has the peak which went from ~7MB to ~3.6MB.

SUPERCILEX:
Oh hey I just realized there are benchmarks!

```
paragraph/new/64        time:   [5.0594 µs 5.0737 µs 5.0832 µs]
                        change: [-16.674% -16.101% -15.446%] (p = 0.00 < 0.05)
                        Performance has improved.
  3 (3.00%) low severe
  5 (5.00%) high mild
  4 (4.00%) high severe
paragraph/new/2048      time:   [207.44 µs 207.81 µs 208.05 µs]
                        change: [-31.424% -31.067% -30.775%] (p = 0.00 < 0.05)
                        Performance has improved.
  4 (4.00%) low severe
  1 (1.00%) low mild
  4 (4.00%) high mild
  4 (4.00%) high severe
paragraph/new/65535     time:   [6.8650 ms 6.8768 ms 6.8904 ms]
                        change: [-62.408% -61.809% -61.232%] (p = 0.00 < 0.05)
                        Performance has improved.
  6 (6.00%) high severe

```

Cargo.toml
```diff
@@ -36,6 +36,7 @@ lru = "0.12.0"
 paste = "1.0.2"
 palette = { version = "0.7.6", optional = true }
 serde = { version = "1", optional = true, features = ["derive"] }
+smallvec = "2.0.0-alpha.6"
```
joshka, line 39: Can you please comment on why you need the alpha release version of smallvec?

clippy.toml
```diff
@@ -14,4 +14,5 @@ allowed-duplicate-crates = [
     "windows_x86_64_gnu",
     "windows_x86_64_gnullvm",
     "windows_x86_64_msvc",
+    "smallvec", # Remove once smallvec 2.0 is released
```
joshka, line 17: Can you add a comment about what the path to duplicates is like above (it's useful to see from a historical perspective why we're facing this and where there might be type incompatibilities.

src/layout/layout.rs
```diff
@@ -1153,29 +1156,29 @@ mod tests {
         let fixed_size_array = [Constraint::Min(0)];
         let layout = Layout::new(Direction::Horizontal, fixed_size_array);
         assert_eq!(layout.direction, Direction::Horizontal);
-        assert_eq!(layout.constraints, [Constraint::Min(0)]);
+        assert_eq!(&*layout.constraints, [Constraint::Min(0)]);
```
joshka, line 1159: This looks pretty ugly - perhaps rephrase as `smallvec![Constraint::Min(0)]`

joshka (commented):
IMO we should generally avoid changing public field types on structs that are used by every ratatui app like this. As the fix is cumbersome (requires adding a new dependency).

The layout and block changes are more reasonable. It would a good idea to split those off from the more controversial text/line changes, which I'm against right now (but I welcome a contrasting view if there's some logic to support it that takes into account the impact on app code).

Cargo.toml
SUPERCILEX: Yes lol. After about 5 hours of pain debugging lifetime errors that didn't make any sense, I discovered this: https://github.com/servo/rust-smallvec/issues/146. I'll be honest I still have pending tabs open to try and understand *why* the errors occurred, but right now the gist is that nothing compiles with the 1.0 release of smallvec.

src/layout/layout.rs
SUPERCILEX: Not sure I agree, (to my eyes) this is pretty standard deref coercion but I'll switch if people really don't like it.

SUPERCILEX:
Ok I created https://github.com/ratatui-org/ratatui/pull/1246 to get some of these changes in. If that PR goes through and https://github.com/ratatui-org/ratatui/pull/1012 makes progress, that would only leave Line to be improved. Regarding API usability, I'll argue a few points:
- I don't think many people use the fields directly (I didn't even know they existed until working on this PR). If codesearch can be used as an unscientific indicator, about ~10% of ratatui files use `lines` and ~1% for `spans`: [this](https://github.com/search?q=%22.lines%22+language%3ARust+ratatui&type=code) and [this](https://github.com/search?q=%22.spans%22+language%3ARust+ratatui&type=code) vs [this](https://github.com/search?q=language%3ARust+ratatui&type=code).
- Of the people that do use the fields, many won't even notice the change because they're using standard APIs like `len` and `push` that are identical to `Vec`.
- Finally, of the people actually using the fields for something funky, most breakages can be fixed with a simple `.into_vec()` which has no worse performance than before as that method does nothing if the SmallVec was heap allocated. For people actually trying to pass mutable references around, they could do this:
  ```rust
  let mut vec = mem::take(&mut line.spans).into_vec();
  my_api(&mut vec);
  line.spans = vec.into();
  ```
  Not great, but at least it's at the call site instead of requiring pervasive changes.

src/layout/layout.rs
joshka: Ugly mostly because test code should look simple. When you bring in deref coercion just to test equality, that's no longer simple.

SUPERCILEX: Ah, ok I agree with that.

joshka:
Giving this some more thought in the layout case, it's worth considering the maximum number of calls you might see on this per frame? I think this would  perhaps be around 100-200 for a really complex UI . I doubt that anyone is going to be hurting for 100 allocations 60 times a second.

It's the same deal with Block::title, the max number of reasonable blocks that might make sense to draw on a screen of ~200x50 cells is maybe somewhere around ~20x15 = 300, so 18,000 per second.

I've definitely been accused of downplaying perf issues in the past, but I err on the side of clean obvious code where possible (vec rather than smallvec is an example of that). My perspectives on this are very similar to those talked about in https://tobeva.com/articles/beyond/. That's not to say this isn't a good place to make this optimization, but I'd like to see some evidence that it is before making this change.

--

Regarding the public side of this:

The tradeoff of introducing smallvec in the public API is that users that need to interact with these parts of the API needs to understand another library and understand why it matters to their app. It also introduces a potential diamond dependency issue (i.e. apps that already use smallvec 1.x and want to use the public API which would be at 2.x now have to do shenanigans to fix this).

src/layout/layout.rs
SUPERCILEX: Actually SmallVec just needs to fix its partialeq impls: https://github.com/servo/rust-smallvec/pull/358

SUPERCILEX:
> I think this would perhaps be around 100-200 for a really complex UI

image

Some of the APIs scale with data size, not what's rendered on screen:
https://github.com/ratatui-org/ratatui/blob/3725262ca384d28a46e03013023a19677b5a35fe/src/text/text.rs#L193-L196

In my benchmark with the huge string, it has a ton of lines (which don't get displayed) but need to be allocated anyway. That's what I'm trying to avoid. As for clean code, I just don't think SmallVec is particularly complicated. Sure the tests became annoying but that will be fixed by the PR I sent and I'll update ratatui if it's released.

Though it's worth noting I'm on the complete opposite end of the spectrum and will optimize things even if they don't matter because I hate the "it's good enough" philosophy, so I'm not impartial here either. :sweat_smile: I do, however, believe there should be a spectrum from most optimized to least along the lines of `kernel -> stdlib/language/runtime -> core libraries/frameworks -> applications`. ratatui falls in the core library category and its optimization decisions affect *all* downstream users which means the bar should be reasonably high.

> The tradeoff of introducing smallvec in the public API is that users that need to interact with these parts of the API needs to understand another library and understand why it matters to their app. It also introduces a potential diamond dependency issue (i.e. apps that already use smallvec 1.x and want to use the public API which would be at 2.x now have to do shenanigans to fix this).

This I agree sucks, so I'd say maybe don't bother changing the API for `Text` since a) that's what was most common in codesearch so will annoy the most people and b) creating a `Text` probably means you'll have several lines so the savings are dubious. If I could only make a single change, it would be to Line's API—that's who's killing my performance.

SUPERCILEX:
Did a bit more ruminating on the trade offs and had a few more thoughts:
- Taking a step back, the big picture goal is to let users call `Paragraph::new` without tanking performance on large strings. That only necessitates fixing Line not Text, so **I do think not messing with Text is the right approach**. For the internal APIs, I don't think this adds any extra complexity except in the tests which will hopefully get fixed as I mentioned.
- When talking about performance and complexity, I feel like people often forget about downstream complexity and "death by a thousand paper cuts."
  - In this scenario, `Paragraph::new` being too slow leads to downstream complexity. Users would need to either fork ratatui and customize it to their application or find workarounds to pass smaller views of the data into paragraph and update the view as the user scrolls. Of course, messing with the public API also adds complexity and I'm not denying that, rather I want to make it clear that there's that meme bell curve going on (but opposite I think?) when it comes to performance: **too slow and you get complexity trying to work around it; too fast and the API is likely convoluted and unpleasant**.
  - "death by a thousand paper cuts" is pretty simple: **if small hits are accepted too many times they take the whole thing down**. That's where my philosophy comes from for optimizing internal APIs. Though again, the previous point has to be taken into account to not accept too much complexity unless you know an API will never need to change again (which is most certainly not the case in a project as active as this).

SUPERCILEX:
Holy shit the smallvec people are fast! My fix is already released: https://github.com/servo/rust-smallvec/pull/358#issuecomment-2253744287

joshka:
> Holy shit the smallvec people are fast! My fix is already released: https://github.com/servo/rust-smallvec/pull/358#issuecomment-2253744287

This is as small targeted changes should be :)

> Some of the APIs scale with data size, not what's rendered on screen:

The 200 note was about layouts, not text / paragraph. A sufficiently complex UI that would need to allocate that many different layouts per screen might be easier to write as just math.

I hear you on the Text/Line stuff / papercuts / memes etc. and need to give it a bit of thought. In a real world app with bad data, I wonder where this part of the problem sits in the stack of bottlenecks. I'd hazard a guess that there's bigger algorithmic problems that would be there to look at that overshadow this perf result.

The other part of doing perf stuff is that I like to have a really good handle on the problem that it was causing, otherwise perf changes become "why is this code like this", "Because SPUERCILEX from the internet said it was slow and made it faster". There's a carrying cost to performance code. I'm not against it outright, but it's code that has to justify its existence much more than plain feature code. I think it's worth having some reasonable large benchmark tests that show off some of the problems in ways that users might actually hit with bigger application data things.

Or put another way, I sometimes have problems visualizing the exact nature of a performance problem expressed generically, as there's often multiple ambiguous versions of the generic problem, and it's difficult to narrow that down to just a concrete example that doesn't accidentally dissolve into something easily straw-manned.

There's a really good URLO comment from one of the smallvec maintainers that's worth a read at https://users.rust-lang.org/t/when-is-it-morally-correct-to-use-smallvec/46375/5?u=joshka

> First, some reasons not to use SmallVec:
> 
> - When it spills onto the heap, it’s just a worse Vec.
> - It wastes more space when empty.
> - It takes up inline space in structs/containers, which can decrease cache locality of the fields outside > of the vector.
> - Moving a SmallVec by value may be much more expensive than moving a Vec, and LLVM does not always ellide > moves perfectly. This means you occasionaly need to re-architect some code in order to use SmallVec > efficiently.
> - It uses unsafe code. While we do a lot to audit and test it, soundness bugs have slipped through a few > times.
> 
> Reasons that SmallVec might be a big win:
> 
> - You know that your vector will hold a very small number of items most of the time, and only rarely need > to hold more. (SmallVec may speed up the common case a lot, while slightly slowing down the uncommon > case.)
> - Your vector is local to the stack of a function that is called frequently. (SmallVec will usually be > allocated on the stack, saving the need for frequent heap allocations and deallocations.)
> - You want cache locality between the items in your vector and other data stored alongside it. (SmallVec > eliminates a pointer indirection between the vector and its container.)
> 
> I'd probably want at least two of these reasons to be true, or some other strong evidence of a performance > win, before I would reach for SmallVec. (Disclosure: I’m one of the maintainers of the smallvec crate.)

Probably the block title case most easily fits the use it criteria. I'm not sure about the other cases.

There's also a fun quote in the same thread:

>One elegant use of smallvec I've encountered is in a terminal application. It has a grid of character cells in which every cell holds a grapheme. With unicode combiners and modifiers and such, this can become quite large (it's hard to put a strict upper bound). However, in practice, by far most cells will contain 'simple' characters like ASCII or at least single codepoints. Using smallvec avoids (in the expected use case) the overhead of a zillion heap allocations and deallocations, as well as a level of indirection when accessing the contents.

SUPERCILEX:
Thanks for sharing https://tobeva.com/articles/beyond/, it was a great read! I really love the idea that **"clean code looks like it was written by someone that cares,"** as the author then goes on to explain that both the OOP style and perf style are written by people who care, but about different things. I think caring about different aspects of the code aptly mirrors some of this thread. He also mentions my point about core stuff (the game engine) needing to be optimized more than than the game itself (user app).

The one major disagreement I have is that he (implicitly) assigns quotas to performance and claims that as long as you're within the quota, there's no need to optimize. This is the "good enough" philosophy I mentioned that feels wasteful to me. As the author points out, the stack goes deep so you can't optimize everything, but if you could press a magic button that made your code faster, everybody would press the button (I guess it's called `-O3` lol). Thus, **if something *can* be optimized with relatively little compromise, I believe that it *should***. To use his example of a simple game that can easily fit in the 16ms budget, using the OOP version of the code means you're needlessly burning power, limiting CPU time available to other applications, limiting the GPU power budget because the CPU is wasting it, etc. Wasteful code is icky to me, even if it's pretty from another angle.

---

> The 200 note was about layouts, not text / paragraph.

Ah, sorry I misunderstood!

> I'd hazard a guess that there's bigger algorithmic problems that would be there to look at that overshadow this perf result.

Absolutely agree, but then this goes back to the article you linked (the discussion about memory layout). The correct way to write all this stuff from a performance standpoint is to not nest the Vecs and instead require all the APIs to pass in various arenas for the different vecs. Or we contort the API to just have a single span buffer that's passed around. Or... I'm sure we could find many ways to make things more efficient by offering lower-level APIs, but at some point the scale starts to tip too far into performance land and things become unusable.

> "why is this code like this", "Because SPUERCILEX from the internet said it was slow and made it faster"

Haha, fair enough. Weird PRs from external contributors are definitely a problem, for both sides.

On my end, I would need to rewrite all of ratatui to achieve a good balance of performance and usability—if we sat in a room for 3 months, we'd inevitably come up with a beautiful API that offers low-level interfaces with next to no overhead and high level abstractions built on top of those for common use cases. But a) that takes forever, b) I can't personally commit that time, and c) I don't even think such sweeping changes would necessarily be welcomed. So for an external contributor you're basically stuck making surface level changes.

On the maintainer's end, new features and more flexibility are likely at the top of the priority list. Until this library settles down into a stable set of features, it's hard to justify re-writing the existing features/APIs if all you'll get is better performance.

These conflicting priorities mean **maintainers don't want surface level changes that will just be annoying to undo when the API changes, but they also can't prioritize fundamental improvements** as, again, the APIs are likely to change meaning those fundamental changes will have to be thrown away.

I think that leaves **two possible reasons surface level improvements should be done**:
- They are easy to undo.
- They are unlikely to need to be undone.

Getting back to this PR, the non-public changes are easy to undo so I think they make sense. `Text` changes are not easy to undo, so probably shouldn't be done. `Line` changes are slightly easier to undo (less people using the field) though not free, but offer a bigger benefit that hopefully outweighs the breakage, though maybe not.

Only when ratatui's feature set stabilizes do I think it makes sense to look holistically at the entire stack and break all the APIs to achieve a final, optimal version.

---

> it's difficult to narrow that down to just a concrete example that doesn't accidentally dissolve into something easily straw-manned

True, but **a 60% improvement** seems pretty good to me. :grin: 
Re-linking in case this got lost in the noise: https://github.com/ratatui-org/ratatui/pull/1240#issuecomment-2253360659

> There's a really good URLO comment from one of the smallvec maintainers that's worth a read

Nice thread! Though the stack overflow concerns are overblown unless you're using recursion IMO.

Trade-off analysis using the smallvec maintainer quote:
- `Layout` and `Block` are both types stored on the stack (please correct me if I'm wrong, are there use cases where you'd store them in a vec?) so their size increase should be irrelevant. Now the question is do we think most applications will fit within the smallvec sizes. Based on some codesearch of `Layout::{vertical,horizontal}`, I've yet to see a layout with more than 4 items, which means maybe we could even bump the smallvec size down, though I really don't think the stack size matters all that much. A few minutes of poking around codesearch shows that people who call `.title` always (that I've seen) have just one title while people who call `.title_{top,bottom}` sometimes use both. So that suggests we should bump up title storage to 2. I think these results point to **smallvec being a good idea for `Layout` and `Block`!** Almost everybody will end up with stack allocations.
- `Text` is a dead horse at this point, so skipping.
- `Line` sucks because it tends to be stored in lists and `Span` is 40 bytes which is pretty beefy. Smallvec-ing `Line` takes it from 48 bytes to 72. This means we *reaaaaaaally* need that extra space to be worth it. Thankfully, I think we're seeing that most lines will have a single span due to the way they're constructed. Though that leads to the question: **why does `Line` generate multiple spans per newline?** Rendering breaks if you get rid of the conversion in `Text` (which makes sense as a line should be a line), but my app seems to work fine if I remove the newline split into spans that's in `Line`. This would further increase the likelihood that smallvec gets used properly (and avoids a scan of the data!).

EdJoPaTo:
Regarding layout… The current layout code is far beyond "simple" code and I personally avoid it completely. Writing some math to calculate the positions of things seems far simpler to me than using the current layouting stuff. (I don't want to run examples or read through `constrains.rs` to understand what might happen)

So introducing another crate there to improve it performance? Maybe simplify it by removing existing complexity? But as I avoid the layout stuff in the first place I don't care much about what happens there.

---

Regarding block… this seems like a good improvement option. I wonder whether we could use a constant here as the amount of titles is something that's (usually) defined at compile time. (Has title at the top left is something that people normally don't change at runtime. The content yes, the existence? Unlikely?)
If that works it wouldn't require an additional dependency but would be breaking. So smallvec would be easier, is hidden from the end user and isn't breaking.
Would be curious about benchmarks there.

---

Regarding line… where I personally see many instances of it is the table. Table with 5 columns and a few hundreds of rows quickly ends up in the thousands of lines. The problem in my eyes there is the generation of all the cells just to render maybe 30 rows. Changing the architecture there to generate the lines of the actually relevant cells seems to be far more effective in my opinion here. (But that requires knowing the height of the cells in order to correctly scroll)

joshka:
I want to simplify this a bit. Can we (succinctly) list the benefits of this change / the things that it fixes? So far I have:

- when creating a paragraph with 65K lines we save 9ms (60%). This is 100% attributable to the `Text::from(String)` call
  (on my machine, this is 3ms (30%)). I'd guess Ed probably could show much larger numbers from running this on a Raspberry Pi though.

The above scenario is the degenerate worst case (because of internal limits). I'm having difficulty envisioning a scenario where it's a reasonable case to consider and where it would have to be in the per-frame hot path.

I'm +1 with Ed's comments about focusing the time to investigate speedups in areas that reduce the amount of work done rather than in speeding up the micro areas like this (unless there's compelling arguments otherwise). My general bar for spending time on performance issues is that there's an actual measured (and not purely hypothetical / synthetic) performance issue that causes problems in a real world app. I find that otherwise these things tend to distract from more useful work. In an ideal situation where we have infinite time, friction is zero, we work on software in a vacuum ;), ... I'd maybe think different, but pragmatism needs to be held as a higher regard.

joshka:
Marking as draft for now - will come back to likely for 0.30

PR #1241, opened 2024-07-26
perf(widgets): Avoid extra allocations when rendering List, and `Table` content.

airblast-dev:
When rendering `List`, the content is copied which causes the `Text` to be cloned. This causes multiple allocations with the `Vec<Line<'_>>`, `Vec<Span<'_>>` in `Line<'_>`, and the `Cow` in `Span<'_>` if an owned value is being cloned. The same situation also applies to `Table`.

I ran the list benchmark, and on average there's roughly a %1-2 improvement, in timings (while the timings were consistently better I'm not exactly sure by how much since I can only test on my personal system). The change is mainly for memory usage and allocations, but wanted to point out the runtime performance improvement as well.

### DIFF
--- a/src/widgets/list/rendering.rs
+++ b/src/widgets/list/rendering.rs
@@ -108,7 +108,7 @@ impl StatefulWidgetRef for List<'_> {
             } else {
                 row_area
             };
-            item.content.clone().render(item_area, buf);
+            item.content.render_ref(item_area, buf);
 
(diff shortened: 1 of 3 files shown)


airblast-dev:
Closing this PR. Going to create separate ones for `List` and `Table` instead since this will likely look ugly in the changelog.

PR #1245, opened 2024-07-26
fix: remove pointless synchronization in layout cache

SUPERCILEX:
The cache is a thread-local, so doesn't make much sense to require synchronized initialization.

### DIFF
--- a/src/layout/layout.rs
+++ b/src/layout/layout.rs
@@ -1,4 +1,4 @@
-use std::{cell::RefCell, collections::HashMap, iter, num::NonZeroUsize, rc::Rc, sync::OnceLock};
+use std::{cell::RefCell, collections::HashMap, iter, num::NonZeroUsize, rc::Rc};
 
 use cassowary::{
     strength::REQUIRED,
@@ -37,7 +37,7 @@ type Cache = LruCache<(Rect, Layout), (Segments, Spacers)>;
 const FLOAT_PRECISION_MULTIPLIER: f64 = 100.0;
 
 thread_local! {
-    static LAYOUT_CACHE: OnceLock<RefCell<Cache>> = const { OnceLock::new() };
+    static LAYOUT_CACHE: RefCell<Option<Cache>> = const { RefCell::new(None) };
 }
 
 /// A layout is a set of constraints that can be applied to a given area to split it into smaller
@@ -210,21 +210,23 @@ impl Layout {
     /// grows until `cache_size` is reached.
     ///
     /// Returns true if the cell's value was set by this call.
-    /// Returns false if the cell's value was not set by this call, this means that another thread
-    /// has set this value or that the cache size is already initialized.
+    /// Returns false if the cell's value was not set by this call, meaning that the cache was
+    /// already initialized.
     ///
     /// Note that a custom cache size will be set only if this function:
     /// * is called before [`Layout::split()`] otherwise, the cache size is
     ///   [`Self::DEFAULT_CACHE_SIZE`].
     /// * is called for the first time, subsequent calls do not modify the cache size.
     pub fn init_cache(cache_size: usize) -> bool {
-        LAYOUT_CACHE
-            .with(|c| {
-                c.set(RefCell::new(LruCache::new(
-                    NonZeroUsize::new(cache_size).unwrap(),
-                )))
-            })
-            .is_ok()
+        LAYOUT_CACHE.with(|c| {
+            let mut cache = c.borrow_mut();
+            if cache.is_none() {
+                *cache = Some(LruCache::new(NonZeroUsize::new(cache_size).unwrap()));
+                true
+            } else {
+                false
+            }
+        })
     }
 
     /// Set the direction of the layout.
@@ -572,16 +574,14 @@ impl Layout {
     /// ```
     pub fn split_with_spacers(&self, area: Rect) -> (Segments, Spacers) {
         LAYOUT_CACHE.with(|c| {
-            c.get_or_init(|| {
-                RefCell::new(LruCache::new(
-                    NonZeroUsize::new(Self::DEFAULT_CACHE_SIZE).unwrap(),
-                ))
-            })
-            .borrow_mut()
-            .get_or_insert((area, self.clone()), || {
-                self.try_split(area).expect("failed to split")
-            })
-            .clone()
+            c.borrow_mut()
+                .get_or_insert_with(|| {
+                    LruCache::new(NonZeroUsize::new(Self::DEFAULT_CACHE_SIZE).unwrap())
+                })
+                .get_or_insert((area, self.clone()), || {
+                    self.try_split(area).expect("failed to split")
+                })
+                .clone()
         })
     }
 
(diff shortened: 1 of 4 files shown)


src/layout/layout.rs
```diff
@@ -37,7 +37,7 @@ type Cache = LruCache<(Rect, Layout), (Segments, Spacers)>;
 const FLOAT_PRECISION_MULTIPLIER: f64 = 100.0;
 
 thread_local! {
-    static LAYOUT_CACHE: OnceLock<RefCell<Cache>> = const { OnceLock::new() };
+    static LAYOUT_CACHE: RefCell<Option<Cache>> = const { RefCell::new(None) };
 }
```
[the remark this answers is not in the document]
SUPERCILEX: Yeah I had a similar thought but the API seemed to be designed to allow initializing a zero capacity cache for people who don't want one. As long as we don't care about that, I see no reason to not just have a cache always present.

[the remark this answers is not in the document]
SUPERCILEX: Hmmm, though actually this would still need a RefCell. Also `init_cache` is now breaking because the returned bool doesn't make sense.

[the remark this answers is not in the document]
joshka: No-one is using init_cache in the wild (at least on a public GitHub repo), so this is safe to break.

https://github.com/search?q=ratatui+init_cache+AND+%28NOT+is%3Afork%29+AND+%28NOT+org%3Aratatui-org%29&type=code

src/layout/layout.rs
```diff
@@ -208,23 +208,10 @@ impl Layout {
...
     /// that subsequent calls with the same parameters are faster. The cache is a `LruCache`, and
     /// grows until `cache_size` is reached.
-    ///
-    /// Returns true if the cell's value was set by this call.
-    /// Returns false if the cell's value was not set by this call, this means that another thread
-    /// has set this value or that the cache size is already initialized.
-    ///
-    /// Note that a custom cache size will be set only if this function:
-    /// * is called before [`Layout::split()`] otherwise, the cache size is
-    ///   [`Self::DEFAULT_CACHE_SIZE`].
```
joshka, line 218: It's worth keeping a note about the default size in the doc comment

src/layout/layout.rs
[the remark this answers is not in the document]
SUPERCILEX: Lol that works :)

src/layout/layout.rs
```diff
@@ -208,23 +208,10 @@ impl Layout {
...
-            .with(|c| {
-                c.set(RefCell::new(LruCache::new(
-                    NonZeroUsize::new(cache_size).unwrap(),
-                )))
-            })
-            .is_ok()
+    pub fn init_cache(cache_size: usize) {
+        LAYOUT_CACHE.with(|c| {
+            *c.borrow_mut() = LruCache::new(NonZeroUsize::new(cache_size).unwrap());
+        });
```
joshka, line 214: This can be simplified to `LAYOUT_CACHE.set()`, or alternatively with_borrow() and then calling [resize()](https://docs.rs/lru/latest/lru/struct.LruCache.html#method.resize) on the cache. Not sure which would be better. I suppose the resize approach means that we leave anything already cached in place. I'm likely bikeshedding at this point though.

I think also since we're breaking this already, do you think it would be worth moving the NonZeroUsize to the method signature here. This would make constructing a valid value the app's problem and not ours.

src/layout/layout.rs
```diff
@@ -572,16 +559,11 @@ impl Layout {
...
-                    NonZeroUsize::new(Self::DEFAULT_CACHE_SIZE).unwrap(),
-                ))
-            })
-            .borrow_mut()
-            .get_or_insert((area, self.clone()), || {
-                self.try_split(area).expect("failed to split")
-            })
-            .clone()
+            c.borrow_mut()
+                .get_or_insert((area, self.clone()), || {
```
joshka, line 563: I'd extract (area, self.clone() into a variable named key. You end up with more expressive code and even save a line as the block no longer wraps.

src/layout/layout.rs
```diff
@@ -1099,37 +1081,19 @@ mod tests {
     }
 
     #[test]
-    fn custom_cache_size() {
-        assert!(Layout::init_cache(10));
-        assert!(!Layout::init_cache(15));
+    fn cache_size() {
         LAYOUT_CACHE.with(|c| {
-            assert_eq!(c.get().unwrap().borrow().cap().get(), 10);
+            assert_eq!(c.borrow().cap().get(), Layout::DEFAULT_CACHE_SIZE);
```
joshka, line 1086: Same here - the borrow can be moved to  `LAYOUT_CACHE.with_borrow()`

src/layout/layout.rs
```diff
@@ -1099,37 +1081,19 @@ mod tests {
...
-            ])
-            .split(target);
-        assert!(!Layout::init_cache(15));
+        Layout::init_cache(15);
         LAYOUT_CACHE.with(|c| {
-            assert_eq!(
-                c.get().unwrap().borrow().cap().get(),
-                Layout::DEFAULT_CACHE_SIZE
-            );
+            assert_eq!(c.borrow().cap().get(), 15);
```
joshka, line 1096: Probably don't need to test this twice now - this test was to show that it didn't change on the second call. Now that it does, this seems unnecessary.

src/layout/layout.rs
SUPERCILEX: No no, this is great. I had looked for `reserve` and `set_capacity` which aren't a thing so I thought you couldn't resize the cache.

> I think also since we're breaking this already, do you think it would be worth moving the NonZeroUsize to the method signature here. This would make constructing a valid value the app's problem and not ours.

Yup, seems better for users to find out zero ain't valid at compile time rather than crashing later.

src/layout/layout.rs
```diff
@@ -37,7 +37,7 @@ type Cache = LruCache<(Rect, Layout), (Segments, Spacers)>;
 const FLOAT_PRECISION_MULTIPLIER: f64 = 100.0;
 
 thread_local! {
-    static LAYOUT_CACHE: OnceLock<RefCell<Cache>> = const { OnceLock::new() };
+    static LAYOUT_CACHE: RefCell<Cache> = RefCell::new(Cache::new(NonZeroUsize::new(Layout::DEFAULT_CACHE_SIZE).unwrap()));
```
joshka, line 40: This is strange that this didn't get caught by the format lint as too wide. Maybe due to being in a macro?

joshka (approved):

joshka:
I'll let one of the other maintainers weigh in on the breaking change here.

Generally we keep a good log of these in the BREAKING CHANGES doc, which I'm almost inclined to skip on this one, but if you were interested, this could be as simple here as "Layout::init_cache no longer returns bool and takes a NonZeroUsize instead of usize".

src/layout/layout.rs
SUPERCILEX: Yeah, fmt never seems to work right in macros.

SUPERCILEX:
> Generally we keep a good log of these in the BREAKING CHANGES doc, which I'm almost inclined to skip on this one, but if you were interested, this could be as simple here as "Layout::init_cache no longer returns bool and takes a NonZeroUsize instead of usize".

I copied you verbatim. :grin:

PR #1248, opened 2024-07-27
docs(examples): add async example

joshka:
This example demonstrates how to use Ratatui with widgets that fetch
data asynchronously. It uses the `octocrab` crate to fetch a list of
pull requests from the GitHub API. You will need an environment
variable named `GITHUB_TOKEN` with a valid GitHub personal access
token. The token does not need any special permissions.


joshka:
image

(There's a bug that causes the style of the first line to be applied to the whole area, which is fixed in https://github.com/ratatui-org/ratatui/pull/1247)

joshka:
This is intended as an example that doesn't involve events, actions, messages, keybindings, channels, etc. Just as a simple one widget, one background task type example. It would be reasonable to extend this with any of those features, or use it as is. This would be quite useful for a multi-pane approach where each pane has asynchronous work to do, that can be handled autonomously without having to coordinate with the entire app.

It's intended as a starting point for conversation for addressing the question "how do I do async simply?"

examples/async.rs
```diff
@@ -0,0 +1,294 @@
...
+    pulls: PullRequestsWidget,
+}
+
+impl App {
+    const FRAMES_PER_SECOND: f32 = 60.0;
+
+    pub async fn run(mut self, mut terminal: Terminal) -> Result<()> {
+        self.pulls.run();
+
+        let mut interval = interval(Duration::from_secs_f32(1.0 / Self::FRAMES_PER_SECOND));
```
kdheepak, line 82:
```suggestion
        let mut interval = tokio::time::interval(Duration::from_secs_f32(1.0 / Self::FRAMES_PER_SECOND));
```

examples/async.rs
```diff
@@ -0,0 +1,294 @@
...
+    init_octocrab()?;
+    let terminal = terminal::init()?;
+    let app_result = App::default().run(terminal).await;
+    terminal::restore();
+    app_result
+}
+
+fn init_octocrab() -> Result<()> {
+    // dotenvy::dotenv()?; // uncomment if you want to use a .env file to load the token
+    let token = std::env::var("GITHUB_TOKEN")?;
```
kdheepak, line 64:
```suggestion
    // https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens#creating-a-fine-grained-personal-access-token
    let token = std::env::var("GITHUB_TOKEN")?;
```

examples/async.rs
kdheepak: I have a minor preference of inlining this import in examples, it is easier to look for `tokio` related code

examples/async.rs
kdheepak: Thoughts on adding a comment to point people to where to generate a `GITHUB_TOKEN`?

examples/async.rs
```diff
@@ -0,0 +1,294 @@
...
+    fn run(&self) {
+        let this = self.clone(); // clone the widget to pass to the background task
+        tokio::spawn(async move {
+            // this runs once, but you could also run this in a loop, using a channel that accepts
+            // messages to refresh on demand, or with an interval timer to refresh every N seconds
+            this.set_loading_state(LoadingState::Loading);
+            match fetch_pulls().await {
+                Ok(page) => this.on_load(&page),
+                Err(err) => this.on_err(&err),
+            }
```
kdheepak, line 168: As an example of why I like inlining `tokio::time::interval`, in an example like this, I'm looking for where are the futures that are being `.await`ed on, and in which parts of the code are they in. Here I like that this code has `tokio::spawn(...)`. 

If I were to go too far, I'd say I even prefer `fetch_pulls` to be inlined or atleast defined inside this `run` method.

examples/async.rs
joshka: 502508eb6

kdheepak (approved):
LGTM!

examples/async.rs
```diff
@@ -0,0 +1,295 @@
...
+    Error(String),
+}
+
+impl PullRequestsWidget {
+    /// Start fetching the pull requests in the background.
+    ///
+    /// This method spawns a background task that fetches the pull requests from the GitHub API.
+    /// The result of the fetch is then passed to the `on_load` or `on_err` methods.
+    fn run(&self) {
+        let this = self.clone(); // clone the widget to pass to the background task
```
kdheepak, line 160: Neat trick!

examples/async.rs
```diff
@@ -0,0 +1,295 @@
...
+
+/// A widget that displays a list of pull requests.
+///
+/// This is an async widget that fetches the list of pull requests from the GitHub API. It contains
+/// an inner `Arc<RwLock<PullRequests>>` that holds the state of the widget. Cloning the widget
+/// will clone the Arc, so you can pass it around to other threads, and this is used to spawn a
+/// background task to fetch the pull requests.
+#[derive(Debug, Clone, Default)]
+struct PullRequestsWidget {
+    inner: Arc<RwLock<PullRequests>>,
```
kdheepak, line 128: I had always used `Mutex` before, good to see alternate ways using `RwLock`

examples/async.rs
joshka: e658be372

examples/async.rs
```diff
@@ -0,0 +1,293 @@
...
+    Loaded,
+    Error(String),
+}
+
+impl PullRequestsWidget {
+    /// Start fetching the pull requests in the background.
+    ///
+    /// This method spawns a background task that fetches the pull requests from the GitHub API.
+    /// The result of the fetch is then passed to the `on_load` or `on_err` methods.
+    fn run(&self) {
```
kdheepak, line 159: Thoughts on renaming this to `spawn_fetches`? or something with `spawn` in the name?

joshka: I'd prefer to keep this fairly standard / generic to indicate that this is something that would be found in most widgets / apps.

PR #1249, opened 2024-07-27
perf: don't split Line content into multiple spans

SUPERCILEX:
WARNING: I DO NOT KNOW IF THIS CHANGE IS CORRECT.

@joshka split off from the discovery in https://github.com/ratatui-org/ratatui/pull/1240#issuecomment-2254272005. It doesn't seem like the newline split here does anything?


src/text/line.rs
```diff
@@ -208,8 +199,6 @@ impl<'a> Line<'a> {
     ///
     /// # Examples
     ///
-    /// Any newlines in the content are removed.
```
EdJoPaTo, line 211: A Line is something without newlines. (Text has multiple Lines)

The \n could be removed and end up in a single span, but that would result in cloned data and less often borrowed one. That would likely end up with more allocations as before.

Maybe the implementation of Line drawing could ensure to ignore `\n` resulting in the same behaviour?

EdJoPaTo (changes requested):

src/text/line.rs
SUPERCILEX: Yeah, I guess my point is that line drawing already seems to ignore lines. I can check again, but it didn't seem like a line with newlines rendered differently with this change.

SUPERCILEX: Yup, just checked again and sticking a `\n` in a `Line::raw` doesn't do anything with or without this change.

EdJoPaTo: The rendering of `\n` in Spans is prevented by this line: https://github.com/ratatui-org/ratatui/blob/b344f95b7cfc3b90221c84cfef8afcf6944bafc1/src/text/span.rs#L274

It's also a control character so set_stringn wouldn't render this either:
https://github.com/ratatui-org/ratatui/blob/b344f95b7cfc3b90221c84cfef8afcf6944bafc1/src/buffer/buffer.rs#L213

But it looks like `Span::render_ref` and `set_stringn` diverged. (Which also results in #1226 being effectless for Spans, so the bug is still there for Spans). Personally I would just remove the logic in Spans and use set_stringn under the hood there, but it seems to contain some strangely complex code from a bugfix #1165 fixing #1160. That's either missing from set_stringn or not a problem at all there?

Tldr… lost way too much time trying to understand that part of the ratatui code…
But they, that sparked #1257 and #1258.

SUPERCILEX: > Tldr… lost way too much time trying to understand that part of the ratatui code…
But they, that sparked #1257 and #1258.

Oof, but also hell yeah! I'm a bit lost—where does that leave this PR? Sounds like you agree that newlines don't need to be removed anymore?

EdJoPaTo (changes requested):
Looks like there is no render test for Span including a newline… Can you add one to prevent a regression in the future, so this behavior stays the same?

Otherwise, I think this is good to be merged. It doesn't change behavior, it includes less magic code, no big `Vec` are creates anymore and probably the biggest win, doesn't clone a bunch of strings when the Cow was Owned already.

kdheepak:
Sorry for not getting to this. I'm not sure about this though. Because I think it'll make it inconsistent with other ways to construct a `Line`. See https://github.com/ratatui-org/ratatui/issues/1111 as an example of something that was fixed.

EdJoPaTo:
> Sorry for not getting to this. I'm not sure about this though. Because I think it'll make it inconsistent with other ways to construct a `Line`. See #1111 as an example of something that was fixed.

Good find! Didn't remember that.

Guess its choosing the other option from #1111, so basically a revert of #1191.
As #1111 also stated both works the same in the end but going with a single span results in less allocations and the ability to reference something rather than cloning stuff. So the performance impact seems more relevant?

SUPERCILEX:
Yup, I agree with @EdJoPaTo. BTW, since `spans` is a public field, it's completely pointless to try and maintain any invariants on it: users can just manually construct an adversarial list. That means whatever we do is merely for convenience's sake, but in this case there's no benefit to having one span per line so simplicity+ perf should win.

SUPERCILEX:
Added the test.

EdJoPaTo:
Oh, you added it for Span. Guess it should be there for Line too. (didn't specify it well enough what I meant I guess? :innocent: )

SUPERCILEX:
Whoops, moar tests it is! One usability issue I'm seeing is that maybe newlines should render as spaces? I don't think that's a job for this PR, but it could be something to think about in the future.

EdJoPaTo:
> One usability issue I'm seeing is that maybe newlines should render as spaces? I don't think that's a job for this PR that behavior hasn't changed, but it could be something to think about in the future.

Personally, I don't think it's really necessary to replace the \n instead of filtering it out. And then it should probably be replaced by something like `` (see https://stackoverflow.com/questions/18927672/newline-symbol-unicode-character#18931703).
But yeah, different discussion.

EdJoPaTo (approved):
As @kdheepak pointed out, we had #1111 and #1191 in the past so we should wait for an approval of @joshka too.

joshka:
I've been out of town a few days and just getting back to things. I want to look at this more deeply, but I want to caution that I'm starting from a position that suggests keeping the filtering based on previous thinking about this. I may come to agreement that filtering the newlines out doesn't make sense once reading / understanding this a bit more, but I wanted to state my priors upfront.

SUPERCILEX:
I'm missing context on previous thinking, but it seems like the original rationale from https://github.com/ratatui-org/ratatui/issues/1111#issuecomment-2111819573 was also performance? Though the conclusion seems wrong? Or rather I don't understand what is meant by

> Two spans makes it possible to use &strs, one span would mean that an owned string would have to be constructed from the two lines.

joshka:
> Or rather I don't understand what is meant by

The quote is in the context of a world where we're filtering, not in the context of a world where we consider filtering during the rendering of the line / span. If we're filtering out the newlines before they get stored in the Line/Span, then we can either create a new single `String` which does not have the newlines, or two `&str`s which don't. The latter doesn't allocate. ("a", "b") The former does. ("ab").

SUPERCILEX:
Ohhh, ok in the context of splitting that does make sense. But what's the context for splitting lol?

joshka:
> Ohhh, ok in the context of splitting that does make sense. But what's the context for splitting lol?

The rationale for that PR was that it aligned with existing behavior of removing newline characters that was not replicated when we added more methods to create / convert into Lines. I'd suggest doing a bit of a history spelunk to understand the context around this a bit more. I'd have to do the same the explain it, and would then only be summarizing things based on what I see in the history. It would be good to have a summary of this before making this change (ala [Chesterson's Fence](https://en.wikipedia.org/wiki/G._K._Chesterton#Chesterton's_fence)).

SUPERCILEX:
Sounds good, found the original discussion: https://github.com/ratatui-org/ratatui/pull/511.

#### Brief summary

**Folks felt `Span`s containing newlines would complicate rendering** so therefore chose to eliminate them in `Line::raw`. "For consistency [with Text::raw], Line should never have a newline character. This also avoids having every widget that takes a line parameter have to work manually handle it." As for why `Line` splits into different `Span`s, that's just how the goal of getting rid of newlines was implemented, but the fact that you end up with multiple spans is somewhat irrelevant as other alternatives were considered.

#### Rebuttal

The rationale that code working with `Line`s will be simplified cannot be valid: assuming rendering (or in general uses of `Line`) must be correct, **anything rendering/using a `Span` (and thus transitively a `Line`) must handle the case where `Span::content` contains newlines**. As mentioned in a previous comment, this is because invariants cannot be maintained due to the fields being public.

---

Next is the argument that getting rid of newlines makes this consistent with `Text`. This is pretty reasonable, but I also think that **a `Line` always containing the contents that was passed in verbatim is a simpler model** from the user's point of view (not the person having to deal with the newlines). The model becomes:

- A paragraph is something with text I can stick in a block.
- Text is something that usually has multiple lines.
- A line (no matter what its contents are!) is something that gets rendered on a single terminal line and ideally contains one span per styling need. Currently, since `Line` is incorrectly assumed by implementors to not contain newlines, a user manually creating a `Line` struct with a newline could see weird behavior. This new model would fix that.
- A span is used when part of a line needs to be styled differently.

This seems quite nice to me from a user's perspective.

---

Finally, I wanted to see how `Span::content` was being used to get a sense of what could break. It gets used in `set_stringn`, `styled_graphemes`, a conversion to `String`, `Span::width`, and a `Display` impl:
- @EdJoPaTo found that **`set_stringn` and `styled_graphemes` already handle newlines**.
- The **`from Line to String` impl I think can be argued to be correct** in maintaining the exact contents of the line as you're trying to leave the ratatui ecosystem.
- **`Display` is broken** as something called `Line` should never emit newlines IMO. **Easy to fix.**
- `**width` is broken** as it counts `\n` as one unit. To fix this we could **update `Span::width` to subtract the number of newlines found OR render newlines as spaces so they match the `width`**. I do think it's more intuitive to render a newline as a space, but that's probably more of a pain.

SUPERCILEX:
Created https://github.com/ratatui-org/ratatui/pull/1270 with fixes for some of my findings. What should we do about `Span::width`?

joshka:
I think you've raised some compelling arguments about keeping the newlines in the Line / Span rather than filtering them. I like where you're going with this, and I'm not opposed to continuing to explore that idea. I'm not sure what makes sense in terms of rendering them however. As a refactoring side of tihs (i.e. just removing the filter on creation), the rendering should not change. As a secondary step / PR though we can design that.

Something to consider as sort of a background idea to all this is how we can approach changes in this area to reduce the amount of processing during rendering. Noting that we have a terminal, which generally has some reasonable sizes for width and height that are in the 100s of characters, what sort of data structure for storage would make things like scrolling / displaying certain portions of a Text/Line/Span ergonomic and fast? I bet it probably looks nothing like what we currently have.

SUPERCILEX:
Yeah, if nothing else this PR has at least generated good branch-off fixes.

> As a refactoring side of tihs (i.e. just removing the filter on creation), the rendering should not change.

Yup, sounds good.

> how we can approach changes in this area to reduce the amount of processing during rendering

Agreed that this would be really cool. In general the answer is of course always *do less stuff*. As you alluded, if somebody asks us to render a huge list, we should try to avoid doing anything to it until we have "sub-view" of the list that we know will be rendered and then only apply rendering transformations to those pieces that will be rendered since they're probably pretty small in a terminal.

> I bet it probably looks nothing like what we currently have.

Maybe? I feel like the public APIs could remain largely the same, but the internals need to be rewritten with the notion of "incremental rendering." I think most performance problems could be solved like this:
1. Do as little work as possible when public APIs are constructed.
2. Compute a start location for where something should start being rendered. Ideally cache it because as you scroll down it will get more expensive to figure where you should start (in a wrapping paragraph for example).
3. Render using the area rect to determine when no more processing would be able to be rendered.

Currently from my itty bit of poking around it seems like internal data structures are created for rendering on the _entire_ input even if only a small sub-view will be visible.

joshka:
> Maybe? I feel like the public APIs could remain largely the same, but the internals need to be rewritten with the notion of "incremental rendering." I think most performance problems could be solved like this:

Mostly on the internal side - I suspect this might be somewhere where we're doing some caching of things, or we're storing the input string instead of the split string and only split it when actually rendering. That sort of thing. Right now we expose too much of the internals publicly though, so we'd need to rethink that. We might also consider whether creating specialized structs to solve this that are opt-in rather than out to make this work nicely but otherwise leave the existing code alone. I don't know fully what that looks like, just spit balling some ideas.

SUPERCILEX:
Yup yup, I'm all for throwing out ideas. :)

> We might also consider whether creating specialized structs to solve this that are opt-in rather than out to make this work nicely but otherwise leave the existing code alone.

I would caution against this: the simple/high-level API really should be built on top of the advanced/low-level API. Otherwise you have 2x the bugs, perf issues, and maintenance burden.

joshka:
Marking as draft for now - will come back to likely for 0.30

SUPERCILEX:
Sounds good! I might take a while to respond as I'll be starting a new job.

PR #1252, opened 2024-07-31
fix(TestBackend): prevent area mismatch

EdJoPaTo:
`TestBackend::new(400, 400)` resulted in a Rect with width/height 255,255 and a locally stored width/height of 400,400 which is wrong.

`TestBackend::new` could have been fixed, but this prevents this kind of error completely as its stored only at a single location now.

Hint for `serde` feature users: The `TestBackend` representation changes. It should be able to read older data, but data generated after this change can't be read by older versions.

### DIFF
--- a/src/backend/test.rs
+++ b/src/backend/test.rs
@@ -34,9 +34,7 @@ use crate::{
 #[derive(Debug, Clone, Eq, PartialEq, Hash)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub struct TestBackend {
-    width: u16,
     buffer: Buffer,
-    height: u16,
     cursor: bool,
     pos: (u16, u16),
 }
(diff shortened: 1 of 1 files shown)


EdJoPaTo:
Is this a breaking change? TestBackend implements serde derive. Reading from an older version should still work but reading from a newer version will not.

joshka (approved):
I'm fine with generally breaking forwards compatibility without noting it as a breaking change in the conventional commit / BREAKING-CHANGES doc. Just make sure the commit (PR comment) calls it out so it gets called out in the changelog.

Ping @a-kenji for their thoughts on this, as the reason TestBackend supports serde is for the reasons mentioned is for an external testing lib mentioned in https://github.com/ratatui-org/ratatui/pull/389. I'm assuming that it's probably ok though, and if kenji doesn't have time to weigh on this, then we can go ahead and merge this as-is.

PR #1253, opened 2024-07-31
feat(layout): add Size::ZERO and Position::ORIGIN constants

EdJoPaTo:
~~Counterpart of `Rect::as_size`.~~

### DIFF
--- a/src/layout/size.rs
+++ b/src/layout/size.rs
@@ -23,21 +23,6 @@ impl Size {
     pub const fn new(width: u16, height: u16) -> Self {
         Self { width, height }
     }
-
-    /// Creates a [`Rect`] from this size at the origin (0,0).
-    pub const fn at_origin(self) -> Rect {
-        self.with_position(Position::ORIGIN)
-    }
-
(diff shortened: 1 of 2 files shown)


src/layout/size.rs
```diff
@@ -16,10 +16,28 @@ pub struct Size {
...
+
     /// Create a new `Size` struct
     pub const fn new(width: u16, height: u16) -> Self {
         Self { width, height }
     }
+
+    /// Creates a [`Rect`] from this size at the origin (0,0).
+    pub const fn at_origin(self) -> Rect {
+        self.with_position(Position::ORIGIN)
+    }
```
joshka, line 30: +1 for having methods that take a size and produce a Rect, but the naming feels subjectively odd. Would it make sense to create these on Rect instead? E.g. `Rect::at_position(size: Size)` / `Rect::at_origin(size: Size)`?

EdJoPaTo: #1254 uses `impl From<(Position, Size)> for Rect` now. Its less readable but works.

joshka (approved):

PR #1254, opened 2024-07-31
refactor(backend)!: return `Size` from `Backend::size` instead of `Rect`

EdJoPaTo:
The `Backend::size` method returns a `Size` instead of a `Rect`.
There is no need for the position here as it was always 0,0.


joshka (approved):
Typo (buffer -> backend) in PR description?

Also Let's be careful to not mark this as a refactor if it changes the public API.

Approving, but would suggest having another maintainer also weigh in on this one as it's a breaking change, and get the other PRs over the line first / rebased etc.

EdJoPaTo:
Idea: Change the cursor methods to use `Position` at the same time, so people can use more specific types on multiple Backend / Terminal related methods?

joshka (approved):
I'd suggest waiting on a second review before merging this one.

joshka:
> Idea: Change the cursor methods to use `Position` at the same release, so people can use more specific types on multiple Backend / Terminal related methods?

I like the idea - we could do this in a non breaking way though by adding methods and deprecating the existing ones. It would be worth doing a quick blast radius check by github search.

EdJoPaTo:
> before merging this one.

Especially as its breaking. Bugfixes like #1226 can't be released when something breaking is merged, so this should wait until a breaking release is actually planned.

joshka:
> Especially as its breaking. Bugfixes like #1226 can't be released when something breaking is merged, so this should wait until a breaking release is actually planned.

I'm checking with the others about doing a 0.28 release soon for the next release. There's a few "sorta breaking" changes here that we could get across the line easily.

EdJoPaTo:
> > Idea: Change the cursor methods to use `Position` at the same release, so people can use more specific types on multiple Backend / Terminal related methods?
> 
> I like the idea - we could do this in a non breaking way though by adding methods and deprecating the existing ones. It would be worth doing a quick blast radius check by github search.

The diff is fairly simple and a deprecated and new method on a trait is somewhat ugly. Even when the deprecated one has a default implementation using the new one. So I'd prefer a clean cut there.

Checking the blast radius might still be a good idea.

joshka:
@orhun / @kdheepak can you please take a look at this change for a second review

src/terminal/terminal.rs
```diff
@@ -197,22 +201,28 @@ where
     ///
     /// Requested size will be saved so the size can remain consistent when rendering. This leads
```
kdheepak, line 202: Should this docstring be changed to "Requested _area_ will be saved"

joshka:
As a followup related to this, something that has bothered me forever is Frame::size. I'd support adding Frame::area and deprecating Frame::size.

src/terminal/terminal.rs
EdJoPaTo: good catch!
e9215f0422f69c8003d2a6252d190f0f882497e6

orhun (approved):

PR #1256, opened 2024-07-31
refactor: internally use Position struct

EdJoPaTo:


joshka:
This seems like it's internal only, and would be easy to make into something that doesn't need the serde change to be merged first (and it only requires Position::ORIGIN if I'm reading it right, not the methods I had some questions about in the other PR), so I'm generally for merging this.

orhun:
wow bencher, nice!

EdJoPaTo:
I removed the `TestBackend` stuff according to https://github.com/ratatui-org/ratatui/pull/1252#pullrequestreview-2214508826 as it wouldn't be compatible with older versions.

joshka (approved):

orhun (approved):

PR #1258, opened 2024-07-31
test(buffer): ensure emojis are rendered

EdJoPaTo:
They are not rendered as expected but not completely wrong either so that's a win?

### DIFF
--- a/src/buffer/buffer.rs
+++ b/src/buffer/buffer.rs
@@ -999,7 +999,7 @@ mod tests {
         let mut buffer = Buffer::filled(Rect::new(0, 0, 7, 1), Cell::new("x"));
         buffer.set_string(0, 0, input, Style::new());
 
-        let expected = Buffer::with_lines([expected]); // Emoji followed by 2 spaces
+        let expected = Buffer::with_lines([expected]);
         assert_eq!(buffer, expected);
     }
 }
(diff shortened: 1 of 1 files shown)


joshka:
Just to be clear - these are characterization tests right? The purpose is to document existing behavior even if it's wrong(-ish)?

joshka (approved):

orhun (approved):

EdJoPaTo:
Yeah. I suspected my change in #1226 might have broken something because of the way more complex emojis are joined together so I wanted to make sure it still renders them.

src/buffer/buffer.rs
```diff
@@ -963,4 +963,43 @@ mod tests {
...
+                char.escape_unicode().to_string(),
+                char.width(),
+                char.is_control()
+            ))
+            .collect::<Vec<_>>());
+
+        let mut buffer = Buffer::filled(Rect::new(0, 0, 7, 1), Cell::new("x"));
+        buffer.set_string(0, 0, input, Style::new());
+
+        let expected = Buffer::with_lines([expected]); // Emoji followed by 2 spaces
```
EdJoPaTo, line 1002: remove old comment
```suggestion
        let expected = Buffer::with_lines([expected]);
```

Issue #1210, opened 2024-06-28
Exports should only be in a single location


EdJoPaTo:
Currently Terminal is exported both as `ratatui::Terminal` and `ratatui::terminal::Terminal`. (There might be more cases like this, haven't checked yet)

> would be breaking, but the public export should only exist in a single location. Either it's something global or it needs structuring in modules. Allowing for both seems strange. General thought here: reduce the stuff that is publicly accessible. One of the two places should be gone.

_Originally posted by @EdJoPaTo in https://github.com/ratatui-org/ratatui/pull/1180#discussion_r1647468130_

orhun:
What would be the best action for this? I realized we are already hiding the exposed struct via `#[doc(inline)]` which makes it not show up in crate docs. We can probably do the same for other duplicate entries (if any) but I'm not sure if it is a big deal.

EdJoPaTo:
If it's hidden, then it can also be removed. There should be exactly one export, not multiple ones.

orhun:
That would be a breaking change. There are projects which use `ratatui::Terminal` and `ratatui::terminal::Terminal`. Unfortunately it is not consistent.

I propose we make the following types only importable from `ratatui` directly instead of `ratatui::terminal` and document it as a breaking change.

```rust
pub use terminal::{CompletedFrame, Frame, Terminal, TerminalOptions, Viewport};
```

What do you think?

EdJoPaTo:
I'm fine with both on their own, as long as it's a consistent one and not duplicated in two locations. There are arguments for both, so I would be fine with both approaches.

joshka:
It's a pretty common pattern for libraries to expose their main types in several locations. In particular you might see some_crate::error::Error as a good example of this. One thing this does give you is an ability to namespace your type refs easily by importing the module name (as e.g. `terminal::Frame`, `terminal::Terminal`), but perhaps this is fine as `ratatui::Frame` `ratatui::Terminal`?

EdJoPaTo:
> It's a pretty common pattern for libraries to expose their main types in several locations.

Then I don't like their code quality :innocent:

Having exactly one way of doing something improves readability and is more clear. There is no need to question which way to choose from or where the differences might be. And when reading code there is no need to look in different positions for the same thing because people might have used different ways for the same thing.

[...]

PR #1260, opened 2024-08-02
fix(terminal)!: make terminal module private

joshka:
This is a simplification of the public API that is helpful for new users
that are not familiar with how rust re-exports work, and helps avoid
clashes with other modules in the backends that are named terminal.

BREAKING CHANGE: The `terminal` module is now private and can not be
used directly. The types under this module are exported from the root of
the crate.

```diff
- use ratatui::terminal::{CompletedFrame, Frame, Terminal, TerminalOptions, ViewPort};
+ use ratatui::{CompletedFrame, Frame, Terminal, TerminalOptions, ViewPort};
```

Fixes: https://github.com/ratatui-org/ratatui/issues/1210

### DIFF
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -75,7 +75,8 @@ This is a quick summary of the sections below:
 [#1160]: https://github.com/ratatui-org/ratatui/pull/1160
 
 The `terminal` module is now private and can not be used directly. The types under this module are
-exported from the root of the crate.
+exported from the root of the crate. This reduces clashes with other modules in the backends that
+are also named terminal, and confusion about module exports for newer Rust users.
 
 ```diff
 - use ratatui::terminal::{CompletedFrame, Frame, Terminal, TerminalOptions, ViewPort};
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -10,8 +10,9 @@ GitHub with a [breaking change] label.
 
 This is a quick summary of the sections below:
 
-- [v0.28.0](#v0280)
+- [v0.28.0](#v0280) (unreleased)
   - `Layout::init_cache` no longer returns bool and takes a `NonZeroUsize` instead of `usize`
+  - `ratatui::terminal` module is now private
 - [v0.27.0](#v0270)
   - List no clamps the selected index to list
   - Prelude items added / removed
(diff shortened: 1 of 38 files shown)


orhun (approved):

BREAKING-CHANGES.md
```diff
@@ -58,6 +59,29 @@ This is a quick summary of the sections below:
...
+- let is_initialized = Layout::init_cache(100);
++ Layout::init_cache(NonZeroUsize::new(100).unwrap());
+```
+
+### `ratatui::terminal` module is now private ([#1160])
+
+[#1160]: https://github.com/ratatui-org/ratatui/pull/1160
+
+The `terminal` module is now private and can not be used directly. The types under this module are
+exported from the root of the crate.
```
EdJoPaTo, line 78: I would add the reason for the why here which is also given on the PR description. It's easier to adopt change when there is a helpful reason.

examples/inline.rs
```diff
@@ -28,10 +28,9 @@ use ratatui::{
     layout::{Alignment, Constraint, Layout, Rect},
     style::{Color, Modifier, Style},
     symbols,
-    terminal::{Frame, Terminal, Viewport},
     text::{Line, Span},
     widgets::{block, Block, Gauge, LineGauge, List, ListItem, Paragraph, Widget},
-    TerminalOptions,
+    Frame, Terminal, TerminalOptions, Viewport,
```
EdJoPaTo, line 33: I like this change. It's already mixed up in the examples (which should be relatively clean in the first place) so it will be even more cleanup in the real world. :sparkles:

EdJoPaTo (approved):
This will create a lot of churn, but I like it. There is exactly one way to import them and its already cleaning up the examples.

BREAKING-CHANGES.md
EdJoPaTo: > Also, I'm inclined to remove the terminal module as it reduces clashes with crossterm::terminal / termwiz::terminal, which are sometimes useful to reference as their module names.
> 
> from https://github.com/ratatui-org/ratatui/issues/1210#issuecomment-2264636585

examples/inline.rs
joshka: 2e400f892

BREAKING-CHANGES.md
joshka: 2e400f892

PR #492, opened 2023-09-11
chore: use modern modules syntax

joshka:
Move xxx/mod.rs to xxx.rs

Fixes https://github.com/ratatui-org/ratatui/issues/488

Issue #488, opened 2023-09-11
Use new module name convention (`src/xxx.rs` instead of `src/xxx/mod.rs`)


joshka:
## Problem
Navigating to module source files named `xxx/mod.rs` is a regular paper cut annoyance as the files in the `tests`, `examples`, and `benches` directories with the same name generally have higher precedence.

## Solution

```shell
mv src/backend/mod.rs src/backend.rs
mv src/text/mod.rs src/text.rs
mv src/widgets/barchart/mod.rs src/barchart.rs
mv src/widgets/canvas/mod.rs src/canvas.rs
mv src/widgets/mod.rs src/widgets.rs
```

## Alternatives
Perhaps there's a way to ensure that VSCode lists modules first?

## Additional context
<img width="602" alt="image" src="https://github.com/ratatui-org/ratatui/assets/381361/6cdb42b3-1206-4699-becf-b7b1a7b17500">

kdheepak:
I'm not so sure how I feel about this.

1. I do understand the pain point, but seems like it is a VSCode specific optimization. For example,

in neovim with telescope:

<img width="1512" alt="image" src="https://github.com/ratatui-org/ratatui/assets/1813121/38b1d0ef-99b8-475f-b067-f49fe09d2ebd">

in terminal with fzf:

<img width="1512" alt="image" src="https://github.com/ratatui-org/ratatui/assets/1813121/7c37fa6c-c5b0-47ee-9ba5-acd9fe5129f0">

2. I have a feeling that this breaks the convention that rust developers might expect. I personally would like to look at more rust crates to see what people are doing. If there's a programmatic way to get statistics that would be awesome, and would motivate this change more strongly for me.

3. Currently, someone new to rust or new to ratatui will be able to see that all "widgets" related code is in "./src/widgets", and similar for the other modules. In my opinion the new version is less intuitive.

4. VSCode uses fuzzy find. It can actually be faster to get to the `mod.rs` files the way it is right now:

<img width="620" alt="image" src="https://github.com/ratatui-org/ratatui/assets/1813121/27d7fe27-3b14-4de2-8225-15334617aeb4">

joshka:
I know that there are ways to avoid this with fuzzy find, but the issue is that there are two types of modules - those with submodules and those without. In the `mod.rs` world, I regularly have to remember which, and then actively choose a different action because of the difference. But that difference has no meaningful value to me.

[...]

PR #1261, opened 2024-08-02
refactor: migrate to mod.rs files

EdJoPaTo:
I prefer the file layout with `mod.rs` files as they indicate on file system level that there are more files in this module. It's either a file or a folder then. Not file and maybe folder. When I don't care about let's say widgets, I can close the widgets folder and everything widget related is hidden.

Some people preach "mod.rs only contains `pub use` and `(pub) mod`, nothing else" which might be a bit extreme but a good general idea where things might be going for more approachable code files. Some files, especially widgets are really huge (ignoring tests) and contain multiple things at once. Maybe "only one pub struct per file" might be a better rule for ratatui?

I mainly want to spark a discussion here and included the PR as a reference about the idea as its quickly created anyway. (Probably needs a rebase when it actually should be merged)
The lint change reflects this preventing the other variant of being used.
It probably should have most maintainers approvals to be merged?


joshka:
Ratatui started with the `mod.rs` approach when we brought it over from `tui-rs`, but we moved away from it in #492 with the reasons documented in #488. I'm a -1 to this PR for the reasons previously documented in the issue. This is a subjective choice (which has been subjectively and objectively rejected by the rust language designers), and one which we've actively moved away from already.

That said, there do feel like a few places where we've pulled things into submodules where a mod.rs almost makes more sense. E.g. `src/text.rs` which only re-exports code from sub modules. For those I'm still against, but not as strongly (perhaps just -0.8). I mainly oppose the change because it adds no value and makes for two ways with a complex rule about which to use.

EdJoPaTo:
That was back when there were even more things meshed into fewer files. Personally I try to do as much as possible in its own files and try to have as little stuff in mod files (with varying success on my own attention). So in the ideal world I wouldn't navigate to a mod file often. (Only when the Repo is quite tiny, but then fuzzy finding is no issue in the first place)

Personally I prefer structure to fuzzy finding (or prelude :ghost:), as I tend to know what im looking for, not how exactly its named. (I search for a layout struct that can do… ah, Margin.) So I seek for a given folder only to find the stuff I seek for outside of given folder. Which would also be kinda gone when more stuff is in their own files rather than a module main file. But that is probably a good idea independent of mod.rs or outside the folder.

Fuzzy find for symbols is also there (Ctrl + T in `VSCode`), but that's probably less useful in an idiomatic code base with [`clippy::module_name_repetitions`](https://rust-lang.github.io/rust-clippy/master/index.html#/module_name_repetitions) enabled. (`ratatui` isn't there yet…)

Thinking about
- `widgets.rs`… The widget trait should be either `widget.rs` or `widgets/widget.rs`.
- `block.rs`… This one seems like a mesh up all together which is in the process of being simplified by getting titles slowly out of it. Especially fun as `borders.rs` is pretty block specific but not part of `block/`
- `table.rs`… which one?
- `layout.rs` is pretty much what I like. There is no real reason to open that file most of the time.
- `backend.rs`… there are tests and structs in it that are not directly related to the trait.

kdheepak:
Thanks for bringing up the discussion! It's good to revisit priors from time to time!

I'll first say I don't feel strongly about either option, as long as we are consistent in the repo and across code in the organization. My preference for not mixing styles is higher than choosing either one of those styles.

I previously used `foo/mod.rs` and at the time of the PRs @joshka linked I was not a fan of moving to `foo.rs`, but reading the PRs, RFC and motivation for this change in rust convinced me that `foo.rs` was worth it. After using it for a while, I'm now either indifferent or leaning toward the status quo.

Here's some of my pros and cons of using the `foo/mod.rs` approach:

**Pros**

1. Fewer top level tree structure because `foo.rs` gets moved to `foo/mod.rs`:

With `mod.rs`:

Screenshot 2024-08-02 at 09 56 33

Without `mod.rs`:

Screenshot 2024-08-02 at 09 56 06

2. Autocomplete in the terminal works ever so slightly more efficiently:

With `mod.rs`:

https://github.com/user-attachments/assets/3b418976-71cd-4e83-b797-a9c68176e9b4

Without `mod.rs`:

https://github.com/user-attachments/assets/4852d737-385d-4a34-a6ef-8489850533a3

**Cons**

In neovim, out of the box with a tabline plugin, this is what my buffers would look like if I opened the `mod.rs` files:

image

If I configure them, this is what they would look like:

image

However, I acknowledge that this might be a neovim / vim plugin specific issue. In vscode, I think this is what they look like out of the box (but I'm not sure if I have it configured somehow and forgot about the configuration), which isn't as bad:

image

---

For me personally the buffer names issue is ergonomically a lot worse. It is annoying enough for me that I personally don't use the `foo/mod.rs` approach unless I have to, even though I like the folder structure better and agree that it is easier to understand the folder structure that way. If I do use the `foo/mod.rs` approach, I'd like there to be as little content inside the `mod.rs` files.

joshka:
This is the main guiding principle I'd fall back on. Moving away from `mod.rs` is a blessed path, and one that's worth becoming familiar with rather than "damn kids get off my lawn"ing it. :)

> Note: Prior to rustc 1.30, using mod.rs files was the way to load a module with nested children. It is encouraged to use the new naming convention as it is more consistent, and avoids having many files named mod.rs within a project.
> -- https://doc.rust-lang.org/reference/items/modules.html

I agree that pro 1 is nice to have. I never have encountered pro 2.

Something that might be a meet in the middle compromise is to use path attributes:
https://github.com/ratatui-org/ratatui/compare/main...joshka:ratatui:jm/move-modules?expand=1

image

I kinda hate it, but less so than switching back to mod.rs.

kdheepak:
Thanks for sharing about this path attributes feature, and thanks to both of you for making commits / PRs so it is easy to see what this would look like! 

I didn't know about this path attribute feature, so I might have to use it in practice to see what it feels like. If I had to guess most new contributors wouldn't know about this? I have a small preference for keeping filesystem as close as possible to representing the module paths, i.e. `ratatui::widgets` -> `ratatui/src/widgets.rs`.

EdJoPaTo:
> I never have encountered pro 2.

I regularly do with `ratatui`… probably one of the reasons which sparked me to do this PR.

> Note: Prior to rustc 1.30, using mod.rs files was the way to load a module with nested children. It is encouraged to use the new naming convention as it is more consistent, and avoids having many files named mod.rs within a project.
> -- https://doc.rust-lang.org/reference/items/modules.html

Thats interesting!

More consistent I'm not sure, but the amout of `mod.rs` files doesn't seem like a problem to me, it actually feels more organized. A folder contains all. But I think I prefer trying to honor a Rust guideline when it works well.

> Something that might be a meet in the middle compromise is to use path attributes

Ugh, seems like more annoyance than helpful… It requires manual configuration of something that should be handled automatically.

---

I think what's something I see from this discussion currently: With either way it's a good idea to split up files and get stuff out of the module root files (whatever their name is).
Ideally only one `pub struct` / `pub enum` per file?

VS Code `explorer.sortOrder` is also helpful. The default of folders then files rips apart the module root file and its folder. It's not perfect, but it's a bit easier to handle with files then folders to get into them. But that should get less relevant when more stuff is in their own files rather than in module root files.

joshka:
> Ugh, seems like more annoyance than helpful… It requires manual configuration of something that should be handled automatically.

Agreed, I did say I hated it ;)

EdJoPaTo:
Closing this being a manageable amount of grumpy sounds good.

kdheepak:
The best part of our governance model is that everyone is always unhappy :)

PR #1263, opened 2024-08-02
docs(terminal): fix imports

EdJoPaTo:
Looks like the failing tests on #1260 where not noticed?


joshka (approved):
Thanks - yup. I missed them.

lint failure seems unrelated.

orhun (approved):

PR #1264, opened 2024-08-02
ci: split up lint job

EdJoPaTo:
This helps with identifying what failed right from the title. Also steps after a failing one are now always executed.

Also shortens the steps a bit by removing obvious names.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,49 +25,45 @@ env:
 # formatting, typos, and missing tests as early as possible. This allows us to fix these and
 # resubmit the PR without having to wait for the comprehensive matrix of tests to complete.
 jobs:
-  lint:
+  rustfmt:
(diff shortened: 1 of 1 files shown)


orhun:
If we are going to step names then we should do it in all places imo.

EdJoPaTo:
> If we are going to step names then we should do it in all places imo, so other workflows as well.
> 
> They do not hurt :)

Some are helpful as easier to read in my opinion, but `Run cargo make clippy-all` for `run: cargo make clippy` is both outdated and exactly the same on the actual run ;)

Can also remove them on other files when they feel duplicated / not at helpful. Mainly didn't want to scroll as much :innocent:

joshka (approved):

joshka:
We can remove the other labels in another PR - I'd like to see how this looks without them, so merging as-is.

PR #1266, opened 2024-08-02
docs(readme): remove MSRV

EdJoPaTo:
This notice was useful when the `Cargo.toml` had no standardized field for this. Now it's easier to look it up in the `Cargo.toml` and it's also a single point of truth. Updating the README was overlooked for quite some time so it's better to just omit it rather than having something wrong that will be forgotten again in the future.


orhun (approved):
LGTM, maybe it makes sense to move this to `lib.rs` docs?

More specifically: https://github.com/ratatui-org/ratatui/blob/a34ff4cae7b02d25f76324ec1684f7414eeb115a/src/lib.rs#L21

EdJoPaTo:
> LGTM, maybe it makes sense to move this to `lib.rs` docs?
> 
> More specifically:
> 
> https://github.com/ratatui-org/ratatui/blob/a34ff4cae7b02d25f76324ec1684f7414eeb115a/src/lib.rs#L21

Will we forget to update it there too? :ghost: 

My point of removing instead of updating it was us being forgetful ;)

orhun:
Good point. I think having it in `lib.rs` is useful, we can update release steps/`RELEASE.md` for not forgetting it. Will look into updating that document very soon anyways.

EdJoPaTo:
As long as stuff like that is automated… But I think checking the `Cargo.toml` is the best place for that as it's a standard and not "somewhere". Especially as the upcoming rust edition 2024 will use that even more.

a-kenji:
I would recommend removing it from duplicate places as well.
`cargo` gives a very clear error message if the MSRV is violated and tools like `crates.io` surface it in an easily digestible manner.

image

README.md
```diff
@@ -365,10 +365,6 @@ Pull Request].
 Please make sure you read the updated [contributing](./CONTRIBUTING.md) guidelines, especially if
 you are interested in working on a PR or issue opened in the previous repository.
 
-## Rust version requirements
-
-Since version 0.23.0, The Minimum Supported Rust Version (MSRV) of `ratatui` is 1.67.0.
-
```
joshka, line 371: I'd suggest replacing this with a policy. E.g. something like:

> Ratatui has an MSRV policy that we will support Rust N-2, but we only generally update the required MSRV when it's necessary to support a feature.

But I'm ok with removing this too.

joshka (approved):

README.md
orhun: +1 for this, makes sense.

EdJoPaTo: Personal opinion: MSRV is a lost game currently. Dependencies randomly update without breaking releases so its kind of impossible to actually use. Personally I would like to allow for people to build my stuff on Debian (Rust 1.63) but it's nearly impossible with the current ecosystem. Rust Edition 2024 (`-Zmsrv-policy`) will allow for MSRV to make sense again. And then the MSRV should be kept at the minimum possible and updates should be checked if they really need the MSRV bump.

So personal opinion: Either target the Rust version of some distro like Debian or argue about every bump without having some policy like n-2 as the current ecosystem mostly assumes stable anyway. Is feature added really worth the bump?
I prefer the first but its kind of impossible without Edition 2024 / `-Zmsrv-policy`.

Tldr: Currently I wouldn't introduce some policy as it's a lost game impossible to hold without lockfiles or Edition 2024.

EdJoPaTo: Also: except for the Cargo.toml lints flag the MSRV was bumped because of dependencies, not because of something ratatui did. Thats out of control for ratatui and without lockfiles impossible to hold.

We should introduce `cargo minimal-versions` instead and allow the users to use `-Zmsrv-policy` effectively.

joshka: The time crate's MSRV policy is N-2, hence transitively that's also effectively our policy.

joshka: Put another way, an MSRV policy doesn't have to be a grand statement thing holds 100% of the time, that is checked and automated. A policy is just words and a commitment to do a thing in a certain way when possible.

Saying N-2 here says to a consumer that Ratatui may or may not work with N-3, but we do guarantee that we won't introduce things which are only available in N-1, and N (unless there's a strong compelling reason to do so like a security issue or incompatibility with our dependencies). This communicates that developers will need to be ok with potentially updating their compiler at least every 3 months, and we will avoid merging changes which don't adhere to that policy and not avoid breaking earlier compiler versions.

An example of this would be the LazyCell/LazyLock stuff stabilized in 1.80.0. In the Layout we had some lazy initialization that had we not removed it recently would be a good candidate for changing over to this. We wouldn't however do so until 1.82.0 was released.

> So personal opinion: Either target the Rust version of some distro like Debian or argue about every bump without having some policy like n-2 as the current ecosystem mostly assumes stable anyway. Is feature added really worth the bump?

I'm not willing to hold back innovation over Debian's choice of compiler version (1.63 IIRC), or any developer's refusal to update to a 3 month old release of Rust. A policy of N-2 mitigates the need to consider trade-offs of whether a feature is worth the bump, as we've already made that determination up front that it is if it's available in N-2.

Introducing minimal versions is an orthogonal concern to this - we should do that in addition to this.

EdJoPaTo: Just to make sure I understand you correctly. You want to deny tools using ratatui from being added to the repos of one of the biggest distros _on purpose_?

Personally I would say, the Rust ecosystem is lost in regard to MSRV until `-Zmsrv-policy` is stable. Then an actually useful policy can be established. Until then, a policy can't do much anyway…

Also, the MSRV of ratatui does not need to be the one the dependencies. It just needs to make sure the minimal-version of dependencies work with its MSRV.

joshka: > Just to make sure I understand you correctly. You want to deny tools using ratatui from being added to the repos of one of the biggest distros on purpose?

I'm saying that I won't personally spend any time on making it possible to run Ratatui software against versions of the rust compiler that are outdated by 2 years (1.63.0 was released August 11 2022), and that anyone that wants to run software that is built with a tool developed today should be willing to use tools that are released recently to compile and run this software. Given the choice of restricting ongoing development of a current project to outdated tooling vs putting a barrier to use the library by a section of the population that does not upgrade, I'm 100% fine with taking the latter approach.

Our dependencies and the features that we use dictate our MSRV
- definitionally: https://github.com/rust-lang/rfcs/blob/master/text/2495-min-rust-version.md
- practically:
```
❯ cargo +1.63.0 check --all-features --all-targets
...
error: package `color-spantrace v0.2.1` cannot be built because it requires rustc 1.65.0 or newer, while the currently active rustc version is 1.63.0
❯ cargo +1.65.0 check --all-features --all-targets
...
error: package `rstest_macros v0.22.0` cannot be built because it requires rustc 1.67.1 or newer, while the currently active rustc version is 1.65.0
❯ cargo +1.67.1 check --all-features --all-targets
...
error: package `clap v4.5.13` cannot be built because it requires rustc 1.74 or newer, while the currently active rustc version is 1.67.1
```

Our MSRV is at 1.74 because we use lint settings in Cargo.toml, and because of a dev dependency on clap (there are other deps as well this just reports the first breaking dependency, which is mostly random - e.g. mio 1.0.1 required by crossterm requires rust 1.70).

joshka:
Merging as is and will add a new PR to add N-2 policy notes

PR #1270, opened 2024-08-03
fix: various newline handling improvements

SUPERCILEX:
Extracted from https://github.com/ratatui-org/ratatui/pull/1249

### DIFF
--- a/src/text/span.rs
+++ b/src/text/span.rs
@@ -565,6 +565,8 @@ mod tests {
         assert_eq!(Span::raw("").width(), 0);
         assert_eq!(Span::raw("test").width(), 4);
         assert_eq!(Span::raw("test content").width(), 12);
+        // Needs reconsideration: https://github.com/ratatui-org/ratatui/issues/1271
+        assert_eq!(Span::raw("test\ncontent").width(), 12);
     }
 
     #[test]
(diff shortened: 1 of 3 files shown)


joshka (approved):

joshka:
I like when we add characterization tests like this :) Thanks!

src/text/span.rs
```diff
@@ -562,6 +565,7 @@ mod tests {
         assert_eq!(Span::raw("").width(), 0);
         assert_eq!(Span::raw("test").width(), 4);
         assert_eq!(Span::raw("test content").width(), 12);
+        assert_eq!(Span::raw("test\ncontent").width(), 12);
```
joshka, line 568: Regarding this - we likely need to filter control characters in the width method to avoid this. It would be ideal if the width method returns the display width and not the unicode width. The crate recently changed to make newlines take up a single character per the unicode spec, but that's not really useful for rendering to the terminal.

SUPERCILEX: Agree, filed https://github.com/ratatui-org/ratatui/issues/1271 to unblock this PR.

joshka: It's worth putting a comment on this line to note it as possibly wrong. and link to #1271

joshka (approved):

SUPERCILEX:
Done.

joshka:
Thanks again!

Issue #1272, opened 2024-08-03
Support for Lines as Chart Labels


nullstalgia:
Continuing from a comment in #1227

Howdy howdy.

I have [an app](https://github.com/nullstalgia/null_iron_heart) I just released using Ratatui, and it was fun learning Rust and this crate! Thanks for maintaining such a cool tool!

And in that app, I have a chart that shows two data sets, correlated, but at different scales. I'm able to do the math to scale them so it appears properly, but I still need to label the bounds of each dataset. I would like to color-code the portions of the labels that correspond to which dataset to differentiate them, but instead I have to rely on parentheses.

Since styling is limited to per-Span, and I can't have a Line as a Label, I'm at an em passe.

![354025593-2675c211-4dda-420b-891e-17ffdc4d9a2b](https://github.com/user-attachments/assets/740dc74f-190c-4c17-ae08-b17d4b857dc7)

Shown is a screenshot of the app in question, and where I would like to apply the stylings.

joshka:
Give #1273 a whirl to check if it does what you need.

nullstalgia:
Wow! I didn't expect a patch so soon!

![image](https://github.com/user-attachments/assets/073309af-ee52-43e9-8e72-607c22f9140e)

Seems to work well! (Although I'll need to think of a smarter way to define them, since the Line and Span constructors are pretty "bulky.")

(My before and after, and after is without the if :P)
```rust
    // let labels = if render_bpm && render_rr {
    //     vec![
    //         format!("{} ({:.1})", bpm_bounds[0], rr_bounds[0]).bold(),
    //         format!("{} ({:.1})", avg_bpm, avg_rr).bold(),
    //         format!("{} ({:.1})", bpm_bounds[1], rr_bounds[1]).bold(),
    //     ]
    // } else if render_bpm {
    //     vec![
    //         format!("{}", bpm_bounds[0]).bold(),
    //         format!("{}", avg_bpm).bold(),
    //         format!("{}", bpm_bounds[1]).bold(),
    //     ]
    // } else {
    //     vec![
    //         format!("{:.1}", rr_bounds[0]).bold(),
    //         format!("{:.1}", avg_rr).bold(),
    //         format!("{:.1}", rr_bounds[1]).bold(),
    //     ]
    // };

    let labels = vec![
        Line::from(vec![
            Span::styled(
                bpm_bounds[0].to_string(),
                Style::default().fg(Color::Red).add_modifier(Modifier::BOLD),
[...]

PR #1273, opened 2024-08-03
feat(barchart)!: allow axes to accept Lines

joshka:
Fixes: https://github.com/ratatui-org/ratatui/issues/1272

### DIFF
--- a/src/widgets/chart.rs
+++ b/src/widgets/chart.rs
@@ -1,7 +1,6 @@
 use std::cmp::max;
 
 use strum::{Display, EnumString};
-use unicode_width::UnicodeWidthStr;
 
 use crate::{
     layout::Flex,
@@ -40,7 +39,7 @@ pub struct Axis<'a> {
     /// Bounds for the axis (all data points outside these limits will not be represented)
     bounds: [f64; 2],
     /// A list of labels to put to the left or below the axis
-    labels: Option<Vec<Span<'a>>>,
+    labels: Option<Vec<Line<'a>>>,
     /// The style used to draw the axis itself
     style: Style,
     /// The alignment of the labels of the Axis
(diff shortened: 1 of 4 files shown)


src/widgets/chart.rs
```diff
@@ -904,17 +908,13 @@ impl<'a> Chart<'a> {
...
-
-        let x = match alignment {
-            Alignment::Left => label_area.left(),
-            Alignment::Center => label_area.left() + label_area.width / 2 - bounded_label_width / 2,
-            Alignment::Right => label_area.right() - bounded_label_width,
+    fn render_label(buf: &mut Buffer, label: &Line, label_area: Rect, alignment: Alignment) {
+        let label = match alignment {
+            Alignment::Left => label.clone().left_aligned(),
+            Alignment::Center => label.clone().centered(),
+            Alignment::Right => label.clone().right_aligned(),
```
joshka, line 915: There are usually a fairly small number of short labels here, so the clone is probably ok here.

tests/widgets_chart.rs
```diff
@@ -203,7 +203,7 @@ fn widgets_chart_handles_long_labels<'line, Lines>(
         "          ",
         "          ",
         "──────────",
-        "AAA  B   C",
+        "AAA B    C",
```
joshka, line 206: This is better aligned with how line treats centering (ambiguous centering preferences to the left)

joshka:
Dependencies action is failing due to docker build.

orhun (approved):

src/widgets/chart.rs
EdJoPaTo: Why is it needed in the first place? :thinking: do the alignment methods require ownership?

src/widgets/chart.rs
```diff
@@ -97,8 +101,8 @@ impl<'a> Axis<'a> {
     ///         .labels(vec!["0".bold(), "25".into(), "50".bold()]);
     /// ```
     #[must_use = "method moves the value of self and returns the modified value"]
-    pub fn labels(mut self, labels: Vec<Span<'a>>) -> Self {
-        self.labels = Some(labels);
+    pub fn labels<T: Into<Line<'a>>>(mut self, labels: Vec<T>) -> Self {
+        self.labels = Some(labels.into_iter().map(Into::into).collect());
```
EdJoPaTo, line 105: Why not `IntoIterator` then too? That would allow for `.labels(["0.0", "5.0", "10.0"])` without `vec![]`.

joshka: Sounds good - want to slap together a quick PR for that?

src/widgets/chart.rs
EdJoPaTo: Ah, it moves the original… well…

PR #1280, opened 2024-08-04
docs: update main lib.rs / README examples

joshka:


EdJoPaTo (approved):
Can we get rid of this duplication somehow? `#[doc(include = "../README.md")]` or something like that?

joshka:
> Can we get rid of this duplication somehow? `#[doc(include = "../README.md")]` or something like that?

No. The README is generated using cargo-rdme, which takes care of updating code links to point to correct web-urls etc.
This also means that this code is formatted by rustfmt.

PR #1281, opened 2024-08-04
docs: document crossterm breaking change

joshka:


BREAKING-CHANGES.md
```diff
@@ -11,6 +11,7 @@ GitHub with a [breaking change] label.
 This is a quick summary of the sections below:
 
 - [v0.28.0](#v0280) (unreleased)
```
EdJoPaTo, line 13: We should get rid of this human generated ToC… enforced automation or the one from GitHub / the markdown reader…

EdJoPaTo (approved):

orhun (approved):

BREAKING-CHANGES.md
joshka: This is about making it easy to see at a glance the problems that each release would cause when upgrading. I don't might automating this, but do mind removing it (even with full knowledge that the GitHub ToC button exists).

PR #1282, opened 2024-08-04
feat(chart): allow removing all the axis labels

EdJoPaTo:
`axis.labels(vec![])` removes all the labels correctly.

Probably never used but prevents weird cases by removing the mix-up of `Option::None` and `Vec::is_empty`.

---

Thought about naming this refactor or feat as its probably something not even used by someone. It's mainly preventing weird edge cases by removing the type mix-up.

### DIFF
--- a/src/widgets/chart.rs
+++ b/src/widgets/chart.rs
@@ -1,4 +1,4 @@
-use std::cmp::max;
+use std::{cmp::max, ops::Not};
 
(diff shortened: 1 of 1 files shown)


joshka (approved):
Nice catch

joshka:
I went with fix as it really fixes up an edge case. Refactor would have been ok too. No big deal though I think.

PR #1284, opened 2024-08-04
feat!: add `get/set_cursor_position()` methods to Terminal and Backend

EdJoPaTo:
Idea from https://github.com/ratatui-org/ratatui/pull/1254#issuecomment-2264919603

Usage search: https://github.com/search?q=ratatui+.set_cursor+AND+(NOT+org%3Aratatui-org)+AND+(NOT+is%3Afork)+AND+(NOT+is%3Aarchived)&type=code

### DIFF
--- a/src/backend.rs
+++ b/src/backend.rs
@@ -104,7 +104,10 @@ use std::io;
 
 use strum::{Display, EnumString};
 
-use crate::{buffer::Cell, layout::Size, prelude::Rect};
+use crate::{
+    buffer::Cell,
+    layout::{Position, Rect, Size},
+};
 
 #[cfg(feature = "termion")]
 mod termion;
@@ -210,7 +213,7 @@ pub trait Backend {
     /// See [`set_cursor`] for an example.
     ///
     /// [`set_cursor`]: Backend::set_cursor
-    fn get_cursor(&mut self) -> io::Result<(u16, u16)>;
+    fn get_cursor(&mut self) -> io::Result<Position>;
 
     /// Set the cursor position on the terminal screen to the given x and y coordinates.
     ///
@@ -220,14 +223,15 @@ pub trait Backend {
     ///
     /// ```rust
     /// # use ratatui::backend::{Backend, TestBackend};
+    /// # use ratatui::layout::Position;
     /// # let mut backend = TestBackend::new(80, 25);
-    /// backend.set_cursor(10, 20)?;
-    /// assert_eq!(backend.get_cursor()?, (10, 20));
+    /// backend.set_cursor(Position { x: 10, y: 20 })?;
+    /// assert_eq!(backend.get_cursor()?, Position { x: 10, y: 20 });
     /// # std::io::Result::Ok(())
     /// ```
     ///
     /// [`get_cursor`]: Backend::get_cursor
-    fn set_cursor(&mut self, x: u16, y: u16) -> io::Result<()>;
+    fn set_cursor<P: Into<Position>>(&mut self, position: P) -> io::Result<()>;
 
     /// Clears the whole terminal screen
     ///
--- a/src/backend.rs
+++ b/src/backend.rs
@@ -195,25 +195,25 @@ pub trait Backend {
     /// # std::io::Result::Ok(())
     /// ```
     ///
-    /// [`show_cursor`]: Backend::show_cursor
+    /// [`show_cursor`]: Self::show_cursor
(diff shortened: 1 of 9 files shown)


joshka:
Given the blast radius (good idea filtering out archives btw), I have a preference to make this change as an additive change with deprecation of the original method. Do you think it would be the worst thing in the world to name these methods like:

- `fn get_cursor_position()`
- `fn set_cursor_position()`

Even though this is more verbose (and causes some repetition), the `_position` suffix seems a bit clearer about the purpose, especially for those coming from a GUI perspective where setting the cursor changes the image that is used for the mouse cursor, and given that it's also possible to change the cursor shape (though not exposed by anything we do in Ratatui).

EdJoPaTo:
> Given the blast radius (good idea filtering out archives btw), I have a preference to make this change as an additive change with deprecation of the original method.

I prefer clean cuts especially as a trait is involved…

> Even though this is more verbose (and causes some repetition), the `_position` suffix seems a bit clearer about the purpose, especially for those coming from a GUI perspective where setting the cursor changes the image that is used for the mouse cursor, and given that it's also possible to change the cursor shape (though not exposed by anything we do in Ratatui).

…, but this is a good argument in the sense of improving the method name at the same time.

joshka:
The rationale for deprecate and then remove is that it provides a way to indicate how to fix a problem when upgrading. This is a way to enable churn without pain for users, and it's a way to build trust. When we can't do this, it's ok, but often we can.

src/terminal/terminal.rs
```diff
@@ -441,12 +457,12 @@ where
...
                 self.backend
-                    .set_cursor(self.viewport_area.left(), self.viewport_area.top())?;
+                    .set_cursor_position(self.viewport_area.as_position())?;
                 self.backend.clear_region(ClearType::AfterCursor)?;
             }
             Viewport::Fixed(area) => {
-                for row in area.top()..area.bottom() {
-                    self.backend.set_cursor(0, row)?;
+                for y in area.top()..area.bottom() {
+                    self.backend.set_cursor_position(Position { x: 0, y })?;
```
joshka, line 465:
```suggestion
                for row in area.rows() {
                    self.backend.set_cursor_position(Position { x: 0, row.y })?;
```

src/backend.rs
```diff
@@ -220,14 +223,31 @@ pub trait Backend {
...
+    /// backend.set_cursor_position(Position { x: 10, y: 20 })?;
+    /// assert_eq!(backend.get_cursor_position()?, Position { x: 10, y: 20 });
     /// # std::io::Result::Ok(())
     /// ```
+    fn set_cursor_position<P: Into<Position>>(&mut self, position: P) -> io::Result<()>;
+
+    /// Get the current cursor position on the terminal screen.
     ///
-    /// [`get_cursor`]: Backend::get_cursor
-    fn set_cursor(&mut self, x: u16, y: u16) -> io::Result<()>;
```
joshka, line 230: Aside (no action necessary just an FYI): Something that can make diffs like this easier to read (on GitHub at least) is to make additive changes like this below the original implementation that stays. E.g. see your change on Terminal below which puts the new method below the old one. I'm guessing here it made sense to put it above because we're grouping provided methods below required methods?

joshka (approved):
I think this is breaking still, but only for someone that's implementing their own backend, not for callers right? Do you think we should add a breaking changes line for this?

Approved as I'd be happy to merge this as-is, or with the suggested change / breaking doc if you prefer.

joshka:
This has merge conflicts (and allow edits by maintainers is turned off so we can't edit them on your behalf)

src/terminal/terminal.rs
EdJoPaTo: As only the y is needed the current approach seems simpler than to calculate a bunch of rects? #1299 reduces the performance overhead, but its still more complex than using a range of numbers.

src/backend.rs
EdJoPaTo: In this case, the diff is kind of horrible in any order as almost all lines contain changes, and it finds empty doc string lines (`///`) to match at random points resulting in a weird diff.

src/terminal/terminal.rs
joshka: no problem with not doing this. To my eyes it looks cleaner and the performance difference is minimal (microseconds / nanoseconds). How many rows per second should the terminal be expected to be able to clear?

joshka (approved):

PR #1285, opened 2024-08-04
feat!: replace Margin and Padding with Gaps

### LINKED ISSUE #931: Better `Margin` features with `Rect`

EdJoPaTo:
This is breaking for sure and will very likely generate a lot of churn.

Fixes #931

Also increases code readability as arguments always keep their intention in the method name. If that's not enough, the Struct pattern `Gaps {left: …, … }` should be used to keep the readability. I didn't create a `new` method for exactly this reason in order to keep readability of code up.

I intentionally `Padding::proportional` as characters on terminals are not 2x1 and it's only a very broad approximation. It's likely a better idea to use something custom for the given use case than some magic `n*2`.

### DIFF
--- a/benches/block.rs
+++ b/benches/block.rs
@@ -1,10 +1,10 @@
 use criterion::{criterion_group, criterion_main, BatchSize, Bencher, Criterion};
 use ratatui::{
     buffer::Buffer,
-    layout::{Alignment, Rect},
+    layout::{Alignment, Gaps, Rect},
     widgets::{
         block::{Position, Title},
-        Block, Padding, Widget,
+        Block, Widget,
     },
 };
 
@@ -30,7 +30,7 @@ fn block(c: &mut Criterion) {
         group.bench_with_input(
             format!("render_all_feature/{width}x{height}"),
             &Block::bordered()
-                .padding(Padding::new(5, 5, 2, 2))
+                .padding(Gaps::horizontal_vertical(5, 2))
                 .title("test title")
                 .title(
                     Title::from("bottom left title")
--- a/examples/block.rs
+++ b/examples/block.rs
@@ -28,12 +28,12 @@ use ratatui::{
         execute,
         terminal::{disable_raw_mode, enable_raw_mode, EnterAlternateScreen, LeaveAlternateScreen},
     },
-    layout::{Alignment, Constraint, Layout, Rect},
+    layout::{Alignment, Constraint, Gaps, Layout, Rect},
     style::{Style, Stylize},
     text::Line,
     widgets::{
         block::{Position, Title},
-        Block, BorderType, Borders, Padding, Paragraph, Wrap,
+        Block, BorderType, Borders, Paragraph, Wrap,
     },
     Frame,
 };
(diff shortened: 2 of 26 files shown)


joshka:
Consider whether this could be implemented with an `Into<Gaps>` approach that is (ignoring type inference) non-breaking on the public API. I'd be inclined to leave the deprecation alone and treat this as a way to add features without breaking existing ones.

I'm not sure that Gaps seems like the right name for this. Consider looking to CSS / Taffy / other GUI-ish libs and crates for inspiration in a name.

Is it possible that `Padding` is the right name to use here and the problem then becomes just convert Margin to Padding in all places?

EdJoPaTo:
> Consider whether this could be implemented with an `Into<Gaps>` approach that is (ignoring type inference) non-breaking on the public API. I'd be inclined to leave the deprecation alone and treat this as a way to add features without breaking existing ones.

I did. Not without removing the `const` possibilities of this and I think having `const` for something that is called often is quite useful.

As it creates a lot of churn I think it might be a good idea to either do a clean cut or just drop it and ignore these inconsistencies to exist (which would be annoying and smelly but well…)

> I'm not sure that Gaps seems like the right name for this. Consider looking to CSS / Taffy / other GUI-ish libs and crates for inspiration in a name.
> 
> Is it possible that `Padding` is the right name to use here and the problem then becomes just convert Margin to Padding in all places?

Not sure about the naming of Gaps either, but neither Padding nor Margin are a good name for it.

CSS: padding and margin are two opposites. They take the same type of arguments (CSS doesn't seem to give it a name), but their concepts are completely opposite.

At first, I implemented both Margin and Padding with a macro as their implementations are nearly identical, but this seemed wrong too. There should be a single struct being used for both concepts and the actual place that uses them should name it.

joshka:
Semantically / Linguistically, a margin is a form of padding, but padding is not a form of a margin, so Padding kinda makes sense still (even if it's not used in the same sense as CSS does). What we're really referring to is the amount of space to allocate on each direction of a rect.

> I did. Not without removing the const possibilities of this and I think having const for something that is called often is quite useful.

What are these concerns?

> As it creates a lot of churn I think it might be a good idea to either do a clean cut or just drop it and ignore these inconsistencies to exist (which would be annoying and smelly but well…)

I'm leaning towards dropping preferentially instead of Gaps (unless there's a really good name that makes sense), but would like to explore Into<Padding> a bit more before dropping this PR fully.

EdJoPaTo:
> > I did. Not without removing the const possibilities of this and I think having const for something that is called often is quite useful.
> 
> What are these concerns?

`const` allows to be run at compile time rather than runtime. When layouting stuff is a constant it can be done ahead of time as much as possible to remove that from runtime. Doesn't work as soon as the current detected area is used but for stuff with fixes sized for example it will not require anything to be computed at runtime at all.

Ideally I would like to have the complete layout module to be `const`.



> Semantically / Linguistically, a margin is a form of padding, but padding is not a form of a margin, so Padding kinda makes sense still (even if it's not used in the same sense as CSS does). What we're really referring to is the amount of space to allocate on each direction of a rect.

Hm… still would like to avoid the naming confusion with CSS.

joshka:
I meant, which parts of this can no longer be const that are performance sensitive?
There's probably some fundamental ratatui design things (stemming from the builder-lite pattern) that are causing this.

EdJoPaTo:
> I meant, which parts of this can no longer be const

Everything with `Into`

EdJoPaTo:
I currently don't see this getting merged as its breaking and not well fitting

PR #1288, opened 2024-08-05
docs(widgets): document stability of WidgetRef

joshka:
Addresses some confusion about when to implement `WidgetRef` vs `impl Widget for &W`. Notes the stability rationale and links to an issue that helps explain the context of where we're at in working this out.


kdheepak (approved):

kdheepak:
I like this improvement to the docs, even if I don't think I feel positively about recommending the 3 (or 6) trait impls:

- impl Widget for Foo
- impl Widget for &Foo
- impl Widget for &mut Foo
- impl StatefulWidget for Foo
- impl StatefulWidget for &Foo
- impl StatefulWidget for &mut Foo

orhun (approved):

Issue #1232, opened 2024-07-17
Rapid Releases


EdJoPaTo:
## Problem

Currently releases are further apart and features and especially bug fixes take time until they are released.

While Alpha releases are automated they are mostly pointless and nearly unused (see graph at the bottom of https://crates.io/crates/ratatui)

## Solution

The alpha is just the main branch. Cargo supports this natively anyway so why not promote this instead.

The existing release automation could just instead publish bugfix and feature releases unless there are breaking changes. Maybe have them trigger manually but do the rest automated like the alpha releases can currently.

This way there are more often releases and features and bug fixes are quicker at the users.

## Additional thoughts

Consider a 1.0.0 release. Currently breaking changes are prevented like it's a stable crate but the current server scheme 0.x.y suggests it's not stable yet. Release numbers would be more clear with a x.y.z version. Breaking change releases are not a big problem and versions starting with higher numbers are fine.

orhun:
First of all, thanks for bringing this up. Release automation is one of the topics I like to brainstorm and discuss a lot!

Before we delve into everything, let me review our current release process by examining its pros and cons to get a clearer picture of how things have been working thus far. I would say our release philosophy has been:

> _Release every 1-1.5 months with a detailed summary, along with automated alpha releases every week_.

### Release Workflow

1. Bump the version and update the changelog via `git-cliff`: https://github.com/ratatui-org/ratatui/pull/1196
2. Prepare a small write-up for the notable changes: https://github.com/ratatui-org/ratatui-website/pull/644
3. Pick a good time, release it, and announce it on Discord, Twitter, Mastodon, LinkedIn, Reddit, TWIR, HackerNews, and Lobsters.

#### Pros

- Every new feature, fix, or breaking change is well-documented through the summary and changelog.
- The content is beneficial for people visiting the website.
- Reduced possibility of regression (debatable).

#### Cons

- Takes time to complete a release due to manual work.
- Requires a social media plan/strategy – I don't think we can automate that.
- Large time gap between releases.

Alpha releases were intended for people who did not want to wait for a new release (and also for those who wanted to create a release on crates.io). It worked fine; I remember a couple of cases where people _really_ needed it. (e.g., @atuinsh used to depend on an RC release – it's definitely better than vendoring the whole Ratatui crate)

---

Now, how can we improve everything? Well, +1 to @joshka:

[...]

PR #1289, opened 2024-08-05
feat(terminal): Add ratatui::init() and restore() methods

joshka:
These are simple opinionated methods for creating a terminal that is
useful to use in most apps. The new init method creates a crossterm
backend writing to stdout, enables raw mode, enters the alternate
screen, and sets a panic handler that restores the terminal on panic.

A minimal hello world now looks a bit like:

```rust
use ratatui::{
    crossterm::event::{self, Event},
    text::Text,
    Frame,
};

fn main() {
    loop {
        terminal
            .draw(|frame: &mut Frame| frame.render_widget(Text::raw("Hello World!"), frame.area()))
            .expect("Failed to draw");
        if matches!(event::read().expect("failed to read event"), Event::Key(_)) {
            break;
        }
    }
    ratatui::restore();
}
```

A type alias `DefaultTerminal` is added to represent this terminal
type and to simplify any cases where applications need to pass this
terminal around. It is equivalent to:
`Terminal<CrosstermBackend<Stdout>>`

We also added `ratatui::try_init()` and `try_restore()`, for situations
where you might want to handle initialization errors yourself instead
of letting the panic handler fire and cleanup. Simple Apps should
prefer the `init` and `restore` functions over these functions.

Corresponding functions to allow passing a `TerminalOptions` with
a `Viewport` (e.g. inline, fixed) are also available (`init_with_options`,
and `try_init_with_options`).

The existing code to create a backend and terminal will remain and
is not deprecated by this approach. This just provides a simple one
line initialization using the common options.


joshka:
This is gross simplification of #1180. It's opinionated and does the right thing for the majority of applications. It also does the right thing with any app that uses `color_eyre::install()?`.

The only downside I see is type inference fails if we later want to implement the same methods for stderr / other backends, but I'm ok with taking a risk and dealing with it then rather than pre-planning to work around it.

src/terminal/terminal.rs
```diff
@@ -568,6 +568,65 @@ where
...
+    /// [`Terminal::with_options`].
+    ///
+    /// # Examples
+    ///
+    /// ```rust,no_run
+    /// let terminal = ratatui::Terminal::init()?;
+    /// # std::io::Result::Ok(())
+    /// ```
+    pub fn init() -> io::Result<Terminal<CrosstermBackend<Stdout>>> {
+        Self::set_panic_hook();
```
kdheepak, line 593: Just thinking out loud, are there instances where a user is going to _not_ want this?

src/terminal/terminal.rs
```diff
@@ -568,6 +568,65 @@ where
...
+    ///
+    /// For more control over the terminal initialization, use [`Terminal::new`] or
+    /// [`Terminal::with_options`].
+    ///
+    /// # Examples
+    ///
+    /// ```rust,no_run
+    /// let terminal = ratatui::Terminal::init()?;
+    /// # std::io::Result::Ok(())
+    /// ```
```
kdheepak, line 591: Should the docstring example use `color_eyre`?

```suggestion
    /// ```rust,no_run
    /// color_eyre::install()?;
    /// let terminal = ratatui::Terminal::init()?;
    /// # std::io::Result::Ok(())
    /// ```
```

joshka: If this is a reasonably acceptable approach, I'd do a pass over the entire crate and fixup examples. Probably not worth adding to the example here however, but it's worth noting some info about anyhow / color-eyre in the doc comments.

src/terminal/terminal.rs
```diff
@@ -568,6 +568,65 @@ where
...
+    /// restored to its original state. This includes disabling raw mode and leaving the alternate
+    /// screen buffer.
+    ///
+    /// # Examples
+    ///
+    /// ```rust,no_run
+    /// ratatui::Terminal::restore()?;
+    /// # std::io::Result::Ok(())
+    /// ```
+    pub fn restore() -> io::Result<()> {
```
EdJoPaTo, line 612: Should it be included in the Terminal Drop implementation? (with a `cfg(feature`)

Probably not as a different backend might have been used.

src/terminal/terminal.rs
```diff
@@ -568,6 +568,65 @@ where
...
+    /// ```
+    pub fn init() -> io::Result<Terminal<CrosstermBackend<Stdout>>> {
+        Self::set_panic_hook();
+        crossterm::terminal::enable_raw_mode()?;
+        crossterm::execute!(std::io::stdout(), crossterm::terminal::EnterAlternateScreen)?;
+        let backend = CrosstermBackend::new(std::io::stdout());
+        Ok(Terminal::new(backend)?)
+    }
+
+    /// Restores the terminal to its original state.
```
EdJoPaTo, line 600: It should be a bit more clear that it is meant to be used with the `crossterm` based init method.

```suggestion
    /// Restores the crossterm terminal on stdout to its original state.
```

EdJoPaTo (commented):
I like exploring this idea, but I think its coupling two currently separated approaches: Backend specific and Backend agnostic. So I would prefer something like this in the actual backend.

joshka:
> So I would prefer something like this in the actual backend.

Can you explain your preference here a bit more?

EdJoPaTo:
> > So I would prefer something like this in the actual backend.
> 
> Can you explain your preference here a bit more?

This is specific to `crossterm` so it should be in a place which is specific to `crossterm`, not generalized for multiple places. → Do stuff like this in `CrosstermBackend`.

src/terminal/terminal.rs
orhun: I we decide to go with this maybe we can also add e.g. `init_with_options(logging: bool)` or something similar in the future as well.

EdJoPaTo: The current approach is the verbose approach. This PR (or better #1180) introduces quick, minimal setup methods to have less setup needed for the common usage. If that's not fitting, the already existing more verbose code can still be used.

src/terminal/terminal.rs
joshka: Not for now for the reason mentioned.

joshka:
> > > So I would prefer something like this in the actual backend.
> > 
> > 
> > Can you explain your preference here a bit more?
> 
> This is specific to `crossterm` so it should be in a place which is specific to `crossterm`, not generalized for multiple places. → Do stuff like this in `CrosstermBackend`.

So my motivation for putting this on Terminal is to:
- reduce the need to "choose" a backend or be aware of that choice.
- reduce the need to know about the backend being a thing - most apps will never call any methods on the backend and so this is really an implementation detail
- be opinionated that crossterm is the backend that we choose to support most thoroughly (as termion is not available on Windows devices and the Termwiz backend is too new to judge even though it might be better)

If we wanted to make that this is explicitly name crossterm as part of the function name, I'd instead recommend some free functions like `ratatui::crossterm_terminal()` or `ratatui::terminal_with_crossterm()`. Even something like `ratatui::init_terminal()` / `ratatui::init()` would be preferable to adding more methods to the backend I think.

I want to understand better where the disconnect on this is:
1. Do we agree that we should have a single method for initialization?
2. Do we agree that choosing a backend for the user is helpful?
3. Do we agree that Crossterm is the right backend to choose as a reasonable default?
4. Do we agree that we should enable raw mode and the alternate screen as a default?
5. Do we agree that if we're choosing the backend, then calling out that choice is not material to library user?
6. For anything that we disagree on, what percentage of apps built with Ratatui do you think would not want to use that setting / approach? How many of these are more advanced use cases where the users can reasonably be expected to just call the more verbose initialization approaches?
7. What similar initialization strategies (e.g. what crates that require init) are we looking to for guidance on common patterns? (I'm looking towards the tracing_subscriber crate as a bit of an influence to this sort of initialization, do you have any similar data points)

orhun (approved):
This is a good start towards #1321, LGTM.

Next steps would be,

- updating the examples (to make this more visible)
- updating the documentation / guides

Is there anything blocking the merge?

joshka:
I think `ratatui::init()` might be an even better idea than `Terminal::init`

orhun:
Yup, then are we going to support multiple backends (based on feature flags) rather than only crossterm?

joshka:
> Yup, then are we going to support multiple backends (based on feature flags) rather than only crossterm?

I think it's reasonable to choose a known good path (crossterm), and make that the default approach. If someone wants to do something different, the other backend configurations are still available.

To answer my 7 questions above:
1. We should have a single method for init
2. Choosing the backend is useful
3. Crossterm is the right backend to choose
4. enabling raw mode and alternate screen is the right thing to do when initializing
5. Making the user have to provide information on which backend to use should not be necessary
6. More advanced use cases can fall back to using the existing code reasonably
7. Already answered with https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/fn.init.html

joshka:
About half the examples redone - more to come.

joshka:
Ready for review now. Everything tidied up in a pretty consistent form - there's still a few examples that don't use an App struct, but they're mostly ok (and I didn't see the need to fully go down that path)

Net lines removed: 1.1K (about 28 lines per example)

image

kdheepak:
I like the idea of a simplification a lot. 

Just wanted to check if this API is tied to `crossterm` in anyway? If in the future, we want to swap `ratatui::init()` to use termwiz, would the API break? 

This also makes it feel like ratatui is blessing `crossterm` as the preferred backend, although we may have been implicitly doing that already. One thing that I've been meaning to do is a more detailed comparison of crossterm and termwiz.

fwiw, I think I'd be more onboard with `ratatui::terminal::crossterm::init()` at the moment.

joshka:
> This also makes it feel like ratatui is blessing crossterm as the preferred backend, although we may have been implicitly doing that already. One thing that I've been meaning to do is a more detailed comparison of crossterm and termwiz.

https://github.com/ratatui-org/ratatui/pull/1289#issuecomment-2286869916 has my answers to this.

Crossterm is used 3:1 in crates, and downloaded 4:1 compared to termion. Termion to termwiz is about 20:1 deps / 4:1 downloads. Termion doesn't work on Windows. The only time that termion should be chosen is for apps that will never make sense to use on windows (e.g. linux / unix specific tooling)

1. Simplifying Ratatui init to a well-known path so that users don't have to make and understand choices early in an app's development is a reasonable, useful, desired approach.
2. A backend choice is an unnecessary one to make when crossterm is the right choice for most apps
3. Exposing information about the backend choice as part of the simplified API's path doesn't make sense when the intent of simplification is that we're choosing the backend on behalf of the user.
4. Nothing about choosing these defaults prevent users from making an active choice to do something different, but choosing crossterm will work for all operating systems, while choosing termion does not. Choosing termwiz doesn't have enough mind share to be worried about.

But looking at forward options:
1. We can choose crossterm and leave it at that
2. We can later change the implementation of `ratatui::init` to `if cfg(crossterm) { init_crossterm() } else if cfg(termion) { init_termion() } else if cfg(termwiz) { init_termwiz() }`. I currently don't think we should do that, but there is a two way door if I'm wrong.
3. The init methods can be replaced internally with something similar to a simplified builder pattern (like what's used in tracing) at some later date if we wanted, but it doesn't add a huge amount of value.

TL;DR: there's significant value in having a single useful default like this. I feel strongly that `ratatui::init` -> `crossterm` is the right one. There's good rationale to back that perspective, and ways to proceed for users that don't want the default, and ways to modify that if this is common.

orhun:
I like the idea of going with the 2nd option (`if cfg(crossterm)...`) based on the level of adoption for `ratatui::init` in the future. Adding `crossterm` as default sounds like the most sensible option to me as well.

Though a detailed comparison of the backends would be nice to have in the documentation I believe.

kdheepak:
If adding the `cfg` approach is possible in the future, I'm good to go forward with this PR.

joshka:
> If adding the `cfg` approach is possible in the future, I'm good to go forward with this PR.

Yep - adding an alternate init module that has termion initialization code and which is gated on not(crossterm) and termion works as expected (noting the caveat that we really shouldn't do this for various reasons - notably it requires breaking feature unification when crossterm and termion features are both enabled)

The following test succeeds:
```
    #[test]
    #[cfg(any(feature = "crossterm", feature = "termion"))]
    fn init() {
        let _terminal = super::init();
        super::restore();
    }
```
With:
```rust
#[cfg(all(not(feature = "crossterm"), feature = "termion"))]
mod init {
    use termion::{raw::IntoRawMode, screen::ToAlternateScreen};

    use super::*;
    use crate::backend::TermionBackend;
    use termion::{raw::IntoRawMode, screen::IntoAlternateScreen};
    pub type DefaultTerminal = Terminal<TermionBackend<std::io::Stdout>>;

    pub fn init() -> DefaultTerminal {
        let stdout = std::io::stdout()
            .into_raw_mode()
            .unwrap()
            .into_alternate_screen()
            .unwrap();
        let backend = TermionBackend::new(stdout);
        Terminal::new(backend).expect("Failed to initialize terminal")
    }

    pub fn restore() {}
```

orhun (approved):
Tested this again and LGTM. We should create a point release pretty soon with the changes in this. 

Maybe also we can also remove alpha releases and update guidelines before that - #1232. Will create a PR.

foxyseta:
Should the tutorials and documentation be updated now?

orhun:
Definitely!

PR #1299, opened 2024-08-06
perf(rect): improve Rect::rows/columns performance

EdJoPaTo:
Use less math involved in `Rect::new` as the partial new Rect will always be smaller than the max area.

### DIFF
--- a/src/layout/rect/iter.rs
+++ b/src/layout/rect/iter.rs
@@ -28,7 +28,11 @@ impl Iterator for Rows {
         if self.current_row >= self.rect.bottom() {
             return None;
         }
-        let row = Rect::new(self.rect.x, self.current_row, self.rect.width, 1);
+        let row = Rect {
+            y: self.current_row,
+            height: 1,
+            ..self.rect
+        };
         self.current_row += 1;
         Some(row)
     }
@@ -62,7 +66,11 @@ impl Iterator for Columns {
         if self.current_column >= self.rect.right() {
             return None;
         }
-        let column = Rect::new(self.current_column, self.rect.y, 1, self.rect.height);
+        let column = Rect {
+            x: self.current_column,
+            width: 1,
+            ..self.rect
+        };
         self.current_column += 1;
         Some(column)
     }


joshka:
Oddly, I see a significant performance regression when testing this:

```
rect_rows/rows/16       time:   [5.6676 ns 5.6870 ns 5.7165 ns]
                        change: [+10.251% +10.967% +12.075%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 5 outliers among 100 measurements (5.00%)
  3 (3.00%) high mild
  2 (2.00%) high severe
rect_rows/rows/1024     time:   [518.40 ns 519.82 ns 521.38 ns]
                        change: [+75.768% +76.510% +77.263%] (p = 0.00 < 0.05)
                        Performance has regressed.
  1 (1.00%) low mild
  3 (3.00%) high mild
  2 (2.00%) high severe
rect_rows/rows/65535    time:   [32.813 µs 32.904 µs 33.007 µs]
                        change: [+79.557% +80.313% +81.059%] (p = 0.00 < 0.05)
                        Performance has regressed.
  2 (2.00%) low mild
  3 (3.00%) high mild
  1 (1.00%) high severe
```
https://github.com/ratatui-org/ratatui/pull/1301

joshka:
> Oddly, I see a significant performance regression when testing this:
> 
> ```
> rect_rows/rows/16       time:   [5.6676 ns 5.6870 ns 5.7165 ns]
>                         change: [+10.251% +10.967% +12.075%] (p = 0.00 < 0.05)
>                         Performance has regressed.
> Found 5 outliers among 100 measurements (5.00%)
>   3 (3.00%) high mild
>   2 (2.00%) high severe
> rect_rows/rows/1024     time:   [518.40 ns 519.82 ns 521.38 ns]
>                         change: [+75.768% +76.510% +77.263%] (p = 0.00 < 0.05)
>                         Performance has regressed.
> Found 6 outliers among 100 measurements (6.00%)
>   1 (1.00%) low mild
>   3 (3.00%) high mild
>   2 (2.00%) high severe
> rect_rows/rows/65535    time:   [32.813 µs 32.904 µs 33.007 µs]
>                         change: [+79.557% +80.313% +81.059%] (p = 0.00 < 0.05)
>                         Performance has regressed.
> Found 6 outliers among 100 measurements (6.00%)
>   2 (2.00%) low mild
>   3 (3.00%) high mild
>   1 (1.00%) high severe
> ```
> 
> #1301

Benchmarks are merged for rect. How does this look on your RPi?

EdJoPaTo:
Whatever is optimized here, its dark magic. There seem to be two performance levels, either the fast one or the slow one. Using `Rect::new` is faster than `Rect {…}`. Which is weird. Removing the math from `Rect::new` also results in the slower performance level. Copying the whole `Rect::new` code into the `next` method also ends up with the slow performance level. So the fast one is only when its both in its own method AND includes the `if` which doesn't even use the floating point math then… Looking at the resulting assembler (`cargo-show-asm`) it includes more branch points and more code, so the method should be significantly faster without that.

Then I changed the benchmark code. Either `Vec::push` or summing up all `row.y` to have some simple call in there and that's a different story. `Vec::push` is relatively slow so not much to see there. They are within error tolerances. Summing happens on the stack, no heap involved and its ~20% faster with this PR. Interestingly a for loop with `+=` is also way faster than `.sum()` so the difference with `.sum()` ends up within the error margins too.

So… whatever the current benchmark code does or what is optimized away there, it's completely opposite of what's happening in the code: less floating point math.

Looking at the absolute numbers its also interesting that the `+=` sum is faster than the current benchmark. So the current benchmark is optimizing something, but it's still doing more than `+= rect.y`?

Benchmarks with stable or nightly Rust channel are about the same. Linux arm64/v8 and Linux armv7 (Pi 2 and 4) are more or less comparable to Linux amd64. 

Tldr: Dark Magic.

---

Personally I would go with the code that does less and is therefore easier to get as less stuff happens in the background.

Also, not sure what to do with the currently existing rect benchmarks as I don't even fully understand what it actually does compared to runtimes with other things tested I tested.

joshka:
I suspect that the current benchmark optimizes the row out, or there is indeed black magic that realizes that this is just an iterator and so most of the floating point logic is never hit for sequential iterations of the loop). I'd suggest updating the benchmark as part of this PR to something that performs some more realistic use case (i.e. uses the row as an index into a buffer of the correct size to read a value). `Sum` / `+=` are both synthetic enough that they don't represent any real use case for this method, and would / are optimized differently.

Your confusion about what to do next stems from not following the correct perf tuning methodology. Starting with a hypothesized problem rather than an actual one often leads to such things. I'd recommend start from an actual app, and pull out some actual code that uses this to understand how to represent some real world performance of this actually is impacted by this code.

Perf tuning approach I use:
1. Establish what is good enough
2. Measure, don't guess
3. Fix the biggest bottleneck first
4. Measure after fixing
And generally:
5. Avoid micro-benchmarks (i.e. things not at the entire system level)
6. Be careful with synthetic data
7. Know your general estimates (e.g. cache, memory, disk, network speeds)

You'd laugh probably about this. I wanted to use some rects in a const context in the Rect benchmark code (e.g. `const AREAS: [Rect; 3] = ... `), but was foiled by the new method not being const.

Taking the broader picture, the right thing to fix here is the existence of the clamping. I think it probably shouldn't exist, but the right thing to do is understand why it does / what problem it was put in place to solve, and then move away from it with the knowledge of what problems it will cause by removing it. I suspect it was entirely about keeping the area return value as u16. But It probably just makes sense to break things and make that u32 / usize.

EdJoPaTo:
> Your confusion about what to do next

The method is called `Iterator::next`… not what to do next…

I said what I would do…

> Starting with a hypothesized problem rather than an actual one often leads to such things.

This benchmark was introduced with #1301, not by me. I did not create this unrealistic benchmark. I looked at the code and simplified the code to have less stuff involved, reducing the complexity.

Getting rid of the math in `Rect::new` by using `u32` for the area would be best.
Fun fact: It would result in this benchmark performing worse.

> Also, not sure what to do with the currently existing rect benchmarks as I don't even fully understand what it actually does

Maybe adding to what I would do: remove the benchmark.

joshka:
> This benchmark was introduced with #1301, not by me. I did not create this unrealistic benchmark. I looked at the code and simplified the code to have less stuff involved, reducing the complexity.

Yup - I added that because this PR says "improves Performance" but doesn't actually measure that. I noticed the oddity that it doesn't fix the performance however. Entirely possible the benchmark is wrong in some way, but if this is about fixing performance then it's important to show that it fixes performance.

There's no floating point arithmetic happening here that is skipped as that path is never taken when you pass 1 as the height here. This PR is skipping mostly fast (nanosecond scale) things:
- a function call,
- integer widening from u16 to u32
- a check whether width*1 > u16::MAX (which is always false and possibly something that gets optimized out if the function is automatically inlined, which it should be)
- variable assignments

I wrote a bunch above about how to address performance concerns, but put more simply "don't fix what ain't broke".

PR #1302, opened 2024-08-06
perf: move benchmarks into a single benchmark harness

joshka:
Consolidates the benchmarks into a single executable rather than having to create a new cargo.toml setting per and makes it easier to rearrange these when adding new benchmarks.

### DIFF
--- a/benches/main/line.rs
+++ b/benches/main/line.rs
@@ -1,6 +1,6 @@
 use std::hint::black_box;
 
-use criterion::{criterion_group, criterion_main, Criterion};
+use criterion::{criterion_group, Criterion};
 use ratatui::{
     buffer::Buffer,
     layout::{Alignment, Rect},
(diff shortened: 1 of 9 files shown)


benches/main/line.rs
```diff
@@ -36,4 +36,4 @@ fn line_render(criterion: &mut Criterion) {
 }
 
 criterion_group!(benches, line_render);
-criterion_main!(benches);
+// criterion_main!(benches);
```
orhun, line 39: Why not delete these? 💭

joshka: I did locally - just missed committing that :)

orhun (approved):

benches/main/line.rs
joshka: a20d8aca4...4384ca479

PR #1307, opened 2024-08-06
feat(logo): shorter logo now in widget form

joshka:
Just an idea that I'm playing with, a subtle modification to the height of the logo so that it can be rendered nicely on 2 lines rather than 2.5

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ cassowary = "0.3"
 compact_str = "0.8.0"
 crossterm = { version = "0.28.1", optional = true }
 document-features = { version = "0.2.7", optional = true }
+indoc = "2"
 instability = "0.3.1"
 itertools = "0.13"
 lru = "0.12.0"
--- a/examples/ratatui-logo.rs
+++ b/examples/ratatui-logo.rs
@@ -13,57 +13,42 @@
 //! [examples]: https://github.com/ratatui/ratatui/blob/main/examples
 //! [examples readme]: https://github.com/ratatui/ratatui/blob/main/examples/README.md
 
-use std::{
-    io::{self},
-    thread::sleep,
-    time::Duration,
+use std::env::args;
+
+use color_eyre::Result;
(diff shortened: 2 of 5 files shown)


Cargo.toml
```diff
@@ -29,6 +29,7 @@ cassowary = "0.3"
 compact_str = "0.8.0"
 crossterm = { version = "0.28", optional = true }
 document-features = { version = "0.2.7", optional = true }
+indoc = "2"
```
joshka, line 32: I don't recall exactly why we removed indoc from the main deps, but it's possible we'd need to avoid it here (I think it was possibly due to someone running on android). Check before any potential merge.

kdheepak: for context, it was removed in this PR: https://github.com/ratatui-org/ratatui/pull/1007

orhun:
Looks good in my terminal, smol ratatui

joshka:
It can even be a little smaller...
image

joshka:
BTW, in that change I also fixed the font kerning around the A / T to look a bit nicer. Altogether:

Current:
image
First update
ratatui-logo
Font Kerning fixed:
ratatui-logo
Tiny:
image

orhun:
> First update
> Font Kerning fixed:

Hmm, the first one looks better to me. Second one's T has a space in it.

joshka:
> > First update
> > Font Kerning fixed:
> 
> Hmm, the first one looks better to me. Second one's T has a space in it.

I'm not sure if you're referring to the 2nd or third image there.

orhun:
Sorry, I meant this one looks better:

> First update
image

Than:

> Font Kerning fixed:
image

orhun:
rebased on main, fixed conflicts & applied clippy suggestions

orhun (approved):

orhun:
let's wrap this up for 0.29.0 :)

joshka:
Tweaked this a bit to add docs, tests, neater code, etc.

Rendered docs:

image

PR #1312, opened 2024-08-07
fix(span): dont render control characters

### LINKED ISSUE #1259: Deprecate/Remove `Buffer::set_line` and `Buffer::set_span`
## Problem

`Span::render` and `Buffer::set_span` exists. Same for line. They use different logic and contain various degrees of bugfixes / complexity.

## Solution

There should be a single way of rendering a Span / Line including all the bugfixes.

Deprecate them and replace their internal usage by the `Span/Line::render` methods.

## Alternatives

Otherwise, point `Span/Line::render` to `Buffer::set_line/span` to reuse the same implementation internally.

But it's probably a good idea to get the higher level stuff out of `Buffer`.

EdJoPaTo:
Fixes #1226 yet another time as `Span` and therefore `Line` and `Text` render differently than #1226 expected.

No there are no tests for this included as there should be a single way of rendering stuff and not multiple ones, so the existing tests should point at the single way of rendering. If #1259 gets fixed the tests from #1226 would also test this.


joshka:
As previously mentioned, Span and Line's implementation intentionally differs from the methods on Buffer as the methods on buffer were both incorrect in a few ways (e.g. not fully styling the entire area of a line), while being so widely used that changing this behavior would be seen as a breaking change in a way that would cause problems for just about every application built on Ratatui.

joshka (approved):
Can you please rebase this on main

joshka (approved):

joshka:
Thanks @EdJoPaTo

PR #1317, opened 2024-08-11
build: simplify Windows build

joshka:
Termion is not supported on Windows, so we need to avoid building it.

Adds a conditional dependency to the Cargo.toml file to only include
termion when the target is not Windows. This allows contributors to
build using the `--all-features` flag on Windows rather than needing
to specify the features individually.

---

Testing:
cargo build --all-features now succeeds on windows (11 Arm 64)

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -5,14 +5,7 @@ skip_core_tasks = true
 
 [env]
 # all features except the backend ones
-ALL_FEATURES = "all-widgets,macros,serde"
-
-[env.ALL_FEATURES_FLAG]
-# Windows does not support building termion, so this avoids the build failure by providing two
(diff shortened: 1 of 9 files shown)


Makefile.toml
```diff
@@ -23,7 +16,7 @@ dependencies = ["lint", "clippy", "check", "test"]
 
 [tasks.lint]
 description = "Lint code style (formatting, typos, docs, markdown)"
-dependencies = ["lint-format", "lint-typos", "lint-docs", "lint-markdown"]
+dependencies = ["lint-format", "lint-typos", "lint-docs"]
```
joshka, line 19: Removed for now. This is unrelated, but will re-enable when this is reconfigured to correctly ignore the generated parts.

orhun (approved):

PR #1318, opened 2024-08-11
chore(ci): use cargo-docs-rs to lint docs

joshka:


orhun (approved):

orhun:
Out of curiosity, can you give an example of a (edge) case we can catch with this CI check?

joshka:
Anything that is doc specific. E.g.: the following mismatched code fence:
````rust
//! ```
//! ``
```
Or a link to a non-existent item:
```rust
/// [`foo`]
```
https://doc.rust-lang.org/rustdoc/lints.html has the lints.

We use nightly for docs because we like several of the features that are only available in nightly
- Writing the feature flag next to code that requires that flag to be enabled https://doc.rust-lang.org/rustdoc/unstable-features.html#doccfg-recording-what-platforms-or-features-are-required-for-code-to-be-present (which we use via https://doc.rust-lang.org/rustdoc/unstable-features.html#doc_auto_cfg-automatically-generate-doccfg
- scraped examples https://doc.rust-lang.org/rustdoc/scraped-examples.html

Switching to cargo-docs-rs over our custom params means that we don't have to track the exact params that docs-rs will use to build our docs, which means that the CI tests the same thing that will be eventually deployed.

orhun:
Got it, thanks for the example 👍🏼

PR #1320, opened 2024-08-11
fix: remove trailing newline from single-line Text::Display

LucasPickering:
The `Display` impl for `Text` intentionally omits a trailing newline for the final line. This logic wasn't working for single-line text though. I added an additional text case, which previously was failing and now passes.


orhun (approved):
LGTM

joshka (approved):
Thanks

PR #1330, opened 2024-08-17
Doc only fixes: its vs it's mistakes

matta:
...nothing too major.

it's --> "it is"
its --> "it" possesses something

I noticed this in the deprecation message for Frame::area, and went looking for more mistakes. I only found three in total.


joshka (approved):

joshka:
Thanks!

Issue #1250, opened 2024-07-29
Column offset and selection in `TableState`


achristmascarl:
## Problem
For tables with many columns, there isn't currently a way to offset the columns that are rendered or to select a column.

## Solution
Add `column_offset` and `column_selected` fields to `TableState` that are used when rendering, similar to how the [row] `offset` and `selected` are currently used.

joshka:
There's a semi-related PR to this which has stalled out due to timing and some incompatible changes that were made in parallel: https://github.com/ratatui-org/ratatui/pull/516

It's a good idea generally that I'd like to see made to work. In particular, this would be useful to hook up to horizontal mouse scrolling or something equivalent (e.g. Shift + Arrow keys) in a demo.

airblast-dev:
Can a maintainer assign this to me? 

I am thinking of making two separate PR's, one for applying styles for the selected column and one for allowing to set the highlight symbol next to the cell (similar to what the mentioned PR offered with `ColumnHighlightSpacing`) as I think doing the changes as seperate PR's would make it easier to review.

orhun:
You got it!

joshka:
I'm a big fan of smaller PRs that just do one thing well. They make it easy to review.

airblast-dev:
I want to give a little update on my progress. Column selection, column highlighting, and cell highlighting are done. These are mostly ready for a PR.

## Problem:
I am not really sure how the column offset should behave with keeping the constraints (the widths field in `Table`) in mind. The point of allowing the user to set a column offset is to give the ability to navigate a large number of columns similar to how we can navigate many rows in `Table`, and `List`.

In the case of `Constraint::Max`, and `Constraint::Length` the intention of shrinking is explicit if the values are lesser than the columns minimally required width.  But with the other constraints (except `Constraint::Min`) we don't really now if shrinking the column is acceptable. We can check if the column could fit, but even then we can't be sure until the layout is created since the `Constraint` is an ambiguous width until the layout is processed. 

 I am not really sure what the correct behavior should be when an offset is set with constraints that contain the mentioned "ambiguous" constraints.

I think checking if the Column minimally fits (through a method such as `Cell::width`) and using its respective `Constraint`  if it does fit, seems the best way to do this. However even doing this isn't great as any externally computed and provided widths now have to be refactored (for example a user might want to allow every column to shrink, except the one that is selected instead of making it scrollable). This change would disallow shrinking were it is not explicit.

[...]

Issue #1215, opened 2024-07-04
Consolidate `List::get_items_bounds` and `Table::get_row_bounds`


kdheepak:
Both `List::get_items_bounds` and `Table::get_row_bounds` perform similar enough tasks that I'm wondering if there's an opportunity to consolidate these. I think these are useful functions that a generalized version could be made `pub`. Related design question - is it possible to move these methods to the `ListState` or `TableState`?

More generally, `Table` and `List` have some overlap and it'd be nice to explore whether there is opportunities for more alignment. At the very least it'd be good to document when a user should use one or the other, what the tradeoffs are etc.

orhun:
> More generally, `Table` and `List` have some overlap and it'd be nice to explore whether there is opportunities for more alignment.

Not sure about the naming but I think we should merge them under a single struct at some point. I had an (useless?) attempt at #1155 which made me realize we have enough overlap to generalize their behavior.

joshka:
One downside to merging the  list and table state would be that when we introduce divergent behavior, we end up with problems splitting the two implementations. E.g. the PR for selecting table columns and cells instead of just rows is not applicable to the list, and we might want to add sorting to the headers, and various other things too.

One possible approach to this might be to use a macroed impl to ensure that these are the same, but the payoff on the code duplication vs complexity is hard to justify for just 2 instances.

josueBarretogit:
I think you guys could do something like this : 
```rust 
use std::marker::PhantomData;

#[derive(Default)]
pub struct TableState;

#[derive(Default)]
pub struct ListState;

#[derive(Default)]
pub struct ScrollState<T> {
    offset: usize,
    selected: Option<usize>,
    _phantom: PhantomData<T>,
}

/// Tablestate-specific methods
impl ScrollState<TableState> {
    pub fn select_header(&mut self) {}
    pub fn select_next_column(&mut self) {}
}

impl ScrollState<ListState> {
    pub fn some_list_state_speficic_method(&mut self) {}
}

/// Both ListState and Tablestate will have access to this methods
impl<T> ScrollState<T> {
    pub fn select_next(&mut self) {}
    pub fn select_previous(&mut self) {}
    pub fn select_last(&mut self) {}
}

fn main() {
    let mut list_state = ScrollState::<ListState>::default();
    list_state.some_list_state_speficic_method();

    let mut table_state = ScrollState::<TableState>::default();
    table_state.select_header();

    list_state.select_next();

    table_state.select_next();
}


```

[...]

PR #1331, opened 2024-08-19
feat(table)!: add support for selecting column and cell

airblast-dev:
Fixes https://github.com/ratatui-org/ratatui/issues/1250

Adds support for selecting a column and cell in `TableState`.  The selected column, and cells style can be set by `Table::column_highlight_style` and `Table::cell_highlight_style` respectively. 

The table example has also been updated to display the new functionality:

https://github.com/user-attachments/assets/e5fd2858-4931-4ce1-a2f6-a5ea1eacbecc

BREAKING CHANGE: The Serialized output of the state will now include the "selected_column" field.  Software that manually parse the serialized the output (with anything other than the `Serialize` implementation on `TableState`) may have to be refactored if the "selected_column" field is not accounted for. This does not affect users who rely on the `Deserialize`, or `Serialize` implementation on the state.

BREAKING CHANGE: The `Table::highlight_style` is now deprecated in favor of `Table::row_highlight_style`.


joshka:
Thanks for the PR. I've asked the previous author of #516 @tonogdlp to take a look at this with some guidance about what to review. Some things to think about when looking at this PR:

- how does this impact existing table users (i.e. if there's an app with a table, what breaks with the new code?) We don't mind breaking things, but I like knowing how big the break is before doing it.
- does the API seem ergonomic?
- does the feature have any edge cases that aren't immediately obvious?
- how well is it documented? Can you understand how it's used from the docs?
- are there any things that would make it difficult to evolve the code in the future?
- subjectively how easy is the new code to maintain? Do you understand the algorithms? Does reading the code require holding a large amount of state in your brain or is this a manageable amount?
- there's probably questions I'm missing. This list is a good one to use in general for code reviews, so feel free to add any that you think would help us more generally review PRs.

airblast-dev:
Sounds good! There is a few parts that I am not exactly happy about, such as not using an enum like the older PR. I have explained the reasons why below.

Compared to https://github.com/ratatui-org/ratatui/pull/516 there's a few things missing and/or different.

This PR does not add any highlight symbol functionality unlike the previous PR. There is already 800+ lines changed, so I think its better to separate the two changes as separate PR's. If it's preferred to have them both in a single PR I am open to working on it and adding it here or in a new PR.

## Design Choices
I decided it would be better to keep the selected column as a field instead of introducing an enum (namely `TableSelection` in mentioned PR). 
There is a few reasons for this: 
- `TableState::selected_mut` returns an `&mut Option<usize>` which means using an enum such as `TableSelection` would not be possible without removing the method.
- Serializing and deserializing the state would break. Though we can use a middleman struct with a derive as a practical workaround. This is required as breaking change in deserialization is not acceptable as every user would have to manually update the stored state to be compatible.
-  `TableState::with_selected` takes an `impl Into<Option<usize>>` which means unlike in the previous PR we can't just accept older arguments since it would require a generic. (there may be a way to do this without a breaking change, but there's still the reasons mentioned above)

I do think that an enum is more elegant in this case. Perhaps we should deprecate `selected_mut`, `selected`, and `with_selected` methods now and use an enum after the next major release (0.29.0)? 
I think it might be a good idea to deprecate methods like `select` in favor of methods named like `select_row` to add specificity to the method name. 

## Doc Changes
- Parts of documentation that mention the rows as "items" were replaced with "rows" instead since it could also be understood as a column or cell.

## Testing
I have added testing for the feature and newly introduced methods. I am open to adding more if needed.

src/widgets/table/table.rs
```diff
@@ -1871,4 +2147,52 @@ mod tests {
                 .remove_modifier(Modifier::CROSSED_OUT)
         );
     }
+
+    use rstest::rstest;
```
orhun, line 2151: It would be nice to move this (and other `rstest` imports) to the top of the tests module.

src/widgets/table/table_state.rs
```diff
@@ -49,6 +53,7 @@
 pub struct TableState {
     pub(crate) offset: usize,
     pub(crate) selected: Option<usize>,
+    pub(crate) selected_column: Option<usize>,
```
orhun, line 56: One thing to keep in mind is how would this new field play out if we decide to consolidate the functionalities of `TableState` and `ListState` in the future. Hint: #1215

orhun (commented):
Obviously there might be some points about the designs and implementation but I overall like this new functionality.

src/widgets/table/table_state.rs
airblast-dev: Yes, I already saw that issue. It is also one of the reasons I decided to leave the selection methods as is for now (instead of renaming the methods `select_row` and so on). Though I am not sure what I should change to avoid a problem. I don't think this is a problem for the consolidation as we will still have to store an `Option<usize>` somewhere in the table state. If you have any ideas on what should be done to make this more compatible, I am happy to make the changes. I just don't know what should be changed here.

src/widgets/table/table.rs
airblast-dev: Moved this one. Not sure how to move the other rstest imports since they are in separate modules.

src/widgets/table/table_state.rs
orhun: Yeah, I'm not sure what to change here as well. It should be fine. I just wanted to bring that up as a point related to `TableState`. I would very much like to avoid a lot of changes between these two states because I believe they should be consolidated in the future.

src/widgets/table/table.rs
orhun: Yup, I think it would be better to import them from the top-level tests module instead of explicitly every time. Not a biggie though.

tonogdlp:
Hi! I'm the one who made #516 .. Happy to give my thoughts as I thought a lot about the possible implementations of this feature. 

My first comment is that this PR is way better structured than the one I made before. It is simpler to read, and I believe it has a much better scope than mine. Tests and docs are super complete. Kudos!


--- 

In a previous PR that was closed by my mistake (#478), I had an implementation similar as the one @airblast-dev has here.. But @joshka suggested using an `enum` here: https://github.com/ratatui-org/ratatui/pull/478#discussion_r1320958860


I think the `enum` implementation is better for final users, even if a little more convoluted to implement internally. One of its main benefits is that  forces you to either explicitly ignore or actively address cases.


Given the selection enum:

```rust
pub enum TableSelection {
    Row(usize),
    Col(usize),
    Cell { row: usize, col: usize },
}
```

we can compare the ability to explicitly match the enum
```rust
match selected {
    Cell(r, c) => {},
    Row(r) => {},
     _ => return   // explicitly ignore Column variant
}
```

with 

```rust
if selected.is_some() and selected_column.is_some() {
.....
}
return //implicitly ignores both the 'Only row selected' and 'Only column selected'
```

IMHO, the `enum` implementation gives you the type safety people have come to expect from Rust.

--- 

In terms of backwards compatibility, we can preserve most (if not all) of the previous code working by implementing the Into/From trait, adding new methods that return `TableSelection` and at the same time deprecating previous methods (while leaving their functionality exactly the same):

```rust
impl From<usize> for TableSelection {
    /// Convert from a usize to a TableSelection::Row, to preserve backwards compatibility
    fn from(row: usize) -> Self {
        Self::Row(row)
    }
}


/// Returns the current selected index, if any
  #[deprecated]
  pub fn selected(&self) -> Option<usize> {
      self.selected_row()
  }

  /// Returns the current selected index, if any
  pub fn selection(&self) -> Option<TableSelection> {
      self.selected
  }
```

---
I also agree with @airblast-dev that other features like the implementation of how the `highlighting_symbol` works should be left for another PR when this one has been ironed out.

src/widgets/table/table.rs
airblast-dev: Done :+1:

airblast-dev:
I definitely agree that using an enum allows us express things better in this case. 

There is a few reasons why I decided not to go through with it:
- `TableState::selected_mut` returns an `&mut Option<usize>`. When using an enum such as the one you used we lose the ability to return it. Based on the [deprecation notice](https://github.com/ratatui-org/ratatui/blob/main/CONTRIBUTING.md#deprecation-notice) we should at least wait one version before removing a method. Not fully sure that this applies here but its at least worth mentioning.
- Currently the `with_selected*` methods accept an `impl Into<Option<usize>>` replacing the `usize` with `Into<TableSelection>` works but now the user will have to provide a generic to these methods. Meaning it is still a breaking change. We can deprecate this to be left as is, and provide a `with_selection` instead.
- Using this enum could cause `TableState` and `ListState` to diverge. Depending on how the state methods are consolidated this might not be an issue, but I certainly think it could make things more difficult. See https://github.com/ratatui-org/ratatui/issues/1215 and https://github.com/ratatui-org/ratatui/pull/1331#discussion_r1723814610.
- Changing the type to the enum will require us to rethink the serialization, and deserialization. Keeping in mind any change must be backwards compatible, and requiring people to manually update the state is not acceptable. The simplest solution would be to derive the traits on a middleman struct and then convert it to the actual state.

Overall I am not sure if using the enum is a good idea considering the future plans and the already existing methods. It definitely made sense when it was suggested, but now it causes multiple breaking changes, requires us to workaround the serialization and deserialization, and possibly diverges the list and table states which @orhun mentioned he wants to eventually get these states consolidated.

airblast-dev:
Found a possible panic when a column is selected, but the provided widths were less than the column count. Fixed, and added testing for the case. Instead of the panic, it will now not apply styling for that column since the selected column is not rendered to begin with.

joshka:
I'd suggest this should be held until after 0.28.1 (because `ratatui::init()` unlocks the ability to document / simplify a bunch of things here, templates, website etc.)

airblast-dev:
I don't mean to rush the review, but is there anything blocking this? Should I replace the `selected_column`, and `selected` fields with an enum? I just wanted to check if there's any design changes to be made on my end.

orhun:
This looks ready for `0.29.0` to me :)

matthewmturner:
Will this enable scrolling columns as well if they are wider than the bounding container?

orhun (approved):

orhun (approved):

orhun:
Updated breaking changes doc, LGTM. Thanks for the contribution!

PR #1333, opened 2024-08-21
feat(color): add hsluv support

### LINKED ISSUE #763: Support hsluv color constructor

du-ob:
closes #763


src/style/color/hsluv.rs
```diff
@@ -0,0 +1,110 @@
...
+        REF_Y * ((l + 16.0) / 116.0).powf(3.0)
+    } else {
+        REF_Y * l / KAPPA
+    };
+    let var_u = u / (13.0 * l) + REF_U;
+    let var_v = v / (13.0 * l) + REF_V;
+
+    let y = var_y * REF_Y;
+    let x = 0.0 - (9.0 * y * var_u) / ((var_u - 4.0) * var_v - var_u * var_v);
+    let z = (9.0 * y - (15.0 * var_v * y) - (var_v * x)) / (3.0 * var_v);
```
orhun, line 48: Is this part panic free? Also, it would be nice to explain these little sections via inner comments for people who are not familiar with the conversion.

src/style/color/hsluv.rs
```diff
@@ -0,0 +1,110 @@
...
+
+    // Convert XYZ to RGB
+    let rgb: Vec<f64> = M
+        .iter()
+        .map(|i| {
+            let dot_product: f64 = i.iter().zip([x, y, z].iter()).map(|(i, j)| i * j).sum();
+            if dot_product <= 0.0031308 {
+                12.92 * dot_product
+            } else {
+                1.055 * (dot_product.powf(1.0 / 2.4)) - 0.055
```
orhun, line 58: What are these constants?

src/style/color/hsluv.rs
```diff
@@ -0,0 +1,110 @@
...
+            }
+        })
+        .collect();
+
+    (rgb[0], rgb[1], rgb[2])
+}
+
+fn max_chroma_for(l: f64, h: f64) -> f64 {
```
orhun, line 66: Would be nice to add unit tests/documentation for these helper functions as well.

orhun (commented):
Overall LGTM, there are some parts that can be improved.

src/style/color/hsluv.rs
du-ob: these are the constants from the transformation formula https://en.wikipedia.org/wiki/SRGB#Specification_of_the_transformation

src/style/color/hsluv.rs
du-ob: yes, all the variables are `f64` and there won't be any division by 0

du-ob: I will add some comments

joshka:
I appreciate the work on this PR, but I think this is the wrong sort of code to include in Ratatui. It's dense, highly specific and belongs in some other crate written once (even if that's a fairly small one that just has this information). @kdheepak what do you think?

kdheepak:
I like having the hsluv functionality (even opened an issue related to this - https://github.com/ratatui/ratatui/issues/763) but would prefer just using `palette` to do the calculations behind a feature flag. Mainly because I don't understand the math here.

joshka:
> I like having the hsluv functionality (even opened an issue related to this - #763) but would prefer just using `palette` to do the calculations behind a feature flag. Mainly because I don't understand the math here.

Yep - I'm ok with that as the idea. It's the implementation / math I don't want. Not the feature itself.

src/style/color.rs
```diff
@@ -390,6 +390,41 @@ impl Color {
...
+    /// assert_eq!(color, Color::Rgb(223, 171, 181));
+    ///
+    /// let color: Color = Color::from_hsluv(0.0, 0.0, 0.0);
+    /// assert_eq!(color, Color::Rgb(0, 0, 0));
+    /// ```
+    #[cfg(feature = "palette")]
+    pub fn from_hsluv(h: f64, s: f64, l: f64) -> Self {
+        use palette::{Clamp, FromColor, Hsluv, Srgb};
+
+        let hsluv = Hsluv::new(h, s, l).clamp();
```
joshka, line 418: Hsluv supports more than just f64. It defaults to f32, and supports other numeric types like u8 for instance. It also supports a whitepoint. I wonder whether we should consider providing methods to convert from palette types instead of handling specific color spaces.

joshka (approved):
This PR is problematic because it doesn't have a real use case that someone has specifically asked for (the issue is more hypothetical that this is needed rather than specific). This makes it difficult to anticipate where the line between generic implementation vs specific is.
- Should whitepoint be supported, should u8 or f32 values instead of f64 be supported? - Should we instead provide conversions from palette types?

Not a blocker per-say (this implements what it implements correctly), so marking as approved, but I'd like to see the questions above considered before merging. I'm certainly not saying we should implement those features, just that we should mindfully notice them if we're not doing so.

orhun (approved):
I think this should be good to go as-is, we can do the improvements in a follow-up PR :)

kdheepak:
Doesn't this also need an update to `Cargo.toml`?

orhun:
Yeah, good point. @du-ob are you still interested in working on this or should we push it over the finish line?

du-ob:
> Yeah, good point. @du-ob are you still interested in working on this or should we push it over the finish line?

Yes I'm still interested in working on this but don't know what additional changes need to be done.

orhun:
I just checked and it is already gated behind the `palette` feature flag. So everything should be good. @kdheepak thoughts?

kdheepak:
Oops, my bad. I thought the `palette` feature didn't exist.

src/style/color.rs
orhun: I'd like to get this in for the next release - would it be fine if we discuss in a separate issue?

joshka:
Running a few of the test values through https://www.colorspaceconverter.com/converter/hsluv-to-rgb gives very different numbers, so there's something wrong with this as it stands.

I think it might be better to implement this conversions from palette types instead of as components and to do the same for the hsl conversion.

(I wrote this a few days ago before this was merged, but was sick and too busy to follow up properly - will make an issue out of it)

PR #1334, opened 2024-08-21
chore: rename ratatui-org to ratatui

joshka:
All urls updated to point at https://github.com/ratatui

To update your repository remotes, you can run the following commands:

```shell
git remote set-url origin https://github.com/ratatui/ratatui
```


orhun:
Only 3 PRs/issues away from **#1337**

orhun (approved):

Issue #1321, opened 2024-08-13
Ratatui's Vision


orhun:
## Problem

Ratatui should have a clear vision regarding how we approach problems and incorporate new features. I wanted to create this issue to share common ideas among maintainers and to clarify things for new contributors who would like to learn more about how things work. This will also help with maintenance, as it will serve as a guide for prioritization and hopefully get everyone on the same page. If you think this issue is missing any points or if you disagree with anything, feel free to discuss it below.

## Solution

Like any other open-source project, we should have guiding principles that we follow when determining our next steps.

I would like to draw your attention to the two most important areas we should be focusing on as a TUI library.

### 1. Ease of Use

There is a fine line between allowing customizations and making something difficult enough to create a frustrating development experience. Right now, I feel like we are standing in the middle, which can be further improved by addressing the following usability-related issues in the long term:

- #1287
- #174
- #293
- #1289 (abstraction over the backends + reducing the boilerplate)
- Simplify the example code, make them consistent to make it easier to read / write the relevant bits.

These and other _universally_ useful issues should be prioritized in development.

### 2. Visual Appeal

_What's a TUI library without a shiny UI?_

We want to draw more attention to Ratatui by making it prettier, and I believe this can be achieved by:

- Making the examples look more polished.
- Allowing widget customization, such as in #1168 and #605.
- Drawing inspiration from other TUI libraries (e.g., Textual, Bubbletea).

### Performance improvements

The current state of the library is not focused around zero allocations or generating the fastest LLVM IR. We take the performance into account while adding new features as long as that are aligned with an actual problem in a real application.

We are willing to make meaningful performance improvements if benchmarks show they are valuable. However, for smaller performance improvements, we will evaluate them on a case-by-case basis, considering factors such as maintainability and potential breakage.

### Breaking changes

We may need to consider breaking changes as we explore these improvements, as they could add significant value in the long term. However, making changing an existing API that would impact large amounts of users is not actionable unless it is really useful. 

### What are the next steps?

We should break down these larger goals into smaller, actionable tasks in a roadmap, ideally with some milestones as well, so we know what to do and when to do it next.

nickspurry:
[...]

PR #1336, opened 2024-08-22
refactor: Remove `itertools` from dependencies

mo8it:
Trying to reduce the dependencies of Rustlings. Everyone who uses Ratatui profits :)

### DIFF
--- a/examples/colors.rs
+++ b/examples/colors.rs
@@ -17,7 +17,6 @@
 // and background colors with their names and indexes.
 
 use color_eyre::Result;
-use itertools::Itertools;
 use ratatui::{
     crossterm::event::{self, Event, KeyCode, KeyEventKind},
     layout::{Alignment, Constraint, Layout, Rect},
@@ -107,7 +106,7 @@ fn render_fg_named_colors(frame: &mut Frame, bg: Color, area: Rect) {
                 .split(*area)
                 .to_vec()
         })
-        .collect_vec();
+        .collect::<Vec<_>>();
     for (i, &fg) in NAMED_COLORS.iter().enumerate() {
         let color_name = fg.to_string();
         let paragraph = Paragraph::new(color_name).fg(fg).bg(bg);
@@ -128,7 +127,7 @@ fn render_bg_named_colors(frame: &mut Frame, fg: Color, area: Rect) {
                 .split(*area)
                 .to_vec()
         })
-        .collect_vec();
+        .collect::<Vec<_>>();
     for (i, &bg) in NAMED_COLORS.iter().enumerate() {
         let color_name = bg.to_string();
         let paragraph = Paragraph::new(color_name).fg(fg).bg(bg);
@@ -199,7 +198,7 @@ fn render_indexed_colors(frame: &mut Frame, area: Rect) {
                 .split(area)
                 .to_vec()
         })
-        .collect_vec();
+        .collect::<Vec<_>>();
 
     for i in 16..=231 {
         let color = Color::Indexed(i);
@@ -236,7 +235,7 @@ fn render_indexed_grayscale(frame: &mut Frame, area: Rect) {
             .split(*area)
             .to_vec()
     })
-    .collect_vec();
+    .collect::<Vec<_>>();
 
     for i in 232..=255 {
         let color = Color::Indexed(i);
--- a/src/buffer/buffer.rs
+++ b/src/buffer/buffer.rs
@@ -618,7 +618,6 @@ impl fmt::Debug for Buffer {
 mod tests {
     use std::iter;
 
-    use itertools::Itertools;
     use rstest::{fixture, rstest};
 
     use super::*;
@@ -924,7 +923,7 @@ mod tests {
             .content
             .iter()
             .map(Cell::symbol)
-            .join("");
+            .fold(String::new(), |acc, x| acc + x);
         assert_eq!(actual_contents, expected);
 
         let actual_styles = small_one_line_buffer.content.iter().map(|c| c.fg);
--- a/src/buffer/buffer.rs
+++ b/src/buffer/buffer.rs
@@ -925,20 +925,16 @@ mod tests {
             .iter()
             .map(Cell::symbol)
             .join("");
-        let actual_styles = small_one_line_buffer
-            .content
-            .iter()
-            .map(|c| c.fg)
-            .collect_vec();
+        assert_eq!(actual_contents, expected);
+
+        let actual_styles = small_one_line_buffer.content.iter().map(|c| c.fg);
 
         // set_line only sets the style for non-empty cells (unlike Line::render which sets the
         // style for all cells)
         let expected_styles = iter::repeat(color)
             .take(content.len().min(5))
-            .chain(iter::repeat(Color::default()).take(5_usize.saturating_sub(content.len())))
-            .collect_vec();
-        assert_eq!(actual_contents, expected);
-        assert_eq!(actual_styles, expected_styles);
+            .chain(iter::repeat(Color::default()).take(5_usize.saturating_sub(content.len())));
+        assert!(actual_styles.eq(expected_styles));
     }
 
     #[test]
--- a/src/layout/layout.rs
+++ b/src/layout/layout.rs
@@ -5,7 +5,6 @@ use cassowary::{
     AddConstraintError, Expression, Solver, Variable,
     WeightedRelation::{EQ, GE, LE},
 };
-use itertools::Itertools;
 use lru::LruCache;
 
 use self::strengths::{
@@ -40,6 +39,22 @@ thread_local! {
     static LAYOUT_CACHE: RefCell<Cache> = RefCell::new(Cache::new(
         NonZeroUsize::new(Layout::DEFAULT_CACHE_SIZE).unwrap(),
     ));
+}
+
+// Similar to `itertools::Itertools::tuple_combinations`.
+fn tuple_combinations<T, E>(
+    slice: &[T],
+    mut f: impl FnMut((&T, &T)) -> Result<(), E>,
+) -> Result<(), E> {
+    if !slice.is_empty() {
+        for (ind, left) in slice[..slice.len() - 1].iter().enumerate() {
+            for right in &slice[ind + 1..] {
+                f((left, right))?;
+            }
+        }
+    }
+
+    Ok(())
 }
 
 /// A layout is a set of constraints that can be applied to a given area to split it into smaller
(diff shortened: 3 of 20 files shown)


orhun (commented):
I am on the fence with this. Primarily because:


- From the crates.io page of [itertools](https://crates.io/crates/itertools/0.13.0/dependencies) it seems like it has only one dependency so the ultimate benefit would be getting rid of that along with `itertools` itself. Is it worth it?

- Also, I see that you added `tuple_combinations` function which takes a crucial part in layout constraints/flex. Should we proceed with a custom implementation, or would it be better to use a potentially better-tested existing solution?

joshka:
I'm against removing itertools, as It's a reasonable widely used dependency (253M downloads) which often adds value / allows code to be expressed concisely. It doesn't appear in the public API, and its only downside is compile time. Is there some particular problem or goal that you're trying to achieve by reducing dependency counts?

In particular I'm for using `collect_vec()` over `.collect::<Vec<_>>()` as a small thing.

I see that there's a bunch of code changes in this where `.collect()` or removing the conversion from iter to vec just works that I am for in this change, and there's a bunch of places where there's also subjectively neater code in this PR too.

examples/colors.rs
```diff
@@ -107,7 +106,7 @@ fn render_fg_named_colors(frame: &mut Frame, bg: Color, area: Rect) {
                 .split(*area)
                 .to_vec()
         })
-        .collect_vec();
+        .collect::<Vec<_>>();
     for (i, &fg) in NAMED_COLORS.iter().enumerate() {
         let color_name = fg.to_string();
         let paragraph = Paragraph::new(color_name).fg(fg).bg(bg);
```
joshka, line 112: this block (can't highlight it fully in the review) can more reasonably written as:
```rust
    let vertical = Layout::vertical([Constraint::Length(1); 2]).split(inner);
    let areas = vertical.iter().flat_map(|area| {
        Layout::horizontal([Constraint::Ratio(1, 8); 8])
            .split(*area)
            .to_vec()
    });
    for (bg, area) in zip(NAMED_COLORS, areas) {
        let color_name = bg.to_string();
        let paragraph = Paragraph::new(color_name).fg(fg).bg(bg);
        frame.render_widget(paragraph, area);
    }
```
Which avoids the allocation and itertools better than the `collect::<Vec<_>>()` code.

This is the same with many of the changes to using that (I haven't checked them all).

This is a more complex changes however. Would you be willing to make it?

src/buffer/buffer.rs
```diff
@@ -924,21 +923,17 @@ mod tests {
             .content
             .iter()
             .map(Cell::symbol)
-            .join("");
-        let actual_styles = small_one_line_buffer
-            .content
-            .iter()
-            .map(|c| c.fg)
-            .collect_vec();
+            .fold(String::new(), |acc, x| acc + x);
```
joshka, line 926: I definitely prefer `join("")` here as it expresses the intent rather than the mechanic and is clearer.

src/layout/layout.rs
```diff
@@ -42,6 +41,22 @@ thread_local! {
...
+        for (ind, left) in slice[..slice.len() - 1].iter().enumerate() {
+            for right in &slice[ind + 1..] {
+                f((left, right))?;
+            }
+        }
+    }
+
+    Ok(())
+}
+
```
joshka, line 59: I'd prefer not to have small utils that are defined in itertools like this as it indicates that we have some sort of special reason that we're doing this differently, which is an extra cognitive burden when maintaining the crate.

src/layout/layout.rs
```diff
@@ -837,29 +849,30 @@ fn configure_fill_constraints(
...
-            _ => unreachable!(),
-        };
-        solver.add_constraint(
-            (right_scaling_factor * left_element.size())
-                | EQ(GROW)
-                | (left_scaling_factor * right_element.size()),
-        )?;
-    }
-    Ok(())
+    tuple_combinations(
```
joshka, line 852: Often I prefer the functional expressions, but I think the for loop here is much better as it the functional approach seems more difficult to reason about.

src/style/stylize.rs
```diff
@@ -345,8 +343,8 @@ mod tests {
         // format!() is used to create a temporary String inside a closure, which suffers the same
         // issue as above without the `Styled` trait impl for `String`
         let items = [String::from("a"), String::from("b")];
-        let sss = items.iter().map(|s| format!("{s}{s}").red()).collect_vec();
-        assert_eq!(sss, vec![Span::from("aa").red(), Span::from("bb").red()]);
+        let sss = items.iter().map(|s| format!("{s}{s}").red());
+        assert!(sss.eq([Span::from("aa").red(), Span::from("bb").red()]));
```
joshka, line 347: Changing this changes the failure mode of the tests from returning the diff between the actual and expected to just returning true / false. That makes it more difficult to diagnose any test failures.

src/text/text.rs
```diff
@@ -632,12 +630,11 @@ impl<T: fmt::Display> ToText for T {
...
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        for (position, line) in self.iter().with_position() {
-            if position == Position::Last || position == Position::Only {
-                write!(f, "{line}")?;
-            } else {
+        if let Some((last, rest)) = self.lines.split_last() {
+            for line in rest {
                 writeln!(f, "{line}")?;
             }
+            last.fmt(f)?;
```
joshka, line 637: I'm curious about what the difference is here over would be compared to calling `write!(f, {last})?;` I suspect that would be clearer.

joshka (commented):
I like the parts in this PR that simplify the code. I really dislike seeing the `collect::<Vec<_>>()` calls, and would suggest removing those in favor of keeping the iterator where possible, using `collect()` where not possible and falling back to `collect_vec()` when necessary. This is likely a more complex change than anticipated however.

See my concerns about removing itertools mentioned in a comment above. I'd like to keep this in the dependency tree as the crate has useful methods which are easy to use and help write intent revealing code. This is a crate with 250 million downloads, so I think it's reasonable to keep in the project.

I'd seek concurrence with either @kdheepak or @orhun on keeping itertools vs dropping it though rather than this just being my perspective / opinion.

mo8it:
> Is it worth it?

It is not significant. But you should know if it is worth it. You will maintain this code :)

> Also, I see that you added tuple_combinations function which takes a crucial part in layout constraints/flex. Should we proceed with a custom implementation, or would it be better to use a potentially better-tested existing solution?

The function isn't complex and should be very easy to test. The `tuple_combinations` implementation in `itertools` is much more complex because it works for a variable number of tuple items.

> its only downside is compile time

No, it also has runtime overhead. See https://github.com/rust-itertools/itertools/issues/817 where one of the maintainers says "I think itertools adaptors are convenient first, not necessarily the fastest in every situation".

Here is a benchmark for the `join("")` method in comparison to my implementation with `fold`:

```rust
use divan::bench;
use itertools::Itertools;

fn main() {
    divan::main();
}

fn str(int: u16) -> &'static str {
    match int % 10 {
        0 => "0",
        1 => "1",
        2 => "2",
        3 => "3",
        4 => "4",
        5 => "5",
        6 => "6",
        7 => "7",
        8 => "8",
        _ => "9",
    }
}

#[bench]
fn fold() -> String {
    (0..u16::MAX).map(str).fold(String::new(), |acc, x| acc + x)
}

#[bench]
fn join() -> String {
    (0..u16::MAX).map(str).join("")
}
```

`join` from `itertools` is about 4 times slower!

Don't get me wrong. I also think that `itertools` is convenient. This is why I kept it in the dev dependencies. But I didn't find it needed in the non-dev code.

joshka:
> No, it also has runtime overhead. See https://github.com/rust-itertools/itertools/issues/817 where one of the maintainers says "I think itertools adaptors are convenient first, not necessarily the fastest in every situation".

That specific code seems like it would be worth optimizing in itertools if possible, which would benefit the entire ecosystem rather than just Ratatui or your own apps. I don't want to own complex code that solves that sort of thing when that complex code belongs elsewhere.


> join from itertools is about 4 times slower!

Regarding perf. 4x means very little unless it's in a hot path that is measured as being a problem in user focused code, and where there is a target performance characteristic which is not being met. The join() call was in a unit test with a small amount of data, and so has no impact on how Ratatui runs within an app. N is usually small (except when it's not), and measuring and using real world numbers rather than synthetic ones is critical to doing performance work.

Don't get me wrong. I agree Perf is important, but it's only important when it's actually causing a problem. This is something that has come up a few times in Ratatui in the past. I do actually appreciate the well meaning attempts to improve the performance, but I prefer these to apply a pragmatic approach that only fixes what is broken rather than every little thing.

A good article that mostly aligns with my perspective on performance is https://tobeva.com/articles/beyond/. It contrasts the performance at all costs approach with a much more pragmatic approach. Ratatui does indeed sit in a weird place where it's often necessary to do performance related things, but those things are mostly likely going to be algorithmic issues rather than smaller micro things like choosing fold over join..

mo8it:
I will open a new PR only with the changes that you want.

joshka:
Thanks mo8it :)

orhun:
Also, you can check out #1321, where I posted our consensus on performance improvements.

mo8it:
See the PR https://github.com/ratatui/ratatui/pull/1345 for the things that you wanted.

Related to this topic: I removed Ratatui from Rustlings after writing the list/table manually using `crossterm` during the weekend. Don't get me wrong, Ratatui is awesome and I will still use it in my other projects. But I think that it is an overkill for the specific use case in Rustlings.
This migration allowed me to avoid duplication with the code that directly used `crossterm` anyway and allowed me to easily implement links and scroll padding in the table. Plus: It was a lot of fun to learn building a TUI in raw mode from scratch :D

I want be completely honest with you: Contributing to Ratatui wasn't a pleasant experience for me. The CI is too annoying, too many restrictions on PRs, formatting nits and too opinionated vision, especially regarding performance.

Please don't take it personally! You were very kind and lovely! I just wanted to give you feedback to improve the experience for the community (but again, I am not a representative for the community, especially because of my performance bias).

Thank you for maintaining Ratatui. Keep cooking! 🥰

kdheepak:
Thanks for sharing! If you can link your commits or PR for porting away from ratatui to a pure crossterm solution, that’ll be great! I think it’ll be a useful reference for people to have.

mo8it:
It is mainly this PR merge: https://github.com/rust-lang/rustlings/commit/360605e2840d7b8876afac6df5178c068484cd57

PR #1337, opened 2024-08-22
chore(ci): update release strategy

### LINKED ISSUE #1232: Rapid Releases

orhun:
closes #1232 

Now we can trigger point releases by pushing a tag (follow the instructions in `RELEASE.md`). This will create a release with generated changelog.

There is still a lack of automation (e.g. updating `CHANGELOG.md`), but this PR is a good start towards improving that.


RELEASE.md
```diff
@@ -1,5 +1,9 @@
 # Creating a Release
 
+Our release strategy is:
+
+> Release major/minor versions with detailed summaries when necessary, while releasing point/patch versions weekly or as needed without extensive announcements.
```
joshka, line 5: nit: wrap at 100 chars

less nit: Our versioning scheme is 0.x.y where x is the major version and y is the minor, so we don't really have patch versions.

.github/workflows/publish-crate.yml
```diff
@@ -0,0 +1,20 @@
...
+jobs:
+  publish-stable:
+    name: Create a stable release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout the repository
+        uses: actions/checkout@v4
+
+      - name: Publish on crates.io
+        uses: actions-rs/cargo@v1
```
joshka, line 17: This action is deprecated / archived since 2023: https://github.com/actions-rs/cargo. Instead it's probably a good idea to install rust using https://github.com/dtolnay/rust-toolchain and then calling cargo publish from a run step.

.github/workflows/publish-crate.yml
joshka, line 1: This seems like it should be in the same workflow as publish stable. Is there a need to split these up?

joshka (approved):
Approved with nits

PR #1339, opened 2024-08-23
perf: only calculate current symbol width once

joshka:
In https://github.com/ratatui/ratatui/issues/1338, it was established that calling .width() in Buffer::diff() was a hotspot for performance. This was being called twice unnecessarily. Caching the value in a variable leads to a 17% performance improvement on the buffer diff of a 200x50 buffer with random text on my machine (MPB M2 Max).

```
buffer/diff             time:   [100.26 µs 100.69 µs 101.15 µs]
                        change: [-18.007% -17.489% -16.929%] (p = 0.00 < 0.05)
                        Performance has improved.
```

---

Running the component sample under flamegraph
```shell
CARGO_PROFILE_RELEASE_DEBUG=true cargo flamegraph --root
```

Before:
image
zooming in
image
image
(the width calculation is about half of the time spent in the diff)

After:
image
image
image
(the width calculation is now only about a quarter of the time spent in the diff)

---

There's likely another slightly perf gain to be had of a similar amount (20μs) that could be found by caching the symbol widths between each frame rather than calculating the previous symbol width when drawing. This change was the simple low hanging fruit change.

---

Ping @gtema for some testing on their machine.

Recommended testing approach:
1. pull ratatui main
2. point the app you're testing at the path of ratatui (config in the cargo.toml)
3. checkout the commit before the change (0d5f3c091)
4. run `CARGO_PROFILE_RELEASE_DEBUG=true cargo flamegraph --root` in your app
5. open the generated svg in something interactive (I use firefox for this, YMMV)
6. Add the benchmark from this PR in ratatui (`git checkout 81c357688 -- benches/buffer.rs`)
7. `cargo bench buffer/diff` (this will set the baseline)
8. checkout the commit after the change (81c357688 - or whatever the latest commit in this PR is)
9. re-run flamegraph and benchmarks
10. compare / cotrast


orhun (approved):

joshka:
Caching in the cell instead of just in the buffer gives a pretty massive speed up:
```
buffer/diff             time:   [54.856 µs 54.994 µs 55.140 µs]
                        change: [-45.189% -44.918% -44.665%] (p = 0.00 < 0.05)
                        Performance has improved.
```
Buffer::diff is now about 20% of the flamegraph time
image

orhun:
Some jobs in CI are failing and there seems to a lint that needs to be fixed.

joshka:
Hrm, this clippy lint is a bit of a worry:

```
error: a `const` item should never be interior mutable
  --> src/buffer/cell.rs:72:5
   |
72 | /     pub const EMPTY: Self = {
73 | |         Self {
74 | |             symbol: CompactString::const_new(" "),
75 | |             fg: Color::Reset,
...  |
82 | |         }
83 | |     };
   | |______^
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#declare_interior_mutable_const
   = note: `-D clippy::declare-interior-mutable-const` implied by `-D warnings`
   = help: to override `-D warnings` add `#[allow(clippy::declare_interior_mutable_const)]`
```

Seems like we can have either the Cell::EMPTY const or the cache.
Might have to check the usage of Cell::EMPTY and replace it with a Cell::empty() function if it's not widely used.
The alternative is move the cache into the diff function, but I think that ends up more convoluted.

The failing tests are more simple to fix.

joshka:
> The failing tests are more simple to fix.

famous last word :D

gtema:
honestly speaking I am confused: flamegraph and diff function measure shows big change, but in the runtime there is absolutely no influence: component-generated template burns 1 CPU core at 50% with and without the change. On flamegraph it looks like time is being re-distributed, but amount of time on try_draw remains same

joshka:
> honestly speaking I am confused: flamegraph and diff function measure shows big change, but in the runtime there is absolutely no influence: component-generated template burns 1 CPU core at 50% with and without the change. On flamegraph it looks like time is being re-distributed, but amount of time on try_draw remains same

This definitely needs to be properly profiled rather than just flamegraphed.

joshka:
Rebased and re-ran benchmark:

Before:
```
❯ cargo bench buffer/diff
   Compiling ratatui v0.28.1 (/Users/joshka/local/ratatui)
    Finished `bench` profile [optimized] target(s) in 20.36s
     Running benches/main.rs (target/release/deps/main-104385649c38cd56)
buffer/diff             time:   [129.09 µs 129.51 µs 130.01 µs]
Found 14 outliers among 100 measurements (14.00%)
  1 (1.00%) low severe
  11 (11.00%) high mild
  2 (2.00%) high severe
```
After:
```
❯ cargo bench buffer/diff
   Compiling ratatui v0.28.1 (/Users/joshka/local/ratatui)
    Finished `bench` profile [optimized] target(s) in 21.68s
     Running benches/main.rs (target/release/deps/main-104385649c38cd56)
buffer/diff             time:   [54.248 µs 54.441 µs 54.666 µs]
                        change: [-58.147% -57.966% -57.793%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 5 outliers among 100 measurements (5.00%)
  1 (1.00%) low mild
  3 (3.00%) high mild
  1 (1.00%) high severe
```
This is on an Macbook M2 Max. At 60 fps that's (129-55)*60 = 4.4ms speedup per second. So 0.4% gain. This doesn't really hit the threshold of pain without some other testing on lower end hardware that might show that this has some sort of more real measuarable benefit. So no huge priority getting this in.

Issue #584, opened 2023-10-22
Inline Viewport Flickering With High Message Throughput


danny-burrows:
## Description

Hi there, I'm encountering a flickering issue when using the `Viewport::Inline` setting for the `Terminal`. On a high throughput of messages (i.e. I'm calling `terminal.insert_before(...)` with high frequency) there is some flickering behaviour likely caused by its scrolling operation.

## To Reproduce

To reproduce you can change just two values in the `inline.rs` example.

1. Set `NUM_DOWNLOADS` to 1000 or more.
2. In `downloads()` change the distribution to be very small. (I.e. Change first line of the function to something like `let distribution = Uniform::new(0, 3);`)

This will have the effect of sending lots of very small downloads through and (for me) reproduces the behaviour. I could imagine the exact values could need tweaking but any number `<10` for the upper bound on the distribution works for me to reproduce the flickering.

## Expected behavior

Ideally this flickering would be eliminated completely.

## Environment(s)

I've tested in all the environments I could get my hands on:

### Linux Laptop
- OS: Linux
- Terminal Emulator: Gnome Terminal
- Font: N/A (Default on install)
- Crate version: v0.23.0
- Backend: **Crossterm**, **Termion**, and **Termwiz**

### Home PC
- OS: Windows
- Terminal Emulator: Windows Terminal
- Font: Consolas (I think that's the default)
- Crate version: v0.23.0
- Backend: **Crossterm**, and **Termwiz**

### Home PC WSL
- OS: Linux (WSL on Windows)
- Terminal Emulator: Kitty
- Font: Consolas
- Crate version: v0.23.0
- Backend: **Crossterm**, **Termion**, and **Termwiz**

joshka:
I took a brief look at this the other day (having not looked at the inline code ever). The code has a terminal clear and then redraws the stuff you want to insert. Doing that many times a second is going to flicker. The most obvious thing to do would be the same thing you'd want to do for any high frequency screen update - batch it. I'm not sure that there's an obvious approach to doing that inside Ratatui for this particular method unfortunately. (I took a stab at using the BeginSynchronizedUpdates commands in Crossterm without any result.

More detailed the algorithm for this is:
1. clear the viewport area
2. add n lines to the terminal shifting the blank area up that number of lines
3. draw the content starting from the top of that newly shifted location
4. Some other stuff after this (the remaining lines / missing lines naming in the method are a little confusing and I'm not sure that I fully grok it yet).

[...]

PR #1340, opened 2024-08-23
feat!(scrolling-regions): use terminal scrolling regions to stop Terminal::insert_before from flickering

nfachan:
The current implementation of `Terminal::insert_before` causes the viewport to flicker. This is described in #584 .

This PR removes that flickering by using terminal scrolling regions (sometimes called "scroll regions"). A terminal can have its scrolling region set to something other than the whole screen. When a scroll ANSI sequence is sent to the terminal and it has a non-default scrolling region, the terminal will scroll just inside of that region.

We use scrolling regions to implement insert_before. We create a region on the screen above the viewport, scroll that up to make room for the newly inserted lines, and then draw the new lines. We may need to repeat this process depending on how much space there is and how many lines we need to draw.

When the viewport takes up the entire screen, we take a modified approach. We create a scrolling region of just the top line (could be more) of the viewport, then use that to draw the lines we want to output. When we're done, we scroll it up by one line, into the scrollback history, and then redraw the top line from the viewport.

A final edge case is when the viewport hasn't yet reached the bottom of the screen. This case, we set up a different scrolling region, where the top is the top of the viewport, and the bottom is the viewport's bottom plus the number of lines we want to scroll by. We then scroll this region down to open up space above the viewport for drawing the inserted lines.

Regardless of what we do, we need to reset the scrolling region. This PR takes the approach of always resetting the scrolling region after every operation. So the `Backend` gets new `scroll_region_up` and `scroll_region_down` methods instead of `set_scrolling_region`, `scroll_up`, `scroll_down`, and `reset_scrolling_region` methods. We chose that approach for two reasons. First, we don't want Ratatui to have to remember that state and then reset the scrolling region when tearing down. Second, the pre-Windows-10 console code doesn't support scrolling regions per se, but it does support scrolling parts of the console.

This PR:
- Adds a new `scrolling-regions` feature.
- Adds two new `Backend` methods: `scroll_region_up` and `scroll_region_down`.
- Implements those `Backend` methods on all backends in the codebase.
- The crossterm and termion implementations use raw ANSI escape sequences. I'm trying to merge changes into those two projects separately to support these functions.
- Adds code to `Terminal::insert_before` to choose between `insert_before_scrolling_regions` and `insert_before_no_scrolling_regions`. The latter is the old implementation.
- Adds lots of tests to the `TestBackend` to for the scrolling-region-related `Backend` methods.
- Adds versions of terminal tests that show that `insert_before` doesn't clobber the viewport. This is a change in behavior from before.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -139,6 +139,10 @@ macros = []
 ## enables conversions from colors in the [`palette`] crate to [`Color`](crate::style::Color).
 palette = ["dep:palette"]
 
+## Use terminal scrolling regions to make some operations less prone to
+## flickering. (i.e. Terminal::insert_before).
+scrolling-regions = []
+
 ## enables all widgets.
 all-widgets = ["widget-calendar"]
 
--- a/src/backend.rs
+++ b/src/backend.rs
@@ -318,6 +318,25 @@ pub trait Backend {
 
     /// Flush any buffered content to the terminal screen.
     fn flush(&mut self) -> io::Result<()>;
+
+    /// Scroll the specified region of the screen up by the provided amount of lines.
+    ///
+    /// The position of the cursor afterwards is undefined.
+    ///
+    /// If the region includes row 0, then lines scrolled out of the top of region should be put
+    /// into the scrollback. Otherwise, they are just ovewritten. Empty lines are scrolled into the
+    /// bottom of the region.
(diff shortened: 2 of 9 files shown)


nfachan:
I closed because I'm going to merge changes and then reopen.

PR #1341, opened 2024-08-23
feat(scrolling-regions)!: use terminal scrolling regions to stop Terminal::insert_before from flickering

nfachan:
The current implementation of Terminal::insert_before causes the viewport to flicker. This is described in #584 .

This PR removes that flickering by using terminal scrolling regions (sometimes called "scroll regions"). A terminal can have its scrolling region set to something other than the whole screen. When a scroll ANSI sequence is sent to the terminal and it has a non-default scrolling region, the terminal will scroll just inside of that region.

We use scrolling regions to implement insert_before. We create a region on the screen above the viewport, scroll that up to make room for the newly inserted lines, and then draw the new lines. We may need to repeat this process depending on how much space there is and how many lines we need to draw.

When the viewport takes up the entire screen, we take a modified approach. We create a scrolling region of just the top line (could be more) of the viewport, then use that to draw the lines we want to output. When we're done, we scroll it up by one line, into the scrollback history, and then redraw the top line from the viewport.

A final edge case is when the viewport hasn't yet reached the bottom of the screen. This case, we set up a different scrolling region, where the top is the top of the viewport, and the bottom is the viewport's bottom plus the number of lines we want to scroll by. We then scroll this region down to open up space above the viewport for drawing the inserted lines.

Regardless of what we do, we need to reset the scrolling region. This PR takes the approach of always resetting the scrolling region after every operation. So the Backend gets new scroll_region_up and scroll_region_down methods instead of set_scrolling_region, scroll_up, scroll_down, and reset_scrolling_region methods. We chose that approach for two reasons. First, we don't want Ratatui to have to remember that state and then reset the scrolling region when tearing down. Second, the pre-Windows-10 console code doesn't support scrolling regio

This PR:
- Adds a new scrolling-regions feature.
- Adds two new Backend methods: scroll_region_up and scroll_region_down.
- Implements those Backend methods on all backends in the codebase.
- The crossterm and termion implementations use raw ANSI escape sequences. I'm trying to merge changes into those two projects separately to support these functions.
- Adds code to Terminal::insert_before to choose between insert_before_scrolling_regions and insert_before_no_scrolling_regions. The latter is the old implementation.
- Adds lots of tests to the TestBackend to for the scrolling-region-related Backend methods.
- Adds versions of terminal tests that show that insert_before doesn't clobber the viewport. This is a change in behavior from before.


nfachan:
The major concern I had with these changes was whether or not scrolling regions would be supported by all terminals, and if not, what the best way was to address #584 while not breaking `Terminal::insert_after` for those terminals that don't support scrolling regions.

My first attempt at that was to create a `scrolling-regions` feature, so that the feature could be enabled by applications that could guarantee they would only be run on appropriate terminals. I don't imagine this is a long-term solution.

From my research, it seems like anything that is vt100-related supports scrolling regions. Window's implementation of ANSI escape sequences supports them, and all Linux and Mac terminal emulators I've tried do support them. The big issue is pre-windows-10 windows support within crossterm.

As a intermediate solution, we could have both implementations compiled in. We'd start with the scrolling-regions one and fall back to the existing implementation. Thus, the but would be fixed on everything but old windows code. The only risk here is that some terminals may not support these ANSI escape sequences. As I said above, I think that's very unlikely: everything I can find that supports ANSI escape sequences supports scrolling regions.

As a long-term solution, we could implement the appropriate behavior in crossterm for windows, and then always rely on scrolling regions. This should be possible, looking at the windows documentation, but I would need to do some work to get it tested and integrated.

One advantage of the intermediate solution is that it wouldn't break external implementations of `Backend`. I don't know how important it is to avoid this type of breaking change. The long-term solution would require that all `Backend`s support scrolling a region of the screen somehow, since the old code would be removed.

nfachan:
I should also note that building maelstrom with this feature enabled results in a much improved TUI. The flickering that was pretty bad before is completely eliminated.

src/backend/test.rs
```diff
@@ -343,6 +343,76 @@ impl Backend for TestBackend {
     fn flush(&mut self) -> io::Result<()> {
         Ok(())
     }
+
+    #[cfg(feature = "scrolling-regions")]
+    fn scroll_region_up(&mut self, region: std::ops::Range<u16>, scroll_by: u16) -> io::Result<()> {
```
joshka, line 348: This method feels a bit more dense than it actually is because of the quantity of conditionals. Do you think it would help the readability / maintainability to flip the nesting to just a single layer and use early returns? e.g.:
```rust
        if region.start > 0 && cells_to_scroll_by >= cell_region.len() {
            // The scroll amount is large enough to clear the whole region.
            self.buffer.content[cell_region].fill_with(Default::default);
            return Ok(());
        }
```

src/backend/test.rs
```diff
@@ -916,4 +986,786 @@ mod tests {
         let mut backend = TestBackend::new(10, 2);
         backend.flush().unwrap();
     }
+
+    #[test]
+    #[cfg(feature = "scrolling-regions")]
+    fn scroll_region_up_whole_screen_by_0() {
+        let mut backend = TestBackend::new(10, 5);
```
joshka, line 993: Consider putting this in a rstest fixture to remove one line from each of the tests.

src/backend/test.rs
```diff
@@ -916,4 +986,786 @@ mod tests {
...
+        backend.buffer = Buffer::with_lines([
+            "aaaaaaaaaa",
+            "bbbbbbbbbb",
+            "cccccccccc",
+            "dddddddddd",
+            "eeeeeeeeee",
+        ]);
```
joshka, line 1000: I normally avoid DRYing tests too much, but I notice that the visual height of the tests seems to get in the way of what they're testing at a glance. The concepts the tests are testing don't really have much to do with the display, but more to do with the idea that the buffer is unchanged. This might make each of the lines seem better as a constant.

Perhaps something like
```rust
// either
const LINE_1: &str = "aaaaaaaaaa";
...
// or
const LINES: [&str; 5] = [...]
const EMPTY_LINE: ...
```
and then:
```rust
backend.assert_buffer_lines([LINE_1, LINE_2, LINE_3, LINE_4, LINE_5]);
```

Alternatively, considering the line length doesn't matter too much (only that there's some width > 1), perhaps collapse the lines to just "aa" etc. and the assertions may naturally wrap at a single line.


Edit: actually, this might be even better purely as an rstest:

```rust
#[rstest]
#[case::whole_screen_by_0(0..5, 0, &[], &[A, B, C, D, E])
#[case::whole_screen_by_2(0..5, 2, &[A, B], &[C, D, E])
...
fn scroll_up(#[case]line_range: Range<u16> #[case]line_count: u16, #[case]expected_scrollback: &[&str], #[case]expected_buffer: &[&str]) {
    let backend = ...
    let backend.buffer = Buffer::with_lines([A,B,C,D,E);
    backend.scroll_region_up(line_range, line_count).unwrap();
    backend_assert_scrollback(expected_scrollback);
    backend_assert_buffer_lines(expected_buffer);
}
```

src/terminal/terminal.rs
```diff
@@ -569,10 +569,22 @@ where
     where
         F: FnOnce(&mut Buffer),
     {
-        if !matches!(self.viewport, Viewport::Inline(_)) {
-            return Ok(());
+        match self.viewport {
+            #[cfg(feature = "scrolling-regions")]
+            Viewport::Inline(_) => self.insert_before_scrolling_regions(height, draw_fn),
+            #[cfg(not(feature = "scrolling-regions"))]
+            Viewport::Inline(_) => self.insert_before_no_scrolling_regions(height, draw_fn),
```
joshka, line 576: Given these are mutually exclusive function definitions would it make sense to name these both the same name (insert_before_impl) and only put the config on the functions instead of within it?

src/terminal/terminal.rs
```diff
@@ -658,6 +670,73 @@ where
         Ok(())
     }
 
+    /// Implement `Self::insert_before` using scrolling regions.
+    #[cfg(feature = "scrolling-regions")]
+    fn insert_before_scrolling_regions(
```
joshka, line 675: Doc comment could be a little bit more detailed. Ideally enough to explain why this is here and how it is different from the previous method. Or put another way, enough to make it easy not to have to read the method implementation to work out what's happening. Nothing huge, just 1-3 extra sentences.

src/terminal/terminal.rs
```diff
@@ -679,7 +758,33 @@ where
         Ok(remainder)
     }
 
+    /// Draw lines at the given vertical offset, assuming that the lines they are replacing on the
+    /// screen are cleared. The slice of cells must contain enough cells /// for the requested
```
joshka, line 762: small typo ///

joshka (approved):
Generally looks good. All comments are mostly nitpicks focused on what I'd preferentially want to maintain in the future rather than things which are wrong

Only the termwiz comment is truly important.

Marking as approved - would like to get another set of eyes on it before merging.

src/terminal/terminal.rs
nfachan: I originally did that, but then I thought about the doc comment. I'm not sure how conditional compilation works with doc comments. I'll look into it.

joshka: Ah that makes sense - I'm fine with leaving this as-is.

src/backend.rs
```diff
@@ -318,6 +318,25 @@ pub trait Backend {
...
     /// Flush any buffered content to the terminal screen.
     fn flush(&mut self) -> io::Result<()>;
+
+    /// Scroll the specified region of the screen up by the provided amount of lines.
+    ///
+    /// The position of the cursor afterwards is undefined.
+    ///
+    /// If the region includes row 0, then lines scrolled out of the top of region should be put
+    /// into the scrollback. Otherwise, they are just overwritten. Empty lines are scrolled into
+    /// the bottom of the region.
```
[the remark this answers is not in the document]
nfachan: I rewrote these comments. Please let me know what you think about the new ones.

src/backend/crossterm.rs
```diff
@@ -271,6 +271,26 @@ where
...
         self.writer.flush()
     }
+
+    #[cfg(feature = "scrolling-regions")]
+    fn scroll_region_up(&mut self, region: std::ops::Range<u16>, amount: u16) -> io::Result<()> {
+        let start = region.start.saturating_add(1);
+        let end = region.end;
+        queue!(self.writer, Print(format!("\x1B[{start};{end}r")))?;
+        queue!(self.writer, crate::crossterm::terminal::ScrollUp(amount))?;
+        queue!(self.writer, Print("\x1B[r"))?;
```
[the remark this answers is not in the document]
nfachan: Done.

src/backend/termwiz.rs
```diff
@@ -249,6 +249,42 @@ impl Backend for TermwizBackend {
...
+                scroll_count: amount as usize,
+            },
+            Change::ScrollRegionUp {
+                first_row: 0,
+                region_size: rows,
+                scroll_count: 0,
+            },
+        ]);
+        Ok(())
+    }
```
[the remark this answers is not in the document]
nfachan: I added a comment to this.

src/backend/termion.rs
```diff
@@ -231,6 +231,26 @@ where
...
         self.writer.flush()
     }
+
+    #[cfg(feature = "scrolling-regions")]
+    fn scroll_region_up(&mut self, region: std::ops::Range<u16>, amount: u16) -> io::Result<()> {
+        let start = region.start.saturating_add(1);
+        let end = region.end;
+        write!(self.writer, "\x1B[{start};{end}r")?;
+        write!(self.writer, "{}", termion::scroll::Up(amount))?;
+        write!(self.writer, "\x1B[r")?;
```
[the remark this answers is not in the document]
nfachan: Okay, done.

src/terminal/terminal.rs
nfachan: Okay, I updated it.

src/terminal/terminal.rs
nfachan: Fixed.

src/backend/test.rs
nfachan: Okay, I did this. Let me know what you think.

src/backend/test.rs
nfachan: Okay, I rewrote it with early returns. Let me know what you think.

nfachan:
I think I have addressed all of the comments.

I'm not sure what to do about the failing coverage check. It's not running coverage checks with `scrolling-regions` enabled.

nfachan:
I have created two separate, mutually exclusive, PRs for crossterm:
- https://github.com/crossterm-rs/crossterm/pull/918
- https://github.com/crossterm-rs/crossterm/pull/923
I have also created a merge request for termion:
- https://gitlab.redox-os.org/redox-os/termion/-/merge_requests/196

If/when these get merged, the appropriate backends should be updated.

nfachan:
What can we do to get this across the line?

src/backend/test.rs
joshka: I'm almost always a fan of early returns over nesting, so this is much better.

src/backend/test.rs
```diff
@@ -343,6 +365,75 @@ impl Backend for TestBackend {
     fn flush(&mut self) -> io::Result<()> {
         Ok(())
     }
+
+    #[cfg(feature = "scrolling-regions")]
+    fn scroll_region_up(&mut self, region: std::ops::Range<u16>, scroll_by: u16) -> io::Result<()> {
+        let width: usize = self.buffer.area.width.into();
+        let region =
+            region.start.min(self.buffer.area.height)..region.end.min(self.buffer.area.height);
```
joshka, line 373: nit: Keeping this in a single region variable over using a start and end variable seems like it hampers the readability a little.

src/backend/test.rs
```diff
@@ -916,4 +995,81 @@ mod tests {
...
+
+        const A: &str = "aaaa";
+        const B: &str = "bbbb";
+        const C: &str = "cccc";
+        const D: &str = "dddd";
+        const E: &str = "eeee";
+        const S: &str = "    ";
+
+        #[rstest]
+        #[case([A, B, C, D, E], 0..5, 0, [],                    [A, B, C, D, E])]
```
joshka, line 1013: This is a really neat way to show how this works succinctly, and the tests are the ones necessary to really cover all the edge cases well (e.g. 0, some, max, >max, for all the regions of note)

src/terminal/terminal.rs
```diff
@@ -657,6 +669,86 @@ where
         Ok(())
     }
 
+    /// Implement `Self::insert_before` using scrolling regions.
```
joshka, line 672: Note to self, read the code below this point more thoroughly - skimmed instead of reading it deeply.

src/terminal/terminal.rs
```diff
@@ -678,7 +770,33 @@ where
...
+        let (to_draw, remainder) = cells.split_at(width * lines_to_draw as usize);
+        if lines_to_draw > 0 {
+            let area = Rect::new(0, y_offset, width as u16, y_offset + lines_to_draw);
+            let old = Buffer::empty(area);
+            let new = Buffer {
+                area,
+                content: to_draw.to_vec(),
+            };
+            self.backend.draw(old.diff(&new).into_iter())?;
+            self.backend.flush()?;
```
joshka, line 793: Is the flush necessary? What's the impact of doing / not doing this?

joshka (approved):
Looks good generally, mostly nits / a couple of questions.

The key comments are the question about using instability vs having a specific feature name, the parts about not making the commands public, and the notes about comments / examples

The tests are really good (and I'd recommend other reviewers read them first to understand this a bit better).

Apologies for not getting back to this sooner.
I'd like to make sure one of the other maintainers takes a bit of a look at this too before we ship it.

jkelleyrtp:
Hi! Test drove this PR, the flickering is gone :)

That being said, there doesn't seem to be a way with the new `insert_before` to insert a line longer than the terminal width. This is useful to emulate `println!()` above a powerline. Lines shorter will get clipped, and while you can wrap them yourself, that's not exactly the same as the line truly being longer than the terminal width (as is common with log lines such that the emulator can wrap them).

I'm not sure if there's anything that can be done in this PR to enable that, but it's worth calling out.

Even an `insert_lines_before` would be nice such that each line is unbounded in length.

I've implemented a workaround using crossterm directly, but it might be a worthwhile addition to this PR to have this be a supported directly.

joshka:
> I've implemented a workaround using crossterm directly, but it might be a worthwhile addition to this PR to have this be a supported directly.

I don't think it would be part of this PR (mainly because if we tried to solve all the problems all at once we'd be here until the world ends reviewing things). But the idea is worth investigating more certainly. What I'd suggest is cutting another issue to list out the problems you're seeing and the specific functionality that you're trying to achieve. Include the workaround that you're using and how it acts under various circumstances.

src/backend.rs
```diff
@@ -318,6 +318,31 @@ pub trait Backend {
...
+
+    /// Scroll a region of the screen downwards, where a region is a range of rows.
+    ///
+    /// Each row in the region is replaced by the row `amount` rows above it, except the top
+    /// `amount` rows, which are replaced by empty rows. If `amount` is equal to or larger than the
+    /// number of rows in the region, then all rows are replaced with empty rows.
+    ///
+    /// The position of the cursor afterwards is undefined.
+    #[cfg(feature = "scrolling-regions")]
+    fn scroll_region_down(&mut self, region: std::ops::Range<u16>, amount: u16) -> io::Result<()>;
```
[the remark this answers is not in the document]
nfachan: Yes, the answer to your first question is as you suspected: that's just what the terminal does (or doesn't do).

I can add some comments about the related ANSI sequences.

I haven't found any great, in-depth documentation on ANSI sequences. Everything I've found basically gives one or two sentence descriptions.

src/backend/crossterm.rs
```diff
@@ -276,6 +276,32 @@ where
...
         self.writer.flush()
     }
+
+    #[cfg(feature = "scrolling-regions")]
+    fn scroll_region_up(&mut self, region: std::ops::Range<u16>, amount: u16) -> io::Result<()> {
+        queue!(
+            self.writer,
+            ScrollUpInRegion {
+                first_row: region.start,
+                last_row: region.end.saturating_sub(1),
```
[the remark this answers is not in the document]
nfachan: Okay.

src/backend.rs
```diff
@@ -318,6 +318,31 @@ pub trait Backend {
...
+    /// number of rows in the region, then all rows are replaced with empty rows.
+    ///
+    /// If the region includes row 0, then `amount` rows are copied into the bottom of the
+    /// scrollback buffer. These rows are first taken from the old contents of the region, starting
+    /// from the top. If there aren't sufficient rows in the region, then the remainder are empty
+    /// rows.
+    ///
+    /// The position of the cursor afterwards is undefined.
+    #[cfg(feature = "scrolling-regions")]
+    fn scroll_region_up(&mut self, region: std::ops::Range<u16>, amount: u16) -> io::Result<()>;
```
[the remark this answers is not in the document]
nfachan: Sure.

src/backend.rs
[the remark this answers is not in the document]
nfachan: Sure.

src/backend.rs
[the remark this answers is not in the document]
nfachan: Okay.

src/backend/crossterm.rs
```diff
@@ -487,6 +513,94 @@ impl From<ContentStyle> for Style {
...
 
+/// A command that scrolls the terminal screen a given number of rows up in a specific scrolling
+/// region.
+///
+/// # Notes
+///
+/// Commands must be executed/queued for execution otherwise they do nothing.
+#[cfg(feature = "scrolling-regions")]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct ScrollUpInRegion {
```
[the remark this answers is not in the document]
nfachan: Okay on the link. And you're right on the accidental pub.

Cargo.toml
```diff
@@ -137,6 +137,10 @@ macros = []
 ## enables conversions from colors in the [`palette`] crate to [`Color`](crate::style::Color).
 palette = ["dep:palette"]
 
+## Use terminal scrolling regions to make some operations less prone to
+## flickering. (i.e. Terminal::insert_before).
+scrolling-regions = []
+
```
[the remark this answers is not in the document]
nfachan: I'm not sure. I was hoping for some guidance from the maintainers on how you want to deal with these changes.

Currently, this feature doesn't affect the public API of the library, but it does add a requirement on the `Backend` trait. So, I'm not sure there's anything to annotate with the instability macros.

nfachan:
> That being said, there doesn't seem to be a way with the new `insert_before` to insert a line longer than the terminal width. This is useful to emulate `println!()` above a powerline. Lines shorter will get clipped, and while you can wrap them yourself, that's not exactly the same as the line truly being longer than the terminal width (as is common with log lines such that the emulator can wrap them).
> 
> I'm not sure if there's anything that can be done in this PR to enable that, but it's worth calling out.

Yeah, that's an interesting issue with the current API. What would your ideal API look like?

I'm curious how the scrolling would interact with the terminal's line wrapping. You might be better off emulating the wrapping yourself so you know how much to scroll by.

> I've implemented a workaround using crossterm directly, but it might be a worthwhile addition to this PR to have this be a supported directly.

Can you share how your workaround works? I'm curious.

src/backend/crossterm.rs
```diff
@@ -487,6 +513,94 @@ impl From<ContentStyle> for Style {
     }
 }
 
+/// A command that scrolls the terminal screen a given number of rows up in a specific scrolling
+/// region.
+///
+/// # Notes
+///
+/// Commands must be executed/queued for execution otherwise they do nothing.
```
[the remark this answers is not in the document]
nfachan: I don't think so, as it's a struct. I removed the comment. I was just following the style in the rest of the crossterm code.

src/terminal/terminal.rs
nfachan: As I understand it, the output to the terminal uses buffered IO. You need to flush to get the changes to take affect on the terminal immediately.

I guess you could do the flush at the very end of the top-level function if you wanted to minimize flushes. But the code as it exists seems to not worry too much about optimizing (i.e. minimizing) the number of flushes.

src/backend.rs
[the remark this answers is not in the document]
nfachan: I added a comment about this.

src/backend/crossterm.rs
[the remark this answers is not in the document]
nfachan: Done.

src/backend.rs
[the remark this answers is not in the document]
nfachan: I referred the reader to look at the tests for the test backend, as I think it would make the doc comment very large to have enough examples to illustrate all of the edge cases.

src/backend.rs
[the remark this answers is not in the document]
nfachan: See above.

src/backend.rs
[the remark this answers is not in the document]
nfachan: Done.

src/backend/test.rs
nfachan: Okay, I broke it apart.

nfachan:
I addressed all of @joshka 's comments, I think. Please let me know what's next.

I'm not sure why the test `backend::test::tests::buffer_view_with_overwrites` is failing in CI. I can't reproduce it on my end and none of the changes in my PR should affect that functionality.

jkelleyrtp:
> Yeah, that's an interesting issue with the current API. What would your ideal API look like?

Ideally I'd be able to widgets/lines into a buffer with an unbounded (or just really large) width and then let the emulator handle the wrapping of the individual lines.

If you try this today with the current API ratatui simply panics. Ideally we could just use today's API but maybe resize the viewport before we push the widget into it, and `insert_before` would calculate the number of lines it needs to scroll up before inserting the line.

> I'm curious how the scrolling would interact with the terminal's line wrapping. You might be better off emulating the wrapping yourself so you know how much to scroll by.

As far as I can tell, you need to scroll the scroll region up to make enough room for the wrapped line (which could wrap to multiple lines) and then call `println!()` at the top line. The emulator apparently will wrap this properly but then also treat the line as a true line (collapsing the line breaks when the viewport is resized).

> Can you share how your workaround works? I'm curious.

The gist of it is that you call `insert_before` with the number of lines the widget will be after it's wrapped, and then you manually call `println!()` at the top of that free space.

This works with really-long lines and multilines:

Screenshot 2024-09-20 at 1 13 01 PM

I do have a bug where the grapheme calculation is slightly off when the incoming ansi sequences seem to cause the line wrapping calculation to be slightly off.

```rust
    fn drain_logs(
        &mut self,
        terminal: &mut Terminal<CrosstermBackend<io::Stdout>>,
    ) -> io::Result<()> {
        use unicode_segmentation::UnicodeSegmentation;

        let Some(log) = self.pending_logs.pop() else {
            return Ok(());
        };

        // Grab out the size and location of the terminal and its viewport
        let frame_rect = terminal.get_frame().area();
        let term_size = terminal.size().unwrap();

        // Create a paragraph by escaping the contents of the log, which is already ansi escaped
        let mut overflowed_lines = 0;
        for line in log.content.lines() {
            let grapheme_count = line.graphemes(true).count() as u16;
            if grapheme_count > term_size.width {
                // Subtract 1 since we already know this line will count as at least one line
                overflowed_lines += grapheme_count.div_ceil(term_size.width) - 1;
            }
        }

        let byte_count = log.content.len() as u16;
        let paragraph = Paragraph::new(log.content.into_text().unwrap());
        let line_count = paragraph.line_count(term_size.width) as u16;

        // We want to get the escaped ansii string and then by dumping the paragraph as ascii codes (again)
        // This is important because the line_count method on paragraph takes into account the width of these codes
        let mut raw_ansi_buf = AnsiStringBuffer::new(3000.max(byte_count), line_count);
        raw_ansi_buf.render_ref(&paragraph, raw_ansi_buf.buf.area);

        // Calculate how many lines we need to draw by adding the real lines to the wrapped lines
        let lines_to_draw = line_count + overflowed_lines;

        let space_available = term_size.height - self.viewport_current_height() - 1;

        // Rendering this line will eat the frame, so just shortcut a more reliable path
        // Render the new line at the top of the viewport, and then some spaces so that when we call "clear"
        // The lines will have been scrolled up
        if space_available < lines_to_draw {
            crossterm::queue!(
                std::io::stdout(),
                crossterm::cursor::MoveTo(0, frame_rect.y),
                crossterm::style::Print(raw_ansi_buf.to_string().trim_end()),
                crossterm::style::Print("\n"),
                crossterm::style::Print(
                    (0..self.viewport_current_height() - 1)
                        .map(|_| "\n")
                        .collect::<String>()
                ),
            )?;
            terminal.clear()?;
            return Ok(());
        }

        // In the case where the log will fit on the screen, we want to make some room for it
        // by adding some lines above the viewport. `insert_before` will eventually use scroll regions
        // in ratatui, so we're just going to use that, even if it has extra flickering in the interim.
        terminal.insert_before(lines_to_draw, |_| {})?;

        // If the viewport is at the bottom of the screen, our new log will be inserted right above
        // the viewport. If not, the viewport will shift down by lines_to_draw *or* the space available
        let y_offset = match frame_rect.y - 1 < space_available {
            true => 0,
            false => lines_to_draw,
        };

        // Finally, print the log to the terminal using crossterm, not ratatui
        // We are careful to handle the case where the log won't fit on the screen, since that will
        // cause this code to be called with the wrong viewport and cause tearing.
        raw_ansi_buf.trim_end();
        let buf = raw_ansi_buf.to_string();

        let mut max_idx = 0_u16;
        for (idx, line) in buf.lines().enumerate() {
            if line.is_empty() {
                continue;
            }
            let start = frame_rect.y.saturating_sub(y_offset) + idx as u16;
            crossterm::queue!(
                std::io::stdout(),
                crossterm::cursor::MoveTo(0, start),
                crossterm::style::Print(line),
                crossterm::style::Print("\n"),
            )?;
            max_idx = idx as _;
        }

        // Just sanity check that our math works out
        assert_eq!(max_idx - line_count, 0);

        Ok(())
    }
```

nfachan:
Ah, I see what you're doing. Thanks for sharing.

I think you're right about doing this with scrolling regions. Assuming the viewport doesn't take up the whole screen, you'd just set the scrolling region, append a lot of lines at the bottom of the region, then reset the scrolling region. Right?

When the viewport takes up the whole screen, I think the top-line trick would work as well, wouldn't it? The terminal emulator would scroll up the top line when doing wrapping. It seems like something to play with.

nfachan:
To implement your function, I think we'd want to split up the backend functions into four, instead of two:
  - Set scrolling region.
  - Reset scrolling region (to whole screen).
  - Scroll up.
  - Scroll down.
I think that's do-able for all backends except for crossterm for old windows terminals (ones that don't support ANSI sequences). Currently, this crossterm implementation doesn't support old windows terminals either, but it could in theory.

joshka:
> I addressed all of @joshka 's comments, I think. Please let me know what's next.
> 
> I'm not sure why the test `backend::test::tests::buffer_view_with_overwrites` is failing in CI. I can't reproduce it on my end and none of the changes in my PR should affect that functionality.

I also can't repro the test failures locally.  I'll rerun them to see if it was some weird intermittent failure. https://github.com/ratatui/ratatui/actions/runs/10964957433?pr=1341

joshka:
I don't have permission to edit the PR, but here's the patch to fix this:

```patch
diff --git a/Cargo.toml b/Cargo.toml
index 4a8bdac9c..7e4597202 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,7 +38,7 @@ termwiz = { version = "0.22.0", optional = true }
 time = { version = "0.3.11", optional = true, features = ["local-offset"] }
 unicode-segmentation = "1.10"
 unicode-truncate = "1"
-unicode-width = "0.1.13"
+unicode-width = "=0.1.13"
 
 [target.'cfg(not(windows))'.dependencies]
 # termion is not supported on Windows
```

nfachan:
I merged in @joshka's fix from main and that seems to fix it.

orhun (approved):

joshka:
Ok, then we're just working out release timing on this one. I'm not sure whether the next release will be 0.28.2 (without this) or 0.29.0 (with this).

nfachan:
> > Yeah, that's an interesting issue with the current API. What would your ideal API look like?
> 
> Ideally I'd be able to widgets/lines into a buffer with an unbounded (or just really large) width and then let the emulator handle the wrapping of the individual lines.
> 
> If you try this today with the current API ratatui simply panics. Ideally we could just use today's API but maybe resize the viewport before we push the widget into it, and `insert_before` would calculate the number of lines it needs to scroll up before inserting the line.
> 
> > I'm curious how the scrolling would interact with the terminal's line wrapping. You might be better off emulating the wrapping yourself so you know how much to scroll by.
> 
> As far as I can tell, you need to scroll the scroll region up to make enough room for the wrapped line (which could wrap to multiple lines) and then call `println!()` at the top line. The emulator apparently will wrap this properly but then also treat the line as a true line (collapsing the line breaks when the viewport is resized).
> 
> I think with scroll regions you can just simply println!() at the bottom of the region and the region will scroll up, so you wouldn't need to implement the wrapping math yourself, hence why I jumped into this PR.
> 
> > Can you share how your workaround works? I'm curious.
> 
> The gist of it is that you call `insert_before` with the number of lines the widget will be after it's wrapped, and then you manually call `println!()` at the top of that free space.
> 
> This works with long lines and multilines (ignore the panic I inserted...):

I opened #1426.

PR #1342, opened 2024-08-24
docs(table): fix incorrect backticks in `TableState` docs

airblast-dev:
The part of the `TableState` docs that mention the examples, use an incorrect number of backticks.

Old:
Old

New:
New


joshka (commented):
Thanks, this is probably from my VSCode setup - it always inserts the wrong number of backticks when I manually add one at the front of a piece of code and then manually add one at the end. Not sure exactly which setting it is that's causing this, but it's somewhat annoying.

joshka (approved):

PR #1345, opened 2024-08-26
refactor: Avoid unneeded allocations

mo8it:
See https://github.com/ratatui/ratatui/pull/1336

### DIFF
--- a/examples/block.rs
+++ b/examples/block.rs
@@ -14,7 +14,6 @@
 //! [examples readme]: https://github.com/ratatui/ratatui/blob/main/examples/README.md
 
 use color_eyre::Result;
-use itertools::Itertools;
 use ratatui::{
     crossterm::event::{self, Event, KeyCode, KeyEventKind},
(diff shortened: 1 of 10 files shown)


mo8it:
I have no idea why the CI is failing.

joshka (approved):
LGTM - thanks @mo8it

joshka:
> I have no idea why the CI is failing.

No idea either. I thought codecov managed to deal with forked repos properly, but seems like it has some sort of failure mode that isn't particularly obvious. Failing this particular check isn't something that matters greatly as we use this mainly used as encouragement to write tests that cover new features / changes rather than as a gate for whether we'll merge something. We have a reasonably high test coverage ratio (~%94), not because we're chasing a number, but because having the tests is useful. Ratatui's tests help us understand when some behavior is changed by a change. We don't mind changing behavior intentionally within reason, but doing it accidentally is a problem we'd generally like to avoid.

PR #1350, opened 2024-09-01
chore(ci): bump git-cliff-action to v4

orhun:
See: https://github.com/orhun/git-cliff-action/releases/tag/v4.0.0

This also makes me think if we should track GitHub Actions dependencies via Dependabot as well.

### DIFF
--- a/.github/workflows/release-alpha.yml
+++ b/.github/workflows/release-alpha.yml
@@ -33,7 +33,7 @@ jobs:
         run: cargo publish --allow-dirty --token ${{ secrets.CARGO_TOKEN }}
 
       - name: Generate a changelog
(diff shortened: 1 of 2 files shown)


Valentin271:
> we should track GitHub Actions dependencies via Dependabot

Yes definitely, no one's going to notice otherwise.

Valentin271 (approved):

orhun:
Just realized it is configured already:

https://github.com/ratatui/ratatui/blob/55956829e6a220e3996109adc0c78e5ea7e78ecc/.github/dependabot.yml#L14-L18

🤔

Valentin271:
You released git-cliff the 26, this monday. Maybe it would have triggered tomorrow ... Guess we'll never know

PR #1351, opened 2024-09-03
feat: add metrics (wip)

joshka:
Adds a variety of metrics to the terminal and backend modules.

Run the metrics example to see the metrics in action.

```shell
cargo run --example metrics
cargo run --example metrics --release
```


joshka:
image

image

Cargo.toml
```diff
@@ -40,6 +40,8 @@ time = { version = "0.3.11", optional = true, features = ["local-offset"] }
 unicode-segmentation = "1.10"
 unicode-truncate = "1"
 unicode-width = "0.1.13"
+metrics = { version = "0.23.0", git = "https://github.com/joshka/metrics.git", branch = "jm/derive-debug" }
```
joshka, line 43: https://github.com/metrics-rs/metrics/pull/504

src/backend.rs
```diff
@@ -127,6 +129,89 @@ pub use self::termwiz::TermwizBackend;
 mod test;
 pub use self::test::TestBackend;
 
+static METRICS: LazyLock<Metrics> = LazyLock::new(Metrics::new);
```
joshka, line 132: LazyLock was stabilized in Rust 1.80 which is past our current MSRV, but I'm using that for now and will replace it with lazy_static or similar until we catch up to a version that has LazyLock.

joshka:
It might be worth experimenting with https://crates.io/crates/autometrics to simplify some of this.

PR #1352, opened 2024-09-04
fix: create `unstable-backend-writer` feature

Patryk27:
https://github.com/ratatui/ratatui/pull/991 created a new unstable feature, but forgot to add it to Cargo.toml, making it impossible to use on newer versions of rustc - this commit fixes it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -156,7 +156,11 @@ underline-color = ["dep:crossterm"]
 #! The following features are unstable and may change in the future:
 
 ## Enable all unstable features.
-unstable = ["unstable-rendered-line-info", "unstable-widget-ref"]
+unstable = [
+  "unstable-rendered-line-info",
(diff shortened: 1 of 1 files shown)


orhun:
shouldn't the feature called "backend-writer" instead?

Patryk27:
No, the unstable- prefix seems to be added implicitly by rustc; I was pretty surprised as well.

orhun:
hmm, strange. Is there any documentation about this?

Maybe it is done by the `stability` crate?

Patryk27:
Ah, right - it's the stability crate:

https://docs.rs/instability/latest/instability/attr.unstable.html

joshka:
Yeah instability is our fork of stability. We needed a feature that took a while to get into stability (attributes on re-exports)

orhun (approved):

joshka (approved):

PR #1353, opened 2024-09-04
fix: allow to resize fixed viewports

Patryk27:
Calling `Terminal::resize()` on a fixed viewport silently does nothing, because the inner `area` variable shadows the `area` parameter specified by user - this feels like a bug, because resizing fixed frames comes handy and I think it should be allowed.

### DIFF
--- a/src/terminal/terminal.rs
+++ b/src/terminal/terminal.rs
@@ -219,7 +219,7 @@ where
                 )?
                 .0
             }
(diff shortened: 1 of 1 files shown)


kdheepak (approved):

kdheepak:
That does seem like a bug, thanks for the PR!

orhun (approved):

joshka (approved):

PR #1356, opened 2024-09-07
fix: clippy lints from rust 1.81.0

fujiapple852:

### DIFF
--- a/src/widgets/reflow.rs
+++ b/src/widgets/reflow.rs
@@ -51,7 +51,7 @@ where
     O: Iterator<Item = (I, Alignment)>,
     I: Iterator<Item = StyledGrapheme<'a>>,
 {
-    pub fn new(lines: O, max_line_width: u16, trim: bool) -> Self {
+    pub const fn new(lines: O, max_line_width: u16, trim: bool) -> Self {
         Self {
             input_lines: lines,
             max_line_width,
@@ -250,7 +250,7 @@ where
     O: Iterator<Item = (I, Alignment)>,
     I: Iterator<Item = StyledGrapheme<'a>>,
 {
-    pub fn new(lines: O, max_line_width: u16) -> Self {
+    pub const fn new(lines: O, max_line_width: u16) -> Self {
         Self {
             input_lines: lines,
             max_line_width,
--- a/src/widgets/sparkline.rs
+++ b/src/widgets/sparkline.rs
@@ -237,7 +237,7 @@ mod tests {
 
     // Helper function to render a sparkline to a buffer with a given width
     // filled with x symbols to make it easier to assert on the result
-    fn render(widget: Sparkline, width: u16) -> Buffer {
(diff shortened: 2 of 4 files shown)


src/widgets/reflow.rs
```diff
@@ -51,7 +51,7 @@ where
     O: Iterator<Item = (I, Alignment)>,
     I: Iterator<Item = StyledGrapheme<'a>>,
 {
-    pub fn new(lines: O, max_line_width: u16, trim: bool) -> Self {
+    pub const fn new(lines: O, max_line_width: u16, trim: bool) -> Self {
```
fujiapple852, line 54: Consider ignoring the `clippy::missing-const-for-fn` lint at the top level, as once a function is const, a changes future to the _implementation_ of a function can impact the ability for it to be const, and that is _technically_ a semver break.

src/widgets/sparkline.rs
```diff
@@ -253,6 +253,7 @@ mod tests {
 
     #[test]
     fn it_does_not_panic_if_max_is_set_to_zero() {
+        #[expect(clippy::unnecessary_min_or_max)]
```
fujiapple852, line 256: This is a known clippy [bug](https://github.com/rust-lang/rust-clippy/issues/13191).  Rather than `ignore` it and link to the issue, I used the new `expect` so that it will fail as soon as the bug is resolved and the lint is no longer needed.

joshka, line 256: expect is 1.8.1, which would raise out MSRV too high for now. I'd like to keep this at least 2 versions back to give people a little time to upgrade (this aligns with the policy of the `time` crate)
```suggestion
        #[allow(clippy::unnecessary_min_or_max)]
```
Also, I think this might be a slightly different form of https://github.com/rust-lang/rust-clippy/issues/13191

src/widgets/sparkline.rs
```diff
@@ -237,7 +237,7 @@ mod tests {
 
     // Helper function to render a sparkline to a buffer with a given width
     // filled with x symbols to make it easier to assert on the result
-    fn render(widget: Sparkline, width: u16) -> Buffer {
+    fn render(widget: Sparkline<'_>, width: u16) -> Buffer {
```
joshka, line 240: I'm curious what this one was...

joshka (approved):
One small fix and a question

src/widgets/reflow.rs
joshka: This module is private, so I'm fine with this change.

src/widgets/sparkline.rs
joshka: Posted as you did I think.

src/widgets/sparkline.rs
joshka: Ah got it. I'm fine with the change. Thanks.

joshka:
Thanks again :)

PR #1358, opened 2024-09-07
feat: add an impl of `DoubleEndedIterator` for `Columns` and `Rows` (#1357)

### LINKED ISSUE #1357: Add an impl of `DoubleEndedIterator` for `Columns` and `Rows`
Add an impl of `DoubleEndedIterator` for `Columns` and `Rows`.

fujiapple852:
Closes #1357

### DIFF
--- a/src/layout/rect/iter.rs
+++ b/src/layout/rect/iter.rs
@@ -3,9 +3,11 @@ use crate::layout::{Position, Rect};
 /// An iterator over rows within a `Rect`.
 pub struct Rows {
     /// The `Rect` associated with the rows.
-    pub rect: Rect,
-    /// The y coordinate of the row within the `Rect`.
-    pub current_row: u16,
+    rect: Rect,
+    /// The y coordinate of the row within the `Rect` when iterating forwards.
+    current_row_fwd: u16,
+    /// The y coordinate of the row within the `Rect` when iterating backwards.
+    current_row_back: u16,
 }
 
 impl Rows {
@@ -13,7 +15,8 @@ impl Rows {
     pub const fn new(rect: Rect) -> Self {
         Self {
             rect,
-            current_row: rect.y,
+            current_row_fwd: rect.y,
+            current_row_back: rect.bottom(),
         }
     }
 }
@@ -25,11 +28,25 @@ impl Iterator for Rows {
     ///
     /// Returns `None` when there are no more rows to iterate through.
     fn next(&mut self) -> Option<Self::Item> {
-        if self.current_row >= self.rect.bottom() {
+        if self.current_row_fwd >= self.current_row_back {
             return None;
         }
-        let row = Rect::new(self.rect.x, self.current_row, self.rect.width, 1);
-        self.current_row += 1;
+        let row = Rect::new(self.rect.x, self.current_row_fwd, self.rect.width, 1);
+        self.current_row_fwd += 1;
+        Some(row)
+    }
+}
+
+impl DoubleEndedIterator for Rows {
+    /// Retrieves the previous row within the `Rect`.
+    ///
+    /// Returns `None` when there are no more rows to iterate through.
+    fn next_back(&mut self) -> Option<Self::Item> {
+        if self.current_row_back <= self.current_row_fwd {
+            return None;
+        }
+        self.current_row_back -= 1;
+        let row = Rect::new(self.rect.x, self.current_row_back, self.rect.width, 1);
         Some(row)
     }
 }
@@ -37,9 +54,11 @@ impl Iterator for Rows {
 /// An iterator over columns within a `Rect`.
 pub struct Columns {
     /// The `Rect` associated with the columns.
-    pub rect: Rect,
-    /// The x coordinate of the column within the `Rect`.
-    pub current_column: u16,
+    rect: Rect,
+    /// The x coordinate of the column within the `Rect` when iterating forwards.
+    current_column_fwd: u16,
+    /// The x coordinate of the column within the `Rect` when iterating backwards.
+    current_column_back: u16,
 }
 
 impl Columns {
@@ -47,7 +66,8 @@ impl Columns {
     pub const fn new(rect: Rect) -> Self {
         Self {
             rect,
-            current_column: rect.x,
+            current_column_fwd: rect.x,
+            current_column_back: rect.right(),
         }
     }
 }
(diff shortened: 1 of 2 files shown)


src/layout/rect/iter.rs
```diff
@@ -38,8 +38,10 @@ impl Iterator for Rows {
 pub struct Columns {
     /// The `Rect` associated with the columns.
     pub rect: Rect,
-    /// The x coordinate of the column within the `Rect`.
+    /// The x coordinate of the column within the `Rect` when iterating forwards.
     pub current_column: u16,
+    /// The x coordinate of the column within the `Rect` when iterating backwards.
+    current_column_back: u16,
```
fujiapple852, line 44: I chose not to make `current_column_back` pub, as I don' think it needs to be.  Given these are not exposed externally I would suggest renaming `current_column` as `current_column_fwd` and making it non-pub.

joshka: I don't recall why I made these pub in the first place - very likely by mistake rather than intention.
Changing this would be a breaking change, but one that very likely would be not seen by anyone.
I'd remove the pub altogether and slap a couple of lines in the docs.

src/layout/rect/iter.rs
```diff
@@ -48,6 +50,7 @@ impl Columns {
         Self {
             rect,
             current_column: rect.x,
+            current_column_back: rect.x + rect.width,
```
joshka, line 53:
```suggestion
            current_column_back: rect.right(),
```

That said, I found it a little confusing that the current column is the next column to return, but the back column is the the next_back + 1 column. Does it make sense here to -1 on this and then use > instead of >= in the checks below?

e.g. for a 3 column area, this will currently be:
current_column = 0
current_column_back = 3 (should be 2)

src/layout/rect/iter.rs
```diff
@@ -59,7 +62,9 @@ impl Iterator for Columns {
     ///
     /// Returns `None` when there are no more columns to iterate through.
     fn next(&mut self) -> Option<Self::Item> {
-        if self.current_column >= self.rect.right() {
+        if self.current_column >= self.rect.right()
+            || self.current_column >= self.current_column_back
```
joshka, line 66: given that current column back is initialized to right(), isn't the first check superfluos?
```suggestion
        if self.current_column >= self.current_column_back
```
(and change to `>` if the above minus 1 change is made

src/layout/rect/iter.rs
```diff
@@ -68,6 +73,26 @@ impl Iterator for Columns {
     }
 }
 
+impl DoubleEndedIterator for Columns {
+    /// Retrieves the previous column within the `Rect`.
+    ///
+    /// Returns `None` when there are no more columns to iterate through.
+    fn next_back(&mut self) -> Option<Self::Item> {
+        if self.current_column_back <= self.rect.x
+            || self.current_column_back <= self.current_column
```
joshka, line 82:
```suggestion
        if self.current_column_back <= self.current_column
```

(and change to `<` if needed)

joshka (approved):
LGTM generally. 
- We should break the use of the pub fields for 0.29.0
- Some small ergonomics of the code for symmetry / better understanding
- This PR should include rows() too

src/layout/rect/iter.rs
fujiapple852: Done and added a section to the `BREAKING-CHANGES.md` document:

```
### Removed public field `[current_column|current_row]` from `layout::rect::[Columns|Rows]` ([#1358])

[#1358]: https://github.com/ratatui/ratatui/pull/1358

The `pub` modifier has been removed from fields `layout::rect::Columns::current_column` and 
`layout::rect::Rows::current_row`. These fields were not intended to be public and should not
have been accessed directly.
```

src/layout/rect/iter.rs
fujiapple852: Changed to use `rect.right()`.

Things seem to work out badly when `current_column_back` points to the (and not past the) final index, mainly with the terminate case for `next_back()`.  I note that `right()` is defined as "the first coordinate outside of the `Rect`" which makes sense to me and seems consistent with this.  I tried to get it to work anyway but couldn't, maybe i'm just being stupid, feel free to have a fiddle.

src/layout/rect/iter.rs
fujiapple852: Good catch, removed.

src/layout/rect/iter.rs
fujiapple852: Likewise.

fujiapple852:
Added `DoubleEndedIterator` for `Rows` as well, and added tests.

src/layout/rect/iter.rs
joshka: Ah, you're right. The thing I was missing was that the indexes need to record when the iterators cross explicitly. A good simple case to test this is having a single item. If the indexes are both 0, then there's nothing to disambiguate whether the iterator has been exhausted already or if the iterator is in its initial state. Having them as 0, 1 makes this possible.

I wonder if there's problems around u16::MAX for this.

src/layout/rect/iter.rs
```diff
@@ -114,20 +148,70 @@ mod tests {
...
+    #[test]
+    fn rows_back() {
+        let rect = Rect::new(0, 0, 2, 3);
+        let mut rows = Rows::new(rect).rev();
+        assert_eq!(rows.next(), Some(Rect::new(0, 2, 2, 1)));
+        assert_eq!(rows.next(), Some(Rect::new(0, 1, 2, 1)));
+        assert_eq!(rows.next(), Some(Rect::new(0, 0, 2, 1)));
+        assert_eq!(rows.next(), None);
+        assert_eq!(rows.next_back(), None);
+    }
```
joshka, line 169: I think this would be better to just test the iterator and not the RevIter wrapped here. Can this be:
```suggestion
    #[test]
    fn rows_back() {
        let rect = Rect::new(0, 0, 2, 3);
        let mut rows = Rows::new(rect);
        assert_eq!(rows.next_back(), Some(Rect::new(0, 2, 2, 1)));
        assert_eq!(rows.next_back(), Some(Rect::new(0, 1, 2, 1)));
        assert_eq!(rows.next_back(), Some(Rect::new(0, 0, 2, 1)));
        assert_eq!(rows.next_back(), None);
        assert_eq!(rows.next, None);
    }
```
And same for cols?

joshka (approved):
LGTM (repeating for other maintainers), hold merging this closer to 0.29 - I'd guess this might be sometime in ~2-4 weeks.

src/layout/rect/iter.rs
fujiapple852: > I wonder if there's problems around u16::MAX for this.

I added a couple of boundary condition tests and I think it is correct.

```rust
/// We allow a total of `65536` columns in the range `(0..=65535)`.  In this test we iterate
/// forward and skip the first `65534` columns, and expect the next column to be `65535` and
/// the subsequent columns to be `None`.
#[test]
fn columns_max() {
    let rect = Rect::new(0, 0, u16::MAX, 1);
    let mut columns = Columns::new(rect).skip(usize::from(u16::MAX - 1));
    assert_eq!(columns.next(), Some(Rect::new(u16::MAX - 1, 0, 1, 1)));
    assert_eq!(columns.next(), None);
}

/// We allow a total of `65536` columns in the range `(0..=65535)`.  In this test we iterate
/// backward and skip the last `65534` columns, and expect the next column to be `0` and the
/// subsequent columns to be `None`.
#[test]
fn columns_min() {
    let rect = Rect::new(0, 0, u16::MAX, 1);
    let mut columns = Columns::new(rect).rev().skip(usize::from(u16::MAX - 1));
    assert_eq!(columns.next(), Some(Rect::new(0, 0, 1, 1)));
    assert_eq!(columns.next(), None);
}
```

joshka: thanks :)
Usually these things are not a huge problem, but someone somewhere is likely to try to break some random tui with this sort of thing.

orhun (approved):

joshka (approved):

src/layout/rect/iter.rs
joshka: I removed the pub on the rect fields too.

PR #1359, opened 2024-09-08
feat(nix-support): add basic .nix files to project root.

pipelight:
Nix and Nixos btw cannot find statically linked libraries, which is why a `cargo build` results in the following error `collect2: fatal error: cannot find 'ld'`.

To ease development under nixos one usually
adds flake.nix and shell.nix files to the repo root.

Then we are able to generate the appropiate developing environment by typing the following command:

```sh
nix develop
```

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,39 @@
+{
+  description = "Ratatui - A Rust crate for cooking up Terminal User Interfaces";
+
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
+    rust-overlay.url = "github:oxalica/rust-overlay";
+    flake-utils.url = "github:numtide/flake-utils";
+    flake-parts.url = "github:hercules-ci/flake-parts";
+  };
+
+  outputs = {
+    self,
+    nixpkgs,
+    rust-overlay,
+    flake-utils,
+    flake-parts,
+  } @ inputs:
+    flake-parts.lib.mkFlake {
+      inherit inputs;
+    } {
+      systems =
+        flake-utils.lib.allSystems;
+      perSystem = {
+        config,
+        self,
+        inputs,
+        pkgs,
+        system,
+        ...
+      }: let
+        overlays = [(import rust-overlay)];
+        pkgs = import nixpkgs {
+          inherit system overlays;
+        };
+      in {
+        devShells.default = pkgs.callPackage ./shell.nix {};
+      };
+    };
+}
--- a/shell.nix
+++ b/shell.nix
@@ -1,0 +1,13 @@
+{pkgs ? import <nixpkgs> {}, ...}:
+pkgs.mkShell {
+  buildInputs = with pkgs.buildPackages; [
+    pkg-config
+    rust-bin.stable.latest.default
+    llvmPackages_latest.bintools
+  ];
+
+  RUSTFLAGS = builtins.map (a: ''-L ${a}/lib'') [
+    pkgs.llvmPackages_latest.bintools
+  ];
+  RUSTC_VERSION = pkgs.lib.readFile ./rust-toolchain.toml;
+}
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,6 @@ Cargo.lock
 *.rs.rustfmt
 .gdb_history
 .idea/
+
+# Nix related artifacts
+flake.lock


orhun:
Are their other Rust libraries doing this? I'm not sure if I fully understood the reason behind this change.

a-kenji:
As a `Nix` user I also don't understand the reason behind this flake.
`Ratatui` itself doesn't have native libraries it needs to link to.

> To ease development under nixos one usually
adds flake.nix and shell.nix files to the repo root.

Devshells are usable by anyone that has `Nix` installed on any Linux, or MacOS system.
So Anyones development might be improved by adding a common way to refer to specific versions of development tools.
I see no reason this really improves anything here.
You can get `rustc` and `cargo` from `nixpkgs` through:
```
nix shell nixpkgs#cargo nixpkgs#rustc --command $SHELL
```
And if upstream doesn't find much value, then maintaining these files becomes a burden.


> Are there other Rust libraries doing this? 

[I do it](https://github.com/a-kenji/tui-term). I find quite the value in it. I build my examples on the rust version that is referenced inside the `rust-toolchain.toml`, I build the examples with the version referenced inside the `MSRV` file. I can be on any computer that has `Nix` and can always get the correct tools, contributors only need *one* single dependency. I run `Nix` in ci and can run the ci checks locally just the same.

flake.nix
```diff
@@ -0,0 +1,39 @@
+{
+  description = "Ratatui - A Rust crate for cooking up Terminal User Interfaces";
+
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
```
joshka, line 5: Add a comment about why using unstable

flake.nix
```diff
@@ -0,0 +1,39 @@
...
+    rust-overlay.url = "github:oxalica/rust-overlay";
```
joshka, line 6: Add a comment about why this overlay is necessary. What problem does this solve? Why not fenix or the mozilla overlay?

flake.nix
```diff
@@ -0,0 +1,39 @@
...
+    flake-utils.url = "github:numtide/flake-utils";
+    flake-parts.url = "github:hercules-ci/flake-parts";
```
joshka, line 8: What problems do these solve?

flake.nix
joshka, line 1: What's the purpose of having a flake?

shell.nix
```diff
@@ -0,0 +1,13 @@
+{pkgs ? import <nixpkgs> {}, ...}:
+pkgs.mkShell {
+  buildInputs = with pkgs.buildPackages; [
+    pkg-config
+    rust-bin.stable.latest.default
+    llvmPackages_latest.bintools
```
joshka, line 6: Why do we need this specifically? Is latest the right approach here? What does rust generally install when installed using rustup?

shell.nix
```diff
@@ -0,0 +1,13 @@
...
+  ];
+
+  RUSTFLAGS = builtins.map (a: ''-L ${a}/lib'') [
+    pkgs.llvmPackages_latest.bintools
+  ];
+  RUSTC_VERSION = pkgs.lib.readFile ./rust-toolchain.toml;
```
joshka, line 12: It's kinda weird that above the bin is stable, but this flag uses the toolchain file. How do these interact.

joshka (commented):
My concern is as a non-nix user I don't understand the design choices here. None of the choices are documented. Is there some standard Rust library setup docs on a source that is nix project related that can be pointed to that would help this?

In the absence of that, the rationale / intent of pretty much every line needs to be explained in a way that makes it easier to judge whether this is right for the users of this. I don't want to be an arbiter of truth when one user wants one way of doing things and another wants a different way.

I want to be clear that I'm not against adding nix related stuff, but I don't necessarily understand it well enough to know what problem is being solved, how that is being solve, whether the solution has downsides or is correct, etc. Because of this, documenting the solution in a way that goes beyond the norm for what you'd need if nix was more common is something I'd suggest.

pipelight:
This pull request is a mistake on my behalf.

I was missing the local libraries needed to build ratatui (rustix deps) due to a bad nixos configuration, and thought it was something with ratatui.

Indeed a toolchain file is more than enough, no need to use flakes here.

Sorry @joshka @orhun @a-kenji for the precious time you lost reviewing this PR.

orhun:
No worries :)

joshka:
All good :)

PR #1361, opened 2024-09-08
docs: frame.size() is deprecated

hosseinnedaee:
`size` is deprecated, use `.area()` as it's the more correct name.


joshka (approved):
LGTM - Thanks for the PR

PR #1362, opened 2024-09-09
refactor: remove unused dependencies; leverage cargo-machete on CI

Veetaha:
`cargo-machete` lints for unused dependencies. See https://github.com/bnjbvr/cargo-machete

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,6 +46,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+      - uses: Swatinem/rust-cache@v2
       - uses: bnjbvr/cargo-machete@v0.6.2
 
   clippy:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,6 +42,13 @@ jobs:
       - uses: actions/checkout@v4
       - uses: EmbarkStudios/cargo-deny-action@v2
 
+  cargo-machete:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


joshka (approved):

Cargo.toml
```diff
@@ -34,7 +34,6 @@ paste = "1.0.2"
 palette = { version = "0.7.6", optional = true }
 serde = { version = "1", optional = true, features = ["derive"] }
 strum = { version = "0.26", features = ["derive"] }
```
joshka, line 36: The removed strum_macros crate in the line below this was added to ensure that strum correctly picked up the right version of that as there was a conflict with something in itertools a while back. When this was added when strum was only available as 0.26.2, and only the macros package was at 0.26.3, so that dependency was actually necessary. Since then, strum released 0.26.3 which references strum_macros 0.26.3, so the right thing to do here is update the main crate version and delete the macros crate as done below.

```suggestion
strum = { version = "0.26.3", features = ["derive"] }
```

.github/workflows/ci.yml
```diff
@@ -42,6 +42,12 @@ jobs:
       - uses: actions/checkout@v4
       - uses: EmbarkStudios/cargo-deny-action@v2
 
+  cargo-machete:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: bnjbvr/cargo-machete@v0.6.2
```
joshka, line 49: Cache to make this fast.
```suggestion
      - uses: Swatinem/rust-cache@v2
      - uses: bnjbvr/cargo-machete@v0.6.2
```

joshka:
Thanks for this :)

PR #1372, opened 2024-09-16
chore(block): deprecate block::Title

joshka:
`ratatui::widgets::block::Title` is deprecated in favor of using `Line` to represent titles.
This removes an unnecessary layer of wrapping (string -> Span -> Line -> Title).

This struct will be removed in a future release of Ratatui (likely 0.31).
For more information see: <https://github.com/ratatui/ratatui/issues/738>

To update your code:
```rust
Block::new().title(Title::from("foo"));
// becomes any of
Block::new().title("foo");
Block::new().title(Line::from("foo"));

Block::new().title(Title::from("foo").position(Position::TOP));
// becomes any of
Block::new().title_top("foo");
Block::new().title_top(Line::from("foo"));

Block::new().title(Title::from("foo").position(Position::BOTTOM));
// becomes any of
Block::new().title_bottom("foo");
Block::new().title_bottom(Line::from("foo"));
```

### DIFF
--- a/src/widgets/block.rs
+++ b/src/widgets/block.rs
@@ -290,9 +290,9 @@ impl<'a> Block<'a> {
     ///
     /// # Future improvements
     ///
-    /// In a future release of Ratatui (likely 0.31) this method will be changed to accept
-    /// `Into<Line>` instead of `Into<Title>`. This allows us to remove the unnecessary `Title`
-    /// struct and store the position in the block itself. For more information see
+    /// In a future release of Ratatui this method will be changed to accept `Into<Line>` instead of
+    /// `Into<Title>`. This allows us to remove the unnecessary `Title` struct and store the
+    /// position in the block itself. For more information see
     /// <https://github.com/ratatui/ratatui/issues/738>.
     ///
     /// [Block example]: https://github.com/ratatui/ratatui/blob/main/examples/README.md#block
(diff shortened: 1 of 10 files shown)


joshka:
I originally wrote this as just the pre-factoring, but after looking at the impact of the refator (which was fairly small), I decided to combine that into this one PR.

src/widgets/block.rs
```diff
@@ -1050,39 +1050,41 @@ mod tests {
...
+        let block = Block::new().title_top("test");
         assert!(block.has_title_at_position(Position::Top));
         assert!(!block.has_title_at_position(Position::Bottom));
 
-        let block = Block::new().title(Title::from("Test").position(Position::Bottom));
+        let block = Block::new().title_bottom("test");
         assert!(!block.has_title_at_position(Position::Top));
         assert!(block.has_title_at_position(Position::Bottom));
 
+        #[allow(deprecated)] // until Title is removed
```
joshka, line 1061: Deprecations left in intentionally until we actually remove the title struct, which helps catch any bugs in the changes until then.

kdheepak (approved):

src/widgets/block.rs
```diff
@@ -288,13 +288,26 @@ impl<'a> Block<'a> {
     /// - [`Block::title_alignment`]
     /// - [`Block::title_position`]
     ///
+    /// # Future improvements
+    ///
+    /// In a future release of Ratatui (likely 0.31) this method will be changed to accept
```
kdheepak, line 293: Minor: imo we don't need to mention `(likely 0.31)`

joshka: Done in 73918092f

joshka:
I'd like to get @orhun and @TadoTheMiner's thoughts on this before merging.

TadoTheMiner (approved):

orhun (approved):
LGTM

PR #1373, opened 2024-09-16
feat(line)!: impl From<Cow<str>> for Line

joshka:
BREAKING-CHANGES: `Line` now implements `From<Cow<str>`

As this adds an extra conversion, ambiguous infered values may no longer
compile.

```rust
// given:
struct Foo { ... }
impl From<Foo> for String { ... }
impl From<Foo> for Cow<str> { ... }

let foo = Foo { ... };
let line = Line::from(foo); // now fails due to ambiguous type inference
// replace with
let line = Line::from(String::from(foo));
```

Fixes: https://github.com/ratatui/ratatui/issues/1367

### DIFF
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -71,7 +71,7 @@ This is a quick summary of the sections below:
 
 [#1373]: https://github.com/ratatui/ratatui/pull/1373
 
-As this adds an extra conversion, ambiguous inferred values may no longer compile.
+As this adds an extra conversion, ambiguous inferred expressions may no longer compile.
 
 ```rust
 // given:
@@ -81,7 +81,7 @@ impl From<Foo> for Cow<str> { ... }
 
 let foo = Foo { ... };
 let line = Line::from(foo); // now fails due to now ambiguous inferred type
-// replace with
+// replace with e.g.
 let line = Line::from(String::from(foo));
 ```
 
@@ -154,7 +154,9 @@ are also named terminal, and confusion about module exports for newer Rust users
 
 This change simplifies the trait and makes it easier to implement.
 
-### `Frame::size` is deprecated and renamed to `Frame::area`
+### `Frame::size` is deprecated and renamed to `Frame::area` ([#1293])
+
+[#1293]: https://github.com/ratatui/ratatui/pull/1293
 
 `Frame::size` is renamed to `Frame::area` as it's the more correct name.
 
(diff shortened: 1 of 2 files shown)


kdheepak:
Does `foo.into()` work? e.g.:

```rust
let line = Line::from(foo.into());
```

kdheepak (approved):

BREAKING-CHANGES.md
```diff
@@ -65,6 +65,26 @@ This is a quick summary of the sections below:
   - MSRV is now 1.63.0
   - `List` no longer ignores empty strings
 
+## v0.29.0 (Unreleased)
+
+### `Line` now implements `From<Cow<str>` ([#1373])
+
+[#1373]: https://github.com/ratatui/ratatui/pull/1373
+
+As this adds an extra conversion, ambiguous inferred values may no longer compile.
```
kdheepak, line 74: Should this be written like this:

```suggestion
As this adds an extra conversion, ambiguously inferred values may no longer compile.
```

?

joshka:
> Does `foo.into()` work? e.g.:
> 
> ```rust
> let line = Line::from(foo.into());
> ```

No, that would have the same problem. both String and Cow are acceptable inferences for Line::from.

BREAKING-CHANGES.md
```diff
@@ -136,8 +156,6 @@ This change simplifies the trait and makes it easier to implement.
 
 ### `Frame::size` is deprecated and renamed to `Frame::area`
 
-[#1293]: https://github.com/ratatui/ratatui/pull/1293
-
```
kdheepak, line 140: Is there a reason this was removed?

BREAKING-CHANGES.md
joshka: I meant it as an adjective attaching to the noun phrase "inferred values", rather than an an adverb attaching to "inferred". "Values" should be "expressions" here. The concrete part of this is that there is an inferred expression, which is ambiguous when evaluated. I lean 80/20 towards the original wording (but would change it to expression).

BREAKING-CHANGES.md
joshka: This was unintentional, and done automatically as it was unused (It's not referenced in the line above, this is an error)

BREAKING-CHANGES.md
joshka: bd5fa963f..eaa180cc6

BREAKING-CHANGES.md
joshka: bd5fa963f..eaa180cc6

joshka:
This is likely safe to keep around until we drop 0.29.0.

orhun (approved):

PR #1374, opened 2024-09-16
fix(block tests docs examples): mark title as deprecated

TadoTheMiner:
This PR marks title as deprecated, adds title_at_position function  and adapts code and function docs to not use Title. The title_at_position function is required as what used to be 
```rust
title(Title::from("foo").position(position))
```
would have to be 
```rust
match position { 
  TitlePosition::Top => title_top("foo"),
  TitlePosition::Bottom => title_bottom("foo"),
}
```
which is too  long compared to 
```rust
title_at_position("foo", position)
```

### DIFF
--- a/src/widgets/sparkline.rs
+++ b/src/widgets/sparkline.rs
@@ -254,6 +254,7 @@ mod tests {
     #[test]
     fn it_does_not_panic_if_max_is_set_to_zero() {
         // see https://github.com/rust-lang/rust-clippy/issues/13191
+        #[allow(unknown_lints)]
         #[allow(clippy::unnecessary_min_or_max)]
         let widget = Sparkline::default().data(&[0, 1, 2]).max(0);
         let buffer = render(widget, 6);
--- a/benches/main/block.rs
+++ b/benches/main/block.rs
@@ -2,10 +2,8 @@ use criterion::{criterion_group, BatchSize, Bencher, Criterion};
 use ratatui::{
     buffer::Buffer,
     layout::{Alignment, Rect},
-    widgets::{
-        block::{Position, Title},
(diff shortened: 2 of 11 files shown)


TadoTheMiner:
Should I change the version in Cargo.toml?

src/widgets/sparkline.rs
```diff
@@ -254,6 +254,7 @@ mod tests {
     #[test]
     fn it_does_not_panic_if_max_is_set_to_zero() {
         // see https://github.com/rust-lang/rust-clippy/issues/13191
+        #[allow(unknown_lints)]
```
joshka, line 257: ?

joshka:
I'd prefer not having title_at_position, as it seems like that would not be a common enough use case to have a runtime configurable title position. The only place I could really anticipate that would need that might is in places where the if statement wouldn't be burdensome. Am I missing something that would make this really useful?

If we don't put this in now, it's something that can be added fairly simply later if that assumption is invalid.

Deprecating Title is handled by #1372, which goes a bit further and cleans up areas in the tests that are over complicated (like `.alignment(Alignment::Center)` -> `.centered()` and places where we convert from a str -> span -> Title).

joshka:
> Should I change the version in Cargo.toml?

No. We handle that automatically with our build scripts.

orhun:
Hey, I was wondering what's the status on this? @TadoTheMiner  👀

joshka:
> Hey, I was wondering what's the status on this? @TadoTheMiner 👀

I don't think there's much left in this to merge. I'd tend to close it as completed by #1372.

Closing it, but if I've done so by mistake and there's code in here that does need to be part of the next release, please feel free to reopen.

PR #1378, opened 2024-09-16
fix(rect)!: Rect::area now returns u32 and Rect::new() no longer clamps area to u16::MAX

joshka:
This change fixes the unexpected behavior of the Rect::new() function to
be more intuitive. The Rect::new() function now clamps the width and
height of the rectangle to keep each bound within u16::MAX. The
Rect::area() function now returns a u32 instead of a u16 to allow for
larger areas to be calculated.

Previously, the Rect::new() function would clamp the total area of the
rectangle to u16::MAX, by preserving the aspect ratio of the rectangle.

BREAKING CHANGE: Rect::area() now returns a u32 instead of a u16.

Fixes: <https://github.com/ratatui/ratatui/issues/1375>

---

TODO

- [x] Rect::new() still needs to clamp the width and height so that the any index bounds computed from x and width / y and 
height fit inside the u16 bounds as otherwise this would lead to non-indexable positions (and generally more overflows).
- [x] Rect::area() should return u32
- [x] All multiplications not in examples / tests should be checked (and tests added) for overflow
- [x] Additions should be investigated as to whether they can also cause overflow, but testing should be a lower priority as any overflows in addition would have already had to be failing

### DIFF
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -71,6 +71,13 @@ This is a quick summary of the sections below:
 
 ## v0.29.0 (Unreleased)
 
+### `Rect::area()` now returns u32 instead of u16 ([#1378])
+
+[#1378]: https://github.com/ratatui/ratatui/pull/1378
+
+This is likely to impact anything which relies on `Rect::area` maxing out at u16::MAX. It can now
+return up to u16::MAX * u16::MAX (2^32 - 2^17 + 1).
+
 ### `Line` now implements `From<Cow<str>` ([#1373])
 
 [#1373]: https://github.com/ratatui/ratatui/pull/1373
--- a/benches/main/buffer.rs
+++ b/benches/main/buffer.rs
@@ -8,12 +8,7 @@ use ratatui::{
 criterion::criterion_group!(benches, empty, filled, with_lines);
 
 const fn rect(size: u16) -> Rect {
-    Rect {
-        x: 0,
-        y: 0,
-        width: size,
-        height: size,
-    }
+    Rect::new(0, 0, size, size)
 }
 
 fn empty(c: &mut Criterion) {
--- a/examples/canvas.rs
+++ b/examples/canvas.rs
@@ -48,7 +48,7 @@ struct App {
 }
 
 impl App {
-    fn new() -> Self {
+    const fn new() -> Self {
         Self {
             x: 0.0,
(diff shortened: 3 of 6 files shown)


BREAKING-CHANGES.md
```diff
@@ -136,8 +143,6 @@ This change simplifies the trait and makes it easier to implement.
 
 ### `Frame::size` is deprecated and renamed to `Frame::area`
 
-[#1293]: https://github.com/ratatui/ratatui/pull/1293
```
orhun, line 139: Is this change related?

orhun (commented):
What do you think of using `usize` instead of `u32` so it scales with the architecture’s pointer size?

kxxt:
> What do you think of using `usize` instead of `u32` so it scales with the architecture’s pointer size?

u32 is guaranteed to be enough for area because we use u16 for width/height. But yes, using `usize` could avoid a future breaking change if we replace u16 with u32/usize for width/height in the future

BREAKING-CHANGES.md
joshka: This is just markdownlint removing unused link refs - I fixed this in another PR IIRC.

joshka:
> What do you think of using `usize` instead of `u32` so it scales with the architecture’s pointer size?

Because the height and width of a terminal is hardcoded to u16, u16*u16 is always < u32::MAX, so usize isn't necessary, and might communicate the wrong thing. We do need to change the Paragraph scroll index to usize though, as that may actually be > u32 (because people want crazy things).

It might at some point be worth considering whether we should make the buffer size able to be larger than the screen max size, for virtual buffer stuff shenanigans, but this is simply about fixing up a long standing problem that's finally actually caused issues.

BREAKING-CHANGES.md
joshka: https://github.com/ratatui/ratatui/pull/1373#discussion_r1760509268

orhun:
Sounds reasonable - fixed clippy lints and rebased on main.

joshka:
I took a quick look to see if there were any obvious overflows and didn't see any that needed new tests.

joshka:
I took a quick look to see if there were any obvious overflows and didn't see any that needed new tests.

orhun (approved):

orhun:
rebased on main and conflicts are fixed

PR #1383, opened 2024-09-20
chore: make Debug output for Text/Line/Span/Style more concise

joshka:
Fixes: https://github.com/ratatui/ratatui/issues/1382

### DIFF
--- a/src/style.rs
+++ b/src/style.rs
@@ -71,6 +71,7 @@
 use std::fmt;
 
 use bitflags::bitflags;
+use color::StylizeDebugKind;
 pub use color::{Color, ParseColorError};
 pub use stylize::{Styled, Stylize};
 
@@ -223,7 +224,7 @@ impl fmt::Debug for Modifier {
 ///     buffer[(0, 0)].style(),
 /// );
 /// ```
-#[derive(Debug, Default, Clone, Copy, Eq, PartialEq, Hash)]
+#[derive(Default, Clone, Copy, Eq, PartialEq, Hash)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub struct Style {
     pub fg: Option<Color>,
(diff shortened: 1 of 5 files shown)


orhun (commented):
I just gave this a try and it works well - however I'm a bit concerned about the complexity that this implementation might gain in the future. I see that the debug output is now as close as possible to the actual API and doing that might require more _string work_ in the future. Also, there are some parts that can be extracted or can be simplified, e.g. `Color` -> snake_case color name conversion.

joshka:
> I just gave this a try and it works well - however I'm a bit concerned about the complexity that this implementation might gain in the future. I see that the debug output is now as close as possible to the actual API and doing that might require more _string work_ in the future. Also, there are some parts that can be extracted or can be simplified, e.g. `Color` -> snake_case color name conversion.

I'm not too worried about the future concerns as we can handle those in the future when we need it. Keeping it simple as possible though makes sense. I wouldn't introduce a crate for case conversion unless it was also used in some other part of Ratatui, but I would  extract color name conversion into a small function and consider making this public if it made sense.

orhun:
Yup, extracting that to a small function would be nice. Also, what else do you have in mind for going forward with this? 

I see some todos which I can take a look if you want.

joshka:
>I see some todos which I can take a look if you want.

Go for it - I'm just catching up after a few days away, but this should be an easy one to complete quickly:
- Modifiers should be .bold() .not_bold() etc.
- alignment should be Alignment::Left, etc. or hidden
- move the implementations of the things that make sense to the stylize module
- I was thinking maybe Span::from("") instead of Span(""), but can take or leave it.
- likely keep everything pub(crate) unless there's an obvious usefulness for it to be pub

A guiding idea is that debug view should not be considered stable, so we can change this as needed.

orhun:
Addressed the comments, can you take a look? :)

src/style/stylize.rs
```diff
@@ -23,6 +25,75 @@ pub trait Styled {
     fn set_style<S: Into<Style>>(self, style: S) -> Self::Item;
 }
 
+/// A helper struct to make it easy to debug using the `Stylize` method names
+pub(crate) struct StylizeDebug {
```
joshka, line 29: name => ColorDebug perhaps and ColorDebugKind below.

joshka (commented):
LGTM, one small nit on naming.

src/style/stylize.rs
orhun: a75abe5ef1b24633cab841336547ac7df797ca02

joshka:
LGTM. I can't approve this as the initial Author

orhun (approved):

PR #1384, opened 2024-09-21
chore: lock unicode-width version to 0.1.13

joshka:
I'm not sure what is causing CI failures, but 0.1.14 contains breakig changes which we'll need to investigate.

Changes https://github.com/unicode-rs/unicode-width/compare/v0.1.13...v0.1.14

Fixes problems found in https://github.com/ratatui/ratatui/pull/1341

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,7 +38,7 @@ termwiz = { version = "0.22.0", optional = true }
 time = { version = "0.3.11", optional = true, features = ["local-offset"] }
 unicode-segmentation = "1.10"
 unicode-truncate = "1"
-unicode-width = "0.1.13"
+unicode-width = "=0.1.13"
 
 [target.'cfg(not(windows))'.dependencies]
 # termion is not supported on Windows


kdheepak (approved):

orhun:
I think this is the culprit: https://github.com/unicode-rs/unicode-width/commit/afab363383557c39942706c9441a5670e321b4a0

> For example, many Brahmic scripts like Devanagari have complex rendering rules
which this crate does not currently handle (and will never fully handle, because
the exact rendering depends on the font):

This test also now fails:

```rust
assert_eq!("👩".width(), 2); // Woman
assert_eq!("🔬".width(), 2); // Microscope
assert_eq!("👩‍🔬".width(), 4); // Woman scientist -> should be 4 but it expect 2
```

I guess they no longer take zero width joiner into account 💀 should we communicate this to upstream or just create a wrapper function for short term?

joshka:
It's unlikely that communicating with upstream will change anything on this. Their stance on stability and communicating changes is incompatible with ours.

orhun:
I will look into the upsides/downsides of making the upgrade vs keeping the current behavior.

PR #1389, opened 2024-09-27
refactor: modularize

joshka:
WIP - this is just a PoC to understand whether this would work fine, termion and termwiz disabled for now

Poc for: https://github.com/ratatui/ratatui/issues/1388


orhun:
I think we are following up on this at #1459

PR #1390, opened 2024-09-28
chore: remove usage of prelude

joshka:
This helps make the doc examples more explicit about what is being used.
It will also makes it a bit easier to do future refactoring of Ratatui,
into several crates, as the ambiguity of where types are coming from
will be reduced.

Additionally, several doc examples have been simplified to use Stylize, and
necessary imports are no longer hidden.

### DIFF
--- a/benches/main/barchart.rs
+++ b/benches/main/barchart.rs
@@ -2,8 +2,7 @@ use criterion::{criterion_group, Bencher, BenchmarkId, Criterion};
 use rand::Rng;
 use ratatui::{
     buffer::Buffer,
-    layout::Rect,
-    prelude::Direction,
+    layout::{Direction, Rect},
     widgets::{Bar, BarChart, BarGroup, Widget},
 };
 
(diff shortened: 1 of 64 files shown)


orhun:
Does it remove the prelude or the usages? I am preparing a presentation/workshop and I kinda depend on it - and I was thinking of making improvements.

joshka:
> Does it remove the prelude or the usages? I am preparing a presentation/workshop and I kinda depend on it - and I was thinking of making improvements.

Just the usages - the prelude is still there and still documented as being available. This just makes the internal code not use it and the doc comment examples also not use it.

kdheepak (approved):

orhun (approved):

orhun:
This removes almost every reference to prelude, so the only way to know about it is to get it suggested by your LSP I guess. It is in crate docs - but I'm still wondering if we should mention it explicitly somewhere.

joshka:
image

orhun:
fair

PR #1395, opened 2024-10-02
test(example): add snapshot test for demo

orhun:
See #1381


joshka:
Neat - likely fully unnecessary for the demo, but it's a good demo of insta. cargo-insta might make a good ratatui-ified app at some point in the future...

PR #1398, opened 2024-10-02
feat!: Add overlap to layout

kdheepak:
This PR adds a new feature for the existing `Layout::spacing` method, and introducing a `Spacing` enum.

Now `Layout::spacing` is generic and can take

- zero or positive numbers, e.g. `Layout::spacing(1)` (current functionality)
- negative number, e.g. `Layout::spacing(-1)` (new)
- variant of the `Spacing` (new)

This allows creating layouts with a shared pixel for segments. When `spacing(negative_value)` is used, spacing is ignored and all segments will be adjacent and have pixels overlapping. `spacing(zero_or_positive_value)` behaves the same as before. These are internally converted to `Spacing::Overlap` or `Spacing::Space`. 

Here's an example output to illustrate the layout solve from this PR:

```rust
#[test]
fn test_layout() {
    use crate::layout::Constraint::*;
    let mut terminal = crate::Terminal::new(crate::backend::TestBackend::new(50, 4)).unwrap();
    terminal
        .draw(|frame| {
            let [upper, lower] = Layout::vertical([Fill(1), Fill(1)]).areas(frame.area());

            let (segments, spacers) = Layout::horizontal([Length(10), Length(10), Length(10)])
                .flex(Flex::Center)
                .split_with_spacers(upper);

            for segment in segments.iter() {
                frame.render_widget(
                    crate::widgets::Block::bordered()
                        .border_set(crate::symbols::border::DOUBLE),
                    *segment,
                );
            }
            for spacer in spacers.iter() {
                frame.render_widget(crate::widgets::Block::bordered(), *spacer);
            }

            let (segments, spacers) = Layout::horizontal([Length(10), Length(10), Length(10)])
                .flex(Flex::Center)
                .spacing(-1) // new feature
                .split_with_spacers(lower);

            for segment in segments.iter() {
                frame.render_widget(
                    crate::widgets::Block::bordered()
                        .border_set(crate::symbols::border::DOUBLE),
                    *segment,
                );
            }
            for spacer in spacers.iter() {
                frame.render_widget(crate::widgets::Block::bordered(), *spacer);
            }
        })
        .unwrap();
    dbg!(terminal.backend());
}
```

```plain
┌────────┐╔════════╗╔════════╗╔════════╗┌────────┐
└────────┘╚════════╝╚════════╝╚════════╝└────────┘
┌─────────┐╔════════╔════════╔════════╗┌─────────┐
└─────────┘╚════════╚════════╚════════╝└─────────┘
```

Currently drawing a border on top of an existing border overwrites it. Future PRs will allow for making the border drawing handle overlaps better.


src/layout/layout.rs
```diff
@@ -2236,6 +2283,32 @@ mod tests {
...
         }
 
+        #[rstest]
+        #[case::length_overlap(vec![(0  , 20) , (20 , 20) , (40 , 20)] , vec![Length(20) , Length(20) , Length(20)] , Flex::Start        , 0)]
+        #[case::length_overlap(vec![(0  , 20) , (19 , 20) , (38 , 20)] , vec![Length(20) , Length(20) , Length(20)] , Flex::Start        , 1)]
+        #[case::length_overlap(vec![(21 , 20) , (40 , 20) , (59 , 20)] , vec![Length(20) , Length(20) , Length(20)] , Flex::Center       , 1)]
+        #[case::length_overlap(vec![(42 , 20) , (61 , 20) , (80 , 20)] , vec![Length(20) , Length(20) , Length(20)] , Flex::End          , 1)]
+        #[case::length_overlap(vec![(0  , 20) , (19 , 20) , (38 , 62)] , vec![Length(20) , Length(20) , Length(20)] , Flex::Legacy       , 1)]
+        #[case::length_overlap(vec![(0  , 34) , (33 , 34) , (66 , 34)] , vec![Length(20) , Length(20) , Length(20)] , Flex::SpaceBetween , 1)]
+        #[case::length_overlap(vec![(0  , 34) , (33 , 34) , (66 , 34)] , vec![Length(20) , Length(20) , Length(20)] , Flex::SpaceAround  , 1)]
```
joshka, line 2293: Try to use unique (distinct) test names here to make it possible to work out which test fails easily. The purpose of ::test_name is to change the test name from `case_01`, ... to `case_01_length_overlap.

src/layout/layout.rs
```diff
@@ -2236,6 +2283,32 @@ mod tests {
...
+            #[case] overlap: u16,
+        ) {
+            let rect = Rect::new(0, 0, 100, 1);
+            let r = Layout::horizontal(constraints)
+                .flex(flex)
+                .overlap(overlap.into())
+                .split(rect);
+            let result = r
+                .iter()
+                .map(|r| (r.x, r.width))
```
joshka, line 2307: Not related to this specific test, but the output of these tests when they fail is particularly poor. Would be nice to have a small struct for this that has start and width instead of the tuple.

E.g.
```
thread 'layout::layout::tests::split::fill_overlap::case_07_flex10' panicked at src/layout/layout.rs:2488:13:
assertion failed: `(left == right)`

Diff < left / right > :
 [
     (
         0,
<        55,
>        50,
     ),
     (
<        45,
<        55,
>        50,
>        50,
     ),
 ]
```

src/layout/layout.rs
```diff
@@ -2236,6 +2283,32 @@ mod tests {
...
+                .collect::<Vec<(u16, u16)>>();
+            assert_eq!(expected, result);
```
joshka, line 2309: Any idea why these are backwards compared to the rest of the lib (actual, expected)?

src/layout/layout.rs
```diff
@@ -657,10 +677,16 @@ impl Layout {
...
 
         let area_size = Element::from((*variables.first().unwrap(), *variables.last().unwrap()));
         configure_area(&mut solver, area_size, area_start, area_end)?;
-        configure_variable_constraints(&mut solver, &variables, area_size)?;
+        configure_variable_in_area_constraints(&mut solver, &variables, area_size)?;
+        if overlap == 0 {
+            configure_variable_ascending_order_constraints(&mut solver, &variables)?;
+        } else {
+            configure_variable_overlap_constraints(&mut solver, &segments, overlap)?;
+        }
```
joshka, line 689: Would be nice if this didn't have to be a special case. I'm surprised that it does have to be. What's different about having an overlap of 1 and an overlap of 0?

src/layout/layout.rs
kdheepak: I agree, the tests failing here are hard to debug. What about something like this?

image

https://github.com/ratatui/ratatui/pull/1398/commits/03be41a02bee5ff0be386b4f4003aa94e60f4bbc

src/layout/layout.rs
kdheepak: Done in https://github.com/ratatui/ratatui/pull/1398/commits/eff582110be19226c57e5778590cc7a99018e1aa

src/layout/layout.rs
kdheepak: There's a number of places in this file where it is backwards. I copy pasted another test from this file and edited. This is something that we should fix in a separate PR.

kdheepak: Opened an issue to track this: https://github.com/ratatui/ratatui/issues/1399

src/layout/layout.rs
kdheepak: overlap 1 only adds constraints for segments but overlap 0 adds constraints for all variables (i.e. segments and spacers).

src/layout/layout.rs
joshka: Perhaps check starts and widths separately? It's tough to know what to do here (and it's a bit of a micro-optimization I guess)

I can see what you're doing there. The idea is good - perhaps could work by showing the rects as full rects with borders.

src/layout/layout.rs
joshka: I think what you're saying here is that the overlap eats up any space that would have been allocated to the spacer, but the spacer doesn't eat any space allocated to the segments. Does that sound correct?

src/layout/layout.rs
kdheepak: I think regardless of what we do here that should happen in a separate PR, because we have to change every test here to use the new format.

kdheepak: I've opened an issue to track this: https://github.com/ratatui/ratatui/issues/1400

src/layout/layout.rs
kdheepak: To make sure we are on the same page, here's the terminology.

We have the following area:

```plain
<───────────────────────────────────area_width─────────────────────────────────>
┌─area_start                                                          area_end─┐
v                                                                              v
┌   ┐┌──────────────────┐┌   ┐┌──────────────────┐┌   ┐┌──────────────────┐┌   ┐
     │     Max(20)      │     │      Max(20)     │     │      Max(20)     │
└   ┘└──────────────────┘└   ┘└──────────────────┘└   ┘└──────────────────┘└   ┘
```

We create elements `Vec<Element>` where each `Element` has a `start` and `end`:

```plain
┌────┬───────────────────┬────┬───element.start───┬────┬───────────────────┐
│    │                   │    │                   │    │                   │
v    v                   v    v                   v    v                   v
┌   ┐┌──────────────────┐┌   ┐┌──────────────────┐┌   ┐┌──────────────────┐┌   ┐
     │     Max(20)      │     │      Max(20)     │     │      Max(20)     │
└   ┘└──────────────────┘└   ┘└──────────────────┘└   ┘└──────────────────┘└   ┘
    ^                   ^    ^                   ^    ^                   ^    ^
    │                   │    │                   │    │                   │    │
    └───────────────────┴────┴───element.end─────┴────┴───────────────────┴────┘
```

"segments" are `Vec<Element>` only for the areas of interest:

```plain
     ┌────────────────────────┬───segment.start────────┐
     │                        │                        │
     v                        v                        v
┌   ┐┌──────────────────┐┌   ┐┌──────────────────┐┌   ┐┌──────────────────┐┌   ┐
     │     Max(20)      │     │      Max(20)     │     │      Max(20)     │
└   ┘└──────────────────┘└   ┘└──────────────────┘└   ┘└──────────────────┘└   ┘
                        ^                        ^                        ^
                        │                        │                        │
                        └──────────segment.end───┴────────────────────────┘
```

"spacers" are `Vec<Element>` the remaining gaps:

```plain
┌────────────────────────┬────────spacer.start────┬────────────────────────┐
│                        │                        │                        │
v                        v                        v                        v
┌   ┐┌──────────────────┐┌   ┐┌──────────────────┐┌   ┐┌──────────────────┐┌   ┐
     │     Max(20)      │     │      Max(20)     │     │      Max(20)     │
└   ┘└──────────────────┘└   ┘└──────────────────┘└   ┘└──────────────────┘└   ┘
    ^                        ^                        ^                        ^
    │                        │                        │                        │
    └────────────────────────┴────spacer.end──────────┴────────────────────────┘
```

joshka: yup - this aligns with my understanding. But adding this image is good for making it easy to visualize conversations, so thanks. The only thing I'd add to the description is that the variable's value really effectively points to the left side of the cell. This subtle distinction helps make it a bit more clear why the end is 1 more than the cell index, and makes the floating point values less ambiguous.

kdheepak: > I think what you're saying here is that the overlap eats up any space that would have been allocated to the spacer, but the spacer doesn't eat any space allocated to the segments. Does that sound correct?

Previously, we were always adding this constraint:

```rust
    // all variables are in ascending order
    for (&left, &right) in variables.iter().tuple_windows() {
        solver.add_constraint(left | LE(REQUIRED) | right)?;
    }
    Ok(())
```

If we use individual variables:

```plain
┌────┬───────────────────┬────┬─────variables─────┬────┬───────────────────┬────┐
│    │                   │    │                   │    │                   │    │
v    v                   v    v                   v    v                   v    v
┌   ┐┌──────────────────┐┌   ┐┌──────────────────┐┌   ┐┌──────────────────┐┌   ┐
     │     Max(20)      │     │      Max(20)     │     │      Max(20)     │
└   ┘└──────────────────┘└   ┘└──────────────────┘└   ┘└──────────────────┘└   ┘
^    ^                   ^    ^                   ^    ^                   ^    ^
└v1  └v2                 └v3  └v4                 └v5  └v6                 └v7  └v8
```

this does:

```
v1 <= v2
v2 <= v3
v3 <= v4
v4 <= v5
v5 <= v6
v6 <= v7
v7 <= v8
```

But when we have overlaps, we cannot do this anymore.

joshka: yes, but we can do generically:

```
v1 - v2 <= 0 // not sure about this one
v2 - v3 <= 0
v3 - v4 <= overlap
v4 - v5 <= 0
v5 - v6 <= overlap
v6 - v7 <= 0
v7 - v8 <= 0 // not sure about this one either
```

kdheepak: I _think_ we have to force `overlap`, i.e.

```
v3 - v4 == overlap
```

instead of 

```
v3 - v4 <= overlap
```

Otherwise we may get overlaps that are not what the user wants.

orhun:
Hmm, I'm curious about where this might be practically useful 🤔

joshka:
> Hmm, I'm curious about where this might be practically useful 🤔

Collapsed borders is the primary use case for this.

src/layout/layout.rs
```diff
@@ -453,6 +453,8 @@ impl Layout {
     pub const fn overlap(mut self, overlap: bool) -> Self {
         if overlap {
             self.spacing = None;
+        } else if self.spacing.is_none() {
+            self.spacing = Some(0)
```
joshka, line 457: This makes `foo.spacing(1).overlap(1)` different to `foo.overlap(1).spacing(1)`. Mostly our builder methods should not have order dependent methods like this.

kdheepak: I agree. The only way around this though is by making a breaking change to `spacing`. I'm open to suggestions on what to do here.

joshka: If these are truly not able to be set together, then they likely deserve to be on an enum with mutually exclusive variants. My initial question on this was on the discord thread sort of had that idea in mind (setting spacing to negative values). If we took that approach, then spacing could be migated to accept `Into<Spacing>` => `enum Spacing { Spaced(u16), Overlap(u16) }`, which would give space for documentation. Alternatively spacing could just become i16 (or i32 if you want to allow layouts that have spacers / overlaps that are > 32767). This would be technically breaking, but not in a way that generally matters to anyone.

kdheepak: I think making it `i16` or `i32` would work. It'd break apps that use `spacing` like we do in constraint explorer:

https://github.com/ratatui/ratatui/blob/edcdc8a8147a2f450d2c871b19da6d6383fd5497/examples/constraint-explorer.rs#L47

But maybe that's fine?

src/layout/layout.rs
```diff
@@ -41,6 +41,46 @@ thread_local! {
...
+impl Default for Spacing {
+    fn default() -> Self {
+        Self::Space(0)
+    }
+}
+
+impl From<u32> for Spacing {
+    fn from(value: u32) -> Self {
+        Self::from(value.clamp(u32::from(u16::MIN), u32::from(u16::MAX)) as u16)
+    }
```
joshka, line 59: Why do we need from u32? There shouldn't be any code that already uses u32 for backwards compat right?

kdheepak: I thought I tried that first. If I don't implement `i32`, I get compile errors on some tests:

image

But it looks like implementing `u32` is not required. Resolved in https://github.com/ratatui/ratatui/pull/1398/commits/3b817e598a49c591ab51b032507a12a9808f1c85

src/layout/layout.rs
kdheepak: Removed `overlap` method in https://github.com/ratatui/ratatui/pull/1398/commits/57dbafe8c7fd42d1901cbc41810b3c609f578ae0

kdheepak:
I made some changes in https://github.com/ratatui/ratatui/pull/1398/commits/ae72e0b2a0b0e787a4e2a796c5a99a5c4d02b737 that change the behavior of overlap in SpaceBetween and SpaceAround. imo, this is more consistent with the behavior of SpaceBetween and SpaceAround with positive spacing, i.e. rules and precedence for positive spacing are the same as negative spacing. 

Apply this patch to `constraint-explorer.rs` to test out this PR:

```
diff --git a/examples/constraint-explorer.rs b/examples/constraint-explorer.rs
index a467551..a729955 100644
--- a/examples/constraint-explorer.rs
+++ b/examples/constraint-explorer.rs
@@ -44,7 +44,7 @@ fn main() -> Result<()> {
 #[derive(Default)]
 struct App {
     mode: AppMode,
-    spacing: u16,
+    spacing: i16,
     constraints: Vec<Constraint>,
     selected_index: usize,
     value: u16,
```

joshka:
@kdheepak can you please update the first comment to cover document this for a changelog entry?

orhun:
Can you wait for my review before merge? I'd like to see how this works :)

joshka (commented):
Need to probably add a change to the constraint explorer to support this too

src/layout/layout.rs
```diff
@@ -111,13 +173,13 @@ thread_local! {
 ///
 /// [`cassowary-rs`]: https://crates.io/crates/cassowary
 /// [Examples]: https://github.com/ratatui/ratatui/blob/main/examples/README.md
-#[derive(Debug, Default, Clone, Eq, PartialEq, Hash)]
+#[derive(Default, Debug, Clone, Eq, PartialEq, Hash)]
```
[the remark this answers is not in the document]
kdheepak: My bad, I implemented a custom `Default` and the undid the change to cause this

joshka:
Oh, and btw, I'm +Approve generally on this now - I wasn't 100% sure at first, but switching to the enum makes it work well

src/layout/layout.rs
```diff
@@ -402,9 +464,11 @@ impl Layout {
     /// Sets the spacing between items in the layout.
     ///
     /// The `spacing` method sets the spacing between items in the layout. The spacing is applied
-    /// evenly between all items. The spacing value represents the number of cells between each
+    /// evenly between all segments. The spacing value represents the number of cells between each
     /// item.
     ///
+    /// Spacing can be negative in which case it will cause overlap between segments.
+    ///
```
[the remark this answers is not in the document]
orhun: I think they are now added in the doc comments of `Spacing`.

src/layout/layout.rs
```diff
@@ -530,7 +605,7 @@ impl Layout {
         self.split_with_spacers(area).0
     }
 
-    /// Wrapper function around the cassowary-r solver that splits the given area into smaller ones
+    /// Wrapper function around the cassowary-rs solver that splits the given area into smaller ones
```
orhun, line 608: We can actually drop the `-rs` suffix here since the library is called "cassowary".

src/layout/layout.rs
```diff
@@ -2540,4 +2717,47 @@ mod tests {
...
+                            .border_set(crate::symbols::border::DOUBLE),
+                        *segment,
+                    );
+                }
+                for spacer in spacers.iter() {
+                    frame.render_widget(crate::widgets::Block::bordered(), *spacer);
+                }
+            })
+            .unwrap();
+        dbg!(terminal.backend());
```
orhun, line 2761: We should probably do a buffer assertion here :)

orhun (approved):
LGTM!

What is in your mind about adding a practical example for this? I think we can either use it in the examples somewhere or mention it in the recipes (e.g. collapsing borders seems related to this).

joshka:
> What is in your mind about adding a practical example for this? I think we can either use it in the examples somewhere or mention it in the recipes (e.g. collapsing borders seems related to this).

This is step one towards collapsing borders, Step 2 is making block able to handle overwriting borders - that can be also pretty complex, as it requires merging symbol sets / building up table(s) of combinations and mappings.

This looks something like:
1. app draws initial block
2. app draws second block - for every border cell, the block checks for a non-blank character and then computes the new character to use based on the desired character from the symbol set, combined with the existing symbol in the current cell.

I imagine this would likely involve hard coding a bunch of combinations to start. It's easy if the lines are in the same line set (normal, dotted, heavy, double, ...), but there combinations for each of the intersection of sets. E.g. all the following are different:
```text
┥ 	┦ 	┧ 	┨ 	┩ 	┪ 	┫ 
```

Then on top of that, because we offer custom sets, this should be something that can be configured for the user's particular choice of sets. This part doesn't have to happen in an initial PR though, but it should happen - we can default to just overwriting when we don't have a mapping.

The thread related to this is at https://discord.com/channels/1070692720437383208/1291063903987433503

src/layout/layout.rs
```diff
@@ -41,6 +41,68 @@ thread_local! {
...
+///
+/// - From `u16`: Directly converts the value to `Spacing::Space`.
+/// - From `i16`: Converts negative values to `Spacing::Overlap` and non-negative values to
+///   `Spacing::Space`.
+/// - From `i32`: Clamps the value to the range of `i16` and converts negative values to
+///   `Spacing::Overlap` and non-negative values to `Spacing::Space`.
+///
+/// See the [`Layout::spacing`] method for details on how to use this enum.
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub enum Spacing {
```
joshka, line 73: needs to be re-exported in src/layout.rs

src/layout/layout.rs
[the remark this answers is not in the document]
kdheepak: Resolved in 276694932388

src/layout/layout.rs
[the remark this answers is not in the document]
kdheepak: Resolved in 590ac83f23f8

src/layout/layout.rs
```diff
@@ -415,13 +479,24 @@ impl Layout {
...
+    ///
+    /// ```rust
+    /// use ratatui::layout::{Constraint::*, Layout};
+    /// let layout = Layout::horizontal([Length(20), Length(20), Length(20)]).spacing(-1);
+    /// ```
+    ///
     /// # Notes
     ///
     /// - If the layout has only one item, the spacing will not be applied.
     /// - Spacing will not be applied for [`Flex::SpaceAround`] and [`Flex::SpaceBetween`]
```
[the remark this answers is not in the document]
kdheepak: Resolved in a45d4dbc43ea

src/layout/layout.rs
kdheepak: Resolved in a0308d68eaf6

src/layout/layout.rs
kdheepak: Made an issue since it might be a good first issue for someone intending to contribute - #1423

src/layout/layout.rs
kdheepak: This was just my function to print the ascii content for the PR description. I've removed the test now.

kdheepak:
> Need to probably add a change to the constraint explorer to support this too

I'd like to do this after we figure out better drawing methods for overlapping.

kdheepak:
This should be good to go! If the PR description looks good, I can merge.

joshka:
I'd suggest cutting the example down to something more concise for the changelog - just two constraints, but otherwise yep - good to merge.

kdheepak:
I wanted to show 3 blocks, because it shows the left and right border for the middle block. I can decrease the width of the output?

Issue #1271, opened 2024-08-03
Discussion: how to fix text width calculation


SUPERCILEX:
See https://github.com/ratatui-org/ratatui/pull/1249#issuecomment-2266337009 and https://github.com/ratatui-org/ratatui/pull/1270#discussion_r1702440545

`UnicodeWidthStr::width` counts newlines as one unit which is not what we want for terminal rendering. We need to figure out how we want to fix this. Some options:
1. Slap some one-off fixes in Line/Span. Clearly a bad idea as there are 45 usages of width so things wouldn't be consistent.
1. Replace `width` with our own extension (maybe called `renderable_width`) which does the right thing for ratatui.
1.  Match rendering to whatever `width` says, e.g. render a space for a newline.

I don't really have a preference between 2 and 3, but I think most people prefer 2?

cc: @joshka @EdJoPaTo

EdJoPaTo:
Not sure where the problem is. Rendering ignores graphemes of width 0 and control characters. Only after that the width is relevant for rendering. At that point `\n` is gone and doesn't influence anything.

The only thing I see here are multiple places of truth, but that should be fixed by reducing the amount of render functions to ideally 1. Currently, `set_stringn` feels like the most basic one which should be used (indirectly) in all the other places. (For example #1259 starts to get rid of some of them.)

joshka:
Option 2, but it we don't need an extension. Text/Line/Span::width are just methods. They should return the width that will be rendered, not the width including control characters.

SUPERCILEX:
> Not sure where the problem is.

I guess my concern is that there a number of usages of `UnicodeWidthStr` that may or may not be correct. Some examples:
- https://github.com/ratatui-org/ratatui/blob/8061813f324c08e11196e62fca22c2f6b9216b7e/src/widgets/scrollbar.rs#L581
- https://github.com/ratatui-org/ratatui/blob/8061813f324c08e11196e62fca22c2f6b9216b7e/examples/table.rs#L301
- https://github.com/ratatui-org/ratatui/blob/7e1bab049bc5acd4a5cb13189e8f86fec5813ab8/src/buffer/buffer.rs#L369

If we only touch Span, I'm worried it'll be a game of wack-a-mole with the other uses.

joshka:
> I guess my concern is that there a number of usages of `UnicodeWidthStr` that may or may not be correct. Some examples:

[...]

PR #1403, opened 2024-10-05
chore(deps)!: pin unicode-width to 0.2.0

orhun:
It is probably a good idea to take a look at those tests and why the output has changed to get a better idea of how this will impact the users.

See #1271

Closes #1385


joshka (approved):
Approved with the logic from #1271

PR #1406, opened 2024-10-06
refactor: Consistent result expected in layout tests

### LINKED ISSUE #1399: `assert_eq!` should be consistently used in ratatui
## Problem

in `./src/layout/layout.rs` we currently use both:

https://github.com/ratatui/ratatui/blob/eff582110be19226c57e5778590cc7a99018e1aa/src/layout/layout.rs#L2529

and 

https://github.com/ratatui/ratatui/blob/eff582110be19226c57e5778590cc7a99018e1aa/src/layout/layout.rs#L1198-L1199

These are inconsistent. 

## Solution

`assert_eq!` should be of the form:

```
assert_eq!(result, expected)
```

farmeroy:
Fixes #1399 
I've looked through all the `assert_eq` and made sure that they follow the `expected, result` pattern. I wasn't sure if it was desired to actually pass result and expected as variables to the assert_eq statements, so I've left everything that seems to have followed the pattern as is.

### DIFF
--- a/src/layout/layout.rs
+++ b/src/layout/layout.rs
@@ -1864,7 +1864,7 @@ mod tests {
                 ])
                 .split(target);
 
-            assert_eq!(target.height, chunks.iter().map(|r| r.height).sum::<u16>());
+            assert_eq!(chunks.iter().map(|r| r.height).sum::<u16>(), target.height);
             chunks.windows(2).for_each(|w| assert!(w[0].y <= w[1].y));
         }
 
--- a/src/layout/layout.rs
+++ b/src/layout/layout.rs
@@ -1957,7 +1957,7 @@ mod tests {
                 .iter()
                 .map(|r| r.width)
                 .collect::<Vec<u16>>();
-            assert_eq!(expected, r);
+            assert_eq!(r, expected);
         }
 
         #[rstest]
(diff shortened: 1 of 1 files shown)


kdheepak (approved):

joshka:
> I wasn't sure if it was desired to actually pass result and expected as variables to the assert_eq statements, so I've left everything that seems to have followed the pattern as is.

This is fine as is. My general rule for tests is to write them such that they are as simple and obvious as possible, balancing:
- consistency: the tests should look similar when testing similar things
- density: shorter (in LoC) tests mean that you can see more tests at once. I'll often use or not use a variable depending on which leads to fewer lines to look at.
- failure mode: should make it easy to understand why the test failed and what to do to fix it

We've talked a little more offline about the layout tests. There's some small annoyances (to me at least) in the them that make their failure mode difficult to reason about. This is because testing just the start and width of a region forces you to do math to work out whether the results are adjacent. Better would be to report these as `start..end` or `start..end (width)`. This is no huge deal though, as it only really adds value when working on changes to the code.

kdheepak:
I agree with @joshka's suggestions. We should open issues to track those kinds of changes and tackle them in future PRs.

PR #1408, opened 2024-10-07
chore: add benchmark for `Table`

airblast-dev:
It is essentially the same as the `List` benchmark, the only logic difference being the variable column counts.

### DIFF
--- a/benches/main/table.rs
+++ b/benches/main/table.rs
@@ -1,0 +1,69 @@
+use criterion::{criterion_group, BatchSize, Bencher, BenchmarkId, Criterion};
+use ratatui::{
+    buffer::Buffer,
+    layout::{Constraint, Rect},
+    widgets::{Row, StatefulWidget, Table, TableState, Widget},
+};
+
+/// Benchmark for rendering a table.
+/// It only benchmarks the render with a different number of rows, and columns.
+fn table(c: &mut Criterion) {
+    let mut group = c.benchmark_group("table");
+
+    for row_count in [64, 2048, 16384] {
+        for col_count in [2, 4, 8] {
+            let bench_sizes = format!("{row_count}x{col_count}");
+            let rows: Vec<Row> = (0..row_count)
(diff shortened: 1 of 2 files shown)


orhun (approved):

benches/main/table.rs
```diff
@@ -0,0 +1,69 @@
...
+        for col_count in [2, 4, 8] {
+            let bench_sizes = format!("{row_count}x{col_count}");
+            let rows: Vec<Row> = (0..row_count)
+                .map(|_| Row::new((0..col_count).map(|_| fakeit::words::quote())))
+                .collect();
+
+            // Render default table
+            group.bench_with_input(
+                BenchmarkId::new("render", &bench_sizes),
+                &Table::new(rows.clone(), [] as [Constraint; 0]),
```
joshka, line 23: Doesn't this render the columns all with 0 width? (and in the one below...)
Maybe `vec![Fill(1); n]` would work here.

airblast-dev: If widths are not provided, it falls back to distributing lengths equally to all columns.
https://github.com/ratatui/ratatui/blob/1153a9ebaf0b98c45982002a659cb718e3c1d137/src/widgets/table/table.rs#L36
https://github.com/ratatui/ratatui/blob/1153a9ebaf0b98c45982002a659cb718e3c1d137/src/widgets/table/table.rs#L831

(Side note: I think the docs actually conflict on what the behavior is.)
https://github.com/ratatui/ratatui/blob/1153a9ebaf0b98c45982002a659cb718e3c1d137/src/widgets/table/table.rs#L26

There is nothing wrong with using `vec![Fill(1); n]` instead so if it is preferred, thats fine.

joshka: Oh yeah, I forgot that we did that that (I wouldn't be surprised if I actually wrote that comment too - the refactoring into a different file hides the original author).

That code that calculates the widths was probably written before Fill was created, and maybe should be moved to that to fully allocate columns rather than allocating like this. Not part of this PR though.

PR #1412, opened 2024-10-13
feat(tabs)!: unselect tab

sanpii:
I don't know if this is a good idea. I use tabs to display shortcuts:

2024-10-13-212428_1680x1050_scrot

In this case, I don’t want to select a tab. Currently I modify the hl style.

### DIFF
--- a/src/widgets/tabs.rs
+++ b/src/widgets/tabs.rs
@@ -51,7 +51,7 @@ pub struct Tabs<'a> {
     /// One title for each tab
     titles: Vec<Line<'a>>,
     /// The index of the selected tabs
-    selected: usize,
+    selected: Option<usize>,
     /// The style used to draw the text
     style: Style,
     /// Style to apply to the selected item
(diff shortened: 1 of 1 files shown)


kdheepak:
I do like the idea. I'd personally prefer for it to be a breaking change with `select` taking `Option<usize>` and not having `unselect`, but that'll be a breaking change. Let's hear from others to decide.

Currently, it is still a breaking change without a new custom `Default` implementation.

orhun:
Agree with kd here, using an Option in `select` would be better and we should add a custom `Default` implementation to `Tabs` to avoid a breaking change.

joshka:
There's an implementation of [`From<T> for Option<T>`](https://doc.rust-lang.org/stable/std/option/enum.Option.html#impl-From%3CT%3E-for-Option%3CT%3E) that makes changing a parameter to `T: Into<Option<T>>` (mostly) non-breaking. The exception is where there is already type inference e.g.:
```rust
let selection = 0u8;
tabs.select(selection.into());
```

The `Default` implementation can't select the first tab as when it's called there are no tabs defined. Instead setting the first tab should happen in `new()` and a not yet created `tabs()` method.

PR #1418, opened 2024-10-14
refactor(color)!: use palette types for Hsl/Hsluv conversions

### LINKED ISSUE #1414: Fix Color hsl / hsluv conversions
Running a few of the test values through https://www.colorspaceconverter.com/converter/hsluv-to-rgb gives very different numbers, so there's something wrong with this as it stands.

I think it might be better to implement this conversions from palette types instead of as components and to do the same for the hsl conversion.

_Originally posted by @joshka in https://github.com/ratatui/ratatui/issues/1333#issuecomment-2409478621_

orhun:
BREAKING-CHANGE: Previously `Color::from_hsl` accepted components
as individual f64 parameters. It now accepts a single `palette::Hsl` value
and is gated behind a `palette` feature flag.

```diff
- Color::from_hsl(360.0, 100.0, 100.0)
+ Color::from_hsl(Hsl::new(360.0, 100.0, 100.0))
```

Fixes: <https://github.com/ratatui/ratatui/issues/1414>

### DIFF
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -73,9 +73,17 @@ This is a quick summary of the sections below:
 
 ## Unreleased
 
-### `Color::from_hsl` is now behind the `palette` feature
+### `Color::from_hsl` is now behind the `palette` feature and accepts `palette::Hsl` ([#1418])
 
-<!-- TODO -->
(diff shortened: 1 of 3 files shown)


joshka (approved):
Added the breaking changes doc

PR #1422, opened 2024-10-15
docs(breaking-changes): mention deprecated block::Title

orhun:
#1372


orhun:
Oops I guess it's not needed, just realized it's not breaking

PR #1427, opened 2024-10-17
fix(terminal): use the latest, resized area when clearing

roberth:
This is the same mistake as 0f4823977894cef51d5ffafe6ae35ca7ad56e1ac

The code that comes after is still incorrect, clearing from the cursor to the bottom of the screen multiple times.
I'm using this in an alternate "Inline" implementation, as described in https://github.com/ratatui/ratatui/issues/1426#issuecomment-2420571210, for which the latter problem doesn't matter, except perhaps to reduce flickering very very slightly.
I hope this patch is acceptable as is.

### DIFF
--- a/src/terminal/terminal.rs
+++ b/src/terminal/terminal.rs
@@ -476,7 +476,8 @@ where
                     .set_cursor_position(self.viewport_area.as_position())?;
                 self.backend.clear_region(ClearType::AfterCursor)?;
             }
-            Viewport::Fixed(area) => {
+            Viewport::Fixed(_) => {
+                let area = self.viewport_area;
                 for y in area.top()..area.bottom() {
                     self.backend.set_cursor_position(Position { x: 0, y })?;
                     self.backend.clear_region(ClearType::AfterCursor)?;


roberth:
The coverage job also fails on `main`.

> > https://github.com/ratatui/ratatui/issues/1426#issuecomment-2420571210

This issue is mostly unrelated to this fix, which stands alone.

orhun (approved):
LGTM.

Out of curiosity, can you share a minimal example for reproducing this bug?

joshka:
How about fixing this by making sure that the viewport::Fixed(area) is corrected instead? This will make it easier to not make this mistake in any future code. Thoughts on this approach?

```diff
--- a/src/terminal/terminal.rs
+++ b/src/terminal/terminal.rs
@@ -237,6 +237,9 @@ where
         self.buffers[self.current].resize(area);
         self.buffers[1 - self.current].resize(area);
         self.viewport_area = area;
+        if let Viewport::Fixed(_) = self.viewport {
+            self.viewport = Viewport::Fixed(area);
+        }
     }
 
     /// Queries the backend for size and resizes if it doesn't match the previous size.
@@ -476,8 +479,7 @@ where
                     .set_cursor_position(self.viewport_area.as_position())?;
                 self.backend.clear_region(ClearType::AfterCursor)?;
             }
-            Viewport::Fixed(_) => {
-                let area = self.viewport_area;
+            Viewport::Fixed(area) => {
                 for y in area.top()..area.bottom() {
                     self.backend.set_cursor_position(Position { x: 0, y })?;
                     self.backend.clear_region(ClearType::AfterCursor)?;
```

orhun:
Yup, makes more sense 👍🏼

roberth:
> ```diff
>          self.viewport_area = area;
> +        if let Viewport::Fixed(_) = self.viewport {
> +            self.viewport = Viewport::Fixed(area);
> +        }
> ```

This would still be a workaround for a redundancy in the data model, which I'm removing in  73b0f122e71c6e205e7d4613411c6046a0c6a023

It's nice to reduce the state space, and more concretely if we ever need to set `viewport_area` in a different place, or do some other refactor, we there's nothing that we can forget to update.

Also the comment on the new type lets us clarify the purpose of retaining the info that we do retain.

roberth:
:+1: I've moved the refactor from this PR to #1428.
It now only contains the first commit with the simple fix.

joshka:
Merged - thanks for the PR

PR #1428, opened 2024-10-18
refactor(terminal): add ViewportType

roberth:
Follow-up to #1427 
See https://github.com/ratatui/ratatui/pull/1427#issuecomment-2422090507
For after the release https://github.com/ratatui/ratatui/pull/1427#issuecomment-2422363853

### DIFF
--- a/src/terminal/terminal.rs
+++ b/src/terminal/terminal.rs
@@ -476,7 +476,8 @@ where
                     .set_cursor_position(self.viewport_area.as_position())?;
                 self.backend.clear_region(ClearType::AfterCursor)?;
             }
-            Viewport::Fixed(area) => {
+            Viewport::Fixed(_) => {
+                let area = self.viewport_area;
                 for y in area.top()..area.bottom() {
                     self.backend.set_cursor_position(Position { x: 0, y })?;
                     self.backend.clear_region(ClearType::AfterCursor)?;
--- a/src/terminal/terminal.rs
+++ b/src/terminal/terminal.rs
@@ -67,7 +67,7 @@ where
     /// Whether the cursor is currently hidden
     hidden_cursor: bool,
     /// Viewport
-    viewport: Viewport,
+    viewport_type: ViewportType,
     /// Area of the viewport
     viewport_area: Rect,
     /// Last known area of the terminal. Used to detect if the internal buffers have to be resized.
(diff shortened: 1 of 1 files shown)


joshka:
I'm swinging back around to review older PRs. I wonder if you would mind summarizing the purpose of this to remind me a bit about the problem and how the solution fixes it? I don't recall the specifics, so this would be really helpful.

joshka:
Ok, I think I possibly understand this a little better now.

Currently, when initializing the terminal with options, the caller can request FullScreen, Inline with a specific height, or a Fixed area viewport. This viewport is stored in the terminal struct. We also store the actual area that the viewport is using in the viewport_area field. From time to time we've made the error of thinking that Terminal::viewport.area is the value that indicates the current viewport's area. It is not. It is the requested area in the absence of resizing information.

This PR changes that a little to add a field name to the inline variant, and remove the Fixed variant's value. The purpose of that is to avoid the confusion.

What's missing (If my understanding is correct) is that both the Inline and Fixed area need to continue to tract these values so that the terminal viewport_area works correctly even when the terminal is resized to a size smaller than the requested height / fixed area. So both the existing values need to be kept. These should be used when resizing below the requested amount (and resizing back to the requested amount in order that the terminal rendering can just work regardless of whether there's enough space to render the requested amount.

Given that understanding, I think there's not a lot of value in adding a secondary enum to track this info with a changed name and removed FixedArea enum value. Instead I think it would be better to document the above in the enum and terminal field, and ensure that there is adequate testing of the mentioned invariants when resizing occurs. Additionally, renaming the params passed to the resize function so that their purpose is clear would be helpful. Last, we should investigate and ensure that the viewport when autoresize works correctly when the fixed viewport is set. I suspect this is not correct right now as we entirely skip resizing when a fixed viewport is set.

roberth:
@joshka Sorry for the late response.

Your understanding of the problem matches mine.
I thought of this as having:
- a user facing type, `Viewport`
- an internal configuration variable without outdated info, `ViewportType`
- the rest of the internal state, including the actual current area

In this context, I merely normalized `ViewportType` in terms of data model.

To me it seemed "efficient" to drop the unnecessary info, but if you prefer to reuse `Viewport`, there might be some benefit to that in terms of amount of code, so that is also fine by me.

Thank you for looking into this!

joshka:
> @joshka Sorry for the late response.

No probs.
 
> Your understanding of the problem matches mine. I thought of this as having:
...
> To me it seemed "efficient" to drop the unnecessary info, but if you prefer to reuse `Viewport`, there might be some benefit to that in terms of amount of code, so that is also fine by me.

Good to get confirmation I hadn't missed something. It was a while since I had really delved into that code.

Originally I agreed with the direction you took when you originally raised this, but on digging a bit deeper I think the rationale for keeping the existing struct makes more sense (because if represents what the caller wants, not what the terminal givens them). I'll slap something together along these lines in a few days (unless you'd like to continue along that line).

PR #1430, opened 2024-10-19
chore: remove the last prelude refs / glob imports from example code

joshka:
Fixes: <https://github.com/ratatui/ratatui/issues/1150>

### DIFF
--- a/src/prelude.rs
+++ b/src/prelude.rs
@@ -1,15 +1,15 @@
 //! A prelude for conveniently writing applications using this library.
 //!
-//! We moved away from using the prelude module universally in Ratatui because it makes it harder to
-//! see which types are coming from the library and which are not and occasionally especially when
-//! reading source code in a non-IDE context (such as on GitHub or in a git diff). See [Issue #1150]
-//! for more information and links to user feedback. We have kept the prelude module in the library
-//! for backwards compatibility and for users who prefer to use it.
+//! The prelude module is no longer used universally in Ratatui, as it can make it harder to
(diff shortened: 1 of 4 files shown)


kdheepak (approved):

src/prelude.rs
```diff
@@ -1,5 +1,15 @@
...
+//! We moved away from using the prelude module universally in Ratatui because it makes it harder to
+//! see which types are coming from the library and which are not and occasionally especially when
+//! reading source code in a non-IDE context (such as on GitHub or in a git diff). See [Issue #1150]
+//! for more information and links to user feedback. We have kept the prelude module in the library
+//! for backwards compatibility and for users who prefer to use it.
+//!
+//! [Issue #1150]: https://github.com/ratatui/ratatui/issues/1150
+//!
+//! # Examples
+//!
```
kdheepak, line 12: I asked chatgpt to rewrite it without using the royal "we":

```suggestion
//! The prelude module is no longer used universally in Ratatui, as it can make it harder to
//! distinguish between library and non-library types, especially when viewing source code
//! outside of an IDE (such as on GitHub or in a git diff). For more details and user feedback, 
//! see [Issue #1150]. However, the prelude is still available for backward compatibility and for 
//! those who prefer to use it.
//!
//! [Issue #1150]: https://github.com/ratatui/ratatui/issues/1150
//!
//! # Examples
//! 
```

joshka: Thanks

PR #1431, opened 2024-10-19
docs(readme): reduce the length

joshka:


joshka:
Motivation for this is that there's a bunch of stuff at the bottom of the Readme that we don't really keep up to date. Instead it's better to link to the places that we do keep this info.

kdheepak (approved):

orhun (approved):

PR #1432, opened 2024-10-19
fix(text): truncate based on alignment

Lunderberg:
This is a follow-up PR to https://github.com/ratatui/ratatui/pull/987, which implemented alignment-aware truncation for the `Line` widget. However, the truncation only checked the `Line::alignment` field, and any alignment inherited from a parent's `Text::alignment` field would not be used.

This commit updates the truncation of `Line` to depend both on the individual `Line::alignment`, and on any alignment inherited from the parent's `Text::alignment`.

### DIFF
--- a/src/text/line.rs
+++ b/src/text/line.rs
@@ -687,7 +687,7 @@ impl Widget for Line<'_> {
 
 impl WidgetRef for Line<'_> {
     fn render_ref(&self, area: Rect, buf: &mut Buffer) {
-        self.render_with_alignment(area, buf, None)
+        self.render_with_alignment(area, buf, None);
     }
 }
 
--- a/src/text/line.rs
+++ b/src/text/line.rs
@@ -568,6 +568,53 @@ impl<'a> Line<'a> {
     pub fn push_span<T: Into<Span<'a>>>(&mut self, span: T) {
         self.spans.push(span.into());
     }
+
+    /// An internal implementation method for `WidgetRef::render_ref`
+    ///
+    /// Allows the parent widget to define a default alignment, to be
+    /// used if `Line::alignment` is `None`.
+    pub(crate) fn render_with_alignment(
+        &self,
+        area: Rect,
+        buf: &mut Buffer,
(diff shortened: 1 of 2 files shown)


src/text/line.rs
```diff
@@ -568,6 +568,53 @@ impl<'a> Line<'a> {
     pub fn push_span<T: Into<Span<'a>>>(&mut self, span: T) {
         self.spans.push(span.into());
     }
+
+    /// An internal implementation method for `WidgetRef::render_ref`
+    ///
+    /// Allows the parent widget to define a default alignment, to be
+    /// used if `Line::alignment` is `None`.
+    pub(crate) fn render_with_alignment(
```
joshka: nit: move this below where it's used so that code is easier to read top to bottom.

Lunderberg:
I couldn't find a clean way to provide the parent's `alignment` field to the `Line` widget through the `WidgetRef` interface, so I ended up moving the majority of `impl WidgetRef for Line<'_>` into a private method, which would be called by both `impl WidgetRef for Line<'_>` and `impl WidgetRef for Text<'_>`.

This also ended up simplifying `impl WidgetRef for Text<'_>`, since `Line` now has enough information to handle alignment that is inherited from the parent.

joshka (approved):
LGTM - made a minor refactor to move the render impl underneath where it's called from. This has the side-effect of making the diff easier to review too.

joshka:
Thanks for the PR!

src/text/line.rs
Lunderberg: Good call, and thank you for the change.  I had been grouping it with the other methods in the `impl Line` block, but on double-checking, those are all publicly-exposed methods.

joshka: I'm tempted to make a rust style lint for method ordering - it's a strong preference of mine that methods should be ordered from abstract to specific like this generally, but I understand that the opposite preference also exists (and so it's worth having both / making that configurable). Having it as a lint would save having to express and repeat the preference as a contributing thing, which would be nice.

PR #1434, opened 2024-10-19
chore(cargo): check in Cargo.lock

joshka:
When kept up to date, this makes it possible to build any git version
with the same versions of crates that were used for any version, without
it, you can only use the current versions. This makes bugs in semver
compatible code difficult to detect.

The Cargo.lock file is not used by downstream consumers of the crate, so
it is safe to include it in the repository (and recommended by the Rust
docs).

This helps with issue <https://github.com/ratatui/ratatui/issues/1433>.


orhun:
I'm not sure about this.

> and recommended by the Rust docs

Can you share some links to that? AFAIK libraries shouldn't commit `Cargo.lock` since it might cause problems with the crates.io package and so on. I need to dig up on my issue list to find an example of that.

joshka:
https://doc.rust-lang.org/cargo/faq.html#why-have-cargolock-in-version-control
https://blog.rust-lang.org/2023/08/29/committing-lockfiles.html

orhun (approved):

PR #1436, opened 2024-10-20
fix(color)!: hsl and hsluv are now clamped before conversion

joshka:
The `from_hsl` and `from_hsluv` functions now clamp the HSL and HSLuv
values before converting them to RGB. This ensures that the input values
are within the expected range before conversion.

Also note that the ranges of Saturation and Lightness values have been
aligned to be consisten with the palette crate. Saturation and Lightness
for `from_hsl` are now in the range [0.0..1.0] while `from_hsluv` are
in the range [0.0..100.0].

Refs:
- <https://github.com/Ogeon/palette/discussions/253>
- <https://docs.rs/palette/latest/palette/struct.Hsl.html>
- <https://docs.rs/palette/latest/palette/struct.Hsluv.html>

Fixes: <https://github.com/ratatui/ratatui/issues/1433>

### DIFF
--- a/src/style/color.rs
+++ b/src/style/color.rs
@@ -397,11 +397,11 @@ impl Color {
     ///
     /// // Minimum Saturation is fully desaturated red = gray
     /// let color: Color = Color::from_hsl(Hsl::new(0.0, 0.0, 0.5));
+    /// assert_eq!(color, Color::Rgb(128, 128, 128));
     ///
     /// // Maximum Saturation is fully saturated red
     /// let color: Color = Color::from_hsl(Hsl::new(0.0, 1.0, 0.5));
     /// assert_eq!(color, Color::Rgb(255, 0, 0));
-    /// assert_eq!(color, Color::Rgb(128, 128, 128));
     /// ```
     #[cfg(feature = "palette")]
     pub fn from_hsl(hsl: palette::Hsl) -> Self {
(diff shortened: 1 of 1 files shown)


src/style/color.rs
```diff
@@ -371,30 +371,42 @@ impl Color {
...
+    /// let color: Color = Color::from_hsl(Hsl::new(0.0, 0.0, 1.0));
     /// assert_eq!(color, Color::Rgb(255, 255, 255));
+    ///
+    /// // Minimum Saturation is fully desaturated red = gray
+    /// let color: Color = Color::from_hsl(Hsl::new(0.0, 0.0, 0.5));
+    /// assert_eq!(color, Color::Rgb(128, 128, 128));
+    ///
+    /// // Maximum Saturation is fully saturated red
+    /// let color: Color = Color::from_hsl(Hsl::new(0.0, 1.0, 0.5));
+    /// assert_eq!(color, Color::Rgb(255, 0, 0));
```
kdheepak, line 404: Can you add a doctest for negative Hue?

kdheepak (approved):

kdheepak:
A good first issue would be to make a HSL <-> RGB tui app.

kdheepak (approved):

src/style/color.rs
joshka: yep

joshka:
> A good first issue would be to make a HSL <-> RGB tui app.

It's probably worth doing a real app for any color format conversion instead. I'm not sure that writing an app that demonstrates a conversion seems all that necessary.

Issue #1440, opened 2024-10-20
Implement Rect::resize similar to Rect::offset


kardwen:
## Problem

`Rect::offset` can be used in a Builder Lite pattern to modify a `Rect`. 
It has the advantage over direct manipulation of `Rect.x` and `Rect.y` that it prevents an integer overflow.
Maybe introducing `Rect::resize` would offer similar benefits?
What I don't like is that it would require to introduce a new `Offset`-like type, for example ~`SizeOffset`~ `ResizeOffset`:

```Rust
pub struct ResizeOffset {
    pub width: i32,
    pub height: i32,
}
```

Edit:

Here is a possible use case:

```Rust
let area = Rect::new(0, 0, 20, 30)
    .offset(Offset { x: 0, y: 4 })
    .resize(ResizeOffset {
        width: -5,
        height: 5,
    });
```

## Solution

I would like to implement it myself, as it seems doable as a first issue, if it fits in.

## Alternatives

What I did so far, manual manipulation of `width` and `height` fields of `Rect`

## Additional context

<https://docs.rs/ratatui/latest/ratatui/layout/struct.Rect.html#method.offset>

joshka:
It's worth checking back on issues about what we discussed on this. I can't recall if we've previously done so.
Would `Size` work for the param here?

kardwen:
`Size` wouldn't work because its fields are defined as `u16` thus not allowing for negative values for resizing (shrinking a `Rect`). It would be nice to evaluate if there is a common use case for this as I can't really tell yet.

joshka:
Oh, I think you're saying that you wanted to do relative resize. I missed that. Yeah, you'd likely need a new type. But something that might also be a way to do this is implement various ops (e.g. `ops:Add<Size> for Rect` => `let rect = Rect::new(1,2,3,4) + Size(5,6)` is now possible.

kardwen:
Interesting, I didn't know that I can also do additions with `Rect`

kardwen:
This is what I had in mind:

```Rust
let area = Rect::new(0, 0, 20, 30)
    .offset(Offset { x: 0, y: 4 })
    .resize(SizeOffset {
        width: -5,
        height: 5,
    });
```

Maybe an associated function `new` for it would also be nice.

kardwen:
Now, that I look at it in more detail I noticed that `Offset` could be reused, or not? The fields are named different, but it would still be a resize operation in `x` for `width` or `y` for `height` respectively.

[...]

PR #1442, opened 2024-10-20
feat(layout): Add rect::resize and layout::ResizeOffset

kardwen:
Discussion in #1440

Hi, this is my first pull request, I hope I did not forget anything!


kardwen:
I could use some help with the semver report and a general review of the changes made.

joshka:
The semver report is fine (and expected). It indicates that we have semver breaking changes in the main branch (these will be released pretty soon)

src/layout/rect.rs
```diff
@@ -462,6 +499,84 @@ mod tests {
         );
     }
 
+    #[test]
```
joshka, line 502: Consider using rstest for the tests to reduce the amount of vertical space. Less code to read without any complexity cost. e.g.:

```rust
#[rstest]
#[case::positiveRect::new(1, 2, 3, 4), ResizeOffset::new(5, 6), Rect::new(1, 2, 8, 10)]
#[case::negative(Rect::new(1, 2, 3, 4), ResizeOffset::new(-2, -3), Rect::new(1, 2, 1, 2)]
// ...
fn resize(#[case] rect: Rect, #[case] resize: ResizeOffset, #[case] expected) {
    assert_eq(rect.resize(resize), expected);
}
```

joshka, line 502: Can you please add a test for i32::MIN and i32::MAX? There's likely no bug, but this prevents any math errors in any future implementation changes.

src/layout/rect.rs
```diff
@@ -166,6 +180,29 @@ impl Rect {
...
+    /// Resizes the `Rect` according to the given offset without modifying its [`x`](Rect::x)
+    /// or [`y`](Rect::y) coordinates.
+    /// - Positive `width` increases the extent of the `Rect` along the x-axis, negative `width`
+    ///   decreases it.
+    /// - Positive `height` increases the extent of the `Rect` along the y-axis, negative `height`
+    ///   decreases it.
+    ///
+    /// See [`ResizeOffset`] for details.
+    #[must_use = "method returns the modified value"]
+    pub fn resize(self, resize_offset: ResizeOffset) -> Self {
```
joshka, line 194: It's possible this make sense named "resized" rather than "resize" as the latter implies an action (e.g. a setter), where-as the former implies a non-modifying method that returns a new value. I'm not 100% sure on this, so it's worth looking at - what do we do that's consistent and what do other libraries that do this do for similar behavior. (To be clear, this is not a request to make this change, but more a request to think about it and push back if I'm wrong. I think this would be fine as `resize` if it's consistent with other things and we never want to have `Rect::resize(&mut self, resize_offset) {}`)

joshka (approved):
Generally LGTM. Approved, but there's some small improvements that could be made and a question to answer (which might be fine to move to the issue).

src/layout/rect.rs
kardwen: Thanks, that was new to me

kardwen: Added two test, that should cover the `ResizeOffset::new` function as well

src/layout/rect.rs
kardwen: I don't know but I also had the new idea to name it
`Rect::shift` and `layout::PositionDelta`, `Rect::resize` and `layout::SizeDelta`:

```Rust
let area = Rect::new(0, 0, 20, 30)
    .shift(PositionDelta { x: 0, y: 4 })
    .resize(SizeDelta {
        width: -5,
        height: 5,
    });
```

delta has the advantage that it just means difference while I associate offset more with a position than a size, so we could just keep `Offset` (but still rename `offset` as function name because it is imprecise):

```Rust
let area = Rect::new(0, 0, 20, 30)
    .shift(Offset { x: 0, y: 4 })
    .resize(SizeDelta {
        width: -5,
        height: 5,
    });
```

I am not sure what I prefer, the second option because it is concise or the first because it relates both types to each other.

orhun (approved):
LGTM!

src/layout/rect.rs
kardwen: I also think so. With 0.29 just out there is some time anyways until the next tagged release.

kardwen:
Only did a rebase on current main

ratatui/src/layout/rect.rs
```diff
@@ -166,6 +187,29 @@ impl Rect {
...
+    /// Resizes the `Rect` by a given delta without modifying its [`x`](Rect::x)
+    /// or [`y`](Rect::y) coordinates.
+    /// - Positive `width` increases the extent of the `Rect` along the x-axis, negative `width`
+    ///   decreases it.
+    /// - Positive `height` increases the extent of the `Rect` along the y-axis, negative `height`
+    ///   decreases it.
+    ///
+    /// See [`SizeDelta`] for details.
+    #[must_use = "method returns the modified value"]
+    pub fn resize_by(self, delta: SizeDelta) -> Self {
```
joshka, line 201: Can you please add a note about the expected behavior when clamping?
Something like "The size is clamped to ensure that the right and bottom bounds of the rect are less than u16::Max" (Feel free to wordsmith that if you want)

joshka (approved):
LGTM - just a small comment about docs.

kardwen:
I have deprecated `Rect::offset'. Maybe we should add a note with which version it will be removed.

ratatui/src/layout/rect.rs
kardwen: I added a note

kardwen:
I noticed that `cargo make format` formatted my test macros in unexpected ways, maybe there is some additional configuration needed?

kardwen:
I wasn't sure if had to create a new pull request for the second commit. My reasoning for adding it here was that the commits are related. Also the pull request title looks so convoluted now.

joshka:
> I wasn't sure if had to create a new pull request for the second commit. My reasoning for adding it here was that the commits are related. Also the pull request title looks so convoluted now.

We use GitHub to squash merge PRs, and then use the PR message verbatim in our changelog. This means if you want two Changelog entries then it's best to split the PR into two. It's generally fine to create a stacked PR that also has a commit from another PR when the second PR is dependent on the first as we'll usually merge the first and then the shared commit disappears from the second.

We could avoid this by doing manual changelog entries, but for now the 1:1 mapping of PRs to changelog entries serves us well enough as it's all automated. It does however put some constraints on how we arrange the PRs.

I'll take another look at the changes in a bit.

kardwen:
> We use GitHub to squash merge PRs, and then use the PR message verbatim in our changelog. This means if you want two Changelog entries then it's best to split the PR into two.

kardwen:
I just now had the case that I want to get the size of a table when it is rendered, where I could use the new feature:

```Rust
        let inner_area = block.inner(area);
        self.content_area = Some(Rect {
            x: inner_area.x,
            y: inner_area.y + 1,
            width: inner_area.width,
            height: inner_area.height.saturating_sub(1),
        });
        // self.content_area = Some(
        //     block
        //         .inner(area)
        //         .move_by(Offset::new(0, 1))
        //         .resize_by(SizeDelta::new(0, -1)),
        // );
```

My current code looks a little bit ugly and also requires some care to prevent an overflow.

Please tell me, if I should split this pull request into two parts.

kardwen:
https://docs.rs/ratatui/latest/ratatui/layout/struct.Layout.html#method.margin

I find this API especially confusing, since it doesn't use `layout::Margin`, which is only used in `Rect::inner`.

For example, when I want to create a `Layout` with only a top margin I can't do this with `Layout::horizontal_margin`. Directly changing the area that I pass to `Layout::split` with `resize_by` and `move_by` seems a lot more straight forward to me.

joshka:
Apologies for reply latency on this one. It's in my bucket of things to look at after the many internal refactoring things we're doing this release. Thanks for being patient while we do that.

ratatui-core/src/layout/position.rs
```diff
@@ -25,30 +25,32 @@ use crate::layout::Rect;
...
-    pub const fn new(x: u16, y: u16) -> Self {
+    pub const fn new(x: T, y: T) -> Self {
         Self { x, y }
     }
 }
 
+impl Position<u16> {
+    /// Position at the origin, the top left edge at 0,0
+    pub const ORIGIN: Self = Self { x: 0, y: 0 };
+}
```
joshka, line 52: TODO: investigate pulling in https://crates.io/crates/num-traits to make Zero a constraint on T and to be able to define this generically rather than only for u16. (and below for Size)

joshka:
I rebased this on main, and then when looking at it had an idea which might be a better one generally. Instead of adding more types to represent size / position delta, we can instead change the Size and Position so that it is generic over the integer type. By default that means u16, which makes all existing code just work (modulo code which infers the x and y values and doesn't otherwise constrain them).  Then, SizeDelta becomes Size<i32> and Offset becomes Position<i32>. I added a type alias for Offset to ensure that existing code is not broken, but used Size<i32> for the resize changes.

This is a bit of a fundamental change and likely should be called out in the changelog as it may affect more users (we should validate this). It probably requires a little more docs to be added about the type parameter at the very least.

@kardwen do you have any thoughts on this approach? I know it's not quite what you had in mind, but does it make sense as a user?

kardwen:
I like the idea to simplify the types and think that using generics to indicate that `Size` or `Position` is used as a relative value (what I called delta) seems pretty elegant. I miss some overview to evaluate if this might cause problems elsewhere but like it so far.

j-g00da:
Can this be closed? I see that this is continued with alternative approach on #1596.
@joshka

orhun:
I think we should finalize #1596 and then we can close this one too

orhun:
closing it to continue in #1596

PR #1447, opened 2024-10-22
fix(buffer): buffer::get_pos() now correctly handles index > u16::MAX

joshka:
Previously this function wrapped the index pass u16::MAX which caused
problems rendering.

Fixes: <https://github.com/ratatui/ratatui/issues/1441>

### DIFF
--- a/src/style/color.rs
+++ b/src/style/color.rs
@@ -480,6 +480,7 @@ mod tests {
 
     #[cfg(feature = "palette")]
     use palette::{Hsl, Hsluv};
+    #[cfg(feature = "palette")]
     use rstest::rstest;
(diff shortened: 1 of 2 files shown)


joshka:
image

src/style/color.rs
```diff
@@ -480,6 +480,7 @@ mod tests {
 
     #[cfg(feature = "palette")]
     use palette::{Hsl, Hsluv};
+    #[cfg(feature = "palette")]
```
joshka, line 483: unrelated - but clippy was complaining when the palette feature is not enabled. not worth a separate PR to fix.

kdheepak (approved):

PR #1452, opened 2024-10-26
fix(table): Allow display of additional table row, if row height > 1

Lunderberg:
Prior to this commit, a table would only display rows whose height could entirely fit within the available area.  As a result, table rows with `row.height > 1`, which require multiple terminal rows to display, could result in empty rows being rendered at the bottom of the table's display area.

This commit updates the rendering of `Table` to also include rows that may be partially visible within the available area.  That is, even if the display area does not have sufficient height to fully display an additional row of the table, it may still display the first few lines of an additional table row.

This change does not affect the change of the `offset`, which adjusts the viewing bounds to ensure that the selected row is displayed.  This is deliberate, since the view should be scrolled to allow the selected row to be displayed in its entirety.

### DIFF
--- a/ratatui-widgets/src/table.rs
+++ b/ratatui-widgets/src/table.rs
@@ -1016,7 +1016,7 @@ impl Table<'_> {
             }
         }
 
-        // If the table contains rows with a height greater than one, then the last row of the
+        // If the table contains rows with a height greater than one, then an additional row of the
         // table may be partially displayed.
         if height < max_height && end < self.rows.len() {
             end += 1;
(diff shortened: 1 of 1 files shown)


Lunderberg:
Converting to a draft, as the test case doesn't seem to catch the error that I'm seeing in my use case.

Lunderberg:
And ready for review.  In order to trigger the bug, the test case needed to have a height that would overflow the available area of the table.

orhun (approved):
The logic seems fine and fixes the reproducible test case. LGTM.

Lunderberg:
Merged main into the PR branch to resolve conflicts with https://github.com/ratatui/ratatui/pull/1459.  I'm guessing there will be additional conflicts when https://github.com/ratatui/ratatui/pull/1460 lands, and can resolve at that point.

joshka:
> Merged main into the PR branch to resolve conflicts with #1459. I'm guessing there will be additional conflicts when #1460 lands, and can resolve at that point.

Yeah - I'm thinking that #1460 takes precedence and then in flight PRs get rebased so that they line up. There might be some delay in landing PRs due to this, but it's easier this way than the opposite.

Lunderberg:
> Yeah - I'm thinking that #1460 takes precedence and then in flight PRs get rebased so that they line up. There might be some delay in landing PRs due to this, but it's easier this way than the opposite.

Definitely agreed.  Splitting up components is a tricky enough task already without the additional complexity of the main branch changing from underneath you.

joshka (approved):

Issue #1449, opened 2024-10-22
Canvas rendering issue when area is huge


JeromeSchmied:
## Description

After a certain point, when characters are teeny-tiny, and therefore `ratatui::widgets::Canvas` becomes extremely huge(around 1000*1500), it doesn't properly fill the frame it's given.

## To Reproduce

1. on [latest commit of main](860e48b0f0ec51c8596bc12985a8b37bad4fba00)
2. `cargo r --release --example canvas`
3. minify font-size (eg <kbd>Ctrl-</kbd>)

## Expected behavior

should continue to fill it's bigger frame

### small font-size, still rendered correctly

![Image](https://github.com/user-attachments/assets/5cd21b28-3cd9-4aa9-80ff-7adcf1b36546)

### even smaller font-size, not rendered correctly

![Image](https://github.com/user-attachments/assets/09492608-38ec-4a8b-a374-1d38d190681b)
 

## Environment

- OS: Arch Linux x86_64
- Terminal Emulator: alacritty 0.13.2
- Font: Recursive (patched)
- Crate version: [latest commit of main](860e48b0f0ec51c8596bc12985a8b37bad4fba00)
- Backend: crossterm

## Additional context

> [!NOTE]
> in debug mode it panics with the following logs 

```log
ratatui (main)$ RUST_BACKTRACE=full cargo r --example canvas
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.11s
     Running `target/debug/examples/canvas`
The application panicked (crashed).
Message:  attempt to multiply with overflow
Location: src/widgets/canvas.rs:126

  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ BACKTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
                                ⋮ 10 frames hidden ⋮
  11: core::panicking::panic_const::panic_const_mul_overflow::h458a35cd46a46d6c
      at /rustc/f6e511eec7342f59a25f7c0534f1dbea00d01b14/library/core/src/panicking.rs:181
  12: ratatui::widgets::canvas::BrailleGrid::new::h8a3bd5247fce3104
      at /home/jero/code/rust/cloned/ratatui/src/widgets/canvas.rs:126
       124 │     /// rows respectively.
       125 │     fn new(width: u16, height: u16) -> Self {
       126 >         let length = usize::from(width * height);
       127 │         Self {
       128 │             width,
[...]

PR #1457, opened 2024-10-30
fix(buffer): avoid overflow while getting the position of indexes

orhun:
This commit fixes an unsafe conversion issue in the `pos_of` method of Buffer where the final position is silently truncated if an index larger than u16::MAX is provided. It was due to the unchecked cast using "as u16" and it has been replaced with a proper u16::try_from call.

`pos_of` now panics if the given position is still not inside the available calculation range, which shouldn't be a problem in most of terminals.

The newly added tests are failing without this change (you can see that the positions are being truncated):

```
thread 'buffer::buffer::tests::pos_of_last_index_does_not_overflow::case_1_bottom_left' panicked at src/buffer/buffer.rs:1250:9:
  left: (16, 3274)
 right: (0, 65534)
```

```
thread 'buffer::buffer::tests::pos_of_last_index_does_not_overflow::case_2_bottom_right' panicked at src/buffer/buffer.rs:1250:9:
  left: (15, 3275)
 right: (19, 65534)
```

Related to #1449
Coded on livestream: https://www.youtube.com/live/jiTtSmdRNd4

### DIFF
--- a/src/buffer/buffer.rs
+++ b/src/buffer/buffer.rs
@@ -309,10 +309,17 @@ impl Buffer {
             "Trying to get the coords of a cell outside the buffer: i={i} len={}",
             self.content.len()
         );
-        (
-            self.area.x + (i as u16) % self.area.width,
-            self.area.y + (i as u16) / self.area.width,
-        )
+        if let (Ok(x_offset), Ok(y_offset)) = (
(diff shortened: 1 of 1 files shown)


orhun:
It seems like I followed a bit behind, it's already fixed in main :) It was a good deep-dive experience to do this live anyways.

PR #1460, opened 2024-10-31
chore: move core types to ratatui-core

joshka:
The buffer, layout, style, symbols, text, and the top level of widgets
modules are moved to ratatui-core. This is the first step in
modularizing the library so that the core types can be used in other
projects without the need for the backend / widgets types.

This helps reduce the need for updating other crates as often due to
semver changes outside of the core types.


joshka:
rebased

joshka:
I think it might be worth pulling the breaking changes for `IntoCrossterm` etc. into their own PR to simplifiy this change to be more of a pure refactoring style change.

ratatui-core/Cargo.toml
```diff
@@ -0,0 +1,70 @@
+[package]
+name = "ratatui-core"
+description = """
+    Core library for the Ratatui terminal UI framework. This crate contains the core types and
+    traits that are used to build terminal user interfaces. Most applications should use the
+    `ratatui`, while widget libraries should use this crate."""
```
orhun, line 6: We should have a shorter description here and move the explanation to the README.md in `ratatui-core`. What do you think?

ratatui/src/backend/termwiz.rs
```diff
@@ -194,7 +194,7 @@ impl Backend for TermwizBackend {
 
     fn get_cursor_position(&mut self) -> io::Result<crate::layout::Position> {
         let (x, y) = self.buffered_terminal.cursor_position();
-        Ok((x as u16, y as u16).into())
+        Ok(crate::layout::Position::new(x as u16, y as u16))
```
orhun, line 197: Let's implicitly import this at the top of the module.

```suggestion
        Ok(Position::new(x as u16, y as u16))
```

orhun (approved):
Good stuff

ratatui-core/Cargo.toml
joshka: This shows up in search results, Ithink that should include the note about widget libs. Eventually this lib will be the second highest ratatui lib to show up by downloads due to being always included when downloading Ratatui.

This definitely does need a README too though.

ratatui/src/backend/termwiz.rs
joshka: 5e2525167..451790c72

ratatui-core/Cargo.toml
orhun: How about making it a bit shorter?

```suggestion
description = """
Core library for the Ratatui terminal UI framework, containing essential types and traits for building terminal UIs. Most applications should use the `ratatui`, while widget libraries should use this crate."""
```

joshka: Simplfied in 8709cdb7d..a8ff86830

Stylistically, I avoid using backticks around crate names unless actually pointing at the crate's module (This is mostly consistent in Ratatui, but there's a few places where that isn't the case).

PR #1461, opened 2024-11-01
ci: replace cargo-make with a custom cargo-xtask

joshka:
This removes the need for cargo-make and replaces it with a custom xtask
binary. See <https://github.com/matklad/cargo-xtask> for info.

Rearranges the CI workflow to use the new xtask and simplify which
workflows that run.


joshka:
image

joshka:
There's a good CI time reduction from this due to lowering the number of jobs from 39 to 31. The latest CI run succeeded in 6mins with another run also running simultaneously (which throttles the number of runners I think). Previously runs were taking 8-15 mins.

.github/workflows/ci.yml
```diff
@@ -114,47 +128,44 @@ jobs:
...
-          toolchain: ${{ matrix.toolchain }}
-      - uses: taiki-e/install-action@v2
-        with:
-          tool: cargo-make,nextest
+      - uses: dtolnay/rust-toolchain@stable
       - uses: Swatinem/rust-cache@v2
-      - run: cargo make test-backend ${{ matrix.backend }}
-        env:
-          RUST_BACKTRACE: full
+      - run: cargo xtask test-backend ${{ matrix.backend }}
```
joshka, line 171: Note: previously this was running each backend on each OS with MSRV and stable toolchains. (3x3x2 - 2 = 16 tests). This seems excessive. I've moved the test that tests all features up and then these just test with OS/backend combinations (3x3 - 1 = 8). There's an error there though that I should have added a matrix to the main lib test on ubuntu.

xtask/src/main.rs
```diff
@@ -0,0 +1,329 @@
...
+        .without_time()
+        .init();
+
+    match args.run() {
+        Ok(_) => (),
+        Err(err) => {
+            tracing::error!("{err}");
+            std::process::exit(1);
+        }
+    }
```
joshka, line 25: I originally had this as just falling through and letting color_eyre printt the error, but changes to this for a little nicer output. It may be worth considering changing back at some point.

orhun:
Does doing this have any workspace-related advantages other than speeding up the CI? I haven't worked with `xtask` before so I'm trying to understand the benefits. I'd be happy if you can provide more information (maybe in the commit summary) why this is good to have.

joshka:
> Does doing this have any workspace-related advantages other than speeding up the CI? I haven't worked with `xtask` before so I'm trying to understand the benefits. I'd be happy if you can provide more information (maybe in the commit summary) why this is better than `cargo-make`.

Regularly when I run cargo make, it tells me there's an update available. That was the bit that annoyed me enough to do this.
I've been thinking about this for some time, and the rationale is mentioned in the link:

> -   It doesn't require any other binaries besides cargo and rustc, it fully bootstraps from them
> -   Unlike bash, it can more easily be cross platform, as it doesn't use the shell.

By using an xtask approach, all the CI steps that previously needed cargo-make already have the necessary tools they need to run installed (and these are likely cached between runs too).

My secondary rationale is a general distaste for implementing imperative logic (here instructions on what and how to build) in declarative languages (here toml) or in shell scripts. The code for detecting the default workspace members to run the `cargo-docs-rs` packages would have been difficult to get right, and would have tended to just be hardcoded instead of letting the existing workspace metadata define what is built. I just like looking at rust for this. It looks pretty simple and obvious.

I also wanted something that would be easy to extend into watching for changes and checking that things are ready for PR in the background (to save having to make small clippy / typos / formatting issue updates), and which could our existing shell scripts (for releases and vhs generation)

Lastly I wanted to experiment with the approach, as it seems like a neat idea generally.

I don't think much of this really belongs in the changelog, which IMO should really contain info about user focused things and less internal things like this.

orhun:
Fair enough.

orhun (approved):

Issue #1388, opened 2024-09-27
Modularize Ratatui crate


joshka:
## Problem
It would be nice to be able for widget crates to rely just on the widget / buffer / etc. types instead of the whole library.
Additionally by extracting a widget crate out of the library we might be able to have quicker incremental build times.
It would also be nice to be able to release non-breaking widget updates in point releases immediately rather than waiting for a library release.

## Solution
Decompose Ratatui into several crates:
- ratatui: the main core library, feature flags to include the parts
- ratatui-core: Widget, WidgetRef, StatefulWidget, StatefulWidgetRef, Terminal, Frame, Backend etc.
- ratatui-{crossterm,termion,termwiz}: backends
- ratatui-widgets: all the existing built-in widgets (currently this is something else, but I'd move the current library to a new name)

It's possible that we could also go more granular than this and have a crate per widget, but I think it's worth doing the first part and evaluating whether doing more is worthwhile.

This will be implemented by making the repo into a cargo workspace, moving the current source under /ratatui/src, and creating several more crates at the root. I know it may be tempting to not do a monorepo for this, but the administrative overhead of having a repo per crate seem prohibitive, while I've had reasonable success using mono-repos (e.g. tui-widgets crate).

The work to do this will be:
- make the actual refactoring changes
- make the CI / CD changes
- make the doc changes (fixup website links etc. to examples)
- test changes in various widget libraries

@ratatui/maintainers what do you think?

# Plan

- [x] move the folder, create the workspace #1459 
- [x] move the main types to ratatui-core #1460 
- [x] Move widgets into ratatui-widgets https://github.com/ratatui/ratatui/pull/1474
- [x] https://github.com/ratatui/ratatui/pull/1479
- [x] https://github.com/ratatui/ratatui/pull/1490
- [x] https://github.com/ratatui/ratatui/pull/1491
- [x] #1508 
- [x] Move examples into /examples - there's a good approach that tokio uses for this where each example gets a crate, and so can have their own dependencies listed etc rather than these being all smooshed into a single dev-dependencies lists.
- [x] Update / make sure that git cliff and release automation are working fine
- [ ] Make an alpha release and then test ratatui-core with some widget libraries
- [x] decide on whether to move frame / terminal to core, or leave them in Ratatui (I'm not certain about this right now, probably should go to core, but there's an argument against it too

joshka:
# Blockers

I'm attempting a PoC of this, and documenting issues that come up

[...]

PR #1462, opened 2024-11-01
docs: fix example link in readme

thomas-tacquet:
README.md contains a link to unexisting example directory like it was before on root bu nowt it's in `ratatui` directory. 

I'm new to this repo this is probably due to a recent change.


joshka (approved):

joshka:
Thanks. I missed that one in a recent change. We're in the process of splitting out the ratatui main crate into a few different crates to make things better for widget crates. See #1388 for info.

PR #1465, opened 2024-11-02
fix(canvas): Draw default-colored blocks with Marker::HalfBlock

Lunderberg:
Prior to this commit, the `HalfBlockGrid` used `Color::Reset` to identify pixels that did not have anything drawn in them.  As a result, elements drawn with `Color::Reset` would be skipped.  This behavior is inconsistent with other implementations of `widgets::canvas::Grid`, which draw elements of `Color::Reset` using the default foreground color.

This commit updates the implementation of `HalfBlockGrid` to store pixels of `Option<Color>`, rather than `Color`.  This allows the implementation of `Grid::save` for `HalfBlockGrid` to distinguish between cells that have not been painted (`None`) and cells that have been painted with the default foreground color (`Some(Color::Reset)`).

### DIFF
--- a/ratatui/src/widgets/canvas.rs
+++ b/ratatui/src/widgets/canvas.rs
@@ -246,7 +246,7 @@ struct HalfBlockGrid {
     /// Height of the grid in number of terminal rows
     height: u16,
     /// Represents a single color for each "pixel" arranged in column, row order
-    pixels: Vec<Vec<Color>>,
+    pixels: Vec<Vec<Option<Color>>>,
 }
 
 impl HalfBlockGrid {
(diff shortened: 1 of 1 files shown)


Lunderberg:
This PR is based on the current `main` branch, which currently includes the workspace migration from https://github.com/ratatui/ratatui/pull/1459, but not the modularization from https://github.com/ratatui/ratatui/pull/1460.

Lunderberg:
Closing this PR in favor of https://github.com/ratatui/ratatui/pull/1466, which also requires updates to `canvas.rs` and includes this fix.

PR #1466, opened 2024-11-02
feat(chart): Allow Braille/Dot charts to overlap with Block

Lunderberg:
The `Chart` widget allows multiple line/scatter plots to be drawn on the same axis.  Prior to this commit, for a given tile of the canvas, the output was entirely determined by the top-most line plot to touch that tile.

In most cases, two overlapping lines cannot be rendered on top of each other.  For example, drawing `Marker::Braille` on top of `Marker::Dot` requires replacing the character being drawn.  However, `Marker::Block` is an exception, as a block may be accurately represented as either `symbols::block::FULL` in the foreground, or with a background color.

This commit updates the rendering of the `Chart` widget to allow two line/scatter plots to overlap, where the first chart uses `Marker::Block`, and the second chart uses either `Marker::Braille` or `Marker::Dot`.

### DIFF
--- a/ratatui/src/widgets/canvas.rs
+++ b/ratatui/src/widgets/canvas.rs
@@ -64,11 +64,22 @@ pub struct Label<'a> {
 /// multiple shapes on the canvas in specific order.
 #[derive(Debug)]
 struct Layer {
-    // A string of characters representing the grid. This will be wrapped to the width of the grid
-    // when rendering
-    string: String,
-    // Colors for foreground and background of each cell
-    colors: Vec<(Color, Color)>,
+    contents: Vec<LayerCell>,
+}
+
+/// A cell within a layer.
+///
+/// If a Context contains multiple layers, then the symbol,
+/// foreground, and background colors for a character will be
+/// determined by the top-most layer that provides a value for that
+/// character.  For example, a chart drawn with Marker::Block may
+/// provide the background color, and a later chart drawn with
+/// Marker::Braille may provide the symbol and foreground color.
+#[derive(Debug)]
+struct LayerCell {
+    symbol: Option<char>,
+    fg: Option<Color>,
+    bg: Option<Color>,
 }
 
 /// A grid of cells that can be painted on.
@@ -116,7 +127,7 @@ struct BrailleGrid {
     utf16_code_points: Vec<u16>,
     /// The color of each cell only supports foreground colors for now as there's no way to
     /// individually set the background color of each dot in the braille pattern.
-    colors: Vec<Color>,
+    colors: Vec<Option<Color>>,
 }
 
 impl BrailleGrid {
@@ -128,7 +139,7 @@ impl BrailleGrid {
             width,
             height,
             utf16_code_points: vec![symbols::braille::BLANK; length],
-            colors: vec![Color::Reset; length],
+            colors: vec![None; length],
         }
     }
 }
@@ -139,15 +150,31 @@ impl Grid for BrailleGrid {
     }
 
     fn save(&self) -> Layer {
-        let string = String::from_utf16(&self.utf16_code_points).unwrap();
-        // the background color is always reset for braille patterns
-        let colors = self.colors.iter().map(|c| (*c, Color::Reset)).collect();
-        Layer { string, colors }
+        let contents = self
+            .utf16_code_points
+            .iter()
+            .zip(&self.colors)
+            .map(|(&code_point, &color)| {
+                let symbol = if code_point == symbols::braille::BLANK {
+                    None
+                } else {
+                    Some(char::from_u32(code_point as u32).unwrap())
+                };
+
+                LayerCell {
+                    symbol,
+                    fg: color,
+                    bg: None,
+                }
+            })
+            .collect();
+
+        Layer { contents }
     }
 
     fn reset(&mut self) {
         self.utf16_code_points.fill(symbols::braille::BLANK);
-        self.colors.fill(Color::Reset);
+        self.colors.fill(None);
     }
 
     fn paint(&mut self, x: usize, y: usize, color: Color) {
@@ -158,7 +185,7 @@ impl Grid for BrailleGrid {
             *c |= symbols::braille::DOTS[y % 4][x % 2];
         }
         if let Some(c) = self.colors.get_mut(index) {
-            *c = color;
+            *c = Some(color);
         }
     }
 }
@@ -173,12 +200,18 @@ struct CharGrid {
     width: u16,
     /// Height of the grid in number of terminal rows
     height: u16,
-    /// Represents a single character for each cell
-    cells: Vec<char>,
     /// The color of each cell
-    colors: Vec<Color>,
+    cells: Vec<Option<Color>>,
+
     /// The character to use for every cell - e.g. a block, dot, etc.
     cell_char: char,
+
+    /// If true, apply the color to the background as well as the
+    /// foreground.  This is used for Marker::Block, so that it will
+    /// overwrite any previous foreground character, but also leave a
+    /// background that can be overlaid with an additional foreground
+    /// character.
+    apply_color_to_bg: bool,
 }
 
 impl CharGrid {
@@ -189,9 +222,16 @@ impl CharGrid {
         Self {
             width,
             height,
-            cells: vec![' '; length],
-            colors: vec![Color::Reset; length],
+            cells: vec![None; length],
             cell_char,
+            apply_color_to_bg: false,
+        }
+    }
+
+    fn apply_color_to_bg(self) -> Self {
+        Self {
+            apply_color_to_bg: true,
+            ..self
         }
     }
 }
@@ -203,14 +243,20 @@ impl Grid for CharGrid {
 
     fn save(&self) -> Layer {
         Layer {
-            string: self.cells.iter().collect(),
-            colors: self.colors.iter().map(|c| (*c, Color::Reset)).collect(),
+            contents: self
+                .cells
+                .iter()
+                .map(|&color| LayerCell {
+                    symbol: color.map(|_| self.cell_char),
+                    fg: color,
+                    bg: color.filter(|_| self.apply_color_to_bg),
+                })
+                .collect(),
         }
     }
 
     fn reset(&mut self) {
-        self.cells.fill(' ');
-        self.colors.fill(Color::Reset);
+        self.cells.fill(None);
     }
 
     fn paint(&mut self, x: usize, y: usize, color: Color) {
@@ -218,10 +264,7 @@ impl Grid for CharGrid {
         // using get_mut here because we are indexing the vector with usize values
         // and we want to make sure we don't panic if the index is out of bounds
         if let Some(c) = self.cells.get_mut(index) {
-            *c = self.cell_char;
-        }
-        if let Some(c) = self.colors.get_mut(index) {
-            *c = color;
+            *c = Some(color);
         }
     }
 }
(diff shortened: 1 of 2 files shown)


Lunderberg:
As an example of the visual effect, screenshots from an interface I'm working on.

Before, where the red `Marker::Braille` and the dark green `Marker::Block` overlap, the dark green disappears entirely:
image


After, where the red `Marker::Braille` and the dark green `Marker::Block` overlap, the dots are drawn on top of the dark green background:
image

joshka:
Neat idea and effect. I may take me a bit of time to dig into the code change for this one.

A small question though, I wonder whether this makes sense as the default, or if it should instead be an option that can be configured? What do you think?

Lunderberg:
> Neat idea and effect. I may take me a bit of time to dig into the code change for this one.

Thank you!  I realized looking at it that there's a larger delta than I had expected.

* [Core functionality change](https://github.com/ratatui/ratatui/pull/1466/files#diff-f354382f44c635aca371c68851a4175e4a8601a7a7a8c0dd698101896434a164R853-R860), to have the symbol, foreground and background each apply separately.
* [Usage of new functionality](https://github.com/ratatui/ratatui/pull/1466/files#diff-f354382f44c635aca371c68851a4175e4a8601a7a7a8c0dd698101896434a164R549), where the `apply_color_to_bg` functionality is only applied for the `Marker::Block` style.
* [Moving](https://github.com/ratatui/ratatui/pull/1466/files#diff-c30ed913844904c34a8845f838d8b9aa5c723cd622963d73c8ae42f945694d91R1021-R1026) the `for dataset in &self.datasets` inside the `.paint` function causes overlapping charts to be resolved within the chart's `Canvas`, rather than within the `Buffer` as a whole.

All other changes are to keep the same user-facing functionality within the updated canvas.

> A small question though, I wonder whether this makes sense as the default, or if it should instead be an option that can be configured? What do you think?

For the core functionality, separating out the symbol/fg/bg, I'd lean to have it enabled unconditionally, mainly because all the existing user-facing functionality can be implemented with symbol/fg/bg potentially coming from separate layers.  It would allow for additional marker types (e.g. a `Marker::Highlight` that would change the foreground color to highlight something, without changing the symbol) that can't be expressed with the current `Layer` class.

For the user-facing functionality, I was going back and forth on whether it should be a separate `Marker` variant (e.g. `Marker::Background`) rather than a special behavior of `Marker::Block`.  I ended up having it be part of `Marker::Block`, because it would have no visual difference when the plots are non-overlapping, and the change to overlapping commands is limited to the `Canvas` widget.

So I think it would make sense as a default behavior for `Marker::Block`, but only because it occurs within a context used for drawing images rather than displaying text.

ratatui-widgets/src/canvas.rs
```diff
@@ -66,11 +66,22 @@ pub struct Label<'a> {
...
+}
+
+/// A cell within a layer.
+///
+/// If a Context contains multiple layers, then the symbol,
+/// foreground, and background colors for a character will be
+/// determined by the top-most layer that provides a value for that
+/// character.  For example, a chart drawn with `Marker::Block` may
+/// provide the background color, and a later chart drawn with
+/// `Marker::Braille` may provide the symbol and foreground color.
```
joshka, line 79: nit: can you please wrap this at 100 chars instead of 60 or 80. Generally that makes it easier to read prose and is the style we generally use throughout.

ratatui-widgets/src/canvas.rs
```diff
@@ -141,15 +152,31 @@ impl Grid for BrailleGrid {
...
+        let contents = self
+            .utf16_code_points
+            .iter()
+            .zip(&self.colors)
+            .map(|(&code_point, &color)| {
+                let symbol = if code_point == symbols::braille::BLANK {
+                    None
+                } else {
+                    Some(char::from_u32(code_point.into()).unwrap())
+                };
```
joshka, line 164: As an aside - this probably indicates a missing (private) helper newtype for braille chars. Something to maybe do after this PR?

ratatui-widgets/src/canvas.rs
```diff
@@ -175,12 +202,18 @@ struct CharGrid {
...
+
     /// The character to use for every cell - e.g. a block, dot, etc.
     cell_char: char,
+
+    /// If true, apply the color to the background as well as the
+    /// foreground.  This is used for `Marker::Block`, so that it will
+    /// overwrite any previous foreground character, but also leave a
+    /// background that can be overlaid with an additional foreground
+    /// character.
+    apply_color_to_bg: bool,
```
joshka, line 216: nit: wrapping as mentioned above.

ratatui-widgets/src/canvas.rs
```diff
@@ -296,45 +339,47 @@ impl Grid for HalfBlockGrid {
             .tuples()
             .flat_map(|(upper_row, lower_row)| zip(upper_row, lower_row));
 
-        // then we work out what character to print for each pair of pixels
-        let string = vertical_color_pairs
-            .clone()
+        // Then we determine the character to print for each pair,
+        // along with the color of the foreground and background.
+        let contents = vertical_color_pairs
```
joshka, line 344: nit: comment wrapping

ratatui-widgets/src/canvas.rs
```diff
@@ -296,45 +339,47 @@ impl Grid for HalfBlockGrid {
...
+                    symbol: Some(symbols::half_block::FULL),
+                    fg: Some(*upper),
+                    bg: Some(*lower),
+                },
+                (Some(upper), Some(lower)) => LayerCell {
+                    symbol: Some(symbols::half_block::UPPER),
+                    fg: Some(*upper),
+                    bg: Some(*lower),
+                },
             })
```
joshka, line 371: Moving the LayerCell outside of the match simplifies this quite a bit:
```rust
.map(|(upper, lower)| {
    let (symbol, fg, bg) = match (upper, lower) {
        (None, None) => (None, None, None),
        (None, Some(lower) => (Some(half_block::LOWER, Some(*lower), None),
        (Some(upper), None) => Some(half_block::UPPER, Some(*upper), None),
        (Some(upper), Some(lower)) if lower == upper => (Some(half_block::FULL), Some(*upper), Some(*lower)),
        (Some(upper), Some(lower)) => (Some(half_block::UPPER), Some(*upper), Some(*lower)),
    }
    LayerCell { symbol, fg, bg }
})

File: ratatui-widgets/src/canvas.rs [hunk 6]
```diff
@@ -462,6 +507,8 @@ impl<'a, 'b> From<&'a mut Context<'b>> for Painter<'a, 'b> {
 /// this as similar to the `Frame` struct that is used to draw widgets on the terminal.
 #[derive(Debug)]
 pub struct Context<'a> {
+    width: u16,
+    height: u16,
```
joshka, line 511: Make a note here of what widths these are referring to as there's a few different units at play and it could easily get confusing. (I'm reading this code top down and I'm not sure I recall exactly myself which would apply here without some extra comments to help, so I know I'd be confused in maintaining this later)

ratatui-widgets/src/canvas.rs
```diff
@@ -520,6 +560,25 @@ impl<'a> Context<'a> {
         }
     }
 
+    fn marker_to_grid(width: u16, height: u16, marker: Marker) -> Box<dyn Grid> {
+        let dot = symbols::DOT.chars().next().unwrap();
+        let block = symbols::block::FULL.chars().next().unwrap();
+        let bar = symbols::bar::HALF.chars().next().unwrap();
```
joshka, line 566: I think these can probably just be e.g. `symbols::DOT[0]` as we know the length of the characters in these values.

Alternatively, at some later change we could also push this down to CharGrid (which is private) and make it accept Strings instead of Char. This would allow multi-byte (though single cell width) unicode chars to be used as the character instead of only single byte chars. Not a big deal, just an observation from reading this the first time.

ratatui-widgets/src/canvas.rs
```diff
@@ -520,6 +560,25 @@ impl<'a> Context<'a> {
...
+        match marker {
+            Marker::Dot => Box::new(CharGrid::new(width, height, dot)),
+            Marker::Block => Box::new(CharGrid::new(width, height, block).apply_color_to_bg()),
+            Marker::Bar => Box::new(CharGrid::new(width, height, bar)),
+            Marker::Braille => Box::new(BrailleGrid::new(width, height)),
+            Marker::HalfBlock => Box::new(HalfBlockGrid::new(width, height)),
+        }
+    }
+
+    /// Change the marker being used in this context
```
joshka, line 576: Given that this is a public function, can you add some notes about what the behavior of this is (resetting the grid, whatever the effect of finish is). Add some info about when / how this might be used and why in the real world.

ratatui-widgets/src/canvas.rs
```diff
@@ -804,19 +863,21 @@ where
 
         // Retrieve painted points for each layer
         for layer in ctx.layers {
-            for (index, (ch, colors)) in layer.string.chars().zip(layer.colors).enumerate() {
-                if ch != ' ' && ch != '\u{2800}' {
```
joshka, line 806: Does the `\u{2800}` char get flattened into None somewhere? Is that still relevant?
I'm assuming that's the empty braille character symbol?
it's possible that I've missed this in reading above.

ratatui-widgets/src/canvas.rs
```diff
@@ -141,15 +152,31 @@ impl Grid for BrailleGrid {
...
+
+                LayerCell {
+                    symbol,
+                    fg: color,
+                    bg: None,
+                }
+            })
+            .collect();
+
+        Layer { contents }
```
joshka, line 174: There's just enough complexity in this method to need a small comment about how it works (mainly capture the intent to make maintaining it easy)

Lunderberg:
> Hey there, My sincere apologies for letting this one languish for so long.

No worries, as I kept meaning to resolve the conflicts on this one as well.  IIRC, there was a larger cleanup/refactor going on when I initially submitted this, and I

> So it might be worth copying some of the relevant info for users into the PR comment, and ensuring that the developer focused info captured in the docs where appropriate.

Good call, and I'm never quite sure which history will be preserved from where.  I've moved the relevant information from the commit messages into the PR comment, so that it will be preserved when squashed.

> I can see how that would be annoying for people who write good commit messages like this, but I hope that's ok with you.

Not a problem at all, especially as different projects have different preferred styles for long-term maintainability.  As someone who tends to make PRs against whichever libraries I'm using in hobby projects, I appreciate the feedback to fit the standards of each library.

I don't have time at the moment to update based on the review feedback, but I hope to get back to it within a week or so.

j-g00da:
> unless the other maintainers find problems

LGTM, I'm happy to approve this once the current comments are addressed :)

orhun:
Gentle nudge @Lunderberg :)
Still interested in finishing this up?

ratatui-widgets/src/canvas.rs
j-g00da: We can't really do that as both str and Chars are not indexable by integers

Lunderberg:
@orhun @j-g00da Apologies, and thank you for picking it up where I had left it off.  Things got a bit busier on my side than expected, and #2165 looks good to me.

ratatui-widgets/src/canvas.rs
j-g00da: Yup, empty braille is flattened to None in the `save` method.

orhun:
Now done in #2165

Issue #1443, opened 2024-10-21
Stylize the highlight_symbol in List Widget


aktaboot:
## Problem

It seems that it's not currently possible to stylize the highlight_symbol in the List Widget,

## Solution

quoting @joshka : "There isn't a way to do just the symbol currently. Likely the symbol should be changed to a span or a line instead of a str."

soumyasen1809:
@orhun @aktaboot 
Hi, I am new to Rust and took the liberty to work on this beginner issue. I have a PR #1467 for this issue. The CI is failing for a job and I am not sure why.

joshka:
Semver failing is fine - that's a check that helps us understand whether the next release will be a patch release (0.29.1) or a minor (0.30.0). We know that there's breaking changes coming up in the next release.

soumyasen1809:
@joshka Hi, thanks for the information. I have marked my PR ready for review. Please let me know if you have any questions and/or if the PR looks fine to merge

PR #1467, opened 2024-11-02
refactor!: Change highlight_symbol to use Line

soumyasen1809:
- Changed highlight_symbol from str to Line
- Adjusted Tests and Benches

Fixes issue #1443


ratatui/src/widgets/list/list.rs
```diff
@@ -290,13 +293,13 @@ impl<'a> List<'a> {
...
     /// let items = ["Item 1", "Item 2"];
-    /// let list = List::new(items).highlight_symbol(">>");
+    /// let list = List::new(items).highlight_symbol(Line::from(">>"));
     /// ```
     #[must_use = "method moves the value of self and returns the modified value"]
-    pub const fn highlight_symbol(mut self, highlight_symbol: &'a str) -> Self {
+    pub fn highlight_symbol(mut self, highlight_symbol: Line<'a>) -> Self {
         self.highlight_symbol = Some(highlight_symbol);
         self
     }
```
orhun, line 305: This function should take `Into<Line>` for allowing flexible conversion between types :)

```suggestion
    pub fn highlight_symbol<T>(mut self, highlight_symbol: T) -> Self
    where
        T: Into<Line<'a>>,
    {
        self.highlight_symbol = Some(highlight_symbol.into());
        self
    }
```

orhun (changes requested):
Thanks for the PR!

ratatui/src/widgets/list/list.rs
soumyasen1809: Added in next commit: https://github.com/ratatui/ratatui/pull/1467/commits/e3ed0feac72fafe1534ad6fdd21444ff6af2f013

orhun:
Can you also revert the changes where the string reference usage was replaced with a `Line`? Those should be backwards compatible now.

As for showcasing this new feature, we can use it somewhere in the examples I think.

soumyasen1809:
> Can you also revert the changes where the string reference usage was replaced with a `Line`? Those should be backwards compatible now.

@orhun Done in the new PR

orhun (approved):

ratatui/src/widgets/list/list.rs
```diff
@@ -296,8 +297,8 @@ impl<'a> List<'a> {
     /// let list = List::new(items).highlight_symbol(">>");
     /// ```
     #[must_use = "method moves the value of self and returns the modified value"]
-    pub const fn highlight_symbol(mut self, highlight_symbol: &'a str) -> Self {
```
joshka, line 299: This is a (small) breaking change, which should be noted in the breaking changes doc and commit (the first PR comment will be used as the squashed commit message).
- breaks inference
- breaks const usage

Can you please search for similar changes in the docs and change log and copy the wording?

ratatui/src/widgets/list/rendering.rs
```diff
@@ -69,7 +68,7 @@ impl StatefulWidgetRef for List<'_> {
         state.offset = first_visible_index;
 
         // Get our set highlighted symbol (if one was set)
-        let highlight_symbol = self.highlight_symbol.unwrap_or("");
+        let highlight_symbol = self.highlight_symbol.clone().unwrap_or_default();
```
joshka, line 71: Does as_ref() work here instead of clone?

soumyasen1809: Updated in https://github.com/ratatui/ratatui/pull/1467/commits/095bf84245697a19568bbce422c5ea9592305fd1

ratatui/src/widgets/list/list.rs
airblast-dev: You can check out older PR's that made breaking changes. 

For example take a look at these PR's
- https://github.com/ratatui/ratatui/pull/1418
- https://github.com/ratatui/ratatui/pull/1331

The first comment contains a general description of the change, followed by the breaking changes written as "BREAKING CHANGE: \<the change that is breaking\>"

As for the title, it should be something like "refactor!: ..." with an exclamation point to indicate that it was a breaking change.

At last you will also want to update the breaking changes doc with the relevant information. See these commits for an example: 
- https://github.com/ratatui/ratatui/pull/1331/commits/ea10843985eb873dae113d550ff63b16e1f2ebf4
- https://github.com/ratatui/ratatui/pull/1418/commits/62c7fb9dca12e11a5510501ba3422f05d87ac398

In simpler terms your initial PR comment is used as a commit message, and the PR title is used as the commit name when the commits are squashed. So your first comment of the PR, and the PR's title should follow [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/).

joshka: Thanks @airblast-dev. +1 to everything there.

The inference part is described in each of the places where we've introduced `Into` in the breaking changes doc: Steal the wording from: https://github.com/ratatui/ratatui/blob/main/BREAKING-CHANGES.md#axislabels-now-accepts-intoiteratorintoline-1273-and-1283

Same deal with const - https://github.com/ratatui/ratatui/blob/main/BREAKING-CHANGES.md#block-style-methods-cannot-be-used-in-a-const-context-720 is an ok example.

joshka: Put more concretely:
- Add a section to BREAKING-CHANGES.md that describes the inference issue. Copy it from somewhere else.
- Update the first PR comment to include a conventional commits `BREAKING CHANGES:` footer which has similar information. This will be used in the changelog when we publish Ratatui next.

PR #1473, opened 2024-11-06
chore: add CI check for up-to-date README.md

orhun:


orhun:
While working on this I also wondered if we should just do `#![doc = include_str!("../../README.md")]` for `ratatui` as well. 

Also, currently there are no Linux binaries published for cargo-rdme which we can use in the CI to avoid `cargo install`. See https://github.com/orium/cargo-rdme/issues/239

joshka:
> While working on this I also wondered if we should just do `#![doc = include_str!("../../README.md")]` for `ratatui` as well.

The cargo-rdme approach is better IMO. Because:
- intralinks are mapped to docs.rs links
- heading levels are adjusted correctly
- documentation is available when directly viewing the source

I was lazy with -core for this, focusing more on getting it done than on worrying about choosing the perfect tooling.

joshka (approved):

.github/workflows/ci.yml
```diff
@@ -116,6 +116,16 @@ jobs:
       - uses: Swatinem/rust-cache@v2
       - run: cargo xtask check
 
+  # Check if README.md is up-to-date with the crate's documentation.
+  check-readme:
+    name: Check README
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: Swatinem/rust-cache@v2
```
joshka, line 125: I think this may make this fast enough.

PR #1474, opened 2024-11-06
chore: move widgets into ratatui-widgets crate

joshka:
Part of: #1388


joshka:
TODO:
- [ ] Readme
- [ ] probably some other bits

ratatui/src/widgets.rs
```diff
@@ -1,8 +1,14 @@
 #![warn(missing_docs)]
 //! `widgets` is a collection of types that implement [`Widget`] or [`StatefulWidget`] or both.
 //!
-//! Widgets are created for each frame as they are consumed after rendered.
-//! They are not meant to be stored but used as *commands* to draw common figures in the UI.
+//! The widgts provided with Ratatui are implemented in the [`ratatui_widgets`] crate, and are
```
orhun, line 4:
```suggestion
//! The widgets provided with Ratatui are implemented in the [`ratatui_widgets`] crate, and are
```

ratatui-widgets/Cargo.toml
orhun, line 1: We should also add a short README.md for this crate.

ratatui-widgets/src/lib.rs
```diff
@@ -0,0 +1,46 @@
...
+pub mod table;
+pub mod tabs;
+
+/// Internal module for reflowing text.
+mod reflow;
+
+#[cfg(feature = "calendar")]
+pub mod calendar;
+
+// use ratatui_core::widgets::Widget;
```
orhun, line 46: Not needed?
```suggestion
```

ratatui-widgets/Cargo.toml
joshka: Go for it. I'm intermittently going to come back to this every few days.

ratatui-widgets/src/lib.rs
joshka: https://github.com/ratatui/ratatui/pull/1474/commits/62a9eeb24202c68e0a0ef2094ddb650d6ca93140

ratatui-widgets/Cargo.toml
joshka: https://github.com/ratatui/ratatui/pull/1474/commits/62a9eeb24202c68e0a0ef2094ddb650d6ca93140

orhun:
The doctest are failing because we need to change e.g. `ratatui::widgets::X` usage to `ratatui_widgets::x:X`. However, I'm not sure if we should do it. Mainly because those widgets are exported in `ratatui` and the documentation is reflected there.

Another issue that I realized is that we are not running the tests _without_ `all-features` so the CI seems to be passing whereas it fails locally for me when I run `cargo build` in `ratatui-widgets` directory.

joshka:
> Looks like adding `ratatui` to `ratatui_widgets` as a dev dependency solved that docs issue. I was expecting to get a rat eating its own tail when I did that but nvm I guess...

Neat, so did I there :D

joshka:
We should probably do the same for -core. I didn't know that dev-deps would be used when running doc tests, so that solves the problem I had with the -core docs well too.

joshka:
> Another issue that I realized is that we are not running the tests _without_ `all-features` so the CI seems to be passing whereas it fails locally for me when I run `cargo build` in `ratatui-widgets` directory.

We should probably add cargo-hack to the testing approach to cover this properly.

joshka:
Merging without auth on this one as Orhun is familiar with the changes and there's a followup PR that would be easier to review without these changes.

PR #1459, opened 2024-10-30
chore: move ratatui crate into workspace folder

joshka:
This is the first step towards modularization. Handling the move
as a separate step rather than combining it should make it easier
to rebase other PRs when necessary.

PR #1484, opened 2024-11-12
docs(readme): correct examples link

HoKim98:
Hello, this PR just corrects the examples hyperlink.
The PR #1459 introduced the workspace directory, but it missed this link.

* Before: https://github.com/ratatui/ratatui/tree/main/examples/README.md
* After: https://github.com/ratatui/ratatui/tree/main/ratatui/examples/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -293,7 +293,7 @@ fn draw(frame: &mut Frame) {
 [Layout]: https://ratatui.rs/how-to/layout/
 [Styling Text]: https://ratatui.rs/how-to/render/style-text/
 [templates]: https://github.com/ratatui/templates/
-[Examples]: https://github.com/ratatui/ratatui/tree/main/examples/README.md
+[Examples]: https://github.com/ratatui/ratatui/tree/main/ratatui/examples/README.md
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -363,7 +363,7 @@ for a curated list of awesome apps/libraries built with `ratatui`!
 
 ## Alternatives
 
-You might want to checkout [Cursive](https://github.com/gyscos/Cursive) or
```
orhun, line 366: Can you revert this formatting change? It seems unrelated :)

joshka: I'm fine with this - the EOL space doesn't belong.

joshka (changes requested):
It may not be obvious, but we generate most of the readme from the lib.rs using cargo-rdme. Can you please also make the same changes there also so that these keep in sync with each other. https://github.com/ratatui/ratatui/actions/runs/11801844681/job/32876120141?pr=1484 is the failing CI check for this.

HoKim98:
Oh it's my mistake

README.md
HoKim98: I think this PR is not relevant to this part. It's my fault by using auto-formatter.

joshka (approved):

PR #1490, opened 2024-11-16
docs: improve docs for workspace crates

orhun:
Overall makes improvements in the documentation of the workspace crates and checking them.


joshka:
I've got a draft change for fixing widgetref all up, but it's fairly big and needs to be broken up (working on that slowly). By making the widgets just implement Widget for &W and making a blanket WidgetRef implementation for any &W it is possible to remove the widgetref stuff from -core, and -widgets.

orhun:
That's nice. Are you fine with adding a default feature in the meantime?

Btw, can you share a bit about what you are working on currently? I'd like to take on splitting the backends but I don't want to overlap with your already existing drafts if any :)

joshka:
Feel free to go ahead with the backend split. I haven't looked at the close change on this yet to answer the question about default yet.

ratatui-widgets/README.md
```diff
@@ -71,6 +75,6 @@ details on contributing, please see the [CONTRIBUTING](CONTRIBUTING.md) document
 
 ## License
 
-This project is licensed under the MIT License. See the [LICENSE](LICENSE) file for details.
+This project is licensed under the MIT License. See the [LICENSE](../LICENSE) file for details.
```
joshka, line 78: I'm curious how this would render in the crates page once packaged, it's likely no big deal, but any idea? Not sure if there's any easy way to test that either.

xtask/src/main.rs
```diff
@@ -172,7 +172,18 @@ fn check() -> Result<()> {
 
 /// Run cargo-rdme to check if README.md is up-to-date with the library documentation
 fn check_readme() -> Result<()> {
-    run_cargo(vec!["rdme", "--workspace-project", "ratatui", "--check"])
+    let meta = MetadataCommand::new()
+        .exec()
+        .wrap_err("failed to get cargo metadata")?;
+    for package in meta.workspace_default_packages() {
```
joshka, line 178: nit: what about making this a function? It's repeated from below.

joshka (approved):
Approved with nits/questions

joshka:
The mentioned widget changes are in https://github.com/ratatui/ratatui/pull/1491

xtask/src/main.rs
orhun: fa25134054c57a7a85d5deb58dde0741a8d5f121

ratatui-widgets/README.md
orhun: Good question. We should probably use an absolute URL in there. (i.e. `https://github.com/ratatui/ratatui/blob/main/LICENSE`) (I ran `cargo package` to analyze the generated assets just to sanity check that.)

README.md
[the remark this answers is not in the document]
orhun: Yup, sounds good. Let's do that sometime, also adding a mermaid diagram would be nice in terms of visualizing the architecture.

PR #1491, opened 2024-11-17
chore: move unstable widget refs to ratatui

joshka:
commit 33b670263

    chore: remove unstable-widget-ref feature flag from -widgets and -core

commit 6df6dee3715454b0b933e97d158bbbee0d475039

    chore!: move {Stateful,}WidgetRef to ratatui from -core
    
    These are less stable than the non-ref traits as we have not yet
    committed to the exact API. This change moves them to ratatui from
    ratatui-core.
    
    To facilitate this:
    - implementations of WidgetRef for all internal widgets are removed and
      replaced with implementations of Widget for references to those
      widgets.
    - Widget is now implemented for Option<W> where W: Widget, allowing for
      rendering of optional widgets.
    - The blanket implementation of Widget for WidgetRef is reversed, to be
      a blanket implementation of WidgetRef for all &W where W: Widget.
    
    BREAKING CHANGE: implementations of WidgetRef no longer have a blanket
    implementation of Widget, so Widgets should generally implement the
    Widget trait on a reference to the widget rather than implementing
    WidgetRef directly. This has the advantage of not requiring unstable
    features to be enabled.

commit 9e510628066294f13c28f2ab6a46bb264205f56e

    fix: warnings in stylize test when default-features=false

commit 58521310cdd1268015b07dab21583e62b5b6cb47

    chore: add cargo-hack to bacon config

commit 7483ccf26a99a5d495b91de9b247d3e2c08126f4

    style: autoformat bacon.toml

---
The rationale for adding cargo-hack here was that problems with using the Ref types show up only when the unstable-widget-ref feature was not enabled.

I'd like to defer adding cargo-hack to CI / xtask from this task as this requires a a bunch of experimenting to get the right combination of coverage and speed. There's thousands of combinations of features, but most are functionally equivalent so making CI wait for cargo-hack to succeed on the full feature powerset would be mostly a waste of time.


ratatui-core/src/style/stylize.rs
joshka, line 1: The problem here was that the `#[cfg]` attributes weren't case specific, but applied to the whole test, which meant that rstest was unused when the underline feature flag was not enabled.

ratatui/src/widgets/widget_ref.rs
```diff
@@ -82,12 +82,23 @@ pub trait WidgetRef {
...
 /// This allows you to render a widget by reference.
-impl<W: WidgetRef> Widget for &W {
-    fn render(self, area: Rect, buf: &mut Buffer) {
-        self.render_ref(area, buf);
+impl<W> WidgetRef for &W
+where
+    for<'a> &'a W: Widget,
+{
+    fn render_ref(&self, area: Rect, buf: &mut Buffer) {
+        self.render(area, buf);
```
joshka, line 101: This is the breaking change, which likely affects [a bunch of places](https://github.com/search?q=ratatui+%22fn+render_ref%22+AND+%28NOT+org%3Aratatui%29+AND+%28NOT+is%3Afork%29&type=code) (294 hits), so should definitely be called out in the highlights when we do change notes. Likely we should do a bit of a deep explanation on this when releasing.

https://github.com/search?q=ratatui%20render_ref%20AND%20(NOT%20org%3Aratatui)%20AND%20(NOT%20is%3Afork)&type=code

ratatui/src/widgets/widget_ref.rs
```diff
@@ -82,12 +82,23 @@ pub trait WidgetRef {
...
     /// to implement a custom widget.
     fn render_ref(&self, area: Rect, buf: &mut Buffer);
+
+    /// Converts the widget into a boxed widget.
+    fn into_boxed(self) -> Box<dyn WidgetRef>
+    where
+        Self: Sized + 'static,
+    {
+        Box::new(self)
+    }
```
joshka, line 92: I'll remove this for now as it didn't end up being necessary - I'm unsure whether this should be on Widget or WidgetRef right now.

ratatui-core/src/widgets/widget.rs
```diff
@@ -82,7 +81,15 @@ impl Widget for &str {
 /// on a [`Buffer`] within the bounds of a given [`Rect`].
 impl Widget for String {
     fn render(self, area: Rect, buf: &mut Buffer) {
-        self.render_ref(area, buf);
+        buf.set_stringn(area.x, area.y, self, area.width as usize, Style::new());
```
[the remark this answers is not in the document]
orhun: uuuh what

ratatui-widgets/src/list/rendering.rs
```diff
@@ -12,39 +12,31 @@ use crate::{
...
-    fn render(self, area: Rect, buf: &mut Buffer, state: &mut Self::State) {
-        StatefulWidgetRef::render_ref(self, area, buf, state);
-    }
-}
 
-impl StatefulWidgetRef for List<'_> {
-    type State = ListState;
-
-    fn render_ref(&self, area: Rect, buf: &mut Buffer, state: &mut Self::State) {
+    fn render(self, area: Rect, buf: &mut Buffer, state: &mut Self::State) {
```
orhun, line 37: This change broke the doctest case in `render_stateful_widget_ref` in `Frame`:

```
---- ratatui/src/terminal/frame.rs - terminal::frame::Frame<'_>::render_stateful_widget_ref (line 170) stdout ----
error[E0277]: the trait bound `List<'_>: StatefulWidgetRef` is not satisfied
   --> ratatui/src/terminal/frame.rs:185:34
    |
18  | frame.render_stateful_widget_ref(list, area, &mut state);
    |       -------------------------- ^^^^ the trait `StatefulWidgetRef` is not implemented for `List<'_>`
    |       |
    |       required by a bound introduced by this call
    |
note: required by a bound in `Frame::<'_>::render_stateful_widget_ref`
   --> /home/orhun/gh/ratatui/ratatui/src/terminal/frame.rs:191:12
    |
189 |     pub fn render_stateful_widget_ref<W>(&mut self, widget: W, area: Rect, state: &mut W::State)
    |            -------------------------- required by a bound in this associated function
190 |     where
191 |         W: StatefulWidgetRef,
    |            ^^^^^^^^^^^^^^^^^ required by this bound in `Frame::<'_>::render_stateful_widget_ref`

error: aborting due to 1 previous error
```

Btw what's the current status with the blanket implementation of `StatefulWidget for &W where W: StatefulWidget`? I guess we need that here.

orhun (commented):
Nice! I pushed some commits to fix the conflicts and failing doctests. There is still one in there though.

ratatui-core/src/widgets/widget.rs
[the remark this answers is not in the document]
joshka: figured I'd give the CoPilot review a try. This is definitely not the right idea :D

ratatui-widgets/src/list/rendering.rs
joshka: f8b6cc1be..1e4ebd730 fixes this

ratatui/src/widgets/stateful_widget_ref.rs
```diff
@@ -64,20 +67,18 @@ pub trait StatefulWidgetRef {
...
-//
-// /// Blanket implementation of `StatefulWidget` for `&W` where `W` implements `StatefulWidgetRef`.
-// ///
-// /// This allows you to render a stateful widget by reference.
-// impl<W: StatefulWidgetRef> StatefulWidget for &W {
-//     type State = W::State;
-//     fn render(self, area: Rect, buf: &mut Buffer, state: &mut Self::State) {
-//         StatefulWidgetRef::render_ref(self, area, buf, state);
-//     }
-// }
```
orhun, line 80: Now that this is removed, does it mean that the widget ref is stabilized?

orhun (approved):

ratatui/src/widgets/stateful_widget_ref.rs
joshka: Not quite - there's still some issues around committing to trait / method naming etc. Thanks for the doctest fixes btw.

PR #1492, opened 2024-11-17
fix(example): fix scrollbar length in table example

kardwen:
This is just a small fix to ensure that the scrollbar is always a long as the table. I noticed that scrollbar changes it size by one while scrolling whenever there is a mismatch with the table viewport in length.

I wonder if there is a nicer way to do this

image

### DIFF
--- a/ratatui/examples/table.rs
+++ b/ratatui/examples/table.rs
@@ -18,7 +18,7 @@ use crossterm::event::KeyModifiers;
 use itertools::Itertools;
 use ratatui::{
     crossterm::event::{self, Event, KeyCode, KeyEventKind},
-    layout::{Constraint, Layout, Margin, Rect},
+    layout::{Constraint, Layout, Rect},
     style::{self, Color, Modifier, Style, Stylize},
     text::Text,
     widgets::{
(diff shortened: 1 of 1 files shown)


orhun:
We should probably find a simpler solution or even better provide a better API for positioning the `ScrollBar` inside an area.

This is not a comment about your PR in particular but our examples really feel like _small demos_ at this point. I'm planning to revamp everything in there before the upcoming release.

ratatui/examples/table.rs
```diff
@@ -202,13 +202,24 @@ impl App {
...
-        self.render_footer(frame, rects[1]);
+        let main_area = areas[0];
+        self.render_table(frame, main_area);
+        // Horizontal margin of 1 left and right
+        // Account for table header
+        let scrollbar_area = Rect {
+            x: main_area.x + 1,
+            y: main_area.y + 1,
+            width: main_area.width.saturating_sub(2),
+            height: (main_area.height.saturating_sub(1) / ITEM_HEIGHT as u16) * ITEM_HEIGHT as u16,
```
joshka, line 217: I don't think I like this much, as it looks a bit weird when the track doesn't touch the border of the footer. But I also don't like the borders on the footer much aesthetically either.

The original intent of the margin was to make the scrollbar match the table highlight block a bit, but I think you're right in observing that the effect looks a bit janky right now.

Some Suggestions for de-janking this example a bit more if you're up for it...
- Make the track render as a solid bar
- Add a theme color for the track part
- Move the scrollbar to the extreme right side rather than being indented
- take up the whole table size with the scrollbar, not just the size of the items
- Remove the double borders on the footer and replace it with a solid background color fitting with the theme (take up 2 rows instead of 4)

joshka:
> provide a better API for positioning the `ScrollBar` inside an area.

The issue here is that we're positioning the Scrollbar and taking note of the title of the table to avoid it.

I don't think Scrollbar should have position inside area functions - we have them already on Rect / Layout, so adding it to Scrollbar might be problematic. Better would be that Widgets that define internal areas expose them when rendering (but this is something that is touches on what we've been generally talking about with respect to what methods exist on Widgets, so I don't want to go much further talking about that in this issue.)

kardwen:
For the area I will document that `viewport_content_length` uses it as fallback in another PR.

The scrollbar style is reversed when a row is highlighted, I think there is no easy way to prevent it, so moving it to the right sounds good.

PR #1501, opened 2024-11-18
feat: add bounds accsess in Painter and off-grid Line points support

renesat:
Support draw Canvas Line if points off-grid and support Canvas bound access in Painter

Fixes: https://github.com/ratatui/ratatui/issues/1489

### DIFF
--- a/ratatui-widgets/src/canvas/line.rs
+++ b/ratatui-widgets/src/canvas/line.rs
@@ -1,3 +1,4 @@
+use line_clipping::{cohen_sutherland, LineSegment, Point, Window};
 use ratatui_core::style::Color;
 
 use crate::canvas::{Painter, Shape};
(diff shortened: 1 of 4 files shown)


renesat:
I'm not sure create one PR or 2, but these features depend.

ratatui-widgets/src/canvas/line.rs
```diff
@@ -124,8 +180,58 @@ mod tests {
     use crate::canvas::Canvas;
 
     #[rstest]
-    #[case::off_grid(&Line::new(-1.0, -1.0, 10.0, 10.0, Color::Red), ["          "; 10])]
-    #[case::off_grid(&Line::new(0.0, 0.0, 11.0, 11.0, Color::Red), ["          "; 10])]
+    #[case::off_grid(&Line::new(-1.0, 0.0, -1.0, 10.0, Color::Red), ["          "; 10])]
```
joshka, line 183: The purpose of the case names (off_grid etc.) is to make it easy to work out which test failed without having to count down the index. Can you please add number suffixes?

renesat: I add named suffixes instead of numbers. It's ok?

renesat:
> Approved generally speaking. I made a quick crate for line clipping that I think may be worth using instead of having this code in Ratatui. Can you take a look and see if it makes sense / add the algorithm here if needed.

Do you mean this repo: https://github.com/joshka/line-clipping ?
Yeah, I think that's a better option. My function is essentially not based, on a specific algorithm. Just half from my head, half StackOverflow :)

ratatui-widgets/src/canvas/line.rs
joshka: The problem is that there are multiple cases with the same name. The names need to be unique to avoid having to count the cases to work out which of them failed.

ratatui-widgets/src/canvas/line.rs
```diff
@@ -32,12 +32,19 @@ impl Line {
...
+            cut_line_in_area(x_bounds, y_bounds, self.x1, self.y1, self.x2, self.y2)
+        else {
             return;
         };
-        let Some((x2, y2)) = painter.get_point(self.x2, self.y2) else {
+        let Some((x1, y1)) = painter.get_point(x1, y1) else {
             return;
         };
+        let Some((x2, y2)) = painter.get_point(x2, y2) else {
+            return;
```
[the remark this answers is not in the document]
joshka: How about using `canvas_` and `grid_` prefixes here? I'm not sure which of these at a glance would be the obvious `world` point.

joshka (approved):
LGTM. Thanks for sticking with this and getting it done.

PR #1503, opened 2024-11-19
chore(examples): move async example to apps

joshka:
Move async example to examples/apps/async as full project.
Simplify a little by removing the need for the github api token.

---

The goal of this PR is to make it easier to have a bit more guidance about how much code ends up in each example. The first step is to get them into a bit of a structure and identify gaps. I think we should move the examples into a similar structure as the [examples section](https://ratatui.rs/examples/) of the website. Right now that's Apps, Layout, Style, Widgets, but these aren't set in stone.

We should generally look to keep the simplest examples in the relevant parts of the repo (as they provide automatic doc snippets, and are good as direct easy references to demonstrate how to use each widget etc., while moving the more advanced full applications with full design decisions into the main examples folder.

This WIP is an example for comment about how this might look.

We might also consider moving the simple widget specific examples into the new ratatui-widgets crate. from the main crate. Ideally these shouldn't be about doing a full app, but really showing how to use the widget.

Ping @kdheepak for feedback on the idea (we've discussed some parts of this previously).


orhun:
This is a good change, however I think we should rename this example to `async-demo` and do something similar for the rest of examples that are a bit _too much_ for a Rust beginner.

Maybe we can update the `examples/` folder with small code snippets for each widget (easily copy/pasteable) and have a separate project for the demos like this.

It's been a while since I looked at the examples and I would like to refactor them a bit... 

What do you think?

joshka:
> This is a good change, however I think we should rename this example to `async-demo` and do something similar for the rest of examples that are a bit _too much_ for a Rust beginner.
> 
> Maybe we can update the `examples/` folder with small code snippets for each widget (easily copy/pasteable) and have separate projects for the demos like this.
> 
> It's been a while since I looked at the examples and I would like to refactor them a bit...
> 
> What do you think?

Yep to all that. I'll throw a small issue up for discussing this more generally and getting some actions items (or there might already be one)

orhun:
Can we close this now? #1512

orhun (approved):
- Rebased and fixed conflicts
- Renamed to `async-github` (as discussed before)
- Added docs for the app
- Update `deny.toml` about `ring`'s license

PR #1507, opened 2024-11-20
fix: round canvas coordinates to nearest grid cell

joshka:
Previously the canvas coordinates were rounded towards zero, which
causes the rendering to be off by one pixel in some cases. It also meant
that pixels at the extreme edges of the canvas can only be drawn if the
point was exactly on the edge of the canvas. This commit rounds the
coordinates to the nearest integer instead.


renesat:
Yes, I think it's more logical to round in Canvas. The circle got more circular :)
But, I noticed that I still need to update the tests `ratatui/tests/widgets_chart.rs`.

joshka:
Yup, I saw that in ci and haven't yet swung back to fix it. Will do sometime in the next few days. I mainly wanted to check that the idea was sound, so thank you for taking a look.

joshka:
Fixed failing tests, removed an integ test that was a bit annoying to update. There's sufficient unit tests on this one to cover legend positions I think.

orhun (approved):

PR #1511, opened 2024-11-24
docs(widgets): Add simple barchart example

joshka:


orhun (approved):
Liked the simplicity!

orhun:
Hmm, are we moving the other examples to `ratatui-widgets` as well?

Maybe sth like this:

- `ratatui-widgets`: Contains widgets (examples)
- `ratatui`: Demos (each one will be a Rust project)

joshka:
> Hmm, are we moving the other examples to `ratatui-widgets` as well?
> 
> Maybe sth like this:
> 
>     * `ratatui-widgets`: Contains widgets (examples)
> 
>     * `ratatui`: Demos (each one will be a Rust project)

Yep - exactly what I was thinking too.

This ends up really really nice in the docs due to scraping too. The trick is have a small method that demos the widget, with a good name.

image

Issue #1512, opened 2024-11-24
Examples rearrangement


joshka:
Follows on from:
- https://github.com/ratatui/ratatui/pull/1503
- #1388 

## Problem
- Simple examples communicate how to use things the best. There's a lot of examples which are mini apps / demos
- App-like demos often do a bit too much

## Constraints / Needs
- Examples are easy to find - contextual
- Examples are easy to run
- Examples are easy to export to the website
- We have at least 4 and maybe more types of examples:
  1. examples for the api docs (scraped by docs.rs)
  2. example full apps / demos (demo, demo2, colors_rgb, async (github))
  3. tools which are useful for doing things (e.g. colors / modifier / constraints-explorer)
  4. small examples of each backend (missing currently we only have demo)

## Solution
- [ ] ratatui-widgets should have super simple examples - generally no more complex than E.g., https://github.com/ratatui/ratatui/pull/1511. These will be used when scraping examples into the docs.rs page.
- [ ] apps either stay under the ratatui/examples folder or move to a top level folder (I'm 70/30 for a top level folder, but would be fine with leaving them under ratatui directly. These should have more context about their running conditions (e.g. a full Cargo.toml / src folder / main.rs etc.)
- [ ] Spend some time making the examples work well in the website. I'm not sure 100% what that looks like. More work / ideas needed here.

orhun:
I want to focus on other areas (website, docs) of the project until next year, so here are the list of examples left if anyone wants to continue the work:

- [ ] [ `colors_rgb`](https://github.com/ratatui/ratatui/blob/main/ratatui/examples/colors_rgb.rs) (rename to `colors-rgb`)
- [ ] [`constraint-explorer`](https://github.com/ratatui/ratatui/blob/main/ratatui/examples/constraint-explorer.rs)
- [ ] [`constraints`](https://github.com/ratatui/ratatui/blob/main/ratatui/examples/constraints.rs)
- [ ] [`flex`](https://github.com/ratatui/ratatui/blob/main/ratatui/examples/flex.rs)
- [ ] [`gauge`](https://github.com/ratatui/ratatui/blob/main/ratatui/examples/gauge.rs)
- [ ] [`hello_world`](https://github.com/ratatui/ratatui/blob/main/ratatui/examples/hello_world.rs)
- [ ] [`minimal`](https://github.com/ratatui/ratatui/blob/main/ratatui/examples/minimal.rs)
[...]

PR #1519, opened 2024-11-25
Move and improve examples

orhun:
This is probably going to take longer than I was expecting...

see #1512

### DIFF
--- a/ratatui-widgets/examples/block.rs
+++ b/ratatui-widgets/examples/block.rs
@@ -1,6 +1,7 @@
-//! # [Ratatui] Block example
+//! # [Ratatui] `Block` example
 //!
-//! The latest version of this example is available in the [examples] folder in the repository.
+//! The latest version of this example is available in the [widget examples] folder in the
+//! repository.
 //!
 //! Please note that the examples are designed to be run against the `main` branch of the Github
 //! repository. This means that you may not be able to compile with the latest release version on
@@ -10,10 +11,11 @@
 //! library you are using.
 //!
 //! [Ratatui]: https://github.com/ratatui/ratatui
-//! [examples]: https://github.com/ratatui/ratatui/blob/main/examples
+//! [widget examples]: https://github.com/ratatui/ratatui/blob/main/ratatui-widgets/examples
 //! [examples readme]: https://github.com/ratatui/ratatui/blob/main/examples/README.md
 
 use color_eyre::Result;
+use itertools::Itertools;
 use ratatui::{
     crossterm::event::{self, Event, KeyCode, KeyEventKind},
     layout::{Alignment, Constraint, Layout, Rect},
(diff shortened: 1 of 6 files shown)


ratatui-widgets/examples/block.rs
joshka, line 1: I was thinking for this example to simplify it a bit more. 3 blocks (bordered, styled, custom borders), no deduplication / complex code, just the code that demonstrates the block, no paragraphs. Aim for `<100` LoC and few function params. Not because it's particularly confusing, but because the goal of these examples should be to quickly demonstrate some usage of just the widget not all the possible real world ways to use things.

ratatui/Cargo.toml
```diff
@@ -58,7 +58,7 @@ all-widgets = ["widget-calendar"]
 #! Widgets that add dependencies are gated behind feature flags to prevent unused transitive
 #! dependencies. The available features are:
 ## enables the [`calendar`](widgets::calendar) widget module and adds a dependency on [`time`].
-widget-calendar = ["ratatui-widgets/calendar"]
+widget-calendar = ["ratatui-widgets/calendar", "dep:time"]
```
joshka, line 61: Side note: what about making this included in default for the widgets lib, but exclude it from the default in ratatui?
(Or even include it in the default in Ratatui as well, with the expectation that people that want smaller deps will generally disable defaults)

ratatui/examples/layout.rs
joshka, line 1: This example can probably can be removed. The constraint explorer is a better demo for showing this. The idea of this example was to show how every combination of constraints interacted. But it no longer represents that with the various changes to support flex, spacing, etc.

joshka (commented):
I think the loop can probably be simplified to:

```rust
	loop {
        terminal.draw(draw)?;
        if matches!(event::read()?, Event::Key(_)) {
            break Ok(());
        }
    }
```

I also want to get some extra `is_` and `as_` methods on the crossterm enums that would simplify that to:

```rust
	loop {
        terminal.draw(draw)?;
        if event::read()?.is_key() {
            break Ok(());
        }
    }
```

or possible even:
```rust
    terminal.draw(draw)?;
	while !event::read()?, Event::Key(_)) {
        terminal.draw(draw)?;
    }
	Ok(())
```

joshka:
BTW - it could be worth doing these examples by one, that way, simple changes can just be merged without being held up on issues in more complex examples.

orhun:
I'll split up this PR :)

ratatui/Cargo.toml
orhun: Yeah, good call. #1521

ratatui/examples/layout.rs
joshka: No, if anything, that's related to -core, but I'd keep it in Ratatui. Better yet, anything which is more of an app should be at the root under /examples not /ratatui/examples and should have all the necessary cargo.toml etc. to work properly from there.

joshka:
> I'll split up this PR :)

Did you want to close this one, or leave it open?

orhun:
I think I addressed everything in here so we can close it.

PR #1523, opened 2024-11-27
chore: Move the demo example to main folder

joshka:
Add a top level examples folder for more app-ish examples
Movethe demo example into the top level folder.


examples/README.md
orhun, line 1: What do you think about renaming this such as:

- `examples/demo` -> `demos/ratatui-demo`

And following that for the rest of the demos, e.g. `demos/constraint-explorer`, `demos/async`, etc.

Just realized having 3 _examples folder_ is a bit too much 💀 Maybe `apps` is a good name too.

examples/README.md
```diff
@@ -0,0 +1,50 @@
...
+[ratatui examples]: ../ratatui/examples
+[ratatui-widgets examples]: ../ratatui-widgets/examples
+
+You can run these examples using:
+
+```shell
+cargo run -p example-name
+```
+
+This folder might use unreleased code. Consider vieing the examples in the `latest` branch instead
```
orhun, line 15:
```suggestion
This folder might use unreleased code. Consider viewing the examples in the `latest` branch instead
```

examples/README.md
```diff
@@ -0,0 +1,50 @@
...
+>   allows you to select any previous tagged version.
+> - To view the code locally, checkout the tag. E.g. `git switch --detach v0.26.1`.
+> - Use the latest [alpha version of Ratatui] in your app. These are released weekly on Saturdays.
+> - Compile your code against the main branch either locally by adding e.g. `path = "../ratatui"` to
+>   the dependency, or remotely by adding `git = "https://github.com/ratatui/ratatui"`
+>
+> For a list of unreleased breaking changes, see [BREAKING-CHANGES.md].
+>
+> We don't keep the CHANGELOG updated with unreleased changes, check the git commit history or run
+> `git-cliff -u` against a cloned version of this repository.
```
orhun, line 37: This warning seems a bit too much, do we actually need it? It can be simplified a lot or just removed.
Also the ratatui version/link there is wrong.

examples/README.md
```diff
@@ -0,0 +1,50 @@
...
+## Demo
+
+This is the original demo example from the main README. It is available for each of the backends.
+[Source](demo/).
+
+```shell
+cargo run --example=demo --features=crossterm
+cargo run --example=demo --no-default-features --features=termion
+cargo run --example=demo --no-default-features --features=termwiz
+```
```
orhun, line 48: The instructions should be in the respective demo directory I think, we can simply have a GIF here and people can check out how to run it if they want.

examples/README.md
joshka: I like having a few folders under this folder (I forgot that I previously moved the async example in https://github.com/ratatui/ratatui/pull/1503). Probably apps rather than demos.

I don't think we need to repeat the `ratatui-` for example projects, but perhaps `-example` could be good.

I'd want the constraint-explorer to be discoverable if you don't know it's there, so tools might be a good name for that sort of thing.

The rationale for multiple example folders is based on example scraping. We want there to be some examples that are simple for -widgets. I'd say these are easiest to create by copying the barchart and putting ~3 examples or so of each widget.

-Core probably doesn't get any examples
ratatui proper probably mostly moves to the root
-each-backend might be worth having a small example of the basic concepts for each

examples/README.md
joshka: Mostly I copied this, and yep it can be simplified, but how about we focus on the content parts of doing this at some later point, and get the organization in place for now?

examples/README.md
joshka:
```suggestion
This is the original demo example from the main README. It is available for each of the backends.
[Source](demo/).
```

examples/README.md
joshka: Sorry - rephrase. I meant, the rationale for putting some examples in the ratatui-widgets/examples folder and some in other folders is that the ones in -widgets will be scraped to provide info for the -widget docs. I suspect also these will show up in the re-exported docs too.

In addition to that, having a top level examples folder is easy to find for people browsing the repo on github.

This should be under /examples/apps in my bikeshed.

orhun: Gotcha, that makes sense. I was concerned about the name "examples" but nvm, we can mention that `examples` are actually "demos" in the `examples/README.md` or somewhere.

orhun (approved):

PR #1525, opened 2024-11-27
chore: prepare alpha modularization release

joshka:
This is just a PR to show what's being published in the first modularization -alpha release. It's a manual publish to ensure that all the crates are properly setup and to set a baseline for comparison in future release-plz checks etc.

This does not update / check the git-cliff setup / changelog

Part of: #1388

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -28,9 +28,9 @@ dependencies = [
 
 [[package]]
 name = "allocator-api2"
-version = "0.2.18"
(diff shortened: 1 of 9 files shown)


orhun (approved):

joshka:
I'm going to merge this as-is without the changelog update. Mostly so that there's a tagged version in the repo that aligns with what was in the alpha release.

PR #1529, opened 2024-11-30
fix(scrollbar): check for area.is_empty() before rendering 

farmeroy:
This adds the `area.is_empty()` back into the scrollbar render method. Without it, the widget panics if the height is 0.
To reproduce this error, you can open the scrollbar demo2 in `/examples` (on the main branch) and continuously reduce the vertical height of your terminal window until it panics.

I also added a test, but I wasn't sure how to write the test correctly, currently there is no assert statement and the test suite will simply panic if the call to `is_empty()` is removed. 
You can also see that in the test, the buffer has a non-zero area. If the buffer area is 0, I don't think this error occurs. 

I was thinking, might make more sense if `scrollbar_area` returned a result instead of exiting with expect? 

Happy to make any changes.

### DIFF
--- a/ratatui-widgets/src/scrollbar.rs
+++ b/ratatui-widgets/src/scrollbar.rs
@@ -500,19 +500,17 @@ impl StatefulWidget for Scrollbar<'_> {
     type State = ScrollbarState;
 
     fn render(self, area: Rect, buf: &mut Buffer, state: &mut Self::State) {
-        if state.content_length == 0
-            || area.is_empty()
-            || self.track_length_excluding_arrow_heads(area) == 0
-        {
+        if state.content_length == 0 || self.track_length_excluding_arrow_heads(area) == 0 {
             return;
         }
 
-        let mut bar = self.bar_symbols(area, state);
-        let area = self.scrollbar_area(area);
-        for x in area.left()..area.right() {
-            for y in area.top()..area.bottom() {
-                if let Some(Some((symbol, style))) = bar.next() {
-                    buf.set_string(x, y, symbol, style);
+        if let Some(area) = self.scrollbar_area(area) {
+            let mut bar = self.bar_symbols(area, state);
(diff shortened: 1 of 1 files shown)


joshka (commented):
The other way to deal with this would be let the scrollbar_area method return Option<Rect>, then the iteration below is something like
```rust
for (Some((symbol, style)), area) in zip(bar, area.flat_map(Rect::cols).flat_map(Rect::columns()) {
    buf.set_string(area.x, area.y, symbol, style);
}
```

Happy if you do make this change, but will merge it without if you don't want to.

joshka (approved):

orhun (approved):

farmeroy:
Ok I can refactor the scrollbar_area to return an Option

farmeroy:
Ok, I couldn't quite wrap my head around using iterators for this :sweat_smile:  but now the scrollbar_area function returns an option and we only try to iterate over it if it is Some

joshka:
> Ok, I couldn't quite wrap my head around using iterators for this 😅 but now the scrollbar_area function returns an option and we only try to iterate over it if it is Some

No problem. I changed it to this to remove an extra for loop layer:
```rust
        if let Some(area) = self.scrollbar_area(area) {
            let areas = area.columns().flat_map(Rect::rows);
            let bar_symbols = self.bar_symbols(area, state);
            for (area, bar) in areas.zip(bar_symbols) {
                if let Some((symbol, style)) = bar {
                    buf.set_string(area.x, area.y, symbol, style);
                }
            }
        }
```
I'm not sure if there's an easy way to get rid of that last `if let Some`, calling `flatten()` seemed to render wrong, so I just left it for now.

joshka:
Thanks for the PR and fix. :)

farmeroy:
> Thanks for the PR and fix. :)

Thanks for cultivating such a great environment to contribute to!

PR #1530, opened 2024-11-30
chore: move terminal types to ratatui-core

joshka:
- Move Terminal, TerminalOptions, ViewPort, CompletedFrame, Frame to
  ratatui-core crate
- Move render_widget_ref() and render_stateful_widget_ref() to extension
  trait (FrameExt) due as the Ref types are unstable and kept in the
  main lib instead of -core
- Fix rustdoc errors / feature config issues

BREAKING CHANGE: to call `Frame::render_widget_ref()` or
`Frame::render_stateful_widget_ref()` you now need to import the
FrameExt trait from `ratatui::widgets` and enable the
`unstable-widget-ref` feature.

### DIFF
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -12,6 +12,7 @@ This is a quick summary of the sections below:
 
 - [Unreleased](#unreleased)
   - The `From` impls for backend types are now replaced with more specific traits
+  - `FrameExt` trait for `unstable-widget-ref` feature
 - [v0.29.0](#v0290)
   - `Sparkline::data` takes `IntoIterator<Item = SparklineBar>` instead of `&[u64]` and is no longer const
   - Removed public fields from `Rect` iterators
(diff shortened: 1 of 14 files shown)


orhun (approved):
LGTM!

I pushed a change to breaking changes document. Feel free to tweak it.

Also, atomic commits for changes that touches upon different areas helps greatly with the review. e.g. this PR could have been 3 commits during review and then we squash it before merge. Not a big deal though, just as an aside for the future :)

orhun:
Created #1531 for fixing the CI.

joshka:
> Also, atomic commits for changes that touches upon different areas helps greatly with the review. e.g. this PR could have been 3 commits during review and then we squash it before merge. Not a big deal though, just as an aside for the future :)

Yup - generally agree. It started small and then I fixed just one more thing and another... I missed the offramp :D

PR #1531, opened 2024-11-30
chore(xtask): check lints for only library targets

orhun:
Fixes the CI.
Also makes it possible to filter workspace packages by their targets which might be useful in the future (e.g. when we want to retrieve all the binary targets / examples, etc.)


joshka (approved):
Ha, I saw this after I wrote very much the same in #1530 :D

PR #1533, opened 2024-11-30
docs(widgets): add example for canvas

orhun:
related #1512


ratatui-widgets/examples/canvas.rs
```diff
@@ -0,0 +1,106 @@
...
+        .x_bounds([-180.0, 180.0])
+        .y_bounds([-90.0, 90.0])
+        .marker(Marker::Braille)
+        .paint(|ctx| {
+            ctx.draw(&Map {
+                resolution: MapResolution::High,
+                color: Color::White,
+            });
+            ctx.layer();
+            ctx.draw(&Line {
```
joshka, line 75: nit: does the new() constructor make this shorter vertically?

joshka (approved):
Idea, how about marking a few points that align with various cities (like Paris for Remy, Seattle, Toronto, your town in Turkey)?

ratatui-widgets/examples/canvas.rs
orhun: Yup, we should probably add that constructor to other canvas structs as well. (`Rectangle::new` etc.)

orhun: #1538

PR #1534, opened 2024-11-30
docs(widget): Improve ScrollbarState documentation

kardwen:


kardwen:
I created this PR as draft because I almost forgot about it. I will refine it tomorrow

ratatui-widgets/src/scrollbar.rs
```diff
@@ -451,7 +471,11 @@ impl ScrollbarState {
         self
     }
 
-    /// Sets the items' size.
+    /// Sets the viewport size.
+    ///
+    /// This needs to be set if the viewport size differs
+    /// from the track length, which may be the case if begin
+    /// and end symbols are set.
```
joshka, line 478: Can you expand on what you're saying with this a bit more. I want to make sure I understand what's being said a bit better.

Nit: wrap at 100 chars (if using vscode, consider installing https://marketplace.visualstudio.com/items?itemName=stkb.rewrap).

ratatui-widgets/src/scrollbar.rs
```diff
@@ -136,15 +136,32 @@ pub enum ScrollbarOrientation {
...
 /// │   second item ║
 /// └───────────────┘
 /// ```
 ///
 /// If you don't have multi-line content, you can leave the `viewport_content_length` set to the
 /// default and it'll use the track size as a `viewport_content_length`.
+///
+/// Note that this scrollbar behaves differently to what you might be used to for example
+/// from your web browser in that it scrolls when selecting an item,
+/// regardless of whether all content fits into the current viewport.
```
joshka, line 164: This is a bit ambiguous. When reading this for the first time, I don't understand what you mean by "it scrolls", also there's no selected item in the scrollbar, only a position. While it might be possible to infer this from the context, I think it would be useful to the language align with the names of the properties.

The flexibility of the scrollbar is that it can be used for dealing with either items or lines, but it's the app that chooses how that works. That makes it difficult to document.

joshka (commented):
There's definitely a deficiency in the way this is documented, so please don't get dismayed by any of my questions / push back here. As a newer user, you're in the best position to help improve the docs (due to the curse of knowledge), so please help us understand what needs to improve. We appreciate it a lot.

ratatui-widgets/src/scrollbar.rs
kardwen: Yes, that's also not precise enough. I think I have equated it to the selected item and not the viewport position. Maybe providing an example would be helpful?

kardwen: The table example also behave likes this, so we should propably introduce another example.

https://github.com/ratatui/ratatui/blob/main/ratatui/examples/table.rs

joshka:
> Another thing I observed is that occasionally the scrollbar handle changes its size by one when moving it (looks a bit like a caterpillar crawling 😄).

This is reasonable if you think about it in terms of how the scrollbar should round.
Example. Scrollbar viewport (no arrows) = 10. Showing some amount which maps to [0.2..2.4]. When rounded that will display a bar from [0..2] Move it a little to [0.4..2.6] and now that rounds instead to [0..3], move it a bit more to [0.6, 2.8] and you get [1..3]. Caterpiller effect.

Something we could do to fix that is make the scrollbar smarter and use partial bars. This would only be possible in the terminal with some symbol sets - 1/8th blocks, quadrant blocks, braille, and the symbols for legacy computing block stuff (sextant / octant). This could be worth considering, but it would add a lot of complexity to the implementation. Because of that I'd suggest experimenting with something like that outside of the Ratatui lib first and then bringing it in paying attention to how previous software works.

kardwen:
I think that is a tradeoff whether you prefer to have a constant handle size or accurate representation of position.
I also think half block would be to complex for the supposed benefit, will they even render correctly everywhere?

joshka:
> I think that is a tradeoff whether you prefer to have a constant handle size or accurate representation of position. I also think half block would be to complex for the supposed benefit, will they even render correctly everywhere?

Another thing to note here is that not rounding like this has some more problems with the way that it displays the bar when positioned close to the end of the bar. If you round to a constant size rather than constant position, then it leads to oddness in the last position. There's perhaps a design that would account for both, but the complexity to maintain it seems intuitively like it would be higher than the benefit (unless written really well).

ratatui-widgets/src/scrollbar.rs
joshka: I misspoke a bit here. What I'm after when asking for expansion here is less "why are you doing this" and more "can you rephrase this a bit so that the unclear points I'm asking about get a bit more clear". I'm not entirely sure what the language should say here, but I know that I don't quite understand exactly this language, so I'm looking for you to help simplify that so that someone without the context of how this works would understand it.

Some parts that seem unknown here (just from reading the docs, not the source). What happens if I don't. Why is it not just handled? Isn't there some calculation that happens that excludes the end symbols? etc.

Feel free to respond in the PR comments to work out how best to phrase this, or if you're confident about what you want to say here, make changes to the PR directly (whichever works best with your workflow).

ratatui-widgets/src/scrollbar.rs
joshka: Same comment as above regarding updating the comments here. I'm looking for suggestions from you (as someone digging into this code and likely the current expert on how it works) to instruct on how best to summarize and communicate what you've learnt. The problem with the current ambiguity is that it states only how something doesn't work by with an ambiguous comparison, without then really clearly stating how it does behave.

orhun:
Ping @kardwen

kardwen:
I'll review and summarise the changes tomorrow as I should have some time on a train journey.

joshka:
Marking as a draft for now, waiting on more feedback from @kardwen on the various comments

orhun:
rebased on main

PR #1535, opened 2024-12-01
"Bell" function abstracts ^G

### LINKED ISSUE #1522: I want to beep
## Problem

I am implementing a terminal text editor. At times, I must alert the user they did something wrong. Terminals have a builtin "bell" capability that does something system-defined like beep or flash the titlebar. I can activate this by saying `print!("\x07");`.

However, **this is wrong!** I am using ratatui which supports multiple backends. As far as I know all those backends are either targeting VT100 emulators or Microsoft Windows (both of which support `\x07`) but in the future, I could imagine a backend targeting something "weird" like the web DOM.

## Solution

The [Action](https://docs.rs/terminal/latest/terminal/enum.Action.html) Enum should have a new member `Beep` (or possibly `Bell`). In existing backends, this should just `print!("\x07");`.

## Alternatives

Things I think you should not do:
- Name the Action something generic like 'Alert'. It should be unambiguous the new action maps to the "ASCII Bell" terminal verb (which has developed a distinct semantic meaning among programmers unique from the literal bell that used to be attached to the terminals).
- Put additional data such as pitch on the bell/beep verb. Although some methods of beeping can have a pitch, this is the *terminal verb* bell which is just a single signal.

## Additional context

Crossterm has a function [crossterm_treminal_ring_bell](https://docs.rs/libcrossterm/latest/crossterm/fn.crossterm_terminal_ring_bell.html). If ratatui doesn't want to add a bell verb, I could potentially call this directly. However I think it would be nice if my program relied on ratatui for terminal interactions and didn't have to take a dependency on a single backend. (At any rate the crossterm backend can call this function.)

mcclure:
Resolves #1522 . As discussed.

Adds a `bell()` to Terminal. Currently on all backends just prints a single `^G` (ASCII BEL) to stdout, but hypothetically on some future backend could do something different.

Things I'm not sure about:

- Right now, per types, it is possible for the bell to fail. I'm not sure this makes sense because it is documented that the bell function may do nothing at all while still returning success. I was thinking maybe if some future backend has the ability to fail and know it's failing it could return an error, but it's actually really irritating to have to `let _ =` the bell function to silence the error. I think maybe I should change the return type to plain `()`.
- Right now, per types, the bell function is &mut, even though the default implementation does not need &mut. This is because I don't know how to predict whether a future backend implementin gbell. might need &mut.
- I suborned the scrollbar example to test the bell function (whenever you press a key that doesn't do anything). However I don't know how to test whether the change works on all backends (I feel certain it must, but this should probably be tested anyway).

### DIFF
--- a/ratatui/src/terminal/terminal.rs
+++ b/ratatui/src/terminal/terminal.rs
@@ -820,6 +820,13 @@ where
         }
         Ok(())
     }
+
+    /// Alert the user in the manner of the VT100 "Bell" function.
+    /// Depending on the backend, the terminal app, and the user's settings,
+    /// this may play a beep, flash the screen, or do nothing at all.
+    pub fn bell(&mut self) -> io::Result<()> {
+        self.backend.bell()
+    }
 }
 
 fn compute_inline_size<B: Backend>(
--- a/ratatui-core/src/backend.rs
+++ b/ratatui-core/src/backend.rs
@@ -370,6 +370,18 @@ pub trait Backend {
         region: std::ops::Range<u16>,
         line_count: u16,
     ) -> io::Result<()>;
+
+    /// Alert the user in the manner of the VT100 "Bell" function.
+    ///
+    /// The exact effect is implementation-defined, and returning Ok(())
+    /// does not guarantee anything happened.
+    /// This default implementation is "good enough" on all known terminals.
+    /// If a backend categorically does not support this, it should override
+    /// and return Err(Errorkind::Unsupported).
+    fn bell(&mut self) -> io::Result<()> {
+        print!("\x07"); // ASCII BEL to STDOUT
+        Ok(())
+    }
 }
 
 #[cfg(test)]
--- a/ratatui/examples/scrollbar.rs
+++ b/ratatui/examples/scrollbar.rs
@@ -78,6 +78,10 @@ impl App {
                                 .horizontal_scroll_state
                                 .position(self.horizontal_scroll);
                         }
+                        // If an unrecognized letter is entered, beep to alert.
+                        KeyCode::Char(_) => {
+                            let _ = terminal.bell();
+                        }
                         _ => {}
                     }
                 }


ratatui/src/terminal/terminal.rs
```diff
@@ -820,6 +820,13 @@ where
         }
         Ok(())
     }
+
+    /// Alert the user in the manner of the VT100 "Bell" function.
+    /// Depending on the backend, the terminal app, and the user's settings,
+    /// this may play a beep, flash the screen, or do nothing at all.
```
joshka, line 826: Same comment re: wrapping.

Leave a line break between the first line and the rest. For doc comments, the first line shows up in the module summary table etc., and the detail only shows up in the main part.

joshka (changes requested):
> Right now, per types, the bell function is &mut, even though the default implementation does not need &mut. This is because I don't know how to predict whether a future backend implementin gbell. might need &mut.

Perhaps check how this pans out by implementing the backend specific changes that write to the backend writers?

Re: returning an error, the same thing goes. I suspect once you're looking at a write call that this might more obviously return an io::Error, which makes returning an error the right thing to do here.

For testing this, it might make sense to make a small simple example that has an implementation for each backend. Something like:
```rust
fn main() {
    #[cfg(feature = ...)
    let terminal = Terminal::new(...Backend...);
    #[cfg(feature = ...)
    let terminal = Terminal::new(...Backend...);
    terminal.bell();
}
```

Be aware that https://github.com/ratatui/ratatui/pull/1530 (merging to main very shortly) moves the terminal struct to ratatui-core, so this PR will need to be rebased.

mcclure:
@joshka , thanks for the feedback, I will wait for 1530 to land and then do another pass.

joshka:
> @joshka , thanks for the feedback, I will wait for 1530 to land and then do another pass.

I merged this just now.

orhun:
Marking as a draft for now, @mcclure still interested in finishing up this one? :)

orhun:
Closing due to inactivity

Issue #1514, opened 2024-11-25
multiline list items + scroll_padding


raylu:
## Description
when selecting a large, multi-line item in a list widget with `scroll_padding`, smaller items sometimes don't render until the next frame

## To Reproduce
here's some code that makes 26 list items. most items are 5 lines long, but every fifth item is 1 line long
```rs
use crossterm::{
	event::{
		self, Event,
		KeyCode::{self, Char},
		KeyEvent,
	},
	execute,
	terminal::{disable_raw_mode, enable_raw_mode, EnterAlternateScreen, LeaveAlternateScreen},
};
use std::{
	error::Error,
	io::{self, Stdout},
};
use tui::{
	backend::CrosstermBackend,
	layout::{Rect, Size},
	style::{Color, Style},
	widgets::{List, ListItem, ListState},
	Frame, Terminal,
};

fn main() {
	let alphabet: Vec<char> = ('a'..='z').into_iter().collect();
	let list_items: Vec<ListItem> = alphabet.into_iter().enumerate().map(|(i, c)| {
		let mut line = c.to_string().repeat(10);
		line.push('\n');
		line.repeat(if i % 5 == 0 { 1 } else { 5 }).into()
	}).collect();
	let app = App {
		list_items, list_state: ListState::default(),
	};
	let mut term = setup().unwrap();
	let res = run_app(&mut term, app);

	teardown(&mut term);
	if let Err(err) = res {
		println!("{:?}", err)
	}
}

pub struct App {
	list_items: Vec<ListItem<'static>>,
	list_state: ListState,
}

type CrosstermTerm = Terminal<CrosstermBackend<Stdout>>;

pub fn setup() -> Result<CrosstermTerm, Box<dyn Error>> {
	enable_raw_mode()?;
	let mut stdout = io::stdout();
	execute!(stdout, EnterAlternateScreen)?;
	let backend = CrosstermBackend::new(stdout);
	Ok(Terminal::new(backend)?)
}

pub fn teardown(terminal: &mut CrosstermTerm) {
	_ = disable_raw_mode();
	_ = execute!(terminal.backend_mut(), LeaveAlternateScreen);
	_ = terminal.show_cursor();
}

pub fn run_app(terminal: &mut CrosstermTerm, mut app: App) -> Result<(), Box<dyn Error>> {
	loop {
		terminal.draw(|frame| ui(frame, &mut app))?;
		if let Event::Key(key) = event::read()? {
[...]

PR #1540, opened 2024-12-03
refactor: rearrange selection_spacing code

raylu:
move the `if` outside the loop
should be a no-op


joshka (approved):
This is a fairly long method (by my preferences at least). Currently reading / maintaining it means reading every line of the method. I wonder if there's a good refactoring that makes the rendering more obvious. E.g., smallish methods that show how each part of the rendering relates to each other. Happy to merge this as-is, but you're welcome to continue simplifying if you'd like.

raylu:
please merge as-is. just random cleanup as I was trying to fix #1514

PR #1541, opened 2024-12-03
fix(crossterm): terminal should keep Bold when removing Dim

MarSik:
The Dim removal should behave the same as the logic for Bold removal that sends NormalIntensity sequence and then restores Dim when needed.

### DIFF
--- a/ratatui-crossterm/src/lib.rs
+++ b/ratatui-crossterm/src/lib.rs
@@ -405,10 +405,18 @@ impl ModifierDiff {
         if removed.contains(Modifier::REVERSED) {
             queue!(w, SetAttribute(CrosstermAttribute::NoReverse))?;
         }
-        if removed.contains(Modifier::BOLD) {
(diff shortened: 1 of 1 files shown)


MarSik:
Ah, the code changed since the last release. Oh well, let me fix that.

orhun (approved):

joshka (changes requested):
Can you please move this closer to the bold part and add a comment about why this is required something like:
```
// removing bold/dim has to start from normal intensity and add back any attributes
if removed.contains bold / dim {
	set normal intensity;
	if contains bold set bold;
	if contains dim set dim;
}
```

MarSik:
> Can you please move this closer to the bold part and add a comment about why this is required something like:
> 
> ```
> // removing bold/dim has to start from normal intensity and add back any attributes
> if removed.contains bold / dim {
> 	set normal intensity;
> 	if contains bold set bold;
> 	if contains dim set dim;
> }
> ```

Something like this?

joshka:
LGTM. Thanks for the fix

PR #1542, opened 2024-12-03
docs(widgets): add example for list

orhun:
related #1512

### DIFF
--- a/ratatui-widgets/examples/list.rs
+++ b/ratatui-widgets/examples/list.rs
@@ -34,14 +34,15 @@ fn main() -> Result<()> {
 
 /// Run the application.
 fn run(mut terminal: DefaultTerminal) -> Result<()> {
-    let mut selected_index = 0;
+    let mut list_state = ListState::default();
+    list_state.select_first();
     loop {
-        terminal.draw(|frame| draw(frame, selected_index))?;
+        terminal.draw(|frame| draw(frame, &mut list_state))?;
         if let Event::Key(key) = event::read()? {
             match key.code {
                 KeyCode::Char('q') => break Ok(()),
-                KeyCode::Down | KeyCode::Char('j') => selected_index = (selected_index + 1) % 4,
-                KeyCode::Up | KeyCode::Char('k') => selected_index = (selected_index + 3) % 4,
+                KeyCode::Down | KeyCode::Char('j') => list_state.select_next(),
+                KeyCode::Up | KeyCode::Char('k') => list_state.select_previous(),
                 _ => {}
             }
         }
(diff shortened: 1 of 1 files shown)


joshka (approved):
I think this probably should be interactive to demonstrate how the state part works. WDYT?

orhun:
Sounds good. Are you fine with using plain variables instead of a state struct etc.? I will try to cook up something basic.

joshka:
yep

ratatui-widgets/examples/list.rs
```diff
@@ -0,0 +1,105 @@
...
+/// Run the application.
+fn run(mut terminal: DefaultTerminal) -> Result<()> {
+    let mut selected_index = 0;
+    loop {
+        terminal.draw(|frame| draw(frame, selected_index))?;
+        if let Event::Key(key) = event::read()? {
+            match key.code {
+                KeyCode::Char('q') => break Ok(()),
+                KeyCode::Down | KeyCode::Char('j') => selected_index = (selected_index + 1) % 4,
+                KeyCode::Up | KeyCode::Char('k') => selected_index = (selected_index + 3) % 4,
```
joshka, line 44: Same comment as from scrollbar example. How about storing the list state as a variable and passing it to the draw function. This would better align with real world usage which suggests that the state vars are kept between frames.

orhun: Very good idea, done in the last commit.

ratatui-widgets/examples/list.rs
```diff
@@ -0,0 +1,103 @@
...
+    let list = List::new(items)
+        .style(Color::White)
+        .highlight_style(Style::new().yellow().italic())
+        .highlight_symbol("> ")
+        .scroll_padding(1)
+        .direction(ListDirection::BottomToTop)
+        .repeat_highlight_symbol(true);
+
+    let mut state = ListState::default();
+    state.select_first();
```
joshka, line 100: What about this one?

orhun: I intentionally left it because I didn't want to control both of these lists using the same state. Thought it might be a more explicit example of how `ListState` is being constructed and how it can be used in a smaller scope.

Also, the directions are different so it looks confusing when you select next/previous for both of them. One of them goes up and the other goes down.

I'm open to improving this!

joshka: A few ideas:
- Simply to just one list (perhaps too simple?)
- Have two states and call next() on one, prev() on the other
- Have an enum to choose which list to pass input to (I think this borders on too complex for the example)
- Leave it as is (this is also fine)

orhun: > Simply to just one list (perhaps too simple?)

Yeah I think one list is too simple. I usually tried to add multiple widgets for each example (`Table` is an exception though).

> Have two states and call next() on one, prev() on the other
> Have an enum to choose which list to pass input to (I think this borders on too complex for the example)

Yeah, I think both of them would complicate the example a bit.

> Leave it as is (this is also fine)

I choose this one! :D

joshka (approved):

PR #1549, opened 2024-12-04
build: Remove cargo lint

joshka:
Duplicate crate lint is too noisy and sensitive to upstream changes


joshka:
See https://github.com/ratatui/ratatui/actions/runs/12168652987/job/33939894968?pr=1546

PR #1551, opened 2024-12-04
docs: add input form example

joshka:
Demonstrates how to manage the cursor and input focus with a simple
form. Uses an enum to track the current input field and passes the
input events to the active field.

This is similar to the json tutorial on the website, but a bit simpler

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1425,6 +1425,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b248f5224d1d606005e02c97f5aa4e88eeb230488bcc03bc9ca4d7991399f2b5"
 
 [[package]]
+name = "input-form"
+version = "0.0.0"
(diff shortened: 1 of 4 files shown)


orhun (approved):
LGTM!

Do you want to plug 3rd party libs (e.g. [focusable](https://github.com/joshka/focusable), [rat-focus](https://github.com/thscharler/rat-focus), etc.) somewhere in the docs maybe? (e.g. "Also, check out xyz")

joshka:
> LGTM!
> 
> Do you want to plug 3rd party libs (e.g. [focusable](https://github.com/joshka/focusable), [rat-focus](https://github.com/thscharler/rat-focus), etc.) somewhere in the docs maybe? (e.g. "Also, check out xyz")

I like this idea, because it gives people more places to jump off to. Probably in this app's readme would be a good place for that info.

PR #1545, opened 2024-12-03
docs(widgets): add example for scrollbar

orhun:
related #1512

PR #1552, opened 2024-12-05
feat(scrollbar): support retrieving the current position of state

orhun:
As of now it is possible to change the position of the Scrollbar but not possible to retrieve the position for further use. e.g.

```rust
let mut state = ScrollbarState::default();
state.next();
```

This commit adds a new method "`current_position`" (since `position` is already taken by the fluent setter) for that purpose:

```rust
let index = state.get_position(); // yay
```

See #1545 for the concrete usage of this.

### DIFF
--- a/ratatui-widgets/src/scrollbar.rs
+++ b/ratatui-widgets/src/scrollbar.rs
@@ -494,6 +494,12 @@ impl ScrollbarState {
             }
         }
     }
+
+    /// Returns the current position within the scrollable content.
+    #[must_use = "returns the current position within the scrollable content"]
+    pub const fn get_position(&self) -> usize {
+        self.position
+    }
 }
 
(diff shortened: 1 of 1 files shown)


ratatui-widgets/src/scrollbar.rs
```diff
@@ -494,6 +494,12 @@ impl ScrollbarState {
             }
         }
     }
+
+    /// Returns the current position within the scrollable content.
+    #[must_use = "returns the current position within the scrollable content"]
+    pub const fn current_position(&self) -> usize {
```
joshka, line 500: What about `get_position` here?

orhun: Sure, updated

joshka (approved):

PR #1559, opened 2024-12-09
docs(widgets): add example for tabs

orhun:
related #1512 

Also removes the tabs example from ratatui crate since it overlaps with
this new example in terms of functionality and it was not following the
general theme of other examples.


joshka:
I think the existing tabs example was nice because it was pretty. Making a note here about the removal, so that we can do a more comprehensive app using that style at some point (not sure exactly what that might look like though.

joshka (approved):

PR #1567, opened 2024-12-10
chore(examples): add weather demo app

orhun:
related to #1512 

As discussed, moves barchart example from ratatui to apps under the name of "weather" with some edits.

This feels a bit too basic compared to other apps so I'm not sure if there is anything else that we can add to it. I looked into fetching the weather data dynamically but that is probably going to complicate things more than we want.

I think it's fine.


joshka (approved):

joshka:
Dynamic weather fetch would be a good async / threaded / sync example.

PR #1571, opened 2024-12-14
chore(examples): add calendar explorer demo app

orhun:
Related to #1512 

As discussed, this moves the calendar example from ratatui to app examples as an "explorer" example. It also adds interactivity where you can press Enter to toggle between different styles of calendars.


joshka:
I tweaked a bunch of things:
- made it interactive (select a date, scroll through months)
- replaced the style index with an enum
- much tidying (consts and things)
- better for loop on the year rendering
etc.

joshka (approved):
Feel free to merge / improve more

examples/apps/calendar-explorer/src/main.rs
```diff
@@ -0,0 +1,241 @@
...
+                .show_surrounding(Style::new().dim())
+                .show_weekdays_header(Style::new().bold().green()),
+            StyledCalendar::MonthHeader => calendar.show_month_header(Style::new().bold().green()),
+            StyledCalendar::MonthAndWeekdaysHeader => {
+                calendar.show_weekdays_header(Style::new().bold().dim().light_yellow())
+            }
+        };
+        frame.render_widget(calendar, area);
+    }
+}
```
orhun, line 187: There were a bit of code repetition but my reasoning for having separate functions was making them easy to copy/paste.

Consider this scenario:

1. User runs the calendar explorer.
2. Likes one of the styles and wants to incorporate that in their app.
3. Simply copies the function to their app.

I guess you could argue that they should just _learn the API_ ™️ to make proper use of the widgets. But most people are lazy and just want to build something quickly.

Also, this is an "app example" so it should be fine. It is still an example though...

What do you think?

orhun (commented):
The refactor looks good, I just have some questions...

examples/apps/calendar-explorer/src/main.rs
joshka: For the explorer apps, I'm less concerned about copy paste, but perhaps inlining the first calendar line might be sufficient here. Mostly I didn't like that it was random methods labelled example1, example2, ...

https://github.com/ratatui/ratatui/pull/1571/commits/315393e2e12e6c4fe056be14664168daf5b519eb

examples/apps/calendar-explorer/src/main.rs
```diff
@@ -0,0 +1,241 @@
...
+        terminal.draw(|frame| render(frame, calendar_style, selected_date))?;
+        if let Event::Key(key) = event::read()? {
+            if key.kind == KeyEventKind::Press {
+                match key.code {
+                    KeyCode::Char('q') => break Ok(()),
+                    KeyCode::Char('s') => calendar_style = calendar_style.next(),
+                    KeyCode::Char('n') | KeyCode::Tab => selected_date = next_month(selected_date),
+                    KeyCode::Char('p') | KeyCode::BackTab => {
+                        selected_date = previous_month(selected_date)
+                    }
```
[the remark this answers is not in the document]
orhun: nvm, I was confusing months with years. Time flies fast

Issue #1158, opened 2024-05-31
Add Stylize methods directly to Style


joshka:
## Problem
I'd like to be able to write:
```rust
const FOO: Style = Style::new().green().bold();
```
But trait methods can't be const.

## Solution
Add the Stylize methods directly to the Style type rather than implementing them via Stylize.

g4titanx:
where can the Style type be found in the codebase?

joshka:
I'd like to get buy-in / thoughts from the other maintainers before implementing this. I've removed the good first issue label for now.

g4titanx:
alright

PR #1572, opened 2024-12-16
feat!: implement stylize methods directly on Style

joshka:
This makes it possible to create constants using the shorthand methods.

```rust
const MY_STYLE: Style = Style::new().blue().on_black();
```

Rather than implementing Styled for Style and then adding extension
methods that implement the Stylize shorthands, this implements the
methods as const functions directly on Style.

BREAKING CHANGE: `Style` no longer implements `Styled`. Any calls to
methods implemented by the blanket implementation of Stylize are now
defined directly on Style. Remove the Stylize import if it is no longer
used by your code.

The `reset()` method does not have a direct replacement, as it clashes
with the existing `reset()` method. Use `Style::reset()` rather than
`some_style.reset()`

Fixes: #1158

### DIFF
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -10,7 +10,7 @@ GitHub with a [breaking change] label.
 
 This is a quick summary of the sections below:
 
-- [Unreleased](#unreleased)
+- [v0.30.0 Unreleased](#v0300-unreleased)
   - The `From` impls for backend types are now replaced with more specific traits
   - `FrameExt` trait for `unstable-widget-ref` feature
   - `List::highlight_symbol` now accepts `Into<Line>` instead of `&str`
@@ -83,7 +83,28 @@ This is a quick summary of the sections below:
   - MSRV is now 1.63.0
   - `List` no longer ignores empty strings
 
-## Unreleased (0.30.0)
+## v0.30.0 Unreleased
+
+### `Style` no longer implements `Styled` ([#1572])
+
+[#1572]: https://github.com/ratatui/ratatui/pull/1572
+
+Any calls to methods implemented by the blanket implementation of `Stylize` are now defined directly
+on `Style`. Remove the `Stylize` import if it is no longer used by your code.
+
+```diff
+- use ratatui::style::Stylize;
+
+let style = Style::new().red();
+```
+
+The `reset()` method does not have a direct replacement, as it clashes with the existing `reset()`
+method. Use the `Style::reset()` method instead.
+
+```diff
+- some_style.reset();
++ Style::reset();
+```
 
 ### Disabling `default-features` suppresses the error message if `show_cursor()` fails when dropping `Terminal` ([#1794])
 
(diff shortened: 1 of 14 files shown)


BREAKING-CHANGES.md
```diff
@@ -75,7 +75,28 @@ This is a quick summary of the sections below:
...
+let style = Style::new().red();
+```
+
+The `reset()` method does not have a direct replacement, as it clashes with the existing `reset()`
+method. Use the `Style::reset()` method instead.
+
+```diff
+- some_style.reset();
++ Style::reset();
+```
```
orhun, line 99: If I understood this correctly, this doesn't effect e.g. `"foo".reset()` so it should be fine.

Maybe we should clarify that? Please correct me if I'm wrong.

ratatui-core/src/style/stylize.rs
```diff
@@ -114,17 +112,33 @@ impl fmt::Debug for ColorDebug {
...
+            }
+        }
+    };
+
+    ( $vis:vis const $color:ident() -> $ty:ty ) => {
+        paste::paste! {
             #[doc = "Sets the foreground color to [`" $color "`](Color::" $color:camel ")."]
             #[must_use = concat!("`", stringify!($color), "` returns the modified style without modifying the original")]
-            fn $color(self) -> T {
+            $vis const fn $color(self) -> $ty {
```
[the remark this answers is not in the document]
orhun: Yeah, I don't think so too. We can create two macros and call the other one from here if we **really** want to avoid repetition though.

I think it's fine like this tbh.

BREAKING-CHANGES.md
joshka: There's a reset method on Stylize: `fn reset(self) -> T`
There's a reset method on Style: `fn reset() -> Style`
There's a blanket implementation of Stylize for T: Styled
Style no longer implements Styled, so no longer implements Stylize.

I'd guess that this has a low risk of impacting any real world code. It's mainly included in case it does and to set expectations that we track even minor problems like this.

orhun: got it, I guess that is a "yes" to my initial assumption :D

orhun (approved):

orhun:
unrelated: I wonder how did you hit this case and decided to fix it this way. I'm generally curious about which part of the codebase you are focusing as of now - to align myself with it :)

BREAKING-CHANGES.md
joshka: yep

joshka:
> unrelated: I wonder how did you hit this case and decided to fix it this way. I'm generally curious about which part of the codebase you are focusing as of now - to align myself with it :)

I was looking at the calendar styles. I pulled them out to constants initially, and wanted to use the shorthand. It's been at the back of my mind to address this for some time.

As a followup to this (and probably in this PR before merging), I think making all code that can use the shorthands would be worthwhile.

kdheepak (approved):
I like this change. I'm assuming this will also help alleviate the incorrect auto-import issue?

joshka:
> I like this change. I'm assuming this will also help alleviate the incorrect auto-import issue?

Yeah, I hope so.

orhun (approved):

Issue #1355, opened 2024-09-06
Add support for Dash borders


0xfalafel:
## Problem

__Dash borders__, even if they are less common, have their use. And are part of the [Unicode Box Drawing](https://en.wikipedia.org/wiki/Box-drawing_characters#Box_Drawing) characters.

At the moment, Ratatui propose different type of borders. But no dashes.

## Solution

* Add __new symbols__ to [ratatui::symbols::border](https://docs.rs/ratatui/latest/ratatui/symbols/border/index.html) that reference the [U+250x box drawing characters](https://en.wikipedia.org/wiki/Box-drawing_characters#Box_Drawing).
* Add one or two __new entries__ in [ratatui::widgets::block::__BorderType__](https://docs.rs/ratatui/latest/ratatui/widgets/block/enum.BorderType.html) for __dash borders__.

## Alternatives

I wanted to use dashes as a border for [hextazy](https://github.com/0xfalafel/hextazy). So I have :

*  printed the `┆`  character to make the separation.
* Added the `┬` character as a title for the pane.

But this leads to 2 issues:

* It doesn't follow the color of the borders when the terminal has a light theme.
* When the window is smaller, the `┬` will not be aligned with the `┆` border.

## Additional context

I have written an [__hexadecimal editor__](https://github.com/0xfalafel/hextazy) that makes use of __dash borders__. I have a __workaround__, but this has some limitations.

__I can__ contribute to ratatui to __implement this feature__. Fixing the limitations of my workaround would probably be more complicated than implementing this feature in ratatui anyway.

![Hextazy](https://github.com/user-attachments/assets/07e03603-5778-4fde-9ccb-8940ec01e529)

joshka:
Seems reasonable to add this. What are the two entries?

As an immediate workaround, take a look at the [`Block::border_set()`](https://docs.rs/ratatui/latest/ratatui/widgets/block/struct.Block.html#method.border_set) method to allow you to specify a custom set of border characters. (This should be documented a little better).

0xfalafel:
I was thinking about `BorderType::Dash` for the `┊` (U+250A) character.

There are other dash characters: 

* `┆` (U+2506)
* `┇` (U+2507) 
* `┋` (U+250B) 

I don't really have an opinion about these. Maybe we can add one as well.

joshka:
Got it - looks like there's double, triple, quadruple dashes as well as heavy variants of some of these. These seem like good things to add in the same batch of changes.

[...]

PR #1573, opened 2024-12-16
feat(symbols): add dashed borders (#1355)

theotchlx:
Closes issue #1355 

- [x] Add dotted border symbols
- [x] Add new entries in BorderType for dash borders
- [x] Name the values closer to their names in the Unicode standards
- [x] Add the heavy and quad versions of the dotted borders

### DIFF
--- a/ratatui-core/src/symbols/border.rs
+++ b/ratatui-core/src/symbols/border.rs
@@ -426,10 +426,10 @@ mod tests {
                 top_right: "b",
                 bottom_left: "c",
                 bottom_right: "d",
-                vertical_left: "g",
-                vertical_right: "h",
-                horizontal_bottom: "i",
-                horizontal_top: "j",
+                vertical_left: "e",
+                vertical_right: "e",
+                horizontal_bottom: "f",
+                horizontal_top: "f",
             }
         );
     }
--- a/ratatui-core/src/symbols/border.rs
+++ b/ratatui-core/src/symbols/border.rs
@@ -18,6 +18,20 @@ impl Default for Set {
     }
 }
 
+// Helper function to convert a line set to a border set
+const fn from_line_set(line_set: line::Set) -> Set {
+    Set {
+        top_left: line_set.top_left,
+        top_right: line_set.top_right,
+        bottom_left: line_set.bottom_left,
+        bottom_right: line_set.bottom_right,
+        vertical_left: line_set.vertical,
+        vertical_right: line_set.vertical,
+        horizontal_top: line_set.horizontal,
+        horizontal_bottom: line_set.horizontal,
+    }
+}
+
 /// Border Set with a single line width
 ///
 /// ```text
@@ -102,16 +116,7 @@ pub const THICK: Set = Set {
 /// ╎xxxxx╎
 /// └╌╌╌╌╌┘
 /// ```
-pub const LIGHT_DOUBLE_DASHED: Set = Set {
-    top_left: line::LIGHT_DOUBLE_DASHED.top_left,
-    top_right: line::LIGHT_DOUBLE_DASHED.top_right,
-    bottom_left: line::LIGHT_DOUBLE_DASHED.bottom_left,
-    bottom_right: line::LIGHT_DOUBLE_DASHED.bottom_right,
-    vertical_left: line::LIGHT_DOUBLE_DASHED.vertical,
-    vertical_right: line::LIGHT_DOUBLE_DASHED.vertical,
-    horizontal_top: line::LIGHT_DOUBLE_DASHED.horizontal,
-    horizontal_bottom: line::LIGHT_DOUBLE_DASHED.horizontal,
-};
+pub const LIGHT_DOUBLE_DASHED: Set = from_line_set(line::LIGHT_DOUBLE_DASHED);
 
 /// Border Set with thick double-dashed border lines
 ///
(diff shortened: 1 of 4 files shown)


ratatui-core/src/symbols/border.rs
```diff
@@ -37,6 +37,25 @@ pub const PLAIN: Set = Set {
     horizontal_bottom: line::NORMAL.horizontal,
 };
 
+/// Border Set with a single dotted line width
+///
+/// ```text
+/// ┌─────┐
+/// │xxxxx│
+/// │xxxxx│
+/// └─────┘
```
joshka, line 46: Is this correct?

theotchlx: Nope! Fixed it, thank you very much :D!

theotchlx:
Ok I tested it in the layout example and it currently looks like this:

image

The above image is using the "┄" ("Box Drawings Light Triple Dash Horizontal") character for dotted horizontal borders, and the "┆" ("Box Drawings Light Triple Dash Vertical") character for dotted vertical borders.

Using a double dash for horizontal borders while keeping a triple dash for vertical borders gives the following result: ("╌" character)

image

But maybe it could instead be used to make a new "dashed border" style using a double dash for both horizontal and vertical borders: ("╌" and "╎" characters)

image

I can add both styles (dotted "┄┆" and dashed "╌╎") if you want. Or one hybrid ("╌┆" double-dash horizontal and triple-dash vertical).

theotchlx:
Personally, I find the mixed one looks better as a "dotted border", but then you don't have separate "triple-dashed" and "double-dashed" borders.  
Here is the box-drawing characters table for reference: [link to Wikipedia](https://en.wikipedia.org/wiki/Box-drawing_characters#Box_Drawing)

theotchlx:
Happy New Year! :)

orhun (approved):
Happy new year, this LGTM!

theotchlx:
Okay it's rebased :+1: :shipit:

joshka (commented):
I know that this is a bit of a late review / change, but thinking about this more, I'd prefer to name the values closer to their names in the Unicode standards. Keeping consistent naming makes it easier to find the values if you're already familiar with them, and understand what will be the outcome of some random code snippet you're reading.

https://www.unicode.org/charts/PDF/U2500.pdf

```
Light and heavy dashed lines
2504 BOX DRAWINGS LIGHT TRIPLE DASH HORIZONTAL
2505 BOX DRAWINGS HEAVY TRIPLE DASH HORIZONTAL
2506 BOX DRAWINGS LIGHT TRIPLE DASH VERTICAL
2507 BOX DRAWINGS HEAVY TRIPLE DASH VERTICAL
2508 BOX DRAWINGS LIGHT QUADRUPLE DASH HORIZONTAL
2509 BOX DRAWINGS HEAVY QUADRUPLE DASH HORIZONTAL
250A BOX DRAWINGS LIGHT QUADRUPLE DASH VERTICAL
250B BOX DRAWINGS HEAVY QUADRUPLE DASH VERTICAL
```

I'd also suggest adding the heavy and quad versions of the line at the same time as this.

So how does making this:
```rust
pub const LIGHT_TRIPLE_DASH_VERTICAL: &str = "┆";
pub const HEAVY_TRIPLE_DASH_VERTICAL: &str = "┇";
pub const LIGHT_QUADRUPLE_DASH_VERTICAL: &str = "┊";
pub const HEAVY_QUADRUPLE_DASH_VERTICAL: &str = "┋";
...
pub const LIGHT_TRIPLE_DASHED: Set = Set { ... }
pub const HEAVY_TRIPLE_DASHED: Set = Set { ... }
...
```

For consistency (and not in this PR), we could later also consider deprecating the existing constants to rename them to the unicode names too.

joshka:
I know that this 

> Personally, I find the mixed one looks better as a "dotted border", but then you don't have separate "triple-dashed" and "double-dashed" borders.

The trick with making the more balanced versions of this might be to define the double/triple/quad sets and then make a separate more balanced set. We do something like that with:
- https://docs.rs/ratatui/latest/ratatui/symbols/border/constant.ONE_EIGHTH_TALL.html
- https://docs.rs/ratatui/latest/ratatui/symbols/border/constant.ONE_EIGHTH_WIDE.html
- https://docs.rs/ratatui/latest/ratatui/symbols/border/constant.PROPORTIONAL_TALL.html
- https://docs.rs/ratatui/latest/ratatui/symbols/border/constant.PROPORTIONAL_WIDE.html

theotchlx:
Great ideas! I'll start working on it ;D

theotchlx:
Well that made me discover the "McGugan box technique". Amazing! :smile:

joshka:
> Well that made me discover the "McGugan box technique". Amazing! 😄

Yeah we were also pleasantly surprised by that one too. https://github.com/ratatui/ratatui/pull/831. I think we probably need to make the docs better about this at some point to make it more discoverable.

@willmcgugan's [article](https://www.willmcgugan.com/blog/tech/post/ceo-just-wants-to-draw-boxes/) about this was the impetus of this. I hope he doesn't mind that Ratatui named that particular border set approach after him :D.

theotchlx:
I added double, triple, quadruple, heavy/light borders (all that I could find), as well as some missing tests.

ratatui-core/src/symbols/border.rs
```diff
@@ -94,6 +94,120 @@ pub const THICK: Set = Set {
...
 
+/// Border Set with light double-dashed border lines
+///
+/// ```text
+/// ┌╌╌╌╌╌┐
+/// ╎xxxxx╎
+/// ╎xxxxx╎
+/// └╌╌╌╌╌┘
+/// ```
+pub const LIGHT_DOUBLE_DASHED: Set = Set {
```
joshka, line 105: How about defining a small (non public) `const fn` for this and using it for all these definitions to tidy up the duplication? I think there's a few where it isn't the same (where the vertical / horizontal have different left/right values), but most of these constants are the same.

```rust
const fn from_line_set(line_set: line::Set) -> BorderSet { ... }
```

joshka (approved):
Thanks. excellent work here. One small suggestion which is non-blocking, but if you don't mind doing it, I think will make this perfect.

ratatui-core/src/symbols/border.rs
theotchlx: Good idea! I added the function, and used it only on the dashed border sets. Should I open a new PR for them when these changes are approved, or do it now?

theotchlx: Thing is yeah, my function doesn't make use of line_set.vertical_left/vertical_right etc...

theotchlx: I added a test for the function

joshka (approved):

ratatui-core/src/symbols/border.rs
joshka: LGTM. Thanks! The vertical_left in the line set is for e.g. `┤` characters. vertical_left in the border is for vertical lines that are on the left side which can be distinctly different than vertical lines on the right side (e.g. for the way that the inner / outer borders are used for the blocky ones.

PR #1581, opened 2024-12-19
feat: add conversions for anstyle

joshka:
https://crates.io/crates/anstyle makes it possible to define colors in
an interoperable way. This makes it possible for applications to easily
load colors from a variety of formats.

This is gated by the anstyle feature flag which is disabled by default.

### DIFF
--- a/ratatui-core/Cargo.toml
+++ b/ratatui-core/Cargo.toml
@@ -24,7 +24,7 @@ rustdoc-args = ["--cfg", "docsrs"]
 [features]
 default = []
 
-## enables conversions to / from colors, modifiers, and styles in the ['answtyle'] crate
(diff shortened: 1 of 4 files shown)


ratatui-core/Cargo.toml
```diff
@@ -51,6 +55,7 @@ palette = { version = "0.7.6", optional = true }
 paste = "1.0.2"
 serde = { workspace = true, optional = true }
 strum.workspace = true
+thiserror = "2"
```
orhun, line 58: Should we put this behind a feature flag too?

joshka: Probably not - this one is fairly generally useful (and most dependency trees will have it in them somewhere)

PR #1584, opened 2024-12-22
feat: Add color to rat logo

Its-Just-Nans:
Why doesn't the logo have color **?**

image

Another question: Why does this logo isn't in `ratatui::widgets::RatatuiLogo`**?**


kdheepak:
I feel indifferent about adding a color to the rat. If we are going to add a color there's a number of considerations

1. It should work in both a light and dark theme on the website.
2. It should have good contrast to the background in both cases

joshka:
I didn't like going too grey on the rat. Take a look at the colors on the actual rat logo on the website for some ideas. It took a while to find the right grey there, but we stuck with white on the demo.

orhun:
I don't think adding colors to that demo is necessary.

> Another question: Why does this logo isn't in ratatui::widgets::RatatuiLogo?

This one is a demo/example rather than a widget.

We should probably rename that already existing widget to `RatatuiTextLogo` or `RatatuiText` and add the rat logo as `RatatuiLogo`.

joshka:
I think that the logo is the Ratatui word, while the Rat holding the terminal is the mascot. That's kinda how I think of it at least (and could be wrong here).

orhun:
Yeah, that also makes sense if we think that way. Then how about adding this mascot as `RatatuiMascot` widget (or even just `Rat` :D)

Though I'm not sure if that's super useful. It could be nice just as a gag.

ratatui-widgets/src/logo.rs
```diff
@@ -168,6 +168,189 @@ impl Size {
...
+        let term_color = self.colors.term;
+        let hat_color = self.colors.hat;
+        let border_color = self.colors.term_border;
+        for (y, (line1, line2)) in RATATUI_MASCOT.iter().tuples().enumerate() {
+            for (x, (ch1, ch2)) in line1.chars().zip(line2.chars()).enumerate() {
+                let x = area.left() + x as u16;
+                let y = area.top() + y as u16;
+                let cell = &mut buf[(x, y)];
+                match (ch1, ch2) {
+                    ('█', '█') => {
```
joshka, line 274: This logic was fine for 2 colors, but once you go beyond that it's worth instead mapping the characters a bit differently. I might be wrong on this as the (literal) edge cases here seem to be a bit hairy, but I think it culd be as simple as something like the following (untested code):
```rust

// given two cells which make up the top and bottom of the character, 
// Foreground color should be the non-space, non-terminal
let (fg, bg) = match (ch1, ch2) {
    (' ', ' ') => (Color::default(), Color::default()),
    (c, ' ') => (color_for(c), Color::default()),
    (' ', c) => (color_for(c), Color::default()),
    ('x', c) = (color_for(c), TERM_BG), // treat the terminal background cells as empty
    (c, 'x') = (TERM_BG, color_for(c)), // rather than having a character
    (c1, c2) => (color_for(c1), color_for(c2)),
}
// symbol should make the empty space or terminal bg as the empty part of the block
let symbol = match (ch1, ch2) {
    (' ', ' ') => ' ',
    ('x', 'x') => ' ', // treat the terminal background as an empty cell
    (c, ' ' | 'x') => '▀',
    (' ' | 'x', c) => '▄',
    (c, d) if c == d => '█',
    (c, d) => '▀',
}
```

joshka (commented):
Quite a few comments. Much of these are in the subjective bucket rather than necessary, which might generally be seen as a bit nitpicky, but I'm pushing for them here because this task is mostly in the same bucket. I hope you're ok with that.

Its-Just-Nans:
Thanks for the great review

Here is the updated code with the updated result

image

ratatui-widgets/src/logo.rs
joshka: This broke the cursor and border pixels - it's likely one of those difficult to get exactly right bits (edge cases :D).

Original
image

First cut:
image

Current:
image

The terminal border pixels and cursor are now broken.

ratatui-widgets/src/mascot.rs
```diff
@@ -0,0 +1,188 @@
...
+                    (c1, c2) => (self.color_for(c1), self.color_for(c2)),
+                };
+                // symbol should make the empty space or terminal bg as the empty part of the block
+                let symbol = match (ch1, ch2) {
+                    (TERM, TERM) => ' ',
+                    (_c, ' ' | TERM) => '▀',
+                    (' ' | TERM, _c) => '▄',
+                    (c, d) if c == d => '█',
+                    (_c, _d) => '▀',
+                };
```
joshka, line 151: Some simplification of the boolean logic (grouping terms etc.) I also put the TERM match before the space match as it reads a little easier. 
```suggestion
                // Combine two character lines from the picture into one cell
                // Make the foreground always the color of the non-empty part of the block
                let (fg, bg) = match (ch1, ch2) {
                    (TERM | ' ', ch2) => (self.color_for(ch2), self.color_for(ch1)),
                    (ch1, ch2) => (self.color_for(ch1), self.color_for(ch2)),
                };
                // symbol should make the empty space or terminal bg as the empty part of the block
                let symbol = match (ch1, ch2) {
                    (TERM | ' ', TERM | ' ') => ' ',
                    (TERM | ' ', _) => '▄',
                    (_, TERM | ' ') => '▀',
                    (c, d) if c == d => '█',
                    (_, _) => '▀',
                };
```

ratatui-widgets/src/mascot.rs
```diff
@@ -0,0 +1,188 @@
...
+    #[must_use]
+    pub const fn set_eye(self, rat_eye: MascotEye) -> Self {
+        Self {
+            eye_state: rat_eye,
+            ..self
+        }
+    }
+    const fn color_for(&self, c: char) -> Option<Color> {
+        match c {
+            '█' => Some(self.rat_color),
```
joshka, line 109: A few constants for these characters would be a good idea here.

ratatui-widgets/src/mascot.rs
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    #[test]
+    fn render_mascot() {
+        let mascot = RatatuiMascot::new();
+        let mut buf = Buffer::empty(Rect::new(0, 0, 32, 32));
+        mascot.render(buf.area, &mut buf);
+        assert_eq!(buf.area.as_size(), (32, 32).into());
+    }
+}
```
joshka, line 188: I don't recall if we added a way to assert lines without style info in tests yet, but if there's something like that elsewhere in other tests, it would make a good test to check the output rendering is correct here.

joshka (commented):
Great work getting this working. I have some small simplifications / fixes noted.

Also, I don't recall when / why the rat moved to the right side. It wasn't part of this PR, but it looks a bit odd. Perhaps flip it back to the left side?

ratatui-widgets/src/logo.rs
Its-Just-Nans: image

Is it correct now ?

ratatui-widgets/src/mascot.rs
joshka: (btw. Make sure to include the space as it will make the match statement neater)

ratatui-widgets/src/mascot.rs
Its-Just-Nans: I added a check for the eye

ratatui-widgets/src/mascot.rs
```diff
@@ -0,0 +1,196 @@
...
+    fn render(self, area: Rect, buf: &mut Buffer) {
+        for (y, (line1, line2)) in RATATUI_MASCOT.lines().tuples().enumerate() {
+            for (x, (ch1, ch2)) in line1.chars().zip(line2.chars()).enumerate() {
+                let x = area.left() + x as u16;
+                let y = area.top() + y as u16;
+                let cell = &mut buf[(x, y)];
+                // given two cells which make up the top and bottom of the character,
+                // Foreground color should be the non-space, non-terminal
+                let (fg, bg) = match (ch1, ch2) {
+                    (' ', ' ') => (None, None),
```
joshka, line 142: nit: Adding a const for EMPTY makes this a bit easier to read.

ratatui-widgets/src/mascot.rs
```diff
@@ -0,0 +1,196 @@
...
+
+    const fn color_for(&self, c: char) -> Option<Color> {
+        match c {
+            RAT => Some(self.rat_color),
+            HAT => Some(self.hat_color),
+            EYE => Some(match self.eye_state {
+                MascotEye::Default => self.rat_eye_color,
+                MascotEye::Red => self.rat_eye_blink,
+            }),
+            TERM => Some(Color::Reset), // use the terminal color
```
joshka, line 120: I missed seeing this before. This is definitely cool (and might make a good option for the widget), but I think the default of the darker known color would be generally better here than making this transparent to the underlying terminal color. Mainly because it makes the demo render consistently on any terminal.

ratatui-widgets/src/mascot.rs
joshka: What I meant was
```rust
assert_eq!(buf, Buffer::with_lines([
    ".....",
    ".....",
]);
```

It's not a big deal to not have this though. Un-resolving so you see this comment. No action necessary.

ratatui-widgets/src/mascot.rs
```diff
@@ -0,0 +1,196 @@
...
+/// State for the mascot's eye
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub enum MascotEye {
+    /// The default eye color
+    #[default]
+    Default,
+
+    /// The red eye color
+    Red,
+}
```
joshka, line 58: nit: this seems like a bit of an ambiguous name (but I don't know a better one). Default is one of those names that tends not to be used as a variant name. Would Normal (or open) and Red (or blink) make sense as variant names?

This is very much in the small irrelevant change bucket, so if you're happy with these I'm fine with it and will leave it until some obvious better name hits my subconscious.

joshka (approved):
A couple of last bits.

Some context on the generation of this demo btw. There was a whole tonne of back and forth between @kdheepak and I one random evening tweaking a whole tonne of things (colors, pixels, etc.) a while back to celebrate https://github.com/ratatui/ratatui/pull/500 and then a collaboration of a bunch of different people helped make it look even better. 

Then we added the destroy function function for [commit 1000](https://github.com/ratatui/ratatui/pull/809) in the repo.

We just missed commit / PR 1500 with this change which is a bit of a bummer, but c'est la vie.

Slapping an approval on this. If the other maintainers want to chime in on any last minute changes, feel free, but I think this is mostly good at this point (with some small nits).

ratatui-widgets/src/mascot.rs
kdheepak: Probably should be `MascotEyeColor`? 

And maybe `Transparent` instead of `Default` as the default variant?

joshka: I'd avoid transparency, because it makes the rendering dependent on what is in the buffer before rendering. This is ok as an option, but not as a default. Similar comment on https://github.com/ratatui/ratatui/pull/1584#discussion_r1896714671

ratatui-widgets/src/lib.rs
```diff
@@ -86,6 +86,7 @@ pub mod clear;
 pub mod gauge;
 pub mod list;
 pub mod logo;
+pub mod mascot;
```
orhun, line 89: We should also mention this in `Available Widgets` above.

ratatui-widgets/src/mascot.rs
orhun: `MascotEyeColor` sounds better to me. Alternatively, we can simply go without an enum and let the user set the `Color` via `set_eye_color`. e.g. `RatatuiMascot::new().set_eye_color(Color::Red);`

orhun (approved):
LGTM!

ratatui-widgets/src/mascot.rs
joshka: I think keeping a bit of control over the logo and how it can be presented is a good idea rather than making this open. That includes specifics of colors etc.

orhun: Sure, that makes sense.

ratatui-widgets/src/mascot.rs
Its-Just-Nans: i've added a test to compare lines ;)

Its-Just-Nans:
Hello

I've updated the code with corrections :)

joshka:
LGTM - I pushed a last change to make the terminal border color 237 because it's (close enough to) the dark mode color for macOS borders.

image

joshka:
Thanks for sticking with this to get it over the line. This is one of those changes that definitely was always going to have some back and forth due to the subjective nature, and I appreciate you for the work you spent getting it over the line.

ratatui-widgets/src/mascot.rs
```diff
@@ -86,7 +86,7 @@ impl Default for RatatuiMascot {
             rat_eye_color: Color::Indexed(236),     // dark_charcoal #303030
             rat_eye_blink: Color::Indexed(196),     // red #ff0000
             term_color: Color::Indexed(232),        // vampire_black #080808
-            term_border_color: Color::Indexed(244), // gray  #808080
+            term_border_color: Color::Indexed(237), // gray  #808080
```
Its-Just-Nans, line 89: @joshka Note that the comment should, in theory, be 

```rs
term_border_color: Color::Indexed(237), // #3a3a3a black_olive
```

joshka: Ha - oops D:

PR #1592, opened 2024-12-24
chore: use cargo xtask for bacon clippy command

joshka:


orhun (approved):
LGTM, we can merge this without the unrelated formatting changes though

joshka:
> LGTM, we can merge this without the unrelated formatting changes though

I just let the formatter plugin do its thing here and don't try to fight it / configure it. Source code definitely matters, config I don't think does (at least this config). I'm not aware of a cross-IDE/editor config / plugin for choosing wrapping options for TOML. So I think the best approach is a just to accept this as-is - does that work for you?

I suspect this change was caused by some tweak in the wrapping threshold of the plugin, or something similar.

orhun:
Yeah that sounds fair. Maybe we should look into [`taplo`](https://github.com/tamasfe/taplo) for more uniform formatting across TOML files in the repo.

joshka:
https://github.com/ratatui/ratatui/pull/1602

PR #1593, opened 2024-12-26
perf: add backend diff comparison benchmark for crossterm

joshka:


ratatui/benches/main/crossterm.rs
```diff
@@ -0,0 +1,72 @@
...
+        for x in 0..area.width {
+            let symbol = (rng.gen::<u8>() % 26 + b'A') as char;
+            let fg = Color::Indexed(rng.gen_range(0..16));
+            let bg = Color::Indexed(rng.gen_range(0..16));
+            let style = Style::default().fg(fg).bg(bg);
+            if let Some(cell) = buffer.cell_mut((x, y)) {
+                cell.set_symbol(&symbol.to_string()).set_style(style);
+            }
+        }
+    }
```
benjajaja, line 70: Maybe set one random cell to a multi-width character, to include that logic in the benchmark.

PR #1595, opened 2024-12-26
feat(list)!: highlight symbol styling

airblast-dev:
Allow styling for `List`'s highlight symbol

Currently there is no way to style the highlight symbol of a `List`. 
This change makes it so anything that implements `Into<Line>` can be used as a highlight symbol.

BREAKING CHANGE: `List::highlight_symbol` can no longer be used in const context
BREAKING CHANGE: `List::highlight_symbol` accepted `&str`. Conversion methods that rely on type inference will need to be rewritten as the compiler cannot infer the type.

closes: https://github.com/ratatui/ratatui/issues/1443

### DIFF
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -13,7 +13,7 @@ This is a quick summary of the sections below:
 - [Unreleased](#unreleased)
   - The `From` impls for backend types are now replaced with more specific traits
   - `FrameExt` trait for `unstable-widget-ref` feature
-  - `List::highlight_symbol` cannot be used in const context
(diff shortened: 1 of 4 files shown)


airblast-dev:
I am aware that https://github.com/ratatui/ratatui/pull/1467 exists and there a few small differences to that PR.

- No clones being performed
- Added a few tests to confirm the behavior
- Fixed a bug where the highlight style could overwrite the style from the highlight symbol

The PR was inactive for a while with very small changes needed. Feel free to close this if the other PR is preferred.

orhun (approved):

joshka:
Hey there, thanks for the PR (and @soumyasen1809 for the original fix)!

PR #1596, opened 2024-12-28
feat: implement Rect ops for moving / resizing

joshka:
Implement `Add`, `AddAssign`, `Sub`, and `SubAssign` on `Rect` for
`Offset` and `Size`. This makes it possible to move or resize rects using
arithmetic operations:

```rust
let rect = Rect::new(1, 2, 3, 4);
let moved = rect + Offset(1, 2);
let moved = rect - Offset(1, 2);
let moved = rect + Offset(-1, -2);
let resized = rect + Size(100, 200);
let resized = rect - Size(1, 2);
```

Additionally Rect, Size, Offset, and Position now all have MIN and MAX
consts.

DEPRECATES: The Rect::offset() method is no longer necessary and will be
removed in a future release (no earlier than 0.32.0)

Fixes: 1440
Alternative to: 1442


joshka (commented):
PIng @kardwen for a review. I was experimenting with working out what it would take to finish out #1440 / #1442, and I think this is likely a better idea than that approach. As mentioned in the issue, the one downside is that you can't do `rect + Size(10, -10)`, but this seems like an edge case (and one that can be easily decomposed to `rect + Size(10, 0) - Size(0, 10)`.

I think this is simpler in that it avoids naming issues (e.g. resize / sizedelta).

kardwen:
I think this approach is easier than the one using generics which may be a little bit confusing for beginners. To me it is not that intuitiv to add a size or an offset to a rectangle, but maybe that is just me. What do you think about keeping the names `move_by` and `resize_by` but changing the signature:

```rust
pub fn move_by(self, x: i32, y: i32) -> Rect
pub fn resize_by(self, x: i32, y: i32) -> Rect
```

I agree that `SizeDelta` and such is awkward but maybe these versions using `i32` as argument type could be introduced in addition to the `Add` trait implementation? I think the naming of the functions itself isn't too bad but this depends on the whole picture and if it is useful to adapt this naming scheme elsewhere (maybe a function `scale_by` is needed somewhere, but probably not for `Rect`).

Btw I don't mind if anything I have written gets merged or not, I think I learned a lot from it already.

orhun:
This is a good idea!

> I think this approach is easier than the one using generics which may be a little bit confusing for beginners. To me it is not that intuitiv to add a size or an offset to a rectangle,

Independently to the implementation details of this PR, I agree with this point. It might be a bit counter intuitive to use `+` / `-` and combinations of `Offset`/`Size`. 

I would prefer we do named methods (`move_by`/`resize_by`) and offer ops as an alternative to that.

joshka:
> What do you think about keeping the names `move_by` and `resize_by` but changing the signature:

I think that would be great. I was definitely struggling to find an upside of removing / deprecating the offset method but just replacing it with a method which takes the same params. It fell in that category of no real improvement just changed name. But changing the params like this makes a positive improvement (avoiding the need to introduce an Offset just to handle moving.

I think adding Rect + Offset makes sense when used in an expression. While Rect + Size makes a little less sense, it seems like something that once you see it once it would be easy to understand what's going on. So in that regard I think it's probably ok as an idea.

joshka:
Todo:
- [ ] Add move_by, move_to, resize_by, resize_to methods which accept Offset / Size and return Rects

joshka:
Self TODO - re-review where this go up to and finish it up

orhun:
ping @joshka - do you have availability to take a look at this soon? I was wondering if we should finalize this and include it in 0.30
Me or @j-g00da can also take this over I believe

joshka:
Yep - I'll take another look at this again this week.

joshka:
I rebased and updated this to only be about moving with Offset.
On reflection, I don't think that adding Size makes much sense, so restricted it to just the move ops (+Offset)
Also, I think it's reasonable to want to keep the offset method around rather than deprecating it. The rationale here is that a method is easily discoverable, while the ops for moving are less so

Previously this was marked as Fixes: #1440 and Alternative to: #1442, but it's kind of neither of those with this change. Followup to come

PR #1539, opened 2024-12-02
docs(widgets): add example for gauge

orhun:
related #1512

PR #1601, opened 2024-12-30
feat(linegauge): customized symbols

sectore:
## Overview

With this PR any symbol (`&str`) can be used to render `filled` and `unfilled` parts of `LineGauge` now.  Before that change, only [`symbols::line::Set`](https://docs.rs/ratatui/latest/ratatui/symbols/line/struct.Set.html) was accepted.

Note: New methods are introduced to define those symbols: `filled_symbol` and `unfilled_symbol`. The method [`line_set`](https://docs.rs/ratatui/latest/ratatui/widgets/struct.LineGauge.html#method.line_set) is still there, but marked as `deprecated`.

Examples and tests have been updated.

## Preview

line_gauge


sectore:
Waiting for #1612 to get CI fixed.

ratatui/examples/vhs/line_gauge.tape
```diff
@@ -5,10 +5,10 @@ Set Theme "Aardvark Blue"
 Set Width 1200
 Set Height 850
 Hide
-Type "cargo run --example=line_gauge --features=crossterm"
+Type "cargo run -p ratatui --example line_gauge --features=crossterm"
```
orhun, line 8: Good call, we need to update the VHS tapes as well, related to #1512

ratatui/examples/line_gauge.rs
orhun, line 1: We are in the process of shuffling things around a bit for better organization for the upcoming release which aims to modularize the codebase (#1388)  For that, we were planning to move these "widget-related" examples under the `ratatui-widgets` directory. (#1512)

For that, do you mind moving this line gauge example to there while also simplifying it a bit? See #1539 and #1544 for an example how that should look like.

orhun (approved):
This overall LGTM! Only had a small comment about the example :)

joshka (approved):
One thing to consider (and something which we've found in other places that accept plain `&str` values in the past is whether this should possibly accept a Span / Line instead. I suspect maybe not, but would like to give it some thought. When you end up with a symbol and style in a struct, then you really just have a `Span` deconstructed  into components there.

One problem here in making this accept `&str`, is that the length will no longer be as constrained as it was. This is likely ok (and the same thing could be said for the value coming from the `line::Set`, so it's not worse. But it's worth thinking about and making some comment on whether this is a problem.

The other thing I was thinking about with respect to this is whether there's a need to also modify the normal gauge in a simile way. I suspect probably not, but it's worth thinking about while were looking at this.

Overall, happy with the idea, looks good, would merge once discussion above is resolved.

ratatui-widgets/src/gauge.rs
```diff
@@ -311,8 +328,27 @@ impl<'a> LineGauge<'a> {
     /// [`NORMAL`](symbols::line::NORMAL), [`DOUBLE`](symbols::line::DOUBLE) and
     /// [`THICK`](symbols::line::THICK).
     #[must_use = "method moves the value of self and returns the modified value"]
+    #[deprecated(
+        since = "0.29.1",
+        note = "You should use `LineGauge::filled_symbol` and `LineGauge::unfilled_symbol` instead."
+    )]
```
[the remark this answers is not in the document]
sectore: Updated to `0.30.0`. Yep, deprecating this function can be reverted in the future if really needed.

sectore:
@orhun  @joshka Thx for your review.

> One thing to consider (and something which we've found in other places that accept plain `&str` values in the past is whether this should possibly accept a Span / Line instead. I suspect maybe not, but would like to give it some thought. When you end up with a symbol and style in a struct, then you really just have a `Span` deconstructed into components there.
> 
> One problem here in making this accept `&str`, is that the length will no longer be as constrained as it was. This is likely ok (and the same thing could be said for the value coming from the `line::Set`, so it's not worse. But it's worth thinking about and making some comment on whether this is a problem.

I don't see any issues by using a `&str` here. With that a developer is free to use any character without any restriction. Creativity can be endless here. IMO the only thing can happen is a broken layout by using an invalid `&str`. But that's something a developer can check and avoid easily. 

> 
> The other thing I was thinking about with respect to this is whether there's a need to also modify the normal gauge in a simile way. I suspect probably not, but it's worth thinking about while were looking at this.

My first attempt of this PR was to add changes to the normal gauge, but I skipped it because of it's implementation. I don't remember all the details and have not pushed these changes into this PR and . But the issue here seemed to be the option to have a multi-line rendered normal gauge or something. If this multi-line thing is not a really needed feature for a gauge, the normal gauge could be removed in favour to use the line gauge only. 

> 
> Overall, happy with the idea, looks good, would merge once discussion above is resolved.

My suggestion is to merge this PR.

joshka:
Makes sense. Thanks for answering the questions.

Issue #1116, opened 2024-05-19
Allow bypassing diff and writing all to screen


itsjunetime:
## Problem
I'm writing a TUI app with ratatui that makes pretty heavy usage of `ratatui_image` to display images. Because of this, the diff process (specifically, the `unicode_width::width` function) takes a lot of the CPU time - much more than it would take to just render the images.

## Solution
I would like to add a field to `Terminal` (or perhaps `Frame`, but I think that would be harder to do while maintaining a simple API) called `bypass_diff` (or something similar) that will bypass the diff call/process when flushing the terminal to the screen and instead just render everything.

## Alternatives
An alternative to this would be simply improving the performance of the diff function. I expect it's possible somehow (I know there are at least some small improvements to be made easily, such as removing the unnecessary access-by-index and double call to `current.symbol().width()`), but it would take a lot more work than just adding an option for developers to bypass the diffing process.

## Additional context
I would be happy to file a PR whenever, as I've already implemented this and verified that it does improve performance in my use case.

joshka:
I'd love to see some benchmarks on the diff function to measure the current performance as one easy-ish part of this.

It's actually possible to skip the entire rendering stack already if you need it:
```rust
terminal.set_cursor(x,y);
let writer = terminal.backend_mut();
write!("{}", your_image_data_as_ansi);
```

That's not super helpful to the ratatui-image widget approach I know, but I'm also not 100% sure what an ergonomic API for this might look like on terminal. It might be useful to think about an approach that doesn't break if we add other similar functionality. Some possible ideas (feel free to add more):
```rust
terminal.draw_direct(...);
terminal.draw_with_options(Options::new().direct(), ...);
terminal.draw(|frame| { frame.skip_diff(true); frame.render_widget(...); }
```

itsjunetime:
Alright, I was able to run some pseudo-benchmarks and get some numbers on the kind of performance hit I'm looking at. To do this, I basically just added the following code around my main render loop:

```rust
let mut end_update = false;
let start_time = std::time::Instant::now();
term.draw(|f| {
	tui.render(f, &main_area, &mut end_update);
	f.bypass_diff = true; // <- I changed this line based on the test
})?;
if end_update {
	execute!(stdout(), EndSynchronizedUpdate)?;
}
let end_time = std::time::Instant::now();
let total_time = end_time.duration_since(start_time);
if redraw_caused_by_crossterm_event {
	render_times.push(total_time.as_millis());
}
```

[...]

PR #1605, opened 2025-01-04
feat: cell diff options

benjajaja:
### feat: cell diff options

Problem: Escape sequences always cause a cell to count as "multiwidth",
even when it doesn't render wider than one cell, or not as wide as the
escape sequence would be computed as.

Solution: Convert `skip:bool` to enum. Add enum option `ForceWidth` to
force a cell width for diffing.

When using the option, this also fixes some bug where diffing is not
idempotent and causes a diff operation for `(symbol.len() - 1)` times.

There are three new specific test cases:

1. Rendering hyperlinks by squeezing the escape sequence into the first
   cell and forcing the width to the unicode width of the text part.
   This is much easier to implement for a Link widget, as it would only
   need to get the unicode-width once and not iterate over graphemes
   like Spans must do.
2. Rendering hyperlinks by squeezing the opening sequence into the first
   cell with the first grapheme and forcing the width to that of the
   first grapheme. Then rendering each grapheme as usual. Then squeezing
   the closing sequence into the last cell with the last grapheme and
   forcing the width to that of the last grapheme.
   This is harder to implement for a Link widget, as it would have to
   iterate over graphemes with their width like Spans do.
3. Kitty image sequence with utf-8 placeholders, similar to 2 but with
   known constant grapheme widths.

### Link widget that leverages this

https://github.com/benjajaja/tui-link

It would be cooler if we could just add something like `.link(url)` to `Span`s, because it would much simpler to insert some link and leverage all the Line/Text/Paragraph wrapping and whatnot. With a custom widget you need to take care of the `Area` where you'd want to render it, so it's not that clean. But we could iterate on this later, if even possible.

<details>
<summary>Click to expand old comment</summary>
Problem: setting escape sequences, e.g. image protocols, always causes a
cell to count as "multiwidth", even when it doesn't render wider than
one cell. For images specifically, when rendering a tiny image into just
one cell, the cell to the right will always be skipped - even if that
cell has nothing to do with the widget / buffer that the image is
operating on.
In general, this affects any escape sequence or otherwise invisible text
in a symbol that does not actually render wider than one cell: the next
cell will be erroneously skipped.

Solution: Convert `skip:bool` to enum. Add enum option `UnitWidth` to
force a cell width of 1 for diffing.

When using the option, this also fixes some bug where diffing is not
idempotent and causes a diff operation for `(symbol.len() - 1)` times.
</details>

### DIFF
--- a/ratatui-core/src/buffer/buffer.rs
+++ b/ratatui-core/src/buffer/buffer.rs
@@ -6,7 +6,7 @@ use core::{cmp, fmt};
 use unicode_segmentation::UnicodeSegmentation;
 use unicode_width::UnicodeWidthStr;
 
-use crate::buffer::Cell;
+use crate::buffer::{Cell, CellDiffOption};
 use crate::layout::{Position, Rect};
 use crate::style::Style;
 use crate::text::{Line, Span};
@@ -494,53 +494,67 @@ impl Buffer {
         let next_buffer = &other.content;
 
         let mut updates: Vec<(u16, u16, &Cell)> = vec![];
-        // Cells invalidated by drawing/replacing preceding multi-width characters:
-        let mut invalidated: usize = 0;
         // Cells from the current buffer to skip due to preceding multi-width characters taking
         // their place (the skipped cells should be blank anyway), or due to per-cell-skipping:
         let mut to_skip: usize = 0;
         for (i, (current, previous)) in next_buffer.iter().zip(previous_buffer.iter()).enumerate() {
-            if !current.skip && (current != previous || invalidated > 0) && to_skip == 0 {
-                let (x, y) = self.pos_of(i);
-                updates.push((x, y, &next_buffer[i]));
+            if to_skip > 0 {
+                to_skip -= 1;
+                continue;
+            }
+            match current.diff_option {
+                CellDiffOption::Skip => {}
+                CellDiffOption::ForcedWidth(width) => {
+                    if current != previous {
+                        let (x, y) = self.pos_of(i);
+                        updates.push((x, y, &next_buffer[i]));
+                    }
+                    to_skip = width.get().saturating_sub(1);
+                }
+                CellDiffOption::None => {
+                    if current != previous {
+                        let (x, y) = self.pos_of(i);
+                        updates.push((x, y, &next_buffer[i]));
 
-                // If the current cell is multi-width, ensure the trailing cells are explicitly
-                // cleared when they previously contained non-blank content. Some terminals do not
-                // reliably clear the trailing cell(s) when printing a wide grapheme, which can
-                // result in visual artifacts (e.g., leftover characters). Emitting an explicit
-                // update for the trailing cells avoids this.
-                let symbol = current.symbol();
-                let cell_width = symbol.width();
-                // Work around terminals that fail to clear the trailing cell of certain
-                // emoji presentation sequences (those containing VS16 / U+FE0F).
-                // Only emit explicit clears for such sequences to avoid bloating diffs
-                // for standard wide characters (e.g., CJK), which terminals handle well.
-                let contains_vs16 = symbol.chars().any(|c| c == '\u{FE0F}');
-                if cell_width > 1 && contains_vs16 {
-                    for k in 1..cell_width {
-                        let j = i + k;
-                        // Make sure that we are still inside the buffer.
-                        if j >= next_buffer.len() || j >= previous_buffer.len() {
-                            break;
+                        // If the current cell is multi-width, ensure the trailing cells are
+                        // explicitly cleared when they previously contained
+                        // non-blank content. Some terminals do not reliably
+                        // clear the trailing cell(s) when printing a wide grapheme, which can
+                        // result in visual artifacts (e.g., leftover characters). Emitting an
+                        // explicit update for the trailing cells avoids
+                        // this.
+                        let symbol = current.symbol();
+                        let cell_width = symbol.width();
+                        // Work around terminals that fail to clear the trailing cell of certain
+                        // emoji presentation sequences (those containing VS16 / U+FE0F).
+                        // Only emit explicit clears for such sequences to avoid bloating diffs
+                        // for standard wide characters (e.g., CJK), which terminals handle well.
+                        let contains_vs16 = symbol.chars().any(|c| c == '\u{FE0F}');
+                        if cell_width > 1 && contains_vs16 {
+                            for k in 1..cell_width {
+                                let j = i + k;
+                                // Make sure that we are still inside the buffer.
+                                if j >= next_buffer.len() || j >= previous_buffer.len() {
+                                    break;
+                                }
+                                let prev_trailing = &previous_buffer[j];
+                                let next_trailing = &next_buffer[j];
+                                if prev_trailing != next_trailing {
+                                    let (tx, ty) = self.pos_of(j);
+                                    // Push an explicit update for the trailing cell.
+                                    // This is expected to be a blank cell, but we use the actual
+                                    // content from the next buffer to handle cases where
+                                    // the user has explicitly set something else.
+                                    updates.push((tx, ty, next_trailing));
+                                }
+                            }
                         }
-                        let prev_trailing = &previous_buffer[j];
-                        let next_trailing = &next_buffer[j];
-                        if !next_trailing.skip && prev_trailing != next_trailing {
-                            let (tx, ty) = self.pos_of(j);
-                            // Push an explicit update for the trailing cell.
-                            // This is expected to be a blank cell, but we use the actual
-                            // content from the next buffer to handle cases where
-                            // the user has explicitly set something else.
-                            updates.push((tx, ty, next_trailing));
+                        if cell_width > 1 {
+                            to_skip = cell_width.saturating_sub(1);
                         }
                     }
                 }
             }
-
-            to_skip = current.symbol().width().saturating_sub(1);
-
-            let affected_width = cmp::max(current.symbol().width(), previous.symbol().width());
-            invalidated = cmp::max(affected_width, invalidated).saturating_sub(1);
         }
         updates
     }
@@ -679,8 +693,9 @@ impl fmt::Debug for Buffer {
 #[cfg(test)]
 mod tests {
     use alloc::format;
-    use alloc::string::ToString;
+    use alloc::string::{String, ToString};
     use core::iter;
+    use core::num::NonZero;
     use std::{dbg, println};
 
     use itertools::Itertools;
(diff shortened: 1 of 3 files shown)


benjajaja:
`UnitWidth` should be at least a partial solution to #1116.

benjajaja:
~~Some information and thoughts I didn't put in the code comments or commit message:~~

~~It seems that inspecting the cell width is mainly done with the intention of detecting graphemes that would be rendered as two cells wide, and then not rendering the next cell which would be "under" the previous.~~

~~If the grapheme width is greater than 2, then _I think_ merging the same buffer twice is no longer idempotent. But I haven't really gotten to the bottom of that issue, nor do I know how this behavior could affect possible usages of long strings in cell symbols - other than escape sequences.~~

ratatui-core/src/buffer/buffer.rs
```diff
@@ -1109,12 +1110,65 @@ mod tests {
...
+        sequences
+            .cell_mut((3, 0))
+            .unwrap()
+            .set_symbol("\u{10EEEE}\x1b[u")
+            .set_diff_option(CellDiffOption::UnitWidth);
+
+        let mut buffer = Buffer::filled(Rect::new(0, 0, 20, 1), Cell::new("x"));
+        buffer.merge(&sequences);
+
+        let diff = buffer.diff(&sequences);
```
benjajaja, line 1162: This test fails, if `Cell::width` returns the symbol width, instead of 1 when `diff_option` is `UnitWidth`.

benjajaja:
I'll leave this as draft until someone weights in on also adding a `Force` variant or not. Also I personally don't need a `Force` option for ratatui-image, just the `UnitWidth` for control sequences.

ratatui-core/src/buffer/cell.rs
```diff
@@ -128,12 +147,9 @@ impl Cell {
...
 
-    /// Sets the cell to be skipped when copying (diffing) the buffer to the screen.
-    ///
-    /// This is helpful when it is necessary to prevent the buffer from overwriting a cell that is
-    /// covered by an image from some terminal graphics protocol (Sixel / iTerm / Kitty ...).
-    pub fn set_skip(&mut self, skip: bool) -> &mut Self {
-        self.skip = skip;
+    /// Sets cell diffing options.
+    pub fn set_diff_option(&mut self, diff_option: CellDiffOption) -> &mut Self {
+        self.diff_option = diff_option;
```
joshka, line 152: One thing we tend to do to make it easier for upgrading between versions is to make changes in a backwards compatible manner. Could this be done by deprecating `set_skip()` and having it call `set_diff_option` with the right value?

There's only a small amount of usages of the existing method, so this might not be a huge deal. https://github.com/search?q=ratatui+set_skip+AND+%28NOT+org%3Aratatui%29+AND+%28NOT+is%3Afork%29&type=code

benjajaja:
@joshka hey, I picked this up again, and personally I'm quite happy with how it looks like now.

orhun (approved):
Neat, I like this. Nothing to remark 👌🏼

ratatui-core/src/buffer/buffer.rs
```diff
@@ -486,53 +486,67 @@ impl Buffer {
...
+                continue;
+            }
+            match current.diff_option {
+                CellDiffOption::Skip => {}
+                CellDiffOption::ForcedWidth(width) => {
+                    if current != previous {
+                        let (x, y) = self.pos_of(i);
+                        updates.push((x, y, &next_buffer[i]));
+                    }
+                    to_skip = width.saturating_sub(1);
```
benjajaja: It just came to me, what happens on `ForcedWidth(0)`? The answer is "nothing much" because of this line, it behaves exactly like `ForcedWidth(1)`. But should something else happen?

Instead of having `Skip` we could have made `ForcedWidth(0)` like `Skip`, but I think it's better to have an explicit intention: we really want to "skip" because something else, anywhere, will render "over" this cell, and `ForcedWidth(_)` rather controls "how many following cells should be skipped".

But from a library perspective, should `ForcedWidth(0)` then be an error or something, until it does something useful (maybe never)?

j-g00da: We can probably just use `NonZeroUsize` there, I agree that this should be more explicit as 0 working like 1 can be a bit confusing.

ratatui-core/src/buffer/cell.rs
```diff
@@ -180,12 +199,23 @@ impl Cell {
...
-        self.skip = skip;
+        self.diff_option = if skip {
+            CellDiffOption::Skip
+        } else {
+            CellDiffOption::None
+        };
+        self
+    }
+
+    /// Sets cell diffing options.
```
j-g00da: This doc needs expanding, currently it's hard to guess what this does without reading `CellDiffOption` docs. Please give it a little summary and point to `CellDiffOption` for more information.

ratatui-core/src/buffer/cell.rs
```diff
@@ -180,12 +199,23 @@ impl Cell {
...
 
-    /// Sets the cell to be skipped when copying (diffing) the buffer to the screen.
-    ///
-    /// This is helpful when it is necessary to prevent the buffer from overwriting a cell that is
-    /// covered by an image from some terminal graphics protocol (Sixel / iTerm / Kitty ...).
+    #[deprecated(
+        since = "0.30.0",
+        note = "use `set_diff_option(CellDiffOption::Skip)` instead"
+    )]
+    /// Set cell diffing option to `CellDiffOption::Skip`.
```
j-g00da:
```suggestion
    /// Set cell diffing option to [`CellDiffOption::Skip`].
```

ratatui-core/src/buffer/cell.rs
```diff
@@ -3,6 +3,25 @@ use compact_str::CompactString;
...
+    ///
+    /// This is helpful when it is necessary to prevent the buffer from overwriting a cell that is
+    /// covered by something from an escape sequence, such as graphics or links.
+    Skip,
+    /// Force a width regardless of the symbol text width.
+    ///
+    /// Escape sequences will have some computed width that does match what is written to the
+    /// screen.
+    ForcedWidth(usize),
+}
```
j-g00da, line 25: nit: not 100% sure about the naming here

j-g00da (commented):
Nice! Mostly nits :)

ratatui-core/src/buffer/buffer.rs
j-g00da: (to be precise, I meant `ForcedWidth(usize)` -> `ForcedWidth(NonZeroUsize)`)

benjajaja: Thanks for the review. `NonZeroUsize` is good!

j-g00da (approved):

ratatui-core/src/buffer/cell.rs
```diff
@@ -1,8 +1,29 @@
...
+    ///
+    /// This is helpful when it is necessary to prevent the buffer from overwriting a cell that is
+    /// covered by something from an escape sequence, such as graphics or links.
+    Skip,
+    /// Force a width regardless of the symbol text width.
+    ///
+    /// Escape sequences will have some computed width that does match what is written to the
+    /// screen.
+    ForcedWidth(NonZeroUsize),
+}
```
joshka, line 25: I agree with j-g00da here. It's possible that we're trying to combine two things in the one enum here:
- the ability to skip diffing against the prior buffer's cell in the same position to decide whether to render the current buffer's symbol
- the ability to skip rendering future cells because we have a width that overlaps horizontal cells
These seem a bit orthogonal.

A quick example of this might be to ask what should be the right value for two images that we render that overlap (e.g. 0,0 - 2x1 image, 1,0 - 2x1 image).

Also, it might be reasonably possible that we want to render zero width items (OSC8 wrapped around a zerowidth joiner or silly things like that). We need to clamp the rendering to 1 though, but I think that's a render level thing, not a cell level possibly (e.g. say I want to be able to put ANSI that is non-rendering in the cell without overwriting anything that's already in the cell from the previous draw call - for whatever reason - not sure of an exact example),

joshka (approved):
Solves the specific problem at hand reasonably well.
I'm going to think about this over the next day or so and see if there's some better ideas aligned a bit with what I mentioned in the comment.

General ok to go ahead with this for now though if this falls below the line of what I can put more time on.

ratatui-core/src/buffer/cell.rs
benjajaja: > I agree with j-g00da here. It's possible that we're trying to combine two things in the one enum here:
> 
>     * the ability to skip diffing against the prior buffer's cell in the same position to decide whether to render the current buffer's symbol
> 
>     * the ability to skip rendering future cells because we have a width that overlaps horizontal cells
>       These seem a bit orthogonal.

Consider removing `Skip`, and having only `None | ForcedSize(Size)` (notice the change to `Size` from `NonZeroUsize`-width) - (or an Option<bool>, that's not important).

We *could* theoretically "create" `Skip`-like private flags at diff time on the *affected cells*. BUT then we would have cases where we would need to clear the `Skip`s that have  been caused by an older `ForcedSize` cell.

E.g. buffer-merge some cell that has a hypothetical `ForcedSize(2x2)` at `0,1`:

|  | 0 | 1 | 2
|-|-|-|-
|0|none|ForcedSize(2x2)|S|
|1|none|S|S|

and on some other merge we write a regular cell with no diff options into `1,0`, we'd have to undo the `Skip` flags on `1,1`, `1,1`, and `1,2`. This could work, but then we would also have to deal with a `ForcedSize` becoming `Skip` from another `ForcedSize` to the top/left of it, e.g.: At `0,0`, when we encounter the `ForcedSize` at `0,1`, we could try to clear all the `Skip` that it had caused. But since `ForcedSize`s could "overlap", this is a bit ambiguous, I think it is possible to "clear if *this* `ForcedSize` that we're undoing is the last one that would cause this `Skip`". 

In other words, a huge rabbit hole we should avoid getting into. Having `Skip` and `ForcedWidth(NonZeroUsize)` are two different things, but they both happen at diffing, this `CellDiffOption`. I'm not saying this is perfect, I think that many users won't know or care that this is all about "diffing", that many would think about this as "rendering" (as you maybe hint below) and may be confused about the "diff" naming. But then again, users that need to concern themselves with this should better be ready to understand what's going on between writing to some intermediate buffer, and what ends up on the terminal screen.

> A quick example of this might be to ask what should be the right value for two images that we render that overlap (e.g. 0,0 - 2x1 image, 1,0 - 2x1 image).

Isn't this the exact same case as having two double-width characters following each other too closely after a buffer-merge? Except in those cases we have the luxury of iterating through the cells of a row and can deal with it (maybe?), but when images occupy multiple rows, I don't think we can do anything about vertical overlaps.

> Also, it might be reasonably possible that we want to render zero width items (OSC8 wrapped around a zerowidth joiner or silly things like that). We need to clamp the rendering to 1 though, but I think that's a render level thing, not a cell level possibly (e.g. say I want to be able to put ANSI that is non-rendering in the cell without overwriting anything that's already in the cell from the previous draw call - for whatever reason - not sure of an exact example),

You could just append the symbol and set `ForcedWidth(1)`. Or prepend (e.g. for OSC8). But specifically for OSC8, I don't think ratatui can simply put the starting part before the text, and the ending part after the end. Because ratatui does all the diffing stuff, and may not write out `<start-seq>link text<end-seq>`, when some characters didn't change (by chance or not). And ultimately, the terminal needs to see the whole thing written out, it doesn't "put it together" by what ends up on screen. This means if we want to write out an OSC8 link, we need to [put the whole `<start-seq>link text<end-seq>` into the first cell (with `ForcedWidth(<text unicode width>)`](https://github.com/benjajaja/tui-link/blob/master/src/lib.rs#L47).

orhun:
Anything blocking here?

ratatui-core/src/buffer/buffer.rs
```diff
@@ -494,53 +494,67 @@ impl Buffer {
...
+
+                        // If the current cell is multi-width, ensure the trailing cells are
+                        // explicitly cleared when they previously contained
+                        // non-blank content. Some terminals do not reliably
+                        // clear the trailing cell(s) when printing a wide grapheme, which can
+                        // result in visual artifacts (e.g., leftover characters). Emitting an
+                        // explicit update for the trailing cells avoids
+                        // this.
+                        let symbol = current.symbol();
+                        let cell_width = symbol.width();
```
junkdog, line 527: since ascii covers most cells, a cheap optimization is to first check if symbol.len() is 1 - else get width from `width()`. this is not a big deal on desktop, but on embedded it's a pretty expensive op.

orhun: let's create a follow up PR to this!

orhun:
I'll merge this for now. We could continue with the rest of the discussion/improvements on separate issues/PRs I believe :)

orhun:
Seems like removing Cell::skip (pub field) was a breaking change. We should do sth about it

thanks for the headsup @junkdog

benjajaja:
@orhun what needs to be done?

orhun:
@junkdog is currently testing this change in tachyonfx. It might be nice to get his input

other than that, we need to update `BREAKING-CHANGES.md` about this

junkdog:
there's also the issue of `Cell` size expanding from 40 to 56 bytes (24->32 on 32bit arch) due to the usize in`CellDiffOptions::ForcedWidth(NonZeroUsize)`. the extra wasted space comes from usize's 8-byte alignment (which is why it doesn't just increase to 48 bytes).

benjajaja:
@junkdog should we use `u16`? Or do something entirely different here?

junkdog:
i don't have the numbers in front of me, but i think there's room for a u8 w/o growing the cell, considering both 32 and 64 bit. all tests passed with u8 width (after some minor syntactic updates). not sure about u16 - can width ever be larger than 2?

benjajaja:
The width was meant to allow for USC8 sequences, so URL+text into one cell.

But I wonder if we could achieve that with ForceWidth(1) and Skip for the following cells. I will try that out and report back. If that works out, then the enum could just be `Skip|One|Two`.

junkdog:
you can take a look at https://github.com/junkdog/ratatui/tree/cell-str-width - where i fixed it along with some width() usage; it uses u8 for width, which i think works, but tbh i didn't consider the use-case

benjajaja:
@junkdog that branch works fine with [tui-link](https://github.com/benjajaja/tui-link), if the link *text* is under 256 characters (the URL size doesn't matter). The real use case I know for this is "wrapped links", something like this:

```
╔═ box ══════╗
║            ║
║http://examp║
║le.com      ║
║            ║
╚════════════╝
```
Where *both* lines would have the OSC8 sequences, with different text but same URL. For context, the "link detection" of terminals would probably detect `http://examp` for the first line which is undesireable.

Bottom line: a 256 limit is basically a limit of the screen width. I don't know what the general screen width target of ratatui is - it surely isn't 80, ratatui is using u16 all over for x,y coordinates, so u8 seems off here. But the size concerns might overrule this concern in this particular case.

<details>
<summary>Technical explanation why `ForceWidth(1)` doesn't work for OSC8</summary>
Ideally, we could split the OSC8 into the start, middle, and end parts, and simply `ForceWidth(1)` start and end, as the middle would be all normal cells.

The problem is in the diffing. For the OSC8 to work, the whole sequence must be flushed out at once, which would happen in the majority of cases, except when the buffer diffing detects that it does not need the "re-render" some of the middle parts. E.g.: Regular text `Slick me?` gets overwritten by OSC8 `Click me!`. The inner `lick me` would not get "sent out", only the beginning `{OSC8,URL,start}C` and ending `!{OSC8-end}`. In practice, terminals then render the first and last character as "clickable".
</details>

orhun:
Should we create a PR from that branch and test it out there?

junkdog:
@orhun  i can put up a PR for it, but it addresses `symbol.width()` calculations first and foremost. it's not related per se to this issue. there are a couple of call-sites that need to be updated before it's PR-worthy. prob sometime today.

thanks for the in-depth explanation, @benjajaja . appreciate it! this isn't really related to the issue, but if i'm understanding this correctly - this means that an url will always apply the initial cell's style along the skipped cells; how does this work if a widget writes to part of the area covered by the url? i'm guessing such changes will be ignored(/skipped) or is it possible to e.g. render the url with a gradient fg color or overlay a block?  the reason i'm asking is because i should maybe mention it somehwhere in tachyonfx's docs.

i had closer look at cell with forcedwidth for usize (current), u16 and u8:

cell-memory-layout

on 64 bit arch, u8 or u16 for `FixedWidth` doesn't matter. on 32 bit archs (embedded and wasm primarily), u16 grows the Cell by 4 bytes, which accumulates to quite a lot on embedded. but here it's maybe better to control it via an "embedded" feature (I guess it relates to my post yesterday, #2379).

so yeah, my suggestion is to go with u16 and then we worry about how to solve it for embedded elsewhere.

I could update my cell width branch with u16 and send a PR, as it's just a one-liner? what do you guys thinks?

RoloEdits:
@junkdog Apologies for the tangent, but was this layout visualization from some software or was this done from scratch?

junkdog:
@RoloEdits it's ad hoc unfortunately, i extracted the field offsets with a macro and then fed the data to my [/diagram](https://github.com/junkdog/agent-skills/tree/main/diagram) agent skill - it generated the diagram using gnuplot. it would be neat to have a cargo command to recursively dump the memory layout of a type; my cursory search didn't find anything.

ratatui-core/src/buffer/cell.rs
```diff
@@ -30,8 +51,8 @@ pub struct Cell {
     /// The modifier of the cell.
     pub modifier: Modifier,
 
-    /// Whether the cell should be skipped when copying (diffing) the buffer to the screen.
-    pub skip: bool,
+    /// Special option applied when copying (diffing) the buffer to the screen (or another buffer).
+    pub diff_option: CellDiffOption,
 }
```
orhun, line 56: We talked about this breaking change w/ @junkdog and I think the option that makes most sense it to bring back `skip: bool` (with restored behavior) and add a deprecation notice with the intention of removing it in the next minor release. Having both `skip` and `diff_option` might complicate the diff logic a bit but I think it's probably a good enough tradeoff.

benjajaja: Are there any known crates that uses `skip` apart from ratatui-image?

junkdog: @benjajaja i asked claude to search github yesterday:

image

looks like it missed ratataui-image though.

benjajaja: Thank you @junkdog, of course there are many more image crates :laughing:

december1981:
I thought this might help, but I've noticed - and I believe I've traced it correctly to this work - that the Full-width plus sign (＋ (U+FF0B))  is not being cleared properly, I'm seeing some kind of trailing artifacts with it when it's no longer being rendered from previous.

At least I can see this issue with GNOME terminal and Alacritty (although not ghostty)

PR #1607, opened 2025-01-05
perf: termion version of colors_rgb

joshka:
Just a quick and dirty version to check perf differences

Left is crossterm, right is termion...

image


orhun:
Here are my findings:

In release mode:

- crossterm: ~50fps
- termwiz: ~300fps
- termion: ~666fps

which is absolutely crazy!

aschey:
I thought the drastic performance difference mentioned above was a bit curious so I took a look at it. The crossterm version is polling for inputs on the same thread whereas the termion version is running it on a separate thread. I changed the input handler to match what the termion version is doing and the performance now seems roughly equal (tested on Linux with a few different terminals). Here's the [change I made](https://github.com/aschey/ratatui/commit/bc121de9d83ec82cdcbdf431c2160c9e47428b8c).

PR #1612, opened 2025-01-10
fix: rust 1.84 clippy lints

joshka:
The canvas map constants are now statics instead.
Fixes https://rust-lang.github.io/rust-clippy/master/index.html\#large_const_arrays


orhun (approved):
LGTM!

PR #1622, opened 2025-01-18
perf(widgets): avoid unnecessary work when rendering paragraphs

alexpasmantier:
Hi,

Thanks for the great work on ratatui 🙏🏻 

### Context
_This PR introduces a slight change that I believe could end up significantly improving rendering times for paragraphs that do not depend on certain kinds of `LineComposer`s._

Currently all `LineComposer`s are considered to be state machines which means rendering a paragraph with a given Y offset requires computing the entire state up to Y before being able to render from Y onwards.

While this makes sense for Composers such as the `WordWrapper` (where one needs to consider all previous lines to determine where a given line will end up), it means it also penalizes Composers which can render a given line "statelessely" (such as the `LineTruncator`) which actually end up doing a lot of unnecessary work (and on the critical rendering path) when the offset gets high.

I came across this while experimenting with previews in [television](https://github.com/alexpasmantier/television) and noticed increased rendering lag as the preview offset got higher for large files.

#### Example
Below is an example of sampling the application and scrolling down a pretty large file in the preview pane while everything else remains static. Note how rendering the paragraph takes longer and longer (in proportion to the rest which I'm assuming shouldn't change much).

https://github.com/user-attachments/assets/1c6dc428-5351-4659-97c1-45fe7fb2ffb4

### Description of the PR

I tried to keep the PR as minimal as possible. The `LineComposer` trait was split into:
- `StatefulLineComposer`: e.g. `WordWrapper`
- `LineComposer`: e.g. `LineTruncator`

When rendering a paragraph, the behavior is now as follows:
- if the line composer is stateful, use the existing code to compute lines up to offset `y` and render from there on up to the area's height
- if the line composer is stateless, just skip ahead until `y` and render from there on up to the area's height

### DIFF
--- a/ratatui-widgets/src/paragraph.rs
+++ b/ratatui-widgets/src/paragraph.rs
@@ -446,10 +446,20 @@ impl Paragraph<'_> {
         });
 
         if let Some(Wrap { trim }) = self.wrap {
-            let line_composer = WordWrapper::new(styled, text_area.width, trim);
+            let mut line_composer = WordWrapper::new(styled, text_area.width, trim);
+            // because WordWrapper is a stateful line composer, we need to compute the lines
+            // iteratively until we reach the desired scroll offset.
+            for _ in 0..self.scroll.y {
+                if line_composer.next_line().is_none() {
+                    return;
+                }
+            }
             self.render_text(line_composer, text_area, buf);
         } else {
-            let mut line_composer = LineTruncator::new(styled, text_area.width);
+            // because LineTruncator is not a stateful line composer, we can skip directly to the
+            // relevant line and start rendering from there which avoids unnecessary work.
+            let mut line_composer =
+                LineTruncator::new(styled.skip(self.scroll.y.into()), text_area.width);
             line_composer.set_horizontal_offset(self.scroll.x);
             self.render_text(line_composer, text_area, buf);
         }
@@ -457,6 +467,7 @@ impl Paragraph<'_> {
 }
 
 impl<'a> Paragraph<'a> {
+    #[allow(clippy::unused_self)]
     fn render_text<C: LineComposer<'a>>(&self, mut composer: C, area: Rect, buf: &mut Buffer) {
         let mut y = 0;
         while let Some(WrappedLine {
@@ -465,24 +476,22 @@ impl<'a> Paragraph<'a> {
             alignment: current_line_alignment,
         }) = composer.next_line()
         {
-            if y >= self.scroll.y {
-                let mut x = get_line_offset(current_line_width, area.width, current_line_alignment);
-                for StyledGrapheme { symbol, style } in current_line {
-                    let width = symbol.width();
-                    if width == 0 {
-                        continue;
-                    }
-                    // If the symbol is empty, the last char which rendered last time will
-                    // leave on the line. It's a quick fix.
-                    let symbol = if symbol.is_empty() { " " } else { symbol };
-                    buf[(area.left() + x, area.top() + y - self.scroll.y)]
-                        .set_symbol(symbol)
-                        .set_style(*style);
-                    x += width as u16;
+            let mut x = get_line_offset(current_line_width, area.width, current_line_alignment);
+            for StyledGrapheme { symbol, style } in current_line {
+                let width = symbol.width();
+                if width == 0 {
+                    continue;
                 }
+                // If the symbol is empty, the last char which rendered last time will
+                // leave on the line. It's a quick fix.
+                let symbol = if symbol.is_empty() { " " } else { symbol };
+                buf[(area.left() + x, area.top() + y)]
+                    .set_symbol(symbol)
+                    .set_style(*style);
+                x += u16::try_from(width).unwrap_or(u16::MAX);
             }
             y += 1;
-            if y >= area.height + self.scroll.y {
+            if y >= area.height {
                 break;
             }
         }
--- a/ratatui-widgets/src/paragraph.rs
+++ b/ratatui-widgets/src/paragraph.rs
@@ -14,14 +14,6 @@ use crate::{
     reflow::{LineComposer, LineTruncator, WordWrapper, WrappedLine},
 };
 
-const fn get_line_offset(line_width: u16, text_area_width: u16, alignment: Alignment) -> u16 {
-    match alignment {
-        Alignment::Center => (text_area_width / 2).saturating_sub(line_width / 2),
-        Alignment::Right => text_area_width.saturating_sub(line_width),
-        Alignment::Left => 0,
-    }
-}
-
 /// A widget to display some text.
 ///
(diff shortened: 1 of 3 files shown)


alexpasmantier:
> Composers which can render a given line "statelessely" (such as the LineTruncator) which actually end up doing a lot of unnecessary work (and on the critical rendering path) when the offset gets high.
>
>I came across this while experimenting with previews in [television](https://github.com/alexpasmantier/television) and noticed increased rendering lag as the preview offset got higher for large files.

Here's an example. The slowdown is kind of subtle but I thinks it's noticeable enough to a user (even if I agree this is somewhat of an edge case).

https://github.com/user-attachments/assets/253b680a-7757-4808-a0bb-ff571d712670

ratatui-widgets/src/paragraph.rs
```diff
@@ -449,15 +449,23 @@ impl Paragraph<'_> {
...
             line_composer.set_horizontal_offset(self.scroll.x);
-            self.render_text(line_composer, text_area, buf);
+            render_text(line_composer, text_area, buf);
         }
     }
 }
 
 impl<'a> Paragraph<'a> {
-    fn render_text<C: LineComposer<'a>>(&self, mut composer: C, area: Rect, buf: &mut Buffer) {
+    fn render_text<C: StatefulLineComposer<'a>>(
```
joshka, line 463: With the way that the paths for truncated vs wrapped text are made are different, this no longer needs to be generic over the type, and in fact probably doesn't even need to construct the wrapper/truncater outside the methods. Does making these two functions 'Paragraph::render_wrapped()` and `Paragraph::render_truncated()` make sense to you?

ratatui-widgets/src/paragraph.rs
```diff
@@ -489,6 +497,35 @@ impl<'a> Paragraph<'a> {
...
+                continue;
+            }
+            // If the symbol is empty, the last char which rendered last time will
+            // leave on the line. It's a quick fix.
+            let symbol = if symbol.is_empty() { " " } else { symbol };
+            buf[(area.left() + x, area.top() + y)]
+                .set_symbol(symbol)
+                .set_style(*style);
+            x += width as u16;
+        }
```
joshka, line 521: This chunk seems like an exact duplicate. Is there a way to not duplicate this (extract a method). (Side note: In general, we arrange supporting child methods below the caller)

ratatui-widgets/src/reflow.rs
```diff
@@ -8,10 +8,14 @@ use unicode_width::UnicodeWidthStr;
...
 /// iterators for that).
-pub trait LineComposer<'a> {
+pub trait StatefulLineComposer<'a> {
     fn next_line<'lend>(&'lend mut self) -> Option<WrappedLine<'lend, 'a>>;
 }
 
+pub trait LineComposer<'a> {
+    fn next_line(&mut self) -> Option<WrappedLine<'_, 'a>>;
+}
+
```
joshka, line 18: If there's only one impl of each of these, then it's likely not worth having a trait. The trait only existed to make the shared behavior accessible via the same methods. What would this look like without the traits?

ratatui-widgets/src/reflow.rs
```diff
@@ -375,31 +379,52 @@ mod tests {
             )
         });
 
-        let mut composer: Box<dyn LineComposer> = match which {
+        match which {
             Composer::WordWrapper { trim } => {
-                Box::new(WordWrapper::new(styled_lines, text_area_width, trim))
+                let mut composer = WordWrapper::new(styled_lines, text_area_width, trim);
```
joshka, line 384: Is there a way to write this that avoids the duplicate code?

joshka (commented):
I'm supportive of the idea, but I don't generally like adding complexity to this code as it's overly complex which makes it a burden to understand and maintain. Look for some simplifications that make for less code if possible here (and less duplication). I wonder if the traits can be removed entirely? Duplicate code tends to otherwise attract divergent behavior, which is a pain to maintain.

It would be good to be able to see the perf differences (either with existing benchmarks or new ones that demonstrate the issue that aligns with the video).

ratatui-widgets/src/paragraph.rs
alexpasmantier: Absolutely. 

TBH I hesitated to remove the traits entirely but didn't feel legitimate to do so.

joshka: Go for it. You're the current expert on how this code works having jumped into to understand it most recently. :D

alexpasmantier:
@joshka Simplified the code quite a bit (finally keeping that trait) and did a bit of benchmarking with criterion using the benches [here](https://github.com/ratatui/ratatui/blob/main/ratatui/benches/main/paragraph.rs).

Found out the benches were actually broken since they were playing with the horizontal scroll when I'm pretty confident the intent was to play with the `y` scroll so I fixed that and made sure we're always rendering a comparable amount of content whatever the scroll offset is.

Here are the results (in a nutshell):

**paragraph/render_scroll_half**
|                | 64 lines  | 2048 lines | u16::MAX lines |
|----------------|:---------:|:----------:|:--------------:|
| `main`           | 292.12 µs | 3.7583 ms  | 113.54 ms      |
| `current branch` | 279.93 µs | 290.45 µs  | 973.97 µs      |
| **speedup** | NA | **-92.268%** | **-99.133%** |

**paragraph/render_scroll_full**
|                | 64 lines  | 2048 lines | u16::MAX lines |
|----------------|:---------:|:----------:|:--------------:|
| `main`           | 316.93 µs | 7.1271 ms  | 228.04 ms      |
| `current branch` | 277.21 µs | 289.69 µs  | 988.34 µs      |
| **speedup** | **-13.585%** | **-95.924%** | **-99.565%** |

All other measures remaining more or less the same.

The plot below might show in a more obvious manner how the PR makes scrolling without wrapping almost independent of the number of lines vs before.

| `main` | `this branch` |
| :--: | :--: |
| Screenshot 2025-01-20 at 15 07 36 | Screenshot 2025-01-20 at 15 08 34 |

joshka:
Are the perf changes due to the change of the benches, or due to the change in implementation (or a combo perhaps)? Are you able to run the new benches against the code in main to double check that / show that the changes are all due to the algo change?

alexpasmantier:
Any thoughts on how to proceed with these changes?
Are there any requirements you feel aren't met yet and should be improved on?

joshka:
> Any thoughts on how to proceed with these changes? Are there any requirements you feel aren't met yet and should be improved on?

Added some small changes that simplify / clarify things a bit in https://github.com/ratatui/ratatui/pull/1622/commits/82f12c97016e162faca52b242cfa938c61774bd6 WDYT?

alexpasmantier:
That looks perfect! Nice job :-)

joshka:
Thanks for the PR

PR #1635, opened 2025-01-30
fix(chart): fix issue with mispositioned x-axis labels

Yomguithereal:
As promised in issue #334, here is a fix for the x-axis labels of the chart widget. I have also edited a related example to demonstrate the fix but I can drop this change if required.

The code is now using linear interpolation to position the center of the label rects under the x-axis.

### DIFF
--- a/ratatui-widgets/src/chart.rs
+++ b/ratatui-widgets/src/chart.rs
@@ -4,7 +4,7 @@ use std::{cmp::max, ops::Not};
 use ratatui_core::{
     buffer::Buffer,
     layout::{Alignment, Constraint, Flex, Layout, Position, Rect},
-    style::{Color, Style, Styled},
+    style::{Color, Style, Styled, Stylize},
     symbols::{self},
     text::Line,
     widgets::Widget,
@@ -539,6 +539,8 @@ pub struct Chart<'a> {
     /// The position determine where the length is shown or hide regardless of
     /// `hidden_legend_constraints`
     legend_position: Option<LegendPosition>,
+    /// Whether to show a grid underneath the graph
+    show_grid: bool,
 }
 
 impl<'a> Chart<'a> {
@@ -578,6 +580,7 @@ impl<'a> Chart<'a> {
             datasets,
             hidden_legend_constraints: (Constraint::Ratio(1, 4), Constraint::Ratio(1, 4)),
             legend_position: Some(LegendPosition::default()),
+            show_grid: false,
         }
     }
 
@@ -737,6 +740,15 @@ impl<'a> Chart<'a> {
     #[must_use = "method moves the value of self and returns the modified value"]
     pub const fn legend_position(mut self, position: Option<LegendPosition>) -> Self {
         self.legend_position = position;
+        self
+    }
+
+    /// Sets whether to show a grid underneath the graph.
+    ///
+    /// This is a fluent setter method which must be chained or used as it consumes self
+    #[must_use = "method moves the value of self and returns the modified value"]
+    pub const fn show_grid(mut self, show: bool) -> Self {
+        self.show_grid = show;
         self
     }
 
@@ -1052,9 +1064,9 @@ impl Widget for &Chart<'_> {
             }
 
             if self.x_axis.tick_marks {
-                if let Some(ticks) = x_ticks {
+                if let Some(ticks) = &x_ticks {
                     for x in ticks {
-                        buf[(x, y)].set_symbol(symbols::line::CROSS);
+                        buf[(*x, y)].set_symbol(symbols::line::CROSS);
                     }
                 }
             }
@@ -1068,9 +1080,9 @@ impl Widget for &Chart<'_> {
             }
 
             if self.y_axis.tick_marks {
-                if let Some(ticks) = y_ticks {
+                if let Some(ticks) = &y_ticks {
                     for y in ticks {
-                        buf[(x, y)].set_symbol(symbols::line::CROSS);
+                        buf[(x, *y)].set_symbol(symbols::line::CROSS);
                     }
                 }
             }
@@ -1180,6 +1192,38 @@ impl Widget for &Chart<'_> {
                     },
                     buf,
                 );
+            }
+        }
+
+        if self.show_grid {
+            if let Some(ticks) = x_ticks {
+                for x in ticks {
+                    for y in graph_area.top()..graph_area.bottom() {
+                        let cell = &mut buf[(x, y)];
+
+                        if cell.symbol() == " " {
+                            cell.set_symbol(symbols::line::VERTICAL)
+                                .set_style(Style::new().dim());
+                        }
+                    }
+                }
+            }
+
+            if let Some(ticks) = y_ticks {
+                for y in &ticks[1..ticks.len().saturating_sub(1)] {
+                    for x in graph_area.left()..graph_area.right() {
+                        let cell = &mut buf[(x, *y)];
+
+                        if cell.symbol() == " " {
+                            cell.set_symbol(symbols::line::HORIZONTAL)
+                                .set_style(Style::new().dim());
+                        } else if cell.symbol() == symbols::line::VERTICAL {
+                            cell.set_symbol(symbols::line::CROSS);
+                        } else {
+                            continue;
+                        }
+                    }
+                }
             }
         }
     }
--- a/ratatui-widgets/src/chart.rs
+++ b/ratatui-widgets/src/chart.rs
@@ -883,6 +883,8 @@ impl<'a> Chart<'a> {
         chart_area: Rect,
         graph_area: Rect,
     ) -> Option<Vec<u16>> {
+        // Linear interpolation returns a point situated at
+        // a percentage of a segment's endpoints.
         fn linear_interpolation(min: u16, max: u16, t: f64) -> u16 {
             ((1.0 - t) * f64::from(min) + t * f64::from(max)) as u16
         }
@@ -894,13 +896,16 @@ impl<'a> Chart<'a> {
             return None;
         }
 
-        let width_per_tick = graph_area.width / labels_len;
+        // We subtract 1 from the width per tick to make sure there is always
+        // at least one space between labels.
+        let width_per_tick = (graph_area.width / labels_len).saturating_sub(1);
+
         let mut ticks = Vec::with_capacity((labels_len - 1) as usize);
 
         let label_area = self.first_x_label_area(
             y,
             labels.first().unwrap().width() as u16,
-            width_per_tick,
+            width_per_tick / 2,
             chart_area,
             graph_area,
         );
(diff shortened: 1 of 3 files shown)


joshka (commented):
One thing that would help show that this is fixed properly would be some unit tests.

I'm envisaging a set of tests where there's a varying number of labels:
- 0 for edge case
- 1 for edge case check
- 2 for lowest meaningful check
- 3 for checking that the label is centered when odd numbers
- 4 for checking that the labels are appropriately placed when there's an even number
- a few larger numbers as spot checks

Then for the numbers greater than 3, try several different graph widths and label widths to show how these get aligned / truncated. Visually showing the output of these makes it easy to see where there's problems a lot easier than just math-ing it.

E.g. just comparing the label line of the graph output might be useful.

```rust
#[case::(&["a", "b", "c"], "a b c")]
#[case::(&["a", "b", "c"], "a  b  c")]
#[case::(&["a", "b", "c"], "a   b   c")]
#[case::(&["a", "b", "c", "d"], "a b c d")]
#[case::(&["a", "b", "c", "d"], "a   b   c   d")]
#[case::(&["aa", "b", "c"], "aa b c")]
fn labels(#[case]input: &[&'static str], expected: &'static str) {...}
```

Additionally, right now we know that we render the labels in the wrong place, so it would be helpful to have these unit tests as characterization tests in a commit that is written before the changes here, so that when the changes here happen, there's a corresponding change in the unit tests that shows the difference. This makes understanding the impact of the change easier to see at a glance (e.g. anyone can see `"a   b c"` became `"a  b  c"`)

Yomguithereal:
I have reworked the code, and added the tick marks and grid feature because it indeed makes the code easier to debug. Will now think about the unit tests, but I need more time.

ratatui-widgets/src/chart.rs
```diff
@@ -895,21 +928,45 @@ impl<'a> Chart<'a> {
...
 
+        // The first label will be drawn in the space offered by y labels on the left
         Self::render_label(buf, labels.first().unwrap(), label_area, label_alignment);
 
+        // Finding positions for the labels by incrementing the
+        // `width_per_tick` each step is not a good solution because
+        // it accumulates rounding error at each iteration.
+        // A correct solution is to find the center position for the
+        // label box, i.e. where the tick mark should be, in terms
+        // of percentage of the graph area's width.
```
joshka, line 939: Nit: for consistency with the rest of the lib, wrap comments at 100 chars. (I generally use https://marketplace.visualstudio.com/items?itemName=stkb.rewrap for this, Other editors / IDEs will have similar solutions.)

Why does this matter? Mostly it's an opinionated subjective preference. I don't like comments to take up large amounts of vertical space. 100 chars hits a good balance for most things in rust IMO while 80 chars often feels overly narrow.

ratatui-widgets/src/chart.rs
```diff
@@ -895,21 +928,45 @@ impl<'a> Chart<'a> {
...
+            // of the label's bounding box.
+            let label_area = Rect::new(x - width_per_tick / 2, y, width_per_tick, 1);
 
             Self::render_label(buf, label, label_area, Alignment::Center);
         }
 
-        let x = graph_area.right() - width_between_ticks;
-        let label_area = Rect::new(x, y, width_between_ticks, 1);
+        // Leaving as much space as possible for the last tick, all while respecting
+        // the average tick bounding box proportions.
```
joshka, line 961: This comment explains the result without the rationale. Why are we leaving this space?

joshka (commented):
Waiting on unit tests - looking good though. A couple of minor comments.

Yomguithereal:
Sorry @joshka. I **will** find the time to finalize this PR at some point. I am just swamped currently.

joshka:
> Sorry @joshka. I **will** find the time to finalize this PR at some point. I am just swamped currently.

No rush. Take your time. Based on my comments it was fairly close to done. If you can add the rationale about why the space that would help. I may be able to find some motivation to complete it in the coming week or two.

ratatui-widgets/src/chart.rs
Yomguithereal: I committed the desired rewrapping.

ratatui-widgets/src/chart.rs
Yomguithereal: Before I amend the comment, let me explain and see whether we are aligned on this particular topic: sometimes the last tick might have more or less columns available that what was allocated per tick on the axis because of rounding variations. The ideal scenario is to leave as much space to the last tick as possible but there are 2 issues with this: 1. this requires knowing the rendered width of the penultimate tick (can be done using `.width()` of course, but do we want to rely on this information?) and 2. it means the last tick might have a different proportion than the others, which is maybe something we also want to avoid?

joshka:
> Does this means we should center the tick box based on the label's width?

Yes. Also note that doing this consistently chooses how to center even values in odd spaces the same way always (I forget whether that's one character left or right, but there's tests for various centering type code for other types).

> The second thing is that each Alignement truncates the string on a different end (I guess ellipsis truncation should be done from userland beforehand?). This means the final label would not be cut the same way as precedent ones and in this screenshot, the 00% label is a bit confusing, no?

If there's an easy way to avoid truncation in the widget, then that makes the most sense. There's enough space to render 100% there even if it ends up right aligned there. The same may happen for the first item when there's not enough space due to no y-axis labels.

Quick question about the y axis zero point - should that be the base line and not a first tick?

This is looking great btw.

j-g00da:
This wasn't updated for a long time now, can I take it over?
@Yomguithereal 
@joshka

Yomguithereal:
Here is what remains to be done:

* Positioning the label boxes relative to the text's width to make sure the tick falls visually in the middle of the label
* Unit tests
* Decide what to do or not to do regarding label truncation (unicode ellipsis truncation is not as straightforward as it seems)

> Quick question about the y axis zero point - should that be the base line and not a first tick?

I guess it could indeed. Maybe it should be within the scope of a later commit?

@j-g00da I should be able to take care of first item quickly at least. For the rest is a bit sketchier time-wise for me.

j-g00da:
Thanks for the todo list!

> I should be able to take care of first item quickly at least.

This would be great! :)

You can ping me when you are done with your part so I can take it over and finish it.

Yomguithereal:
@j-g00da first item is taken care of by [87bacf0](https://github.com/ratatui/ratatui/pull/1635/commits/87bacf03580ac4ae2d4bc708c198e52a9ca3132a)

As for your question @joshka:

> Quick question about the y axis zero point - should that be the base line and not a first tick?

Here is the reason why not :)

Screenshot from 2025-06-04 16-29-00

Points with a 0 on the y coordinate will overwrite x axis, if you want to keep the proportions of the graph correct.

I edited the `chart` example with the following dataset to make the problem apparent:

```rust
  let dataset = Dataset::default()
        .name("Stonks")
        .marker(Marker::Braille)
        .graph_type(GraphType::Line)
        .style(Color::Blue)
        .data(&[
            (0.0, 1.0),
            (1.0, 3.0),
            (2.0, 0.0), // Here
            (3.0, 2.5), // Here
            (4.0, 0.8),
            (5.0, 4.0),
            (6.0, 1.0),
            (7.0, 6.0),
            (8.0, 3.0),
            (10.0, 10.0),
        ]);
```

joshka:
That actually looks kind of nice really.
It would be worth considering how best to display zero points like this.
How does that look with the code prior to this change?

Yomguithereal:
> How does that look with the code prior to this change?

Screenshot from 2025-06-06 09-54-54

j-g00da:
> @j-g00da first item is taken care of by [87bacf0](https://github.com/ratatui/ratatui/pull/1635/commits/87bacf03580ac4ae2d4bc708c198e52a9ca3132a)
> 
> 
> 
> As for your question @joshka:
> 
> 
> 
> > Quick question about the y axis zero point - should that be the base line and not a first tick?
> 
> 
> 
> Here is the reason why not :)
> 
> 
> 
> Screenshot from 2025-06-04 16-29-00

This makes more sense to me and IMO looks good. Maybe this should be configurable? But then I think I would prefer the starting point of both x and y be offset by 1.

x=0 being at the origin and y=0 being one character higher just feels off.

Yomguithereal:
One edge case you might want to consider also is that if the first y axis label is wide and the first x axis label is also wide you will not have the one line visual separation between both I think. It might look ugly, I dunno :)

joshka:
I agree with @j-g00da here. It makes sense for the default on both to be offset by 1, but also to make the zero axis point configurable on both. It could be reasonable for the axis to be fully movable such that graphs which swing positive and negative around the axis are able to be handled. I'm guessing that currently we only allow the axes to be rendered bottom and left?

I know that likely blows the implementation out a bit from a simple fix, so perhaps let's stay with keeping this PR as 1 based to match the existing behavior, and add another one to make the axis fully configurable. (Unless you think it makes more sense to work through that part of the problem in this PR because that is really coupled with the problem).

Yomguithereal:
> I'm guessing that currently we only allow the axes to be rendered bottom and left?

Indeed. It would be nice at some point to have the option to have the axis elsewhere, or even one axis on the right and another on the left for a different series (even if I would argue this is a dubious dataviz practice to mix different scales in the same space).

> I know that likely blows the implementation out a bit from a simple fix, so perhaps let's stay with keeping this PR as 1 based to match the existing behavior, and add another one to make the axis fully configurable. (Unless you think it makes more sense to work through that part of the problem in this PR because that is really coupled with the problem).

I agree it could be part of a different PR. I think this should remain an option though, because some people might like one way or the other.

joshka:
I invited @sbarral over from https://github.com/ratatui/ratatui/issues/2224 to take a look at this.

sbarral:
As I replied in issue #2224 , I wasn't aware of this effort and certainly do not want to derail it. My suggestion was less ambitious, but also had different goals. Namely, my focus was on drawing ticks, but only when they can be positioned exactly. In the general case, and assuming that extra symbols from _Symbols for Legacy Computing Supplement_ can be used, this only covers the case of 2 or 3 ticks.

My understanding is that this PR outlines a more ambitious goal of drawing every tick but consciously accepting its drawback, namely the nearest-neighbor (mis)positioning. Likewise, its use of vanilla box-drawing symbols promotes wider compatibility, but incurs a loss of 1 cell in the height (which is a bit unfortunate--I am often battling with vertical real-estate).

So within this different set of objectives, my only suggestion would be *not* to align the ticks with the axes. As I understand it, the x-axis is anyway already misaligned. Aligning the y-axis with the first tick means that we cannot draw the data at the first x-tick location without overwriting the y-axis, and the data at the first x-tick location is often important.
I personally do not see the axis/ticks misalignment as a problem, and even consider it a better practice when the first tick is not x=0 or y=0.

Below is an illustration of the differences with my approach, with the caveat again that my approach does not attempt to draw more than the ticks at the bounds and in the center of the interval.

This PR (note the misalignment of the x=1 and y=0 lines with the ticks):

Image

My solution (uses additional symbols):

Image

[Edit: replaced raw characters by images]

```

Companion:
Hi @Yomguithereal I opened #2751 to address the x axis label positioning issue discussed here I wanted to link it rather than duplicate your work silently and would appreciate your thoughts on the approach

Both changes address the same root cause but #2751 is deliberately limited to label positioning:

* It does not add the `tick_marks` or `show_grid` APIs, change y axis rendering or modify the chart example
* It uses integer interpolation across the actual plotted data columns avoiding floating point positioning and accumulated rounding
* It preserves the existing first and last label handling while centring intermediate labels on their corresponding plotted data positions
* Intermediate label areas are bounded by endpoint and neighbouring labels so long labels are truncated instead of overwriting one another
* Tests cover hand calculated positions rendered output, plotted data alignment narrow areas and overlapping labels

Your analysis here was useful in identifying the original segment counting problem and ive referenced this PR in #2751 any feedback on whether the narrower implementation misses relevant behaviour from your work would be welcome

Yomguithereal:
@Companion I don't have much time currently to review your code unfortunately but I don't have any issue with your PR superseding mine if you fix stuff I did not yet address. I must admit I don't even remember what were the remaining blocking points were.

Companion:
> @Companion I don't have much time currently to review your code unfortunately but I don't have any issue with your PR superseding mine if you fix stuff I did not yet address. I must admit I don't even remember what were the remaining blocking points were.

No worries thanks for confirming! ive kept #2751 focused on x axis label positioning and added coverage for plotted data alignment, wide label truncation and preventing adjacent labels from overwriting each other your original analysis was very helpful.

PR #1636, opened 2025-01-31
fix: avoid extra line break on whitespace only lines when wrapping paragraphs

dotdash:
Currently whitespace only lines produces an extra line break when trimming is disabled, because both the trimmed as well as the non-trimmed line get inserted. Fix this by only inserting the non-trimmed one.


orhun (approved):
LGTM!

joshka:
Thanks for the PR

PR #1652, opened 2025-02-05
feat: add ratatui-macros crate

joshka:
I think the time has come to put the ratatui-macros crate into the main repo. This is a unrelated histories merge of https://github.com/ratatui/ratatui-macros/.

The only relevant commits are:

prep commit: 92c4078413fc79fcc83f5d3d8708abb58696ff1a
merge commit: d95a67549131e2c72de7e82e258030e815ea1bb5

These:
- remove the github workflows
- remove husky pre-commit / pre-push hooks
- move the folder under ratatui-macros.
- add a macros feature that imports the macros under ratatui::macros

> [!CAUTION]
> This PR should be merged rather than squashed

As a side note, I've come around to the constraint macros being ok. I wouldn't personally use them as I think they're a bit magical for my personal preferences, but I think they're a fine thing to publish in the library.


ratatui-macros/CHANGELOG.md
orhun, line 1: Don't we want to publish a single changelog for the all the members of the workspace via `release-plz`? In that case this file should be removed.

ratatui-macros/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "ratatui-macros"
+version = "0.7.0-alpha.0"
+edition = "2021"
+authors = ["The Ratatui Developers"]
+description = "Macros for Ratatui"
+license = "MIT"                                              # or any other appropriate license
+repository = "https://github.com/ratatui-org/ratatui-macros"
```
orhun, line 8:
```suggestion
repository = "https://github.com/ratatui/ratatui"
```

ratatui-macros/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+documentation = "https://docs.rs/ratatui-macros"
+keywords = ["ratatui", "macros", "tui", "ui"]
+categories = ["command-line-interface"]
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
orhun, line 13:
```suggestion
```

ratatui-macros/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+
+[dependencies]
+ratatui-core = "0.1.0-alpha.2"
+ratatui-widgets = "0.3.0-alpha.1"
```
orhun, line 17: Depend on the crates from the workspace

orhun (commented):
Nice!

ratatui-macros/CHANGELOG.md
joshka: If we go that direction, it can happen as a single atomic step. Showing this file as a delete and then a recreate from the automation is a good thing to see (in a future PR).

ratatui-macros/Cargo.toml
joshka: d95a67549...d4415204e

joshka: d95a67549...d4415204e

ratatui-macros/README.md
```diff
@@ -0,0 +1,200 @@
...
+
+## Contributing
+
+Contributions to `ratatui-macros` are welcome! Whether it's submitting a bug report, a feature
+request, or a pull request, all forms of contributions are valued and appreciated.
+
+[Crates.io badge]: https://img.shields.io/crates/v/ratatui-macros?logo=rust&style=flat-square
+[License badge]: https://img.shields.io/crates/l/ratatui-macros
+[CI Badge]:
+  https://img.shields.io/github/actions/workflow/status/ratatui-org/ratatui-macros/ci.yml?logo=github&style=flat-square
```
orhun, line 193:
```suggestion
  https://img.shields.io/github/actions/workflow/status/ratatui/ratatui-macros/ci.yml?logo=github&style=flat-square
```

ratatui-macros/README.md
```diff
@@ -0,0 +1,200 @@
...
+[Docs.rs badge]: https://img.shields.io/docsrs/ratatui-macros?logo=rust&style=flat-square
+[Crate Downloads badge]: https://img.shields.io/crates/d/ratatui-macros?logo=rust&style=flat-square
+[ratatui_macros crate]: https://crates.io/crates/ratatui-macros
+[API Docs]: https://docs.rs/ratatui-macros
+[CI Status]: https://github.com/kdheepak/ratatui-macros/actions
+[Ratatui]: https://github.com/ratatui-org/ratatui
```
orhun, line 199:
```suggestion
[Ratatui]: https://github.com/ratatui/ratatui
```

ratatui-macros/README.md
```diff
@@ -0,0 +1,200 @@
...
+
+[Crates.io badge]: https://img.shields.io/crates/v/ratatui-macros?logo=rust&style=flat-square
+[License badge]: https://img.shields.io/crates/l/ratatui-macros
+[CI Badge]:
+  https://img.shields.io/github/actions/workflow/status/ratatui-org/ratatui-macros/ci.yml?logo=github&style=flat-square
+[Docs.rs badge]: https://img.shields.io/docsrs/ratatui-macros?logo=rust&style=flat-square
+[Crate Downloads badge]: https://img.shields.io/crates/d/ratatui-macros?logo=rust&style=flat-square
+[ratatui_macros crate]: https://crates.io/crates/ratatui-macros
+[API Docs]: https://docs.rs/ratatui-macros
+[CI Status]: https://github.com/kdheepak/ratatui-macros/actions
```
orhun, line 198: We should update this link

kdheepak:
I'm open to dropping the `constraints!` macro from this PR. We don't know how it will pan out with taffy in the future. We can always add it in a future release if we'd like.

joshka:
> I'm open to dropping the `constraints!` macro from this PR. We don't know how it will pan out with taffy in the future. We can always add it in a future release if we'd like.

This would actually be a little more complicated than is strictly necessary. I think it's fine to keep this and make the taffy stuff additive rather than a replacement.

joshka:
@kdheepak brought up that the commit history will automatically link to incorrect issues due to the bare PR ref numbers in the commit messages. This is annoying, but I think noting it is all we can really do here. GitHub should fix this thing, but it likely will never do so.

orhun:
that's unfortunate... is squashing all those commits a naughty thing to do? 👀

joshka:
> that's unfortunate... is squashing all those commits a naughty thing to do? 👀

My reason for grafting the history like this is that it's often useful to be able to see the rationale for why something happened a particular way. Often that need is discovered long after the fact. You may not need it today, but you cannot know that you won't need it in the future. I know that the source repo will likely stick around forever as an archive, but then the history is bifurcated, which is annoying when digging through the history for any reason.

A particular use case for this is learning from more experienced folk about why they did something, I like that it's possible to dig into the design decisions of open source project and throwing away all that info is a pain.

orhun (approved):

kdheepak:
Before merging, it might be worth interactively rebasing this branch and just editing all the commit messages to remove the `#xx`, keeping everything the same.

joshka:
The effort involved in doing that feels like it outweighs any gain from doing so. The commits that are in this PR will have been linked to the original issues (I think that's permanent, but not 100% sure if the commits cease to exist - subject to github's implementation whims). I'm significantly more inclined to squash merge rather than to edit previous commits to remove the PR linkage.

orhun:
Makes sense. I'm okay with proceeding with merging this. No additional remarks.

PR #1662, opened 2025-02-09
chore: sort dependencies in Cargo.toml

canac:
Move `termion` up in the dependencies list in `Cargo.toml` so that they are sorted alphabetically and so that the `unicode-width` comment is directly above the `unicode-width` dependency.


joshka (approved):

joshka:
Thanks

PR #1668, opened 2025-02-11
build: Building ratatui fails on windows

thscharler:
I've tried to build rat-salsa without using a dependency on ratatui, just ratatui-core and ratatui-widgets.

These two dev-dependencies make the build fail on windows. 

With them cargo always tries to build termion too.

I don't know if they are needed for other things, but without them I could build everything after fixing my imports.

### DIFF
--- a/ratatui-crossterm/Cargo.toml
+++ b/ratatui-crossterm/Cargo.toml
@@ -40,7 +40,7 @@ instability.workspace = true
 ratatui-core = { workspace = true }
 
 [dev-dependencies]
-ratatui = { path = "../ratatui", features = ["crossterm"] }
+#ratatui = { path = "../ratatui", features = ["crossterm"] }
 rstest.workspace = true
 
 [lints]
--- a/ratatui-widgets/Cargo.toml
+++ b/ratatui-widgets/Cargo.toml
@@ -63,7 +63,7 @@ unicode-width.workspace = true
 [dev-dependencies]
 color-eyre.workspace = true
 pretty_assertions.workspace = true
-ratatui = { path = "../ratatui" }
+# ratatui = { path = "../ratatui" }
 rstest.workspace = true
 
 [lints]


ratatui-crossterm/Cargo.toml
```diff
@@ -40,7 +40,7 @@ instability.workspace = true
 ratatui-core = { workspace = true }
 
 [dev-dependencies]
-ratatui = { path = "../ratatui", features = ["crossterm"] }
+#ratatui = { path = "../ratatui", features = ["crossterm"] }
```
orhun, line 43: Maybe try `[target.'cfg(not(windows))'.dev-dependencies]` instead of commenting this out?

The CI is failing due to this and we cannot accept this PR in this state.

orhun (commented):
Strange, `ratatui-crossterm` shouldn't be built on Windows...

Unfortunately none of the devs have a Windows system (IRC) so this makes the testing a bit difficult on our side.

ratatui-crossterm/Cargo.toml
joshka: > Strange, ratatui-crossterm shouldn't be built on Windows...

You're thinking ratatui-termion. ratatui-crossterm [builds fine on Windows in CI](https://github.com/ratatui/ratatui/actions/runs/13273214121/job/37057282301#step:5:17).

It's entirely possible that these deps should not be in here, but right now they're passing all CI steps fine, so I'd like to understand what's failing before fixing it. I can spin up a Windows VM, but would like some more details about what the problem is before doing that as the only info I currently have is that it works fine. @thscharler can you post details about the build failure and perform some diagnostic steps to understand what it's doing. I.e. look at `cargo tree` and `cargo tree --edges features` etc.

thscharler: I'm sorry, I left out an important detail.

I was trying to use the repository as a local dependency. 
To bypass the ratatui workspace I created a new one:

```
[workspace]
resolver = "2"
members = [
    "rat-salsa/rat-text",
]

[workspace.package]
authors = []
documentation = ""
repository = ""
homepage = ""
keywords = []
categories = []
readme = ""
license = "MIT"
edition = "2021"
rust-version = "1.74.0"

[workspace.lints.rust]
unsafe_code = "forbid"

[workspace.dependencies]
bitflags = "2.7.0"
color-eyre = "0.6.3"
crossterm = "0.28.1"
document-features = "0.2.7"
indoc = "2.0.5"
instability = "0.3.7"
itertools = "0.13.0"
pretty_assertions = "1.4.1"

ratatui = { path = "ratatui/ratatui", version = "0.30.0-alpha.1", features = ["unstable-widget-ref" ] }
ratatui-core = { version = "0.1.0-alpha.2", path="ratatui/ratatui-core"}
ratatui-crossterm = { path = "ratatui/ratatui-crossterm", version = "0.1.0-alpha.1" }
ratatui-macros = { path = "ratatui/ratatui-macros", version = "0.7.0-alpha.0" }
ratatui-termion = { path = "ratatui/ratatui-termion", version = "0.1.0-alpha.1" }
ratatui-termwiz = { path = "ratatui/ratatui-termwiz", version = "0.1.0-alpha.1" }
ratatui-widgets = { path = "ratatui/ratatui-widgets", version = "0.3.0-alpha.1" }

rstest = "0.24.0"
serde = { version = "1.0.217", features = ["derive"] }
serde_json = "1.0.138"
strum = { version = "0.26.3", features = ["derive"] }
termion = "4.0.0"
termwiz = { version = "0.22.0" }
unicode-segmentation = "1.12.0"
# See <https://github.com/ratatui/ratatui/issues/1271> for information about why we pin unicode-width
unicode-width = "=0.2.0"
```

In my crate I only referenced ratatui-core and ratatui-widgets. 

With these two lines cargo starts to pull ratatui with what it seems all features enabled, which 
obviously wont work on windows. 

When I look at it today it seems like a mediorce idea at best, and I shouldn't have bothered you with this PR.

joshka: No probs. Thanks for the update.

Issue #1667, opened 2025-02-10
Paragraph panics when drawn outside of frame


jwodder:
The following code:

```rust
use ratatui::{
    crossterm::event::{self, Event, KeyEvent},
    layout::Rect,
    widgets::Paragraph,
    DefaultTerminal,
};

fn main() -> std::io::Result<()> {
    let terminal = ratatui::init();
    let r = run(terminal);
    ratatui::restore();
    r
}

fn run(mut terminal: DefaultTerminal) -> std::io::Result<()> {
    loop {
        terminal.draw(|frame| {
            frame.render_widget(
                Paragraph::new("Beyond the pale"),
                Rect {
                    x: 200,
                    y: 200,
                    width: 10,
                    height: 10,
                },
            );
        })?;
        if let Event::Key(KeyEvent { .. }) = event::read()? {
            break;
        }
    }
    Ok(())
}
```

when run on an insufficiently large terminal panics with:

```
thread 'main' panicked at /Users/jwodder/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ratatui-core-0.1.0-alpha.2/src/buffer/buffer.rs:253:13:
index outside of buffer: the area is Rect { x: 0, y: 0, width: 203, height: 62 } but index is (200, 200)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

If `Text` or `Line` is used instead of `Paragraph`, then the program draws nothing instead of panicking, which is what I believe should happen.  I have not checked any other widgets.

## Environment

- OS: macOS 14.7.2
- Terminal Emulator: wezterm 20240203-110809-5046fc22
- Font: —
- Crate version: both 0.29.0 and 0.30.0-alpha.1
- Backend: crossterm 0.28.1

joshka:
Thanks for the bug report, would you consider submitting a PR to fix this?

jwodder:
@joshka PR to fix this: https://github.com/ratatui/ratatui/pull/1670

joshka:
This one should probably be finished up for 0.30. I need to take another look at the PR and get back up to speed on it.

PR #1670, opened 2025-02-11
fix: Fix panic when rendering a `Paragraph` out of bounds

jwodder:
Fixes #1667.

### DIFF
--- a/ratatui-widgets/src/paragraph.rs
+++ b/ratatui-widgets/src/paragraph.rs
@@ -406,6 +406,7 @@ impl Widget for Paragraph<'_> {
 
 impl Widget for &Paragraph<'_> {
     fn render(self, area: Rect, buf: &mut Buffer) {
+        let area = area.intersection(buf.area);
         buf.set_style(area, self.style);
         self.block.as_ref().render(area, buf);
         let inner = self.block.inner_if_some(area);
@@ -500,6 +501,7 @@ mod tests {
     use ratatui_core::style::{Color, Modifier, Style, Stylize};
     use ratatui_core::text::{Line, Span, Text};
     use ratatui_core::widgets::Widget;
+    use rstest::rstest;
 
     use super::*;
     use crate::block::TitlePosition;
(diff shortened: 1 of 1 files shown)


orhun (approved):

ratatui-widgets/src/paragraph.rs
```diff
@@ -477,7 +477,11 @@ fn render_line(wrapped: &WrappedLine<'_, '_>, area: Rect, buf: &mut Buffer, y: u
         // Make sure to overwrite any previous character with a space (rather than a zero-width)
         let symbol = if symbol.is_empty() { " " } else { symbol };
         let position = Position::new(area.left() + x, area.top() + y);
-        buf[position].set_symbol(symbol).set_style(*style);
+        if let Some(cell) = buf.cell_mut(position) {
+            cell.set_symbol(symbol).set_style(*style);
+        } else {
+            break;
+        }
```
joshka, line 484: Another approach that might also be worth considering for this is pre-trimming the area function in the render function. (e.g. `let area = buf.area.intersection(area);` But I think that would change how paragraphs that are partially out of the area are rendered, which may not be desirable. Longer note about this in the test below.

ratatui-widgets/src/paragraph.rs
```diff
@@ -1203,4 +1207,18 @@ mod tests {
...
     }
+
+    #[test]
+    fn test_render_paragraph_out_of_bounds() {
+        let paragraph = Paragraph::new("Beyond the pale");
+        for area in [
+            Rect::new(0, 5, 15, 1),
+            Rect::new(20, 0, 15, 1),
+            Rect::new(20, 5, 15, 1),
+        ] {
```
joshka, line 1218: These areas are fully outside of the buffer area. So the tests here are perfectly correct. I wonder though if there's a second edge case (as mentioned above) where the render area is partially outside the buffer area. There's a couple of ways that this could be intended to render:
1. crop the text outside of the buffer area
2. restrict the paragraph to the buffer area so that it fills correctly

Both of these seem like they'd be useful in some cases. The first one would be useful where you have moving text which you're fine with rendering partially off screen. The second is where you have text which you want to render fully regardless of the screen bounds. So I'm not 100% sure what the default behavior should be here. Either way, it's worth capturing what our current behavior is as a characterization test and then we can work out what it should be later.

jwodder: @joshka So... are you asking for further changes to this PR, or are these things for other people to do in later PRs?

j-g00da: I think 1 seems like a good default behavior (so current implementation), one can always do an intersect if they want it to be restricted.

> Either way, it's worth capturing what our current behavior is as a characterization test and then we can work out what it should be later.

Agree

j-g00da:
@jwodder would you like to continue working on this PR, or would you prefer that I take it over?

jwodder:
@j-g00da You take it over.

joshka:
whoops - beat you to it :)

PR #1684, opened 2025-02-24
feat: implement styled for other primitives

aschey:
Implements `Styled` for other primitive types so you can do something like `1.red()` instead of `1.to_string().red()`. I also added the implementation for `Cow<'a str>` since it was missing. It would be nice to have a blanket impl of `Styled` for any type implementing `Display`, but that would mean calling `to_string()` on borrowed types and creating extra allocations, so I think this will have to suffice.


orhun (approved):
I can't think of any side effects of this so LGTM!

PR #1706, opened 2025-03-06
fix: examples moving selection twice per key press

TeamDman:
This caused two movements for a single key event


joshka:
Thanks for the PR - I've got another PR just now (https://github.com/ratatui/ratatui/pull/1707) that also modifies all the examples. I updated the same code there to check for Press events. The next version of crossterm should also have some nice additions for checking this more simply (e.g. `event::read()?.as_key_press()`).

PR #1707, opened 2025-03-07
feat: add ratatui::run() method

joshka:
This introduces a new `ratatui::run()` method which runs a closure with
a terminal initialized with reasonable defaults for most applications.
This calls `ratatui::init()` before running the closure and
`ratatui::restore()` after the closure completes, and returns the result
of the closure.

A minimal hello world example using the new `ratatui::run()` method:

```rust
use crossterm::event;

fn main() -> Result<(), Box<dyn std::error::Error>> {
    ratatui::run(|terminal| loop {
        terminal.draw(|frame| frame.render_widget("Hello, world!", frame.area()))?;
        if matches!(event::read()?, event::Event::Key(_)) {
            break Ok(());
        }
    })
}
```

Of course, this also works both with apps that use free methods and structs:

```rust
fn run(terminal: &mut DefaultTerminal) -> Result<(), AppError> { ... }

ratatui::run(run)?;
```

```rust
struct App { ... }

impl App {
    fn new() -> Self { ... }
    fn run(mut self, terminal: &mut DefaultTerminal) -> Result<(), AppError> { ... }
}

ratatui::run(|terminal| App::new().run(terminal))?;
```

### DIFF
--- a/ratatui-widgets/examples/sparkline.rs
+++ b/ratatui-widgets/examples/sparkline.rs
@@ -22,28 +22,19 @@ use ratatui::layout::{Constraint, Layout, Rect};
 use ratatui::style::{Color, Style, Stylize};
 use ratatui::text::{Line, Span};
 use ratatui::widgets::{RenderDirection, Sparkline};
-use ratatui::{DefaultTerminal, Frame, symbols};
+use ratatui::{Frame, symbols};
 
 fn main() -> Result<()> {
     color_eyre::install()?;
-    let terminal = ratatui::init();
-    let result = run(terminal);
-    ratatui::restore();
-    result
-}
-
-/// Run the application.
-fn run(mut terminal: DefaultTerminal) -> Result<()> {
-    loop {
-        terminal.draw(render)?;
-        // Ensure that the animations renders at 50 FPS (GIF speed)
-        if !event::poll(Duration::from_secs_f64(1.0 / 50.0))? {
-            continue;
+    let frame_timeout = Duration::from_secs_f64(1.0 / 60.0); // run at 60 FPS
+    ratatui::run(|terminal| {
+        loop {
+            terminal.draw(render)?;
+            if event::poll(frame_timeout)? && event::read()?.is_key_press() {
+                break Ok(());
+            }
         }
-        if event::read()?.is_key_press() {
-            return Ok(());
-        }
-    }
+    })
 }
 
 /// Render the UI with various sparklines.
--- a/examples/apps/advanced-widget-impl/src/main.rs
+++ b/examples/apps/advanced-widget-impl/src/main.rs
@@ -21,10 +21,7 @@ use ratatui::widgets::{Widget, WidgetRef};
 
 fn main() -> Result<()> {
     color_eyre::install()?;
-    let terminal = ratatui::init();
-    let result = App::default().run(terminal);
-    ratatui::restore();
-    result
+    ratatui::run(|terminal| App::default().run(terminal))
 }
 
 #[derive(Default)]
@@ -36,15 +33,15 @@ struct App {
 }
 
 impl App {
-    fn run(mut self, mut terminal: DefaultTerminal) -> Result<()> {
+    fn run(mut self, terminal: &mut DefaultTerminal) -> Result<()> {
         while !self.should_quit {
-            self.draw(&mut terminal)?;
+            self.render(terminal)?;
             self.handle_events()?;
         }
         Ok(())
     }
 
-    fn draw(&mut self, tui: &mut DefaultTerminal) -> Result<()> {
+    fn render(&mut self, tui: &mut DefaultTerminal) -> Result<()> {
         tui.draw(|frame| frame.render_widget(self, frame.area()))?;
         Ok(())
     }
--- a/examples/apps/calendar-explorer/src/main.rs
+++ b/examples/apps/calendar-explorer/src/main.rs
@@ -22,21 +22,18 @@ use time::{Date, Month, OffsetDateTime};
 
 fn main() -> Result<()> {
     color_eyre::install()?;
-    let terminal = ratatui::init();
-    let result = run(terminal);
-    ratatui::restore();
-    result
+    ratatui::run(run)
 }
 
 /// Run the application.
-fn run(mut terminal: DefaultTerminal) -> Result<()> {
+fn run(terminal: &mut DefaultTerminal) -> Result<()> {
     let mut selected_date = OffsetDateTime::now_local()?.date();
     let mut calendar_style = StyledCalendar::Default;
     loop {
         terminal.draw(|frame| render(frame, calendar_style, selected_date))?;
         if let Some(key) = event::read()?.as_key_press_event() {
             match key.code {
-                KeyCode::Char('q') | KeyCode::Esc => return Ok(()),
+                KeyCode::Char('q') => break Ok(()),
                 KeyCode::Char('s') => calendar_style = calendar_style.next(),
                 KeyCode::Char('n') | KeyCode::Tab => selected_date = next_month(selected_date),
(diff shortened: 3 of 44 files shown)


kdheepak:
I like it! Thanks for making it easy to review by splitting the commits + pointing us to the exact diff. You make this too easy :) 

Would it be useful to get some user feedback on this? I'm assuming we are going to be advocating for this to be the new way users are introduced to ratatui.

joshka:
> Would it be useful to get some user feedback on this? I'm assuming we are going to be advocating for this to be the new way users are introduced to ratatui.

Yup. Mentioned it in discord. This started named with_terminal, but I think run works much better.

TadoTheMiner:
Perfect, but why take a mutable reference and not give ownership?

orhun (approved):
LGTM.

I'm okay with this being `crossterm` specific, like the other convenience functions such as `init` etc.

For the naming, maybe we can do something like `run_with_terminal` but I mean... what else are we going to run it with? Oh! Maybe `run_with_web` in [ratzilla](https://docs.rs/ratzilla/) in the future. I think `run` should be good for now.

P.S. `0.30.0` release is going to be really packed 🫠

joshka:
> I don't see an advantage of giving a mutable reference over ownership. I think it makes more sense to give ownership to give the user more control over the terminal.

Choosing to move the terminal rather than pass a mutable reference here cedes control over when the terminal value is dropped to the closure. Passing it as a mutable reference means that it will be dropped after the closure completes, just before the terminal is restored.

Also, none of the methods on Terminal require ownership, so there's no real upside to doing so that I can see.

Some questions to think about to help clarify this:
- Is there something specific you're looking for that would need ownership of the terminal?
- Is there some specific downside you're trying to avoid with not wanting this to be a mutable ref?
- Are the problematic cases advanced enough that you might generally want to avoid the simple method? If you grab a copy of this PR locally, there's around 7 examples where I didn't update the examples to use `ratatui::run()` because those places did more interesting things than could be easily represented with this method (enabling mouse capture, running with inline options, etc.). They might inform real world cases where this method isn't useful. Are there others you know about?
- Are there other libraries you can point to that have similar approaches where a closure is used to give temporary access to an initialized value that will be restored after the call where this is provided as an owned value rather than a mutable ref?

joshka:
From [discord](https://discord.com/channels/1070692720437383208/1347410533820923914):

>8:37 PM]raylu: seems great? is there any downside? if you don't want it you can just use whatever you were doing before, right?
[10:30 PM]joshka: The main downsides I can think of are:
this is crossterm specific (like init and restore)
a closure as the primary interface is a little more complex than other approaches
the generic name (run vs run_with_terminal or something like that)
the lack of an approach that would be additive to the use case
>
> For each of those I really come up on the side of "yeah, but so what?":
we've already past that mile marker for init and restore, this is harmonious with the previous decision
it's 2025, closures are part of normal developer mindset
run is simple and obvious
there might be better ideas, show me...

coolreader18:
Would it make sense to have `restore` be called in a drop guard? That way a panic won't leave the terminal in raw mode.

joshka:
> Would it make sense to have `restore` be called in a drop guard? That way a panic won't leave the terminal in raw mode.

Doing the right thing in a terminal on panics, errors, and normal flow is surprisingly difficult to get right, and you end up choosing an option that's not necessarily ideal, but good enough. `ratatui::init()` is sort of that compromise. AI coding tools like CoPilot regularly tell me that I might have been a little wrong on that one by completing it as `ratatui::init()?` implying that the method intuitively should have returned a result. But we do have a `try_init()` method for that use case.

The major problem with this idea is that the panic hooks which display the error information run before the drop methods. In general apps that crash should disable raw mode and leave the alternate screen before displaying the panic messages. `ratatui::init()` sets up a panic handler that calls `ratatui::restore()` before handling a panic.

The second (less impactful) problem is that `Terminal` doesn't own the application's raw/alternate screen state. The user's app does. Users can still create apps where they choose not to use the `ratatui::init()` and `ratatui::restore()` methods. The expectation there should be that the user gets to decide when to enable / disable these states. This is meaningful for the narrow use cases where apps call out to other apps (e.g. an editor) and want to be in full control of these states before and after.

Third, the code to actually do the restore can in some circumstances itself crash, so we want that code to be intentionally part of the app's flow rather than part of the drop scope handling. Otherwise the error messages are a pain.

The metric for validating how well `init()` / `restore()` works comes down to looking at how often we get people asking about why their terminal is not being correctly restored. The only messages I recall seeing about those since these methods were introduced is by people not following the gilded path which this PR codifies into a small run method.

TadoTheMiner:
I personally store the terminal in a separate struct in my still in development xkcd viewer. It being a mutable reference wouldn't imply much except me adding a few &muts and 'as. It would be a minor annoyance but a minor annoyance is still worse than no annoyance

ratatui-widgets/examples/sparkline.rs
```diff
@@ -24,29 +24,22 @@ use ratatui::{
...
-
-/// Run the application.
-fn run(mut terminal: DefaultTerminal) -> Result<()> {
-    loop {
-        terminal.draw(draw)?;
-        if event::poll(Duration::from_millis(16))? && matches!(event::read()?, Event::Key(_)) {
+    let frame_timeout = Duration::from_secs_f64(1.0 / 60.0); // run at 60 FPS
+    ratatui::run(|terminal| loop {
+        terminal.draw(render)?;
+        if event::poll(frame_timeout)? && matches!(event::read()?, Event::Key(_)) {
```
TeamDman, line 35:
```suggestion
        if event::poll(frame_timeout)? && matches!(event::read()?, Event::Key(x) if x.kind == KeyEventKind::Press ) {
```

joshka: For the most part, because this immediately exits, I'm not too concerned about checking for whether it's a key press or release. Currently unreleased in crossterm there is code that adds an `Event::is_key_press()` method, so these checks will become e.g. `if event::poll(frame_timeout)? && event::read()?.is_key_press() {` when that is released.

The only real concern that really require this is that I have heard it rumored that sometimes windows terminals will pass through the key release event of the enter key that was used to run the command that runs the app. I don't use windows often enough to have seen this personally though, so it's a hypothetical problem right now.

joshka: > The only real concern that really require this is that I have heard it rumored that sometimes windows terminals will pass through the key release event of the enter key that was used to run the command that runs the app. I don't use windows often enough to have seen this personally though, so it's a hypothetical problem right now.

I've seen this actually happen a few times since then, but the change would be to check `event::read()?.is_key_press()` which was introduced in crossterm 0.29 instead.

orhun:
Should we rebase and finalize this PR?

joshka:
Yep. This is on my list for today. The examples changed enough that it doesn't clearly rebase easily, so it's a bit of work to do this.

ratatui-widgets/examples/sparkline.rs
joshka: updated everything to use is_key_press()

janTatesa:
Shouldn't it be generic over error instead of returning `Box<dyn Error>`? SUch as `fn run<F: FnOnce(&mut DefaultTerminal) -> Result<(), E>, E>(F) -> Result<(), E>`

joshka:
> Shouldn't it be generic over error instead of returning `Box<dyn Error>`? SUch as `fn run<F: FnOnce(&mut DefaultTerminal) -> Result<(), E>, E>(F) -> Result<(), E>`

It (`run()`) is generic over any return type whether that's an error or not an error.

```rust
pub fn run<F, R>(f: F) -> R
where
    F: FnOnce(&mut DefaultTerminal) -> R,
{
    let mut terminal = init();
    let result = f(&mut terminal);
    restore();
    result
}
```

PR #1728, opened 2025-03-19
style: use Module imports_granularity

joshka:
I was swayed by the arguments about this made by the compiler team In
<https://github.com/rust-lang/compiler-team/issues/750> and decided to
look at how this organization affects ratatui. I found this reduces the
number of lines across the codebase by about 350 and makes the imports
more readable and definitely more greppable as you usually only have
to read a single line. I've found in the past that maintaining imports
regularly leads to merge conflicts which have to be resolved by hand
and this change should reduce the likelihood of that happening.

Main change is in rustfmt.toml, and the rest is just the result of
running `cargo xtask format`.

While implementing this, cargo machete brings up that the various
backend crates are unused by the example crates.

The re-export of each backend crate under ratatui is to make it possible
for libs that rely on a specific version of ratatui to use the same
version of the backend crate. Apps in general should use the backend
crate directly rather than through ratatui as this is less confusing.

-    Removes all usages of `ratatui::{crossterm, termion, termwiz}`` in the
    examples.
-    Adds the backend crate to the dependencies of the examples that use
    the backend crate directly.


rustfmt.toml
```diff
@@ -1,9 +1,12 @@
...
+
+use_field_init_shorthand = true
+
+# unstable options
 comment_width = 100
 format_code_in_doc_comments = true
 format_macro_matchers = true
 group_imports = "StdExternalCrate"
-imports_granularity = "Crate"
+imports_granularity = "Module"
```
joshka, line 10: This is the only manual change

joshka:
fix: use backend specific imports for examples

The re-export of each backend crate under ratatui is to make it possible
for libs that rely on a specific version of ratatui to use the same
version of the backend crate. Apps in general should use the backend
crate directly rather than through ratatui as this is less confusing.

This fixes the cargo machete errors that find that the backend crate
is not used directly in many of the examples.

- Removes all usages of `ratatui::{crossterm, termion, termwiz}`` in the
  examples.
- Adds the backend crate to the dependencies of the examples that use
  the backend crate directly.

orhun (approved):
Nice, less code less bugs (not really)

joshka:
> Nice, less code less bugs (not really)

fewer merge conflicts, so if you class that as sort of a bug...

PR #1729, opened 2025-03-20
Very basic support of block border merging

pauladam94:
## Description
This is a feature that has been proposed for the next release of ratatui (0.30.0) [here](https://ratatui.rs/highlights/v029/).
> Future versions will enhance border drawing by combining borders to handle overlaps better.

## Example of the Result (Before Left - After Right)
<p align="center">
  
  
</p>

## Remark 
- The code is very basic, not optimal and should not be merged like that. However, it has the merit to be simple so it may break nothing. 
- I don't know of any idea to achieve it a better way. Indeed, there is a lot of `if then else` in my code. I have only tackled a few cases. Those could be replaced by macros, but this is not a good idea I think. I was thinking of a big match case but this is not possible to compare &str.
- Is it always wanted to have this behavior ? You can without this behavior simulate some depth and the user could understand which block is before another one.

### Functions implemented
The only thing implemented are the function :
```rust
pub fn merge_symbol(first: &str, second: &str) -> Option<&'static str>;
```
and the method for the `Cell` type :
```rust
    pub fn merge_symbol(&mut self, symbol: &str) -> &mut Self;
```
that is a drop-in replacement of the `set_symbol` method of `Cell`
```rust
    pub fn set_symbol(&mut self, symbol: &str) -> &mut Self;
```
### Line and Corner behavior
When the line is drawn for the border block, the last border character is also drawn. This poses a problem because it might merge with the corner in a wrong way.

There is 2 way to fix that :
- draw the line one pixel before and after to not merge with the corner (what I have done because it changes the less code)
- let the line merge and delete completely the call to draw the corner (it will merge into a corner by it self)

### `merge_symbol` returns an `Option`
 The merge function returns an Option if there is some merge to do or not.
 In a lot of cases, this will return None. It can be better for performance to have an early return for most of the cases. Especially if we call `merge_symbol` instead of `set_cell` all the time.

## Future Work
- Better way to handle cases
- handle all cases
- maybe use the `merge_symbol` instead for the `set_cell` function all the time

### DIFF
--- a/ratatui-core/src/buffer/cell.rs
+++ b/ratatui-core/src/buffer/cell.rs
@@ -1,6 +1,7 @@
 use compact_str::CompactString;
 
 use crate::style::{Color, Modifier, Style};
+use crate::symbols::merge::merge_symbol;
 
 /// A buffer cell
 #[derive(Debug, Clone, Eq, PartialEq, Hash)]
@@ -59,6 +60,19 @@ impl Cell {
     #[must_use]
     pub fn symbol(&self) -> &str {
         self.symbol.as_str()
+    }
+
+    /// Merge the symbol of the cell with the one already on the cell.
+    pub fn merge_symbol(&mut self, symbol: &str) -> &mut Self {
+        match merge_symbol(self.symbol.as_str(), symbol) {
+            None => {
+                self.symbol = CompactString::new(symbol);
+            }
+            Some(new_symbol) => {
+                self.symbol = CompactString::new(new_symbol);
+            }
+        }
+        self
     }
 
     /// Sets the symbol of the cell.
--- a/ratatui-core/src/symbols.rs
+++ b/ratatui-core/src/symbols.rs
@@ -9,5 +9,6 @@ pub mod braille;
 pub mod half_block;
 pub mod line;
 pub mod marker;
+pub mod merge;
 pub mod scrollbar;
 pub mod shade;
--- a/ratatui-core/src/symbols/merge.rs
+++ b/ratatui-core/src/symbols/merge.rs
@@ -1,0 +1,66 @@
+use super::line;
+
+/// Merge the two given symbols.
+/// Returns `None` if there is nothing to merge.
+/// Nothing to merge means you have to choose which one to display.
+///
+/// TODO : so much "if then else" are
+/// - bad pratice
+/// - bad performance
+/// - easy to achieve with macro (but not a good idea)
+/// - compiler optimized ?
+/// I didn't achieved it simply with a match case
+pub fn merge_symbol(first: &str, second: &str) -> Option<&'static str> {
+    if first == line::NORMAL.vertical {
+        if second == line::NORMAL.vertical {
+            None
+        } else if second == line::NORMAL.horizontal {
+            Some(line::NORMAL.cross)
+        } else if second == line::NORMAL.top_right {
+            Some(line::NORMAL.vertical_left)
+        } else if second == line::NORMAL.top_left {
+            Some(line::NORMAL.vertical_right)
+        } else if second == line::NORMAL.bottom_right {
+            Some(line::NORMAL.vertical_left)
+        } else if second == line::NORMAL.bottom_left {
+            Some(line::NORMAL.vertical_right)
+        } else if second == line::NORMAL.vertical_left {
+            None
+        } else if second == line::NORMAL.vertical_right {
+            None
+        } else if second == line::NORMAL.horizontal_down {
+            Some(line::NORMAL.cross)
+        } else if second == line::NORMAL.horizontal_up {
+            Some(line::NORMAL.cross)
+        } else {
+            None
+        }
+    } else if first == line::NORMAL.horizontal {
+        if second == line::NORMAL.vertical {
+            Some(line::NORMAL.cross)
+        } else if second == line::NORMAL.horizontal {
+            None
+        } else if second == line::NORMAL.top_right {
+            Some(line::NORMAL.horizontal_down)
+        } else if second == line::NORMAL.top_left {
+            Some(line::NORMAL.horizontal_down)
+        } else if second == line::NORMAL.bottom_right {
+            Some(line::NORMAL.horizontal_up)
+        } else if second == line::NORMAL.bottom_left {
+            Some(line::NORMAL.horizontal_up)
+        } else if second == line::NORMAL.vertical_left {
+            Some(line::NORMAL.cross)
+        } else if second == line::NORMAL.vertical_right {
+            Some(line::NORMAL.cross)
+        } else if second == line::NORMAL.horizontal_down {
+            None
+        } else if second == line::NORMAL.horizontal_up {
+            None
+        } else {
+            None
+        }
+    } else {
+        None
+    }
+    // ... other cases
+}
--- a/ratatui-widgets/src/block.rs
+++ b/ratatui-widgets/src/block.rs
@@ -647,9 +647,9 @@ impl Block<'_> {
 
     fn render_left_side(&self, area: Rect, buf: &mut Buffer) {
         if self.borders.contains(Borders::LEFT) {
-            for y in area.top()..area.bottom() {
+            for y in area.top() + 1..area.bottom() - 1 {
                 buf[(area.left(), y)]
-                    .set_symbol(self.border_set.vertical_left)
+                    .merge_symbol(self.border_set.vertical_left)
                     .set_style(self.border_style);
             }
         }
@@ -657,9 +657,9 @@ impl Block<'_> {
 
     fn render_top_side(&self, area: Rect, buf: &mut Buffer) {
         if self.borders.contains(Borders::TOP) {
-            for x in area.left()..area.right() {
+            for x in area.left() + 1..area.right() - 1 {
                 buf[(x, area.top())]
-                    .set_symbol(self.border_set.horizontal_top)
+                    .merge_symbol(self.border_set.horizontal_top)
                     .set_style(self.border_style);
             }
         }
@@ -668,9 +668,9 @@ impl Block<'_> {
     fn render_right_side(&self, area: Rect, buf: &mut Buffer) {
         if self.borders.contains(Borders::RIGHT) {
             let x = area.right() - 1;
-            for y in area.top()..area.bottom() {
+            for y in area.top() + 1..area.bottom() - 1 {
                 buf[(x, y)]
-                    .set_symbol(self.border_set.vertical_right)
+                    .merge_symbol(self.border_set.vertical_right)
                     .set_style(self.border_style);
             }
         }
@@ -679,9 +679,9 @@ impl Block<'_> {
     fn render_bottom_side(&self, area: Rect, buf: &mut Buffer) {
         if self.borders.contains(Borders::BOTTOM) {
             let y = area.bottom() - 1;
-            for x in area.left()..area.right() {
+            for x in area.left() + 1..area.right() - 1 {
                 buf[(x, y)]
-                    .set_symbol(self.border_set.horizontal_bottom)
+                    .merge_symbol(self.border_set.horizontal_bottom)
                     .set_style(self.border_style);
             }
         }
@@ -690,7 +690,7 @@ impl Block<'_> {
     fn render_bottom_right_corner(&self, buf: &mut Buffer, area: Rect) {
         if self.borders.contains(Borders::RIGHT | Borders::BOTTOM) {
             buf[(area.right() - 1, area.bottom() - 1)]
-                .set_symbol(self.border_set.bottom_right)
+                .merge_symbol(self.border_set.bottom_right)
                 .set_style(self.border_style);
         }
     }
@@ -698,7 +698,7 @@ impl Block<'_> {
     fn render_top_right_corner(&self, buf: &mut Buffer, area: Rect) {
         if self.borders.contains(Borders::RIGHT | Borders::TOP) {
             buf[(area.right() - 1, area.top())]
-                .set_symbol(self.border_set.top_right)
+                .merge_symbol(self.border_set.top_right)
                 .set_style(self.border_style);
         }
     }
@@ -706,7 +706,7 @@ impl Block<'_> {
     fn render_bottom_left_corner(&self, buf: &mut Buffer, area: Rect) {
         if self.borders.contains(Borders::LEFT | Borders::BOTTOM) {
             buf[(area.left(), area.bottom() - 1)]
-                .set_symbol(self.border_set.bottom_left)
+                .merge_symbol(self.border_set.bottom_left)
                 .set_style(self.border_style);
         }
     }
@@ -714,7 +714,7 @@ impl Block<'_> {
     fn render_top_left_corner(&self, buf: &mut Buffer, area: Rect) {
         if self.borders.contains(Borders::LEFT | Borders::TOP) {
             buf[(area.left(), area.top())]
-                .set_symbol(self.border_set.top_left)
+                .merge_symbol(self.border_set.top_left)
                 .set_style(self.border_style);
         }
     }


ratatui-widgets/src/block.rs
```diff
@@ -637,19 +637,19 @@ impl Block<'_> {
 
     fn render_left_side(&self, area: Rect, buf: &mut Buffer) {
         if self.borders.contains(Borders::LEFT) {
-            for y in area.top()..area.bottom() {
+            for y in area.top() + 1..area.bottom() - 1 {
```
pauladam94, line 640: Here the first and last draw calls are removed. Indeed they overlap with the border so the merging is done where there should be no merging.

ratatui-widgets/src/block.rs
```diff
@@ -637,19 +637,19 @@ impl Block<'_> {
...
                 buf[(area.left(), y)]
-                    .set_symbol(self.border_set.vertical_left)
+                    .merge_symbol(self.border_set.vertical_left)
```
pauladam94, line 642: Here I replaced `set_symbol` with the drop-in replacement `merge_symbol`. The same operation is done but if a border (or something else) has already drawnn something here, the previous and new character will be merged.

ratatui-core/src/symbols/merge.rs
```diff
@@ -0,0 +1,156 @@
+use crate::symbols::line::{DOUBLE, NORMAL, ROUNDED, THICK};
+use std::collections::HashMap;
+use std::sync::OnceLock;
+
+/// Map to know how to merge the two given symbols.
+/// If a couple of symbols are not is the map there is nothing to merge.
+/// This map should be symetric.
+static MERGE_MAP: OnceLock<HashMap<(&'static str, &'static str), &'static str>> = OnceLock::new();
+
+macro_rules! insert_merge_rules {
```
pauladam94, line 10: A macro because those merging rules are generic to any style (`s` in the code) of border.

ratatui-core/src/symbols/merge.rs
```diff
@@ -0,0 +1,156 @@
+use crate::symbols::line::{DOUBLE, NORMAL, ROUNDED, THICK};
+use std::collections::HashMap;
+use std::sync::OnceLock;
+
+/// Map to know how to merge the two given symbols.
+/// If a couple of symbols are not is the map there is nothing to merge.
+/// This map should be symetric.
+static MERGE_MAP: OnceLock<HashMap<(&'static str, &'static str), &'static str>> = OnceLock::new();
```
pauladam94, line 8: Here we use `OnceLock` which demands no other dependency. A more idiomatic way to do this would be to use `once_cell` or `lazy_static` but this adds a dependency.

The behavior wanted is that this hashmap should be initialize once at first call. It is a static hashmap so after that it is O(1) lookup time.

ratatui-core/src/symbols.rs
```diff
@@ -9,5 +9,6 @@ pub mod braille;
 pub mod half_block;
 pub mod line;
 pub mod marker;
+pub mod merge;
```
pauladam94, line 12: Because this code is about merging characters. I found that the right place was next to all the definition of specific unicode character used in ratatui.

ratatui-core/src/symbols/merge.rs
```diff
@@ -0,0 +1,156 @@
...
+        $map.insert(($s.cross, $s.vertical_left), $s.cross);
+        $map.insert(($s.cross, $s.vertical_right), $s.cross);
+        $map.insert(($s.cross, $s.horizontal_down), $s.cross);
+        $map.insert(($s.cross, $s.horizontal_up), $s.cross);
+        // $map.insert(($s.cross, $s.cross), $s.cross);
+    };
+}
+
+/// Lazy initialization of `MERGE_MAP`
+pub fn get_merge_map() -> &'static HashMap<(&'static str, &'static str), &'static str> {
```
pauladam94, line 147: This is not a free operation. This also allocates a pretty big hash map. This is a cost that we have to take into account, however the cost is mainly in startup time and in space. In terms of complexity, this is O(1) for a normal application.

ratatui-core/src/symbols/merge.rs
joshka: What about using [LazyLock](https://doc.rust-lang.org/std/sync/struct.LazyLock.html)? Introduced in 1.80 (we're fine to bump the MSRV up to at most 1.83 if necessary given stable is 1.85).

It might also be nice if users are able to modify this value so that their own sets use the combination characters. How would the design of this change if you considered that problem?

ratatui-core/src/symbols/merge.rs
joshka: I'm not sure I'm seeing the need for this to be a macro - isn't s always symbol::line::Set in the calling code below?

ratatui-widgets/src/block.rs
joshka: Add a comment about the +1 / -1. Seems easy that there would be some off by one errors that are difficult to guess about here.

ratatui-core/src/symbols/merge.rs
```diff
@@ -0,0 +1,156 @@
...
+    MERGE_MAP.get_or_init(|| {
+        let mut map = HashMap::new();
+        insert_merge_rules!(map, NORMAL);
+        insert_merge_rules!(map, ROUNDED);
+        insert_merge_rules!(map, THICK);
+        insert_merge_rules!(map, DOUBLE);
```
joshka, line 153: An oversight of missing characters from this is combination characters e.g. `┍` U+250D (Down light + right heavy). These combinations take this from pretty useful to must have. You can then make UIs where the focused block uses a thicker + colored border for instance. Note that the order of overlap likely matters here too (with any first drawn part of the line being overwritten).

joshka (commented):
There's another block border merging approach by @thscharler in https://github.com/ratatui/ratatui/issues/1468 which is worth collaborating with them on to work out what's best. I think the ideal solution is somewhere between this and that solution. (I like the simplicity of this solution, the other solution felt a bit complex to me but I haven't dug in to it in earnest to understand whether that's intrinsic or extrinsic complexity (i.e. caused by the problem or caused by the choice of solution). I suspect a bit of both.

ratatui-core/src/symbols/merge.rs
pauladam94: > What about using [LazyLock](https://doc.rust-lang.org/std/sync/struct.LazyLock.html)? Introduced in 1.80 (we're fine to bump the MSRV up to at most 1.83 if necessary given stable is 1.85).
> 
> It might also be nice if users are able to modify this value so that their own sets use the combination characters. How would the design of this change if you considered that problem?

Totally agree. I might use LazyLock. Dont know about specific thread safe difference. I used `OnceLock` for no specific reason.

ratatui-core/src/symbols/merge.rs
pauladam94: Definitely something interesting for this PR. Those criteria can be added to the hashmap... those features are clearly a step by step process adding the merge of symbols we want.

pauladam94:
> I'm not sure I'm seeing the need for this to be a macro - isn't s always symbol::line::Set in the calling code below?
I think this comes forced with the approach I have taken. The MERGE_MAP don't know where the symbols are coming from.

This can be a problem we cannot use as is the separation of the symbol::line::Set struct.

However this approach is very general and will be applicable easily to any widget drawing some line or special character.

pauladam94:
> There's another block border merging approach by @thscharler in https://github.com/ratatui/ratatui/issues/1468

This approach is actually specific to the block. (what I have seen of it quite quickly). Mine will even be useful for people writing symbol one by one to make their own widget, or for people drawing things with multiple lines. They would have to change the function from set_symbol to merge_symbol of course.

Cargo.toml
```diff
@@ -24,7 +24,7 @@ readme = "README.md"
 license = "MIT"
 exclude = ["assets/*", ".github", "Makefile.toml", "CONTRIBUTING.md", "*.log", "tags"]
 edition = "2021"
-rust-version = "1.74.0"
+rust-version = "1.80.0"
```
pauladam94, line 27: change of rust version because of the use of LazyLock.

pauladam94:
What breaks all the test is the +1/-1 change when drawing a line. 

Indeed, when the corner is not present before a line will still be drawn until the end.
This can happen for different reason like choosing to not one of the border of the block.

We get :
```
        " ───────Title┐ "
        "             │ "
        " ────────────┘ "
```
instead of (there is the first character missing at the left).
```
        "  ──────Title┐ "
        "             │ "
        "  ───────────┘ "

**Comment by pauladam94** [comment]
@joshka should I implement the solution of #1732 here since it is related to this PR ? Or create a new PR ?

File: ratatui-core/src/buffer/cell.rs [hunk 9]
```diff
@@ -61,6 +62,19 @@ impl Cell {
...
+        match MERGE_MAP.get(&(self.symbol.as_str(), symbol)) {
+            None => {
+                self.symbol = CompactString::new(symbol);
+            }
+            Some(new_symbol) => {
+                self.symbol = CompactString::new(new_symbol);
+            }
+        }
+        self
+    }
```
orhun, line 76:
```suggestion
    /// Merge the symbol of the cell with the one already on the cell.
    pub fn merge_symbol(&mut self, symbol: &str) -> &mut Self {
        match MERGE_MAP.get(&(self.symbol.as_str(), symbol)) {
            None => self.set_symbol(symbol),
            Some(new_symbol) => self.set_symbol(new_symbol),
        }
    }
```

orhun (commented):
Nice, I like the simplicity of this PR as well

ratatui-core/src/buffer/cell.rs
joshka: Option::unwrap_or() works too

ratatui-core/src/symbols/line.rs
```diff
@@ -58,6 +58,738 @@ pub const CROSS: &str = "┼";
 pub const DOUBLE_CROSS: &str = "╬";
 pub const THICK_CROSS: &str = "╋";
 
+pub enum LineStyle {
```
pauladam94, line 61: Some other type in the ratatui code base may be use instead of this one.

ratatui-core/src/symbols/merge.rs
```diff
@@ -0,0 +1,32 @@
+use crate::symbols::line::*;
+
+pub fn merge_border(prev: BorderSymbol, next: BorderSymbol) -> BorderSymbol {
```
pauladam94, line 3: simpler code for merging because of the data structure introduced.

ratatui-core/src/symbols/line.rs
orhun: Yup, this looks like [`BorderType`](https://docs.rs/ratatui/latest/ratatui/widgets/block/enum.BorderType.html)

ratatui-core/src/symbols/line.rs
```diff
@@ -58,6 +58,738 @@ pub const CROSS: &str = "┼";
...
+        }
+    }
+}
+
+impl TryFrom<&str> for BorderSymbol {
+    type Error = ();
+    fn try_from(value: &str) -> Result<Self, Self::Error> {
+        use LineStyle::*;
+        match value {
+            "╷" => Ok(BorderSymbol::new(None, None, None, Some(Normal))),
```
joshka: From a visual perspective, this fit on the screen better without wrapping as a tuple (alternatively this is a place to turn off automatic formatting).

I wonder if a compile time hashtable would be best here (e.g. phf crate)

You might be able to also add a none / empty vairant to the enum rather than using Option... Not sure about that though:

Regarding ordering these, I'd like to see the order arranged in the same order as the unicode symbol table.

I suspect that there's a way (perhaps with a small amount of macro magic) to write the table once and generate both the forward and backward table)

joshka (commented):
An idea for a declarative macro approach:

```rust
enum Weight {
    None,
    Light,
    Heavy,
    Double,
    TripleDash,
    QuadrupleDash,
    HeavyTripleDash,
    HeavyQuadrupleDash,
}

macro_rules! define_symbols {
    /// TBD - copilot is getting pretty good at generating these sorts of things from an example of the output btw.
}

define_symbols! {
    // symbol => (left, right, up, down) - not sure about this order, possible also match the
    // unicode naming order which is bottom/top, left/right, also perhaps splitting out the
    // horizontal and vertical weights would work too
    "─" => (Lignt, Light, None, None),
    "━" => (Heavy, Heavy, None, None),
    "│" => (None, None, Light, Light),
    "┃" => (None, None, Heavy, Heavy),
    "┄" => (TripleDash, TripleDash, None, None),
    "┅" => (HeavyTripleDash, HeavyTripleDash, None, None),
    "┆" => (None, None, TripleDash, TripleDash),
    "┇" => (None, None, HeavyTripleDash, HeavyTripleDash),
    "┈" => (QuadrupleDash, QuadrupleDash, None, None),
    "┉" => (HeavyQuadrupleDash, HeavyQuadrupleDash, None, None),
    "┊" => (None, None, QuadrupleDash, QuadrupleDash),
    "┋" => (None, None, HeavyQuadrupleDash, HeavyQuadrupleDash),
    "┌" => (None, Light, None, Light),
    "┍" => (None, Heavy, None, Heavy),
    ...
}
```

Should generate whatever hashtable / match statement / combination is needed

ratatui-core/src/symbols/line.rs
pauladam94: - Seems legit to turn off auto formatting.
- Compile time hashtable could be the way, I was trying to avoid that in favor of functions that may be const someday.
- Ordering is an issue indeed. So ordered croissant from there unicode value ?
- yes some macro may be able to do the backward table for sure

joshka: croissant? 🥐

pauladam94:
> (left, right, up, down) - not sure about this order
I have choose the right - up - left - down order because of math. This is a traditional way when describing increasing angle in radian.

joshka:
> > (left, right, up, down) - not sure about this order
> 
> I have chosen the right - up - left - down order because of math. This is a traditional way when describing increasing angle in radian.

got it - that is definitely one way of ordering. There's too many obvious ways to order. I'm aiming here for one which makes most sense with respect to the characters.

E.g.
- Css (top, right, bottom, left)
- Unicode box drawing order (bottom, top, right, left) - I suspect that if the symbols are ordered as they appear in the table, then this will appear fairly naturally as the right order.
- Natural terminal / reading order (top, bottom, left, right)

Mostly this affects how the maintenance of checking that a symbol exists and is catered for correctly

Also, regarding None, the macro approach might map None onto Option::None while the others onto Some(...) saving a bit of horizontal space.

ratatui-core/src/symbols/line.rs
pauladam94: "croissant" means "increasing" in french .. 🫣 (yes this has the spelling and pronunciation as the pastry)

joshka: Ha, ok - I thought it was a typo, but couldn't work out what for...
Je n'ai qu'un peu de français

pauladam94:
Yep I am looking into doing a nice macro for that. Completely agree that we want a ordering that makes sense with the character ordering in unicode and having the up-left... in the same order it was intended in unicode characters.

My reference is of course [wikipedia](https://en.wikipedia.org/wiki/Box-drawing_characters) especially this picture :
(the image has a dark background so it might no be great to look at in default github settings)
image

ratatui-core/src/symbols/line.rs
```diff
@@ -58,6 +58,212 @@ pub const CROSS: &str = "┼";
...
+}
+
+pub struct BorderSymbol {
+    pub right: LineStyle,
+    pub up: LineStyle,
+    pub left: LineStyle,
+    pub down: LineStyle,
+}
+
+macro_rules! define_symbols {
```
pauladam94, line 81: Macro that define the back and forth translation between `BorderSymbol` and the character that corresponds to it.

pauladam94:
This is not supported by the model I have presented here :
image

Indeed, the merging function sometimes returns a border structure for which we have no character that represents it.

pauladam94:
> It's sort of a shame that all the combinations of box drawing characters aren't part of the unicode standard (e.g. double up / down / left / right, double rounded, ...

Because of that we get some behavior like that (because the combination : top normal, left double, down double don't exists !) :
image

joshka:
> > It's sort of a shame that all the combinations of box drawing characters aren't part of the unicode standard (e.g. double up / down / left / right, double rounded, ...
> 
> Because of that we get some behavior like that (because the combination : top normal, left double, down double don't exists !) :
> 
>  

One thing you might consider there is whether `╦` and `╣` might be better than not linking the boxes (or perhaps making the outcome there configurable?)

One thing that would be useful to see is tests that show how each of the borders interact visually. I wonder what the minimal picture for that might look like that shows each interaction for the various combinations?

pauladam94:
> making the outcome there configurable

I think there should be a parameter like that somewhere :
```Rust
enum MergeLevel {
    Basic, // just merge when a "perfect" character works for the merge
    Extensive // try merging into the closest character that works
}

// ...
    level: Option<MergeLevel>, // None being the actual behavior that don't do anything
// ...
```

This could be a parameter of each Widget (Block and Line for example) to know if they will call `set_symbol` or `merge_symbol`.

joshka:
> This could be a parameter of each Widget (Block and Line for example) to know if they will call `set_symbol` or `merge_symbol`.

It likely only needs to apply to the Block widget for now. This likely should have a variant which controls that the symbols are not merged. e.g.:
```rust
#[derive(Default)
enum MergeBorders {
    #[default]
    Disabled,
    // Only merge symbols where there's an exact match
    Exact,
    // Merge symbols that ... (wording for how to express this would be difficult)
    BestFit,
}
```

> Also, a lot of test don't work because of the behavior when removing a border of a block. What do you think about those tests ? Should we change the test because it has a different behavior, or make the merging differently to pass those tests ?

If you're referring to existing tests being broken, then this indicates places where users of Ratatui will see their apps change their display behavior. This will not be desirable in many cases. Setting MergeBorders::Disabled as the default is the key to making this backwards compatible and opt-in.

> I agree, a large part of this PR should include some additional test.

There's an arrangement that would visually test most of the cross over parts in a small amount of diagrams:
- a plain bordered block with a overlapping thick blocks at each corner (tests that two corners become a cross)
- a plain bordered block where the overlapping blocks are shifted 2 characters inside in width (tests vertical T intersections and overlapping horizontal lines)
- a plain bordered block where the overlapping blocks are shifted 2 characters inside in height (tests horizontal T intersections and overlapping vertical lines)
- a plain bordered block where the overlapping blocks are shifted inside for both width and height (tests horizontal and vertical crossing lines)

I prefer the visual rendering tests for this sort of thing over more character based tests as it makes it easy understand that the tests are correct.

pauladam94:
There are still 3 three things to do :
- have some sort of "best fit" algorithm to find merged when they're no "pixel perfect" one
- arrange in the block widget code the parameter and how to toggle it
- add 2 good test for this feature. One that is clear and comes up with the most cases. 2 to have one for the "Exact" and the "Best Fit" part. This will be the same code (except the MergeBorders option) but with different output to show the cases where Exact mght not find "good" characters.

pauladam94:
> got it - that is definitely one way of ordering. There's too many obvious ways to order. I'm aiming here for one which makes most sense with respect to the characters.

For now I didn't change the order of the (left, right, up, down) issue thing ... I don't mind changing it with a simple regexp in the macro call. Is there somewhere in the ratatui code base where such a question already happened ? Maybe we can choose that to stay consistent. It is not that important and will be able to change easily in the future.

joshka:
> > got it - that is definitely one way of ordering. There's too many obvious ways to order. I'm aiming here for one which makes most sense with respect to the characters.
> 
> For now I didn't change the order of the (left, right, up, down) issue thing ... I don't mind changing it with a simple regexp in the macro call. Is there somewhere in the ratatui code base where such a question already happened ? Maybe we can choose that to stay consistent. It is not that important and will be able to change easily in the future.

I can't think of any. Mostly I was curious whether there's an order that makes it intentionally show some sort of obvious relationship and order to the combinations (e.g. how 0001, 0010, 0010, 0011 counts)

pauladam94:
> obvious relationship and order to the combinations
Yeah, I thought like that too. But there is not such a thing, they have ordered it for some sort of groups of characters and no so much more things. There doesn't seem to be any binary underlying logic that we could use, unfortunately.

pauladam94:
This PR is near the finish line.

Here is the test I have come up with.
Left is for `None` (default and past behavior) | Middle if for `Some(MergeStyle::Exact)` | Right is for `Some(MergeStyle::BestFit)`

image

joshka:
Hey, it may be a 2+ weeks before I get back to looking at this one. Ping me if you haven't heard back around the end of April.

pauladam94:
@joshka I think it's the moment to ping you.
I don't have that much time right now for this PR, I think it only needs some documentation comment for each new function.
I made it so that nothing breaks, 3 more tests and a new parameter for the `Block` struct.

joshka:
Thanks for the reminder on this. I'll take a look tomorrow for you.

ratatui-widgets/src/block.rs
```diff
@@ -1647,4 +1666,76 @@ mod tests {
...
+            .border_type(BorderType::Rounded)
+            .merge_style(merge_style.cloned())
+            .render(Rect::new(3, 0, 3, 2), buffer);
+        Block::bordered()
+            .merge_style(merge_style.cloned())
+            .render(buffer.area, buffer);
+    }
+
+    #[test]
+    fn render_non_merging_blocks() {
```
orhun, line 1692: Are these t3 ests sufficient to cover all the cases? Maybe use `rstest` and add some different test variants to each one.

orhun (commented):
This evolved nicely since last time I checked! Really pleasant work.

I left some QoL comments - looks pretty good to me.

j-g00da:
> I don't have that much time right now for this PR

Hey @pauladam94, let me know if you want a hand wrapping this up - I really like this feature!

pauladam94:
> Hey @pauladam94, let me know if you want a hand wrapping this up - I really like this feature!

Totally, I don't that much time for it right now (student exams sooon).
I definitely would take your help @j-g00da with this PR.

To wrap it up completely :
- documentation of each new structure (like @orhun pointed out)
- perhaps one more test to handle more cases
- every comment of @orhun is right and should be addressed

ratatui-core/src/symbols/line.rs
```diff
@@ -58,6 +58,270 @@ pub const CROSS: &str = "┼";
...
+    Nothing,
+    Plain,
+    Rounded,
+    Double,
+    Thick,
+    DoubleDash,
+    TripleDash,
+    TripleDashThick,
+    QuadrupleDash,
+    QuadrupleDashThick,
```
j-g00da, line 72:
```suggestion
    /// Represents the absence of a line.
    /// Example: (No visible line)
    Nothing,
    /// A single line.
    /// Example: ─ │
    Plain,
    /// A rounded line style, only applicable in corner symbols.
    /// Example: ╭ ╮ ╰ ╯
    Rounded,
    /// A double line.
    /// Example: ═ ║
    Double,
    /// A thickened line.
    /// Example: ━ ┃
    Thick,
    /// A dashed line with a double dash pattern.
    /// Example: ╌ ╎
    DoubleDash,
    /// A dashed line with a triple dash pattern.
    /// Example: ┄ ┆
    TripleDash,
    /// A thicker variant of the triple dash.
    /// Example: ┅ ┇
    TripleDashThick,
    /// A dashed line with four dashes.
    /// Example: ┈ ┊
    QuadrupleDash,
    /// A thicker variant of the quadruple dash.
    /// Example: ┉ ┋
    QuadrupleDashThick,
```

ratatui-core/src/symbols/line.rs
```diff
@@ -58,6 +58,270 @@ pub const CROSS: &str = "┼";
...
+    fn replace(self, from: &Self, to: &Self) -> Self {
+        if self == *from {
+            to.clone()
+        } else {
+            self
+        }
+    }
+}
+
+pub struct BorderSymbol {
```
j-g00da, line 85:
```suggestion
/// Represents a border symbol made of individual line components.
pub struct BorderSymbol {
```

ratatui-core/src/symbols/line.rs
```diff
@@ -58,6 +58,270 @@ pub const CROSS: &str = "┼";
...
+    pub fn replace(mut self, from: &LineStyle, to: &LineStyle) -> Self {
+        self.up = self.up.replace(from, to);
+        self.right = self.right.replace(from, to);
+        self.down = self.down.replace(from, to);
+        self.left = self.left.replace(from, to);
+        self
+    }
+}
+
+macro_rules! define_symbols {
```
j-g00da, line 138:
```suggestion
// Defines a translation between `BorderSymbol` and the corresponding character.
macro_rules! define_symbols {
```

ratatui-core/src/symbols/line.rs
```diff
@@ -58,6 +58,270 @@ pub const CROSS: &str = "┼";
...
+pub struct BorderSymbol {
+    pub right: LineStyle,
+    pub up: LineStyle,
+    pub left: LineStyle,
+    pub down: LineStyle,
+}
+
+impl BorderSymbol {
+    /// Find the best representation of the `BorderSymbol`
+    /// that will be able to be transform into a character
```
j-g00da, line 94:
```suggestion
    /// Finds the closest representation of the `BorderSymbol`, that has a corresponding unicode
    /// character.
```

ratatui-core/src/symbols/line.rs
```diff
@@ -58,6 +58,270 @@ pub const CROSS: &str = "┼";
...
+        }
+        // Rouned border character are only available for corners.
+        if self.contains(&Rounded) {
+            self = self.replace(&Rounded, &Plain);
+        }
+
+        self
+    }
+
+    pub fn contains(&self, style: &LineStyle) -> bool {
```
j-g00da, line 124:
```suggestion
    /// Checks if any of the line components making the `BorderSymbol` matches the `style`.
    pub fn contains(&self, style: &LineStyle) -> bool {
```

ratatui-core/src/symbols/line.rs
```diff
@@ -58,6 +58,270 @@ pub const CROSS: &str = "┼";
...
+    #[must_use]
+    pub const fn new(right: LineStyle, up: LineStyle, left: LineStyle, down: LineStyle) -> Self {
+        Self {
+            right,
+            up,
+            left,
+            down,
+        }
+    }
+}
```
j-g00da, line 323: Moved to the other impl block.
```suggestion
```

ratatui-core/src/symbols/line.rs
```diff
@@ -58,6 +58,270 @@ pub const CROSS: &str = "┼";
...
+
+pub struct BorderSymbol {
+    pub right: LineStyle,
+    pub up: LineStyle,
+    pub left: LineStyle,
+    pub down: LineStyle,
+}
+
+impl BorderSymbol {
```
j-g00da, line 92: Move `new` from the other impl block, add comments.

```suggestion
impl BorderSymbol {
    /// Creates a new `BorderSymbol`, based on individual line styles.
    #[must_use]
    pub const fn new(right: LineStyle, up: LineStyle, left: LineStyle, down: LineStyle) -> Self {
        Self {
            right,
            up,
            left,
            down,
        }
    }
```

ratatui-core/src/symbols/merge.rs
```diff
@@ -0,0 +1,42 @@
...
+
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Default)]
+pub enum MergeStyle {
+    #[default]
+    Exact,
+    // Merge symbols in a visual pleasing way.
+    // An example is that " " and " " will be merged to : ""
+    BestFit,
+}
+
```
j-g00da, line 11:
```suggestion
/// Defines the merge strategy of overlaping characters.
#[derive(Debug, Clone, Eq, PartialEq, Hash, Default)]
pub enum MergeStyle {
    /// Merges symbols only if an exact composite unicode character exists.
    /// Example: `┐` and `┗` will be merged into `╄`
    #[default]
    Exact,
    /// Merges symbols even if an exact composite unicode character doesn't exist,
    /// using the closest match.
    /// Example: `╮` and `└` will be merged into `┼`
    BestFit,
}

```

ratatui-widgets/src/block.rs
```diff
@@ -530,6 +534,14 @@ impl<'a> Block<'a> {
         self
     }
 
+    /// TODO documentation
+    /// TODO small example with two block merging / not merging
```
j-g00da, line 538: I think it doesn't need an example, as it links to MergeStyle, which has examples.
```suggestion
    /// Sets the block's [`MergeStyle`] for overlapping characters. Setting it to `None`
    /// never merges characters.
```

pauladam94:
@j-g00da thanks a lot ! All the comments have been merged. I now ping @joshka for the end, maybe.
Remark: typos is failing, but for a typo not introduced by this PR on the changelog.

joshka:
> Remark: typos is failing, but for a typo not introduced by this PR on the changelog.

Ha, sounds like we should do one or more of:
- run typos on commit messages
- not run typos on the changelog

ratatui-core/src/symbols/merge.rs
```diff
@@ -0,0 +1,46 @@
...
+        merge_line_style(&prev.left, &next.left),
+        merge_line_style(&prev.down, &next.down),
+    );
+    match style {
+        MergeStyle::BestFit => exact_result.best_fit(),
+        MergeStyle::Exact => exact_result,
+    }
+}
+
+pub fn merge_line_style(prev: &LineStyle, next: &LineStyle) -> LineStyle {
```
orhun, line 29:
```suggestion
/// Merges two line styles into one.
pub fn merge_line_style(prev: &LineStyle, next: &LineStyle) -> LineStyle {
```

ratatui-core/src/symbols/merge.rs
```diff
@@ -0,0 +1,46 @@
...
+    /// Example: `┐` and `┗` will be merged into `╄`
+    #[default]
+    Exact,
+    /// Merges symbols even if an exact composite unicode character doesn't exist,
+    /// using the closest match.
+    /// Example: `╮` and `└` will be merged into `┼`
+    BestFit,
+}
+
+pub fn merge_border(prev: &BorderSymbol, next: &BorderSymbol, style: &MergeStyle) -> BorderSymbol {
```
orhun, line 16:
```suggestion
/// Merges two border symbols into one.
pub fn merge_border(prev: &BorderSymbol, next: &BorderSymbol, style: &MergeStyle) -> BorderSymbol {
```

ratatui-widgets/src/block.rs
```diff
@@ -126,6 +127,8 @@ pub struct Block<'a> {
     style: Style,
     /// Block padding
     padding: Padding,
+    /// Merging Style
```
orhun, line 130: nit:
```suggestion
    /// Merging style
```

orhun (commented):
We're pretty close with this... just needs some docs tweaks + documenting `LineStyle`

orhun:
yup, more tests == more better == more rat

ratatui-core/src/symbols/line.rs
```diff
@@ -58,6 +58,270 @@ pub const CROSS: &str = "┼";
 pub const DOUBLE_CROSS: &str = "╬";
 pub const THICK_CROSS: &str = "╋";
 
+#[derive(PartialEq, Clone)]
+pub enum LineStyle {
```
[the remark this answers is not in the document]
j-g00da: Hey, I think you closed this by mistake, can you apply this suggestion?

ratatui-core/src/symbols/line.rs
j-g00da: Same with this one

j-g00da:
Can we merge this?

ratatui-core/src/symbols/merge.rs
```diff
@@ -0,0 +1,48 @@
+use crate::symbols::line::{BorderSymbol, LineStyle};
+
+/// Defines the merge strategy of overlapping characters.
+#[derive(Debug, Clone, Eq, PartialEq, Hash, Default)]
+pub enum MergeStyle {
+    /// Merges symbols only if an exact composite unicode character exists.
+    /// Example: `┐` and `┗` will be merged into `╄`
+    #[default]
+    Exact,
```
joshka, line 9: Comment should cover a bit about why this is needed in a narrative form (e.g. There are some symbol combinations that are missing such as rounded corners combining with straight lines and thick lines with double lines.

I'd think in general that BestFit would be a better default, but given that I'm suggesting that Overwrite should be a variant, that makes more sense rather than having a two tier default (None / Some(Exact))

ratatui-widgets/src/block.rs
```diff
@@ -530,6 +534,14 @@ impl<'a> Block<'a> {
         self
     }
 
+    /// Sets the block's [`MergeStyle`] for overlapping characters. Setting it to `None`
+    /// never merges characters.
+    #[must_use = "method moves the value of self and returns the modified value"]
+    pub const fn merge_style(mut self, merge_style: Option<MergeStyle>) -> Self {
```
joshka, line 540: This should talk about how this is related to the border and perhaps be named merge_borders or something similar. It definitely should not have style in the name.

joshka (changes requested):
Apologies for the lag in getting to this review. I wanted to give it some deep review which required really digging in and understanding it in context rather than a shallower one that focuses on does it work etc. Because this affects things in `ratatui-core`, we want to try to get this as close to perfect as possible so we can avoid needing to bump `-core` soon after release.

As an over all implementation, I think the idea of making this part of Cell is a great idea. I think it works better over the alternative which only implemented this for Block.

I think MergeStyle (MergeAction) would be clearer with an Overwrite variant rather than using Option to represent this. This is the main blocking comment I have. It should also be non_exhaustive I think. This will make it possible to allow users to add their own merge functionality in the future (e.g. to handle custom border sets).

I think I'd like to start out making LineStyle and BorderSymbol private rather than having them in the public API. Doing so makes the merge functionality defined by the implementation rather than being constrained to follow / align with types which can't be changed without bumping `ratatui-core` in the future.

By making those two changes, we get this to a point where the internal details / nits etc. are not breaking with respect to semver and this PR becomes not too problematic to merge soon.

--

In particular, I think at some point in the future it makes sense to have a BoxDrawingCharacter type (or something similar), which represents the Unicode metadata about these symbols nicely (and which helps this implementation).

An alternative approach over using LineStyle is to have a `(symbol, symbol) => result` function / hashmap. As a table, this would be 16kB (128 chars in the block ^ 2), so that's not particularly large (not enough to worry about at least).

Broadly applicable comment:
- leave a newline before doc comments on enum variants and functions. This makes the comment more easy to align with the code that it's commenting (I don't think there's a rustfmt option for this, but I'd turn it on if there was).


There's a few things which seem like they're oddly placed:
- functions which seem like they're good candidates to be methods
- merge implementation in Cell rather than as a method MergeStyle

On the testing side of this, I think it makes sense to have some basic symbol, symbol => symbol type tests of the merging in addition to the checks on block.

For the block tests, I can visualize some simpler block drawings that might help test the block combinations a bit better too, but they're harder to describe than implement.
- A block, surrounded on each corner by 4 blocks (or partial blocks to save space). I think this works out to 6x6 in terms of necessary rendered space.
- 4 combinations of the above (I think this would be renderable in 24x6)
  - touching corners exactly
  - touching corner and vertical sides
  - touching corner and horizontal sides
  - crossing vertical and horizontal sides
- for each of the combinations between (plain, thick, double, rounded) add a line (16 options, so this becomes 24x96) or go across then down (96x24) could also work pretty neatly and fit on a screen and be fairly easy to look at and see that it's correct
I'll slap together a quick demo of what I mean here shortly.

@j-g00da if you've got a bunch of motivation to continue this PR, you're welcome to. If you want to add your own commits on top and have them run through as a normal PR, feel free to clone this, add your commits and make a new PR. We'll credit the eventual single commit to @pauladam94, and add a Co author on it.

Otherwise, there's a bunch of things I've mentioned above that could make sense for me to just do that I can update and push to Paul's branch. Let me know how you'd like to proceed?

j-g00da:
@joshka I can get back to this on the weekend

j-g00da:
I've opened a new draft PR: #1874, @joshka can you apply your code suggestions here, so I can rebase it on my branch?

ratatui-core/src/buffer/cell.rs
```diff
@@ -61,6 +63,24 @@ impl Cell {
         self.symbol.as_str()
     }
 
+    /// Merge the symbol of the cell with the one already on the cell.
+    pub fn merge_symbol(&mut self, symbol: &'static str, style: Option<&MergeStyle>) -> &mut Self {
```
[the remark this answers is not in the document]
j-g00da: Resolved in #1874

ratatui-core/src/symbols/merge.rs
```diff
@@ -0,0 +1,48 @@
...
+        MergeStyle::BestFit => exact_result.best_fit(),
+        MergeStyle::Exact => exact_result,
+    }
+}
+
+/// Merges two line styles into one.
+pub fn merge_line_style(prev: &LineStyle, next: &LineStyle) -> LineStyle {
+    use LineStyle::{
+        DoubleDash, Nothing, Plain, QuadrupleDash, QuadrupleDashThick, TripleDash, TripleDashThick,
+    };
```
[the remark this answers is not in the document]
j-g00da: Resolved in #1874

ratatui-core/src/symbols/merge.rs
```diff
@@ -0,0 +1,48 @@
...
+        merge_line_style(&prev.down, &next.down),
+    );
+    match style {
+        MergeStyle::BestFit => exact_result.best_fit(),
+        MergeStyle::Exact => exact_result,
+    }
+}
+
+/// Merges two line styles into one.
+pub fn merge_line_style(prev: &LineStyle, next: &LineStyle) -> LineStyle {
```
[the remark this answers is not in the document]
j-g00da: Resolved in #1874

ratatui-core/src/symbols/merge.rs
```diff
@@ -0,0 +1,48 @@
...
+pub enum MergeStyle {
+    /// Merges symbols only if an exact composite unicode character exists.
+    /// Example: `┐` and `┗` will be merged into `╄`
+    #[default]
+    Exact,
+    /// Merges symbols even if an exact composite unicode character doesn't exist,
+    /// using the closest match.
+    /// Example: `╮` and `└` will be merged into `┼`
+    BestFit,
+}
```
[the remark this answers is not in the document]
joshka: Strategy is a much better choice I think.

orhun:
Implemented in #1874

PR #1735, opened 2025-03-24
feat!: rename Alignment to HorizontalAlignment and add VerticalAlignment

joshka:
We don't anticipate removing or deprecating the type alias in the near
future, but it is recommended to update your imports to use the new
name.

Added a VerticalAlignment enum to make the API more consistent. We don't
have a specific use case for it yet, but it's better to add it now and
be able to use it in the future.

BREAKING-CHANGE: The `Alignment` enum has been renamed to
`HorizontalAlignment` to better reflect its purpose. A type alias has
been added to maintain backwards compatibility, however there are some
cases where type aliases are not enough to maintain backwards
compatibility. E.g. when using glob imports to import all the enum
variants. This should not affect most users, but it is recommended to
update your imports to use the new name.

```diff
- use ratatui::layout::Alignment;
+ use ratatui::layout::HorizontalAlignment;

- use Alignment::*;
+ use HorizontalAlignment::*;
```

### DIFF
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -14,6 +14,7 @@ This is a quick summary of the sections below:
   - The `From` impls for backend types are now replaced with more specific traits
   - `FrameExt` trait for `unstable-widget-ref` feature
   - `List::highlight_symbol` now accepts `Into<Line>` instead of `&str`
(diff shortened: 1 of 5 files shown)


joshka:
Feel free to merge this if approving.

orhun (approved):

PR #1741, opened 2025-03-31
feat: Add elm architecture example

Tuurlijk:
A list-detail app that uses the Elm architecture

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2551,6 +2551,7 @@ dependencies = [
  "cassowary",
  "compact_str",
  "document-features",
+ "hashbrown",
  "indoc",
  "itertools 0.13.0",
  "lru",
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -36,7 +36,7 @@ actions](.github/workflows/cd.yml) and triggered by pushing a tag.
 ## Alpha Releases
 
 Alpha releases are automatically released every Saturday via [cd.yml](./.github/workflows/cd.yml)
-and can be manually be created when necessary by triggering the [Continuous
+and can be manually created when necessary by triggering the [Continuous
 Deployment](https://github.com/ratatui/ratatui/actions/workflows/cd.yml) workflow.
 
 We automatically release an alpha release with a patch level bump + alpha.num weekly (and when we
--- a/examples/apps/colors-rgb/src/main.rs
+++ b/examples/apps/colors-rgb/src/main.rs
@@ -115,7 +115,7 @@ impl App {
             if let Event::Key(key) = event::read()? {
                 if key.kind == KeyEventKind::Press && key.code == KeyCode::Char('q') {
                     self.state = AppState::Quit;
-                };
+                }
             }
         }
         Ok(())
--- a/examples/apps/constraint-explorer/src/main.rs
+++ b/examples/apps/constraint-explorer/src/main.rs
@@ -142,7 +142,7 @@ impl App {
             | Constraint::Fill(v)
             | Constraint::Percentage(v) => *v = v.saturating_add(1),
             Constraint::Ratio(_n, d) => *d = d.saturating_add(1),
-        };
+        }
     }
 
     fn decrement_value(&mut self) {
@@ -156,7 +156,7 @@ impl App {
             | Constraint::Fill(v)
             | Constraint::Percentage(v) => *v = v.saturating_sub(1),
             Constraint::Ratio(_n, d) => *d = d.saturating_sub(1),
-        };
+        }
     }
 
     /// select the next block with wrap around
@@ -466,7 +466,7 @@ impl ConstraintBlock {
             } else {
                 main_color
             };
-            if let Some(last_row) = area.rows().last() {
+            if let Some(last_row) = area.rows().next_back() {
                 buf.set_style(last_row, border_color);
             }
         }
--- a/examples/apps/inline/src/main.rs
+++ b/examples/apps/inline/src/main.rs
@@ -107,7 +107,7 @@ fn input_handling(tx: mpsc::Sender<Event>) {
                     event::Event::Key(key) => tx.send(Event::Input(key)).unwrap(),
                     event::Event::Resize(_, _) => tx.send(Event::Resize).unwrap(),
                     _ => {}
-                };
+                }
             }
             if last_tick.elapsed() >= tick_rate {
                 tx.send(Event::Tick).unwrap();
@@ -212,9 +212,9 @@ fn run(
                             break;
                         }
                     }
-                };
+                }
             }
-        };
+        }
     }
     Ok(())
 }
--- a/examples/apps/input-form/src/main.rs
+++ b/examples/apps/input-form/src/main.rs
@@ -236,7 +236,7 @@ impl AgeField {
             KeyCode::Up | KeyCode::Char('k') => self.increment(),
             KeyCode::Down | KeyCode::Char('j') => self.decrement(),
             _ => {}
-        };
+        }
     }
 
     fn increment(&mut self) {
--- a/examples/apps/table/src/main.rs
+++ b/examples/apps/table/src/main.rs
@@ -80,14 +80,20 @@ impl Data {
         [&self.name, &self.address, &self.email]
     }
 
+    // https://github.com/rust-lang/rust/issues/139338
+    #[allow(clippy::missing_const_for_fn)]
     fn name(&self) -> &str {
         &self.name
     }
 
+    // https://github.com/rust-lang/rust/issues/139338
+    #[allow(clippy::missing_const_for_fn)]
     fn address(&self) -> &str {
         &self.address
     }
 
+    // https://github.com/rust-lang/rust/issues/139338
+    #[allow(clippy::missing_const_for_fn)]
     fn email(&self) -> &str {
         &self.email
     }
--- a/examples/apps/todo-list/src/main.rs
+++ b/examples/apps/todo-list/src/main.rs
@@ -105,7 +105,7 @@ impl App {
             terminal.draw(|frame| frame.render_widget(&mut self, frame.area()))?;
             if let Event::Key(key) = event::read()? {
                 self.handle_key(key);
-            };
+            }
         }
         Ok(())
     }
--- a/ratatui-core/Cargo.toml
+++ b/ratatui-core/Cargo.toml
@@ -48,6 +48,7 @@ bitflags = "2.3"
 cassowary = "0.3"
 compact_str = "0.8.0"
 document-features = { workspace = true, optional = true }
+hashbrown = "0.15.2"
 indoc.workspace = true
 itertools.workspace = true
 lru = "0.12.0"
--- a/ratatui-core/src/backend.rs
+++ b/ratatui-core/src/backend.rs
@@ -100,6 +100,7 @@
 //! [Examples]: https://github.com/ratatui/ratatui/tree/main/ratatui/examples/README.md
 //! [Backend Comparison]: https://ratatui.rs/concepts/backends/comparison/
 //! [Ratatui Website]: https://ratatui.rs
+use core::ops;
 use std::io;
 
 use strum::{Display, EnumString};
@@ -341,8 +342,7 @@ pub trait Backend {
     /// For examples of how this function is expected to work, refer to the tests for
     /// [`TestBackend::scroll_region_up`].
     #[cfg(feature = "scrolling-regions")]
-    fn scroll_region_up(&mut self, region: std::ops::Range<u16>, line_count: u16)
-        -> io::Result<()>;
+    fn scroll_region_up(&mut self, region: ops::Range<u16>, line_count: u16) -> io::Result<()>;
 
     /// Scroll a region of the screen downwards, where a region is specified by a (half-open) range
     /// of rows.
@@ -363,11 +363,7 @@ pub trait Backend {
     /// For examples of how this function is expected to work, refer to the tests for
     /// [`TestBackend::scroll_region_down`].
     #[cfg(feature = "scrolling-regions")]
-    fn scroll_region_down(
-        &mut self,
-        region: std::ops::Range<u16>,
-        line_count: u16,
-    ) -> io::Result<()>;
+    fn scroll_region_down(&mut self, region: ops::Range<u16>, line_count: u16) -> io::Result<()>;
 }
 
 #[cfg(test)]
--- a/ratatui-core/src/backend/test.rs
+++ b/ratatui-core/src/backend/test.rs
@@ -1,8 +1,9 @@
 //! This module provides the `TestBackend` implementation for the [`Backend`] trait.
 //! It is used in the integration tests to verify the correctness of the library.
 
-use std::fmt::{self, Write};
-use std::{io, iter};
+use core::fmt::{self, Write};
+use core::{iter, ops};
+use std::io;
 
 use unicode_width::UnicodeWidthStr;
 
@@ -54,7 +55,7 @@ fn buffer_view(buffer: &Buffer) -> String {
             } else {
                 overwritten.push((x, c.symbol()));
             }
-            skip = std::cmp::max(skip, c.symbol().width()).saturating_sub(1);
+            skip = core::cmp::max(skip, c.symbol().width()).saturating_sub(1);
         }
         view.push('"');
         if !overwritten.is_empty() {
@@ -363,7 +364,7 @@ impl Backend for TestBackend {
     }
 
     #[cfg(feature = "scrolling-regions")]
-    fn scroll_region_up(&mut self, region: std::ops::Range<u16>, scroll_by: u16) -> io::Result<()> {
+    fn scroll_region_up(&mut self, region: ops::Range<u16>, scroll_by: u16) -> io::Result<()> {
         let width: usize = self.buffer.area.width.into();
         let cell_region_start = width * region.start.min(self.buffer.area.height) as usize;
         let cell_region_end = width * region.end.min(self.buffer.area.height) as usize;
@@ -409,11 +410,7 @@ impl Backend for TestBackend {
     }
 
     #[cfg(feature = "scrolling-regions")]
-    fn scroll_region_down(
-        &mut self,
-        region: std::ops::Range<u16>,
-        scroll_by: u16,
-    ) -> io::Result<()> {
+    fn scroll_region_down(&mut self, region: ops::Range<u16>, scroll_by: u16) -> io::Result<()> {
         let width: usize = self.buffer.area.width.into();
         let cell_region_start = width * region.start.min(self.buffer.area.height) as usize;
         let cell_region_end = width * region.end.min(self.buffer.area.height) as usize;
@@ -1026,7 +1023,7 @@ mod tests {
         #[case([A, B, C, D, E], 2..2, 2, [],                    [A, B, C, D, E])]
         fn scroll_region_up<const L: usize, const M: usize, const N: usize>(
             #[case] initial_screen: [&'static str; L],
-            #[case] range: std::ops::Range<u16>,
+            #[case] range: ops::Range<u16>,
             #[case] scroll_by: u16,
             #[case] expected_scrollback: [&'static str; M],
             #[case] expected_buffer: [&'static str; N],
@@ -1060,7 +1057,7 @@ mod tests {
         #[case([A, B, C, D, E], 2..2, 2, [A, B, C, D, E])]
         fn scroll_region_down<const M: usize, const N: usize>(
             #[case] initial_screen: [&'static str; M],
-            #[case] range: std::ops::Range<u16>,
+            #[case] range: ops::Range<u16>,
             #[case] scroll_by: u16,
             #[case] expected_buffer: [&'static str; N],
         ) {
--- a/ratatui-core/src/buffer/buffer.rs
+++ b/ratatui-core/src/buffer/buffer.rs
@@ -1,5 +1,5 @@
-use std::fmt;
-use std::ops::{Index, IndexMut};
+use core::ops::{Index, IndexMut};
+use core::{cmp, fmt};
 
 use unicode_segmentation::UnicodeSegmentation;
 use unicode_width::UnicodeWidthStr;
@@ -102,6 +102,8 @@ impl Buffer {
     }
 
     /// Returns the content of the buffer as a slice
+    // https://github.com/rust-lang/rust/issues/139338
+    #[allow(clippy::missing_const_for_fn)]
     pub fn content(&self) -> &[Cell] {
         &self.content
     }
@@ -497,8 +499,8 @@ impl Buffer {
 
             to_skip = current.symbol().width().saturating_sub(1);
 
-            let affected_width = std::cmp::max(current.symbol().width(), previous.symbol().width());
-            invalidated = std::cmp::max(affected_width, invalidated).saturating_sub(1);
+            let affected_width = cmp::max(current.symbol().width(), previous.symbol().width());
+            invalidated = cmp::max(affected_width, invalidated).saturating_sub(1);
         }
         updates
     }
@@ -590,7 +592,7 @@ impl fmt::Debug for Buffer {
                 } else {
                     overwritten.push((x, c.symbol()));
                 }
-                skip = std::cmp::max(skip, c.symbol().width()).saturating_sub(1);
+                skip = cmp::max(skip, c.symbol().width()).saturating_sub(1);
                 #[cfg(feature = "underline-color")]
                 {
                     let style = (c.fg, c.bg, c.underline_color, c.modifier);
@@ -636,7 +638,7 @@ impl fmt::Debug for Buffer {
 
 #[cfg(test)]
 mod tests {
-    use std::iter;
+    use core::iter;
 
     use itertools::Itertools;
     use rstest::{fixture, rstest};
--- a/ratatui-core/src/layout/constraint.rs
+++ b/ratatui-core/src/layout/constraint.rs
@@ -1,4 +1,4 @@
-use std::fmt;
+use core::fmt;
 
 use strum::EnumIs;
 
--- a/ratatui-core/src/layout/layout.rs
+++ b/ratatui-core/src/layout/layout.rs
@@ -1,12 +1,12 @@
-use std::cell::RefCell;
-use std::collections::HashMap;
-use std::iter;
-use std::num::NonZeroUsize;
-use std::rc::Rc;
+use alloc::rc::Rc;
+use core::cell::RefCell;
+use core::iter;
+use core::num::NonZeroUsize;
 
 use cassowary::strength::REQUIRED;
 use cassowary::WeightedRelation::{EQ, GE, LE};
 use cassowary::{AddConstraintError, Expression, Solver, Variable};
+use hashbrown::HashMap;
 use itertools::Itertools;
 use lru::LruCache;
 
@@ -1404,7 +1404,7 @@ mod tests {
     /// - underflow: constraint is for less than the full space
     /// - overflow: constraint is for more than the full space
     mod split {
-        use std::ops::Range;
+        use core::ops::Range;
 
         use itertools::Itertools;
         use pretty_assertions::assert_eq;
--- a/ratatui-core/src/layout/margin.rs
+++ b/ratatui-core/src/layout/margin.rs
@@ -1,4 +1,4 @@
-use std::fmt;
+use core::fmt;
 
 #[derive(Debug, Default, Clone, Copy, Eq, PartialEq, Hash)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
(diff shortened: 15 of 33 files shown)


orhun:
Thanks for the PR! This is definitely useful 👍🏼 

But maybe we can add this to [`templates`](https://github.com/ratatui/templates) instead so that it can be generated as a project template. We already have a [component architecture](https://github.com/ratatui/templates/blob/main/component/README.md) template and would be nice to have elm there as well.

WDYT?

CONTRIBUTING.md
```diff
@@ -32,7 +32,7 @@ guarantee that the behavior is unchanged.
 ### Code formatting
 
 Run `cargo xtask format` before committing to ensure that code is consistently formatted with
-rustfmt. Configuration is in [`rustfmt.toml`](./rustfmt.toml).
```
joshka, line 35: Let's put these changes in another PR as they're not really that related to this change

examples/apps/table-detail-tea/src/main.rs
```diff
@@ -0,0 +1,239 @@
+use std::io::stdout;
+/// A Ratatui example that demonstrates the Elm architecture with a basic list - detail
```
joshka, line 2: Nit: wrap at 100 chars

examples/apps/table-detail-tea/src/main.rs
```diff
@@ -0,0 +1,239 @@
...
+    self, DisableMouseCapture, EnableMouseCapture, Event, KeyCode, MouseEventKind,
+};
+use crossterm::ExecutableCommand;
+use ratatui::layout::{Constraint, Layout};
+use ratatui::style::{Modifier, Style};
+use ratatui::text::Line;
+use ratatui::widgets::{Block, Borders, Padding, Paragraph, Row, Table, TableState};
+use ratatui::Frame;
+
+static INFO_TEXT: &str = " (Esc/q) quit | (↑) move up | (↓) move down ";
```
joshka, line 22: nit: move to a constant in the method (putting text like this closer to where it's used is often a good approach)

examples/apps/table-detail-tea/src/main.rs
```diff
@@ -0,0 +1,239 @@
...
+
+/// Application data model and state
+#[derive(Debug, Default)]
+struct Model {
```
joshka, line 26: Would it be worth calling this AppModel - given the comment, this would seem to be aligned?

examples/apps/table-detail-tea/src/main.rs
```diff
@@ -0,0 +1,239 @@
...
+    Done,
+}
+
+/// Data model
+#[derive(Debug, Default)]
+struct Data {
+    name: String,
+    address: String,
+    email: String,
+}
```
joshka, line 45: Would it be worth calling this Item instead of data - slightly less abstract
Or even Person?

examples/apps/table-detail-tea/src/main.rs
```diff
@@ -0,0 +1,239 @@
...
+#[derive(PartialEq, Copy, Clone)]
+enum Message {
+    SelectNext,
+    SelectPrevious,
+    Quit,
+}
+
+fn main() -> color_eyre::Result<()> {
+    tui::install_panic_hook();
+    let mut terminal = tui::init_terminal()?;
```
joshka, line 56: use `ratatui::init()` here instead of having a full tui module
For disabling mouse capture, have a small method which just handles resetting the mouse capture, it will run before the ratatui installed hook

examples/apps/table-detail-tea/src/main.rs
```diff
@@ -0,0 +1,239 @@
...
+    let mut model = Model {
+        table_items: generate_some_fans(),
+        ..Default::default()
+    };
```
joshka, line 60: put this in a Default implementation

```rust
impl Default for Model {
    fn default() -> Self {
        Self {
            table_items: generate_some_fans(),
..
```

examples/apps/table-detail-tea/src/main.rs
```diff
@@ -0,0 +1,239 @@
...
+    stdout().execute(EnableMouseCapture)?;
```
joshka, line 61: Move before the model init

examples/apps/table-detail-tea/src/main.rs
```diff
@@ -0,0 +1,239 @@
...
+        terminal.draw(|f| view(&mut model, f))?;
+
+        // Handle events and map to a Message
+        let mut current_msg = handle_event(&model)?;
+
+        // Process updates as long as they return a non-None message
+        while current_msg.is_some() {
+            current_msg = update(&mut model, current_msg.unwrap());
+        }
+    }
```
joshka, line 73: Failing in the main part of the app (and propagating the error using `?`) should not cause the restore terminal code to run. The easiest way to avoid this problem is to extract this loop to a method (e.g. name it run) and store the result, but only evaluate it after restoring the terminal). See the rest of the examples for this pattern.

examples/apps/table-detail-tea/src/main.rs
```diff
@@ -0,0 +1,239 @@
...
+    stdout().execute(DisableMouseCapture)?;
+    tui::restore_terminal()?;
+    Ok(())
+}
+
+fn view(model: &mut Model, frame: &mut Frame) {
```
joshka, line 79: There's a bunch of methods that all have Model as the subject, this generally implies that this will make a good self parameter. Extract this out so these methods are all in an `impl Model` block

examples/apps/table-detail-tea/src/main.rs
```diff
@@ -0,0 +1,239 @@
...
+    let [top, bottom] = Layout::vertical([Constraint::Fill(1); 2]).areas(frame.area());
+
+    // Table setup
```
joshka, line 82: This method is a bit long - split it into render_table and render_detail

examples/apps/table-detail-tea/src/main.rs
```diff
@@ -0,0 +1,239 @@
...
+            Constraint::Fill(1),
+        ],
+    )
+    .header(header)
+    .row_highlight_style(Style::default().add_modifier(Modifier::REVERSED | Modifier::ITALIC))
+    .block(
+        Block::default()
+            .borders(Borders::ALL)
+            .title(" Terminal fans "),
+    );
```
joshka, line 110: nit: you can often save a bunch of vertical height due to automatic wrapping by adding an intermediate variables (e.g. widths here and using Block::bordered()) It's not much, but in example code lines of code are often worth saving to make it feel like less code to have to worry about.

```suggestion
        let widths = [
            Constraint::Length(20),
            Constraint::Min(10),
            Constraint::Fill(1),
        ];
        let table = Table::new(rows, widths)
            .header(header)
            .row_highlight_style(Style::new().add_modifier(Modifier::REVERSED | Modifier::ITALIC))
            .block(Block::bordered().title(" Terminal fans "));
```

examples/apps/table-detail-tea/src/main.rs
```diff
@@ -0,0 +1,239 @@
...
+
+    frame.render_stateful_widget(table, top, &mut model.table_state);
+
+    // Select the first row if no row is selected
+    if model.table_state.selected().is_none() {
+        model.table_state.select_first();
+    }
```
joshka, line 117: Does this need to go before the render call? Should it be in the main part of the app during some sort of update path rather than in the render code?

examples/apps/table-detail-tea/src/main.rs
```diff
@@ -0,0 +1,239 @@
...
+        selected_item.email, selected_item.address
+    ))
+    .block(
+        Block::default()
+            .borders(Borders::ALL)
+            .title(format!(" {} ", selected_item.name))
+            .title_bottom(Line::from(INFO_TEXT))
+            .padding(Padding::new(1, 1, 1, 1)),
+    );
+    frame.render_widget(detail, bottom);
```
joshka, line 132:
```suggestion
    let selected_item = &self.table_items[self.table_state.selected().unwrap()];
    let block = Block::bordered()
        .title(format!(" {} ", selected_item.name))
        .title_bottom(Line::from(INFO_TEXT))
        .padding(Padding::new(1, 1, 1, 1));
    let text = format!("{}\n\n{}", selected_item.email, selected_item.address);
    let detail = Paragraph::new(text).block(block);
    frame.render_widget(detail, bottom);
```

examples/apps/table-detail-tea/src/main.rs
```diff
@@ -0,0 +1,239 @@
...
+            let email = contact::email();
+
+            Data {
+                name,
+                address,
+                email,
+            }
+        })
+        .collect()
+}
```
joshka, line 239: Consider generating a model that has all the fields and then doing the to string parts at render time. E.g. `impl From<Person> for Row` and 
```
struct Person {
    name: String,
    address: Address,
    email: String,
}


impl Person {
    fn fake() -> Self {
        Self {
            name: fakeit::name::full(),
            address: Address::fake(),
            email: fakeit::contact::email(),
        }
    }
}

impl Address {
    fn fake() -> Self { ... }
}
```

examples/vhs/table-detail-tea.tape
```diff
@@ -0,0 +1,20 @@
+# This is a vhs script. See https://github.com/charmbracelet/vhs for more info.
+# To run this script, install vhs and run `vhs ./examples/vhs/table-detail-tea.tape`
+Output "target/table-detail-tea.gif"
+Set Theme "Aardvark Blue"
+Set Width 1400
```
joshka, line 5: Constrain this to max 1200 - otherwise GitHub readmes look bad when they scale the image

examples/vhs/table-detail-tea.tape
joshka, line 1: Can you run VHS publish on this and add link to the screenshot in the main README. We can sub in the generated version linking to the images branch just before merging this.

examples/apps/table-detail-tea/src/main.rs
```diff
@@ -0,0 +1,239 @@
...
+        let original_hook = panic::take_hook();
+        panic::set_hook(Box::new(move |panic_info| {
+            stdout().execute(LeaveAlternateScreen).unwrap();
+            disable_raw_mode().unwrap();
+            original_hook(panic_info);
+        }));
+    }
+}
+
+fn generate_some_fans() -> Vec<Data> {
```
joshka, line 217: What are fans in this context?

examples/apps/table-detail-tea/src/main.rs
```diff
@@ -0,0 +1,239 @@
...
+            let current_index = model.table_state.selected().unwrap_or(0);
+            if current_index > 0 {
+                model.table_state.select(Some(current_index - 1));
+            }
+        }
+    };
+    None
+}
+
+mod tui {
```
joshka, line 183: Mostly this would be dropped - see the comment further up

joshka (commented):
Thanks for the PR - a bunch of improvements to take this from good code to (mostly subjectively) better code.

orhun:
@joshka what do you think about moving this to templates?

examples/vhs/table-detail-tea.tape
Tuurlijk: Ok, I'll also commit changes for table.tape, where I copied it from.

examples/apps/table-detail-tea/src/main.rs
Tuurlijk: Verbatim copy from the manual. Please adjust as you see fit.

examples/apps/table-detail-tea/src/main.rs
Tuurlijk: That may be better, but for the sake of the example, this may be easier to understand.

```rust
fn generate_some_people() -> Vec<Person> {
    (0..50)
        .map(|_| Person {
            name: name::full(),
            address: generate_fake_address(),
            email: contact::email(),
        })
        .collect()
}

fn generate_fake_address() -> String {
    format!(
        "{}\n{}, {} {}",
        address::street(),
        address::city(),
        address::state(),
        address::zip()
    )
}
```

examples/vhs/table-detail-tea.tape
Tuurlijk: Can this `vhs publish` be documented in the vhs readme?

examples/apps/table-detail-tea/src/main.rs
Tuurlijk: https://github.com/ratatui/ratatui/actions/runs/14189010632/job/39749447988?pr=1741

???

examples/apps/table-detail-tea/src/main.rs
Tuurlijk: Moved to main. Right after table init. Update only fires when there is a `current_msg`.

examples/apps/table-detail-tea/src/main.rs
Tuurlijk: Please adjust as you see fit. Copied it from: https://ratatui.rs/concepts/application-patterns/the-elm-architecture/

joshka:
> @joshka what do you think about moving this to templates?

I think it would be useful to have as an example. One caveat though is that it's worth going back to the elm architecture source and looking at that as the definitive approach for a TEA app.

But from there it's also worth noting when and how idiomatic rust would do something different. E.g. TEA works very well with the idea of immutable states which are modified by messages etc. and which return a new state. Here we're mutating the model instead of replacing. The idea is mostly the same result, but it's not quite TEA. That said, I'm not super familiar with TEA in general. I guess the takeaway is treat the ratatui website as mostly fan fiction with regard to TEA, and outdated already with regards to ratatui.

To move this to a template, I'd want someone to invest time into documenting the design decisions a bit more, and extracting something from real world usage. The biggest questions I'd have are "what problems are being solved", and "what differentiates this solution from the other approaches" as well as "how is this similar to elm in other languages" and which concepts are mapped into which concrete code ideas. (I have a vague recollection of "Ports" being a thing in TEA... and that's not something that seems to be converted here at all.)

examples/apps/table-detail-tea/src/main.rs
joshka: My apologies. It was the period that made it 101 :D
I copied the line into wc -c because it looked short, and added application manually, but without the .

joshka:
Bumped a bunch of things in https://github.com/ratatui/ratatui/pull/1741/commits/5db04d30e29e2f8ba5e2f25e306ff8c516671ab4

Any thoughts before merging @Tuurlijk

orhun:
> To move this to a template, I'd want someone to invest time into documenting the design decisions a bit more, and extracting something from real world usage. 

Isn't adding this to the templates helpful for that? We can get actual user feedback that way.

I think majority of the time people use templates for structuring their application, not the examples. I believe adding them to examples would make this pretty much invisible, especially with the `-tea` suffix which is not super clear from first sight.

This either needs better documentation on the website for more visibility or to become a part of the templates for practical usage.

joshka:
> > To move this to a template, I'd want someone to invest time into documenting the design decisions a bit more, and extracting something from real world usage.
> 
> Isn't adding this to the templates helpful for that? We can get actual user feedback that way.
> 
> I think majority of the time people use templates for structuring their application, not the examples. I believe adding them to examples would make this pretty much invisible, especially with the `-tea` suffix which is not super clear from first sight.
> 
> This either needs better documentation on the website for more visibility or it should become a part of the templates for practical usage.

Making something into a template bakes in architecture choices even if they're bad choices. TEA definitely needs more "why" before "how"

Tuurlijk:
> Bumped a bunch of things in [5db04d3](https://github.com/ratatui/ratatui/commit/5db04d30e29e2f8ba5e2f25e306ff8c516671ab4)
> 
> Any thoughts before merging @Tuurlijk

No, drawing blanks. Go for it!

joshka:
TODO - update readme images.

examples/apps/table-detail-tea/README.md
```diff
@@ -0,0 +1,11 @@
+# Table detail view
+
+This example shows how to create a table of items and a detail view.
+
+It uses the Elm architecture or TEA approach.
```
orhun, line 5: Can we link to some resources about this? I get tea pictures and brewing methods when I search for TEA approach.

examples/apps/table-detail-tea/src/main.rs
```diff
@@ -0,0 +1,212 @@
+use std::io::stdout;
+use std::panic;
```
orhun, line 2: Move imports after the comment at top

examples/apps/table-detail-tea/src/main.rs
```diff
@@ -0,0 +1,212 @@
...
+    self, DisableMouseCapture, EnableMouseCapture, Event, KeyCode, MouseEventKind,
+};
+use crossterm::ExecutableCommand;
+use fakeit::{address, contact, name};
+use ratatui::layout::{Constraint, Layout, Rect};
+use ratatui::style::{Modifier, Style};
+use ratatui::widgets::{Block, Padding, Paragraph, Row, Table, TableState};
+use ratatui::{DefaultTerminal, Frame};
+
+/// Application data model and state
```
orhun, line 23: hmm is this comment really useful? This should be obvious imo

examples/apps/table-detail-tea/src/main.rs
```diff
@@ -0,0 +1,212 @@
...
+                if message == Message::Quit {
+                    return Ok(());
+                }
+                maybe_message = self.update(message);
+            }
+        }
+    }
+
+    fn update(&mut self, msg: Message) -> Option<Message> {
+        match msg {
```
orhun, line 98: nit for consistency

```suggestion
    fn update(&mut self, message: Message) -> Option<Message> {
        match message {
```

examples/apps/table-detail-tea/README.md
Tuurlijk: I guess the manual/book needs to be updated too then? Who added the section in the first place? Can that person elaborate? I can asked Grok for an explanation and links:

https://x.com/i/grok/share/NnbjuMB0KQzwMFmj9DZcVVnaI

But since the whole Elm thing seems to be a question, should we not ask what the best architecture for TUI apps on top of immediate mode rendering apps on top of ratatui is.

Then  Grok comes back with TEA.

Alternatives Considered
MVC: Too heavy and better suited for retained mode.
ECS: Overkill for TUI simplicity.
MVVM: Less compatible with immediate mode’s redraw-every-frame approach.

TEA’s balance of simplicity and structure outperforms these for this context.
Conclusion
The Elm Architecture (TEA) is the best architectural approach for a Ratatui TUI application using Immediate Rendering Mode. It delivers a clean, modular, and efficient solution that meets all specified goals while leveraging Rust’s strengths and Ratatui’s capabilities. Whether for simple or complex TUIs, TEA scales effectively with minimal overhead.

orhun: I mean I just meant some general documentation about Elm architecture and TEA approach, if possible.

joshka: https://guide.elm-lang.org/architecture/ and https://github.com/charmbracelet/bubbletea/tree/main/tutorials are the two main places I'm aware of

Realistically though, I think Ratatui's approach here only passingly resembles both those.

examples/apps/table-detail-tea/src/main.rs
```diff
@@ -0,0 +1,212 @@
+use std::io::stdout;
+use std::panic;
+/// A Ratatui example that demonstrates the Elm architecture with a basic list - detail
+/// application.
+///
+/// This example runs with the Ratatui library code in the branch that you are currently
+/// reading. See the [`latest`] branch for the code which works with the most recent Ratatui
+/// release.
+///
+/// [`latest`]: https://github.com/ratatui/ratatui/tree/latest
```
joshka, line 10:
```suggestion
//! A Ratatui example that demonstrates the Elm architecture with a basic list - detail
//! application. See <https://ratatui.rs/concepts/application-patterns/the-elm-architecture/>
//!
//! This example runs with the Ratatui library code in the branch that you are currently
//! reading. See the [`latest`] branch for the code which works with the most recent Ratatui
//! release.
//!
//! [`latest`]: https://github.com/ratatui/ratatui/tree/latest
use std::io::stdout;
use std::panic;
```

examples/apps/table-detail-tea/src/main.rs
joshka, line 23:
```suggestion
```

examples/apps/table-detail-tea/README.md
orhun: Ah alright then

examples/apps/table-detail-tea/README.md
```diff
@@ -0,0 +1,11 @@
+# Table detail view
+
+This example shows how to create a table of items and a detail view.
+
+It uses the Elm architecture or TEA approach as described on [the ratatui website](https://ratatui.rs/concepts/application-patterns/the-elm-architecture/).
```
joshka, line 5: This will make markdownlint angry due to line length

Fix:
```suggestion
It uses [the Elm architecture(TEA)](https://ratatui.rs/concepts/application-patterns/the-elm-architecture/).
```

examples/apps/table-detail-tea/src/main.rs
Tuurlijk: Line has been removed

examples/apps/table-detail-tea/src/main.rs
Tuurlijk: Done

orhun (approved):

examples/apps/table-detail-tea/src/main.rs
```diff
@@ -0,0 +1,211 @@
...
+
+    let model = AppModel::default();
+    let result = model.run(terminal);
+
+    stdout().execute(DisableMouseCapture)?;
+    ratatui::restore();
+    result
+}
+
+pub fn install_panic_hook() {
```
joshka, line 64: nit: add a comment here. a small note like:

```suggestion
/// Ensure that mouse mode is disabled if the app crashes
pub fn install_panic_hook() {
```

examples/apps/table-detail-tea/src/main.rs
```diff
@@ -0,0 +1,211 @@
...
+            terminal.draw(|frame| self.view(frame))?;
+
+            let mut maybe_message = handle_event()?;
+            while let Some(message) = maybe_message {
+                if message == Message::Quit {
+                    return Ok(());
+                }
+                maybe_message = self.update(message);
+            }
+        }
```
joshka, line 93: It would be nice to add some docs about how this loop works. Mostly because it's a while loop within another loop, so from an initial read it's unclear how the interaction between the loops is intended to work.

As a reviewer seeing this for the first time (well I read it before, but that was a while ago), I don't know specifically what the intent is, I can only see that there's a nested loop. I don't have any intent to judge its correctness on.

examples/apps/table-detail-tea/src/main.rs
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    fn update(&mut self, message: Message) -> Option<Message> {
+        match message {
+            Message::Quit => {}
+            Message::SelectNext => self.table_state.select_next(),
+            Message::SelectPrevious => self.table_state.select_previous(),
+        };
+        None
+    }
```
joshka, line 103: Given this always returns None, there should be some documentation about what the purpose and why it needs to have a return value here at all. Alternatively add a use case where the return value is used (I'm not sure how that looks)

examples/apps/table-detail-tea/src/main.rs
```diff
@@ -0,0 +1,211 @@
...
+
+    /// Render the current view
+    fn view(&mut self, frame: &mut Frame) {
+        let [top, bottom] = Layout::vertical([Constraint::Fill(1); 2]).areas(frame.area());
```
joshka, line 107: a small update for 0.30:

```suggestion
        let [top, bottom] = frame.area().layout(Layout::vertical([Constraint::Fill(1); 2]));
```

examples/apps/table-detail-tea/src/main.rs
```diff
@@ -0,0 +1,211 @@
...
+        self.render_table(frame, top);
+        self.render_detail(frame, bottom);
+    }
+
+    fn render_table(&mut self, frame: &mut Frame, area: Rect) {
+        let header = Row::new(vec!["Name", "Address", "Email"])
+            .style(Modifier::BOLD)
+            .height(1);
```
joshka, line 115: nit: What about moving the row header to the Person impl?
(is height 1 the default btw?)

E.g.:
```suggestion
        let header = Person::header_row().bold();
```

or 
```suggestion
        let header = Row::from(Person::headers()).bold();
```

Not sure which is nicer - just a thought (especially since there's a conversion from Person to Row already below)

examples/apps/table-detail-tea/src/main.rs
```diff
@@ -0,0 +1,211 @@
...
+        let rows = self.people.iter().map(Row::from);
+        let widths = [
+            Constraint::Length(20),
+            Constraint::Min(10),
+            Constraint::Fill(1),
+        ];
+        let table = Table::new(rows, widths)
+            .header(header)
+            .row_highlight_style(Style::new().add_modifier(Modifier::REVERSED | Modifier::ITALIC))
```
joshka, line 124:
```suggestion
            .row_highlight_style(Style::new().reversed().italic()
```

(and `use ratatui::style::Stylize`)

examples/apps/table-detail-tea/src/main.rs
```diff
@@ -0,0 +1,211 @@
...
+            .block(Block::bordered().title(" People "));
+
+        frame.render_stateful_widget(table, area, &mut self.table_state);
+    }
+
+    fn render_detail(&self, frame: &mut Frame, area: Rect) {
+        let selected_item = &self.people[self.table_state.selected().unwrap()];
```
joshka, line 131: Generally let's avoid unwrap in examples - here this is used because we expect there always to be a selection. We should probably error out if this isn't true. Maybe use expect, but it would be better to render an error instead and not crash.

examples/apps/table-detail-tea/src/main.rs
```diff
@@ -0,0 +1,211 @@
...
+        let detail = Paragraph::new(text).block(block);
+        frame.render_widget(detail, area);
+    }
+}
+
+/// Handle events and map to a Message
+fn handle_event() -> color_eyre::Result<Option<Message>> {
+    if !event::poll(Duration::from_millis(250))? {
+        return Ok(None);
+    }
```
joshka, line 146: nit: we've tended to move away from examples incorporating polls and re-rendering when there's no events to handle like this. Polling with a duration allows your app to do other things on fixed intervals (like animations, etc.) If that's actually necessary in this app, it's worth calling out the behavior / rationale.

So maybe either remove this, add some information about how it's useful, or actually make use of it. All 3 are valid approaches, though I'd prefer the first or the last. The last adds complexity though.

examples/apps/table-detail-tea/src/main.rs
```diff
@@ -0,0 +1,211 @@
...
+    match event::read()? {
+        Event::Key(key) if key.kind == event::KeyEventKind::Press => Ok(handle_key(key)),
```
joshka, line 148: crossterm 0.29 change

```suggestion
        Event::Key(key) if key.is_press() => Ok(handle_key(key)),
```

examples/vhs/table-detail-tea.tape
joshka: The vhs publish command is covered in their readme:
https://github.com/charmbracelet/vhs?tab=readme-ov-file#publish-tapes

Expanding on what I meant to ask for here is to show the output of how this looks. These easiest way is to generate it and publish using vhs directly (run `vhs examples/vhs/table-detail-tea.tape --publish` and paste the markdown link).

We store images on a separate branch of the repo to avoid bloating the repo in a way that makes it difficult to reduce over time as images get edited etc. But PRs don't handle working on two branches at the same time, so we handle the step of updating images manually rather than as part of a PR.

We've got some documents on this in a few places, but it's mostly something that we just do rather than worry too much about capturing exactly how it works:
- https://github.com/ratatui/ratatui/blob/main/examples/README.md (at the bottom)
- https://github.com/ratatui/ratatui/blob/main/RELEASE.md sort of covers part of this
- https://github.com/ratatui/ratatui/blob/main/examples/vhs/generate.bash sort of covers some of it

examples/apps/table-detail-tea/src/main.rs
```diff
@@ -0,0 +1,211 @@
+/// A Ratatui example that demonstrates the Elm architecture with a basic list - detail
+/// application.
+///
+/// This example runs with the Ratatui library code in the branch that you are currently
+/// reading. See the [`latest`] branch for the code which works with the most recent Ratatui
+/// release.
+///
+/// [`latest`]: https://github.com/ratatui/ratatui/tree/latest
```
joshka, line 8: It would be good to summarize how this example demonstrates TEA here (also, use //! instead of ///)

joshka (commented):
Hey there, a few nits / small changes, but generally ok with this.

The main documentation comments are the key ones to cover here. I'm putting my user hat on and asking questions like:
- what can I learn from this example in terms of how TEA works?; and
- what are the interesting parts of this example that are novel design choices?

There may be other things, but given you're querying LLMs for info on TEA, perhaps add some of the sorts of info that you've asked about into the example too.

Also, if you want to go a bit meta, the example data could use different types of tea instead of fake people :D

PR #1747, opened 2025-04-01
perf(table): replace while loop with simple min operation

EdJoPaTo:
While looking into the performance of table I noticed this. I changed it towards [the same approach used as in my table widget](https://github.com/EdJoPaTo/tui-rs-tree-widget/blob/6a5243243a7c4a608ef5f0b632b527baae649752/src/lib.rs#L208-L210).

It shouldn’t have much impact on small changes to offset / selection. Bigger changes upwards no longer require checking the height of many rows multiple times as the function has one less loop involved. The first loop still makes sure everything is correct.

This case might also exist in other widgets, haven’t checked so far.

### DIFF
--- a/ratatui-widgets/src/table.rs
+++ b/ratatui-widgets/src/table.rs
@@ -910,6 +910,11 @@ impl Table<'_> {
     fn visible_rows(&self, state: &TableState, area: Rect) -> (usize, usize) {
         let last_row = self.rows.len().saturating_sub(1);
         let mut start = state.offset.min(last_row);
+
+        if let Some(selected) = state.selected {
+            start = start.min(selected);
+        }
+
         let mut end = start;
         let mut height = 0;
 
@@ -931,16 +936,6 @@ impl Table<'_> {
                 while height > area.height {
                     height = height.saturating_sub(self.rows[start].height_with_margin());
                     start += 1;
-                }
-            }
-
-            // scroll up until the selected row is visible
-            while selected < start {
-                start -= 1;
-                height = height.saturating_add(self.rows[start].height_with_margin());
-                while height > area.height {
-                    end -= 1;
-                    height = height.saturating_sub(self.rows[end].height_with_margin());
                 }
             }
         }


joshka (approved):
LGTM, nice find. It's a good simplification of the logic regardless of perf, but what were the perf results?

joshka:
> This case might also exist in other widgets, haven’t checked so far.

I'd guess probably in the list widget too.

joshka:
Taking a deeper look through the tests for this code, the only parts which are currently tested are items with height = 1 and no top/bottom margins. I was initially thinking that this would mean that the change possibly isn't covered, but I think it's likely still covered just as much.

https://github.com/ratatui/ratatui/blob/c27fba06f1824543fa8807d85cdbdc1c23108c75/ratatui-widgets/src/table.rs#L1641-L1670

ratatui-widgets/src/table.rs
```diff
@@ -910,6 +910,11 @@ impl Table<'_> {
     fn visible_rows(&self, state: &TableState, area: Rect) -> (usize, usize) {
         let last_row = self.rows.len().saturating_sub(1);
         let mut start = state.offset.min(last_row);
+
+        if let Some(selected) = state.selected {
+            start = start.min(selected);
+        }
+
```
joshka, line 917:
```suggestion
        let mut start = state
            .selected
            .map_or(state.offset, |i| state.offset.min(i))
            .min(last_row);
```

joshka (approved):

ratatui-widgets/src/table.rs
EdJoPaTo: I think this is way harder to read, even when `i` is replaced with `selected`

EdJoPaTo:
> what were the perf results?

Not significant in comparison to creating thousands of Rows and near a million cells on every render which is why I have to build my own widget anyway.

The significant performance change is only on scrolling upward. And it's a difference of O(n) vs O(1), where the logic is mostly arithmetic and branching, so the branching is likely the expensive part here. When LLVM is smart it shrinks the branching into simple arithmetics, so any benchmarking here would end up benchmarking the optimizations LLVM finds for the given CPU architecture. I don't think its worth the effort of benchmarking this. 3 loops into 2 loops is the significant part here.

orhun (approved):
LGTM, nothing to remark

PR #1753, opened 2025-04-03
feat(no_std): remove redundant `std` usages in `ratatui-core`

j-g00da:
Resolves https://github.com/ratatui/ratatui/issues/1751

### DIFF
--- a/ratatui-core/src/backend.rs
+++ b/ratatui-core/src/backend.rs
@@ -341,8 +341,11 @@ pub trait Backend {
     /// For examples of how this function is expected to work, refer to the tests for
     /// [`TestBackend::scroll_region_up`].
     #[cfg(feature = "scrolling-regions")]
-    fn scroll_region_up(&mut self, region: std::ops::Range<u16>, line_count: u16)
-        -> io::Result<()>;
+    fn scroll_region_up(
+        &mut self,
+        region: core::ops::Range<u16>,
+        line_count: u16,
+    ) -> io::Result<()>;
 
     /// Scroll a region of the screen downwards, where a region is specified by a (half-open) range
     /// of rows.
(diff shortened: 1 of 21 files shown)


ratatui-core/src/backend.rs
```diff
@@ -341,8 +341,11 @@ pub trait Backend {
     /// For examples of how this function is expected to work, refer to the tests for
     /// [`TestBackend::scroll_region_up`].
     #[cfg(feature = "scrolling-regions")]
-    fn scroll_region_up(&mut self, region: std::ops::Range<u16>, line_count: u16)
-        -> io::Result<()>;
+    fn scroll_region_up(
+        &mut self,
+        region: core::ops::Range<u16>,
```
joshka, line 346: nit: use an import instead of fully qualifying this here (and below)

ratatui-core/src/lib.rs
```diff
@@ -38,6 +38,8 @@
 //!
 //! This project is licensed under the MIT License. See the [LICENSE](../LICENSE) file for details.
 
+extern crate alloc;
```
joshka, line 41: Can you use the `use` syntax here, or `extern crate` required?

joshka (approved):
General nit: particularly where the change makes the wrapping worse, I'd use an import rather than just the direct change.

Hashbrown change is fine - this doesn't affect the public API surface as we don't accept HashMap as a parameter / return it anywhere.

ratatui-core/src/lib.rs
j-g00da: Yes it's required. `alloc` unlike `core` has to be explicitly added using `extern crate`.

j-g00da:
Ready, merge #1755 first.

joshka (approved):

PR #1755, opened 2025-04-03
fix: correct clippy errors introduced by rust 1.86.0 update

j-g00da:
New version of rust (1.86.0) caused CI to fail.


j-g00da:
There is one last error in `ratatui-core/src/buffer/buffer.rs:105`. 
Clippy wants to make this function const, but then returns:
`error[E0015]: cannot perform non-const deref coercion on `std::vec::Vec<buffer::cell::Cell>` in constant functions`
What? You wanted that.

Should I just leave it not-const and silence clippy in that line?

joshka:
> Should I just leave it not-const and silence clippy in that line?

yes

I raised this as an issue with the lint - it's fixed in beta already though.

https://github.com/rust-lang/rust/issues/139338

joshka (approved):
LGTM - thanks

PR #1759, opened 2025-04-04
chore(lint): add `std` instead of `core`/`alloc` lints to `ratatui-core`

### LINKED ISSUE #1752: `ratatui-core` - Add linter rules to prefer `core` and `alloc` over `std` and `core` over `alloc`.
Inspiration: https://github.com/bevyengine/bevy/pull/15281

j-g00da:
Resolves #1752

### DIFF
--- a/ratatui-core/Cargo.toml
+++ b/ratatui-core/Cargo.toml
@@ -65,5 +65,51 @@ pretty_assertions.workspace = true
 rstest.workspace = true
 serde_json.workspace = true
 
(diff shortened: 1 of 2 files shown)


joshka:
What about adding the lint to lib.rs as an attribute?

j-g00da:
> What about adding the lint to lib.rs as an attribute?

Yes, that's probably a better idea.

joshka (approved):

PR #1763, opened 2025-04-05
chore(lint): add `std` instead of `core`/`alloc` lints to `ratatui-widgets`

### LINKED ISSUE #1761: `ratatui-widgets` - Add linter rules to prefer core and alloc over std and core over alloc

j-g00da:
Resolves #1761 

Requires #1762 merged first.


EdJoPaTo:
Why not in the `Cargo.toml`? Would also lint examples and tests at the same time but I don't think it's that bad?

j-g00da:
> Why not in the `Cargo.toml`? Would also lint examples and tests at the same time but I don't think it's that bad?

https://github.com/rust-lang/cargo/issues/13157

We only want these lints in core and widgets for now, and cargo doesn't support overriding workspace lints.

orhun (approved):

joshka (approved):

Issue #1745, opened 2025-03-31
Unwanted whitespaces are added to chinese characters, breaking layout


piscopancer:
for some reason when i go to the right pane ratatui decides to render an additional whitespace between certain chinese characters, pushing the whole layout to the right, even making alacritty render right-edge characters on the next line

https://github.com/user-attachments/assets/79f6c58c-9d10-4573-a679-db49697c9955

is it a know issue?

piscopancer:
![Image](https://github.com/user-attachments/assets/6e92e310-1997-4d4c-b3c1-84576ccb6f98)

tested in vscode, breaks there too, it's not the terminals fault

orhun:
Thanks for reporting! There were some similar issues in the past related to word wrapping / reflow / unicode but I don't remember seeing this one.

Can you share a minimal example (or some text) for reproducing the issue?

(Coincidentally I am in Hong Kong right now :D)

joshka:
This is a problem with 2-3 parts:
1. There isn't a clearly defined size that terminals should move the cursor when rendering characters 
2. We use unicode-width to give us a width, but it's not 100% accurate
3. Even if there were a definition, terminals implement things differently

I was thinking about a generic way that we might be able to solve this problem. Maybe we could update the code that sends widths to the backend to allow for an Unknown width, which the backend works out by checking the cursor location after (and then caches that value).

Some things we'd have to do to get this sort of thing correct are:
1. Render a bunch of characters that are known and unknown across all the terminals
2. Look at what's actually rendered to understand if cursor location matches actual reported location
3. Feed that back into the implementation

There's a lot of complexity in getting this right. It's a big bucket of edge cases basically.

piscopancer:
✔ figured out a way to fix it, i just do not need to render a debug widget, may not even remove the area, it will not jitter
```rust
let debug_p = Paragraph::new(
  vec![Line::from(format!("sec [{:?}] ", self.focused_el))]
).light_magenta();

// frame.render_widget(&debug_p, debug_area);
```
it's that very first line

⚠ though not rendering a widget that has no solid reason for not rendering is not a way out

![Image](https://github.com/user-attachments/assets/2e488533-801e-4cc3-9577-14cc9681234a)

(plus some artifacts in inputs on the right, but that's not too much unless you are a perfectionist)

orhun:
Not sure if I got how you fixed it, you just don't render it?

piscopancer:
What helped is not rendering that widget at the very top
[...]

PR #1764, opened 2025-04-05
Fixed  issue 1745

ZhangHanDong:
This PR fixes issue [#1745](https://github.com/ratatui/ratatui/issues/1745) where Chinese characters and other multi-width characters were being rendered with additional unwanted whitespace between them. The problem occurred in the set_stringn method when handling characters that occupy multiple cells in the terminal.

The fix ensures proper handling of multi-width characters by explicitly resetting cells covered by these characters and advancing the cursor position by the exact width of each character. This prevents the insertion of unexpected whitespace between characters and maintains correct alignment.

A comprehensive test suite has been added to verify both pure Chinese text and mixed English/Chinese text rendering, confirming that multi-width characters occupy the correct number of cells without additional spacing.

**However, since I was unable to reproduce the error in issue 1745 locally, I tried adding Chinese to the todo list, and it seems to work fine.**

Screenshot 2025-04-05 at 17 05 29

### DIFF
--- a/ratatui-core/src/buffer/buffer.rs
+++ b/ratatui-core/src/buffer/buffer.rs
@@ -354,7 +354,7 @@ impl Buffer {
                 remaining_width = remaining_width.saturating_sub(width);
                 Some((symbol, width))
             });
-        
+
         let style = style.into();
         for (symbol, width) in graphemes {
             self[(x, y)].set_symbol(symbol).set_style(style);
@@ -663,7 +663,7 @@ mod tests {
         let expected = "Buffer {\n    area: Rect { x: 0, y: 0, width: 0, height: 0 }\n}";
         assert_eq!(result, expected);
     }
-    
+
     #[test]
     fn chinese_characters_rendering() {
         let area = Rect::new(0, 0, 10, 1);
@@ -671,7 +671,7 @@ mod tests {
 
         // Test with a string containing Chinese characters
         buffer.set_string(0, 0, "你好世界", Style::default());
-        
+
         // Chinese characters are each 2 cells wide, so "你好世界" should take 8 cells
         // Check that cells are properly filled (not with extra spaces)
         let contents = buffer.content.iter().map(Cell::symbol).collect::<Vec<_>>();
@@ -686,11 +686,11 @@ mod tests {
         // Cells 8-9 are untouched
         assert_eq!(contents[8], " ");
         assert_eq!(contents[9], " ");
-        
+
         // Reset and test with mixed English and Chinese
         let mut buffer = Buffer::empty(area);
         buffer.set_string(0, 0, "hi你好", Style::default());
-        
+
         let contents = buffer.content.iter().map(Cell::symbol).collect::<Vec<_>>();
         assert_eq!(contents[0], "h");
         assert_eq!(contents[1], "i");
--- a/ratatui-core/src/buffer/buffer.rs
+++ b/ratatui-core/src/buffer/buffer.rs
@@ -342,22 +342,31 @@ impl Buffer {
         let mut remaining_width = self.area.right().saturating_sub(x).min(max_width);
         let graphemes = UnicodeSegmentation::graphemes(string.as_ref(), true)
             .filter(|symbol| !symbol.contains(char::is_control))
-            .map(|symbol| (symbol, symbol.width() as u16))
+            .map(|symbol| {
+                let width = symbol.width() as u16;
+                (symbol, width)
+            })
             .filter(|(_symbol, width)| *width > 0)
             .map_while(|(symbol, width)| {
-                remaining_width = remaining_width.checked_sub(width)?;
+                if width > remaining_width {
+                    return None;
+                }
+                remaining_width = remaining_width.saturating_sub(width);
                 Some((symbol, width))
             });
+        
         let style = style.into();
         for (symbol, width) in graphemes {
             self[(x, y)].set_symbol(symbol).set_style(style);
-            let next_symbol = x + width;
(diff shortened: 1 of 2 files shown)


ratatui-core/src/buffer/buffer.rs
```diff
@@ -345,19 +345,25 @@ impl Buffer {
...
             self[(x, y)].set_symbol(symbol).set_style(style);
-            let next_symbol = x + width;
-            x += 1;
-            // Reset following cells if multi-width (they would be hidden by the grapheme),
-            while x < next_symbol {
-                self[(x, y)].reset();
-                x += 1;
+            // For each grapheme, advance x by exactly its width
+            // This ensures that multi-width characters like Chinese characters
+            // are properly handled without adding extra whitespace
```
joshka, line 360: nit: wrap comments at 100 chars

ratatui-core/src/buffer/buffer.rs
```diff
@@ -345,19 +345,25 @@ impl Buffer {
...
+            for i in 1..width {
+                if x + i < self.area.right() {
+                    self[(x + i, y)].reset();
+                }
             }
+            x += width;
```
joshka, line 366: This is definitely better, but it's not clear that the if check is required (the purpose of the map_while above is to exclude characters which would make `x + width >= area.right()`

ratatui-core/src/buffer/buffer.rs
```diff
@@ -655,6 +661,47 @@ mod tests {
         assert_eq!(result, expected);
     }
 
+    #[test]
+    fn chinese_characters_rendering() {
```
joshka, line 665: I checked this test against the current code and it succeeds already.

joshka (commented):
It's not clear what's actually changed here that would fix issue 1745. From my reading, of this it seems like there's actually no logical change here.

I suspect that the problems would be around the edge cases that occur when a wide character is at the end of line.

The way to show that the problem is fixed is to create a small unit test that actually triggers the problem, and to see that unit test go from failing to succeeding with the change. It's not clear that's happening here.

orhun:
I agree that adding chinese characters to the examples is a good idea (and I like the test suite) but I'm not so sure if this PR touches any existing rendering issues...

ratatui-core/src/buffer/buffer.rs
ZhangHanDong: I’ve added more tests locally and found that the current main branch code will fail when executing these tests. I’ll submit the modifications after conducting more tests and pinpointing some issues.

ZhangHanDong:
@joshka 

I added two new tests that fail in the buffer.rs file of the main branch. 

```rust
    #[test]
    fn chinese_at_buffer_boundary() {
        let area = Rect::new(0, 0, 5, 1);

        // Should fit exactly: "a" + "你" + "b"
        buffer.set_string(0, 0, "a你b", Style::default());

        let contents = buffer.content.iter().map(Cell::symbol).collect::<Vec<_>>();
        assert_eq!(contents[0], "a");
        assert_eq!(contents[1], "你");
        assert_eq!(contents[2], " "); // Should be reset/empty
        assert_eq!(contents[3], "b");

        // Now try rendering where the last character would be cut off
        buffer = Buffer::empty(area);
        buffer.set_string(0, 0, "a你好", Style::default());

        let contents = buffer.content.iter().map(Cell::symbol).collect::<Vec<_>>();
        assert_eq!(contents[0], "a");
        assert_eq!(contents[1], "你");
        assert_eq!(contents[2], " "); // Should be reset/empty
        // Character "好" should not appear as it doesn't fit
        assert_eq!(contents[3], " ");
        assert_eq!(contents[4], " ");
    }

    #[test]
    fn visual_chinese_formatting() {
        // Create a buffer with border characters and Chinese text
        let area = Rect::new(0, 0, 14, 3);

        // Draw a box
        buffer.set_string(0, 0, "┌────────┐", Style::default());
        buffer.set_string(0, 1, "│        │", Style::default());
        buffer.set_string(0, 2, "└────────┘", Style::default());

        // Put Chinese characters inside the box
        buffer.set_string(2, 1, "你好世界", Style::default());

        // The entire buffer content should look correct with proper alignment
        let expected = [
            "┌────────┐",
            "│你好世界│", // 4 Chinese characters taking 8 spaces
            "└────────┘"
        ];

        // Verify each row
        for (i, row) in expected.iter().enumerate() {
            let mut line = String::new();
            for j in 0..area.width {
                line.push_str(buffer[(j, i as u16)].symbol());
            }
            assert_eq!(line, *row);
        }
    }
```

The test results are as follows:

```
---- buffer::buffer::tests::chinese_at_buffer_boundary stdout ----

thread 'buffer::buffer::tests::chinese_at_buffer_boundary' panicked at ratatui-core/src/buffer/buffer.rs:1396:9:
  left: "好"
 right: " "

---- buffer::buffer::tests::visual_chinese_formatting stdout ----

thread 'buffer::buffer::tests::visual_chinese_formatting' panicked at ratatui-core/src/buffer/buffer.rs:1449:13:
  left: "┌────────┐    "
 right: "┌────────┐"


failures:
    buffer::buffer::tests::chinese_at_buffer_boundary
    buffer::buffer::tests::visual_chinese_formatting
```

ZhangHanDong:
I'm adding more multi-byte related test cases to locate and identify issues. 

Screenshot 2025-04-07 at 11 12 18

I’ll send a new update if there’s a final fix.

 @orhun @joshka

joshka:
marking as a draft for now until it's resolved / questions are answered.

PR #1782, opened 2025-04-11
refactor(no_std): Make usages of std explicit in ratatui-core.

ed-2100:
### This commit does the following:

- Adds `#[no_std]` to `lib.rs`.
- Adds `extern crate std;` to `lib.rs`.
- Updates `ratatui-core` to explicitly `use` items from std and alloc.
- Prefers `use`-ing alloc over std when possible.

### Explanation:

This allows usages of `std` in `ratatui-core` to be clearly pointed out and dealt with individually.

Eventually, when `std` is to be feature gated, the associated commit will be much cleaner.


j-g00da:
Thanks!

orhun (approved):

joshka (approved):

joshka:
LGTM - thanks

PR #1791, opened 2025-04-15
Cow

cgzones:
Use a Copy-on-Write type for the contents of tables and lists so their content can be cached and only recomputed on actual changes.
This avoids unnecessary computations for the rendering in case of unchanged data.

Related to https://github.com/ratatui/ratatui/issues/1004.

Use a Copy-on-Write data type for the content of texts and lines to avoid unnecessary memory allocations.


joshka (commented):
This change adds extra lifetimes to several built-ins (text, line, span, etc). It seems to me like this would probably break almost every Ratatui application out there in a significant annoying way that would require every developer to perform manual fixes to their app to fix. This intuitively feels like too large an ask to reasonably take on.

It is possible that I've misunderstood this, I'd suggest an exercise to help show I'm wrong would be to extract just the text/span/line changes into a small commit (against the v0.29 tag if that's feasible), and then taking a reasonable sized app (e.g. yazi or something similar), and seeing what amount of effort it would take to apply this change to that app. I suspect this would take a non-trivial amount of hours.

ratatui-core/src/text/line.rs
```diff
@@ -179,18 +179,18 @@ use crate::widgets::Widget;
...
+pub struct Line<'lend, 'data> {
     /// The style of this line of text.
     pub style: Style,
 
     /// The alignment of this line of text.
     pub alignment: Option<Alignment>,
 
     /// The spans that make up this line of text.
-    pub spans: Vec<Span<'a>>,
+    pub spans: Cow<'lend, [Span<'data>]>,
```
Patryk27, line 190: `Cow` is covariant in `T`, so I don't think we need two lifetimes in here - e.g. this works as-is:
```rust
use std::borrow::Cow;

struct Line<'a> {
    spans: Cow<'a, [Span<'a>]>,
}

#[derive(Clone)]
struct Span<'a> {
    str: Cow<'a, str>,
}

fn line<'a>(s: &'a str) -> Line<'a> {
    Line {
        spans: Cow::Owned(vec![Span { str: Cow::Borrowed(s) }]),
    }
}
```

j-g00da:
Superseded by ratatui/ratatui#1900

PR #1795, opened 2025-04-17
feat(no_std)!: option to disable layout cache for `no_std` compatibility

j-g00da:
Resolves #1780 

BREAKING CHANGE: Disabling `default-features` will now disable layout cache, which can have a negative impact on performance. `Layout::init_cache` and `Layout::DEFAULT_CACHE_SIZE` are now only available if `layout-cache` feature is enabled.

### DIFF
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -84,6 +84,20 @@ This is a quick summary of the sections below:
 
 ## Unreleased (0.30.0)
 
+## Disabling `default-features` will now disable layout cache, which can have a negative impact on performance ([#1795])
+
+[#1795]: https://github.com/ratatui/ratatui/pull/1795
+
+Layout cache is now opt-in in `ratatui-core` and enabled by default in `ratatui`.
+If app doesn't make use of `no_std`-compatibility, and disables `default-feature`,
+it is recommended to explicitly re-enable layout cache.
+Not doing so may impact performance.
+
+```diff
+- ratatui = { version = "0.29.0", default-features = false }
++ ratatui = { version = "0.30.0", default-features = false, features = ["ratatui-core/thread-local-cache"] }
+```
+
 ## The MSRV is now 1.81.0 ([#1786])
 
 [#1786]: https://github.com/ratatui/ratatui/pull/1786
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -84,7 +84,7 @@ This is a quick summary of the sections below:
 
 ## Unreleased (0.30.0)
 
-## Disabling `default-features` will now disable layout cache, which can have a negative impact on performance ([#1795])
+### Disabling `default-features` will now disable layout cache, which can have a negative impact on performance ([#1795])
 
 [#1795]: https://github.com/ratatui/ratatui/pull/1795
 
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -19,6 +19,7 @@ This is a quick summary of the sections below:
   - `Backend` now requires an associated `Error` type and `clear_region` method
   - `Backend` now uses `Self::Error` for error handling instead of `std::io::Error`
   - `Terminal<B>` now uses `B::Error` for error handling instead of `std::io::Error`
+  - Disabling `default-features` will now disable layout cache, which can have a negative impact on performance
 - [v0.29.0](#v0290)
   - `Sparkline::data` takes `IntoIterator<Item = SparklineBar>` instead of `&[u64]` and is no longer const
   - Removed public fields from `Rect` iterators
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -88,10 +88,9 @@ This is a quick summary of the sections below:
 
 [#1795]: https://github.com/ratatui/ratatui/pull/1795
 
-Layout cache is now opt-in in `ratatui-core` and enabled by default in `ratatui`.
-If app doesn't make use of `no_std`-compatibility, and disables `default-feature`,
-it is recommended to explicitly re-enable layout cache.
-Not doing so may impact performance.
+Layout cache is now opt-in in `ratatui-core` and enabled by default in `ratatui`. If app doesn't
+make use of `no_std`-compatibility, and disables `default-feature`, it is recommended to explicitly
+re-enable layout cache. Not doing so may impact performance.
 
 ```diff
 - ratatui = { version = "0.29.0", default-features = false }
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -21,6 +21,7 @@ This is a quick summary of the sections below:
   - `Terminal<B>` now uses `B::Error` for error handling instead of `std::io::Error`
   - `TestBackend` now uses `core::convert::Infallible` for error handling instead of `std::io::Error`
   - Disabling `default-features` will now disable layout cache, which can have a negative impact on performance
+  - `Layout::init_cache` and `Layout::DEFAULT_CACHE_SIZE` are now only available if `layout-cache` feature is enabled
 - [v0.29.0](#v0290)
   - `Sparkline::data` takes `IntoIterator<Item = SparklineBar>` instead of `&[u64]` and is no longer const
   - Removed public fields from `Rect` iterators
@@ -85,6 +86,13 @@ This is a quick summary of the sections below:
 
 ## Unreleased (0.30.0)
 
+### `Layout::init_cache` and `Layout::DEFAULT_CACHE_SIZE` are now only available if `layout-cache` feature is enabled ([#1795])
+
+[#1795]: https://github.com/ratatui/ratatui/pull/1795
+
+Previously, `Layout::init_cache` and `Layout::DEFAULT_CACHE_SIZE` were available independently of
+enabled feature flags.
+
 ### Disabling `default-features` will now disable layout cache, which can have a negative impact on performance ([#1795])
 
 [#1795]: https://github.com/ratatui/ratatui/pull/1795
@@ -95,7 +103,7 @@ re-enable layout cache. Not doing so may impact performance.
 
 ```diff
 - ratatui = { version = "0.29.0", default-features = false }
-+ ratatui = { version = "0.30.0", default-features = false, features = ["ratatui-core/layout-cache"] }
++ ratatui = { version = "0.30.0", default-features = false, features = ["layout-cache"] }
 ```
 
 ### `TestBackend` now uses `core::convert::Infallible` for error handling instead of `std::io::Error` ([#1823])
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -95,7 +95,7 @@ Not doing so may impact performance.
 
 ```diff
 - ratatui = { version = "0.29.0", default-features = false }
-+ ratatui = { version = "0.30.0", default-features = false, features = ["ratatui-core/thread-local-cache"] }
++ ratatui = { version = "0.30.0", default-features = false, features = ["ratatui-core/layout-cache"] }
 ```
 
 ## The MSRV is now 1.81.0 ([#1786])
--- a/ratatui-core/Cargo.toml
+++ b/ratatui-core/Cargo.toml
@@ -59,6 +59,8 @@ thiserror = "2"
 unicode-segmentation.workspace = true
 unicode-truncate = "2"
 unicode-width.workspace = true
+once_cell = "1.21.3"
+critical-section = "1.2.0"
 
 [dev-dependencies]
 pretty_assertions.workspace = true
--- a/ratatui-core/Cargo.toml
+++ b/ratatui-core/Cargo.toml
@@ -27,8 +27,8 @@ default = []
 ## enables std
 std = []
 
-## enables layout cache based on thread-local LRU
-thread-local-cache = ["std"]
+## enables layout cache
+layout-cache = ["std"]
 
 ## enables conversions to / from colors, modifiers, and styles in the ['anstyle'] crate
 anstyle = ["dep:anstyle"]
--- a/ratatui-core/Cargo.toml
+++ b/ratatui-core/Cargo.toml
@@ -22,7 +22,13 @@ all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
 
 [features]
-default = []
+default = ["tls-layout-cache"]
+
+## enables std support
+std = []
+
+## enables layout cache based on thread-local storage
+tls-layout-cache = ["std"]
 
 ## enables conversions to / from colors, modifiers, and styles in the ['anstyle'] crate
 anstyle = ["dep:anstyle"]
@@ -46,12 +52,14 @@ serde = ["dep:serde", "bitflags/serde", "compact_str/serde"]
 anstyle = { version = "1", optional = true }
 bitflags = "2.3"
 compact_str = "0.9.0"
+critical-section = "1.2.0"
 document-features = { workspace = true, optional = true }
 hashbrown = "0.15.2"
 indoc.workspace = true
 itertools.workspace = true
 kasuari = "0.4.0"
 lru = "0.12.0"
+once_cell = "1.21.3"
 palette = { version = "0.7.6", optional = true }
 serde = { workspace = true, optional = true }
 strum.workspace = true
@@ -59,13 +67,12 @@ thiserror = "2"
 unicode-segmentation.workspace = true
 unicode-truncate = "2"
 unicode-width.workspace = true
-once_cell = "1.21.3"
-critical-section = "1.2.0"
 
 [dev-dependencies]
 pretty_assertions.workspace = true
 rstest.workspace = true
 serde_json.workspace = true
+critical-section = {version = "1.2.0", features = ["std"]}
 
 [lints]
 workspace = true
--- a/ratatui-core/Cargo.toml
+++ b/ratatui-core/Cargo.toml
@@ -69,10 +69,10 @@ unicode-truncate = "2"
 unicode-width.workspace = true
 
 [dev-dependencies]
+critical-section = { version = "1.2.0", features = ["std"] }
 pretty_assertions.workspace = true
 rstest.workspace = true
 serde_json.workspace = true
-critical-section = {version = "1.2.0", features = ["std"]}
 
 [lints]
 workspace = true
--- a/ratatui-core/Cargo.toml
+++ b/ratatui-core/Cargo.toml
@@ -22,13 +22,13 @@ all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
 
 [features]
-default = ["tls-layout-cache"]
+default = []
 
-## enables std support
+## enables std
 std = []
 
-## enables layout cache based on thread-local storage
-tls-layout-cache = ["std"]
+## enables layout cache based on thread-local LRU
+thread-local-cache = ["std"]
 
 ## enables conversions to / from colors, modifiers, and styles in the ['anstyle'] crate
 anstyle = ["dep:anstyle"]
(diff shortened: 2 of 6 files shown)


j-g00da:
I'm not really sure about the current implementation, I think it can be done some better way, I'm open to suggestions.

The problems I see with the current one:
1. Since we don't use TLS, Rects have to implement Send+Sync, that's why I had to change `Rc` to `Arc`, adding some overhead (should `tls-layout-cache` still compile using `Rc`? Probably.)
2. Two new dependencies are added that in fact are only needed when `tls-layout-cache` is _disabled_. Cargo doesn't support something like this, so to not include unnecessary dependencies, we would have to create a reverse `no_std` feature instead of `std` which would be an anti pattern, quoting The Cargo Book: 
> (...) features should be additive. That is, enabling a feature should not disable functionality, and it should usually be safe to enable any combination of features. A feature should not introduce a [SemVer-incompatible change](https://doc.rust-lang.org/cargo/reference/features.html#semver-compatibility).

> For example, if you want to optionally support [no_std](https://doc.rust-lang.org/reference/names/preludes.html#the-no_std-attribute) environments, do not use a no_std feature. Instead, use a std feature that enables std.

joshka:
> Since we don't use TLS, Rects have to implement Send+Sync, that's why I had to change Rc to Arc, adding some overhead (should tls-layout-cache still compile using Rc? Probably.)

Can you quantify the overhead of Arc vs Rc here? I'd anticipate it's likely fairly small (especially since the default cache size is 500 items).

One option to consider, which might be the simplest approach would be to not use the cache for no_std.

Alternatively, because no_std doesn't have threads (`std::thread`), perhaps a single static field without the thread local would work fine?

ratatui-core/Cargo.toml
```diff
@@ -22,7 +22,13 @@ all-features = true
...
 
 [features]
-default = []
+default = ["tls-layout-cache"]
+
+## enables std support
+std = []
+
+## enables layout cache based on thread-local storage
+tls-layout-cache = ["std"]
```
joshka, line 31: I'd prefer this not use `tls` as that intuitively means Transport Layer Security to many people rather than Thread Local Storage as intended here. So perhaps:
```suggestion
thread-local-cache = ["std"]
```

ratatui-core/Cargo.toml
```diff
@@ -46,12 +52,14 @@ serde = ["dep:serde", "bitflags/serde", "compact_str/serde"]
...
 bitflags = "2.3"
 compact_str = "0.9.0"
+critical-section = "1.2.0"
 document-features = { workspace = true, optional = true }
 hashbrown = "0.15.2"
 indoc.workspace = true
 itertools.workspace = true
 kasuari = "0.4.0"
 lru = "0.12.0"
+once_cell = "1.21.3"
```
joshka, line 62: core has OnceCell at https://doc.rust-lang.org/core/cell/struct.OnceCell.html, so I'd prefer avoiding this dep unless there's a rationale for it.

ratatui-core/src/layout/layout.rs
```diff
@@ -39,12 +43,39 @@ type Cache = LruCache<(Rect, Layout), (Segments, Spacers)>;
...
+    CsMutex::new(RefCell::new(LruCache::new(
+        NonZeroUsize::new(Layout::DEFAULT_CACHE_SIZE).unwrap(),
+    )))
+});
+
+fn with_layout_cache<F, R>(f: F) -> R
+where
+    F: FnOnce(&mut Cache) -> R,
+{
+    #[cfg(feature = "tls-layout-cache")]
```
joshka, line 64: Where the body of a method is entirely different given different feature flags, I prefer to put the flags on the method declaration and define it twice, this makes it easier to see when looking at a source file. Subjective opinion though. WDYT?

ratatui-core/src/layout/layout.rs
```diff
@@ -1,24 +1,28 @@
...
+use once_cell::sync::Lazy;
 
 use self::strengths::{
     ALL_SEGMENT_GROW, FILL_GROW, GROW, LENGTH_SIZE_EQ, MAX_SIZE_EQ, MAX_SIZE_LE, MIN_SIZE_EQ,
     MIN_SIZE_GE, PERCENTAGE_SIZE_EQ, RATIO_SIZE_EQ, SPACER_SIZE_EQ, SPACE_GROW,
 };
 use crate::layout::{Constraint, Direction, Flex, Margin, Rect};
 
-type Rects = Rc<[Rect]>;
+type Rects = Arc<[Rect]>;
```
joshka, line 25: This would be a breaking change as although the type isn't public (it's a private alias to simplify the code), the actual type appears in the public interface (`pub fn split(&self, area: Rect) -> Rects`), so if this is where we land, then it needs to be called out in the commit (easiest to put this in the PR body to make a squash merge work well) and in a BREAKING changes doc item. (Happy to defer this until we work out what the actual impl looks like though).

ratatui-core/src/layout/layout.rs
j-g00da: Sure, makes sense.

ratatui-core/Cargo.toml
joshka: Ah got it - makes sense. Given the proposed alternatives this might go away possibly still, but I see why it is necessary for this approach.

j-g00da:
> Can you quantify the overhead of Arc vs Rc here? I'd anticipate it's likely fairly small (especially since the default cache size is 500 items).

It is not much, but for a constrained embedded device it _might_ make some difference. (Don't have a proper way to test it yet.)


> One option to consider, which might be the simplest approach would be to not use the cache for no_std.

Thought about it, but I think it still might be beneficial. Ideally we would want both (no_std layout cache as opt-in) so one can decide between memory or speed optimization. On the other hand this might be added in the future.

> Alternatively, because no_std doesn't have threads (`std::thread`), perhaps a single static field without the thread local would work fine?

We would probably still have to use an unsafe block for that. Also the lack of `std::thread` doesn't imply that one can't create threads, that's why `Send` is in `core`. I wouldn't like to add unsafe block and instruct in docs to never use x in non-main thread (AFAIK there is no way to enforce that).

joshka:
> Thought about it, but I think it still might be beneficial. Ideally we would want both (no_std layout cache as opt-in) so one can decide between memory or speed optimization. On the other hand this might be added in the future.

What about getting it working without a cache, then getting an understanding of the necessity for a cache in real world apps with performance data. This might be the easiest way to get past the initial blocker that this is presenting. (Most of that blocker is in lack of familiarity with the replacement libraries necessary here, so not really having a good idea about whether they're the right ones or not).

As an aside, one of the things that I found confusing when initially learning tui-rs / ratatui was that the layout methods returned `Rc<[Rect]>` instead of a vec or array, and there wasn't an obvious reason for it. This meant certain iterators etc. didn't compile. Several iterators have since been added to the compiler and I obviously know a lot more about the details since that initial look a couple of years ago to understand this design, but I feel that this could still be something that trips up newer rust users in the same way as it did me back then.

j-g00da:
Sure, going with just no cache (and current TLS cache as opt-in) for now makes sense.

ratatui-core/src/layout/layout.rs
j-g00da: Anyway, despite deciding on using no cache for no_std for now, I would still want to leave this function, as this will make it easier to add more layout cache implementations in the future - just copy it with different cfg. WDYT @joshka @orhun?

ratatui-core/src/layout/layout.rs
```diff
@@ -281,7 +312,7 @@ impl Layout {
     ///
     /// By default, the cache size is [`Self::DEFAULT_CACHE_SIZE`].
     pub fn init_cache(cache_size: NonZeroUsize) {
-        LAYOUT_CACHE.with_borrow_mut(|c| c.resize(cache_size));
+        with_layout_cache(|c| c.resize(cache_size));
     }
```
j-g00da, line 316: If we want no layout cache for no_std, should this whole method not be compiled (breaking change) or just not take any effect (with appropriate docs update)? Same question for `DEFAULT_CACHE_SIZE`. @joshka @orhun

ratatui-core/src/layout/layout.rs
orhun: Sounds reasonable, as long as we document this function well and maybe add a note about this.

>  more layout cache implementations in the future

Not sure what else is _implementable_ out there though...

ratatui-core/src/layout/layout.rs
orhun: I'd go for NOP

ratatui-core/src/layout/layout.rs
```diff
@@ -653,11 +668,18 @@ impl Layout {
...
+        #[cfg(feature = "thread-local-cache")]
+        {
+            with_layout_cache(|c| {
+                let key = (area, self.clone());
+                c.get_or_insert(key, split).clone()
+            })
+        }
+
+        #[cfg(not(feature = "thread-local-cache"))]
+        split()
```
j-g00da, line 682: I didn't copy the whole method for cfg, as this would cause the docs to display only for the first implementation.

ratatui/Cargo.toml
```diff
@@ -25,9 +25,16 @@ rustdoc-args = ["--cfg", "docsrs"]
...
-default = ["crossterm", "underline-color", "all-widgets", "macros"]
+## which allows you to set the underline color of text, the `macros` feature which provides
+## some useful macros and `ratatui-core/thread-local-cache` which speeds up layout cache calculations
+## in `std`-enabled environments.
+default = [
+  "crossterm",
+  "underline-color",
+  "all-widgets",
+  "macros",
+  "ratatui-core/thread-local-cache",
```
j-g00da, line 36: I don't enable `thread-local-cache` in core by default but here instead, as this would cause third party widgets be not compatible with no_std by default.

With the current implementation, this is not a breaking change, but apps that disable `default-features` can experience performance drop. I think we should address that somehow?

ratatui-core/src/layout/layout.rs
```diff
@@ -1200,13 +1222,15 @@ mod tests {
...
             assert_eq!(c.cap().get(), Layout::DEFAULT_CACHE_SIZE);
         });
 
         Layout::init_cache(NonZeroUsize::new(10).unwrap());
-        LAYOUT_CACHE.with_borrow(|c| {
+
+        with_layout_cache(|c| {
             assert_eq!(c.cap().get(), 10);
         });
     }
```
j-g00da, line 1236: CI tests run with `default-features` disabled, so now they will run with no cache, how would you like me to handle this? Ideally we should run all tests with and without cache.

ratatui/Cargo.toml
j-g00da: Btw. please remove Breaking Change label.

orhun: Yeah... this is a bit tricky...

Maybe we should enable this in core instead and expose a feature flag here for disabling it? Can you list out the possible options with their trade offs for better evaluation of this case?

> but apps that disable default-features can experience performance drop.

I think that's a breaking change from the performance perspective 🫠

ratatui-core/src/layout/layout.rs
orhun: I think this will be addressed with your question below... but we can simply update our xtask config to do some [`cargo-hack`]-like testing logic here:

https://github.com/ratatui/ratatui/blob/2739391950eddd129c25cd4ffddb28b99eed8cf5/xtask/src/commands/backend.rs#L47-L65

Speaking of, maybe we should run [`cargo-hack`] in CI for not worrying about these feature-matrix issues in the future? 

[`cargo-hack`]: https://github.com/taiki-e/cargo-hack

ratatui-core/src/layout/layout.rs
orhun: fair!

ratatui/Cargo.toml
j-g00da: > expose a feature flag here for disabling it

I don't know if I understand correctly, do you mean an inversed feature flag?

> Can you list out the possible options with their trade offs for better evaluation of this case?

I think the current solution works the best and aligns better with what The Cargo Book suggests. If we moved it to core's default, creators of third-party widget crates would have to explicitly disable `default-features` to make the widgets compatible with `no_std`, and I think this compatibility should be the default. Especially since most widgets won't really need `std` (adding `no_std` compatibility was merely a few easy changes in `ratatui-widgets` and I can create a short guide for the website).

> I think that's a breaking change from the performance perspective 🫠

I can add a note to breaking changes if this solution is accepted.

ratatui-core/src/layout/layout.rs
j-g00da: Hmm this would be the best solution, the only question is which combinations of features would we want to test, as I don't see testing every single one - this would take insanely long.

orhun: Yup... this should be done in a separate issue/PR. We can _hack_ the xtask config in this PR if needed.

Created a tracking issue here: https://github.com/ratatui/ratatui/issues/1820

j-g00da: > We can hack the xtask config in this PR if needed.

Done

ratatui/Cargo.toml
orhun: As we have talked on Discord, I think this is fine 👍🏼

j-g00da:
Added a note to BREAKING-CHANGES.md, I think this is ready.

joshka (commented):
Overall I'm good with the approach. Some small things to get it over the line though.

ratatui-core/src/layout/layout.rs
```diff
@@ -279,9 +290,13 @@ impl Layout {
...
+    /// If no cache feature flag is enabled, method takes no effect.
+    /// Currently, the only available cache is `thread-local-cache`, which requires `std`.
+    ///
     /// By default, the cache size is [`Self::DEFAULT_CACHE_SIZE`].
-    pub fn init_cache(cache_size: NonZeroUsize) {
-        LAYOUT_CACHE.with_borrow_mut(|c| c.resize(cache_size));
+    pub fn init_cache(#[allow(unused_variables)] cache_size: NonZeroUsize) {
+        #[cfg(feature = "thread-local-cache")]
+        with_layout_cache(|c| c.resize(cache_size));
     }
```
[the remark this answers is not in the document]
j-g00da: I asked about it here https://github.com/ratatui/ratatui/pull/1795#discussion_r2072542251

Just want to make sure first if I should go with feature flagging it whole and adding a note to breaking changes.

ratatui-core/src/layout/layout.rs
```diff
@@ -188,6 +198,7 @@ impl Layout {
     /// bit more to make it a round number. This gives enough entries to store a layout for every
     /// row and every column, twice over, which should be enough for most apps. For those that need
     /// more, the cache size can be set with [`Layout::init_cache()`].
+    /// This const is unused if layout cache is disabled.
     pub const DEFAULT_CACHE_SIZE: usize = 500;
```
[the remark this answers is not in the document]
j-g00da: Same as below, https://github.com/ratatui/ratatui/pull/1795#discussion_r2072542251

Let me know if I should proceed with this change.

ratatui-core/Cargo.toml
```diff
@@ -24,6 +24,12 @@ rustdoc-args = ["--cfg", "docsrs"]
 [features]
 default = []
 
+## enables std
+std = []
+
+## enables layout cache based on thread-local LRU
+thread-local-cache = ["std"]
```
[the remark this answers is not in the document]
j-g00da: I left it that way (and same for the `with_layout_cache` function) in case we wanted to add alternative no_std cache in the future. I really think this can be helpful, especially after benchmarks you have shown on discord. MCUs are much slower so the difference will be noticable much sooner.

If we change this to just layout-cache, we would introduce a breaking change if we wanted to add a new cache in the future. The way it is now, it won't break anything.

As for the `with_layout_cache`, this will help introducing this future cache.

I want this discussed before I go with suggested changes.

WDYT?

ratatui-core/src/layout/layout.rs
```diff
@@ -39,12 +38,23 @@ type Cache = LruCache<(Rect, Layout), (Segments, Spacers)>;
...
+// This function should be copied with appropriate feature
+// for any future layout cache implementations.
+#[cfg(feature = "thread-local-cache")]
+fn with_layout_cache<F, R>(f: F) -> R
+where
+    F: FnOnce(&mut Cache) -> R,
+{
+    LAYOUT_CACHE.with_borrow_mut(f)
+}
+
```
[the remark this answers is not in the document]
j-g00da: https://github.com/ratatui/ratatui/pull/1795#discussion_r2076786483

ratatui/Cargo.toml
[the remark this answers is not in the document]
j-g00da: https://github.com/ratatui/ratatui/pull/1795#discussion_r2076786483

ratatui/Cargo.toml
```diff
@@ -25,9 +25,16 @@ rustdoc-args = ["--cfg", "docsrs"]
 #!
 ## By default, we enable the crossterm backend as this is a reasonable choice for most applications
 ## as it is supported on Linux/Mac/Windows systems. We also enable the `underline-color` feature
-## which allows you to set the underline color of text and the `macros` feature which provides
-## some useful macros.
-default = ["crossterm", "underline-color", "all-widgets", "macros"]
+## which allows you to set the underline color of text, the `macros` feature which provides
+## some useful macros and `ratatui-core/thread-local-cache` which speeds up layout cache calculations
```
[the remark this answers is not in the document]
j-g00da: https://github.com/ratatui/ratatui/pull/1795#discussion_r2076786483

BREAKING-CHANGES.md
```diff
@@ -19,6 +19,7 @@ This is a quick summary of the sections below:
   - `Backend` now requires an associated `Error` type and `clear_region` method
   - `Backend` now uses `Self::Error` for error handling instead of `std::io::Error`
   - `Terminal<B>` now uses `B::Error` for error handling instead of `std::io::Error`
+  - Disabling `default-features` will now disable layout cache, which can have a negative impact on performance
```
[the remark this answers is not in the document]
j-g00da: Oh I overlooked this in the previous PRs too, should I open a new PR for the previous entries that are missing details?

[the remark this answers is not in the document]
joshka: Thanks. Yeah, let's go with another PR for the breaking changes documentation to avoid holding this up.
Mostly the idea is to be able to make it easy to hit find on the doc and generally have a good relevant result.

ratatui-core/Cargo.toml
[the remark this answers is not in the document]
joshka: I think that the thread-local-ness of the cache is really an implementation detail rather than something consumers of the library should be concerned about. If we work out the no-std cache with something simple, then we can probably use the same feature flag rather than a different one. I tend to adhere to YAGNI for things which aren't well spec-ed.

ratatui-core/src/layout/layout.rs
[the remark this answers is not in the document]
joshka: The rationale for this is that it's prepping for something which may or may not come (or may be quite different). It's good generally to abstract away things which have a known reason to change, but I think this replaces one line of simple code with 9 without any obvious benefit. We can extract it if it makes sense when doing the no-std cache work.

ratatui-core/src/layout/layout.rs
joshka: I think I'd be inclined to make it breaking (feature flagged out). This makes anyone that is actively using the cache functionality and who disables default features break and have to update their code. This is a good break IMO as the behavior is changed and the compiler tells them this.

ratatui-core/src/layout/layout.rs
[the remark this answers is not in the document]
joshka: Same rationale as above.

ratatui-core/src/layout/layout.rs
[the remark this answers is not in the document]
joshka: Rationale above

ratatui-core/Cargo.toml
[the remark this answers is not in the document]
j-g00da: Ok! Makes sense.

orhun (approved):

j-g00da:
Ready, I also decided to add
`layout-cache = ["ratatui-core/layout-cache"]`
to `ratatui/Cargo.toml`, to make enabling/disabling the cache less awkward.

joshka (approved):

joshka:
Thanks for the effort in this PR.

PR #1797, opened 2025-04-17
feat(chart): support multi color line charts

jaysonmaw:
Added the ability to have line charts where values in specific ranges are drawn in different colors than the default. 

Needed this for my own project, and thought maybe it would be useful to someone else.

Example use:
```rust
    let datasets_cpu = vec![
            Dataset::default()
                .marker(symbols::Marker::Braille)
                .graph_type(GraphType::Line)
                .multi_color_line(
                    MultiColorLineConfig::default()
                        .add_color_range(0.0..3.0, Color::Red)
                        .add_color_range(3.0..6.0, Color::Yellow),
                )
                .style(Style::default().fg(Color::Green))
                .data(&temp_cpu_history),
        ];
```

Result:
image

## Limitations
Only used in GraphType::Line, mainly just because the other types you could already do this by just splitting up your data into different datasets based on values and then assign different colors there. The Line type needed custom code so that a continuous line could be drawn across boundaries while changing colors.


ratatui-widgets/src/chart.rs
```diff
@@ -170,6 +170,85 @@ pub enum GraphType {
     Bar,
 }
 
+/// Used to specify a range of values that will be drawn in another color.
+/// If the value does not fall inside any range specified the color will fallback to the style
+/// foreground color.
+///
+/// Has no effect in any [`GraphType`] other than [line](GraphType::Line)
```
orhun, line 177:
```suggestion
/// Used to specify a range of values that will be drawn in another color.
///
/// If the value does not fall inside any range specified the color will fallback to the style
/// foreground color.
///
/// Has no effect in any [`GraphType`] other than [line](GraphType::Line)
```

ratatui-widgets/src/chart.rs
```diff
@@ -170,6 +170,85 @@ pub enum GraphType {
...
+#[derive(Debug, Default, Clone, PartialEq)]
+pub struct MultiColorLineConfig {
```
orhun, line 179: I'd like to rename this to `MultiColorLine` since we don't usually use `Config` suffix with types.

```suggestion
pub struct MultiColorLine {
```

ratatui-widgets/src/chart.rs
```diff
@@ -170,6 +170,85 @@ pub enum GraphType {
...
+
+impl MultiColorLineConfig {
+    /// Add a range of values to be set to a specified Color
+    #[must_use = "method moves the value of self and returns the modified value"]
+    pub fn add_color_range(mut self, range: Range<f64>, color: Color) -> Self {
+        self.ranges.push(ColorRange { range, color });
+        self
+    }
+
+    /// Internal Helper function to be used during Widget render impl
```
orhun, line 191: Would be nice to explain what this function is doing instead of this

ratatui-widgets/src/chart.rs
```diff
@@ -395,6 +476,38 @@ impl<'a> Dataset<'a> {
...
+    /// ```rust
+    /// use ratatui::widgets::{Dataset, GraphType, MultiColorLineConfig};
+    /// use ratatui::style::{Color, Style, Stylize};
+    ///
+    /// let dataset = Dataset::default()
+    ///     .graph_type(GraphType::Line)
+    ///     .multi_color_line(
+    ///         MultiColorLineConfig::default()
+    ///             .add_color_range(0.0..1.0, Color::Green)
+    ///             .add_color_range(0.0..5.0, Color::Red) //Note: Ranges are searched in order of creation, meaning this range is also equivalent to 1.0..5.0
```
orhun, line 500: Interesting... we should probably note this in a separate `# Note` section.

ratatui-widgets/src/chart.rs
```diff
@@ -395,6 +476,38 @@ impl<'a> Dataset<'a> {
...
+    /// If the value does not fall inside any range specified the color will fallback to the style
+    /// foreground color.
+    ///
+    /// Has no effect in any [`GraphType`] other than [line](GraphType::Line)
+    ///
+    /// Ordering matters, ranges added first take precedence. (see example)
+    ///
+    /// This is a fluent setter method which must be chained or used as it consumes self
+    /// # Example
+    /// ```rust
```
orhun, line 491:
```suggestion
    /// This is a fluent setter method which must be chained or used as it consumes self
    ///
    /// # Example
    ///
    /// ```rust
```

orhun (commented):
Neat feature :)

jaysonmaw:
Implemented your suggested changes.
Also realized that the code was making the assumption that the ranges were contiguous, and that they were sorted/ascending as that was all I had ever needed for my own use case. Added extra branches to handle these cases.

jaysonmaw:
I ended up implementing the ability to switch the colors to be based on the x axis as mentioned in last comment, ended up being a fairly simple add. 

I'll add that as a commit as well, in case there is interest. Otherwise that commit can be dropped. 

Here's an example of what that looks like:
image

ratatui-widgets/src/chart.rs
```diff
@@ -170,6 +171,158 @@ pub enum GraphType {
...
+                } else if range.range.contains(&y2) {
+                    //line end is inside range, but not start
+                    let y3 = range.range.start;
+                    let x3 = ((y3 - y1) * (x2 - x1)) / (y2 - y1) + x1;
+                    draw(x3, y3, x2, y2, range.color);
+                    let y4 = range.range.start - f64::EPSILON;
+                    let x4 = ((y4 - y1) * (x2 - x1)) / (y2 - y1) + x1;
+                    x2 = x4;
+                    y2 = y4;
+                } else if (y1..=y2).contains(&range.range.start)
```
orhun, line 291: Hmm... the logic here gets quite complex... I think we should think about how to make this code simpler and more maintainable somehow...

jaysonmaw: Absolutely fair and I agree, I went ahead and refactored it a fair bit. I think the main `fn draw_lines_on_canvas_based_on_values()` where the majority of the logic was at is now much more clear/easier to follow. I was able to "hide" the main bit of trig away in its own function. I think a big part of it looking complex before was essentially that same trig function with slight tweaks being copied all over the place and that's no longer needed.

I'll get that refactor committed shortly, and I'll resolve this once I'm able to commit. (Should be in an hour or so, having to do some computer maintenance that's a blocker.)

jaysonmaw:
Sorry that took awhile to get that refactor out, in doing the refactor I noticed a small visual bug. I thought I had introduced it in the refactor, but couldn't find anything there. I thought it could be in the trig function so I rewrote that from scratch. There is still a slight visual bug that occasionally shows up where the color seems to bleed outside the range. I'm going to leave it for now, maybe its actually an artifact of how `ratatui` rendering works? Or maybe someone else can see my mistake if there is one? As for now I'm out of any other ideas on it.

jaysonmaw:
I did confirm that this is a rendering quirk of `ratatui`, believe its just a case of the terminal only having so much precision.
I added some debug assertions to panic on lines with cords outside the range bound being drawn in that color. (This did help me find one bug actually, but not the cause of the rendering bug)

Zooming in to the range boundary shows that its actually working as expected, and never draws across it:
image

While zooming out makes it much easier to happen:
image

I may end up porting the draw code here over to something with a full pixel buffer just to fully confirm / demonstrate, but I'm fairly confident that is the case.

jaysonmaw:
I ended up significantly simplifying the way this works, hopefully that also lessens the burden to get this accepted. 

I noticed the crate already had the [line-clipping crate ](https://docs.rs/line-clipping/0.3.0/line_clipping/) made by @joshka as a dependency, and that I could use that to simplify this a whole bunch. 

The result is the api for this new feature stays the exact same, though the way it works under the hood does change slightly.

The main thing that moving to this changes, is that you would want to recommended actually not setting a default foreground color, as if a default color is set it draws the entire line in that default color and then draws the color ranges on top, which can cause some small "z-fighting" visual artifacts. The best workaround for that is to just not set a default color, and then instead just make sure the ranges you set cover your entire data ranges. 

I feel like this is sort of the best of all worlds, it keeps the internal code simple, it doesn't add a dependency as it was already being included, and the very small visual "z-fighting" bug is easily worked around. 

Also just to point it out, you could get rid of the color range api using range syntax and instead just directly take in a ```Vec<line_clipping::Window>``` but then obviously the Window type becomes part of ratatui's public api which I wasn't sure if that would be wanted.

ratatui-widgets/src/chart.rs
```diff
@@ -395,6 +481,42 @@ impl<'a> Dataset<'a> {
...
+    ///     .multi_color_line(
+    ///         MultiColorLine::default()
+    ///             .add_color_range(0.0..1.0, Color::Green)
+    ///             .add_color_range(0.0..5.0, Color::Red), // equivalent to 1.0..5.0
+    ///     )
+    ///     .style(Style::default().cyan())
+    ///     .data(&[(0.0, 0.0), (1.0, 10.0)]);
+    /// ```
+    /// # Note
+    /// Ordering matters, ranges added first take precedence.
```
orhun, line 511: style nit:

```suggestion
    /// ```
    ///
    /// # Note
    ///
    /// Ordering matters, ranges added first take precedence.
```

ratatui-widgets/src/chart.rs
```diff
@@ -328,6 +412,8 @@ pub struct Dataset<'a> {
     graph_type: GraphType,
     /// Style used to plot this dataset
     style: Style,
+    /// If using [`GraphType::Line`] you can specify different color ranges by value
+    multi_color_line_config: Option<MultiColorLine>,
```
orhun, line 416: I think we can rename `multi_color_line_config` -> `multi_color_line` throughout this PR

orhun:
I'm also running into an interesting case with this.. I tried the example from the function docs:

image

Here you can see there are green dots where the red ones should be.

jaysonmaw:
Side note, not sure why that clippy lint in the CI started flagging as I didn't touch any of those files.

jaysonmaw:
Okay,

I have [this branch here](https://github.com/jaysonmaw/ratatui/tree/fix-z-fighting) which fixes the z-fighting issue by introducing line splitting again. 

One thing to note with it, is the ```create_point_that_ends_before_boundary``` helper function looks way more complicated than is should, just due to the current MSRV. Once that is bumped to ```1.86.0``` that gets dramatically simpler, and I added a TODO on that function to that effect.

orhun:
> Let me know if I should pull that extra branch into this pull request so you can do any reviews on that, if that's the direction we end up going.

Yup I think you can pull those changes here for a review

jaysonmaw:
> > Let me know if I should pull that extra branch into this pull request so you can do any reviews on that, if that's the direction we end up going.
> 
> Yup I think you can pull those changes here for a review

@orhun Sounds good, done 👍

joshka (commented):
Hey there, sorry it's taken me so long to get to this review. It came in while I was traveling without my laptop, and I figured it needed a bit more attention than I could give it on my phone.

At a high level:
- Doc comments need to be targeted at explaining the problem context a bit more. It's easy to forget when you're already deep in the solution that the person you're writing for hasn't gone on the same journey as you.
- I wonder if this would be better configured on the axis functions rather than on the graph itself?
- A simplification that makes lines always multi-colored (but just with the single default style) is to set the range as `[(f64::MIN..f64::MAX, None)]`
- Longer methods (>10-15 lines) should be shorter or have good docs about what each part of the method does to make it easier to maintain
- I think it might make sense to not make this exclusively GraphType::Line (could be a follow up PR). Scatter plots would benefit by not having to specify the style for each point. Bar charts would be rendered with each part of the bar being colored reasonably.

There's a bunch there and my delay in getting to this probably means that you're well past thinking about this for your own app, but I think my comments here push this in the direction of maximum utility for other user's apps. I hope that disconnect isn't too severe.

ratatui-widgets/src/chart.rs
```diff
@@ -170,6 +173,192 @@ pub enum GraphType {
     Bar,
 }
 
+/// Used to specify a range of values that will be drawn in another color.
+///
+/// If the value does not fall inside any range specified the color will fallback to the style
+/// foreground color.
+///
+/// Has no effect in any [`GraphType`] other than [line](GraphType::Line)
```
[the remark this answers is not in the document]
jaysonmaw: Updated Doc comment

> Note: this is an intentionally adversarial comment that looks at it from the perspective of someone that is unfamiliar with this code, while reading top to bottom in the PR. I'd assume that this is likely the worse case scenario for understanding this code and comments and most would be reading outside in encountering this from where it's used and then drilling into this type to understand more. The future comments in this PR all are from the same perspective.

I appreciate this being explicitly stated even if its generally assumed.

ratatui-widgets/src/chart.rs
```diff
@@ -170,6 +173,192 @@ pub enum GraphType {
...
+
+        let mut lines_to_draw = vec![line];
+        while let Some(line) = lines_to_draw.pop() {
+            for range in self.ranges.iter().chain(core::iter::once(&default_range)) {
+                let window = if self.x_axis {
+                    Window::new(range.range.start, range.range.end, f64::MIN, f64::MAX)
+                } else {
+                    Window::new(f64::MIN, f64::MAX, range.range.start, range.range.end)
+                };
+                if let Some(line_within_window) = clip_line(line, window) {
```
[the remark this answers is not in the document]
jaysonmaw: > overlapping color ranges

>I wonder if making this only store contiguous ranges and making later ranges take precendence over former. E.g. the unset / default range is effectively [f64::MIN..f64::MAX], adding 0.0..1.0 makes this [f64::MIN..0.0, 0.0..1.0, 1.0..f64::MAX], which removes the need to handle default range any differently from a specified range.

Covering both of these here, because even without overlapping ranges, that would only change this if you also did the default range system you mentioned in another comment.

I see what you are saying with having a default range that technically covers all ranges not covered by the user added ranges, I actually did have the same thought and had at one point had started implementing it like that. I ended up not going that direction though. 

From my view it ended up just ended up moving the complexity instead of actually getting rid of it. In the end the current version it just treats the default range the exact same as a user added range with the only difference being that it is always checked last.

I'm hoping my latest refactor / comments make it more obvious that the current implementation isn't actually doing anything complicated, as well we save on doing extra `clip_line` calls. 

> as well we save on doing extra `clip_line` calls. 

By this I mean, if there were added default ranges covering the in-between sections of each of the user ranges, we would be adding an extra calls to `clip_line` for each of those in-between sections, where as right now we essentially get those sections for "free" (minus a couple bit shifts)

>Concretely, if the ranges looks something like [0..1.0, 1.0..2.0, 3.0..4.0] (as an inclusive left, exclusive right range) then drawing a line from 0 to 5 would draw 0..1, 1..2 (overwriting the point at 1), 2..3 (overwriting the point at 2) etc.

Good catch on this, when changing to using `clip_line` instead of my own custom impl, I was no longer following the non-inclusive range semantics. Fixed that.

ratatui-widgets/src/chart.rs
```diff
@@ -395,6 +586,44 @@ impl<'a> Dataset<'a> {
...
+    ///     )
+    ///     .style(Style::default().cyan())
+    ///     .data(&[(0.0, 0.0), (1.0, 10.0)]);
+    /// ```
+    ///
+    /// # Note
+    ///
+    /// Ordering matters, ranges added first take precedence.
+    /// In practice this means that range `0.0..1.0` followed by `0.0..5.0` is equivalent to range
+    /// `0.0..1.0` followed by `1.0..5.0`
```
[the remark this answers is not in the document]
jaysonmaw: covered in other comment here https://github.com/ratatui/ratatui/pull/1797#discussion_r2096470768

ratatui-widgets/src/chart.rs
```diff
@@ -395,6 +586,44 @@ impl<'a> Dataset<'a> {
...
+    #[must_use = "method moves the value of self and returns the modified value"]
+    pub fn multi_color_line(mut self, multi_color_line: MultiColorLine) -> Self {
```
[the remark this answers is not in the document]
jaysonmaw: I don't believe so, just because if it was on the `Axis` struct it would then be possible to be "enabled" on both axis's and its unclear what you would do in this case. Keeping it as its own thing means we can enforce it only applying to one axis by them selecting the specific axis for it to apply to. 

Let me know if I missed something there though

jaysonmaw:
> Hey there, sorry it's taken me so long to get to this review. It came in while I was traveling without my laptop, and I figured it needed a bit more attention than I could give it on my phone.

No problem at all, has given me time to come back and look at it more anyways. Looking at it with fresh eyes has helped. 

> * Doc comments need to be targeted at explaining the problem context a bit more. It's easy to forget when you're already deep in the solution that the person you're writing for hasn't gone on the same journey as you.

Hopefully Covered

> * I wonder if this would be better configured on the axis functions rather than on the graph itself?

I don't believe so, just because if it was on the Axis struct it would then be possible to be "enabled" on both axis's and its unclear what you would do in this case. Keeping it as its own thing means we can enforce it only applying to one axis by them selecting the specific axis for it to apply to.

> * A simplification that makes lines always multi-colored (but just with the single default style) is to set the range as `[(f64::MIN..f64::MAX, None)]`

I did end up making the chart always multi colored, and with no color ranges set it will just fall back to the default color of either the foreground or Reset, which matched how it worked before. This means its no longer an `Option<MultiColorLine>` and instead we just always expect the `MultiColorLine` to do the drawing

> * Longer methods (>10-15 lines) should be shorter or have good docs about what each part of the method does to make it easier to maintain

Hopefully Covered

> * I think it might make sense to not make this exclusively GraphType::Line (could be a follow up PR). Scatter plots would benefit by not having to specify the style for each point. Bar charts would be rendered with each part of the bar being colored reasonably.

I agree, I had assumed adding to bar was going to be different as I assumed there was a `CanvasRect` struct (instead of `CanvasLine`) being used here, but since its just a `CanvasLine`, adding Bar support was trivial. Adding scatter support did require its own function, but its much simpler. 

To that end `MultiColorLine` and all of its functions were renamed to `MultiColorChart`

Here's some quick examples of the 3 chart types with the same data being drawn by this new system with a handful of ranges set
ratatui_example_line
ratatui_example_bar
ratatui_example_scatter


> There's a bunch there and my delay in getting to this probably means that you're well past thinking about this for your own app, but I think my comments here push this in the direction of maximum utility for other user's apps. I hope that disconnect isn't too severe.

Again no problem, my own project is done but I'm still happy to help get a cool feature to land for others to use. 


The only other thing I can think to add is that the impl for the scatter drawing multi color allocates more than I'd like and I was looking at it and realized something like a `PointsIter` would actually be really useful in this specific case, and could completely get rid of those extra allocations. I created a version of it just to test and it does work pretty well so I figured I'd just through it out there, but this PR is not really the place to actually get it added so more just something to think about. 

This code almost certainly is not perfect, just a proof of concept to show what I'm talking about. 
```rust
/// A group of points with a given color
#[derive(Default, Clone, PartialEq)]
pub struct PointsIter<I> {
    /// List of points to draw
    pub coords: core::cell::RefCell<I>,
    /// Color of the points
    pub color: Color,
}

impl<I> PointsIter<I>{
    pub fn new<T>(coords: T, color: Color) -> Self
    where
        T: IntoIterator<Item = (f64,f64),IntoIter = I>,
    {
        Self { coords: core::cell::RefCell::new(coords.into_iter()), color: color }
    }
}

impl<I> Shape for PointsIter<I>
where
    I: Iterator<Item = (f64,f64)>,
{
    fn draw(&self, painter: &mut Painter){
        let mut coords_iter = self.coords.borrow_mut();
        for (x, y) in &mut *coords_iter {
            if let Some((x, y)) = painter.get_point(x, y) {
                painter.paint(x, y, self.color);
            }
        }
    }
}
```

Thanks again for your help and the great review.

ratatui-widgets/src/chart.rs
```diff
@@ -170,6 +173,192 @@ pub enum GraphType {
...
+                    break;
+                }
+            }
+        }
+    }
+}
+
+/// A range of values to be drawn in the specified color
+#[derive(Debug, Default, Clone, PartialEq)]
+pub struct ColorRange {
```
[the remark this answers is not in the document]
joshka: Yeah, I tend to avoid tuples too, except for locally scoped things, or where the tuple make significant sense as a result type. This is in that narrow region where I think the struct is fine.

orhun:
We'll get back to this at some point. Sorry for the delay.

orhun:
Adding this to 0.31 milestone

PR #1801, opened 2025-04-20
chore: Upgrade to latest Rust version and to 2024 Rust edition

MatrixFrog:
Fixes https://github.com/ratatui/ratatui/issues/1727

This looks like a large PR but most of it is just formatting changes.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,8 +23,8 @@ categories = ["command-line-interface"]
 readme = "README.md"
 license = "MIT"
 exclude = ["assets/*", ".github", "Makefile.toml", "CONTRIBUTING.md", "*.log", "tags"]
-edition = "2021"
-rust-version = "1.81.0"
+edition = "2024"
+rust-version = "1.86.0"
 
 [workspace.dependencies]
 bitflags = "2.9.0"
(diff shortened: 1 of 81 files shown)


Cargo.toml
```diff
@@ -23,8 +23,8 @@ categories = ["command-line-interface"]
 readme = "README.md"
 license = "MIT"
 exclude = ["assets/*", ".github", "Makefile.toml", "CONTRIBUTING.md", "*.log", "tags"]
-edition = "2021"
-rust-version = "1.81.0"
+edition = "2024"
+rust-version = "1.86.0"
```
MatrixFrog, line 27: it looks like 1.87.0 will be out in a couple days

https://releases.rs/docs/1.87.0/

so I should be ready to send this for review by next weekend

joshka:
I really broke this merge a bunch - apologies. Attempting to fix.

joshka:
The migration approach for going from 2021 should generally be:
1. bump rust-version to 1.85 (not 1.86)
2. run `cargo fix --edition`
3. check the various fixes that are introduced and work out which are necessary. Some may be relevant, some may not, but we do want to document the choices we make regardless. There's some complexity in suggested fixes that would take a bit more effort than usual to dig into and understand.
4. update edition to 2024
5. run `cargo xtask fmt`

This PR skipped a bit of the process by jumping directly to step 4, so we need to redo that approach rather than merging this as is.

Additionally, a bunch of stuff moved around in the intervening time since it was raised. This PR would mostly be a full replacement. Closing it out for now rather than attempting to rebase / merge / etc.

PR #1814, opened 2025-04-27
feat(rect): centering

TadoTheMiner:
Resolves #617

### DIFF
--- a/ratatui-core/src/layout/rect.rs
+++ b/ratatui-core/src/layout/rect.rs
@@ -393,7 +393,9 @@ impl Rect {
     /// use ratatui_core::layout::Constraint;
     /// use ratatui_core::terminal::Frame;
     /// fn render(frame: &mut Frame) {
-    ///     let area = frame.area().centered(Constraint::Ratio(1, 2), Constraint::Ratio(1, 3));
+    ///     let area = frame
+    ///         .area()
+    ///         .centered(Constraint::Ratio(1, 2), Constraint::Ratio(1, 3));
     /// }
     /// ```
     #[must_use]
--- a/ratatui-core/src/layout/rect.rs
+++ b/ratatui-core/src/layout/rect.rs
@@ -7,6 +7,8 @@ use crate::layout::{Margin, Position, Size};
 mod iter;
 pub use iter::*;
 
+use super::{Constraint, Flex, Layout};
+
 /// A Rectangular area.
 ///
 /// A simple rectangle used in the computation of the layout and to give widgets a hint about the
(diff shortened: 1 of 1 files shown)


orhun (commented):
Hmm this PR seems like it needs some cleanup, what's up with the Nix stuff :D

.envrc
joshka, line 1: Remove (from this PR)

.gitignore
joshka, line 1: Remove

flake.lock
joshka, line 1: Remove

flake.nix
joshka, line 1: Remove

ratatui-core/src/layout/rect.rs
```diff
@@ -351,6 +353,35 @@ impl Rect {
         }
     }
 
+    /// Centers the rect horizontally. The constraint indicates the result rects width.
```
joshka, line 356: The usual rust (and often other languages) convention for doc comments is simple single sentence as the first line. This appears anywhere where a short description is needed. The detail generally is separated by a newline and then more information that talks about how this works.

```suggestion
    /// Returns a new Rect, centered horizontally based on the provided constraint.
```

It would be good to add a small example of use here. E.g. something like:

```rust
fn render(frame &mut Frame)
    let area = frame.area().centered_horizontally(Constraint::Ratio(1, 2));
}
```

(This comment also applies to the other methods below)

ratatui-core/src/layout/rect.rs
```diff
@@ -351,6 +353,35 @@ impl Rect {
...
+    #[must_use]
+    pub fn center_horizontally(self, constraint: Constraint) -> Self {
```
joshka, line 358: Bikeshedding the name. I think I prefer `centered_` over `center_` as an adjective generally suggests that a new item that is modified will be returned. `center` is ambiguously either a noun or a verb, which is problematic. It's difficult to intuit whether it modifies the existing item or returns a new item. Does this make sense? (It's possible that I'm wrong here if there's other existing code examples that would use this name like this. I think this is unlikely, but only at around 90% confidence).

ratatui-core/src/layout/rect.rs
```diff
@@ -662,4 +694,23 @@ mod tests {
...
     }
+
+    #[test]
+    fn center() {
+        let rect = Rect {
+            x: 0,
+            y: 0,
+            width: 5,
+            height: 5,
+        };
```
joshka, line 705: Generally try to make unit tests not take up too much vertical space as when there's lots (as there should be), vertical space causes fatigue when reading. Use Rect::new() here instead.

ratatui-core/src/layout/rect.rs
```diff
@@ -662,4 +694,23 @@ mod tests {
             }
         );
     }
+
+    #[test]
+    fn center() {
```
joshka, line 699: Add tests for the other two methods

joshka (commented):
Generally this is a good idea I think. I like the idea of having this as a method on Rect like this. Some comments about naming are the main thing to think about.

TadoTheMiner:
Those files are required for NixOS users. It could be bypassed by using a devshell from a different project but it is more convienent to have it here as everytime I switch to the directory i automatically enter it. If you really don't want it there, is there something like .gitignore but the file itself is not tracked by git?

orhun:
> If you really don't want it there, is there something like .gitignore but the file itself is not tracked by git?

It's `.git/info/exclude`

joshka:
> Those files are required for NixOS users. It could be bypassed by using a devshell from a different project but it is more convienent to have it here as everytime I switch to the directory i automatically enter it. If you really don't want it there, is there something like .gitignore but the file itself is not tracked by git?

Those files are required for your use case - using nix, but none of the core devs on ratatui are using nix (at least last I checked), and the files aren't required for this PR. A good general rule is to do one thing in a PR at a time. We can certainly have the discussion about nix and do something about it perhaps, but it shouldn't be part of the discussion about centering rects. There might already be an issue about it (I'm not sure if it's in this issue or another ratatui related one).

When you're adding files to the git index to become part of a commit, you can choose which files are included and which are excluded at that time. You don't need to add all files. See https://git-scm.com/docs/git-add for more info. If you're using a GUI or TUI, they should generally support commits that don't include all files. Take a look at https://sukima.github.io/GitFixUm/ for more info on fixing up git commits in various ways (this might make a good TUI / CLI app...).

You can also setup your own personal ignore files either at a user level or a repo level by following the config options mentioned in https://git-scm.com/docs/gitignore#_description (e.g. $GIT_DIR/info/exclude probably makes sense here).

ratatui-core/src/layout/rect.rs
TadoTheMiner: Okay we should do this for other tests aswell though

TadoTheMiner:
Cargo fmt --all doesn't change anything, why does the check fail?

joshka:
> Cargo fmt --all doesn't change anything, why does the check fail?

We use a bunch of unstable options in rustfmt.toml, for which you'll need a nightly rust toolchain installed.
Run `cargo xtask fmt` (or `cargo +nightly fmt`) to format the code.

It's also possible that if this is what you're doing already then something changed in the nightly version. Most of the time this is transient, but it does happen. Compare the version in the CI failure with your local version in this case.

TadoTheMiner:
I struggle with setting it up so I have fmt nightly and everything else stable without using rustup. Could you please format it for me?

ratatui-core/src/layout/rect.rs
```diff
@@ -351,6 +353,61 @@ impl Rect {
         }
     }
 
+    /// Returns a new Rect, centered horizontally based on the provided constraint.
+    /// # Examples
```
orhun, line 357: nit:

```suggestion
    /// Returns a new Rect, centered horizontally based on the provided constraint.
    ///
    /// # Examples
```

(applies for other doc comments as well)

joshka (approved):

joshka:
I resolved the small formatting nits and merged this. Thanks for the PR :)

PR #1817, opened 2025-05-01
feat(error)!: introduce `ratatui_core::error::Error`

j-g00da:
Follow up to #1778
Introduce `ratatui_core::error::Error` and require backends to implement `From<ratatui_core::error::Error>` for associated `Error` type.

WIP, the problem I'm facing:

```rust
pub trait Backend {
    type Error: core::error::Error + Into<crate::error::Error>;
}
```
This does _not_ make type parameters `B: Backend` 
satisfy `crate::error::Error: From<B as Backend>::Error` bound.

I also tried:
```rust
pub trait Backend 
where
    Self::Error: core::error::Error,
    crate::error::Error: From<Self::Error>,
{
    type Error;
}
```

But this still produces the same problem:
```rust
impl<B> Terminal<B>
where
    B: Backend,
{
...
```
> the trait bound `error::Error: From<<B as Backend>::Error>` is not satisfied [E0277]
> the trait `From<<B as Backend>::Error>` is not implemented for `error::Error`
> Note: required by a bound in `Backend`

Generally this can be fixed by adding another bound to where clause, everywhere where `B: Backend` type param is used, but this doesn't feel right. I don't know if this is a compiler limitation, but I can't find a way to make it work without it.

I'm open to suggestions.

### DIFF
--- a/examples/apps/demo/src/crossterm.rs
+++ b/examples/apps/demo/src/crossterm.rs
@@ -43,14 +43,11 @@ pub fn run(tick_rate: Duration, enhanced_graphics: bool) -> Result<(), Box<dyn E
     Ok(())
 }
 
-fn run_app<B: Backend>(
-    terminal: &mut Terminal<B>,
+fn run_app<W: io::Write>(
+    terminal: &mut Terminal<CrosstermBackend<W>>,
     mut app: App,
     tick_rate: Duration,
-) -> Result<(), Box<dyn Error>>
-where
-    B::Error: 'static,
-{
+) -> Result<(), Box<dyn Error>> {
     let mut last_tick = Instant::now();
     loop {
         terminal.draw(|frame| ui::draw(frame, &mut app))?;
--- a/examples/apps/demo/src/termion.rs
+++ b/examples/apps/demo/src/termion.rs
@@ -32,14 +32,11 @@ pub fn run(tick_rate: Duration, enhanced_graphics: bool) -> Result<(), Box<dyn E
     Ok(())
 }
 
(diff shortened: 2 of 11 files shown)


aschey:
Type bounds aren't always propagated in the way you would expect, unfortunately. This might be an instance of [this issue](https://github.com/rust-lang/rust/issues/20671). There are some workarounds listed in that thread, but I'm not sure if any of them are worth the complexity.

Considering most use cases will probably involve handling errors coming from a `Terminal` instance rather than the specific backend, perhaps it would be cleaner to remove the backend-specific error type and have each backend return `ratatui_core::error::Error` instead? For scenarios like this, I often see a shared error type that enumerates common scenarios that can be used by most/all backend types, and then provides a `Custom` variant that backends can use to express something that doesn't fall into one of the standard categories. `std::io::Error` follows this pattern.

j-g00da:
I'm having second thoughts if we need ratatui Error at all. I think we should discuss pros and cons of each solution.
What is currently on main has a pretty big advantage of letting built-in backends still use `std::io::Error`, which won't break most existing apps. Additionally a lot of errors might be highly backend-specific and IMO it's better to let backends define their own errors instead of putting everything to `Other` variant. So my solution would be to just close this PR and leave it as it is. WDYT?

cc: @orhun @joshka

orhun:
Sounds reasonable to me ngl

orhun:
Any verdict on this? @j-g00da

j-g00da:
> Any verdict on this? @j-g00da 

We are not implementing this.

PR #1818, opened 2025-05-04
add `ratatui-console` backend

JeromeSchmied:
create a `Backend` for `console`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -576,6 +576,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "console"
+version = "0.15.11"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


JeromeSchmied:
I guess the implementation from my side is quite complete, but there are many changes to be made in `console` to get this working. dunno whether it's worth the effort, I'll keep going though. see `unimplemented!()`-s in ratatui-console/src/lib.rs

JeromeSchmied:
I'm afraid making it to the end would require an insane amount of work on `console` as it was simply designed for completely different things, and therefore lacks many many features a `ratatui` integration would require, eg: persistent raw mode. see todos in previous comment, these features generally wouldn't be difficult to implement, but being so many and not having basic functionality to build upon makes the whole backend not worth it.

orhun:
Thanks for looking into this and giving it a shot, it's definitely very valuable!

I'll keep an eye on those console issues

PR #1827, opened 2025-05-08
feat(rect): add flex functions

### LINKED ISSUE #1826: Rect flex functions
## Problem

The centered function saves you from doing the layout thing with centering but it doesn't save you when you don't want to center but to say flex to the right
## Solution

flex_bottom flex_top flex_left flex_right functions on rect, each of them taking a constraint, and behaving just like their centering counterparts except that they use their corresponding flex parameters

TadoTheMiner:
Resolve #1826

### DIFF
--- a/ratatui-core/src/layout/rect.rs
+++ b/ratatui-core/src/layout/rect.rs
@@ -417,6 +417,80 @@ impl Rect {
             .centered_vertically(vertical_constraint)
     }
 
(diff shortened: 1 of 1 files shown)


orhun:
I think we need to decide things in #1826 (https://github.com/ratatui/ratatui/discussions/1879) before implementing this

joshka:
Marking this as a draft until the discussing is resolved

j-g00da:
Closing until the discussion is resolved.

PR #1828, opened 2025-05-08
chore(breaking-changes): add details to `no_std`-related breaking changes

j-g00da:
Some corrections and added details to BREAKING-CHANGES.md.

I decided to remove:

  - `Backend` now uses `Self::Error` for error handling instead of `std::io::Error`
  - `Terminal<B>` now uses `B::Error` for error handling instead of `std::io::Error`

...as we are still using `std::io::Error` in built-in backends, so this will only be breaking if a third-party backend decides to use a custom error other than `std::io::Error`, which would be a breaking change in downstream and not `ratatui`.

The exception to that is `TestBackend`, which uses `Infallible`, but this already has its own breaking changes entry.

### DIFF
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -22,6 +22,7 @@ This is a quick summary of the sections below:
   - `TestBackend` now uses `core::convert::Infallible` for error handling instead of `std::io::Error`
   - Disabling `default-features` will now disable layout cache, which can have a negative impact on performance
   - `Layout::init_cache` and `Layout::DEFAULT_CACHE_SIZE` are now only available if `layout-cache` feature is enabled
+  - Disabling `default-features` suppresses the error message if `show_cursor()` fails when dropping `Terminal`
 - [v0.29.0](#v0290)
   - `Sparkline::data` takes `IntoIterator<Item = SparklineBar>` instead of `&[u64]` and is no longer const
   - Removed public fields from `Rect` iterators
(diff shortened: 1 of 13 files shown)


j-g00da:
Do you think, the other entries in the breaking changes cover this sufficiently?
I'm not 100% sure about that, since the fact that third party backends _may_ change the return type of methods in `Terminal`, might still be considered a breaking change on our side...?
WDYT, should I revert removing these two?

orhun:
I'm missing the context on why the default implementation on `clear_region` was removed. The other changes look reasonable to me.

> since the fact that third party backends may change the return type of methods in Terminal, might still be considered a breaking change on our side...?

Not sure about those breakages, I need to switch to ratatui `main` for Ratzilla and see it for myself.

j-g00da:
> I'm missing the context on why the default implementation on `clear_region` was removed. The other changes look reasonable to me.

It used `std::io::Error` explicitly, this should be handled by concrete backend implementations.

BREAKING-CHANGES.md
```diff
@@ -121,6 +119,13 @@ re-enable layout cache. Not doing so may impact performance.
 Since `TestBackend` never fails, it now uses `Infallible` as associated `Error`. This may require
 changes in test cases that use `TestBackend`.
 
+### `Backend` now requires an associated `Error` type and `clear_region` method ([#1778])
+
+[#1778]: https://github.com/ratatui/ratatui/pull/1778
+
+Custom `Backend` implementations must now define an associated `Error` type for method `Result`s
+and implement the `clear_region` method, which no longer has a default implementation.
+
```
orhun, line 128: Maybe we can mention why this change has been made as well (i.e. we can say "for no_std" somewhere)

j-g00da: Added:

This change was made to provide greater flexibility for custom backends, particularly to remove the explicit dependency on `std::io` for backends that want to support `no_std` targets.

orhun (approved):

PR #1833, opened 2025-05-09
fix(no_std): final patches needed for no_std

j-g00da:
NOT FOR MERGE

This is FYI only, I will cut this into smaller PRs.
Line-clipping uses git here as it's not released yet.


joshka:
I think this can be closed as it got chopped into the other PRs right?

j-g00da:
yup, this was just a POC branch

PR #1835, opened 2025-05-09
chore: move time to dev-dependencies

j-g00da:
Since `time` is only used in `ratatui` in tests, I believe it should be moved to dev-dependencies.

### DIFF
--- a/ratatui/tests/widgets_calendar.rs
+++ b/ratatui/tests/widgets_calendar.rs
@@ -1,4 +1,4 @@
-#![cfg(feature = "widget-calendar")]
+#![cfg(test)]
 use ratatui::backend::TestBackend;
 use ratatui::buffer::Buffer;
 use ratatui::style::Style;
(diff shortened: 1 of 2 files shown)


ratatui/tests/widgets_calendar.rs
```diff
@@ -1,4 +1,4 @@
-#![cfg(feature = "widget-calendar")]
+#![cfg(test)]
```
joshka, line 1: Is this necessary? Doesn't this get defined by default if you're compiling files in test?

joshka (approved):
Approved subject to question

fivemhostingserver:
Great implementation of this feature. I particularly like the approach to error handling.

ratatui/tests/widgets_calendar.rs
j-g00da: Right

orhun (approved):

joshka (approved):

PR #1837, opened 2025-05-10
fix: `padding_right()` should set right padding instead of left

sxyazi:
Fixes https://github.com/ratatui/ratatui/issues/1836


joshka (approved):

joshka (commented):
Would you mind adding two tests (for left and right)?

sxyazi:
Done!

joshka (approved):
LGTM - thanks again

PR #1838, opened 2025-05-10
fix(no_std): propagate `std` feature flag to dependencies

j-g00da:
Disables `std` feature flags in dependencies and only enables them with `ratatui` and `ratatui-core`'s `std` feature flag.
~~Additionally bumps version of `kasuari`~~ _(moved to #1844)_, to get upstream changes that allow disabling dependency on `std`.
This partially fixes the issue of still depending on `std`, when `std` feature flag is disabled.


j-g00da:
This requires #1840 to compile properly (that's why tests fail).

joshka (approved):

PR #1840, opened 2025-05-10
fix(no_std): provide `f64` polyfills for `no_std` compatibility

j-g00da:
Since a lot of float methods are not included in `core`, I decided to make a simple polyfills crate, so we don't have to conditionally compile every place where these methods are used.

This adds a dependency, that does nothing if we use `std`. We can either:
1. Leave as it is, with additional dependency.
2. If we don't want to add a dependendy by default, we can introduce a `std-polyfills` (or `no-std`) feature flag (_additional_ to `std`). I did something like that here: #1833.
This has one drawback - we _have to_ use either `std` or `std-polyfills` feature flag or the library won't compile.
To make it easier on the user, we can add a check in lib.rs that emitts compiler error "either std or std-polyfills feature flag must be enabled" in such cases.
3. I can try to make the `float-polyfills` crate lighter, not depending on libm (or gate libm behind feature flag).

Let me know what you think and if you have some other ideas for this.

Related: https://github.com/rust-lang/rust/issues/137578


j-g00da:
Added these directly to core and widgets instead of adding a dependency as discussed on discord.

The uncovered lines are not important, it would be hard to test the polyfills trait implications directly as std shadows it (and we need it to test against something), so I put the implementations in standalone functions and test these instead.

ratatui-widgets/src/polyfills.rs
```diff
@@ -0,0 +1,157 @@
...
+    fn floor(self) -> f64;
+
+    /// Approximates the sine of a number (in radians) with max error of `0.002`.
+    fn sin(self) -> f64;
+
+    /// Approximates the cosine of a number (in radians) with max error of `0.002`.
+    fn cos(self) -> f64;
+}
+
+impl F64Polyfills for f64 {
```
orhun, line 66: Seems like there are some warnings in GitHub interface about " not covered by tests", are we sure we are running the necessary tests (via `cargo-xtask`)?

ratatui-widgets/src/polyfills.rs
```diff
@@ -0,0 +1,157 @@
...
+fn cos(val: f64) -> f64 {
+    let mut x = val;
+    x *= FRAC_1_PI / 2.0;
+    x -= 0.25 + floor(x + 0.25);
+    x *= 16.0 * (x.abs() - 0.5);
+    x += 0.225 * x * (x.abs() - 1.0);
+    x
+}
+
+#[allow(dead_code)]
```
orhun, line 47: Why is this necessary?

ratatui-widgets/src/polyfills.rs
```diff
@@ -0,0 +1,157 @@
+//! Polyfills providing pure rust fallback implementations of various `f64` methods for `no_std`
+//! compatibility. These implementations may not be as accurate as their built-in counterparts.
+//! This must be taken into account when using floating point math in this crate.
+//!
+//! Implementations based on `micromath` crate:
+//! <https://github.com/tarcieri/micromath>
+//!
+//! Related rust tracking issues:
+//! <https://github.com/rust-lang/rust/issues/50145>
+//! <https://github.com/rust-lang/rust/issues/137578>
```
orhun, line 10:
```suggestion
//! Implementations based on [`micromath`](https://github.com/tarcieri/micromath) crate.
//!
//! Related Rust tracking issues:
//! 
//! - <https://github.com/rust-lang/rust/issues/50145>
//! - <https://github.com/rust-lang/rust/issues/137578>
```

ratatui-widgets/src/polyfills.rs
j-g00da: Compiler thinks this is never used, as all the methods are shadowed by std. This is something we actually want as this gives us nice fallback in `no_std`, and in `std` the imports of this trait just do nothing.

ratatui-widgets/src/polyfills.rs
j-g00da: https://github.com/ratatui/ratatui/pull/1840#issuecomment-2869681518

ratatui-widgets/src/polyfills.rs
orhun: Can we maybe use feature flags instead of allowing lints such as this one and `unused_imports`? I see that it is convenient but it's a bit hard to derive that from the code especially when there is not comments why it is `dead_code` etc.

j-g00da: I'm pretty sure we would still need these despite conditionally compiling it as this is a compiler quirk, but I will check this later.

Alternatively I can add a comment why this is needed.

j-g00da: Yuup, I'm getting same `never used` and `unused import` errors even with `#[cfg(not(feature = "std"))]`.
I will probably open an issue upstream if this isn't already reported.

j-g00da: Added a clarification comment to this.

j-g00da: FYI, opened issue for clippy, let's see what they say. Maybe I'm missing something.
https://github.com/rust-lang/rust-clippy/issues/14795

j-g00da: Ok I was wrong, managed to fix that.

joshka (approved):
LGTM. It's possible that a few of these methods will no longer be necessary after some future work on the circle algorithms to avoid trig functions, but that can happen at a later date. For now this works to get over the no_std line.

ratatui-widgets/src/polyfills.rs
joshka: I'm fine with merging this with those warnings

Issue #1820, opened 2025-05-05
Consider integrating `cargo-hack` for testing crate features


orhun:
## Problem

As we add more feature flags to the workspace crates, it gets difficult to tweak the `cargo-xtask` configuration accordingly and make sure all the necessary features are activated while running the tests.

## Solution

[`cargo-hack`](https://github.com/taiki-e/cargo-hack) is a cargo plugin that supports running cargo commands with combination of features, individual features and so on.

We might consider integrating that into our current `cargo-xtask` set up. 

## Additional context

See the discussion in https://github.com/ratatui/ratatui/pull/1795#discussion_r2073107212

joshka:
Yup - I think it was mentioned before as a thing also.

PR #1841, opened 2025-05-10
feat(crossterm): allow multiple crossterm versions

joshka:
This commit introduces feature flags to make it possible for widget
library authors to depend on a specific version of crossterm without
causing version conflicts. This should make it easier for libraries and 
apps to update crossterm versions more easily.

The available feature flags are `crossterm_0_28` and `crossterm_0_29`.
By default, the latest version is enabled. If a multiple features are
enabled we choose the latest version. We will in general support at
least the last two major (0.x) versions of crossterm, and will only
remove versions in a major version bump.


ratatui-crossterm/Cargo.toml
```diff
@@ -14,11 +14,31 @@ exclude.workspace = true
...
+#! enabled. This will always default to the most recent version of crossterm, but you can override
+#! this by specifying the version you want to use in your Cargo.toml file. The purpose of these
+#! features is to allow widget libraries to depend on specific versions of crossterm in a way that
+#! doesn't cause version conflicts. A compiler error will be raised if you try to use multiple
+#! versions of crossterm at the same time.
+
+## Enables crossterm 0.28.x
+crossterm_0_28 = ["dep:crossterm_0_28"]
+## Enables crossterm 0.29.x
+crossterm_0_29 = ["dep:crossterm_0_29"]
```
orhun, line 34: It might get a bit annoying to add a new feature flag for the each new version of crossterm, even the patches. Or are you thinking of only doing this for the minor versions?

Are we fine with that? If they decide to speed up their release cycle we might end up with so many feature flags. Maybe we can let the upstream know that we are doing this and hear what they think?

ratatui-crossterm/src/lib.rs
```diff
@@ -6,16 +6,40 @@
...
 //!
+//! ## Crossterm Version and Re-export
+//!
+//! `ratatui-crossterm` requires you to specify a version of the [Crossterm] library to be used.
+//! This is managed via feature flags. You **must** enable one, and only one, of the available
+//! `crossterm_xx` features (e.g., `crossterm_28`, `crossterm_29`). These features determine which
+//! version of Crossterm is compiled and used by the backend. Attempting to enable none or multiple
+//! `crossterm_xx` features will result in a compile-time error.
+//!
+//! To promote interoperability within the [Ratatui] ecosystem, the selected Crossterm crate is
```
orhun, line 20:
```suggestion
//! To promote interoperability within the [Ratatui] ecosystem, the selected [Crossterm] crate is
```

ratatui-crossterm/src/lib.rs
```diff
@@ -6,16 +6,40 @@
...
-//! the [Crossterm] crate to interact with the terminal.
+//! This crate provides [`CrosstermBackend`], an implementation of the [`Backend`] trait for the
+//! [Ratatui] library. It uses the [Crossterm] library for all terminal manipulation.
 //!
+//! ## Crossterm Version and Re-export
+//!
+//! `ratatui-crossterm` requires you to specify a version of the [Crossterm] library to be used.
+//! This is managed via feature flags. You **must** enable one, and only one, of the available
+//! `crossterm_xx` features (e.g., `crossterm_28`, `crossterm_29`). These features determine which
+//! version of Crossterm is compiled and used by the backend. Attempting to enable none or multiple
```
orhun, line 17:
```suggestion
//! version of [Crossterm] is compiled and used by the backend. Attempting to enable none or multiple
```

ratatui-crossterm/src/lib.rs
```diff
@@ -6,16 +6,40 @@
...
+//! `crossterm_xx` features will result in a compile-time error.
+//!
+//! To promote interoperability within the [Ratatui] ecosystem, the selected Crossterm crate is
+//! re-exported as `ratatui_crossterm::crossterm`. This re-export is essential for authors of widget
+//! libraries or any applications that need to perform direct Crossterm operations while ensuring
```
orhun, line 22:
```suggestion
//! libraries or any applications that need to perform direct [Crossterm] operations while ensuring
```

ratatui-crossterm/Cargo.toml
joshka: It's only necessary for semver incompatible versions (0.x -> 0.x+1). Patch version doesn't matter.
It's also fairly unlikely that crossterm will speed up development IMO. Either way, we can roll with it as needed.
Perhaps we support last 2 versions or something like that. The rationale here is to make it possible for downstream widget crates to support several versions as needed.

ratatui-crossterm/src/lib.rs
joshka: We don't need to link every time this is mentioned. It's enough to link it in the first paragraph I think. Sound good?

xtask/src/commands/check.rs
```diff
@@ -13,9 +13,47 @@ pub struct Check {
...
+            // `underline-color` is part of default features for ratatui-crossterm,
+            // but with `--no-default-features`, we must add it explicitly if desired.
+            let common_features = [
+                "serde",
+                "underline-color",
+                "scrolling-regions",
+                "unstable",
+                "unstable-backend-writer",
+            ]
+            .join(",");
```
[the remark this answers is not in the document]
joshka: I don't think this scenario quite fits into cargo-hack. I may be wrong though. The problem is that the crossterm features are mutually exclusive.

ratatui-crossterm/Cargo.toml
orhun: Yeah, supporting only the last 2-3 versions sound reasonable to me. We should then update these docs accordingly (we only support blah blah)

I still believe we should adjust this PR a bit to make it easier to add new versions (see my other comments)

xtask/src/commands/check.rs
```diff
@@ -13,9 +13,47 @@ pub struct Check {
...
+
+            // Run `cargo check` on `ratatui-crossterm` with specific crossterm versions
+            for crossterm_feature in ["crossterm_0_28", "crossterm_0_29"] {
```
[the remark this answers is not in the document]
joshka: I'm mostly treating xtask as the place where this info is defined. Think of it as a makefile defined in rust. The available versions are not a runtime thing. It maybe makes sense to have a const in the xtask lib, but I think it's clear enough as-is on this. Put another way, you only have to look at this file to understand what the cargo xtask check command will do.

xtask/src/commands/clippy.rs
```diff
@@ -12,19 +12,57 @@ pub struct Clippy {
...
+        .join(",");
+
+        let clippy_options = ["--", "-D", "warnings"];
+
+        // Run Clippy on `ratatui-crossterm` with `crossterm_0_28` and `crossterm_0_29`
+        for crossterm_feature in ["crossterm_0_28", "crossterm_0_29"] {
+            let mut command = clippy_command.clone();
+            let features = format!("{common_features},{crossterm_feature}");
+            // Note that adding --tests or --benches causes clippy to pick up the default features.
+            // I'm not sure why this is the case (JM 2025-05-10).
```
joshka, line 40: Any ideas on this?

joshka (commented):
Note: this is still a WIP (getting CI to be happy with it)

xtask/src/commands/check.rs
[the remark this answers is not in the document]
orhun: I see... I guess we need to play around with `cargo-hack` a bit to see how that would work

ratatui-crossterm/src/lib.rs
orhun: I'd link every time due to OCD... but sure :D

xtask/src/commands/check.rs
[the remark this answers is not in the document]
orhun: Yup... then maybe let's do a const there? I really don't like the idea of modifying loops directly whenever a new version is out.

xtask/src/commands/clippy.rs
orhun: Never ran into it... But intuitively I'd guess that using those flags will use default features...

ratatui-crossterm/src/lib.rs
joshka: I draw the line when it's already mentioned in the same paragraph as excessive links can make the sentences more difficult to read. Here, Crossterm is the entire focus of this crate, so linking every occurrence would be annoyingly frequent. I too have tendencies that like a strongly consistent way of doing things, but if it helps, the rules that I'd apply here is less an "every time" rule and instead a "specifically on the first time only" rule.

xtask/src/commands/check.rs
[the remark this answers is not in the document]
joshka: I'm wrong - cargo hack has a `mutually-exclusive` option:

```
        --mutually-exclusive-features <FEATURES>...
            Space or comma separated list of features to not use together.

            To specify multiple groups, use this option multiple times:
            `--mutually-exclusive-features a,b --mutually-exclusive-features c,d`

            This flag can only be used together with --feature-powerset flag.
```

That said, I'd do that independently of this PR as we want to check how it affects CI.

xtask/src/commands/check.rs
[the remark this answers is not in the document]
joshka: done

xtask/src/commands/clippy.rs
```diff
@@ -12,19 +12,57 @@ pub struct Clippy {
...
+            "scrolling-regions",
+            "unstable",
+            "unstable-backend-writer",
+        ]
+        .join(",");
+
+        let clippy_options = ["--", "-D", "warnings"];
+
+        // Run Clippy on `ratatui-crossterm` with `crossterm_0_28` and `crossterm_0_29`
+        for crossterm_feature in ["crossterm_0_28", "crossterm_0_29"] {
```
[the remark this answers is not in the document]
joshka: done

xtask/src/commands/check.rs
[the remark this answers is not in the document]
joshka: Also, turns out that getting the build exactly right on this is kinda painful, whether we use cargo-hack or not.

ratatui-crossterm/Cargo.toml
joshka: done

joshka: Added a note to the lib and readme about the intention to support 2 versions + extras if needed.

joshka:
~~This is ready for review now~~

Ugh - coverage is failing, I assumed that was just the normal failure without checking. Instead it's caused by --all-features no longer working at the library level.

Edit: fixed

joshka:
> Probably related: #1820
> 
> Do you think this can be easily split into two PRs?

@orhun mentioned that link in a resolved comment. https://github.com/ratatui/ratatui/pull/1841#discussion_r2083452956

I used cargo-hack in this PR because it simplified code that otherwise wasn't working (see the history of broken builds in https://github.com/ratatui/ratatui/actions?query=workflow%3A%22Continuous+Integration%22+branch%3Ajm%2Fcrossterm-multiple-versions). Specifically, it makes the combination of builds to run just right.

I don't want to worry about the deeper concerns mentioned in #1820 right now, which are about actually testing combinations of feature flags. The problem with splitting this PR into two PRs is that the build fixes in cargo xtask are purely because we can no longer build with --all-features as `ratatui-crossterm/crossterm_0_28` and `ratatui-crossterm/crossterm_0_28` are mutually exclusive flags and have to be tested independently.

Put another way, cargo-hack isn't a necessary tool outside of this PR, but it's extremely useful in solving the problems that this PR causes, so it makes sense to keep it as part of this PR (I think).

xtask/src/commands/coverage.rs
orhun, line 1: Can you explain the rationale behind these changes? I'm a bit lost here 🤔

xtask/src/commands/clippy.rs
```diff
@@ -12,19 +12,57 @@ pub struct Clippy {
...
+        .join(",");
+
+        let clippy_options = ["--", "-D", "warnings"];
+
+        // Run Clippy on `ratatui-crossterm` with `crossterm_0_28` and `crossterm_0_29`
+        for crossterm_feature in CROSSTERM_VERSION_FEATURES {
+            let mut command = clippy_command.clone();
+            let features = format!("{common_features},{crossterm_feature}");
+            // Note that adding --tests or --benches causes clippy to pick up the default features.
+            // I'm not sure why this is the case (JM 2025-05-10).
```
orhun, line 40: Should probably be reported to cargo upstream

orhun (approved):
Looking all good, just had a small Q

xtask/src/commands/clippy.rs
joshka: I think it was probably feature unification due to having ratatui in the dev deps

xtask/src/commands/coverage.rs
joshka: This runs coverage with all features for all the packages except crossterm, which cannot be run with all-features (as you can't have two crossterm versions compiled in).

(Actually, I suppose one way of making the feature addictive would be to just choose the latest version as the one that's in use). That might make some of this simpler.

j-g00da: I had this thought exactly, I think it might be worth just choosing the highest version selected.

joshka: I'll experiment with this a bit and see if it works out. It would likely simplify a bunch of the xtask changes.

xtask/src/commands/clippy.rs
orhun: ah yes might be

aschey:
Another potential way to handle this - instead of making the feature flags mutually exclusive, you could configure it so that `crossterm_0_29` takes precedence over `crossterm_0_28` and so on. The downside would be that error messages caused by incompatible versions are a bit less clear, but this approach would be more congruent with how features are intended to work.

Anecdotally, I have my editor configured to always pass in `--all-features`, and if I'm working on a project that breaks this workflow, I have to figure out the correct feature flags to pass in so it builds correctly without having a bunch of stuff `cfg`-ed out. Not the worst thing in the world, but it is a bit inconvenient.

joshka:
> Another potential way to handle this - instead of making the feature flags mutually exclusive, you could configure it so that `crossterm_0_29` takes precedence over `crossterm_0_28` and so on. The downside would be that error messages caused by incompatible versions are a bit less clear, but this approach would be more congruent with how features are intended to work.

Yeah, I think I like that approach a bit better given that I hit the same problems fairly regularly. Thanks for the suggestion on this. I'm going to do it this way.

ratatui-crossterm/README.md
```diff
@@ -2,10 +2,40 @@
...
+This crate provides [`CrosstermBackend`], an implementation of the [`Backend`] trait for the
+[Ratatui] library. It uses the [Crossterm] library for all terminal manipulation.
+
+### Crossterm Version and Re-export
+
+`ratatui-crossterm` requires you to specify a version of the [Crossterm] library to be used.
+This is managed via feature flags. You **must enable one, and only one**, of the available
+`crossterm_0_xx` features (e.g., `crossterm_0_28`, `crossterm_0_29`). These features determine
+which version of Crossterm is compiled and used by the backend. Attempting to enable none or
+multiple `crossterm_0_xx` features will result in a compile-time error.
```
j-g00da: This should be updated

ratatui-crossterm/Cargo.toml
```diff
@@ -14,11 +14,31 @@ exclude.workspace = true
...
+
 [features]
-default = ["underline-color"]
+default = ["underline-color", "crossterm_0_29"]
+
+#! One of the following versions of crossterm must be enabled. By default, the latest version is
+#! enabled. This will always default to the most recent version of crossterm, but you can override
+#! this by specifying the version you want to use in your Cargo.toml file. The purpose of these
+#! features is to allow widget libraries to depend on specific versions of crossterm in a way that
+#! doesn't cause version conflicts. A compiler error will be raised if you try to use multiple
```
j-g00da: Compiler error is only raised if none of the version features are enabled.

j-g00da (approved):
Just one tiny correction needed, approving so it can be merged.

PR #1842, opened 2025-05-10
fix(layout): feature flag cache related types

joshka:


orhun (approved):

nanookclaw:
Good catches. Will address:
1) Clamping scroll delta to widget height — will add `delta = delta.min(self.items.len().saturating_sub(1))` before the offset update
2) Named constants for scroll amounts in tests — will replace magic numbers with `const SCROLL_AMOUNT: u16` declarations

Will push an update shortly.

PR #1858, opened 2025-05-14
chore: move dependency management to workspace

joshka:
Move all dependency management to the workspace level. This makes it
easier to manage dependencies across multiple crates in the workspace.

This also changes the versions of each dependency to track based on the
semver compatible version of the dependency (e.g. 0.1 instead of 0.1.0
or 2.9 instead of 2.9.0 to avoid having to regularly update the toml
files and to communicate that Ratatui will still generally work with
versions of the dependencies that are not the fully latest version. The
exact version of the dependencies is still tracked in the Cargo.lock
file.

Several dependencies that are fairly stable are changed to track a less
specific version (e.g. serde 1 instead of 1.0.x).

The following dependencies are updated to their latest versions:
- bitflags (2.3 -> 2.9)
- strum (0.26 -> 0.27)
- strum_macros (0.26 -> 0.27)
- all other semver compatible updates

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,15 +27,26 @@ edition = "2021"
 rust-version = "1.81.0"
 
 [workspace.dependencies]
-bitflags = "2.9.0"
-color-eyre = "0.6.3"
-crossterm = "0.29.0"
-document-features = "0.2.11"
-hashbrown = "0.15.3"
-indoc = "2.0.6"
-instability = "0.3.7"
-itertools = { version = "0.13.0", default-features = false, features = ["use_alloc"] }
-pretty_assertions = "1.4.1"
+anstyle = "1"
+bitflags = "2.9"
+color-eyre = "0.6"
+compact_str = { version = "0.9", default-features = false }
+criterion = { version = "0.5", features = ["html_reports"] }
+crossterm = "0.29"
+document-features = "0.2"
+fakeit = "1"
+futures = "0.3"
+hashbrown = "0.15"
+indoc = "2"
+instability = "0.3"
+itertools = { version = "0.14", default-features = false, features = ["use_alloc"] }
+kasuari = { version = "0.4", default-features = false }
+line-clipping = "0.3"
+lru = "0.14"
+palette = { version = "0.7" }
+pretty_assertions = "1"
+rand = "0.9"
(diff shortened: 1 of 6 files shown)


joshka:
I'm hoping that this will make dependabot updates just have to deal with the cargo.lock file most of the time.

Cargo.toml
```diff
@@ -27,29 +27,48 @@ edition = "2021"
...
+fakeit = "1"
+futures = "0.3"
+hashbrown = "0.15"
+indoc = "2"
+instability = "0.3"
+itertools = { version = "0.14", default-features = false, features = ["use_alloc"] }
+kasuari = { version = "0.4", default-features = false }
+line-clipping = "0.3"
+lru = "0.14"
+palette = { version = "0.7" }
```
orhun, line 46:
```suggestion
palette = "0.7"
```

Cargo.toml
```diff
@@ -27,29 +27,48 @@ edition = "2021"
...
-strum = { version = "0.26.3", default-features = false, features = ["derive"] }
-termion = "4.0.5"
-termwiz = { version = "0.23.3" }
-unicode-segmentation = "1.12.0"
+rstest = "0.25"
+serde = { version = "1", features = ["derive"] }
+serde_json = "1"
+strum = { version = "0.27", default-features = false, features = ["derive"] }
+termion = "4"
+termwiz = { version = "0.23" }
```
orhun, line 62:
```suggestion
termwiz = "0.23"
```

Cargo.toml
```diff
@@ -27,29 +27,48 @@ edition = "2021"
...
+thiserror = { version = "2", default-features = false }
+time = { version = "0.3", default-features = false }
+tokio = { version = "1" }
```
orhun, line 65:
```suggestion
tokio = "1"
```

Cargo.toml
```diff
@@ -27,29 +27,48 @@ edition = "2021"
...
+tracing = "0.1"
+tracing-appender = "0.2"
+tracing-subscriber = { version = "0.3" }
```
orhun, line 68:
```suggestion
tracing-subscriber = "0.3"
```

orhun (approved):

PR #1859, opened 2025-05-15
fix: specify correct msrv

joshka:
iter::repeat_n was stabilized in 1.82


joshka:
Probably would just wait on this and bump to 1.85 for picking up the edition change.

joshka:
#1860 supersedes this

Issue #1727, opened 2025-03-19
Migrate to Rust edition 2024 (Rust 1.87 - 23 May 2025)


joshka:
## Problem
Rust 1.85 introduced the 2024 Edition. It's not a huge set of new functionality, but it would be nice to keep things current.

## Solution

We've discussed (but not formally committed) that our MSRV policy is to support the current version and 2 previous versions of the stable compiler (this is the same policy as the time crate which we depend on, so it's transitively our policy). This generally means that we give users 3+ months to upgrade their compiler.

So once Rust 1.87.0 is released, we should update to edition 2024. This is expected to be released on 16 May 2025 (the compiler has a regular 6 week release cadence).

## Alternatives
We could consider holding back on doing this if there's specific blockers that downstream libs / apps come to us with. Except Debian's usage of the 1.63 compiler - that's not really a good basis for pushing back.

MatrixFrog:
I was curious what would happen so I tried it out: https://github.com/ratatui/ratatui/pull/1801/files

Mostly formatting changes, and then clippy wanting a lot of functions to be marked as `const fn`

joshka:
It would be worth breaking this up into smaller reviewable pieces. I'd expect at least some parts of the PR would probably be valid in the current MSRV and could be merged early. It looks like the formatting changes the ordering of nested types and modules, so that would wait until the edition update. Does this match your understanding?

MatrixFrog:
That sounds right. I wonder if it would be good to also do the toolchain version update in a separate PR from the rust edition update, in case there's a subtle bug introduced from the latter, then it can more easily be rolled back separately.

MatrixFrog:
trying again #1860

MatrixFrog:
The Plan (from https://github.com/ratatui/ratatui/pull/1801/#issuecomment-2887648927)

1. bump rust-version to 1.85 (not 1.86) (done: #1860)
1. run `cargo fix --edition`
1. check the various fixes that are introduced and work out which are necessary. Some may be relevant, some may not, but we do want to document the choices we make regardless. There's some complexity in suggested fixes that would take a bit more effort than usual to dig into and understand.
1. update edition to 2024
1. run `cargo xtask fmt`

MatrixFrog:
`cargo fix --edition` produces changes in 10 files:
 * 2 additions of `+ use<>` and one of `+ use<'_>` -- see https://blog.rust-lang.org/2024/09/05/impl-trait-capture-rules/
[...]

PR #1861, opened 2025-05-18
refactor: Swap const and non-const arms in color! and modifier! macros

MatrixFrog:
This gets us a little bit closer to Edition 2024 support (#1727)

I am still learning but I think I'm understanding [this part of the edition 2024 guide](https://doc.rust-lang.org/edition-guide/rust-2024/macro-fragment-specifiers.html) correctly: If we leave these macro definitions unchanged and upgrade to Edition 2024, the first arm will always match, because `expr` will start matching `const` expressions that it didn't match in Edition 2021. And then the second arm will never match.

So, we simply switch the order to explicitly match const expressions first. This is a no-op for Edition 2021 but will make these work as expected under Edition 2024.

### DIFF
--- a/ratatui-core/src/style/stylize.rs
+++ b/ratatui-core/src/style/stylize.rs
@@ -112,20 +112,6 @@ impl fmt::Debug for ColorDebug {
 /// }
 /// ```
 macro_rules! color {
-    ( $variant:expr, $color:ident(), $on_color:ident() -> $ty:ty ) => {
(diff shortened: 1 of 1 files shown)


MatrixFrog:
Nevermind, as long as we have `pub` this isn't necessary. If this was more complicated and allowed various combinations of having/not having `pub` and having/not having `const` then we might need to tweak it, but I think this works fine as is.

PR #1862, opened 2025-05-18
chore: Use edition.workspace so that all crates are on the same edition

MatrixFrog:
The workspace edition is currently 2021 so this doesn't actually change anything yet. It just means that when the workspace edition does change, it will change for all crates at once.

Part of #1727


joshka:
We can change the workspace edition in this PR too. No need to split quite this atomically :)

PR #1863, opened 2025-05-19
chore: upgrade to Rust Edition 2024

### LINKED ISSUE #1727: Migrate to Rust edition 2024 (Rust 1.87 - 23 May 2025)

MatrixFrog:
https://doc.rust-lang.org/edition-guide/rust-2024/index.html

Fixes #1727


joshka (approved):

joshka:
Thanks again for putting in the work for this. I added the formatting changes as we've previously looked at these. There's a bunch that given the rust 2024 formatting I'd probably write the code a bit neater, but those changes can come later.

PR #1865, opened 2025-05-20
feat(no_std): make `ratatui-macros` no-std

j-g00da:


j-g00da:
I had to explicitly link alloc in docs to make it work properly.
`fails.stderr` was overwritten as line numbers in stderr changed.

joshka (approved):
Generally LGTM - 1 question

ratatui-macros/src/lib.rs
```diff
@@ -112,6 +113,7 @@
 //! a `;` followed by the format string and arguments.
 //!
 //! ```rust
+//! # extern crate alloc;
```
joshka, line 116: Do these macros work fine (i.e without that crate having to add extern crate alloc) when called from other crates that aren't no_std? Is it worth it to add that check as an explicit test, or is this something that's obvious enough that it should always work?

j-g00da: This should work fine in std crates. `extern crate alloc` just links alloc, which is already linked if you don't explicitly disable std with `#![no_std]`.

No-std crates will on the other hand have to explicitly link alloc, but we can include this in some no-std guide on the website.

joshka: How about adding that to the lib.rs (and readme via cargo-rdme?)

orhun (approved):
LGTM

joshka (approved):
LGTM

joshka:
Thanks again for getting this over the line.

j-g00da:
No problem 👌

PR #1874, opened 2025-05-27
feat: block border merging

j-g00da:
Co-authored by @pauladam94
Fork of #1729


ratatui-core/src/symbols/line.rs
```diff
@@ -57,6 +59,321 @@ pub const THICK_HORIZONTAL_UP: &str = "┻";
 pub const CROSS: &str = "┼";
 pub const DOUBLE_CROSS: &str = "╬";
 pub const THICK_CROSS: &str = "╋";
+/// A visual style defining the appearance of a single line making up a block border.
```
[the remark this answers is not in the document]
j-g00da: Ahh ok, I missinterpreted it as I wanted to add a blank lines before "Example" anyway and thought the comment was about this. Yeah these definitely need a blank line.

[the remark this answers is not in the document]
joshka: It's sort of a nit - I definitely prefer to read source where each item has a space before it starts, mainly because it makes the eyes associate the comment with the next item, not the previous when I read the source.

ratatui-core/src/symbols/merge.rs
```diff
@@ -0,0 +1,30 @@
...
+    /// Merges symbols only if an exact composite unicode character exists.
+    ///
+    /// Example: `┐` and `┗` will be merged into `╄`
+    #[default]
+    Exact,
+    /// Merges symbols even if an exact composite unicode character doesn't exist,
+    /// using the closest match.
+    ///
+    /// Example: `╮` and `└` will be merged into `┼`
+    BestFit,
```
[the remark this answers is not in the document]
j-g00da: Fuzzy makes sense and I like it being one word.

ratatui-core/src/symbols/line.rs
```diff
@@ -58,6 +60,331 @@ pub const CROSS: &str = "┼";
...
+            // (Thick, Plain | Thick) | (Plain, Thick) => Thick,
+            // (Double, Plain | Double) | (Plain, Double) => Double,
+            (Plain, Plain) => Plain,
+            (DoubleDash, DoubleDash) => DoubleDash,
+            (TripleDash, TripleDash) => TripleDash,
+            (TripleDashThick, TripleDashThick) => TripleDashThick,
+            (QuadrupleDash, QuadrupleDash) => QuadrupleDash,
+            (QuadrupleDashThick, QuadrupleDashThick) => QuadrupleDashThick,
+            (_, other) => other,
+        }
```
j-g00da, line 115: I think we can remove this match statement? Most cases do nothing, as they are redundant as they are already covered by `(_, other) => other`. This whole thing can be just:
```rust
if other != Nothing { other } else { self }
```

ratatui-core/src/symbols/line.rs
```diff
@@ -57,6 +59,321 @@ pub const THICK_HORIZONTAL_UP: &str = "┻";
...
+    /// Merges line styles.
+    #[must_use]
+    pub const fn merge(self, other: Self) -> Self {
+        #[allow(clippy::enum_glob_use)]
+        use LineStyle::*;
+        match (self, other) {
+            (Nothing, Nothing) => Nothing,
+            (s, Nothing) | (Nothing, s) => s,
+            // (Thick, Plain | Thick) | (Plain, Thick) => Thick,
+            // (Double, Plain | Double) | (Plain, Double) => Double,
```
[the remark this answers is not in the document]
j-g00da: This would make Thick and Double override Plain. Do we want this? I think the current solution causes less ambiguity.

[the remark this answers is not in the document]
joshka: Don't we want to merge as if the last rendered border takes priority. Imagine rendering a single border around some panes but rendering a double or thick border for the selected pane that overwrites any single borders.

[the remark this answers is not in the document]
j-g00da: But that's how it works now. The last one overwrites previous one. The commented lines would make Thick and Double take priority over Plain regardless of order.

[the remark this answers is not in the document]
j-g00da: Maybe I don't see something you do, I will try to visualize how I see this and why IMO  this is a problem regardless of this being public or not.

Let's render three boxes from left to right, so that boxes overlap like this, set merge strategy to fuzzy:
```
┌─┬─┬─┐
│ │ │ │
└─┴─┴─┘ 
```
Now let's change the border type of the middle box to double dash and assume these two lines are **not** commented out.
```
┌─┬╌┬─┐
│ ╎ │ │
└─┴╌┴─┘ 
```
That makes sense, the last box draws on top of the middle one.
Let's try thick quad dash:
```
┌─┲┉┭─┐
│ ┋ │ │
└─┺┉┵─┘ 
```
Works the same way, great.
Now let's change it to thick:
```
┌─┲━┱─┐
│ ┃ ┃ │
└─┺━┹─┘ 
```
Suddenly it appears like the render order changed. I think user would expect it to work the same way no matter the border type.

If we would really want to go that way, we would have to make rules for all cases, but it's not that easy - general rule would be to overwrite plain with stylized borders, but:
- What about plain + dashed? Wouldn't it make more sense if plain overwrites dashed?
- What about e.g. double dash and triple dash? should they create some dash or plain line?
- Double dash + thick double dash would create a thick double dash, but what about triple dash and thick double dash?
- What about thick + double?
- Imagine a rectangle with a smaller rectangle inside, with one common vertex. One rectangle has rounded border, should the common vertex be rounded?

And so on...
I think using the last rendered one makes it much more predictable for user.

ratatui-core/src/symbols/merge.rs
```diff
@@ -0,0 +1,328 @@
...
+            return self;
+        }
+        // Rouned border character are only available for corners.
+        if self.contains(Rounded) {
+            self = self.replace(Rounded, Plain);
+        }
+
+        self
+    }
+
```
j-g00da, line 87: I think this needs refactoring, it doesn't take all cases into account.
Initial pseudocode:
```
if not horizontal or vertical line:
    replace all dashes with plain
    replace all thick dashes with thick
if not corner:
    replace all rounded with plain
if has thick and double:
    replace all double with thick
if <check for missing plain/double combinations>:
    replace all double with thick
```
@joshka

ratatui-widgets/src/block.rs
```diff
@@ -639,19 +651,23 @@ impl Block<'_> {
 
     fn render_left_side(&self, area: Rect, buf: &mut Buffer) {
```
joshka, line 652: While we're refactoring / changing this. I think there's a refactoring worth doing that makes this neat. Translate these 4 functions into a loop over an array of values which reduces a bunch of repeated code. This is valuable here because we're changing the same code in the same way in multiple places, so it reduces the maintenance burden in future updates / reading of this code.

```rust
let sides = [
    (Borders::LEFT, area.left()..area.left(), area.top()..area.bottom()),
    (Borders::TOP, area.left()..area.right(), area.top()..area.top()),
    (Borders::RIGHT, area.right()..area.right(), area.top()..area.bottom()),
    (Borders::BOTTOM, area.left()..area.right(), area.bottom()..area.bottom()),
for (border, y_range, x_range, symbol) in sides {
    if self.borders.contains(border) { ... }
}
```
The corners would be similar I think.

ratatui-widgets/src/block.rs
```diff
@@ -530,6 +534,14 @@ impl<'a> Block<'a> {
         self
     }
 
+    /// Sets the block's [`MergeStrategy`] for overlapping characters. Setting it to `None`
+    /// never merges characters.
+    #[must_use = "method moves the value of self and returns the modified value"]
+    pub const fn border_merge_strategy(mut self, merge_strategy: MergeStrategy) -> Self {
```
joshka, line 540: Simplification on naming:

```suggestion
    pub const fn merge_borders(mut self, strategy: MergeStrategy) -> Self {
```

WDYT?

ratatui-core/src/symbols/merge.rs
j-g00da:
```rust
    pub fn fuzzy(mut self) -> Self {
        #[allow(clippy::enum_glob_use)]
        use LineStyle::*;

        // Dashes only include vertical and horizontal lines.
        if !self.is_uniform_line() {
            self = self
                .replace(DoubleDash, Plain)
                .replace(TripleDash, Plain)
                .replace(QuadrupleDash, Plain)
                .replace(DoubleDashThick, Thick)
                .replace(TripleDashThick, Thick)
                .replace(QuadrupleDashThick, Thick);
        }

        // Rounded has only corner variants.
        if !self.is_uniform_corner() {
            self = self.replace(Rounded, Plain);
        }

        // There are no Double + Thick variants.
        // Additionally, some Plain + Double variants don't exist.
        if TryInto::<&str>::try_into(self).is_err() {
            self = self.replace(Double, Thick);
        }
        self
    }

```
I'm not fully satisfied with this, but it covers all cases.
Requires two new methods - `is_uniform_line` and `is_uniform_corner`, but makes `contains` redundant.
WDYT?

joshka (changes requested):

ratatui-core/src/symbols/merge.rs
joshka: I think your comment was actually written before I hit complete review.
Bikeshed comment: What about `is_straight()` and `is_corner()` for the naming?

ratatui-widgets/src/block.rs
j-g00da: Ok so I made it work, but it required some hacks.
This passes the tests, let me know if you have some better ideas or is it good enough. Would you like me to inline the `to_range` closure? `#[allow(clippy::range_plus_one)]` is needed, because we need `Range` and not `InclusiveRange`, or else the types in array won't match. I can just make all ranges inclusive, but this would on the other hand make the offseted ranges less readable.

```rust
    fn render_borders(&self, area: Rect, buf: &mut Buffer) {
        let offset = u16::from(self.merge_strategy != MergeStrategy::Replace);
        #[allow(clippy::range_plus_one)]
        let as_range = |value: u16| value..value + 1;
        let sides = [
            (
                Borders::LEFT,
                as_range(area.left()),
                area.top() + offset..area.bottom() - offset,
                self.border_set.vertical_left,
            ),
            (
                Borders::TOP,
                area.left() + offset..area.right() - offset,
                as_range(area.top()),
                self.border_set.horizontal_top,
            ),
            (
                Borders::RIGHT,
                as_range(area.right() - 1),
                area.top() + offset..area.bottom() - offset,
                self.border_set.vertical_right,
            ),
            (
                Borders::BOTTOM,
                area.left() + offset..area.right() - offset,
                as_range(area.bottom() - 1),
                self.border_set.horizontal_bottom,
            ),
        ];
        for (border, x_range, y_range, symbol) in sides {
            if self.borders.contains(border) {
                for x in x_range {
                    for y in y_range.clone() {
                        buf[(x, y)]
                            .merge_symbol(symbol, self.merge_strategy)
                            .set_style(self.border_style);
                    }
                }
            }
        }

        // TODO: replace these too
        self.render_bottom_right_corner(buf, area);
        self.render_top_right_corner(buf, area);
        self.render_bottom_left_corner(buf, area);
        self.render_top_left_corner(buf, area);
    }
```

This would obviously require some comments, overall the loops themselves are pretty straightforward, the sides array itself on the other hand is off-by-one hell.

ratatui-core/src/symbols/merge.rs
```diff
@@ -0,0 +1,328 @@
...
+    /// Merges symbols even if an exact composite unicode character doesn't exist,
+    /// using the closest match.
+    ///
+    /// Example: `╮` and `└` will be merged into `┼`
+    Fuzzy,
+}
+
+impl MergeStrategy {
+    pub fn merge<'a>(self, prev: &'a str, next: &'a str) -> &'a str {
```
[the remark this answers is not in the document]
j-g00da: With replace, I can just do it like this:

```rust
#[test]
fn replace_merge_strategy() {
    let strategy = MergeStrategy::Replace;
    let symbols = [
        "─", "━", "│", "┃", "┄", "┅", "┆", "┇", "┈", "┉", "┊", "┋", "┌", "┍", "┎", "┏", "┐",
        "┑", "┒", "┓", "└", "┕", "┖", "┗", "┘", "┙", "┚", "┛", "├", "┝", "┞", "┟", "┠", "┡",
        "┢", "┣", "┤", "┥", "┦", "┧", "┨", "┩", "┪", "┫", "┬", "┭", "┮", "┯", "┰", "┱", "┲",
        "┳", "┴", "┵", "┶", "┷", "┸", "┹", "┺", "┻", "┼", "┽", "┾", "┿", "╀", "╁", "╂", "╃",
        "╄", "╅", "╆", "╇", "╈", "╉", "╊", "╋", "╌", "╍", "╎", "╏", "═", "║", "╒", "╓", "╔",
        "╕", "╖", "╗", "╘", "╙", "╚", "╛", "╜", "╝", "╞", "╟", "╠", "╡", "╢", "╣", "╤", "╥",
        "╦", "╧", "╨", "╩", "╪", "╫", "╬", "╭", "╮", "╯", "╰", "╴", "╵", "╶", "╷", "╸", "╹",
        "╺", "╻", "╼", "╽", "╾", "╿",
    ];

    for a in symbols {
        for b in symbols {
            assert_eq!(strategy.merge(a, b), b);
        }
    }
}
```
The bigger problem are exact and fuzzy, as these would require either 15625 asserts or an output array of size 15625 each.

ratatui-widgets/src/block.rs
joshka: I feel the closure makes it difficult to read. What about simplifying with a few intermediate variables instead?

`let right = area.right() - 1;
let right_offset = area.right() - offset;
// etc.
`

And then using an inclusive range for the non changing range makes sense.

ratatui-core/src/symbols/merge.rs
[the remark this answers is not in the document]
joshka: A good way would be to categorize these into a list of items that are mostly the same. E.g the T pieces with one of the sides being thick are mostly the same as each other. You can just pick one of the symbols and add a few tests that arrive at that by exact methods and some that arrive there by the fuzzy method.

This can be a bit "white box" testing because you know the algorithm is mostly agnostic to which border part is of a particular type, so a rotated piece hits the same logic

Arrange the assertions in Unicode order by the destination symbol and it would be easy to add tests if needed.

ratatui-widgets/src/block.rs
j-g00da: I can't mix Range with InclusiveRange - the inclusive one wouldn't be on the same position in each tuple, so I still have to silence clippy.  But yeah, this looks cleaner now anyway.
```rust
        let offset = u16::from(self.merge_strategy != MergeStrategy::Replace);
        
        let left = area.left();
        let left_offset = area.left() + offset;
        let top = area.top();
        let top_offset = area.top() + offset;
        let right = area.right() - 1;
        let right_offset = area.right() - offset;
        let bottom = area.bottom() - 1;
        let bottom_offset = area.bottom() - offset;
        
        #[allow(clippy::range_plus_one)]
        let sides = [
            (
                Borders::LEFT,
                left..left + 1,
                top_offset..bottom_offset,
                self.border_set.vertical_left,
            ),
            (
                Borders::TOP,
                left_offset..right_offset,
                top..top + 1,
                self.border_set.horizontal_top,
            ),
            (
                Borders::RIGHT,
                right..right + 1,
                top_offset..bottom_offset,
                self.border_set.vertical_right,
            ),
            (
                Borders::BOTTOM,
                left_offset..right_offset,
                bottom..bottom + 1,
                self.border_set.horizontal_bottom,
            ),
        ];
```

joshka: Or could rangebounds make this work in the iterator. https://doc.rust-lang.org/std/ops/trait.RangeBounds.html (not sure if that makes it more complex)

j-g00da: IMO the last code snippet I pasted is the most readable approach here.

ratatui-core/src/buffer/cell.rs
```diff
@@ -61,6 +62,13 @@ impl Cell {
         self.symbol.as_str()
     }
 
+    /// Merge the symbol of the cell with the one already on the cell.
+    pub fn merge_symbol(&mut self, symbol: &str, strategy: MergeStrategy) -> &mut Self {
+        let next = strategy.merge(self.symbol(), symbol);
+        self.symbol = CompactString::new(next);
```
orhun, line 68: nit: `merged` sounds more correct than `next` here, WDYT?

```suggestion
        let merged = strategy.merge(self.symbol(), symbol);
        self.symbol = CompactString::new(merged);
```

ratatui-core/src/symbols/merge.rs
```diff
@@ -0,0 +1,431 @@
...
+}
+
+/// Represents a composite border symbol using individual line components.
+#[derive(Debug, Copy, Clone)]
+struct BorderSymbol {
+    pub right: LineStyle,
+    pub up: LineStyle,
+    pub left: LineStyle,
+    pub down: LineStyle,
+}
```
orhun, line 44: `pub` fields needs to be documented

ratatui-core/src/symbols/merge.rs
```diff
@@ -0,0 +1,431 @@
...
+        // Additionally, some Plain + Double variants don't exist.
+        // TODO: check if this is good enough after adding tests
+        if TryInto::<&str>::try_into(self).is_err() {
+            self = self.replace(Double, Thick);
+        }
+        self
+    }
+
+    /// Return true only if the symbol is a line and both parts have the same [`LineStyle`].
+    fn is_straight(self) -> bool {
```
orhun, line 91: Hmm, why do these methods consume `self`?

ratatui-core/src/symbols/merge.rs
```diff
@@ -0,0 +1,431 @@
...
+        use LineStyle::Nothing;
+        (self.up == self.down && self.left == self.right)
+            && (self.up == Nothing || self.left == Nothing)
+    }
+
+    /// Return true only if the symbol is a corner and both parts have the same [`LineStyle`].
+    fn is_corner(self) -> bool {
+        use LineStyle::Nothing;
+        // TODO: maybe this can be simplified?
```
orhun, line 100: I'm okay with this at its current state, simplifying it might make it less readable imo

j-g00da:
I think it's just a matter of expanding the docs now, can you confirm? @joshka @orhun

ratatui-core/src/symbols/merge.rs
```diff
@@ -0,0 +1,439 @@
...
+        if !self.is_corner() {
+            self = self.replace(Rounded, Plain);
+        }
+
+        // There are no Double + Thick variants.
+        // Additionally, some Plain + Double variants don't exist.
+        // TODO: check if this is good enough after adding tests
+        if TryInto::<&str>::try_into(self).is_err() {
+            self = self.replace(Double, Thick);
+        }
```
j-g00da: Let me know if this is ok so I can remove the TODO.

ratatui-core/src/symbols/merge.rs
```diff
@@ -0,0 +1,439 @@
...
+        assert_eq!(strategy.merge("│", "┕"), "┝");
+        assert_eq!(strategy.merge("┏", "│"), "┝");
+        assert_eq!(strategy.merge("┏", "┆"), "┝");
+        assert_eq!(strategy.merge("│", "┏"), "┢");
+        assert_eq!(strategy.merge("╽", "┕"), "┢");
+        assert_eq!(strategy.merge("│", "─"), "┼");
+        assert_eq!(strategy.merge("┆", "─"), "┼");
+        assert_eq!(strategy.merge("┘", "┌"), "┼");
+        assert_eq!(strategy.merge("┘", "╭"), "┼");
+        assert_eq!(strategy.merge("┵", "╞"), "┿");
```
joshka: Intuitively, it seems to me that `╪` would be a better fit for this because the double line is in the character that replaces the existing character.

j-g00da: This is caused by how the merging works right now - fuzzy is applied when the new BorderSymbol is already created, so we don't preserve information about _line_ overlaps. I will try to rework the logic there. Ideally I think BorderSymbol should have different merge methods for each strategy instead of applying fuzzy _after_ the merge, so we can have full control over this.

j-g00da: Looked into the code, this would be even easier and can be achevied by just passing last merged in symbol into fuzzy method.

```rust
        // There are no Double + Thick variants.
        if self.contains(Double) && self.contains(Thick) {
            // Decide whether to use Double or Thick, based on the last merged-in symbol.
            if last_symbol.contains(Double) {
                self = self.replace(Thick, Double);
            } else {
                self = self.replace(Double, Thick);
            }
        }

        // Some Plain + Double variants don't exist, use Thick instead in such case.
        if TryInto::<&str>::try_into(self).is_err() {
            self = self.replace(Double, Thick);
        }
        self
```

j-g00da: I'm thinking about it now from the user's perspective and what would make sense in real scenarios, do we want to really change double to thick or thick to double at all? This pretty much always ends up looking quite bad.

Look how double merges with thick, I don't think this provides any value to most users.
```
┌─┬┬─┐
│┏┿╅─┤
├╂┘┃ │
│┃╔╬╗│
│┗╬┛║│
└─╨─╨┘
```

Other option I see is to make fuzzy affect only rounded (-> plain), dash variants (-> plain) and thick dash variants (-> thick). These actually look good.

Currently:
```
┵ + ╞ -> ┿
╎ + ┉ -> ┿
┘ + ╭ -> ┼
```
My last comment (option A):
```
┵ + ╞ -> ╪
╎ + ┉ -> ┿
┘ + ╭ -> ┼
```
This comment (option B):
```
┵ + ╞ -> ╞
╎ + ┉ -> ┿
┘ + ╭ -> ┼
```

WDYT? @joshka @orhun

joshka: The example output is a little too complex to clearly see the impact in a way that a user would see as real. I'd start with the end result in mind. Look at some real merged blocks. Take a block with borders and show what happens when you intersect it in several ways:

- for each border type
  - with each border type
    - for each corner show 4 versions:
      - touching corners
      - touching horizontal
      - touching vertical
      - intersecting

3 nested for loops, a small amount of 0,1,-1 type math to draw a block and 4 surrounding blocks (can be partial as only the intersecting sides need be drawn). This solves the question visually instead of abstractly.

j-g00da: I think something like this is enough:
https://gist.github.com/j-g00da/9f083fe1a9e07e3a1b1bcec9a794fb6d

Each file is a result produced by different implementation.

joshka: Looks great, the other idea for joining double corners over plain or thick sides might be to use `╦`. It's wrong, but at least it's got two right double parts instead of just one. WDYT?

j-g00da: Can you show some example?

joshka: Given:
```plain
                           
 ┌──┐   ┌──┐   ┌──┐   ┌──┐ 
 │  │   │  │   │  │   │  │ 
 │  │   │  │   │  │   │  │ 
 └──┘   └──┘   └──┘   └──┘ 
```
With the following merged over:
```
╔╗  ╔╗ ╔═╗╔═╗ ╔╗  ╔╗ ╔═╗╔═╗
╚╝  ╚╝ ╚═╝╚═╝ ║║  ║║ ║ ║║ ║
              ╚╝  ╚╝ ╚═╝╚═╝
              ╔╗  ╔╗ ╔═╗╔═╗
╔╗  ╔╗ ╔═╗╔═╗ ║║  ║║ ║ ║║ ║
╚╝  ╚╝ ╚═╝╚═╝ ╚╝  ╚╝ ╚═╝╚═╝
```

With the current implementation, this currently gives:

Replace:
```
╔╗  ╔╗ ╔═╗╔═╗ ╔╗  ╔╗ ╔═╗╔═╗
╚╝──╚╝ ╚═╝╚═╝ ║║──║║ ║┌║║┐║
 │  │   │  │  ╚╝  ╚╝ ╚═╝╚═╝
 │  │   │  │  ╔╗  ╔╗ ╔═╗╔═╗
╔╗──╔╗ ╔═╗╔═╗ ║║──║║ ║└║║┘║
╚╝  ╚╝ ╚═╝╚═╝ ╚╝  ╚╝ ╚═╝╚═╝
```
Exact:
```
╔╗  ╔╗ ╔═╗╔═╗ ╔╗  ╔╗ ╔═╗╔═╗
╚╝──╚╝ ╚╤╝╚╤╝ ║╟──╢║ ║┌╫╫┐║
 │  │   │  │  ╚╝  ╚╝ ╚╪╝╚╪╝
 │  │   │  │  ╔╗  ╔╗ ╔╪╗╔╪╗
╔╗──╔╗ ╔╧╗╔╧╗ ║╟──╢║ ║└╫╫┘║
╚╝  ╚╝ ╚═╝╚═╝ ╚╝  ╚╝ ╚═╝╚═╝
```
Fuzzy:
```
╔╗  ╔╗ ╔═╗╔═╗ ╔╗  ╔╗ ╔═╗╔═╗
╚╃──╄╝ ╚╤┹┺╤╝ ║╟──╢║ ║┌╫╫┐║
 │  │   │  │  ╚┩  ┡╝ ╚╪╝╚╪╝
 │  │   │  │  ╔┪  ┢╗ ╔╪╗╔╪╗
╔╅──╆╗ ╔╧┱┲╧╗ ║╟──╢║ ║└╫╫┘║
╚╝  ╚╝ ╚═╝╚═╝ ╚╝  ╚╝ ╚═╝╚═╝
```
I think fuzzy should actually be:
```
╔╗  ╔╗ ╔═╗╔═╗ ╔╗  ╔╗ ╔═╗╔═╗
╚╬──╬╝ ╚╤╩╩╤╝ ║╟──╢║ ║┌╫╫┐║
 │  │   │  │  ╚╣  ╠╝ ╚╪╝╚╪╝
 │  │   │  │  ╔╣  ╠╗ ╔╪╗╔╪╗
╔╬──╬╗ ╔╧╦╦╧╗ ║╟──╢║ ║└╫╫┘║
╚╝  ╚╝ ╚═╝╚═╝ ╚╝  ╚╝ ╚═╝╚═╝
```
Note the use of `╬ ╩ ╦ ╣╠`

The overwritten corners have two double segments with two plain segments. The double is the last written border so it should be given some amount of preference. That seems like it should render as `╬` to me.

Same with double corner over a single line. Ends up with 2 double segments and one line segment, which seems like this should render as ╩.

If the plain was rendered over double, then the symbols would be single cross / T symbols.

My rationale for this is that if you're rendering a block last, that block should likely be the one with the highest z-index, and so it's the one which the borders make most sense to be as uniform as possible. Going from double to thick when there's no thick borders anywhere in the interface seems confusing and difficult to justify.

I'll commit (https://github.com/ratatui/ratatui/pull/1874/commits/d81ebdbaec86316bed57af15e0d41ea50df0d2c8) the block test I was using to generate these this so you can use it / modify it as needed. That said it could be worth just grabbing the inputs and outputs instead as these would be simple tests of just the merge strategy instead of the block border code.

j-g00da: Oh okay now I get it, thanks!
[I've added results to the gist.](https://gist.githubusercontent.com/j-g00da/9f083fe1a9e07e3a1b1bcec9a794fb6d/raw/2c9b42758d0e82269bd78f34e7d9e3088308abb3/plain_double_last_precedence.txt)

ratatui-widgets/src/block.rs
```diff
@@ -1647,4 +1670,2839 @@ mod tests {
...
+    }
+
+    #[test]
+    #[allow(clippy::too_many_lines)]
+    fn render_replace_merge() {
+        let buffer = render_merging_borders(MergeStrategy::Replace);
+
+        #[rustfmt::skip]
+        let expected = Buffer::with_lines([
+            "┌───┐    ┌───┐  ┌───┌───┐┌───┐",
```
joshka: This definitely deserves its own file or files, but is very good otherwise.

ratatui-widgets/src/block/merge/exact.txt
orhun, line 1: Maybe rename these `.txt` files to `.snap` since they are mostly snapshots?

joshka: These are text files, so txt is correct IMO. If we change to using insta at some point in the future (I'm adding a comment on the test right now about that), then we should use the conventional names that insta uses - not sure if that's snap or txt off the top of my head though.

orhun: Yes, insta uses `.snap` extension.

Then maybe at least move them to `tests/` or something? It seems off to have txt files among source files.

ratatui-core/src/symbols/merge.rs
```diff
@@ -0,0 +1,740 @@
...
+        match (self.up, self.right, self.down, self.left) {
+            (up, right, Nothing, Nothing) => up == right,
+            (Nothing, right, down, Nothing) => right == down,
+            (Nothing, Nothing, down, left) => down == left,
+            (up, Nothing, Nothing, left) => up == left,
+            _ => false,
+        }
+    }
+
+    // TODO: not removing this yet as fuzzy strategy implementation may change
```
orhun: Should this be removed now?

j-g00da: Yes!

ratatui-widgets/src/block/merge/exact.txt
j-g00da: Yeah I didn't really know where to put these

joshka: I'm moving them to tests/block/merge_exact.txt etc.

joshka: https://github.com/ratatui/ratatui/pull/1874/commits/d86bfe2e5f8a43f3e0825ba4db4e5be7574abfe5

ratatui-widgets/src/block.rs
```diff
@@ -1647,4 +1714,148 @@ mod tests {
         ]);
         assert_eq!(buffer, expected);
     }
+
+    #[test]
+    fn render_partial_borders() {
```
j-g00da: This test would pass even with no fix as it only checks with `Replace`, which already worked. IMO this should be parametrized to check with all `MergeStrategy` variants.

j-g00da (commented):
LGTM just one thing

ratatui-widgets/src/block.rs
```diff
@@ -1647,4 +1714,148 @@ mod tests {
...
+            .borders(Borders::LEFT | Borders::RIGHT)
+            .render(buffer.area, &mut buffer);
+        #[rustfmt::skip]
+        let expected = Buffer::with_lines([
+            "│        │",
+            "│        │",
+            "│        │",
+        ]);
+        assert_eq!(buffer, expected);
+    }
```
j-g00da:
```suggestion
    #[rstest]
    #[case::replace(MergeStrategy::Replace)]
    #[case::exact(MergeStrategy::Exact)]
    #[case::fuzzy(MergeStrategy::Fuzzy)]
    fn render_partial_borders(#[case] strategy: MergeStrategy) {
        let mut buffer = Buffer::empty(Rect::new(0, 0, 10, 3));
        Block::new()
            .borders(Borders::TOP | Borders::LEFT | Borders::RIGHT | Borders::BOTTOM)
            .merge_borders(strategy)
            .render(buffer.area, &mut buffer);
        #[rustfmt::skip]
        let expected = Buffer::with_lines([
            "┌────────┐",
            "│        │",
            "└────────┘",
        ]);
        assert_eq!(buffer, expected);

        let mut buffer = Buffer::empty(Rect::new(0, 0, 10, 3));
        Block::new()
            .borders(Borders::TOP | Borders::LEFT)
            .merge_borders(strategy)
            .render(buffer.area, &mut buffer);
        #[rustfmt::skip]
        let expected = Buffer::with_lines([
            "┌─────────",
            "│         ",
            "│         ",
        ]);
        assert_eq!(buffer, expected);

        let mut buffer = Buffer::empty(Rect::new(0, 0, 10, 3));
        Block::new()
            .borders(Borders::TOP | Borders::RIGHT)
            .merge_borders(strategy)
            .render(buffer.area, &mut buffer);
        #[rustfmt::skip]
        let expected = Buffer::with_lines([
            "─────────┐",
            "         │",
            "         │",
        ]);
        assert_eq!(buffer, expected);

        let mut buffer = Buffer::empty(Rect::new(0, 0, 10, 3));
        Block::new()
            .borders(Borders::BOTTOM | Borders::LEFT)
            .merge_borders(strategy)
            .render(buffer.area, &mut buffer);
        #[rustfmt::skip]
        let expected = Buffer::with_lines([
            "│         ",
            "│         ",
            "└─────────",
        ]);
        assert_eq!(buffer, expected);

        let mut buffer = Buffer::empty(Rect::new(0, 0, 10, 3));
        Block::new()
            .borders(Borders::BOTTOM | Borders::RIGHT)
            .merge_borders(strategy)
            .render(buffer.area, &mut buffer);
        #[rustfmt::skip]
        let expected = Buffer::with_lines([
            "         │",
            "         │",
            "─────────┘",
        ]);
        assert_eq!(buffer, expected);

        let mut buffer = Buffer::empty(Rect::new(0, 0, 10, 3));
        Block::new()
            .borders(Borders::TOP | Borders::BOTTOM)
            .merge_borders(strategy)
            .render(buffer.area, &mut buffer);
        #[rustfmt::skip]
        let expected = Buffer::with_lines([
            "──────────",
            "          ",
            "──────────",
        ]);
        assert_eq!(buffer, expected);

        let mut buffer = Buffer::empty(Rect::new(0, 0, 10, 3));
        Block::new()
            .borders(Borders::LEFT | Borders::RIGHT)
            .merge_borders(strategy)
            .render(buffer.area, &mut buffer);
        #[rustfmt::skip]
        let expected = Buffer::with_lines([
            "│        │",
            "│        │",
            "│        │",
        ]);
        assert_eq!(buffer, expected);
    }
```

ratatui-widgets/src/block.rs
j-g00da: https://github.com/ratatui/ratatui/pull/1874#discussion_r2125599985

joshka: good catch - thanks

orhun (approved):
LGTM

joshka:
Updated commit message was force pushed in 671c2b4fd4ee277ba0dc36d152d66c40cfa1f030

joshka:
Thanks @pauladam94 for kicking this off, and @j-g00da for jumping on it and doing a bunch of work to get it over the line.

PR #1882, opened 2025-05-30
refactor: remove the style field from ListItem and delegate it to the inner Text instead

janTatesa:
I think it's unnecessary to have two separate styles. So styling `ListItem` will actually style the inner text

### DIFF
--- a/ratatui-widgets/src/list/item.rs
+++ b/ratatui-widgets/src/list/item.rs
@@ -225,21 +225,21 @@ mod tests {
     #[test]
     fn new_from_str() {
         let item = ListItem::new("Test item");
-        assert_eq!(item.content, Text::from("Test item"));
+        assert_eq!(item.0, Text::from("Test item"));
         assert_eq!(item.0.style, Style::default());
     }
 
     #[test]
     fn new_from_string() {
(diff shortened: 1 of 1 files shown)


joshka:
This will cause setting style directly on the item to overwrite all styles rather than effectively combine the styles.

This test will fail:
```rust
#[test]
fn combines_style() {
    let bold = Text::styled("foo", Modifier::BOLD);
    let item = ListItem::new(bold).style(Modifier::ITALIC);
    assert_eq!(
        item,
        ListItem::new(Text::styled("foo", Modifier::BOLD | Modifier::ITALIC))
    );
}
```

Calling stylize methods on the ListItem will work, because they get the style, combine it and then set the style.

Additionally, the space allocated for the highlight symbol as well as any empty space after the text will also now be styled using the text's style where previously this would be only styled using the list item's style. This is a change which would be breaking to various UIs that assume that this is not how things work.

I'm pretty sure there isn't a resolution to both those points. I mentioned the second point in the related issue (not sure if you caught it https://github.com/ratatui/ratatui/issues/1692#issuecomment-2920464144)

joshka:
See https://github.com/ratatui/ratatui/pull/1884 for added tests

Issue #1877, opened 2025-05-29
Add Serialize, Deserialize to more structs


aurreland:
## Problem

My problem is i want to make a dynamic structure for my app, where user could in the config file define like LayoutNode:
```rust
#[derive(Debug, Clone, Deserialize)]
#[serde(tag = "type")]
pub enum LayoutNode {
    Vertical {
        constraints: Vec<Constraint>,
        children: Vec<LayoutNode>,
    },
    Horizontal {
        // Not possible as Constraint does not implement Deserialize
        constraints: Vec<Constraint>, 
        children: Vec<LayoutNode>,
    },
    Widget {
        name: String,
        #[serde(default)]
        options: WidgetOptions
    },
}

#[derive(Debug, Default, Clone, Deserialize)]
pub struct WidgetOptions {
    // Same as before Block does not implement Deserialize
    pub block: Option<Block>
}
```
But I can't deserialize/serialize simple structs like Constraint that is just an enum. I then need to either implement Serialize/Deserialize manually which is no good or implement a parser function.

## Solution

The solution would to make public struct that can derive Serialize/Deserialize, implement these when the feature is enabled.

joshka:
`Constraint` should definitely be serializable.
Are there any other specific types that are on your list?

aurreland:
WIP list of struct it could be useful :
- Constraint
- Block
- (Scrollbar)

joshka:
One problem with introducing Serialization into widgets is that this inherently adds a cross version semver problem. E.g. if we update block to add or remove a field, then we cause problems with serializations done against previous versions. That in turn adds to the maintenance burden.

I don't mind adding Constraint. That's fairly stable (albeit, I think we need to add a `#[non_exhaustive]` attribute to this for 0.30, but that's just a side note - we need to do this across the board a bit.

But I know that we plan to break Block due to the way titles are currently added - we want to deprecate the Title struct and just use Line for that (because it already has the positioning info built-in).
Scrollbar is probably something that will have some churn too in comparison to all the other widgets. It's one of the newest widgets, so it tends to have some things which we want to fix / change / improve the most.

aurreland:
I think it should be good to first do Constraint and BorderType which are both enums, and BorderType could be a good way to start for Block. 

Adressing widgets I think it's ok to let users do it themselves I think, but simple types like enums should be covered.

aurreland:
I'm starting to add the derives.

Though I have a question : 
[...]

PR #1883, opened 2025-05-31
feat(serde): derive Serialize/Deserialize for additional structs/enums

aurreland:
### Summary

This PR adds `#[derive(Serialize, Deserialize)]` to the following structs:
- `Constraint`
- `Direction`
- `Spacing`
- `Layout`
- `AccentedPalette`
- `NonAccentedPalette`
- `Palette`
- `Padding`
- `Borders`
- `BorderType`
- `ListDirection`
- `ScrollbarOrientation`
- `ScrollDirection`
- `RenderDirection`
- `HighlightSpacing`

Fixes #1877

### DIFF
--- a/ratatui-core/src/layout/constraint.rs
+++ b/ratatui-core/src/layout/constraint.rs
@@ -46,6 +46,7 @@ use strum::EnumIs;
 /// let constraints = Constraint::from_fills([1, 2, 1]);
 /// ```
 #[derive(Debug, Clone, Copy, Eq, PartialEq, Hash, EnumIs)]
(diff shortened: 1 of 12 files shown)


joshka:
> This might be leading to breaking changes in deserialization in the future due to changed fields in the structs but it should be fine as long as we're careful + document them well.

Yeah, I was thinking the same. Likely what we should do (at least for -core) is be a bit better about marking things as non_exhaustive.

joshka (approved):

joshka:
Thanks for the PR

PR #1884, opened 2025-05-31
chore: add more tests to cover how list styles are combined

joshka:


ratatui-widgets/src/list.rs
```diff
@@ -492,4 +496,132 @@ mod tests {
...
+        let list = List::new([item])
+            .highlight_symbol(">>")
+            .highlight_spacing(HighlightSpacing::Always);
+        let mut buffer = Buffer::empty(Rect::new(0, 0, 10, 1));
+
+        list.render(buffer.area, &mut buffer, &mut ListState::default());
+
+        assert_eq!(
+            buffer,
+            Buffer::with_lines([Line::from(vec!["  ".italic(), "Item 1  ".bold().italic(),])])
```
orhun:
```suggestion
            Buffer::with_lines([Line::from(vec!["  ".italic(), "Item 1  ".bold().italic()])])
```

orhun (approved):
Cool, I'm glad that everything works as expected :D

PR #1891, opened 2025-06-03
fix: Change order of operations for length calculation of pixel buffers fo…

Daksh14:
Fixes: https://github.com/ratatui/ratatui/issues/1449 

The issue 1449 was about trying to render ratatui canvas on a very small font size and terminal scaled `Ctrl+-` all the way down, because of the high value of integers, which caused some integer overflows.

NOTE: We have to double check if the current behavior by the canvas is valid under the scaled down circumstance currently of we want to change the way canvas grid works?

### DIFF
--- a/ratatui-widgets/src/canvas.rs
+++ b/ratatui-widgets/src/canvas.rs
@@ -967,11 +967,11 @@ mod tests {
         let mut b_grid = BrailleGrid::new(u16::MAX, 3);
         let mut c_grid = CharGrid::new(u16::MAX, 3, 'd');
 
-        let max_u16 = u16::MAX as usize;
+        let max = u16::MAX as usize + 10;
 
         // see if we can paint outside bounds
-        b_grid.paint(max_u16 + 10, max_u16 + 10, Color::Red);
-        c_grid.paint(max_u16 + 10, max_u16 + 10, Color::Red);
+        b_grid.paint(max, max, Color::Red);
+        c_grid.paint(max, max, Color::Red);
         // see if we can paint usize max bounds
         b_grid.paint(usize::MAX, usize::MAX, Color::Red);
         c_grid.paint(usize::MAX, usize::MAX, Color::Red);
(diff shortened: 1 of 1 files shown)


ratatui-widgets/src/canvas.rs
```diff
@@ -937,4 +940,17 @@ mod tests {
...
     }
+
+    // The canvas methods work a lot with arithmetic so here we enter various width and height
+    // values to check if there are any integer overflows we just initialize the canvas painters
+    #[test]
+    fn check_integer_overflow_grids() {
+        let max = u16::MAX;
+        let mut b_grid = BrailleGrid::new(max, 2);
+        let mut c_grid = CharGrid::new(max, 2, 'd');
+        let max = usize::MAX;
```
joshka: reassigning max here makes the test a little confusing, it would be better to instead use u16::MAX / usize::MAX directly inline on the params.

ratatui-widgets/src/canvas.rs
```diff
@@ -937,4 +940,17 @@ mod tests {
...
+
+        b_grid.paint(max, max, Color::Red);
+        c_grid.paint(max, max, Color::Red);
```
joshka: These would be generally out of grid regardless of the size of the grid dimensions (because usize::MAX > u16::MAX). So it seems like this test is testing a couple of things at once, but not being explicit about it:
- That we can create grids that support u16::MAX * u16::MAX
- That we can paint to characters outside of the grid dimensions without panicking.

I think it would be best to treat those two things as distinct things to check. So how about changing this test so that it checks that the size supports u16::MAX*u16::MAX for each of the grid types and that we can write to (0, 0), (0, max), (max, 0), (max, max)? And then adding another test which checks that we don't fail when painting to (u16::MAX + 1, u16::MAX + 1) and (usize::MAX, usize::MAX)? WDYT?

Daksh14: @joshka Sounds reasonable, I will seperate those tests

joshka:
```
[2:56 PM]weegee: @joshka Constructing buffer of size u16::MAX*u16::MAX will take its sweet time, slow the CI also 
[2:57 PM]weegee: so I just did u16::MAX * 2 to check overflow of u16
[2:57 PM]joshka: how long?
[2:57 PM]weegee: longer than 60 seconds
[2:58 PM]joshka: ah ok - yeah that's no good. *2 is fine then
```

joshka (approved):

PR #1893, opened 2025-06-03
feat: block shadows

### LINKED ISSUE #1892: Block shadows
## Problem
It would be nice to have a built-in support for shadows in Ratatui.
This is a pretty common thing, that would be great for example in popups, floating windows etc.
Examples in existing UI:
- popups in AMI BIOS
- raspi-config

## Solution
Introduce:
- `Shadow` struct used for configuring a shadow,
- `ShadowType` enum for different shadow types,
- `Block::shadow(shadow: Shadow)` method for adding shadow to `Block`

### Shadow types
TBD

### API
```rust
Block::bordered().shadow(Shadow::default());
```
```
┌Box──┐
│     │█
└─────┘█
 ███████
```
```rust
Block::bordered().shadow(Shadow::default().offset(Offset::new(-1, 1)));
```
```
 ┌Box──┐
█│     │
█└─────┘
███████
```
```rust
Block::bordered().shadow(Shadow::new(ShadowType::LightShade));
```
```
┌Box──┐
│     │░
└─────┘░
 ░░░░░░░
```
```rust
Block::bordered().shadow(Shadow::new(ShadowType::DoubleBorder));
```
```
┌Box──┐
│     │╗
└─────┘║
 ╚═════╝
```
We also probably want to add option to set style:
```rust
Shadow::default().style(...);
```
and some shortcut constructors:
```rust
Shadow::light_shade();
```

Let me know what you think and what border types would you want to see. I will start working on it right after border merging lands on main.

j-g00da:
Resolves #1892 

Image
Image
Image
Image


j-g00da:
I've pushed a very initial, dirty PoC. Works on blocks only - this will change _a lot_.

orhun:
I like the shade options, looks like magic :D

ratatui-widgets/src/block/shadow.rs
```diff
@@ -0,0 +1,137 @@
...
+        let symbol = self.shadow_type.get_char();
+
+        for y in shadow_area.top()..shadow_area.bottom() {
+            for x in shadow_area.left()..shadow_area.right() {
+                if base_area.contains(Position { x, y }) {
+                    continue;
+                }
+                if symbol != ' ' {
+                    buf[(x, y)].set_char(symbol);
+                }
```
joshka: Would it be worth making symbol `Option<String>` here?

j-g00da: This will disappear anyway, all single char fills should probably be under `ShadowType::Fill(&'static str)` not to mix up with other types.

`ShadowType::LightShade` -> `ShadowType::Fill(ratatui::symbols::shade::LIGHT)`
`ShadowType::Custom('#')` -> `ShadowType::Fill("#")`
etc.

This check won't be needed as this will be handled by some fill method.

ratatui-widgets/src/block/shadow.rs
```diff
@@ -0,0 +1,144 @@
...
+    };
+    (|$x:ident: u16, $y:ident: u16, $buf:ident: &mut Buffer| $body:expr) => {
+        $crate::cell_filter!(|$x: u16, $y: u16, $buf: &mut Buffer| { $body; })
+    };
+
+    // fill
+    ($s:expr) => {
+        $crate::cell_filter!(|x: u16, y: u16, buf: &mut Buffer| buf[(x, y)].set_symbol($s))
+    };
+}
```
j-g00da, line 132: This is WIP and can be improved, but works. Just wanted to show this PoC.

ratatui-widgets/src/block/shadow.rs
```diff
@@ -0,0 +1,144 @@
...
+
+    /// TODO: docs
+    pub fn medium_shade() -> Self {
+        Self::new(crate::cell_filter!(shade::MEDIUM))
+    }
+
+    /// TODO: docs
+    pub fn dark_shade() -> Self {
+        Self::new(crate::cell_filter!(shade::DARK))
+    }
```
j-g00da, line 48: Not sure if we want these at all, IMO just clutters the API.

ratatui-widgets/src/block/shadow.rs
```diff
@@ -0,0 +1,144 @@
...
+            filter,
+            style: Style::default(),
+            offset: Offset::new(1, 1),
+        }
+    }
+
+    /// TODO: docs
+    pub fn overlay() -> Self {
+        Self::new(|_, _, _| {})
+    }
```
j-g00da, line 28: This only makes sense with `.style(...)`, so maybe we can move this to `Default` and change this to 
```suggestion
    /// TODO: docs
    pub fn overlay<S: Into<Style>>(style: S) -> Self {
        Self::default().style(style)
    }
```

I'm also not sure about the name "overlay".

j-g00da (commented):
Left some comments, WDYT? @joshka @orhun 
Related: https://github.com/ratatui/ratatui/issues/1892#issuecomment-2959240719

orhun:
This is looking pretty good tbh, should we finalize it and include it in the release?

j-g00da:
> This is looking pretty good tbh, should we finalize it and include it in the release?

Right now it's just a PoC, I would have to clean it up, move stuff around and add some tests and docs but can be done in 1 week probably.

cc: @joshka

ratatui-widgets/src/block/shadow.rs
```diff
@@ -0,0 +1,144 @@
+use ratatui_core::buffer::Buffer;
+use ratatui_core::layout::{Offset, Position, Rect};
+use ratatui_core::style::{Color, Modifier, Style};
+use ratatui_core::symbols::shade;
+use ratatui_core::widgets::Widget;
+
+/// TODO: docs
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
```
orhun, line 8: Implementing `PartialEq` gives a lint warning:

> --> ratatui-widgets/src/block/shadow.rs:31:5
|
29 | `#[derive(Debug,` Clone, Copy, PartialEq, Eq, Hash)]
| --------- in this derive macro expansion
30 | pub struct Shadow {
31 | filter: fn(Rect, Rect, &mut Buffer),
| ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
|
= note: the address of the same function can vary between different codegen units
= note: furthermore, different functions could have the same address after being merged together
= note: for more information visit <https://doc.rust-lang.org/nightly/core/ptr/fn.fn_addr_eq.html>



I think we need to implement it separately without the filter function:

```rust
// Define equality without comparing the function pointer.
impl PartialEq for Shadow {
    fn eq(&self, other: &Self) -> bool {
        self.style == other.style && self.offset == other.offset
    }
}

// Hash only stable fields, without the function pointer.
impl Hash for Shadow {
    fn hash<H: Hasher>(&self, state: &mut H) {
        self.style.hash(state);
        self.offset.hash(state);
    }
}
```

orhun (commented):
Just had a look at this PR while trying out RustRover and did a small thing

orhun:
Any chance we could revive this for 0.30.1 @j-g00da ?

j-g00da:
> Any chance we could revive this for 0.30.1 @j-g00da ?

I hope so, but it's hard for me to guarantee anything currently :/

orhun:
I gave this a shot in #2481

orhun:
implemented in #2481

PR #1896, opened 2025-06-04
docs(breaking-changes): change MSRV for v0.30.0 to 1.85

j-g00da:
1. Should I remove this? Ratatui only supports the latest version anyway, 1.81 >= might work but it's not guaranteed anyway.
> This is due to the use of `#[expect]` in the codebase, which is only available in Rust 1.81.0 and later.`
2. If so, should the linked PR be changed to #1860?


orhun (approved):

joshka:
Yeah, change the link to the other PR.

j-g00da:
> Yeah, change the link to the other PR.

Done

PR #1899, opened 2025-06-06
docs(block): add collapsed border example

joshka:


joshka:
There's a bit of a problem with this:
Given the following, where top is using thick borders, but is drawn before bottom, then bottom overwrites:
```rust
frame.render_widget(left_block, left);
frame.render_widget(top_block, top);
frame.render_widget(bottom_block, bottom);
```
image

If we flip it so bottom is written before top, the border is fine, but the heading is a problem:

```rust
frame.render_widget(left_block, left);
frame.render_widget(bottom_block, bottom);
frame.render_widget(top_block, top);
```
image

Perhaps the merge borders function needs a way to allow normal printable characters to not be overwritten?

@j-g00da @pauladam94 any thoughts here?

orhun:
Yes, I think the solution would be to only collapse border characters and don't touch the normal characters such as title.

Or we don't change the implementation because the overlap actually makes sense... the second example can be fixed by setting a bottom left title on the top block.

j-g00da:
> Yes, I think the solution would be to only collapse border characters and don't touch the normal characters such as title.

This can cause weird glitches (holes in the border) when rendering a border over any text.

joshka:
> Or we don't change the implementation because the overlap actually makes sense... the second example can be fixed by setting a bottom left title on the top block.

So the idea here is to attach titles to the blocks that they're relevant to. Sure the title is at the bottom of the top block, but moving it there when you've got a more complex set of panes is the complexity that we're trying to avoid here. It would mean that each pane no longer just cares about how to render itself, it has to know about the other panes in the app in order to render correctly.

> This can cause weird glitches (holes in the border) when rendering a border over any text.

Oh yeah, spaces in the titles would be handled weirdly :/

pauladam94:
The standard in GUIs is to render on top whatever is rendered last (or have a depth parameter to control specificaly that). The merging characters was a attempt to be closer to what happened on GUIs where rendering multiple character should just merge together them.

Conclusion it is normal that the text is overwritten in the second case.

Here the problem comes from that "horizontal line" merged with "horizontal line thick" don't comes up thick in the first example 🤔

I think it should comes up thick and we can fix that by changing a bit the merging algorithm.

joshka:
So my general rule for Ratatui is that whatever is rendered last wins, but having the merge strategy there allows us to modify that a bit so we can easily make UIs that work well. There's a couple of options for making this work well I think.
- Add a strategy that doesn't overwrite text.
- Add a strategy that preferences "thicker" over "thinner" symbols

I think the second option fails a bit in that it makes styling more difficult to get right (e.g, I want to mark the active pane with a thick red border), but the first option should probably work ok.

The question I have is, should this be just part of the existing exact / fuzzy strategies or a new strategy? I think it's probably new, and we add more descriptive fallbacks. Perhaps renaming the existing variants to MergeBorderSymbolsExact or something like that?

j-g00da:
I have two solutions, with a different approach:
1. Provide a way to modify existing borders. You would draw all of these as plain, and then if you would like to make one of these "selected", you would call something like `change_border(area, buf, Thick, Exact);` and it would change all the border symbols on the perimeter of the area.
2. Make it possible to render a title even when there is no border (if this isn't already possible), this way you would just render the bottom right block with no top border.

joshka:
The problem with both those approaches is that it requires rendering out of order, which puts significant burden on the app side rather than the Ratatui side of this. The ideal solution to my understanding requires not too much modification other than reordering things from the app side.

I think part of the underlying problem here is that we don't distinguish empty cells (which have never been written), from cells with a space character in them (which have). If we did, then we could merge `character + border = character`, `space + border = space`, and `empty + border = border` when rendering the overwriting border. This might be possible to fix while maintaining the same buffer idea. I'm not 100% certain though.

We're a bit limited in comparison to other libraries (e.g textual) here, which understand z-indexes when rendering, allowing a screen to be compositied from multiple layers (the borders being lower, the titles being higher). This isn't something we can fix without fully revamping how rendering works. So fixing this sort of thing would be something to think if we ever made a Ratatui2 where we rewrite everything with all the things we've learnt, and break all apps.

j-g00da:
> I think part of the underlying problem here is that we don't distinguish empty cells (which have never been written), from cells with a space character in them (which have). If we did, then we could merge character + border = character, space + border = space, and empty + border = border when rendering the overwriting border. This might be possible to fix while maintaining the same buffer idea. I'm not 100% certain though.

I like this one a lot, best solution so far IMO. The behavior makes sense and it wouldn't clutter the API.

j-g00da:
Opened #1939 for this.

joshka:
Made with VHS

ratatui-widgets/examples/collapsed-borders.rs
```diff
@@ -0,0 +1,115 @@
...
+                .title(title)
+                .merge_borders(MergeStrategy::Exact);
+            frame.render_widget(block, area);
+        }
+    }
+    // Render the selected pane last (so it appears on top) with a thick border
+    if let Some(&(area, title)) = panes.get(&selected_pane) {
+        let block = Block::bordered()
+            .title(title)
+            .merge_borders(MergeStrategy::Exact)
```
j-g00da, line 111: Nit: Maybe we can add some comment here, to highlight that this is the line that makes it work. It may be easy to miss.

joshka: done

orhun (approved):

PR #1900, opened 2025-06-07
feat: use CoW for List and Table contents

cgzones:
Use a Copy-on-Write type for the contents of tables and lists so their content can be cached and only recomputed on actual changes.  This avoids unnecessary computations for the rendering in case of unchanged data.

Related to ratatui/ratatui-table#5.

Contains first commit of ratatui/ratatui#1791, with `'lend` and `'data` lifetimes combined into `'a`.

### DIFF
--- a/ratatui-widgets/src/list.rs
+++ b/ratatui-widgets/src/list.rs
@@ -27,6 +27,9 @@ mod state;
 ///
 /// [`Table`]: crate::table::Table
 ///
+/// List items can be created on the stack with the macro `list_items!` and used via
+/// [`List::from`] to avoid dynamic allocations.
+///
 /// List items can be aligned using [`Text::alignment`], for more details see [`ListItem`].
 ///
 /// [`List`] is also a [`StatefulWidget`], which means you can use it with [`ListState`] to allow
@@ -73,15 +76,18 @@ mod state;
 /// ```rust
 /// use ratatui::Frame;
 /// use ratatui::layout::Rect;
+/// use ratatui::macros::list_items;
 /// use ratatui::style::{Style, Stylize};
 /// use ratatui::widgets::{Block, List, ListState};
 ///
 /// # fn ui(frame: &mut Frame) {
 /// # let area = Rect::default();
+///
 /// // This should be stored outside of the function in your application state.
 /// let mut state = ListState::default();
-/// let items = ["Item 1", "Item 2", "Item 3"];
-/// let list = List::new(items)
+/// let items = list_items!["Item 1", "Item 2", "Item 3"];
+///
+/// let list = List::from(&items)
 ///     .block(Block::bordered().title("List"))
 ///     .highlight_style(Style::new().reversed())
 ///     .highlight_symbol(">>")
@@ -154,19 +160,21 @@ impl<'a> List<'a> {
     /// From a slice of [`&str`]
     ///
     /// ```
+    /// use ratatui::macros::list_items;
     /// use ratatui::widgets::List;
     ///
-    /// let list = List::new(["Item 1", "Item 2"]);
+    /// let list = List::new(list_items!["Item 1", "Item 2"]);
     /// ```
     ///
     /// From [`Text`]
     ///
     /// ```
+    /// use ratatui::macros::list_items;
     /// use ratatui::style::{Style, Stylize};
     /// use ratatui::text::Text;
     /// use ratatui::widgets::List;
     ///
-    /// let list = List::new([
+    /// let list = List::new(list_items![
     ///     Text::styled("Item 1", Style::new().red()),
     ///     Text::styled("Item 2", Style::new().red()),
     /// ]);
@@ -434,11 +442,10 @@ impl<'a> List<'a> {
     /// # Example
     ///
     /// ```
-    /// use ratatui::widgets::{List, ListItem};
+    /// use ratatui::widgets::List;
     ///
-    /// let items = [ListItem::new("Item 1"), ListItem::new("Item 2"), ListItem::new("Item 3")];
-    /// let list = List::from(&items);
-    /// assert!(list.is_borrowed());
+    /// let list = List::new(["Item 1", "Item 2", "Item 3"]);
+    /// assert!(list.is_owned());
     /// ```
     pub const fn is_owned(&self) -> bool {
         match self.items {
@@ -449,13 +456,27 @@ impl<'a> List<'a> {
 
     /// Returns whether the list borrows its items.
     ///
-    /// # Example
+    /// # Examples
     ///
     /// ```
+    /// use ratatui::widgets::{List, ListItem};
+    ///
+    /// let items = [
+    ///     ListItem::new("Item 1"),
+    ///     ListItem::new("Item 2"),
+    ///     ListItem::new("Item 3"),
+    /// ];
+    /// let list = List::from(&items);
+    /// assert!(list.is_borrowed());
+    /// ```
+    ///
+    /// ```
+    /// use ratatui::macros::list_items;
     /// use ratatui::widgets::List;
     ///
-    /// let list = List::new(["Item 1", "Item 2", "Item 3"]);
-    /// assert!(list.is_owned());
+    /// let items = list_items!["Item 1", "Item 2", "Item 3"];
+    /// let list = List::from(&items);
+    /// assert!(list.is_borrowed());
     /// ```
     pub const fn is_borrowed(&self) -> bool {
         !self.is_owned()
--- a/ratatui-widgets/src/list.rs
+++ b/ratatui-widgets/src/list.rs
@@ -1,6 +1,7 @@
 //! The [`List`] widget is used to display a list of items and allows selecting one or multiple
 //! items.
 
+use alloc::borrow::Cow;
 use alloc::vec::Vec;
 
 use ratatui_core::style::{Style, Styled};
@@ -105,12 +106,12 @@ mod state;
 /// [`Text::alignment`]: ratatui_core::text::Text::alignment
 /// [`StatefulWidget`]: ratatui_core::widgets::StatefulWidget
 /// [`Widget`]: ratatui_core::widgets::Widget
-#[derive(Debug, Clone, Eq, PartialEq, Hash, Default)]
+#[derive(Debug, Clone, Default, Eq, Hash, PartialEq)]
 pub struct List<'a> {
     /// An optional block to wrap the widget in
     pub(crate) block: Option<Block<'a>>,
     /// The items in the list
-    pub(crate) items: Vec<ListItem<'a>>,
+    pub(crate) items: Cow<'a, [ListItem<'a>]>,
     /// Style used as a base style for the widget
     pub(crate) style: Style,
     /// List display direction
@@ -182,6 +183,7 @@ impl<'a> List<'a> {
     /// ```
     ///
     /// [`Text`]: ratatui_core::text::Text
+    #[must_use = "constructor"]
     pub fn new<T>(items: T) -> Self
     where
         T: IntoIterator,
@@ -426,6 +428,38 @@ impl<'a> List<'a> {
     pub fn is_empty(&self) -> bool {
         self.items.is_empty()
     }
+
+    /// Returns whether the list owns its items.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use ratatui::widgets::{List, ListItem};
+    ///
+    /// let items = [ListItem::new("Item 1"), ListItem::new("Item 2"), ListItem::new("Item 3")];
+    /// let list = List::from(&items);
+    /// assert!(list.is_borrowed());
+    /// ```
+    pub const fn is_owned(&self) -> bool {
+        match self.items {
+            Cow::Owned(_) => true,
+            Cow::Borrowed(_) => false,
+        }
+    }
+
+    /// Returns whether the list borrows its items.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use ratatui::widgets::List;
+    ///
+    /// let list = List::new(["Item 1", "Item 2", "Item 3"]);
+    /// assert!(list.is_owned());
+    /// ```
+    pub const fn is_borrowed(&self) -> bool {
+        !self.is_owned()
+    }
 }
 
 impl Styled for List<'_> {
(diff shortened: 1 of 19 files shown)


ratatui-widgets/src/list.rs
```diff
@@ -142,6 +145,41 @@ pub enum ListDirection {
...
+///     .block(Block::bordered().title("List"))
+///     .highlight_style(Style::new().reversed())
+///     .highlight_symbol(">>")
+///     .repeat_highlight_symbol(true);
+///
+/// frame.render_widget(list, area);
+/// # }
+/// ```
+#[macro_export]
+macro_rules! items {
```
joshka: How about we call this `list_items` (just from an ambiguity perspective)?

ratatui-widgets/src/list.rs
```diff
@@ -142,6 +145,41 @@ pub enum ListDirection {
...
+/// # let area = Rect::default();
+/// let items = items!["Item 1", "Item 2", "Item 3"];
+/// let list = List::from(&items)
+///     .block(Block::bordered().title("List"))
+///     .highlight_style(Style::new().reversed())
+///     .highlight_symbol(">>")
+///     .repeat_highlight_symbol(true);
+///
+/// frame.render_widget(list, area);
+/// # }
```
joshka: This example could be a bit shorter, it doesn't really have to show how the items are used. Mainly it should show how the macro is used. So it should highlight any things that are interesting about how this facilitates Cow semantics (or not) and how it handles conversions to list item.

It's sort of similar to how the `vec!` macro works, should there be some syntax that accepts `; n`?

ratatui-widgets/src/list.rs
```diff
@@ -418,11 +457,15 @@ impl<'a> List<'a> {
     }
 
     /// Returns the number of [`ListItem`]s in the list
+    #[must_use]
+    #[allow(clippy::missing_const_for_fn)]
```
joshka: Why does this lint trigger? Wse `#[expect]` with some info about the problem if it's not fixable,or fix the problem if it is. (and same question below)

ratatui-widgets/src/list.rs
```diff
@@ -73,14 +74,16 @@ mod state;
...
+/// use ratatui::widgets::{Block, List, ListState, items};
 ///
 /// # fn ui(frame: &mut Frame) {
 /// # let area = Rect::default();
+///
 /// // This should be stored outside of the function in your application state.
 /// let mut state = ListState::default();
-/// let items = ["Item 1", "Item 2", "Item 3"];
-/// let list = List::new(items)
+/// let items = items!["Item 1", "Item 2", "Item 3"];
```
joshka: The macro probably should be introduced in the doc text above somewhere so it's not a surprise in the example.

ratatui-widgets/src/list.rs
```diff
@@ -142,6 +145,41 @@ pub enum ListDirection {
     BottomToTop,
 }
 
+/// Creates a fixed sized array of [`ListItem`]s.
+///
+/// Can be used together with `List::from` to only generate the [`ListItem`]s if
+/// any changed instead of each time the [`List`] is rendered.
```
joshka: nit: try to extend doc comments all the way to 100 chars (not 80) for consistency with the rest of the library.

ratatui-widgets/src/list.rs
```diff
@@ -477,14 +566,52 @@ mod tests {
...
+        let items = [
+            ListItem::from("Item0"),
+            ListItem::from("Item1"),
+            ListItem::from("Item2"),
+        ];
+        let expected = List::new(items);
+        assert!(matches!(collected.items, Cow::Owned(_)));
+        assert!(matches!(expected.items, Cow::Owned(_)));
+        assert_eq!(collected, expected);
+    }
```
joshka, line 583: This test got a whole lot more complex and it's now much less clear what it's testing. Is it testing the items macro, the from_iter(), ...? Try to rewrite this in a way where just one property of the list is being tested and there's one obvious way that this could fail. I.e. for good tests, thing not: "you can use List like this and like this, and this is the same", but instead: "when you use list like this, this is the result". Don't be afraid to add more than one test with a good name if that helps make it clear.

ratatui-widgets/src/list.rs
```diff
@@ -477,14 +566,52 @@ mod tests {
...
+        assert!(matches!(collected.items, Cow::Owned(_)));
+        assert!(matches!(expected.items, Cow::Owned(_)));
         assert_eq!(collected, expected);
     }
 
     #[test]
     fn can_be_stylized() {
         assert_eq!(
-            List::new::<Vec<&str>>(vec![])
+            List::new::<Vec<ListItem<'_>>>(vec![])
```
joshka: Can this be written more simply with an empty items![] call?

ratatui-widgets/src/list.rs
```diff
@@ -527,6 +654,47 @@ mod tests {
         );
     }
 
+    #[test]
+    fn macro_items() {
```
joshka: Given that the macro has its own tests, perhaps pull it out into a separate file and make each test here smaller.

ratatui-widgets/src/list.rs
```diff
@@ -527,6 +654,47 @@ mod tests {
...
+        let items = items![];
+        let list = List::new(items);
+        assert!(matches!(list.items, Cow::Owned(_)));
+        assert_eq!(*list.items, []);
```
joshka: Be generally wary of using `assert!` followed by `assert_eq!` in tests. If this fails, then there's no information about why it failed because the initial failure is against a binary condition. So this hampers developing a fix without also first modifying the tests to show more information.

Instead if you can, go straight to an assert_eq! test. Here `assert_eq!(list.items, Cow::Owned([]))` or something like that. (and below)

ratatui-widgets/src/list.rs
```diff
@@ -527,6 +654,47 @@ mod tests {
...
+
+        let items = items!["Item0"];
+        let list = List::new(items);
+        assert!(matches!(list.items, Cow::Owned(_)));
+        assert_eq!(
+            *list.items,
+            [ListItem {
+                content: Text::from(Line::from("Item0")),
+                style: Style::new(),
+            }]
```
joshka: In general, test should avoid having too much logic and should be fairly linear and understandable without having to read other parts. The one place I'd make an exception is that when you have small obvious data like types like ListItem is it's worth just using the same value everywhere if it's easy to construct like this.

So perhaps simplify the repetition in these tests by sticking something like `let item0 = ListItem::from("Item0")` somewhere appropriate in the test. Adding this in each test is fine, but you could also put this in an rstest fixture and pass it as a test parameter like so:

```rust
#[fixture]
fn item_0() -> ListItem {
    ListItem::from("Item 0");
}

#[rstest]
fn item_test(item_0: ListItem) {
    // ...
    assert_eq!(list.items, Cow::Owned([item_0));
}
```

I think either would be fine here (a one liner vs a rstest fixture param are effectively equivalent in my perspective of maintenance burden).

ratatui-widgets/src/list/rendering.rs
```diff
@@ -346,8 +346,8 @@ mod tests {
 
     #[test]
     fn does_not_render_in_small_space() {
-        let items = vec!["Item 0", "Item 1", "Item 2"];
-        let list = List::new(items.clone()).highlight_symbol(">>");
+        let items = items!["Item 0", "Item 1", "Item 2"];
+        let list = List::from(&items).highlight_symbol(">>");
```
joshka: For the changes in this file, are these made due to necessity or from a perspective of cleaner code?

I have a general view that when writing unit tests, it's usually better to keep them simple and exercise the least amount of the code as possible (i.e. avoiding the macro here). I'm not sure what the rationale is here for changing to the macro over plain arrays / vecs. Can you add some detail?

Perhaps it would be better to make these generally use `List::from(["item 0", "item 1"])` where necessary.

ratatui-widgets/src/table.rs
```diff
@@ -271,10 +272,44 @@ pub struct Table<'a> {
     flex: Flex,
 }
 
+/// Creates a fixed sized array of [`Row`]s.
```
joshka: Let's move the macro to its own file (simplify testing etc.)

ratatui-widgets/src/table.rs
```diff
@@ -271,10 +272,44 @@ pub struct Table<'a> {
...
+///
+/// Can be used together with `Table::from` to only generate the [`Row`]s if
+/// any changed instead of each time the [`Table`] is rendered.
```
joshka: I don't understand this comment. It's difficult to parse without context, so it might be worth describing the scenario / problem this solves and what this means a bit better.

nit: Line length (wrap at 100 instead of 80)

ratatui-widgets/src/table.rs
```diff
@@ -271,10 +272,44 @@ pub struct Table<'a> {
...
+/// ];
+/// let table = Table::from(&rows).widths(widths);
+/// ```
+#[macro_export]
+macro_rules! rows {
+    () => (
+        ([] as [$crate::table::Row<'_>; 0])
+    );
+    ($([$($x:expr),+ $(,)?]),+ $(,)?) => (
+        // TODO: avoid to_vec()
```
joshka: Why?

ratatui-widgets/src/table.rs
```diff
@@ -1048,14 +1128,16 @@ mod tests {
...
     use crate::table::Cell;
 
     #[test]
     fn new() {
         let rows = [Row::new(vec![Cell::from("")])];
         let widths = [Constraint::Percentage(100)];
         let table = Table::new(rows.clone(), widths);
-        assert_eq!(table.rows, rows);
+        assert!(matches!(table.rows, Cow::Owned(_)));
+        assert_eq!(*table.rows, rows);
```
joshka: Same notes as above about assert! and then assert_eq!

ratatui-widgets/src/table.rs
```diff
@@ -1231,6 +1315,148 @@ mod tests {
         assert_eq!(table.widths, [Constraint::Percentage(100)], "vec ref");
     }
 
+    #[test]
+    fn macro_rows() {
```
joshka: Split this into multiple tests and move to a new file.

ratatui-widgets/src/table/row.rs
```diff
@@ -73,13 +74,37 @@ use super::Cell;
...
     pub(crate) height: u16,
     pub(crate) top_margin: u16,
     pub(crate) bottom_margin: u16,
     pub(crate) style: Style,
 }
 
+/// Creates a fixed sized array of [`Cell`]s.
+///
+/// Can be used together with `Row::from` to only generate the [`Cell`]s if
+/// any changed instead of each time the [`Row`] is rendered.
```
joshka: Same comments as above (new file, comment length, explain the context)

ratatui-widgets/src/table/row.rs
```diff
@@ -301,6 +371,38 @@ mod tests {
         assert_eq!(row.cells, cells);
     }
 
+    #[test]
+    fn from_vec() {
+        let cells = vec![Cell::from("")];
+        let row = Row::from(cells);
+        assert!(matches!(row.cells, Cow::Owned(_)));
+        assert_eq!(*row.cells, [Cell::new("")]);
```
joshka: Same comments as above for assert then assert_eq

joshka (commented):
For the macros, it could be worth putting these in the ratatui-macros project which has some stuff for table rows already. WDYT?

https://docs.rs/ratatui-macros/latest/ratatui_macros/#row

I think in general, this idea seems fine. The main question I'd have are about how this affects existing code. I'm assuming that it should continue to work fine, but there's lots of changes to constructor type code which I'm unsure whether it was due to necessity or preference. Leaving a few clear tests that show the existing call patterns being undisturbed would generally allay that concern.

ratatui-widgets/src/table.rs
cgzones: reworded

ratatui-widgets/src/table.rs
cgzones: to avoid the allocation in `.to_vec()`

ratatui-widgets/src/list.rs
cgzones: Since `ListItem` does not implement copy, the trivial implementation does not work

cgzones:
Thanks for the review.
Rebased and suggestions adopted.

ratatui-widgets/src/list/list_items.rs
orhun, line 1: As Josh pointed out earlier, what do you think about moving this to `ratatui-macros`? I think there it would be more appropriate.

cgzones: When trying to extract the macro I get the following errors:
```
error[E0271]: type mismatch resolving `<[ListItem<'_>] as ToOwned>::Owned == Vec<ListItem<'_>>`
   --> ratatui-widgets/src/list.rs:743:40
    |
743 |         let cow: Cow<[ListItem<'_>]> = Cow::Owned(items);
    |                                        ^^^^^^^^^^^^^^^^^ expected `ratatui::widgets::ListItem<'_>`, found `item::ListItem<'_>`
    |
note: the crate `ratatui_widgets` is compiled multiple times, possibly with different configurations
   --> ratatui-widgets/src/list/item.rs:73:1
    |
 73 | pub struct ListItem<'a> {
    | ^^^^^^^^^^^^^^^^^^^^^^^ this is the found type `item::ListItem`
    |
   ::: /var/home/user/Coding/workspaces/ratatui/ratatui-widgets/src/list/item.rs:73:1
    |
 73 | pub struct ListItem<'a> {
    | ^^^^^^^^^^^^^^^^^^^^^^^ this is the expected type `ratatui::widgets::ListItem`
    |
   ::: ratatui-widgets/src/list.rs:523:9
    |
523 |     use ratatui::macros::list_items;
    |         ------- one version of crate `ratatui_widgets` used here, as a dependency of crate `ratatui_macros`
    = help: you can use `cargo tree` to explore your dependency tree
```

orhun: That looks odd. Are you sure you are using `ratatui-widgets` from the workspace?

ratatui-widgets/src/table/row.rs
```diff
@@ -263,6 +266,42 @@ impl Styled for Row<'_> {
...
+}
+
+impl<'a, const N: usize> From<&'a [Cell<'a>; N]> for Row<'a> {
+    fn from(value: &'a [Cell<'a>; N]) -> Self {
+        Self {
+            cells: Cow::Borrowed(value),
+            ..Self::default()
+        }
+    }
+}
```
hasezoey, line 340: Wouldnt this implementation be already handled by `From<&'a [Cell<'a>]>` above?

ratatui-widgets/src/list/list_items.rs
cgzones: Moved the tests using the new macros into the macro crate.

ratatui-widgets/src/table/row.rs
cgzones: It seems not:

```
error[E0277]: the trait bound `ratatui_widgets::table::Row<'_>: core::convert::From<&[ratatui_widgets::table::Cell<'_>; 0]>` is not satisfied
  --> ratatui-macros/src/row_cells.rs:34:19
   |
34 |         let row = Row::from(&cells);
   |                   ^^^ the trait `core::convert::From<&[ratatui_widgets::table::Cell<'_>; 0]>` is not implemented for `ratatui_widgets::table::Row<'_>`
   |
   = help: the following other types implement trait `core::convert::From<T>`:
             `ratatui_widgets::table::Row<'_>` implements `core::convert::From<&[ratatui_widgets::table::Cell<'_>]>`
             `ratatui_widgets::table::Row<'_>` implements `core::convert::From<&alloc::vec::Vec<ratatui_widgets::table::Cell<'_>>>`
             `ratatui_widgets::table::Row<'_>` implements `core::convert::From<alloc::borrow::Cow<'_, [ratatui_widgets::table::Cell<'_>]>>`
             `ratatui_widgets::table::Row<'_>` implements `core::convert::From<alloc::vec::Vec<ratatui_widgets::table::Cell<'_>>>`
```

joshka:
The Table experiment now lives at https://github.com/ratatui/ratatui-table, with focused successor tracking for its Table portion at https://github.com/ratatui/ratatui-table/issues/12. List-owned work can continue in the main Ratatui repository.

@cgzones, would you be interested in recreating the Table portion as a focused PR against the extracted 0.1.0 baseline? We would also like to offer you opt-in Write access so proposal authors can review and collaborate without waiting on the core Ratatui repository. Releases and protected configuration remain maintainer-controlled; reply here if you would like the invitation.

PR #1907, opened 2025-06-13
perf(ratatui-core): add unified text to prevent unnecessary heap allocation in table and listitem

### LINKED ISSUE #1904: Use smallvec for Line and Text to prevent unnecessary heap allocation
## Problem

Stuff like ListItem or Cell currently accept an `Into<Text>`. If the thing is a Line then you get an unnecessary heap allocation, and if a Span then you have two of them. This can impact performance, as when you have a big table where each cell is a span then you get Rows * Columns * 2 heap allocations, which is a lot of heap allocations for big tables
## Solution

Replace the spans and lines fields of text and line with SmallVec<[Span;1]> and SmallVec<[Line;1]>. Note that [Span;1] doesn't mean that it's elements are arrays of spans of size one, it means that it's elements are spans and 1 is the amount of elements stored on the stack. This is a backwards compatible change as SmallVec provides the same APIs as Vec.
## Alternatives

An alternative would be instead of Text to have an enum
```rust
enum Cell<'a> {
  Span(Span<'a>),
  Line(Line<'a>),
  Text(Text<'a>)  
}
```
Same for list item. But the problem is that this wouldn't apply to external crates that use Text

janTatesa:
Resolves #1904
`Cell` and `ListItem` now use `UnifiedText` which is either a `Span` a `Line` or `Text`. This avoids unnecessary heap allocation. The enum is not yet documented properly, but it is 100% compatible as everything convertible to `Text` is also convertible to `UnifiedText`

### DIFF
--- a/examples/apps/hyperlink/src/main.rs
+++ b/examples/apps/hyperlink/src/main.rs
@@ -13,7 +13,7 @@ use ratatui::DefaultTerminal;
 use ratatui::buffer::Buffer;
 use ratatui::layout::Rect;
 use ratatui::style::Stylize;
-use ratatui::text::{Line, Text};
+use ratatui::text::{Line, UnifiedText};
 use ratatui::widgets::Widget;
 
 fn main() -> Result<()> {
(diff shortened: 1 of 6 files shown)


joshka:
Resolve notes on https://github.com/ratatui/ratatui/issues/1904#issuecomment-2969730279 before considering / reviewing this.

joshka:
Hey, I'm closing this out for now as a solution where the problem isn't something that has a clear definition. The problem with this is that although it presents a neat idea and is generally well written etc., it does so by adding a layer of indirection that increases the complexity and makes things more difficult to understand. The complexity added here has to justify its existence by being the simplest solution to a problem that's worth solving.

It may be in the future that you're definitely right (I'm not saying that this code is wrong or incorrect in any way). It's just that it's not clear that it's needed right now.

PR #1908, opened 2025-06-14
docs(breaking-changes): improve migration guide for `Backend::Error`

j-g00da:
Related: https://github.com/fujiapple852/trippy/pull/1588

### DIFF
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -148,6 +148,32 @@ and implement the `clear_region` method, which no longer has a default implement
 This change was made to provide greater flexibility for custom backends, particularly to remove the
 explicit dependency on `std::io` for backends that want to support `no_std` targets.
 
+If your app or library uses the `Backend` trait directly - for example, by providing a generic
+implementation for many backends - you may need to update the referenced error type.
+
+```diff
+- fn run<B: Backend>(mut terminal: Terminal<B>) -> io::Result<()> {
++ fn run<B: Backend>(mut terminal: Terminal<B>) -> Result<(), B::Error> {
+```
+
+Alternatively, you can explicitly require the associated error to be `std::io::Error`. This approach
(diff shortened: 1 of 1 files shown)


BREAKING-CHANGES.md
```diff
@@ -148,6 +148,32 @@ and implement the `clear_region` method, which no longer has a default implement
...
+
+```diff
+- fn run<B: Backend>(mut terminal: Terminal<B>) -> io::Result<()> {
++ fn run<B: Backend>(mut terminal: Terminal<B>) -> Result<(), B::Error> {
+```
+
+Alternatively, you can explicitly require the associated error to be `std::io::Error`. This approach
+may require fewer changes in user code but is generally not recommended, as it limits compatibility
+with third-party backends. Additionally, the error type used by built-in backends may or may not
+change in the future, making this approach less future-proof compared to the previous one.
```
joshka, line 162: Mostly only libraries should in general be generic over the Backend, apps should mostly be concrete (except in the case of providing an ability to use the TestBackend I guess), so while this is probably correct advice, it might be the sort that's not usually necessary for apps. I think I'd like to see this rephrased mostly targeting libs rather than apps.

I'm not quite sure what that looks like though.

orhun:
I think we can have this more detailed section in the release highlights and keep the breaking changes document simple.

e.g. https://ratatui.rs/highlights/v029/

BREAKING-CHANGES.md
joshka: Actually, I think I'm fine with this.

joshka (approved):

joshka:
> I think we can have this more detailed section in the release highlights and keep the breaking changes document simple.

I'm going to merge this one. We can use it for documentation later. It's better than what was there. Perhaps unnecessary, but we get questions all the time on design decisions, so it's worth leaving this in the doc I think.

PR #1909, opened 2025-06-14
feat: add ergonomic methods for layouting Rects

joshka:
This commit introduces new methods for the `Rect` struct that simplify
the process of splitting a `Rect` into sub-rects according to a given
`Layout`. By putting these methods on the `Rect` struct, we make it a
bit more natural that a layout is applied to the `Rect` itself, rather
than passing a `Rect` to the `Layout` struct to be split.

Adds:
- `Rect::layout` and `Rect::try_layout` methods that allow splitting a
  `Rect` into an array of sub-rects according to a given `Layout`.
- `Rect::layout_vec` method that returns a `Vec` of sub-rects.
- `Layout::try_areas` method that returns an array of sub-rects, with
  compile-time checks for the number of constraints. This is added
  mainly for consistency with the new `Rect` methods.

```rust
use ratatui_core::layout::{Layout, Constraint, Rect};
let area = Rect::new(0, 0, 10, 10);
let layout = Layout::vertical([Constraint::Fill(1); 2]);

// Rect::layout() infers the number of constraints at compile time:
let [top, main] = area.layout(&layout);

// Rect::try_layout() and Layout::try_areas() do the same, but return a
// Result:
let [top, main] = area.try_layout(&layout)?;
let [top, main] = layout.try_areas(area)?;

// Rect::layout_vec() returns a Vec of sub-rects:
let areas_vec = area.layout_vec(&layout);

// you can also explicitly specify the number of constraints:
let areas = area.layout::<2>(&layout);
let areas = area.try_layout::<2>(&layout)?;
let areas = layout.try_areas::<2>(area)?;
```

### DIFF
--- a/ratatui-core/src/layout/rect.rs
+++ b/ratatui-core/src/layout/rect.rs
@@ -585,7 +585,6 @@ impl Rect {
     ///
     /// [`Vec`]: alloc::vec::Vec
     #[must_use]
-    #[cfg(feature = "std")]
     pub fn layout_vec(self, layout: &Layout) -> alloc::vec::Vec<Self> {
         layout.split(self).as_ref().to_vec()
     }


joshka:
Draft because I want to modify the examples to use this too

joshka:
Something to think about when reviewing this is whether these methods should perhaps return an Iterator over the cells instead of an array / vec. This would make them inherently able to be iterated over when building UIs, which would simplify things like https://jm-grid-layout-recipe.ratatui.pages.dev/recipes/layout/grid/

```rust
let cells = area.layout_vec(&vertical).iter().flat_map(|row| row.layout_vec(&horizontal));
```
Might become instead
```rust
let cells = area.layout_iter(&vertical).flat_map(|row| row.layout_iter(&horizontal));
```
Naming of the method might be better without the _iter suffix there - I'm not 100% sure.

j-g00da:
> Something to think about when reviewing this is whether these methods should perhaps return an Iterator over the cells instead of an array / vec.

I like this a lot

ratatui-core/src/layout/rect.rs
```diff
@@ -417,6 +414,100 @@ impl Rect {
...
+    /// let areas = area.layout_vec(&layout);
+    /// assert_eq!(areas, vec![Rect::new(0, 0, 10, 1), Rect::new(0, 1, 10, 9),]);
+    /// ```
+    ///
+    /// [`Vec`]: alloc::vec::Vec
+    #[must_use]
+    #[cfg(feature = "std")]
+    pub fn layout_vec(self, layout: &Layout) -> alloc::vec::Vec<Self> {
+        layout.split(self).as_ref().to_vec()
+    }
```
j-g00da: Why does this require `std`?

joshka: I suspect it was probably due to Vec and then I changed it to use alloc

joshka:
I think I'll add the iter part before merging this.

orhun (approved):
lgtm

ratatui-core/src/layout/rect.rs
joshka: (fixed)

joshka:
Actually, the way the iter has to be implemented is to convert the Rc<[Rect]> into vec and then call .iter() on it. That makes having the function not worth the effort I think.

PR #1919, opened 2025-06-23
feat(no_std): make palette and serde features depends on std

j-g00da:
This is a nitpick, main thing it does is making it easier to see which features require std.


joshka:
Thanks!

PR #1926, opened 2025-06-24
chore: remove block::Title

joshka:
The title alignment is better expressed in the `Line` as this fits more
coherently with the rest of the library.

BREAKING CHANGES:
- `widgets::block` is no longer exported
- `widgets::block::Title` no longer exists
- `widgets::block::Position` is now `widgets::TitlePosition`
- `Block::title()` now accepts `Into::<Line>` instead of `Into<Title>`
- `BlockExt` is now exported at widgets::`BlockExt`

Closes: https://github.com/ratatui/ratatui/issues/738


BREAKING-CHANGES.md
```diff
@@ -85,6 +85,47 @@ This is a quick summary of the sections below:
 
 ## v0.30.0 Unreleased
 
+### `block::Title` no longer exists ([#1926])
```
j-g00da, line 92: This should be also in the summary.

joshka: Good catch. Will fix

j-g00da (approved):
Needs formatting, but other than that LGTM

PR #1929, opened 2025-06-24
feat: add Rect::outer()

joshka:
Fixes: https://github.com/ratatui/ratatui/issues/211

### DIFF
--- a/ratatui-core/src/layout/rect.rs
+++ b/ratatui-core/src/layout/rect.rs
@@ -157,22 +157,30 @@ impl Rect {
 
     /// Returns a new `Rect` outside the current one, with the given margin applied on each side.
     ///
-    /// If the margin causes the `Rect`'s bounds to be negative or larger than the maximum value of
-    /// `u16`, the `Rect` will be clamped to keep the bounds within `u16`. This will modify the size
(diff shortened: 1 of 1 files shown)


ratatui-core/src/layout/rect.rs
```diff
@@ -155,6 +155,26 @@ impl Rect {
...
+    #[must_use = "method returns the modified value"]
+    pub const fn outer(self, margin: Margin) -> Self {
+        Self::new(
+            self.x.saturating_sub(margin.horizontal),
+            self.y.saturating_sub(margin.vertical),
+            self.width
+                .saturating_add(margin.horizontal.saturating_mul(2)),
+            self.height
+                .saturating_add(margin.vertical.saturating_mul(2)),
+        )
```
j-g00da: I think if `x` or `y` saturates, then `width`/`height` can be off (still adding `2*margin`), which I think would be a bug.
I would add a test with the initial `Rect` starting at `(0,0)`.

joshka: You're right, currently this clamps bounds at zero and truncates size at u16::max (due to Rect::new()). It should probably truncate at both ends.

PR #1931, opened 2025-06-24
fix: fix truncation of left aligned block titles

joshka:
truncate the right side of left aligned titles rather than the left side
of right aligned titles. This is more obvious as the left side of text
often contains more important information.

Fixes: https://github.com/ratatui/ratatui/issues/358

### DIFF
--- a/ratatui-widgets/src/block.rs
+++ b/ratatui-widgets/src/block.rs
@@ -771,9 +771,9 @@ impl Block<'_> {
 
     fn render_title_position(&self, position: Position, area: Rect, buf: &mut Buffer) {
         // NOTE: the order in which these functions are called defines the overlapping behavior
-        self.render_right_titles(position, area, buf);
-        self.render_center_titles(position, area, buf);
         self.render_left_titles(position, area, buf);
+        self.render_center_titles(position, area, buf);
+        self.render_right_titles(position, area, buf);
     }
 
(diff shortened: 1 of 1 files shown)


ratatui-widgets/src/block.rs
```diff
@@ -771,9 +771,9 @@ impl Block<'_> {
 
     fn render_title_position(&self, position: Position, area: Rect, buf: &mut Buffer) {
         // NOTE: the order in which these functions are called defines the overlapping behavior
-        self.render_right_titles(position, area, buf);
-        self.render_center_titles(position, area, buf);
         self.render_left_titles(position, area, buf);
+        self.render_center_titles(position, area, buf);
+        self.render_right_titles(position, area, buf);
```
j-g00da, line 776: I'm not sure if changing the rendering order should be treated as just a fix.
Since there's no obvious "correct" behavior, IMO this should be considered a breaking change.

joshka: Every big fix is breaking for some workflow... https://xkcd.com/1172/

For things that only affect edge cases on how something displays, our approach has been to not document them as breaking, otherwise every change to rendering ends up being a breaking change. Instead we draw a fuzzy subjective line at changes that make it difficult to compile existing software.

j-g00da (approved):

PR #1932, opened 2025-06-25
docs: document widgets module

joshka:
Adds a good overview of the use and implementation of widget traits.

Goals with the doc rewrite:
- document the rationale for the ratatui-widgets crate with info for app builders and widget makers.
- Show how to use the widgets for rendering as well as implement the traits- document the differences and reasons for each trait
- document the historical perspective (to make it easy to understand older Ratatui apps as well as migrate to newer approaches
- give recommended approaches to implementing traits
- explain the differences between Consuming and Shared / Mutable Reference implementations of Widget
- explain the differences between using StatefulWidget and Mutable References
- Explain the use case for WidgetRef and StatefulWidgetRef
- Link out to third part widget lists

Fixes: https://github.com/ratatui/ratatui/issues/366

### DIFF
--- a/ratatui/src/widgets.rs
+++ b/ratatui/src/widgets.rs
@@ -1,16 +1,513 @@
 #![warn(missing_docs)]
-//! `widgets` is a collection of types that implement [`Widget`] or [`StatefulWidget`] or both.
+//! Widgets are the building blocks of user interfaces in Ratatui.
 //!
-//! The widgets provided with Ratatui are implemented in the [`ratatui_widgets`] crate, and are
-//! re-exported here. The [`Widget`] and [`StatefulWidget`] traits are implemented in the
-//! [`ratatui_core`] crate and are also re-exported in this module. This means that you can use
-//! these types directly from the `ratatui` crate without having to import the `ratatui_widgets`
-//! crate.
+//! They are used to create and manage the layout and style of the terminal interface. Widgets can
+//! be combined and nested to create complex UIs, and can be easily customized to suit the needs of
+//! your application.
 //!
-//! Widgets are created for each frame as they are consumed after rendered. They are not meant to be
-//! stored but used as *commands* to draw common figures in the UI.
+//! Ratatui provides a wide variety of built-in widgets that can be used to quickly create UIs.
+//! Additionally, [`String`], [`&str`], [`Span`], [`Line`], and [`Text`] can be used as widgets
+//! (though often [`Paragraph`] is used instead of these directly as it allows wrapping and
+//! surrounding the text with a block).
 //!
-//! The available widgets are:
+//! # Crate Organization
+//!
+//! Starting with Ratatui 0.30.0, the project was split into multiple crates for better modularity:
+//!
+//! - **[`ratatui-core`]**: Contains the core widget traits ([`Widget`], [`StatefulWidget`]) and
+//!   text-related types ([`String`], [`&str`], [`Span`], [`Line`], [`Text`])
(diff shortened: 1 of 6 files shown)


joshka:
ping @hasezoey for a review

joshka:
Screenshot to save having to render this:

Screen Shot 2025-06-25 at 02 08 50

ratatui/src/widgets.rs
```diff
@@ -1,16 +1,513 @@
...
+//!
+//! In Ratatui, widgets are implemented as Rust traits, which allow for easy implementation and
+//! extension. The main traits for widgets are:
+//!
+//! - [`Widget`]: Basic trait for stateless widgets that are consumed when rendered
+//! - [`StatefulWidget`]: Trait for widgets that maintain state between renders
+//! - [`WidgetRef`]: Trait for rendering widgets by reference (unstable)
+//! - [`StatefulWidgetRef`]: Trait for rendering stateful widgets by reference (unstable)
+//!
+//! ## Widget
```
hasezoey: Is this also meant to be wrapped in inline code blocks like the other headers for `*Widget*`?

```suggestion
//! ## `Widget`
```

ratatui/src/widgets.rs
```diff
@@ -1,16 +1,513 @@
...
+//! Widgets that need to maintain state between frames use `Frame::render_stateful_widget()`. This
+//! method takes both the widget and a mutable reference to its state, allowing the widget to read
+//! and modify state during rendering (such as updating scroll positions or handling selections).
+//!
+//! ```rust
+//! # use ratatui::{backend::TestBackend, Terminal};
+//! # use ratatui::widgets::{List, ListItem, ListState};
+//! # let backend = TestBackend::new(10, 3);
+//! # let mut terminal = Terminal::new(backend).unwrap();
+//! # let mut list_state = ListState::default();
```
hasezoey:
```suggestion
//! let mut list_state = ListState::default();
```

This is relevant to the types used, otherwise it could be confusing where or what `list_state` is.

hasezoey (approved):
Looks good to me and answers basically all questions, though reading the recommendations, i have the following understanding:

- `WidgetRef` and `StatefulWidgetRef` are basically just helper traits to allow dyn compatability for `&Widget` and `&StatefulWidget`
- it is preferred to use `impl Widget for &SomeWidget` for reference taking / reusable widgets (over implementing `WidgetRef`)
- `WidgetRef` and `StatefulWidgetRef` are not meant to replace parts of the "old" widget traits
- `impl Widget for &mut SomeWidget` vs `impl StatefulWidget for SomeWidget` is basically just choice of the maker of `SomeWidget`

joshka:
Yep to all the above
WidgetRef in 0.29 was intended to be the replacement, but we worked out it was easy to flip the blanket implementation around and it leads to better code generally.  (one of the reasons that it's still unstable was that we weren't 100% sure the direction was correct.

j-g00da:
I know there will be a more detailed section about no-std in the version highlights (and would be also good to dedicate a separate page for that), but I still would like to see here a short section about recommendations for widget libraries that expands on the note in "Crate organization" section (with a link to a page with details).

WDYT? I would add this in a separate PR.

joshka:
Definitely agree. This comment is mainly about clarifying that there could be some confusion about which things exist in ratatui:: widgets and which are in ratatui-widgets. There needs to be a good summary of this on the main level too. Repeating this info is fine as most people will likely read only a small portion of any doc they're reading.

PR #1934, opened 2025-06-25
fix: fix handling of multi-byte chars in bar chart

joshka:
The split_at method requires that the split point is at a valid utf8
character boundary.

Fixes: https://github.com/ratatui/ratatui/issues/1928


hodgesds:
Thanks for fixing this so fast!

joshka:
We're ramping up for a large 0.30 release in a couple of weeks time, so it might not be available until then. We should have a beta release out before then though. Keep an eye out.

PR #1935, opened 2025-06-25
feat: add lifetime to symbol sets

joshka:
This makes it possible to create symbol sets at runtime with non-static
lifetimes.

Fixes: https://github.com/ratatui/ratatui/issues/1722


j-g00da:
Just for the sake of curiosity:
Is there any particular reason we are not using `char` type for symbols other than the fact that migrating `&str` to `char` would break things?

joshka:
> Just for the sake of curiosity:
> Is there any particular reason we are not using `char` type for symbols other than the fact that migrating `&str` to `char` would break things?

Unicode

https://doc.rust-lang.org/std/primitive.char.html#representation has a good description.

j-g00da:
Ohh thanks!

orhun (approved):

PR #1942, opened 2025-06-26
chore: remove OpenSSL license

joshka:
Only keep licenses in the allow list that are actually used


joshka:
```
warning[license-not-encountered]: license was not encountered
   ┌─ /home/runner/work/ratatui/ratatui/deny.toml:12:4
   │
12 │   "OpenSSL",
   │    ━━━━━━━ unmatched license allowance
```

PR #1948, opened 2025-06-27
docs: improve layout related docs

### LINKED ISSUE #1937: Improve layout docs
## Problem

The api docs for the Layout struct are mostly about the implementation rather than the use of layouts. There's some docs on the website that go into more detail, but the api docs should cover a few more bits about common use cases.

## Solution
Write docs that cover:
- nested layouts
- how the cache works
- split() vs areas()
- flex
- Spacing and spacers

Add layout module level docs that talk about what each struct is used for and provides a good overview of the layout system. Consider how the std lib splits module level and type level docs and arrange the topics as appropriate.

## Alternatives
We could treat the website as the primary documentation source, but in doing so, we would have to judiciously link to things. That's difficult to do well.

## Additional context

https://github.com/ratatui/ratatui/issues/1906

https://discord.com/channels/1070692720437383208/1072907135664529508/1386410418855084032

joshka:
Adds module level docs and more comprehensive docs on all the types in the layout module

Fixes #1937

### DIFF
--- a/ratatui-core/src/layout.rs
+++ b/ratatui-core/src/layout.rs
@@ -89,10 +89,27 @@
 //!
 //! // Divide it vertically into two equal parts using Layout
 //! let layout = Layout::vertical([Constraint::Percentage(50), Constraint::Percentage(50)]);
-//! let areas: [Rect; 2] = layout.areas(area);
+//! let [top, bottom] = layout.areas(area);
 //!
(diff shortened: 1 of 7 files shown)


orhun:
I think this is linked to #1937

j-g00da (approved):

PR #1949, opened 2025-06-27
chore(gitignore): add `.env` to `.gitignore`

j-g00da:


orhun:
Do we even use this file? Not sure if updating the `.gitignore` is necessary

joshka:
It's worth it to avoid committing secrets generally

PR #1953, opened 2025-06-28
docs: improve Block docs

joshka:


joshka:
Rendered
Screen Shot 2025-06-28 at 03 20 50

kdheepak (approved):

ratatui-widgets/src/block.rs
```diff
@@ -360,16 +471,18 @@ impl<'a> Block<'a> {
...
     ///
-    /// Block::new()
+    /// Block::bordered()
     ///     .title_position(TitlePosition::Bottom)
-    ///     .title_top("top")
-    ///     .title("foo")
-    ///     .title("bar");
+    ///     .title("foo") // will be at the bottom
+    ///     .title("bar") // will be at the bottom
+    ///     .title_top("top"); // will be at the top
```
j-g00da: nit: changing order may demonstrate it better than calling `.title(...)` two times in a row
```suggestion
    ///     .title("foo") // will be at the bottom
    ///     .title_top("top"); // will be at the top
    ///     .title("bar") // will be at the bottom
```

ratatui-widgets/src/block.rs
```diff
@@ -22,87 +22,194 @@ use crate::borders::{BorderType, Borders};
 
 mod padding;
 
-/// Base widget to be used to display a box border around all other built-in widgets.
+/// A widget that displays borders, titles, and padding around other widgets.
```
j-g00da: nit: `displays` sounds off to me in context of padding (but it can be just me)
```suggestion
/// A widget that renders borders, titles, and padding around other widgets.
```

j-g00da (approved):
LGTM, added some nitpick comments, but I'm fine with merging it as-is too

ratatui-widgets/src/block.rs
joshka: Changed it to renders instead of adds.

PR #1955, opened 2025-06-28
docs: add docs for authoring widget crates

### LINKED ISSUE #1941: Add no-std docs and recommendations
## Problem
We want future third-party widget libraries to maintain no-std compatibility. Currently there are no recommendations for widget crates other than depending on `ratatui-core`.

## Solution
Add recommendations for creating no-std compatible widget crates.

## Additional context

#1932

j-g00da:
Resolves #1941

### DIFF
--- a/ratatui/src/widgets.rs
+++ b/ratatui/src/widgets.rs
@@ -39,6 +39,62 @@
 //! [`ratatui-core`]: https://crates.io/crates/ratatui-core
 //! [`ratatui-widgets`]: https://crates.io/crates/ratatui-widgets
 //!
+//! # Built-in Widgets
+//!
+//! Ratatui provides a comprehensive set of built-in widgets:
+//!
+//! - [`Block`]: a basic widget that draws a block with optional borders, titles and styles.
+//! - [`BarChart`]: displays multiple datasets as bars with optional grouping.
+//! - [`calendar::Monthly`]: displays a single month.
+//! - [`Canvas`]: draws arbitrary shapes using drawing characters.
+//! - [`Chart`]: displays multiple datasets as a lines or scatter graph.
+//! - [`Clear`]: clears the area it occupies. Useful to render over previously drawn widgets.
+//! - [`Gauge`]: displays progress percentage using block characters.
+//! - [`LineGauge`]: display progress as a line.
+//! - [`List`]: displays a list of items and allows selection.
+//! - [`Paragraph`]: displays a paragraph of optionally styled and wrapped text.
+//! - [`Scrollbar`]: displays a scrollbar.
+//! - [`Sparkline`]: display a single data set as a sparkline.
+//! - [`Table`]: displays multiple rows and columns in a grid and allows selection.
+//! - [`Tabs`]: displays a tab bar and allows selection.
+//! - [`RatatuiLogo`]: displays the Ratatui logo.
+//! - [`RatatuiMascot`]: displays the Ratatui mascot.
+//!
+//! Additionally, primitive text types implement [`Widget`]:
+//! - [`String`]: renders the owned string content
+//! - [`&str`]: renders the string slice content
+//! - [`Line`]: renders a single line of styled text spans
+//! - [`Span`]: renders a styled text segment
+//! - [`Text`]: renders multiple lines of styled text
+//!
+//! For more information on these widgets, you can view the widget showcase and examples.
+//!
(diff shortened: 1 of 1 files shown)


j-g00da:
Some details:
1. The code snippet is with warning as I had to ignore it. I'm not sure how to build it in docs without causing linker errors.
2. I can expand it a bit with some other details, but I think `ratatui-widgets` codebase would be more helpful.
3. Rendered docs:
image

ratatui/src/widgets.rs
```diff
@@ -268,6 +268,74 @@
...
+//!
+//! To do this, first add the `![no_std]` attribute at the top of your `lib.rs`.
+//!
+//! When working in a `no_std` environment:
+//!
+//! - Use `core::` instead of `std::` for basic functionality
+//! - Use `alloc::` for heap-allocated types like `String`, `Vec`, and `Box`
+//! - Add `extern crate alloc;` to access allocation types
+//!
+//! Here's a complete example of a `no_std` compatible widget:
```
joshka: I think this deserves an overview here like you've put, but a deeper doc that's linked to (probably a /concepts/some-category/no-std on the website).

ratatui/src/widgets.rs
```diff
@@ -268,6 +268,74 @@
...
+//!
+//! - **Broader compatibility**: Your widgets work in embedded environments and other `no_std`
+//!   contexts
+//! - **Lighter dependencies**: Users don't pull in backend code they don't need
+//! - **Better compile times**: Fewer dependencies mean faster builds
+//! - **Future-proofing**: Your library remains compatible as Ratatui evolves its architecture
+//!
+//! Even if you haven't worked with `no_std` development before, the changes are typically minimal
+//! for widget libraries since most widget logic involves basic data manipulation which work well
+//! within `no_std` constraints.
```
joshka: I'd push these benefits up into the above two sections (on depending on core and making things no-std). WDYT?

One not obvious thing I've been trying to do with the docs rewrite stuff is avoid small one paragraph sections and generally have a few paragraphs in each section. This makes it easier to not be assaulted by section headings every few lines and makes the documentation a bit more easier.

ratatui/src/widgets.rs
```diff
@@ -268,6 +268,74 @@
...
+//! use ratatui_core::buffer::Buffer;
+//! use ratatui_core::layout::Rect;
+//! use ratatui_core::text::Line;
+//! use ratatui_core::widgets::Widget;
+//!
+//! struct MyWidget {
+//!     content: String,
+//! }
+//!
+//! impl Widget for &MyWidget {
```
joshka: The problem with putting an example here of the shared reference is that this section comes before the section about choosing state strategies. So it seems a bit out of place to be presenting a general example at this point in the docs flow.

(Also, it reparents the below H2 sections under this H1)

What about making this into a "Releasing a widget library" section (h2) as the last section in the custom widgets section?

(side note: re-thinking about this a little, the built-in and third party widget sections probably deserve to be either first or second to help users find widgets in the library). WDYT?

j-g00da: Oh I just realized I've put it in wrong place, yes the headings are totally wrong now.

joshka: (Can you move the built-in and third party sections up in this PR as well - doesn't need a new PR for it)

ratatui/src/widgets.rs
```diff
@@ -268,6 +268,74 @@
...
+//!     fn render(self, area: Rect, buf: &mut Buffer) {
+//!         Line::raw(&self.content).render(area, buf);
+//!     }
+//! }
+//! ```
+//!
+//! ## Benefits of This Approach
```
orhun: I think you are referring to `no_std` here so this should be a sub-section.

```suggestion
//! ### Benefits of This Approach
```

j-g00da: I actually refer to both, but I will merge content of this into the sub-sections above anyway. 
(context: https://github.com/ratatui/ratatui/pull/1955#discussion_r2173538364)

ratatui/src/widgets.rs
j-g00da: I'll add separate issue for this on website repo, we can link it here when it's ready.

j-g00da:
Screen Shot 2025-06-30 at 00 12 12

joshka (approved):

PR #1956, opened 2025-06-28
feat(text): implement `AddAssign` for `Text`

acuteenvy:
It's possible to add two instances of `Text` together via the [`Add` trait](https://docs.rs/ratatui/0.29.0/ratatui/text/struct.Text.html#impl-Add-for-Text%3C'a%3E), but the `AddAssign` 
impl doesn't exist.

I added an `AddAssign` implmentation that does the same thing as `Add`.


joshka:
Thanks for the PR :)

PR #1961, opened 2025-06-30
ci: audit github workflows with zizmor

joshka:
Fixes https://github.com/ratatui/ratatui/issues/1950

### DIFF
--- a/.github/workflows/check-pr.yml
+++ b/.github/workflows/check-pr.yml
@@ -1,6 +1,13 @@
 name: Check Pull Requests
 
+# Set the permissions of the github token to the minimum and only enable what is needed
+# See https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions
+permissions: {}
+
 on:
+  # this workflow is required to be run on pull_request_target as it modifies the PR comments
+  # care should be taken that the jobs do not run any untrusted input
+  # zizmor: ignore[dangerous-triggers]
   pull_request_target:
     types:
       - opened
@@ -8,23 +15,21 @@ on:
       - synchronize
       - labeled
       - unlabeled
-  merge_group:
-
-permissions:
-  pull-requests: write
 
 jobs:
   check-title:
+    permissions:
+      pull-requests: write
     runs-on: ubuntu-latest
     steps:
       - name: Check PR title
         if: github.event_name == 'pull_request_target'
-        uses: amannn/action-semantic-pull-request@v5
+        uses: amannn/action-semantic-pull-request@0723387faaf9b38adef4775cd42cfd5155ed6017 # v5
         id: check_pr_title
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       # Add comment indicating we require pull request titles to follow conventional commits specification
-      - uses: marocchino/sticky-pull-request-comment@v2
+      - uses: marocchino/sticky-pull-request-comment@52423e01640425a022ef5fd42c6fb5f633a02728 # v2
         if: always() && (steps.check_pr_title.outputs.error_message != null)
         with:
           header: pr-title-lint-error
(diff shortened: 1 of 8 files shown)


.github/workflows/check-pr.yml
```diff
@@ -1,30 +1,35 @@
 name: Check Pull Requests
 
+# Set the permissions of the github token to the minimum and only enable what is needed
+# See https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions
+permissions: {}
+
 on:
+  # this workflow is required to be run on pull_request_target as it modifies the PR comments
+  # care should be taken that the jobs do not run any untrusted input
+  # zizmor: ignore[dangerous-triggers]
```
joshka, line 10: This is the only ignored rule - the threat is that running any external code in the context of the repo could make it possible for a PR to do malicious things. The jobs in this workflow avoid running any untrusted input.

.github/workflows/check-pr.yml
```diff
@@ -1,30 +1,35 @@
...
   pull_request_target:
     types:
       - opened
       - edited
       - synchronize
       - labeled
       - unlabeled
-  merge_group:
```
joshka, line 11: We no longer use merge groups

.github/workflows/release-alpha.yml
```diff
@@ -20,29 +24,31 @@ jobs:
       contents: write
     steps:
       - name: Checkout the repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
         with:
           fetch-depth: 0
+          # probably needs to be set to true for git-cliff to work - check build
```
joshka, line 30: We will have to check this for failure when using this workflow next

.github/workflows/release-stable.yml
```diff
@@ -13,20 +17,22 @@ jobs:
       contents: write
     steps:
       - name: Checkout the repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
         with:
           fetch-depth: 0
+          # this possibly needs to be set to true for git-cliff to work - check build results
```
joshka, line 23: We likely need to check this also.

joshka:
(added the zizmor workflow)

.github/workflows/ci.yml
```diff
@@ -40,18 +46,22 @@
...
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
-      - uses: dtolnay/rust-toolchain@stable
-      - uses: taiki-e/install-action@cargo-deny
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
+        with:
+          persist-credentials: false
+      - uses: dtolnay/rust-toolchain@4305c38b25d97ef35a8ad1f985ccf2d2242004f2 # stable
+      - uses: taiki-e/install-action@6bacb65d2d3b6af9f35e32e03022b97653a66b84 # cargo-deny
```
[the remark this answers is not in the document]
joshka: This is a bit odd - I definitely copied the hashes from the install-action repo, unsure why they're no longer part of the repo. Perhaps it they force push to update the branches?

https://github.com/taiki-e/install-action/commit/6bacb65d2d3b6af9f35e32e03022b97653a66b84
current latest in that branch: https://github.com/taiki-e/install-action/commit/4294a5359c6effcec012fc3b158b96042ad0a709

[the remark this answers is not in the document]
joshka: raised in https://github.com/taiki-e/install-action/issues/1023

.github/workflows/release-alpha.yml
orhun: Hmm, how is this related to git-cliff I wonder

joshka: The question is whether git-cliff needs the git creds that are saved by the actions/checkout call, in which case it will fail. We'll see that next time we run this workflow.

bacon.toml
```diff
@@ -60,6 +60,22 @@ command = ["cargo", "xtask", "hack"]
 [jobs.format]
 command = ["cargo", "xtask", "format"]
 
+[jobs.zizmor-offline]
+# zizmor checks the workflow files for security issues. The fffline version is generally faster, but
```
j-g00da:
```suggestion
# zizmor checks the workflow files for security issues. The offline version is generally faster, but
```

orhun (approved):

.github/workflows/ci.yml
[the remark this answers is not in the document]
joshka: The answer to this is https://github.com/ratatui/ratatui/pull/1961/commits/7652dcfeac4eedfeb6cef442604f8e408d64ac71

These actions (dtolnay/rust-toolchain, dtolnay/install, taiki-e/install-action) use an approach where they update the branch by rewriting it from the main branch and then add a single commit on top that handles providing the tool name / toolchain name. This is a reasonable approach, but it makes the previous hashes not part of the commit tree and triggers zizmor to fail as it detects the changes as malicious absent other information.

To address this I switched these to point at hashes on the master / version tagged branches which follow a more standard linear history, and instead of using the shortcut branch names we pass the toolname / toolchain name into the action.

PR #1963, opened 2025-07-01
Fix centered block title truncation

### LINKED ISSUE #1936: Centered block titles truncate oddly
## Description
Truncation of Centered titles truncate the titles one by one rather than the whole list of titles.

## To Reproduce

```rust
    #[test]
    fn center_titles_truncated() {
        let mut buffer = Buffer::empty(Rect::new(0, 0, 10, 1));
        Block::new()
            .title(Line::from("C12345").centered())
            .title(Line::from("C67890").centered())
        assert_eq!(buffer, Buffer::with_lines(["C12345 678"])); // incorrect current behavior
    }
```

## Expected behavior

The expected result should be "12345 C678" or "2345 C6789".
I forget how we choose between centering odd character counts in even spaces, but this should be consistent with all other usages.

ognis1205:
## Description

WIP

## Related Issue

Closes #1936

## Checklist

- [ ] All unit tests pass
- [ ] Added regression test for multi-title center truncation
- [ ] Documentation updated if necessary

## Notes

WIP


joshka:
I ❤️ Unit tests before the fix to show that the fix actually fixes things

ognis1205:
> I ❤️ Unit tests before the fix to show that the fix actually fixes things

Thanks, @joshka !
The fix area and the general approach are decided, but I’d appreciate your help when I get stuck or have questions.

ognis1205:
This should now be ready for review. cc: @joshka  — let me know if anything needs clarification.

joshka:
Wow, this turned into something much larger than expected. Thanks for going to the effort. It might take a bit more time to get a good review of things here, so I'll apologize in advance for any delay.

ognis1205:
Hi @joshka — thanks again for the feedback!

Just to add a bit of context:

 - I’m honestly not sure if this was the best possible strategy, so if there’s a simpler or more idiomatic approach, please feel free to suggest — I’m very open to revising things.
 - I did consider adding something like `Direction::{Horizontal, Vertical}` to `Text` to handle this more generally, but I hesitated since that would be more of a new feature rather than a bug fix.
 - Also, I’m not 100% confident that the name `Inline` is the most appropriate — it might sound more like a layout component than a text structure.

Happy to adjust direction if you think there’s a better path. Thanks as always for your time!

joshka:
The approach that I expected to see here was something like the following level of complexity:

```rust
let centered_titles = titles.filter(|title| title.alignment == Alignment::Center);
let total_width = centered_titles.map(|title| title.width() + 1) - 1;
let excess = total_width.saturating_sub(area.width);
let mut left_trim = excess / 2; // round down
let mut right_trim = (excess + 1) / 2; // round up
for each title in centered_titles {
    // draw each title, taking into account left and right trim to ignore titles
    // that are not visible and partially draw others
    // this is the most complex part of the change, but it's still reasonably simple - like 5-10 LoC
}
```

Don't fear though, I'm fine with this blowing out significantly from that. I think you may have solved a much more difficult problem, and that's one that a few people have asked about (e.g. [earlier today on discord](https://discord.com/channels/1070692720437383208/1072907135664529508/1391877980879458405)). That question is effectively "how do I put multiple line parts on the same line aligned differently". Have I got that right?

ognis1205:
Thanks @joshka — your proposed approach is definitely simpler and more elegant! I actually wanted to implement something along those lines as well, though I wasn’t able to reduce it to something quite as simple and elegant as what you’ve suggested.

At the time, I considered using [Line::spans_after_width](https://github.com/ratatui/ratatui/blob/32ccf38606a992aebf1155ad79ea9d2a9c3a34c2/ratatui-core/src/text/line.rs#L748), but since it’s a private method — and this PR wasn’t meant to introduce a new feature — I ended up reimplementing the logic as a new, undocumented API instead.

Your approach might actually have been a better match for the issue’s scope than what I ended up with. Thanks for sharing the idea — it’s very helpful!

I also saw the Discord thread. While the current implementation doesn’t directly support that use case, I believe it could be extended with something like `Inline::justify(Space::Between)`. That would allow for automatic spacing based on the rendering width — essentially replicating `justify-content: space-between` from CSS:

https://www.w3schools.com/css/tryit.asp?filename=trycss3_flexbox_justify-content_space-between

I haven’t looked deeply into the `layout` module yet, so whether this belongs in `layout` or `text` is probably worth discussing.

joshka:
I wonder if you'd be keen to solve only the narrower title problem with the simple approach I mentioned, while keeping this PR open for a longer discussion about the ideas in it? I haven't had a chance to go through this thoroughly yet, but I definitely want to if only to understand where it fits. I know it's going to require a bit of deep thinking, so I need to put it off a little until I have the time and energy to put into it.

ognis1205:
That sounds totally reasonable to me — I’ll try to extract the narrower title fix into a separate PR, and we can continue exploring the broader ideas here over time. Thanks for taking the time to consider it so thoughtfully!

ognis1205:
Ah, I just realized there's already a `Flex` implementation in the `ratatui_core::layout` module — the context from the Discord thread makes more sense now.

While `Inline` wasn’t originally designed for the use case discussed in that thread, I think it could be adapted to support similar behavior — rendering multiple `Line`s, each with its own alignment, on the same row. That would make it somewhat of a true horizontal counterpart to `ratatui_core::text::Text`.

Of course, the exact behavior and API design would still need discussion, and we should probably be cautious about introducing layout-like functionality into the `text` module — especially if it involves more complex layout logic. That said, I think this direction is worth exploring. I’m still getting up to speed with `ratatui`, but the Discord thread made me wonder whether this use case could be handled within `layout::Flex`, or whether a specialized solution within `text` might actually make sense.

ognis1205:
^ updated.

joshka:
I took a quick skim. I think I misinterpreted what you wrote a bit originally.
My understanding is that this implementation is a bit narrower than I assumed.

Inline is not of a:
- replacement for a horizontal group of Lines each with their own alignment

it's actually a:
- replacement for horizontal groups of lines all sharing the same alignment

So, this is definitely a bit overkill for solving the direct problem.

I'd say though that the first problem is probably worth solving at some point. It makes a bunch of things a bit easier (e.g. title bars, menus, status bars, table of contents, ...). So this isn't wasted effort.

And to be clear, if you do tackle the simpler approach, please feel free to open another PR for it rather than modifying this one. Grab the unit tests from this though if they apply.

ognis1205:
# Inline Layout Behavior (RFC)

## Premise

- `W`: Total rendering width.
- `L`, `C`, `R`: The display width (in characters or cells) of each respective text.

## Rendering Range of Each Text

```
left_start   = 0
left_end     = L               if L > 0 else 0

center_start = (W - C) / 2     if C > 0 else ∞
center_end   = center_start + C

right_start  = W - R           if R > 0 else ∞
right_end    = W
```

## Overlap Detection (automatically skipped if either side is empty)

```
overlap_lc = max(0, min(left_end, center_end) - max(left_start, center_start)) if L > 0 and C > 0 else 0
overlap_cr = max(0, min(center_end, right_end) - max(center_start, right_start)) if C > 0 and R > 0 else 0
overlap_lr = max(0, min(left_end, right_end) - max(left_start, right_start)) if L > 0 and R > 0 else 0
```

## Shrinking Logic (distribute overlapping width evenly)

```
if overlap_lc > 0:
    L -= overlap_lc / 2
    C -= overlap_lc / 2

if overlap_cr > 0:
    C -= overlap_cr / 2
    R -= overlap_cr / 2

if overlap_lr > 0:
    L -= overlap_lr / 2
    R -= overlap_lr / 2
```

## Examples

```
Area: 50, No Overlap, Spacer 1
|__________________________________________________|
|LLLLL                                             |
|                     CCCCCCC                      |
|                                            RRRRRR|

Result:
|LLLLL                CCCCCCC                RRRRRR|


Area: 50, Overlapped, Spacer 1 (._, represent spacers)
|__________________________________________________|
|LLLLL.LLLLL.LLLLL.LLLLL                           |
|             CCCCCCC_CCCCCCC_CCCCCCC              |
|                              RRRRRR,RRRRRR,RRRRRR|

Result:
|LLLLL.LLLLL.LLLLLCCC_CCCCCCC_CCCCCRR,RRRRRR,RRRRRR|


Area: 50, Overlapped, Spacer 1 (._, represent spacers)
|__________________________________________________|
|LLLLL.LLLLL.LLLLL.LLLLL                           |
|             CCCCCCC_CCCCCCC_CCCCCCC              |
|                                                  |

Result:
|LLLLL.LLLLL.LLLLLCCC_CCCCCCC_CCCCCCC              |


Area: 50, Overlapped, Spacer 1 (._, represent spacers)
|__________________________________________________|
|                                                  |
|             CCCCCCC_CCCCCCC_CCCCCCC              |
|                              RRRRRR,RRRRRR,RRRRRR|

Result:
|             CCCCCCC_CCCCCCC_CCCCCRR,RRRRRR,RRRRRR|


Area: 50, Overlapped, Spacer 1 (._, represent spacers)
|__________________________________________________|
|LLLLL.LLLLL.LLLLL.LLLLL.LLLLL                     |
|                                                  |
|                       RRRRRR,RRRRRR,RRRRRR,RRRRRR|

Result:
|LLLLL.LLLLL.LLLLL.LLLLL.LLRRR,RRRRRR,RRRRRR,RRRRRR|
```

ognis1205:
## Regarding the naming of `Span`, `Line`, and `Text`

As we plan to introduce a new `Inline` (tentative name) widget, I would also like to propose renaming based on its functionality.

### Proposal 1 (breaking change)
 - `Span` → `Span`
 - `Line` → `Text`
 - `Text` → `VTextGroup`
 - `Inline` → `HTextGroup`

### Proposal 2
 - `Span` → `Span`
 - `Line` → `Line`
 - `Text` → `VText`
 - `Inline` → `HText`

### Proposal 3
 - `Span` → `Span`
 - `Line` → `Line`
 - `Text` and `Inline` → unified as `Text` + `Direction` enum

joshka:
Generally speaking, I'd set a fairly high bar on benefits when considering whether a breaking change that would affect almost every application built with Ratatui, so it's important to make sure that this really obviously captures that. It might be helpful to take a small step back from the implementation detail and ground the solution by talking more about the problem being solved here and then link that to the solution with a narrative about how each of the proposed approaches would solve the problem.

TL;DR:
- explain why this change needs to be made
- show the alternate solutions and how they solve the problem
- make sure the pros and cons of the solutions are considered (particularly downstream impact - github searches can be helpful here)
- assess the tradeoffs and make a recommendation

I would strongly anticipate that after doing that, that the likely result is that we don't make a breaking change in this area.

ognis1205:
@joshka

Thank you very much for taking the time to give such detailed and thoughtful feedback.

I realize now that suggesting a breaking change, even with good intentions around API consistency, carries significant weight—especially in a mature and widely used project like Ratatui. I really appreciate you pointing out the importance of considering downstream impact and long-term maintenance.

My proposal regarding the naming was motivated by a desire to introduce the new `Inline` feature in a way that feels coherent with the rest of the API, but I now see that this might be better approached incrementally, without a breaking rename—at least for now.

I’ll go back, do some more research, and see if I can reframe the proposal in a way that clearly identifies the underlying problem, compares possible solutions (including non-breaking ones), and evaluates the tradeoffs more carefully. I’ll also make sure to write this up more cleanly as an RFC.

Thanks again for your patience and guidance!

ognis1205:
Just a quick note to myself (again) — I’ll try to write up a short RFC/design doc for this.

ognis1205:
## RFC

title: Support for Inline Widget in `ratatui_core::text`
author: @ognis1205 
status: Draft
created: 2025-08-07

### Summary

This RFC proposes the addition of an inline widget API in the `ratatui_core::text` module. While the existing `Text` abstraction represents a vertical (line-wise) composition — i.e., a collection of `Line` structs stacked top-to-bottom — this RFC introduces a counterpart API that models horizontal (column-wise) composition of `Line`s. This would allow for richer, inline-style rendering within text regions.

### Motivation

"How do I put multiple line parts on the same line aligned differently?"

The current API makes this difficult to achieve with the same level of simplicity that `Text` offers. Developers must either:

 - Use `Block::inner` and manually render line fragments with different alignments, which is error-prone and risks overwriting content if rendering order is not carefully managed.
 - Use `ratatui::layout::Layout` and `ratatui::layout::Constraint` to split a line into aligned fragments. This requires (pseudo) manual layout computation and can become quite complex when combined with style management.

The proposed Inline API aims to solve this alignment problem with the same level of abstraction and ease of use as the existing `Text` type, reducing boilerplate and increasing composability.

See also:

 - [Text](https://ratatui.rs/recipes/render/display-text/)
 - [Layout](https://ratatui.rs/concepts/layout/)
 - [Constraint](https://docs.rs/ratatui/latest/ratatui/layout/enum.Constraint.html)
 - [Discord Discussion](https://discord.com/channels/1070692720437383208/1072907135664529508/1391877980879458405)
 
### Guide-level Explanation

#### Before (current approach with Layout + manual alignment):

```rust
let block = Block::bordered();
let inner = block.inner(area);
let [left, right] = inner.layout(Layout::horizontal(Constraint::from_fills([score.width(), time.width()]);

let score = format!("Score: {score}").blue();
let time = format!("Time: {time}").yellow().to_line().right_aligned();
score.render(left, buf);
time.render(right, buf);
```

#### After (with proposed Inline API):

```rust
let score = format!("Score: {score}").blue().left_aligned();
let time = format!("Time: {time}").yellow().right_aligned();
let inline = Inline::from(vec![score, time])
    .style(style)
    .spacer(1);
let block = Block::bordered();
let paragraph = Paragraph::new(inline)
    .block(block);
    .render(area, buf);
```

#### Before (current block.rs title rendering):

```rust
fn render_title_position(&self, position: TitlePosition, area: Rect, buf: &mut Buffer) {
    self.render_left_titles(position, area, buf);
    self.render_center_titles(position, area, buf);
    self.render_right_titles(position, area, buf);
}
```

#### After (with Inline API):

```rust
fn render_title_position(&self, position: TitlePosition, area: Rect, buf: &mut Buffer) {
    self.render_titles(position, area, buf);
}
```

#### Benefit:

The Inline API eliminates the need for separate rendering functions and manual ordering for each alignment, significantly improving maintainability.

### Reference-level Explanation

#### Internally, the Inline type could:
 - Contain a collection of Lines tagged with alignment metadata (Left, Center, Right).
 - Implement rendering logic that computes the visual horizontal alignment across a shared y-coordinate (line), handling spacing and overlapping according to configuration.
 
#### Examples:

```
Area: 50, No Overlap, Spacer 1
|__________________________________________________|
|LLLLL                                             |
|                     CCCCCCC                      |
|                                            RRRRRR|

Result:
|LLLLL                CCCCCCC                RRRRRR|


Area: 50, Overlapped, Spacer 1 (._, represent spacers)
|__________________________________________________|
|LLLLL.LLLLL.LLLLL.LLLLL                           |
|             CCCCCCC_CCCCCCC_CCCCCCC              |
|                              RRRRRR,RRRRRR,RRRRRR|

Result:
|LLLLL.LLLLL.LLLLLCCC_CCCCCCC_CCCCCRR,RRRRRR,RRRRRR|


Area: 50, Overlapped, Spacer 1 (._, represent spacers)
|__________________________________________________|
|LLLLL.LLLLL.LLLLL.LLLLL                           |
|             CCCCCCC_CCCCCCC_CCCCCCC              |
|                                                  |

Result:
|LLLLL.LLLLL.LLLLLCCC_CCCCCCC_CCCCCCC              |


Area: 50, Overlapped, Spacer 1 (._, represent spacers)
|__________________________________________________|
|                                                  |
|             CCCCCCC_CCCCCCC_CCCCCCC              |
|                              RRRRRR,RRRRRR,RRRRRR|

Result:
|             CCCCCCC_CCCCCCC_CCCCCRR,RRRRRR,RRRRRR|


Area: 50, Overlapped, Spacer 1 (._, represent spacers)
|__________________________________________________|
|LLLLL.LLLLL.LLLLL.LLLLL.LLLLL                     |
|                                                  |
|                       RRRRRR,RRRRRR,RRRRRR,RRRRRR|

Result:
|LLLLL.LLLLL.LLLLL.LLLLL.LLRRR,RRRRRR,RRRRRR,RRRRRR|
```

### Drawbacks

 - Requires a new layout algorithm specific to inline rendering
 - Could increase confusion between Text and Inline unless clearly documented

### Alternatives

 - Extend Text to support alignment metadata per Line (might complicate existing API semantics)
 - Ask users to continue using Layout and Block patterns manually

### Unresolved Questions

- Should Inline live inside the text module?

### Future Possibilities

 - Combine Text and Inline into a more unified abstraction

ognis1205:
Hey @joshka @orhun @j-g00da ,

I drafted an RFC for an Inline widget in `ratatui_core::text` to simplify horizontal alignment. Could you take a look and share any feedback?

Link: https://github.com/ratatui/ratatui/pull/1963#issuecomment-3162893062

Thanks!

joshka:
Bikeshed: I think I'd call this InlineText perhaps. This might makes it more obvious that it's expected to be similar in every way to using Text, but without the lines being rendered vertically. It would also help make it a little more disambiguated from the Inline viewport mode.

Spacer: It looks like (currently) that this is used fairly similarly to the [layout::Spacing](https://docs.rs/ratatui/latest/ratatui/layout/enum.Spacing.html) enum. I wonder if there's a way that this makes sense to be combined?

Use cases:
- I wonder if there's any other obvious places where this makes sense to use in Ratatui to simplify things. Maybe Tabs?
- I wonder if there's other places where we accept Line where it might be worth considering `Into<InlineText>` instead.

Someone presented an idea a long time ago about whether Text should have been designed as a trait instead of a concrete type (I think maybe it was in the discussions, but it's a difficult thing to search for). I wonder if there's any basis for that that crosses over with this (I think probably not, I'm bringing it up as something related that has been considered in the past).

I will definitely dive a bit more into this after we release 0.30

ognis1205:
@joshka 

> Bikeshed: I think I'd call this InlineText perhaps. This might makes it more obvious that it's expected to be similar in every way to using Text, but without the lines being rendered vertically. It would also help make it a little more disambiguated from the Inline viewport mode.

I don’t have a strong opinion on the naming, so I think `InlineText` works well. It makes the functionality provided in relation to `Text` clear and easy to understand.

> Spacer: It looks like (currently) that this is used fairly similarly to the `layout::Spacing` enum. I wonder if there's a way that this makes sense to be combined?

Regarding `layout::Spacing`, I’ll take a closer look at the source code. I was aware of the API but haven’t fully understood it yet. I’ll investigate whether it makes sense to align the design more cleanly with the existing API.

> - I wonder if there's any other obvious places where this makes sense to use in Ratatui to simplify things. Maybe `Tab`s?
> - I wonder if there's other places where we accept `Line` where it might be worth considering `Into<InlineText>` instead.

I’ll also look into the other two points you mentioned: potential use in `Tab`s and places where we currently accept `Line` but might consider `Into<InlineText>` instead. I’m not very familiar with `Tab`s yet, but I think implementing `Into<InlineText>` where needed is a good idea.

> Someone presented an idea a long time ago about whether `Text` should have been designed as a trait instead of a concrete type (I think maybe it was in the discussions, but it's a difficult thing to search for). I wonder if there's any basis for that that crosses over with this (I think probably not, I'm bringing it up as something related that has been considered in the past).

As for the past discussion about whether `Text` should have been designed as a trait instead of a concrete type, I’ll try to search Discord and GitHub to see if I can find any related information. Thanks for pointing that out.

> I will definitely dive a bit more into this after we release 0.30

For now, the implementation in this PR isn’t exactly what I proposed in the RFC, so I plan to rework it once as a PoC for the sake of discussion (without the need for it to be merged). I’d also like to use Ratatui in one of my own projects, so I’ll continue working on that too. After the 0.3.0 release, it would be great if you could take another look when you have time. Thanks a lot for the feedback!

ognis1205:
I plan to resume work on this PR in the next couple of days.

orhun:
Cool, I'm also planning to take a look at it soon

orhun:
feel free to ping me whenever you need a review

ognis1205:
Hey @orhun ,

Thanks! The implementation of `InlineText` as a widget should be something I can handle quickly.

For the other two points:

> Spacer: It looks like (currently) that this is used fairly similarly to the layout::Spacing enum. I wonder if there’s a way that this makes sense to be combined?

> Someone presented an idea a long time ago about whether Text should have been designed as a trait instead of a concrete type (I think maybe it was in the discussions, but it’s a difficult thing to search for). I wonder if there’s any basis for that that crosses over with this (I think probably not, I’m bringing it up as something related that has been considered in the past).

I’d like to dig into GitHub Issues/Discussions and Discord to get more context before making changes, so I’ll probably ask for reviews in stages. Thanks a lot for your support 🙏

orhun:
Sure, take your time. I think potentially using the `Spacing` enum here would be nice. (not sure if it is possible without making {breaking,} changes)

I just had a quick look at the changes and they look pretty clean already :) I think it's a matter of simplifying some parts and making stuff more ergonomic now.

ognis1205:
### Progress Checklist

- [ ] Refactor `Inline` widget into `InlineTest` widget
- [ ] Implement rendering logic to handle multiple `Line`s at once, regardless of alignment (see RFC)
- [ ] Reconsider usage of `layout::Spacing` at the specification level
- [ ] Investigate whether this widget can be reused in other components (e.g., `Tab`)
- [ ] Review `Into<InlineText>` implementation and where it should be applied
- [ ] Revisit past Discord/Discussion threads to reconsider whether `Text` should be defined as a trait

ognis1205:
**WIP: Benchmark results before and after PR**

Here are the benchmark results for preliminary evaluation:

**Before PR:**
```sh
Gnuplot not found, using plotters backend
block/render_empty/100x50
                        time:   [6.3066 µs 6.3116 µs 6.3173 µs]
                        change: [−1.4435% −0.9652% −0.6276%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high mild
block/render_all_feature/100x50
                        time:   [8.7098 µs 8.7278 µs 8.7503 µs]
                        change: [−1.7369% −1.3129% −0.8518%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 15 outliers among 100 measurements (15.00%)
  6 (6.00%) high mild
  9 (9.00%) high severe
block/render_empty/200x50
                        time:   [12.247 µs 12.250 µs 12.254 µs]
                        change: [−0.4233% −0.3659% −0.3085%] (p = 0.00 < 0.05)
                        Change within noise threshold.
  3 (3.00%) high mild
  3 (3.00%) high severe
block/render_all_feature/200x50
                        time:   [15.629 µs 15.636 µs 15.642 µs]
                        change: [−1.2350% −1.1542% −1.0782%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild
block/render_empty/256x256
                        time:   [80.602 µs 80.648 µs 80.711 µs]
                        change: [−0.1695% −0.1015% −0.0025%] (p = 0.01 < 0.05)
                        Change within noise threshold.
Found 2 outliers among 100 measurements (2.00%)
  1 (1.00%) high mild
  1 (1.00%) high severe
block/render_all_feature/256x256
                        time:   [86.056 µs 86.104 µs 86.155 µs]
                        change: [−0.0940% −0.0416% +0.0251%] (p = 0.16 > 0.05)
                        No change in performance detected.
Found 7 outliers among 100 measurements (7.00%)
  6 (6.00%) high mild
  1 (1.00%) high severe
```

**After PR:**
```sh
Gnuplot not found, using plotters backend
block/render_empty/100x50
                        time:   [6.3603 µs 6.3779 µs 6.4065 µs]
                        change: [+0.7812% +0.9742% +1.2143%] (p = 0.00 < 0.05)
                        Change within noise threshold.
  9 (9.00%) high mild
  4 (4.00%) high severe
block/render_all_feature/100x50
                        time:   [8.8289 µs 8.8398 µs 8.8531 µs]
                        change: [+0.5900% +1.0513% +1.4226%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 14 outliers among 100 measurements (14.00%)
  9 (9.00%) high mild
  5 (5.00%) high severe
block/render_empty/200x50
                        time:   [12.304 µs 12.359 µs 12.454 µs]
                        change: [+0.4843% +0.7528% +1.1738%] (p = 0.00 < 0.05)
                        Change within noise threshold.
  5 (5.00%) high mild
  7 (7.00%) high severe
block/render_all_feature/200x50
                        time:   [15.831 µs 16.048 µs 16.448 µs]
                        change: [+1.7035% +2.6225% +4.2100%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 15 outliers among 100 measurements (15.00%)
  9 (9.00%) high mild
  6 (6.00%) high severe
block/render_empty/256x256
                        time:   [80.938 µs 81.284 µs 81.842 µs]
                        change: [+0.5529% +1.0274% +1.6827%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 14 outliers among 100 measurements (14.00%)
  6 (6.00%) high mild
  8 (8.00%) high severe
block/render_all_feature/256x256
                        time:   [86.079 µs 86.123 µs 86.173 µs]
                        change: [−0.0337% +0.0296% +0.0861%] (p = 0.35 > 0.05)
                        No change in performance detected.
Found 3 outliers among 100 measurements (3.00%)
  2 (2.00%) high mild
  1 (1.00%) high severe
```

orhun:
Here is a more readable summary :)

| Benchmark                     | Before PR (µs) | After PR (µs) | Change   | Note              |
|-------------------------------|----------------|---------------|----------|-------------------|
| block/render_empty/100x50     | 6.31           | 6.38          | +1.0%    | Within noise      |
| block/render_all_feature/100x50 | 8.73         | 8.84          | +1.1%    | Within noise      |
| block/render_empty/200x50     | 12.25          | 12.36         | +0.8%    | Within noise      |
| block/render_all_feature/200x50 | 15.64        | 16.05         | **+2.6%** | ⚠️ Regressed      |
| block/render_empty/256x256    | 80.65          | 81.28         | +1.0%    | Within noise      |
| block/render_all_feature/256x256 | 86.10       | 86.12         | 0.0%     | No change         |

ognis1205:
@orhun 
Thanks!

ognis1205:
Note:

`ratatui_widgets::reflow::WordWrapper` provides similar functionality, but for the sake of a PoC, I will implement the necessary equivalent functionality here. It might be better to eventually move `InlineText` under `ratatui_widgets` as an `Inline` widget, rather than keeping it under `ratatui_core::text`.

ognis1205:
temp2

I've implemented a new text-related widget called `InlineText`, and also integrated it into the `Block` widget to support left and right titles in addition to the existing top and bottom ones.

It's still in a proof-of-concept stage, but here's a quick summary of what it does and where it's heading:

### Purpose

 - Renders multiple lines with different alignments and performs text wrapping inline — essentially a horizontal counterpart of the `Text` widget.
 
### Current usage

 - Now used for rendering block titles, including the new `title_left` and `title_right` variants, in addition to the existing top and bottom titles.
 
### Behavior change

 - Previously, block titles were rendered sequentially (left → center → right), often overwriting each other. Now, `InlineText` automatically adjusts and truncates each title based on overlap.

### Features

 - Supports text wrapping for inline rendering.
 - Can render multiple independently aligned text segments within the same line.
 
### Future direction

 - May later be applied to other widgets if this PoC proves useful.

ognis1205:
Note:

We could either use `ratatui_core::layout` internally, or provide an API in `InlineText` that internally calls layout to calculate the drawing area for left/center/right alignment. This should be considered in light of existing API design consistency.

ognis1205:
This PR has grown quite large, so for now I am suspending the following three items:

> - [ ] Investigate whether this widget can be reused in other components (e.g., Tab)
>    - [x] Block: Replaced the existing title rendering with InlineText to unify rendering behavior and make the rendering order independent of implementation details. Also added support for left and right titles.
> - [ ] Review Into<InlineText> implementation and where it should be applied
> - [ ] Revisit past Discord/Discussion threads to reconsider whether Text should be defined as a trait

The main focus for review can be on the `ratatui_core::text::inline_text` module. Changes to `ratatui_widgets::Block` are also included, but these are mainly:

 - Replacing the existing `render_titles` with `InlineText` to simplify the code.
 - Making the rendering order independent of implementation details for different title positions.
 - Adding new left and right title support.
 
 You can consider the `Block` changes primarily as an example usage of `InlineText`. Apologies for the large PR!

ognis1205:
Note:

As an InlineText implemented on the same layer as Text, I personally feel that the current level of abstraction and implementation details are just right. That said, looking at implementations like Paragraph, it might be the correct approach in the future to introduce an abstraction layer for Iterators over Position and Area, allowing text layout and placement to be handled more uniformly.

If InlineText is literally an inline element, it could be treated as a line element within Text and added naturally as part of its line structure.

ognis1205:
Note:

 - https://github.com/ratatui/ratatui/discussions/164#discussioncomment-7033466
 - https://discord.com/channels/1070692720437383208/1072879985762762812/1103128396906627132

joshka:
This is quite large, and with starting a new job recently I've been very time constrained. I'm not sure that this will let up soon. What I'd suggest here is to simplify as much as possible, extract small ideas out of this if possible, and present a good summary. You might find that doing so allows you to present portions of this change that are chewable rather than an entire meal. A good place to try to aim for is something that can be easily reviewed in about 15 minutes (or less). This realistically means 1-300 lines of code (tending to the lower end of this except for pure refactoring changes, or places where there's significant testing / documentation that bloats the numbers).

Do you think this is possible to do with this?

Another approach might be to see what AI tooling can do with summarizing the way to quickly grok the code changes. Give me context, show what the problem being solved is and simply explain the solution before then expanding top down into the details.

ognis1205:
Thanks for the comment, @joshka !

I also agree that this PR has grown quite large, and I appreciate your practical suggestions. Here's a proposal for how we could proceed:

 1. Move this PR back to PoC / Do Not Merge
 - Continue work here to explore where this new widget can be used in existing implementations and what improvements it brings (not for review purposes, but to validate the usefulness of InlineText).
 - This exploratory work will help make future smaller PRs easier to review.
 
 2. Create an issue to track tasks
 - List items to accomplish, such as core logic implementation, documentation, benchmarks, tests, and applying to existing implementations.
 - Use the issue to monitor progress.
 
 3. Create PRs per task
 - Submit code in reviewable chunks.
 - The core logic is likely to generate the most review work, but separating documentation and unit tests into separate PRs should reduce reviewer burden.
 - There may be room for discussion on whether to split core logic, documentation, and unit tests into separate PRs.
 
 Additionally, we could consider using AI tooling in the future to help summarize changes and make them easier to quickly understand.

If this approach sounds good, I can move this PR to draft and continue exploring potential applications first.
Does this sound reasonable to you?

PR #1964, opened 2025-07-02
feat: add set_viewport_height for inline terminal

Marlinski:
See it in action:

https://github.com/user-attachments/assets/a9a6a9bb-6479-499c-a1e2-a5308e18d896


examples/apps/resizable-inline/Cargo.toml
joshka, line 1: small nit: let's call this inline-resizable so that it sorts adjacent to the other inline example.

joshka, line 1: Add this to examples/README.md

examples/apps/resizable-inline/src/main.rs
```diff
@@ -0,0 +1,266 @@
...
+/// This example runs with the Ratatui library code in the branch that you are currently
+/// reading. See the [`latest`] branch for the code which works with the most recent Ratatui
+/// release.
+///
+/// [`latest`]: https://github.com/ratatui/ratatui/tree/latest
+use std::{
+    sync::mpsc,
+    thread,
+    time::{Duration, Instant},
+};
```
joshka: nit: `cargo +nightly fmt` (or cargo xtask fmt`) will put these on lines alone like
```suggestion
use std::sync::mpsc;
use std::sync::thread;
use std::sync::time::{Duration, Instant};
```

examples/apps/resizable-inline/src/main.rs
```diff
@@ -0,0 +1,266 @@
+/// A Ratatui example that demonstrates how to use the resizable inline viewport.
+///
+/// This example shows how to create an inline viewport with a fixed height that can be
+/// dynamically resized during runtime using the new `set_viewport_height()` method.
+///
```
joshka: nit: include the standard header about how examples only work for the main branch (copy it from another example). This generally cuts down answering the question we get a bunch: "I pasted the example and it didn't work, what gives".

examples/apps/resizable-inline/src/main.rs
```diff
@@ -0,0 +1,266 @@
...
+    input_handling(tx.clone());
+    
+    let mut app_state = AppState::new();
+    let app_result = run(&mut terminal, &mut app_state, rx);
+
+    ratatui::restore();
+    app_result
+}
+
+struct AppState {
```
joshka: Let's call this App

examples/apps/resizable-inline/src/main.rs
```diff
@@ -0,0 +1,266 @@
...
+}
+
+fn run<B: Backend>(
+    terminal: &mut Terminal<B>,
+    app_state: &mut AppState,
+    rx: mpsc::Receiver<Event>,
+) -> Result<()>
+where
+    B::Error: Send + Sync + 'static,
+{
```
joshka: Let's make this `App::run(self, terminal: DefaultTerminal, rx: Receiver<Event>)` instead of a free function.
Mainly I do that when there's an obvious self that the function is all about.
You see it here that app_state is called a few times in the body.

examples/apps/resizable-inline/src/main.rs
```diff
@@ -0,0 +1,266 @@
...
+    
+    loop {
+        if redraw {
+            terminal.draw(|frame| render(frame, app_state))?;
+        }
+        redraw = true;
+
+        match rx.recv()? {
+            Event::Input(key_event) => {
+                if key_event.kind == KeyEventKind::Press {
```
joshka:
```suggestion
            Event::Input(key_event) => if key_event.is_press() {
```

examples/apps/resizable-inline/src/main.rs
```diff
@@ -0,0 +1,266 @@
...
+                    match key_event.code {
+                        KeyCode::Char('q') => break,
+                        KeyCode::Char('+') | KeyCode::Char('=') => {
+                            app_state.increase_height();
+                            terminal.set_viewport_height(app_state.current_height)?;
```
joshka: this could be run just before rendering instead of changing during the key handling part. That would help make sure that the visible effects are seen at the same time. It also tends to keep the abstract UI state away from the concrete UI display code. This is a good thing IMO.

examples/apps/resizable-inline/src/main.rs
```diff
@@ -0,0 +1,266 @@
...
+                            let now = app_state.start_time.elapsed();
+                            terminal.insert_before(1, |buf| {
+                                Paragraph::new(format!(
+                                    "📝 Inserted line #{} at {:02}:{:02}",
+                                    count,
+                                    now.as_secs() / 60,
+                                    now.as_secs() % 60
+                                ))
+                                .style(Style::default().fg(Color::Green).add_modifier(Modifier::BOLD))
+                                .render(buf.area, buf);
```
joshka: If you're not wrapping / surrounding with a block, then Text or Line (or even Span) can be used instead of Paragraph. In fact you can probably here just use a string.
```suggestion
                                let mins = now.as_secs() / 60;
                                let secs = now.as_secs() % 60;
                                format!("📝 Inserted line #{count} at {mins:02}:{secs:02}")
                                    .green()
                                    .bold()
                                    .render(buf.area, buf);
```

examples/apps/resizable-inline/src/main.rs
```diff
@@ -0,0 +1,266 @@
...
+                            })?;
+                        }
+                        _ => {}
+                    }
+                }
+            }
+            Event::Resize => {
+                terminal.autoresize()?;
```
joshka: Terminal::autoresize() is called first inside terminal.draw(), most of the time it's unnecessary to call it manually. Do we need it here?

examples/apps/resizable-inline/src/main.rs
```diff
@@ -0,0 +1,266 @@
...
+            Event::Tick => {
+                app_state.frame_count += 1;
+                redraw = false; // Don't redraw on every tick unless content changed
+            }
+        }
+    }
+    Ok(())
+}
+
+fn render(frame: &mut Frame, app_state: &AppState) {
```
joshka: Same deal as above - make this either `App::render()`, or perhaps `impl Widget for &App`

examples/apps/resizable-inline/src/main.rs
```diff
@@ -0,0 +1,266 @@
...
+    
+    let mut app_state = AppState::new();
+    let app_result = run(&mut terminal, &mut app_state, rx);
+
+    ratatui::restore();
+    app_result
+}
+
+struct AppState {
+    current_height: u16,
```
joshka: nit: every variable in a program is "current" unless otherwise mentioned
```suggestion
    height: u16,
```

examples/apps/resizable-inline/src/main.rs
```diff
@@ -0,0 +1,266 @@
...
+}
+
+fn render(frame: &mut Frame, app_state: &AppState) {
+    let area = frame.area();
+    
+    // Create a block with border and title showing current height
+    let main_block = Block::default()
+        .borders(Borders::ALL)
+        .title(format!(" Resizable Viewport (Height: {}) ", app_state.current_height))
+        .title_style(Style::default().fg(Color::Cyan).add_modifier(Modifier::BOLD));
```
joshka: nit: use stylize shortcuts and bordered() constructor:
```suggestion
    let main_block = Block::bordered()
        .title(format!(" Resizable Viewport (Height: {}) ", app_state.current_height).cyan().bold());
```

(and move the title on its own line if that needs to wrap)
```
let title = ...

File: examples/apps/resizable-inline/src/main.rs [hunk 12]
```diff
@@ -0,0 +1,266 @@
...
+    // Status line
+    let elapsed = app_state.start_time.elapsed();
+    let status_line = Paragraph::new(format!(
+        "Runtime: {:02}:{:02} | Frames: {} | Current Height: {}",
+        elapsed.as_secs() / 60,
+        elapsed.as_secs() % 60,
+        app_state.frame_count,
+        app_state.current_height
+    ))
+    .style(Style::default().fg(Color::Green));
```
joshka:
```suggestion
    .green()
```

examples/apps/resizable-inline/src/main.rs
```diff
@@ -0,0 +1,266 @@
...
+        Constraint::Length(1), // Status line
+        Constraint::Length(2), // Controls info
+        Constraint::Min(0),    // Messages list
+        Constraint::Length(1), // Bottom info
+    ]);
+    let [status_area, controls_area, messages_area, bottom_area] = vertical_layout.areas(inner_area);
+    
+    // Status line
+    let elapsed = app_state.start_time.elapsed();
+    let status_line = Paragraph::new(format!(
```
joshka:
```suggestion
    let status_line = format!(
```

examples/apps/resizable-inline/src/main.rs
```diff
@@ -0,0 +1,266 @@
...
+        app_state.frame_count,
+        app_state.current_height
+    ))
+    .style(Style::default().fg(Color::Green));
+    frame.render_widget(status_line, status_area);
+    
+    // Controls info
+    let controls = Paragraph::new(vec![
+        Line::from(vec![
+            Span::styled("Controls: ", Style::default().fg(Color::Yellow).add_modifier(Modifier::BOLD)),
```
joshka:
```suggestion
            "Controls: ".yellow().bold(),
```

examples/apps/resizable-inline/src/main.rs
```diff
@@ -0,0 +1,266 @@
...
+        .take(messages_area.height as usize)
+        .enumerate()
+        .map(|(i, msg)| {
+            let style = if i == 0 {
+                Style::default().fg(Color::Yellow) // Highlight newest message
+            } else {
+                Style::default().fg(Color::White)
+            };
+            ListItem::new(Line::from(Span::styled(msg, style)))
+        })
```
joshka: style() accepts `Into<Style>`, so this can be simplified
```suggestion
        .map(|(i, msg)| {
            let color = if i == 0 { Color::Yellow } else { Color::White };
            ListItem::new(Line::from(msg).styled(color))
        })
```

ratatui-core/src/terminal/terminal.rs
```diff
@@ -522,6 +522,60 @@ where
...
+    ///     viewport: Viewport::Inline(8),
+    /// })?;
+    /// 
+    /// // Later, resize the viewport to 12 lines
+    /// terminal.set_viewport_height(12)?;
+    /// ```
+    pub fn set_viewport_height(&mut self, new_height: u16) -> Result<(), B::Error> {
+        match &mut self.viewport {
+            Viewport::Inline(height) => {
+                if *height != new_height {
```
joshka:
```suggestion
        if let Viewport::Inline(height) && *height != new_height {
```

ratatui-core/src/terminal/terminal.rs
```diff
@@ -522,6 +522,60 @@ where
...
+                    *height = new_height;
+                    
+                    // Keep the same position but update the height
+                    let new_viewport_area = Rect {
+                        x: self.viewport_area.x,
+                        y: self.viewport_area.y,
+                        width: self.viewport_area.width,
+                        height: new_height,
+                    };
```
joshka:
```suggestion
                    let new_viewport_area = Rect {
                        height: new_height,
                        ..self.viewport_area
                    };
```

ratatui-core/src/terminal/terminal.rs
```diff
@@ -522,6 +522,60 @@ where
...
+    /// use ratatui::{Terminal, TerminalOptions, Viewport};
+    /// 
+    /// let mut terminal = Terminal::with_options(backend, TerminalOptions {
+    ///     viewport: Viewport::Inline(8),
+    /// })?;
+    /// 
+    /// // Later, resize the viewport to 12 lines
+    /// terminal.set_viewport_height(12)?;
+    /// ```
+    pub fn set_viewport_height(&mut self, new_height: u16) -> Result<(), B::Error> {
```
joshka: **This is the main comment of the PR**

How does this differ from the resize() method? (Docs should clarify both if there's a significant difference).

I think adding tests to help clarify some of this. This is an area that's currently lacking a bit in the terminal as we haven't prioritized it. The tests belong as unit tests in this file (not in a tests/ folder) and should hit the TestBackend to confirm various like the state of the terminal display before and after this call.

In particular, there's some undocumentd logic in the computed_inline_size method that I know exists, but haven't really dug into why it's there / what its intent is. It would be great to dig into this a bit more as part of this - mainly to help answer the questions:
- why use this method instead of resize?
- how does resize work?
- should this be a separate method, or just a fix to the resize method?

joshka (commented):
Looking good. Thanks for the PR on this.
Take a look at the "main comment" first as it's a bit of a why / design focused question, and it will help ensure that this is easy to maintain going forward.

The other comments are nits / implementation details / stylistic things that help keep the examples simple so that the amount of code is generally as small as possible.

ratatui-core/src/terminal/terminal.rs
Marlinski: Thanks for your super fast review!

The reason that motivated this PR is because I needed to be able to resize my inline viewport height to fit different modal of different size. I first tried doing it with resize but it takes a rect with X and a Y and width which I wasn't sure how to reliably get. The width I can get from terminal.size(). So in short:

- why use this method instead of resize? 
  - not sure how to use resize for this job, required information seems gated behind private attributes
  - simpler and more direct rather than first fetching parameters needed (x,y,width) to make the resize call

- how does resize work?
  - I am not sure that current  resize method address the scrolling issue that is specific to inline viewport.

- should this be a separate method, or just a fix to the resize method?
  - I suppose we could have a single entry point to resize
  - maybe also have handy methods like move_x, move_y, set_height, set_width that would feed into the same resize?
  - this solution is good enough for now?

Marlinski:
I commited a fix to my PR to handle scrolling the terminal when hitting the bottom in similar fashion with insert_before(). Now increasing the height push the scroll buffer down to make room for the new height.

ratatui-core/src/terminal/terminal.rs
joshka: First up, I want to validate the use case - resizing like you're doing here is perfectly valid and something I want to support. Nothing in this should distract you from that, my push back is mostly about how this is achieved and what that means for API methods.

Let me clarify this a little from several perspectives

## User perspective:
Resize is a problem from a discoverability perspective for inline viewports because the area that the used as the size is computed, and because the area isn't something that is ever provided by the user in their configuration. Instead they only initially provide a TerminalOptions param with a Viewport::Inline(length) field.

This observation and the fact this method is only applicable to Viewport::Inline leads me to the idea of resetting the viewport rather than making this a set_viewport_height method. E.g. what about fn reset_viewport(viewport: ViewPort) here, which handles properly changing from fixed / fullscreen / inline and also handles sizing changes? This seems like a good way to avoid this method only working when the mode is inline.

We invest a lot in making sure the docs on docs.rs provide all the info a user needs to be able to use every function in Ratatui without too much ambiguity. I personally see it as a bug in the docs if users have to ask how something works, and there's nowhere I can point to as an RTFM response.

## Maintainer perspective:
The existing code that handles the inline mode stems from with_options and the resize function. It currently does some shenanigans with calculating the height and areas. This is not well documented right now, so I don't actually know how it works at all (mostly I haven't dived into it at all ever). This PR skips a bunch of that logic when resizing, so now I don't know the underlying logic + I don't know what is skipped and the impact of that. To get this PR over the line, the rationale and intent about the existing logic has to be captured and documented. I'd suggest docs + unit tests for this. This is the clean kitchen before cooking phase, and unfortunately this is a place where it's necessary before making this change.

Additionally to that, when documenting the behavior of methods like this it's important to capture the intent behind any of the not obvious parts. It's easy to read the code and see what's happening, but the code doesn't capture intent. I don't know if what the code is doing is what it's supposed to be doing, or if what it's doing is a bug. That information can only come from comments that reveal the intention.

Making these changes without cleaning up and documenting adds a risk that the behavior of the new approach is incorrect in some way and will need to be changed. Breaking changes in -core (new in the upcoming 0.30) will be something that we try to avoid doing too often, to make sure that widget libraries don't need to be upgraded as often as they do presently. So getting this right first up is much more important than a widget change which can be iterated on.

Marlinski: Gotcha! 

Make sense. I'll try to dig a little deeper when I have time to see how it works with relation with other resizing method and identify if:

- it isn't messing internal state by resizing "manually" instead of using the other method
- if it could be implemented in another method

Marlinski:
there's another more deeper issue. Whenever the viewport height reach the terminal maximum, the logic breaks because the viewport position (y) cannot be negative (encoded as u16) and as such, it isn't possible at the moment to have a viewport that's larger than the terminal height.

j-g00da:
In theory, would just making it signed be enough to make it work?

joshka:
There's probably some painful edge cases around resizing the height and resizing the terminal at the same time here too. viewport height should never be negative (I'm not sure I understand the proposed idea for that though).

Marlinski:
@joshka if you ever used claude code, there's sometime a permission modal (if it writes a large file at once) that can be high enough to be larger than your terminal height in which case you have to scroll the terminal up to see the entire file. Once you click on yes, the modal disappear and become push into the scroll before (in a similar fashion as insert_before). If I want to replicate this behaviour, I need a way to increase my viewport to be bigger than terminal height so that I can then delete the modal. I don't see how I can do that currently. I can't push to insert before either because once it is pushed to the scroll buffer I can't delete it (at least not easily, not sure how I would do that either).

joshka:
To the best of my knowledge, there's no way a viewport larger than the terminal can work in Ratatui. The abstraction is purely that an inline viewport is an allocation of a fixed rectangular area positioned on the screen as-if that area is where the cursor started. The addressable cursor locations on the screen are limited to the actual number of rows and columns on the screen.

Concretely, If I have a 200x50 terminal, and run an app. Say 2 lines are used to display the output of `cargo run`, so my app starts running on the third line (index 2 as Ratatui uses 0 based lines). If I have an Inline viewport height 10, then the Rect used for rendering is `Rect { x: 0, y: 2, width: 200, height: 10 }`. All cursor movement calls are able to address x: 0..199 and y: 0..49. Those calls will be y: 2..11 with the inline viewport described. There's nothing that we can do to make those access negative areas or areas beyond the available space. When drawing the buffer, it just draws to the locations on the underlying right place.

So because of this understanding, I'm having difficulty understanding how Claude would be doing something different.
I haven't used claude code yet, is there any way that you can show a small vid of that happening?
If you're on a mac, cmd+shift+5 pops up a tool that can screenshot + record video easily (PC / linux 🤷)

It would be great to see how this works to visualize it a bit better. Also OpenAI Codex expressed a similar need in a recent issue, (and is being rewritten using Ratatui...), so knowing how other frameworks do it is something useful to know.

It's indeed possible that there are more complex ways of interacting with the terminal that defeat this somehow - the VT and ECMA standards are deep, so I wouldn't be surprised if there's some way to modify off screen parts of the terminal.

Marlinski:
here you go:

https://github.com/user-attachments/assets/0ee446bc-55b5-4098-859b-5023f90c668f

joshka:
That's super interesting. I'm trying to wrap my head around how it works still. I have a hypothesis, but need more information...

Can you do that same task again (perhap just "a file with the numbers 1 to 100 each on a line" to save time). When it's done, scroll up to where the text in your shell is before you started claude is visible? I'm wondering if this is a really clever fake out, by using a combination of the primary and alternate screen and rendering things twice? Also, do a full scroll up after accepting the change, to see if that's any different.

Marlinski:
dang you are right, I did not even noticed that !

https://github.com/user-attachments/assets/fa980c2c-abe6-49f2-a880-26e0c65d4d1b

so what do you think happens here ? because even if it deleted what was before it can still grow very long.

joshka:
> so what do you think happens here ? because even if it deleted what was before it can still grow very long.

I think it definitely started on the main screen, flipped to the alternate screen when it generated 100 lines. Pushed that all to the full buffer as plain old lines in the buffer. Then it didn't go back to the main screen (yet?). I suspect the delete part would be switching back to the main screen where the long text never happened.

Shenanigans are afoot ... :)

What I'm not seeing in the second demo is the clearing of the long permission modal, so it's difficult to tell beyond a guess.
It's possible that something like a line being cleared and then a backspace being processed to remove the line from the terminal, but I'm not 100% sure. Without seeing the full process again as mentioned, I can't confirm the hypothesis.

If you're using iTerm2, the **Session | Terminal State** menu should help show whether you're in alternate screen mode to confirm.

Marlinski:
hmm now I am confused haha:

https://github.com/user-attachments/assets/631bab0f-2c1f-4a09-9918-2a5e3fa30dd1

joshka:
Ok, cool - it's not using the alternate screen. Instead I think it's probably clearing the entire scrollback history and rewriting it? I'm assuming that it's doing so from the start through to the current output and not somehow deleting lines backwards.

(You could always ask claude to inspect its own code to describe what's happening there btw... ;))

Marlinski:
ok so to replicate:

I would insert_before(my permission modal) while remembering the number of lines I pushed to the scroll before. resize my viewport to a minimum, and when it is accepted or deny, I delete the scroll buffer by the same amount of line. 

Currently is there a way to delete content, something like delete_before() ?

joshka:
> Currently is there a way to delete content, something like delete_before() ?

Not as far as I'm aware.

That's where I'm really curious to understand what's happening there. Maybe check it out with Ghostty? It has a "Terminal Inspector" feature that can capture what's going on with a tonne of detail.

kdheepak:
I believe Claude Code is using React Ink - https://github.com/vadimdemedes/ink

And they must be storing the entire state of the application when rendering to the terminal (as HTML elements). And when that is passed to React Ink, React Ink only changes the parts of the terminal that need to be changed. 

I haven't looked into how React Ink implements this though.

joshka:
> I believe Claude Code is using React Ink - https://github.com/vadimdemedes/ink
> 
> And they must be storing the entire state of the application when rendering to the terminal (as HTML elements). And when that is passed to React Ink, React Ink only changes the parts of the terminal that need to be changed.
> 
> I haven't looked into how React Ink implements this though.

Looks like there's some deep info on this at:

https://github.com/vadimdemedes/ink/issues/382

joshka:
I think this is probably something that we should defer to after 0.30, so marking it for 0.31 milestone.

joshka:
Made with VHS

ratatui-core/src/terminal/terminal.rs
```diff
@@ -522,6 +522,61 @@ where
...
+        let Viewport::Inline(height) = &mut self.viewport else {
+            return Ok(());
+        };
+        if *height == new_height {
+            return Ok(());
+        }
+
+        let old_height = std::mem::replace(height, new_height);
```
j-g00da:
```suggestion
        let old_height = core::mem::replace(height, new_height);
```

ratatui-core/src/terminal/terminal.rs
```diff
@@ -522,6 +522,61 @@ where
...
+        self.clear()?;
+
+        let new_y = match new_height.cmp(&old_height) {
+            std::cmp::Ordering::Greater => {
```
j-g00da:
```suggestion
            core::cmp::Ordering::Greater => {
```

j-g00da (changes requested):

joshka:
The approach that I'm taking on this is to go a little deeper into the various methods involved here (e.g. `resize`, `compute_inline_size`, `set_viewport_area`) and make sure that they're clear, obvious and tested (as much as is possible). This is an area that has not had a good deal of focus until now, so digging in and making sure it's solid will help ensure that this gets implemented right. I actually also think we can make this part of the forthcoming 0.30 release instead of deferring to 0.31.

joshka:
Made with VHS

joshka:
@Marlinski just a heads up, I'm pushing changes to your PR branch fairly liberally here. This is something we regularly do in this repo, but may not be something that not all github users are super familiar with.

If you want to "play along at home", make sure that you pull the branch locally to see what's going on and replace your existing code. If you want to keep another branch pointing at what you originally submitted for reference, that could be worth it too (`git branch foo 439b1fe`). It's not strictly necessary to do this, but can be helpful.

Marlinski:
No issue @joshka and thanks for taking the time to work on this awaited PR ! Until this one is merged, [shai](https://github.com/ovh/shai) will keep pointing to my own original work (in the forked repo) but I'll make sure to point it back when it is done :)

joshka:
> Until this one is merged, [shai](https://github.com/ovh/shai) will keep pointing to my own original work (in the forked repo)

Sounds good. I didn't even think about the problems this sort of approach could cause for projects that point at git branches like this. But it looks like you're pointing at a different branch than this PR is cut from, so that's not a problem here. Something I'll keep in mind for future though (and will update our contrib docs to add some info about it - https://github.com/ratatui/ratatui/pull/2029).

joshka:
One kind of annoying bug in this is that resizing horizontal breaks the display hard:

https://github.com/user-attachments/assets/3cf76e65-04ca-484b-acac-9349b052450a

(I have a hypothesis that his is caused by a bug in the way that the buffer resize code works, but I haven't dug into it yet.

This doesn't really look correct to me:

https://github.com/ratatui/ratatui/blob/71ef65b6248bd8b22fe775125447948071868cbe/ratatui-core/src/buffer/buffer.rs#L408-L418

aldanor:
Confirmed, something's off with horizontal resizes. @joshka would you have any ideas?

On a side note, is there any help needed with this PR? Seems like it's mostly there, would be great to push it to the finish line.

aldanor:
@joshka Seems like there might be something janky with horizontal resizing on main as well, possibly unrelated to this PR in particular, is this kind of behaviour expected with inline viewport? (this is the official inline viewport example, on main branch)

https://github.com/user-attachments/assets/1ea6a908-8f4d-4081-a8b8-94ba5add521e

---
Also, if you resize down aggressively, you will lose content (I think because of that `.content.truncate()` in the buffer resize). So you will end up straight up losing lines in the history so that the inline viewport example output will look like this
```
...
Finished download 16 in 44ms
Finished download 13 in 388ms
Finished download 170 in 1436ms
Finished download 173 in 609ms
...
```

---
On a side note, I could also make it crash (not this PR, but main branch) by resizing aggressively:

```rust
The cursor position could not be read within a normal duration

Location:
   src/main.rs:179
```
which is this line in the inline-view example:
```rust
            Event::Resize => {
                terminal.autoresize()?;
            }

```

aldanor:
Added a separate issue (https://github.com/ratatui/ratatui/issues/2086) so as not to divert this one.

With that in mind (horizontal resize doesn't seem to be related to this branch in particular), is there anything else currently blocking this PR?

orhun:
Yeah, CI needs to be fixed & reviews required.

aldanor:
@orhun @Marlinski Seems like `scroll_up()` method no longer exists hence the errors. In particular, this method doesn't exist at all when "scrolling-regions" feature is enabled. What's the correct behaviour here under scrolling-regions?

> Sounds good. I didn't even think about the problems this sort of approach could cause for projects that point at git branches like this. But it looks like you're pointing at a different branch than this PR is cut from, so that's not a problem here. Something I'll keep in mind for future though (and will update our contrib docs to add some info about it - #2029).

@joshka Pointing to a particular unreleased branch of ratatui causes massive problems in real projects, because you now have to fork every single tui-* crate as well, otherwise it won't compile due to different versions of ratatui being used (unless you use patch.crates-io of course... which is a bit of a hack in itself). So it would be very nice to have this branch merged eventually.

Marlinski:
@aldanor @joshka  not sure what's going on with scroll_up, was it deleted ? 
as for the resize issue as it isn't specific to this PR and should be dealt with separately.

orhun:
`scroll_up` is only available when `scrolling-regions` feature is _not_ enabled.

That's why passing `--all-features` is failing the build I guess.

retorted:
Feedback note: I've been previewing this feature recently by pulling the crate directly from the fork at `Marlinski/viewport-inline-resize`, and it's great :raised_hands:

orhun:
Thanks for the feedback!

Not sure what's left to do here but it's added to the milestone 0.31. Not sure if we can make this into 0.30 which should be soon ™️

joshka:
Hey there - apologies for not getting back onto this earlier. I was traveling for the last month or so. I'll take another look soon.

atkinson:
What do we need to do to get this merged? I'm happy to give some time to getting it done.

joshka:
I will take a proper look at this asap. Probably this weekend.

It's mostly been in my bucket of too deep to dig into with the limited time I've had to spend on Ratatui (life, travel, new job, travel for new job, etc.)

Marlinski:
@joshka congratulation for your new job and looking forward to merging this PR! I wonder, will you treat the resize issue as a separate PR ?

atkinson:
I have time and skills and am very happy to provide whichever combination is needed to progress this feature.

joshka:
The TLDR of the status on this is that I want to ensure that the existing behavior is well documented and well tested before making this change.

I updated some docs in https://github.com/ratatui/ratatui/pull/2312 - those interested in this, please take a look there to validate that things look right.

The next step would be adding a decent amount of characterization tests for the existing behavior.

joshka:
https://github.com/ratatui/ratatui/pull/2315 follows that up with a slew of new tests. It also refactors the terminal into smaller chunks of related behavior to make it easy to have the code and relevant tests side by side and obviously do the right thing. There are also a few bugs that it brought up in how clearing the terminal works.

Once those are merged, this PR will have to be re-done so as to fit with the new docs and tests. The base need on this is to evaluate how it fits in the library as a whole.

The horizontal resizing problem that happens when inline mode is on seems unrelated, but adding more complexity (at least in how many paths you can get to a state) makes it a little bit more difficult to solve that problem. I'd like to ensure that that problem is properly fixed here too.

I think the right API on this could be to add `set_viewport(ViewPort)` allowing the terminal to be switched from FullScreen / Fixed / Inline rather than `set_viewport_height` which only supports changing the Inline. Anyone got thoughts on this?

examples/apps/inline-resizable/README.md
```diff
@@ -0,0 +1,51 @@
...
+- **Event Handling**: Separate thread for input and tick events
+- **Render Function**: Displays the UI with status information, controls, message log, and height gauge
+- **Resize Logic**: Handles viewport height changes with validation and feedback
+
+## Implementation Notes
+
+- The viewport height is constrained between 3 and 20 lines for demonstration purposes
+- Each resize operation is logged to show the functionality in action
+- The example uses a tick-based rendering system to update runtime statistics
+- Auto-resize is handled for terminal window changes
```
orhun, line 51:
```suggestion
# Inline resizable demo

This example demonstrates the resizable inline viewport functionality. It shows how
to create an inline viewport with a fixed height that can be dynamically resized during runtime
using the `terminal.set_viewport_height()` method.

To run this demo:

```shell
cargo run -p inline-resizable
```
```

**Review by orhun — APPROVED** [review]
LGTM!

Note to myself: record the VHS tape before merging this.

File: examples/apps/inline-resizable/src/main.rs [hunk 22]
```diff
@@ -0,0 +1,202 @@
...
+            .collect::<List>();
+        frame.render_widget(message_list, messages_area);
+
+        // Bottom info with dynamic gauge showing height percentage
+        let height_percentage = (f64::from(self.height) / 20.0).min(1.0);
+        let gauge = Gauge::default()
+            .gauge_style(Color::Blue)
+            .ratio(height_percentage)
+            .label(format!("Height: {}/20", self.height));
+        frame.render_widget(gauge, bottom_area);
```
orhun, line 200: the max height is 15

```suggestion
        // Bottom info with dynamic gauge showing height percentage
        let height_percentage = (f64::from(self.height) / 15.0).min(1.0);
        let gauge = Gauge::default()
            .gauge_style(Color::Blue)
            .ratio(height_percentage)
            .label(format!("Height: {}/15", self.height));
        frame.render_widget(gauge, bottom_area);
```

PR #1973, opened 2025-07-10
fix(widgets): fix centered block title truncation

### LINKED ISSUE #1936: Centered block titles truncate oddly

ognis1205:
## Description

- Expose `ratatui_core::text::Line::spans_after_width` using `#[doc(hidden)]` to support the expected behavior in `ratatui_widgets::Block::render_center_title`.
- Considered refactoring `spans_after_width` into a proper method on the `Line` struct, but opted against it to avoid introducing new public APIs in a bugfix-oriented PR.
- Also considered adding a `skip_width: Option<usize>` field to `Line`, but decided against it for the same reason.

## Related Issue

Closes #1936

## Checklist

- [x] All unit tests pass
- [x] Added regression test for multi-title center truncation
- [x] Documentation updated if necessary

### DIFF
--- a/ratatui-widgets/src/block.rs
+++ b/ratatui-widgets/src/block.rs
@@ -1937,6 +1937,16 @@ mod tests {
     }
 
     #[test]
+    fn left_titles() {
+        let mut buffer = Buffer::empty(Rect::new(0, 0, 10, 1));
+        Block::new()
+            .title("L12")
+            .title("L34")
+            .render(buffer.area, &mut buffer);
+        assert_eq!(buffer, Buffer::with_lines(["L12 L34   "]));
+    }
+
+    #[test]
     fn left_titles_truncated() {
         let mut buffer = Buffer::empty(Rect::new(0, 0, 10, 1));
         Block::new()
@@ -1944,6 +1954,16 @@ mod tests {
             .title("L67890")
             .render(buffer.area, &mut buffer);
         assert_eq!(buffer, Buffer::with_lines(["L12345 L67"]));
+    }
+
+    #[test]
+    fn center_titles() {
+        let mut buffer = Buffer::empty(Rect::new(0, 0, 10, 1));
+        Block::new()
+            .title(Line::from("C12").centered())
+            .title(Line::from("C34").centered())
+            .render(buffer.area, &mut buffer);
+        assert_eq!(buffer, Buffer::with_lines([" C12 C34  "]));
     }
 
     /// Note: this test is probably not what you'd expect, but it is how it works in the current
@@ -1959,7 +1979,17 @@ mod tests {
             .title(Line::from("C12345").centered())
             .title(Line::from("C67890").centered())
             .render(buffer.area, &mut buffer);
-        assert_eq!(buffer, Buffer::with_lines(["C12345 678"]));
+        //assert_eq!(buffer, Buffer::with_lines(["12345 C678"]));
+    }
+
+    #[test]
+    fn right_titles() {
+        let mut buffer = Buffer::empty(Rect::new(0, 0, 10, 1));
+        Block::new()
+            .title(Line::from("R12").right_aligned())
+            .title(Line::from("R34").right_aligned())
+            .render(buffer.area, &mut buffer);
+        assert_eq!(buffer, Buffer::with_lines(["   R12 R34"]));
     }
 
     #[test]
--- a/ratatui-widgets/src/block.rs
+++ b/ratatui-widgets/src/block.rs
@@ -13,7 +13,7 @@ use ratatui_core::layout::{Alignment, Rect};
 use ratatui_core::style::{Style, Styled};
 use ratatui_core::symbols::border;
 use ratatui_core::symbols::merge::MergeStrategy;
-use ratatui_core::text::{Line, spans_after_width};
+use ratatui_core::text::Line;
 use ratatui_core::widgets::Widget;
 use strum::{Display, EnumString};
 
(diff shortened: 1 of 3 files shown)


ognis1205:
cc: @joshka 

This PR is ready for your review when convenient. Let me know if anything needs clarification!

ratatui-widgets/src/block.rs
```diff
@@ -933,41 +933,80 @@ impl Block<'_> {
     }
 
     /// Render titles in the center of the block
-    ///
-    /// Currently this method aligns the titles to the left inside a centered area. This is not
-    /// ideal and should be fixed in the future to align the titles to the center of the block and
-    /// truncate both sides of the titles if the block is too small to fit all titles.
     #[expect(clippy::similar_names)]
     fn render_center_titles(&self, position: TitlePosition, area: Rect, buf: &mut Buffer) {
```
joshka, line 936: This method is now a bit longer than I'd ideally like to see (~75 Loc). I started looking at how this might be broken up a bit and ended up with something that's a bit smaller.

ratatui-widgets/src/block.rs
```diff
@@ -933,41 +933,80 @@ impl Block<'_> {
...
             .collect_vec();
         let total_width = titles
             .iter()
-            .map(|title| title.width() as u16 + 1) // space between titles
+            // Add space width between titles.
+            .map(|title| title.width() as u16 + 1)
             .sum::<u16>()
-            .saturating_sub(1); // no space for the last title
+            // Remove redundant space width, i.e., the trailing one.
+            .saturating_sub(1);
```
joshka: I changed the comments a bunch to be more succinct

ratatui-widgets/src/block.rs
```diff
@@ -933,41 +933,80 @@ impl Block<'_> {
...
-                break;
-            }
-            let title_width = title.width() as u16;
-            let title_area = Rect {
-                width: title_width.min(titles_area.width),
+        // If the total title width fits within the available area, center the titles.
+        if total_width <= titles_area.width {
+            // Adjust the rendering area to center the titles horizontally.
+            titles_area = Rect {
+                x: titles_area.left() + (titles_area.width.saturating_sub(total_width) / 2),
```
joshka: Added more temp vars to help make the line density a bit lower

ratatui-widgets/src/block.rs
```diff
@@ -933,41 +933,80 @@ impl Block<'_> {
...
             };
-            buf.set_style(title_area, self.titles_style);
-            title.render(title_area, buf);
-
-            // bump the titles area to the right and reduce its width
-            titles_area.x = titles_area.x.saturating_add(title_width + 1);
-            titles_area.width = titles_area.width.saturating_sub(title_width + 1);
+            for title in titles {
+                if titles_area.is_empty() {
+                    break;
```
joshka: This check is not necessary for area > width - it will never be triggered (codecov agrees)

ratatui-widgets/src/block.rs
```diff
@@ -933,41 +933,80 @@ impl Block<'_> {
...
+                        width: span_width.min(title_area.width),
+                        ..title_area
+                    };
+                    span.render(span_area, buf);
+                    // Move the rendering cursor to the right.
+                    titles_area.x = titles_area.x.saturating_add(span_width);
+                    titles_area.width = titles_area.width.saturating_sub(span_width);
+                });
+                // Advance skip_width by the full title width.
+                skip_width = skip_width.saturating_sub(title_width);
```
joshka: I replaced this with `title.clone().right_aligned().render(...)`, while it does do a clone of the line, the number of titles on screen and their length is likely not enough to worry about in any reasonable program, so I'm unconcerned about any performance hit on this.

ratatui-widgets/src/block.rs
joshka: With my changes it's down to 50, which is still a bit high. Suggestions below to address that.

ratatui-widgets/src/block.rs
```diff
@@ -933,41 +933,56 @@ impl Block<'_> {
...
-        let mut titles_area = Rect {
-            x: titles_area.left() + (titles_area.width.saturating_sub(total_width) / 2),
-            ..titles_area
-        };
-        for title in titles {
-            if titles_area.is_empty() {
-                break;
+            .saturating_sub(1);
+
+        if total_width <= area.width {
```
joshka, line 948: For reducing the size of this method, either:
- chop the inner part of the if blocks (e.g. render_center_titles_fitting, render_center_titles_exceeds); or
- see if the for loop can be factored so that the fitting and exceeds parts have a common implementation - these are fairly close with this new approach. I think probably extracting the methods is probably better as it makes it clear that the truncation code isn't necessary unless the titles are chopped.

ratatui-widgets/src/block.rs
```diff
@@ -933,41 +933,56 @@ impl Block<'_> {
...
+            // is truncated by the area width.
+            let mut offset = total_width.saturating_sub(area.width) / 2;
+            for title in titles {
+                if area.is_empty() {
+                    break;
+                }
+                let width = area.width.min(title.width() as u16).saturating_sub(offset);
+                let title_area = Rect { width, ..area };
+                buf.set_style(title_area, self.titles_style);
+                buf.set_style(title_area, title.style);
```
joshka: not necessary as the title.render() call below sets this

ratatui-widgets/src/block.rs
```diff
@@ -1946,6 +1971,16 @@ mod tests {
...
+        Block::new()
+            .title(Line::from("C12").centered())
+            .title(Line::from("C34").centered())
+            .render(buffer.area, &mut buffer);
+        assert_eq!(buffer, Buffer::with_lines([" C12 C34  "]));
+    }
+
     /// Note: this test is probably not what you'd expect, but it is how it works in the current
     /// implementation. Update this if the behavior changes.
     ///
```
joshka: Remove

joshka (commented):
Self review of update

ratatui-widgets/src/block.rs
j-g00da: First option sounds like the right way to go

j-g00da (changes requested):
`render_center_titles` has to be split but other than that LGTM.

joshka (approved):

joshka:
Thanks for the PR @ognis1205

ognis1205:
@joshka @j-g00da 

Thanks for the merge and the refinements! Appreciate the cleanup — learned a lot from your refactor. Glad I could help!

In the original PR, I’ll start by outlining the intended behavior to clarify the spec. In addition to that, I’ll move forward with a PoC implementation. I might also start exploring another issue in parallel depending on bandwidth.

PR #1974, opened 2025-07-11
chore: use Option in Cell instead of " "

joshka:
This change makes the `Cell::symbol` field an `Option<CompactString>`,
allowing it to represent an empty cell as `None` instead of an empty
string. The rationale for this is to later allow the merge symbol
functionality to act differently based on whether a cell has previously
held a symbol or not, rather than always merging with an empty string.
This will help make it possible to merge borders with titles with spaces
and other symbols, without assuming that an empty string is always
equivalent to no symbol.

- Default is now derived as `Option::None` works correctly.
- PartialEq and Eq implementations are updated to treat `None` the same
  as an empty string.
- merge_symbol against an empty cell will now just set the symbol
  rather than calling MergeStrategy::merge with an empty string.
- PartialEq, and Hash are manually implementated instead of being
  derived, and are updated to treat `None` equal to an empty string.

### DIFF
--- a/ratatui-core/src/buffer/cell.rs
+++ b/ratatui-core/src/buffer/cell.rs
@@ -4,7 +4,7 @@ use crate::style::{Color, Modifier, Style};
 use crate::symbols::merge::MergeStrategy;
 
 /// A buffer cell
-#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+#[derive(Debug, Default, Clone)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub struct Cell {
     /// The string to be drawn in the cell.
@@ -15,7 +15,7 @@ pub struct Cell {
     /// buffer for short strings.
     ///
     /// See <https://github.com/ratatui/ratatui/pull/601> for more information.
-    symbol: CompactString,
+    symbol: Option<CompactString>,
 
     /// The foreground color of the cell.
     pub fg: Color,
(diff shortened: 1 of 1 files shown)


joshka:
Remember merge is new enough that there's no existing behavior to break. Currently we use the default strategy which is to always replace regardless of character right? I'm fine with breaking the alpha behavior either by adding a new option or modifying one of the merge options so that what to do with non border characters is properly handled.

To draw a block over existing text when a merge mode is active, you would want to draw a Clear widget first (we might want to make this use empty rather than spaces if it's not already). It might be neat to build that into the Block options too. E.g. `Block::bordered().clear_first()` or something like that.

j-g00da:
The default behavior still merges with replace and this is my main issue as this would be a breaking change from 0.29 and not just the alpha. 

Do you want to create an exception for replace to not have this rule?
This seems fine but would need some docs as this can be confusing I think

> `Block::bordered().clear_first()`

Yes, this would be nice to have.

j-g00da (approved):

orhun (approved):
LGTM

orhun:
Commit message suggestion: `chore(cell): use Option instead of empty string for symbol`

joshka:
> Commit message suggestion: `chore(cell): use Option instead of empty string for symbol`

It's actually a space " ", not an empty string

joshka:
> The default behavior still merges with replace and this is my main issue as this would be a breaking change from 0.29 and not just the alpha.

I'm not sure I understand. In 0.29, there was no merge strategy. So rendering a block border always replaces any symbol.

In 0.30, the block border merge strategy is replace, which should do render the same thing.

orhun:
> It's actually a space " ", not an empty string

Ah yes sorry that's what I meant

j-g00da:
In 0.30 technicaly we _always merge_, just the default strategy is `Replace`, so when comparing with 0.29 we should take the behavior of `MergeStrategy::Replace` to the account. That's why now we have to make an exception for Replace when it comes to merging with titles (#1977)

joshka:
> In 0.30 technicaly we _always merge_, just the default strategy is `Replace`, so when comparing with 0.29 we should take the behavior of `MergeStrategy::Replace` to the account. That's why now we have to make an exception for Replace when it comes to merging with titles (#1977)

Ah got it - yes. The Replace strategy should act as a replacement regardless of what the character is.

Issue #1939, opened 2025-06-25
Preserve block titles while merging borders


j-g00da:
## Problem
Currently when merging borders, titles are completely overwritten.

## Solution
Titles should always be preserved while merging block borders.

## Alternatives
-

## Additional context
https://github.com/ratatui/ratatui/pull/1899#issuecomment-2954292552

PR #1977, opened 2025-07-11
feat: preserve block titles when merging borders

j-g00da:
Resolves #1939 

This uses #1974, which should be merged first.


ratatui-core/src/symbols/merge.rs
```diff
@@ -300,10 +300,14 @@ impl MergeStrategy {
     /// [Box Drawing Unicode block]: https://en.wikipedia.org/wiki/Box_Drawing
     /// [`Cell::merge_symbol`]: crate::buffer::Cell::merge_symbol
     pub fn merge<'a>(self, prev: &'a str, next: &'a str) -> &'a str {
+        // Replace should always just return the last symbol.
+        if self == Self::Replace {
+            return next;
+        }
```
j-g00da: I don't like this here as it complicates the logic, let me know if you have a better idea where to put this check.

orhun: Hmm I'm not sure what this does without looking into the whole implementation again tbh - it's not clear to me how the problem is being fixed from the PR or commit description :D

j-g00da: Context:
https://github.com/ratatui/ratatui/pull/1974#issue-3221733135
https://github.com/ratatui/ratatui/pull/1899#issuecomment-2954292552

We can now distinguish between space symbols (e.g. in title) and empty cell, so if cells contains text instead of border characters, we don't override it so the titles are still visible. We don't want to do this for replace, as this should always overwrite (so we return `next` here).

joshka: I think this seems reasonable to me to have here. It's the simple degenerate case. I much prefer early exits for this sort of thing (guard clauses etc.)

ratatui-core/src/symbols/merge.rs
```diff
@@ -300,10 +300,14 @@ impl MergeStrategy {
...
         let (Ok(prev_symbol), Ok(next_symbol)) =
             (BorderSymbol::from_str(prev), BorderSymbol::from_str(next))
         else {
-            return next;
+            return prev;
```
joshka: I think this actually needs to work out which of the two items is a border and preferentially choose the non-border character.
- "A" + "-" => "A"
- "-" + "A" => "A"

We should maybe also consider whether we need to rename any of the variant names (or add extra ones) to make this clear (and allow for future behavior where this is not the case). (Not 100% sure of this right now though).

j-g00da: > I think this actually needs to work out which of the two items is a border and preferentially choose the non-border character.

Right! Will add a test for this too.

> We should maybe also consider whether we need to rename any of the variant names (or add extra ones) to make this clear (and allow for future behavior where this is not the case). (Not 100% sure of this right now though).

Or maybe this should be a separate setting?

ratatui-core/src/symbols/merge.rs
```diff
@@ -300,11 +300,19 @@ impl MergeStrategy {
...
+        let Ok(prev_symbol) = BorderSymbol::from_str(prev) else {
+            return prev;
         };
+        let Ok(next_symbol) = BorderSymbol::from_str(next) else {
+            return next;
+        };
+
         if let Ok(merged) = prev_symbol.merge(next_symbol, self).try_into() {
             return merged;
         }
```
j-g00da: Actually I think in case of both symbols not being a border, we should use `next`.

j-g00da: (changed in https://github.com/ratatui/ratatui/pull/1977/commits/e4d0caf8482fe2f545d2b8fda76f25d17c0b30bd)

j-g00da:
It still needs docs but let me know if this implementation makes sense to you.
Wdyt about adding a separate setting for this behavior? I'm not really sure about it.

joshka:
Ps the way the tests updated to show how this change affects things shows that the tests were really well chosen to be good example style tests.

j-g00da:
Let's merge this as-is then.

PR #1985, opened 2025-07-13
chore: change Borders::NONE to a proper const

joshka:
https://docs.rs/bitflags/latest/bitflags/#zero-bit-flags

> Flags with no bits set should be avoided because they interact strangely with [Flags::contains](https://docs.rs/bitflags/latest/bitflags/trait.Flags.html#method.contains) and [Flags::intersects](https://docs.rs/bitflags/latest/bitflags/trait.Flags.html#method.intersects). A zero-bit flag is always contained, but is never intersected. The names of zero-bit flags can be parsed, but are never formatted.

Removing this simplifies the manual Debug impl that previously had to check for Borders::NONE and now does not.

---

This is technically breaking (for anyone using contains / intersects), but not in a way that should have been reasonable to ever been relied upon as defined behavior. I'm leaning towards not documenting this as a breaking change because it's such an unlikely edge case to hit.

### DIFF
--- a/ratatui-widgets/src/borders.rs
+++ b/ratatui-widgets/src/borders.rs
@@ -10,8 +10,6 @@ bitflags! {
     #[derive(Default, Clone, Copy, Eq, PartialEq, Hash)]
     #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
     pub struct Borders: u8 {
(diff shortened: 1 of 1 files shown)


joshka:
Incidentally, crossterm has similar bitflags issues.

PR #1993, opened 2025-07-15
chore: remove cargo_metadata dep from xtask

joshka:
Removed due to hard N-2 MSRV requirement, whereas we use a soft N-2 (only update when necessary). This makes it painful to be able to test our actual msrv (as the xtask has to be built with the version that its deps support, while still wanting to check the msrv version, so we'd end up with 2 versions in the one CI task and this would get annoying to check).

See https://github.com/rust-lang/cargo/issues/15746 for more details.

Partially implements #1820 - mainly as a problem solution rather than a specific goal to use cargo-hack


j-g00da (approved):
CI needs to be updated but LGTM, approving so it can be merged later.

PR #1996, opened 2025-07-18
fix: panic when rendering widgets on too small buffer

j-g00da:
Fixes panic on overflow on horizontal `Barchart` and `RatatuiMascot`, adds proper tests to all widgets.

### DIFF
--- a/ratatui-widgets/src/barchart.rs
+++ b/ratatui-widgets/src/barchart.rs
@@ -1473,6 +1473,7 @@ mod tests {
         ]);
         assert_eq!(buffer, expected);
     }
+
     #[rstest]
     #[case::horizontal(Direction::Horizontal)]
     #[case::vertical(Direction::Vertical)]
--- a/ratatui-widgets/src/barchart.rs
+++ b/ratatui-widgets/src/barchart.rs
@@ -1477,7 +1477,7 @@ mod tests {
     #[rstest]
     #[case::horizontal(Direction::Horizontal)]
     #[case::vertical(Direction::Vertical)]
-    fn buffer_overflow(#[case] direction: Direction) {
+    fn render_in_minimal_buffer(#[case] direction: Direction) {
         let chart = BarChart::default()
             .data(&[("A", 1), ("B", 2)])
             .bar_width(3)
--- a/ratatui-widgets/src/barchart.rs
+++ b/ratatui-widgets/src/barchart.rs
@@ -1487,5 +1487,21 @@ mod tests {
         let mut buffer = Buffer::empty(Rect::new(0, 0, 1, 1));
         // This should not panic, even if the buffer is too small to render the chart.
         chart.render(buffer.area, &mut buffer);
+        assert_eq!(buffer, Buffer::with_lines([" "]));
+    }
+
+    #[rstest]
+    #[case::horizontal(Direction::Horizontal)]
+    #[case::vertical(Direction::Vertical)]
+    fn render_in_empty_buffer(#[case] direction: Direction) {
+        let chart = BarChart::default()
+            .data(&[("A", 1), ("B", 2)])
+            .bar_width(3)
+            .bar_gap(1)
+            .direction(direction);
+
+        let mut buffer = Buffer::empty(Rect::new(0, 0, 0, 0));
+        // This should not panic, even if the buffer is empty.
+        chart.render(buffer.area, &mut buffer);
     }
 }
--- a/ratatui-widgets/src/barchart.rs
+++ b/ratatui-widgets/src/barchart.rs
@@ -515,7 +515,7 @@ impl BarChart<'_> {
             let margin = u16::from(label_size != 0);
             Rect {
                 x: area.x + label_size + margin,
-                width: area.width.saturating_sub(label_size + margin),
+                width: area.width.saturating_sub(label_size).saturating_sub(margin),
                 ..area
             }
         };
--- a/ratatui-widgets/src/barchart.rs
+++ b/ratatui-widgets/src/barchart.rs
@@ -515,7 +515,7 @@ impl BarChart<'_> {
             let margin = u16::from(label_size != 0);
             Rect {
                 x: area.x + label_size + margin,
-                width: area.width - label_size - margin,
+                width: area.width.saturating_sub(label_size + margin),
                 ..area
             }
         };
@@ -579,10 +579,10 @@ impl BarChart<'_> {
     }
 
     fn render_vertical(&self, buf: &mut Buffer, area: Rect) {
-        let label_info = self.label_info(area.height - 1);
+        let label_info = self.label_info(area.height.saturating_sub(1));
 
         let bars_area = Rect {
-            height: area.height - label_info.height,
+            height: area.height.saturating_sub(label_info.height),
             ..area
         };
 
@@ -722,6 +722,7 @@ mod tests {
     use ratatui_core::layout::Alignment;
     use ratatui_core::style::{Color, Modifier, Stylize};
     use ratatui_core::text::Span;
+    use rstest::rstest;
 
     use super::*;
     use crate::borders::BorderType;
@@ -1471,5 +1472,19 @@ mod tests {
             "\u{202f}███",
         ]);
         assert_eq!(buffer, expected);
+    }
+    #[rstest]
+    #[case::horizontal(Direction::Horizontal)]
+    #[case::vertical(Direction::Vertical)]
+    fn buffer_overflow(#[case] direction: Direction) {
+        let chart = BarChart::default()
+            .data(&[("A", 1), ("B", 2)])
+            .bar_width(3)
+            .bar_gap(1)
+            .direction(direction);
+
+        let mut buffer = Buffer::empty(Rect::new(0, 0, 1, 1));
+        // This should not panic, even if the buffer is too small to render the chart.
+        chart.render(buffer.area, &mut buffer);
     }
 }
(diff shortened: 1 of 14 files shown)


ratatui-widgets/src/clear.rs
```diff
@@ -69,4 +69,12 @@ mod tests {
...
         assert_eq!(buffer, expected);
     }
+
+    #[test]
+    fn buffer_overflow() {
+        let mut buffer = Buffer::empty(Rect::new(0, 0, 1, 1));
+        let clear = Clear;
+        // This should not panic, even if the buffer is too small to render the clear widget.
+        clear.render(buffer.area, &mut buffer);
+    }
```
j-g00da: Ok this one doesn't make sense, will remove it.

ratatui-widgets/src/scrollbar.rs
```diff
@@ -970,6 +970,8 @@ mod tests {
     #[case::position_8("<---#####>", 8, 10)]
     #[case::position_9("<----####>", 9, 10)]
     #[case::position_one_out_of_bounds("<----####>", 10, 10)]
+    #[case::position_few_out_of_bounds("<----####>", 15, 10)]
+    #[case::position_very_many_out_of_bounds("<----####>", 500, 10)]
```
j-g00da, line 974: This is just copied from `render_scrollbar_with_symbols`

ratatui-widgets/src/scrollbar.rs
```diff
@@ -1000,7 +1002,9 @@ mod tests {
     #[case::position_8("<---#####>", 8, 10)]
     #[case::position_9("<----####>", 9, 10)]
     #[case::position_one_out_of_bounds("<----####>", 10, 10)]
-    fn render_scrollbar_vertical_rightl(
+    #[case::position_few_out_of_bounds("<----####>", 15, 10)]
+    #[case::position_very_many_out_of_bounds("<----####>", 500, 10)]
+    fn render_scrollbar_vertical_right(
```
j-g00da, line 1007: Same, copied from `render_scrollbar_with_symbols`.
Additionally fixed a typo "rightl" -> "right", I believe there's no reason to do this in a separate PR.

ratatui-widgets/src/barchart.rs
```diff
@@ -1472,4 +1473,18 @@ mod tests {
         ]);
         assert_eq!(buffer, expected);
     }
+    #[rstest]
```
orhun, line 1477: nit: newline before tests

```suggestion

    #[rstest]
```

ratatui-widgets/src/barchart.rs
```diff
@@ -1472,4 +1473,18 @@ mod tests {
...
+    #[case::horizontal(Direction::Horizontal)]
+    #[case::vertical(Direction::Vertical)]
+    fn buffer_overflow(#[case] direction: Direction) {
```
orhun: It sounds like we are testing the buffer overflow, which is something that does not happen. Maybe renaming this to `does_not_overflow` or `no_buffer_overflow` would be more appropriate. WDYT?

This comment also applies to the rest of the PR.

ratatui-widgets/src/block.rs
```diff
@@ -2134,4 +2134,14 @@ mod tests {
             .render(buffer.area, &mut buffer);
         assert_eq!(buffer, Buffer::with_lines(["  C1R67890"]));
     }
+
+    #[test]
+    fn buffer_overflow() {
+        let mut buffer = Buffer::empty(Rect::new(0, 0, 1, 1));
+        // This should not panic, even if the buffer is too small to render the block.
+        Block::bordered()
+            .title("I'm too big for this buffer")
```
orhun, line 2143: yeah I was also too big for her heart I guess

ratatui-widgets/src/calendar.rs
```diff
@@ -302,4 +302,15 @@ mod tests {
     fn test_today() {
         CalendarEventStore::today(Style::default());
     }
+
+    #[test]
+    fn buffer_overflow() {
+        let mut buffer = Buffer::empty(Rect::new(0, 0, 1, 1));
+        let calendar = Monthly::new(
+            Date::from_calendar_date(2023, Month::January, 1).unwrap(),
```
orhun: can we avoid unwrap here? Maybe expect it or do proper error handling.

Also:

```suggestion
            Date::from_calendar_date(1984, Month::January, 1).unwrap(),
```

ratatui-widgets/src/gauge.rs
```diff
@@ -586,4 +586,20 @@ mod tests {
             }
         );
     }
+
+    #[test]
+    fn buffer_overflow_gauge() {
```
orhun:
```suggestion
    fn gauge_does_not_overflow() {
```

ratatui-widgets/src/gauge.rs
```diff
@@ -586,4 +586,20 @@ mod tests {
...
+        let mut buffer = Buffer::empty(Rect::new(0, 0, 1, 1));
+        let gauge = Gauge::default().percent(50);
+        // This should not panic, even if the buffer is too small to render the gauge.
+        gauge.render(buffer.area, &mut buffer);
+    }
+
+    #[test]
+    fn buffer_overflow_line_gauge() {
```
orhun:
```suggestion
    fn line_gauge_does_not_overflow() {
```

ratatui-widgets/src/logo.rs
```diff
@@ -236,4 +236,12 @@ mod tests {
             ])
         );
     }
+
+    #[test]
+    fn buffer_overflow() {
+        let mut buffer = Buffer::empty(Rect::new(0, 0, 1, 1));
+        let logo = RatatuiLogo::new(Size::Tiny);
```
orhun: Use parameterized tests for different sizes

ratatui-widgets/src/tabs.rs
```diff
@@ -676,4 +676,14 @@ mod tests {
...
     }
+
+    #[test]
+    fn buffer_overflow() {
+        let mut buffer = Buffer::empty(Rect::new(0, 0, 1, 1));
+        let tabs = Tabs::new(vec!["Tab1", "Tab2", "Tab3", "Tab4"])
+            .select(1)
+            .divider("|");
+        // This should not panic, even if the buffer is too small to render the tabs.
+        tabs.render(buffer.area, &mut buffer);
```
orhun, line 687: I'm curious, what does this actually render? Anything? A single char?

ratatui-widgets/src/calendar.rs
j-g00da: lmao

ratatui-widgets/src/barchart.rs
j-g00da: Right... I thought about it a bit and I think I will go with something like `render_in_minimal_buffer`, as this seems the most descriptive.

ratatui-widgets/src/calendar.rs
joshka: Unwrap is often fine in tests. It generally indicates that you expect that call never to fail, and if it does then the test is wrong and you mainly should be looking at a stack trace etc.. Not sure what the linter does with it from memory.

ratatui-widgets/src/barchart.rs
```diff
@@ -515,7 +515,7 @@ impl BarChart<'_> {
             let margin = u16::from(label_size != 0);
             Rect {
                 x: area.x + label_size + margin,
-                width: area.width - label_size - margin,
+                width: area.width.saturating_sub(label_size + margin),
```
joshka: Does this break when margin = u16::MAX?

ratatui-widgets/src/barchart.rs
```diff
@@ -1472,4 +1473,19 @@ mod tests {
...
+    fn render_in_minimal_buffer(#[case] direction: Direction) {
+        let chart = BarChart::default()
+            .data(&[("A", 1), ("B", 2)])
+            .bar_width(3)
+            .bar_gap(1)
+            .direction(direction);
+
+        let mut buffer = Buffer::empty(Rect::new(0, 0, 1, 1));
+        // This should not panic, even if the buffer is too small to render the chart.
+        chart.render(buffer.area, &mut buffer);
```
joshka, line 1489: Not panicking is the default assertion of a test. I'm wondering if we can assert more here. Would it be worth checking the buffer output in these tests (here and below) instead of only checking that the render does not panic?

joshka (commented):
I think it may be worth checking empty buffer size in this as well as minimal buffer size. 

Also taking a step out from this, doing area outside of buffer for all widgets (perhaps as another PR on that part). WDYT?

j-g00da:
Yup, I was already thinking about it, will add this.

ratatui-widgets/src/barchart.rs
j-g00da: I wasn't sure how meaningful would it be to check a single character in such edge cases, and that was why I initially didn't check the buffers.

Thinking about it now, maybe it would be worth checking it anyway, it wouldn't take much time to introduce this.

ratatui-widgets/src/barchart.rs
j-g00da: That's true, will just do saturating sub two times instead

ratatui-widgets/src/tabs.rs
j-g00da: added assertions to all minimal buffer tests

orhun (approved):

ratatui-widgets/src/paragraph.rs
```diff
@@ -1219,4 +1219,25 @@ mod tests {
             ])
         );
     }
+
+    #[test]
+    fn render_in_minimal_buffer() {
+        let mut buffer = Buffer::empty(Rect::new(0, 0, 1, 1));
+        let paragraph = Paragraph::new(
+            "This is a long paragraph that should not panic when rendered in a very small buffer area.",
+        );
```
joshka: nit: we don't need quite so large a paragraph to test this :D

PR #1997, opened 2025-07-18
feat(block): enable clearing the area before rendering

quietpigeon:
An implementation of `Block::bordered().clear_first()`, which renders `Clear` before rendering a `Block`.

Related to https://github.com/ratatui/ratatui/issues/1975.


orhun:
Looking good to me. Would be also nice to add unit tests :)

ratatui-widgets/src/block.rs
```diff
@@ -290,6 +294,19 @@ impl<'a> Block<'a> {
         block
     }
 
+    /// Creates a new block with [Clear] rendered first.
```
j-g00da: This should be phrased differently, it now suggests that this is a constructor.

ratatui-widgets/src/block.rs
```diff
@@ -781,12 +798,19 @@ impl Widget for &Block<'_> {
             return;
         }
         buf.set_style(area, self.style);
+        self.render_clear(area, buf);
```
j-g00da: `render_clear` should go before setting style, as otherwise it will reset the block style too.

ratatui-widgets/src/block.rs
```diff
@@ -781,12 +798,19 @@ impl Widget for &Block<'_> {
...
         self.render_titles(area, buf);
     }
 }
 
 impl Block<'_> {
+    fn render_clear(&self, area: Rect, buf: &mut Buffer) {
+        if self.clear_first {
+            Clear.render(area, buf);
+        }
+    }
```
j-g00da: I don't think this has to be a separate method, I would prefer this directly in `render`.

j-g00da:
Oh I didn't realize this is still a draft, sorry if I commented on something you already planned to change!

ratatui-widgets/src/block.rs
```diff
@@ -290,6 +294,32 @@ impl<'a> Block<'a> {
...
+    /// Clears the area to allow overdrawing with [Clear] before rendering.
+    ///
+    /// ```
+    /// use ratatui::widgets::Block;
+    ///
+    /// let block = Block::bordered().clear_first();
+    /// ```
+    ///
+    /// When rendered, it is equivalent to:
+    /// ```
```
orhun, line 307: nit:

```suggestion
    /// When rendered, it is equivalent to:
    ///
    /// ```
```

ratatui-widgets/src/block.rs
quietpigeon: Fixed!

ratatui-widgets/src/block.rs
quietpigeon: Fixed.

ratatui-widgets/src/block.rs
quietpigeon: I have rephrased the doc string and have added some more examples. I don't have a lot of experience in writing doc strings, so please let me know if there's anything to add or fix :)

ratatui-widgets/src/block.rs
```diff
@@ -1402,6 +1456,21 @@ mod tests {
...
+            .clear_first()
+            .render(buffer.area, &mut buffer);
+        #[rustfmt::skip]
+        let expected = Buffer::with_lines([
+            "┌────────┐",
+            "│        │",
+            "└────────┘",
+        ]);
+        assert_eq!(buffer, expected);
+    }
```
j-g00da, line 1506: nit: maybe add a second test with no `.clear_first()` too

ratatui-widgets/src/block.rs
```diff
@@ -290,6 +294,32 @@ impl<'a> Block<'a> {
         block
     }
 
+    /// Clears the area to allow overdrawing with [Clear] before rendering.
```
j-g00da:
```suggestion
    /// Clears the area with [Clear] before rendering.
```

ratatui-widgets/src/block.rs
```diff
@@ -233,6 +234,8 @@ pub struct Block<'a> {
     padding: Padding,
     /// Border merging strategy
     merge_borders: MergeStrategy,
+    /// Whether or not to clear the area before rendering
```
j-g00da:
```suggestion
    /// Whether to clear the area before rendering
```

j-g00da (commented):
Generally LGTM, left some minor comments.
It would be also good to go through `Clear` usages in the examples to see if some of these can be changed to use this method instead.

ratatui-widgets/src/block.rs
```diff
@@ -1402,6 +1456,21 @@ mod tests {
         );
     }
 
+    #[test]
+    fn render_clear_block() {
+        let mut buffer = Buffer::with_lines(["xxxxxxxxxx"; 3]);
+        Block::bordered()
+            .clear_first()
+            .render(buffer.area, &mut buffer);
```
orhun: I think what would be cooler is to either modify this test or add another test to treat this as a popup:

```suggestion
        let mut buffer = Buffer::with_lines(["xxxxxxxxxxxx"; 8]);
        Block::bordered().clear_first().render(
            buffer.area.inner(Margin {
                horizontal: 2,
                vertical: 2,
            }),
            &mut buffer,
        );
```

Results in:

```        content: [
            "xxxxxxxxxxxx",
            "xxxxxxxxxxxx",
            "xx┌──────┐xx",
            "xx│      │xx",
            "xx│      │xx",
            "xx└──────┘xx",
            "xxxxxxxxxxxx",
            "xxxxxxxxxxxx",
        ],
```

Which I think shows the intention of `clear_first` more _clearly_.

orhun (approved):

examples/apps/demo2/src/tabs/about.rs
```diff
@@ -47,8 +47,11 @@ fn render_crate_description(area: Rect, buf: &mut Buffer) {
         vertical: 4,
         horizontal: 2,
     });
-    Clear.render(area, buf); // clear out the color swatches
-    Block::new().style(THEME.content).render(area, buf);
+    // clear out the color swatches before rendering the block
+    Block::new()
+        .clear_first()
```
j-g00da: nit: this comment should be on the `.clear_first()` line

examples/apps/popup/src/main.rs
```diff
@@ -52,10 +52,9 @@ fn render(frame: &mut Frame, show_popup: bool) {
     frame.render_widget(Block::bordered().title("Content").on_blue(), content);
 
     if show_popup {
-        let popup = Block::bordered().title("Popup");
-        let popup_area = centered_area(area, 60, 20);
         // clears out any background in the area before rendering the popup
```
j-g00da: nit: maybe we can remove this comment now? I think the code is very self-descriptive.
```suggestion
```

examples/apps/demo2/src/tabs/about.rs
quietpigeon: Fixed.

examples/apps/popup/src/main.rs
quietpigeon: Removed.

ratatui-widgets/src/block.rs
joshka: This should be:
```suggestion
    /// Clears the area with [`Clear`] before rendering.
```

ratatui-widgets/src/block.rs
joshka: oooh, the pun... savage :D

ratatui-widgets/src/block.rs
```diff
@@ -290,6 +294,33 @@ impl<'a> Block<'a> {
         block
     }
 
+    /// Clears the area with [Clear] before rendering.
+    ///
+    /// ```
```
joshka, line 299: The doc comments produce the API docs, so they should in general capture enough information that the reader never has to wonder about why the method exists. The first line should be a simple sentence about the effect, this likely doesn't need to capture how it does things (e.g. it doesn't need to state it uses the Clear widget). Then there's a blank line and a more detailed narrative. Try to capture the essence of the PR, and information from the Clear widget and the popup article on the website in this part. Think about this from the perspective of someone that asks at least the following questions:
- how do I clear the background behind a block
- why do I need to do this
- how do I render a popup that displays over existing text on the screen

These shape people's search terms as well as their reading.

ratatui-widgets/src/block.rs
```diff
@@ -290,6 +294,33 @@ impl<'a> Block<'a> {
...
+    /// use ratatui::widgets::{Block, Clear};
+    ///
+    /// fn draw_block_on_clear(f: &mut Frame, area: Rect) {
+    ///     let block = Block::bordered();
+    ///     f.render_widget(Clear, area);
+    ///     f.render_widget(block, area);
+    /// }
+    /// ```
+    #[must_use = "method moves the value of self and returns the modified value"]
+    pub const fn clear_first(mut self) -> Self {
```
joshka, line 319: I like the implementation as it's reasonably simple and straightforward. I'm not convinced though that clear_first is 100% the best name for this (it's adequate though if we can't find a better one). As a name, it sort of captures how this acts rather than what the effect is. I'm not sure though what a better name might be here. Perhaps [bikeshed](https://en.wikipedia.org/wiki/Law_of_triviality) a few good options before we merge this?

One approach to this might be to consider whether in the future we might extend this idea to allow some transparency to see the items behind the block. In that case assume that the bool becomes an enum and has some sort of name that captures how it works. What would you call the two cases that we're handling here today (ignoring transparency)? What is the concept?

(This semi- relates back to https://rust-lang.github.io/api-guidelines/type-safety.html#c-custom-type, which doesn't explicitly say don't use bools, but it helps frame boolean options as sometimes having intention revealing names rather than just yes/no)

joshka (commented):
General approval of implementation.
My discussion point is mainly about whether we want to spend a little bit more time on the naming before committing to it. Happy to hear push back on that point, but I think it's worth considering.

ratatui-widgets/src/block.rs
quietpigeon: I would call the two cases `Inline` and `Floating` as clearing the area gives a `floating` effect to the block. I could make that into an enum, something like: 
```rust
pub enum DisplayMode {
    #[default]
    Inline,
    Floating,
}
```
Which gives us
```rust
let block = Block::bordered().display(DisplayMode::Floating);
```
What do you think?

j-g00da: What about `OverwriteMode` (or simply `Overwrite`) with variants `All` and `Diff`.

Seems descriptive to me.

Or maybe we can stick to `DisplayMode`/`RenderMode` but change variants to a single `Overwrite(bool)`, then we can in the future add `Filter(fn(...) -> ...)` variant that can apply e.g. transparency.

WDYT?

quietpigeon: What does `OverwriteMode::Diff` imply?

j-g00da: Widgets generally only overwrite cells they edit, which you can say crates sort of a "diff", this is the base case, where it only overwrites using this diff. I'm open to other names for that.

So I've been thinking about this a bit more:

(Just a pseudocode, this should derive Copy, Clone, Default etc.)
```rust
// Not sure about the name, probably needs some suffix like "Kind"?
// Otherwise we end up with `Overwrite(Overwrite)` in `RenderStrategy`
pub enum Overwrite {
    #[default]
    Diff,
    All,
}

enum RenderStrategy {
    #[default]
    Overwrite(Overwrite),
    // added in the future
    // this can probably just take another widget?
    // can be in some way unified in the v0.31 with shadows
    Filter(fn(&mut Buffer, Rect)),
}

impl From<OverwriteKind> for RenderStrategy {
// ...
}

// probably won't work like that, but it's just to give the idea
// it's out of scope for this PR anyway
impl From<fn(&mut Buffer, Rect)> for RenderStrategy {
// ...
}
```

Then we can implement a method in Block like this:
```rust
fn render_strategy(self, strategy: Into<RenderStrategy>) {
    self.render_strategy = strategy.into()
    self
}
```

And use it like this:
```rust
Block::bordered().render_strategy(Overwrite::All)
// or in the future
Block::bordered().render_strategy(|&mut buffer, area| {...})
```

---

Alternative option which just came to my mind writing this. Since we can probably use just widgets as filters... and `Clear` is a widget...

We can probably make a method that is something like this:

```rust
fn pre_render<W: Widget>(self, widget: W) {
```

Then we can clear the area like this:

```rust
Block::bordered().pre_render(Clear)
```
Or apply some kind of filter same way...

WDYT?

cc: @joshka

quietpigeon: I see. Yeah, this sounds good to me. 

I can implement this right away, or should we wait for @joshka 's opinion on this?

joshka: I think adding the pre-render approach seems like adds a bunch of complexity that doesn't really differentiate itself from just rendering the widget beforehand. Does that make sense?

One related thing that I just thought about is how we might want to think about this with the new border merging code, so we might want to differentiate how we handle the border and the inner area of this to make that work.

We may also want to leave the styles and symbols differently.

what about something like:

```rust
let block = Block::bordered().clear_symbols(Clear::Inner);
let block = Block::bordered().clear_styles(Clear::Outer);
```

or flipping the method / enum:

```rust
let block = Block::bordered().clear_inner(Clear::Symbols);
let block = Block::bordered().clear_outer(Clear::Styles);
```

(default would be `Option<Clear>::None`)

(also using Clear as the type name clashes with the Clear widget, so this isn't quite right either).

j-g00da: My proposition would be to first come up with test cases that use border merging over some existing buffer content, that are awkward to achieve now and see what would help making it easier.

Intuitively clearing the whole area and clearing the inner area seems to be the most useful, can't think of any use case for clearing the border only.

joshka: +1 advocating for a user oriented design is always a good idea.

quietpigeon: Pardon me, I think I understand what `clear_inner()` and `clear_outer()` mean, but not `clear_symbols()` and `clear_styles`. Would you mind elaborating on what are their expected behaviours?

joshka: The difference is whether we're resetting Cell::symbol, Cell::style (or perhaps both I guess). Mostly there I'm experimenting with some ideas for what it might look like if we take a perspective of the user interacting with the API. But @j-g00da suggests one better step, thinking about what problems we can solve with this change that are related, and how those displays might look.

The hold up here is mainly that we're really only replacing `Clear::render(area, buf)`, with something that on the surface is added complexity, so we're kind of looking for the reasons to make it more general. It may be that this is a solution looking for a problem. That's a bit backwards unfortunately and I usually hate doing it when it's avoidable. So perhaps let's let this one lie a bit for now, and we can come back to it when needed.

orhun:
I think this is a nice feature to include in the new release and I'm afraid we will drag this longer if we keep it open for improvements/bikeshedding. How about we get this in for now and track the new API in a follow up PR / issue?

j-g00da:
> I think this is a nice feature to include in the new release and I'm afraid we will drag this longer if we keep it open for improvements/bikeshedding. How about we get this in for now and track the new API in a follow up PR / issue?

I'm not sure about merging something that we know will most likely be modified and will a cause breaking change in v0.31

ratatui-widgets/src/block.rs
kdheepak: > Intuitively clearing the whole area and clearing the inner area seems to be the most useful, can't think of any use case for clearing the border only.

Would a border animation fall under this category?

PR #2000, opened 2025-07-19
docs: update heading image for Ratatui 0.30.0 release 🎉

j-g00da:


orhun:
PR 2000, damn,

j-g00da:
Added initial implementation to examples, still needs refactoring but this is how it looks:

image

joshka:
Super nitpick. I think the empty space after the version number looks a little off.
How about adding 1 more gradient to make it an extra line to push the version number to the last line, and then right align the version number (it could also look ok only right aligning the version).

examples/apps/ratatui-release-header/src/main.rs
```diff
@@ -0,0 +1,181 @@
...
+            GRADIENT_RED[index],
+            GRADIENT_AMBIENT[index],
+            GRADIENT_BLUE[index],
+        ),
+    };
+    Color::Rgb(r, g, b)
+}
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let viewport = Viewport::Inline(10);
```
joshka: It may be worth considering whether we want to go with the same size approach as the current demo2 dimensions. This was chosen to fit the github automatic repo preview sizes. That said maybe we leave demo2 on those parts and make this new image show up in docs.rs and the Readme to give people a break from the demo2 destroy gif.

examples/apps/ratatui-release-header/src/main.rs
```diff
@@ -0,0 +1,181 @@
...
+    let mut terminal = ratatui::init_with_options(TerminalOptions { viewport });
+    loop {
+        terminal.draw(render_header)?;
```
joshka: If this returns an error, it will prevent ratatui::restore running.

examples/apps/ratatui-release-header/src/main.rs
```diff
@@ -0,0 +1,181 @@
...
+
+        for (rect, color) in letters.iter().zip(ROYGBIV) {
+            let gradient_color = get_gradient_color(&color, idx as usize);
+            frame.render_widget(
+                Block::new().style(Style::new().fg(gradient_color).dim()),
+                rect.offset(offset),
+            );
+        }
+        frame.render_widget(RatatuiLogo::small(), logo_area.offset(offset));
+    }
```
joshka: Neat idea. So this basically:
- renders 5x7 blocks, with each effectively setting the foreground color of the logo
- renders the logo 5 times, overlapping each by one line

examples/apps/ratatui-release-header/src/main.rs
```diff
@@ -0,0 +1,181 @@
...
+
+    let inner_backends = menu_block.inner(backends_area);
+    frame.render_widget(
+        Paragraph::new(vec![
+            Line::from("> ratatui-crossterm"),
+            Line::from("> ratatui-termion"),
+            Line::from("> ratatui-termwiz"),
+        ]),
+        inner_backends,
+    );
```
joshka: Can this part use Paragraph::new(...).block(...) with the new overlapping block behavior?

examples/apps/ratatui-release-header/src/main.rs
j-g00da: I have that in mind, I want to add some space around and just center the whole thing.

j-g00da:
> How about adding 1 more gradient to make it an extra line to push the version number to the last line, and then right align the version number (it could also look ok only right aligning the version).

Makes sense.
Btw. I thought about maybe adding some short line next to version number that we can change in each release, e.g. "Even More Cheese!" or "Chef's Pick!".
WDYT? @ratatui/maintainers

j-g00da:
image
image

Added one more gradient layer and made the dimmensions 81x18 (as in demo2).
If we decide to add some more text there, then leaving version number on the left would make more sense.
I also dimmed it, because leaving it white crated an optical illusion where it seemed to be too low.

j-g00da:
Interesting. My previous screenshots were using kitty terminal. VHS handles dimming differently:
ratatui-release-header
Arguably a bug in ttyd, as I don't think it should be possible to have three colors in a single cell:
image

joshka:
That tty thing could be a gif encoding problem perhaps, try adding a screenshot command to the tape file to see how it looks in PNG. (In case you haven't used it before, VHS requires least 1 frame after the Screenshot command, so add a small sleep after).

The way the logo's cell positions are aligned is intentionally a bit weird so that the letters are more nicely kerned rather than leaving a full cell width space between the A and the T. I wonder if there's a bug with dim + quadrant char + rendering several times to the same cell?

examples/apps/ratatui-release-header/src/main.rs
```diff
@@ -0,0 +1,215 @@
...
+use ratatui::crossterm::execute;
+use ratatui::crossterm::terminal::{EnterAlternateScreen, LeaveAlternateScreen};
+use ratatui::layout::{Offset, Spacing};
+use ratatui::prelude::*;
+use ratatui::symbols::merge::MergeStrategy;
+use ratatui::widgets::{Block, BorderType, Padding, Paragraph, RatatuiLogo};
+use ratatui::{DefaultTerminal, TerminalOptions, Viewport};
+
+const SEMVER: &str = "0.30.0";
+const RELEASE_NAME: &str = "Bryndza";
```
orhun: idea:

How about we set a version in `Cargo.toml` and use `CARGO_PKG_VERSION` here?

Similarly, I believe we can do the same for `RELEASE_NAME` as well. (`CARGO_PKG_DESCRIPTION`)

examples/apps/ratatui-release-header/src/main.rs
```diff
@@ -0,0 +1,215 @@
...
+];
+const BACKENDS: [&str; 3] = [
+    "> ratatui-crossterm",
+    "> ratatui-termion",
+    "> ratatui-termwiz",
+];
+
+const FG_COLOR: Color = Color::Rgb(246, 214, 187);
+const BG_COLOR: Color = Color::Rgb(20, 20, 50);
+const MENU_BORDER_COLOR: Color = Color::Rgb(255, 255, 160);
```
orhun:
```suggestion
const FG_COLOR: Color = Color::Rgb(246, 214, 187); // #F6D6BB
const BG_COLOR: Color = Color::Rgb(20, 20, 50);    // #141432
const MENU_BORDER_COLOR: Color = Color::Rgb(255, 255, 160); // #FFFFA0
```

this would look better on most editors:

image

examples/apps/ratatui-release-header/src/main.rs
```diff
@@ -0,0 +1,215 @@
...
+        .set_style(area, Style::new().fg(FG_COLOR).bg(BG_COLOR));
+
+    let horizontal_layout = Layout::horizontal([Constraint::Length(31), Constraint::Length(23)])
+        .spacing(Spacing::Overlap(1));
+    let [left_area, right_area] = area
+        .centered(
+            Constraint::Length(53),
+            Constraint::Length(MAIN_DISHES.len() as u16 + BACKENDS.len() as u16 + 3),
+        )
+        .layout(&horizontal_layout);
```
orhun: The constraint numbers look so arbitrary lol

Any chance that we dynamically set them without messing up the code? Not a big deal though

examples/apps/ratatui-release-header/Cargo.toml
orhun, line 1: I think we can remove the `ratatui-` prefix from the directory and example name.

examples/apps/ratatui-release-header/src/main.rs
```diff
@@ -0,0 +1,215 @@
...
+fn render_menu(frame: &mut Frame, area: Rect) {
+    let vertical_layout = Layout::vertical([
+        Constraint::Length(MAIN_DISHES.len() as u16 + 2),
+        Constraint::Fill(BACKENDS.len() as u16 + 2),
+    ])
+    .spacing(Spacing::Overlap(1));
+    let [main_dishes_area, backends_area] = area.layout(&vertical_layout);
+
+    render_menu_block(frame, main_dishes_area, "Main dishes", &MAIN_DISHES);
+    render_menu_block(frame, backends_area, "Backends", &BACKENDS);
```
orhun: What do you think about this cosmetic change?

```suggestion
    render_menu_block(frame, main_dishes_area, "┤Main dishes├", &MAIN_DISHES);
    render_menu_block(frame, backends_area, "┤Backends├", &BACKENDS);
```

j-g00da:
> try adding a screenshot command to the tape file to see how it looks in PNG.

This already is a png generated using screenshot command.

> I wonder if there's a bug with dim + quadrant char + rendering several times to the same cell?

Most probably, I bet dim changes opacity and we can just see some previous char. I will try to make a repro and open an issue later. For now I can think of at least two workarounds for this so it shouldn't be a blocker.

examples/apps/ratatui-release-header/src/main.rs
j-g00da: It came through my mind but I'm not sure if this helps anything or just complicates things, especially release name in description seems like a hack.

examples/apps/ratatui-release-header/src/main.rs
joshka: I don't mind the real numbers here. These would be the sizes of the logo and width of the text in the boxes. They seem reasonable as arbitrary numbers. In libraries, numbers like this often need a constant that conveys the intent. In an app, it's less so as they represent a real concrete measured values. It's 31 because that's the size you want it to be to display the items in the right place, not because 31 has some symbolic intent.

examples/apps/ratatui-release-header/src/main.rs
joshka: What about making the background color of the titles a solid color instead?

j-g00da: Not sure about this, would probably look better on bigger blocks:

image

joshka: agreed

j-g00da:
image

image

examples/apps/release-header/README.md
```diff
@@ -0,0 +1,14 @@
+
```
orhun:
```suggestion
```

orhun (approved):

joshka:
I tweaked things a bunch in [cf4cf3c](https://github.com/ratatui/ratatui/pull/2000/commits/cf4cf3c46e48ef4b1a35d097e1b3a1881c5023ea). It should still render the same, just a bit simpler code in places, various shortcuts, using layout methods, etc.

examples/apps/release-header/src/main.rs
```diff
@@ -11,14 +11,16 @@
 //! [`latest`]: https://github.com/ratatui/ratatui/tree/latest
 
 use std::io::stdout;
+use std::iter::zip;
 
-use ratatui::crossterm::execute;
 use ratatui::crossterm::terminal::{EnterAlternateScreen, LeaveAlternateScreen};
-use ratatui::layout::{Offset, Spacing};
-use ratatui::prelude::*;
```
joshka: Avoid prelude generally

examples/apps/release-header/src/main.rs
```diff
@@ -39,12 +41,7 @@ const FG_COLOR: Color = Color::Rgb(246, 214, 187); // #F6D6BB
 const BG_COLOR: Color = Color::Rgb(20, 20, 50); // #141432
 const MENU_BORDER_COLOR: Color = Color::Rgb(255, 255, 160); // #FFFFA0
 
-const GRADIENT_RED: [u8; 6] = [41, 43, 50, 68, 104, 156];
-const GRADIENT_GREEN: [u8; 6] = [24, 30, 41, 65, 105, 168];
-const GRADIENT_BLUE: [u8; 6] = [55, 57, 62, 78, 113, 166];
-const GRADIENT_AMBIENT: [u8; 6] = [17, 18, 20, 25, 40, 60];
```
joshka: Moved to `impl Rainbow`

examples/apps/release-header/src/main.rs
```diff
@@ -54,51 +51,8 @@ enum RainbowColor {
     Violet,
 }
 
-const ROYGBIV: [RainbowColor; 7] = [
```
joshka: Moved to `impl Rainbow`

examples/apps/release-header/src/main.rs
```diff
@@ -54,51 +51,8 @@ enum RainbowColor {
...
-    RainbowColor::Red,
-    RainbowColor::Orange,
-    RainbowColor::Yellow,
-    RainbowColor::Green,
-    RainbowColor::Blue,
-    RainbowColor::Indigo,
-    RainbowColor::Violet,
-];
-
-fn get_gradient_color(color: &RainbowColor, index: usize) -> Color {
```
joshka: Moved to `impl Rainbow` and simplified

examples/apps/release-header/src/main.rs
```diff
@@ -54,51 +51,8 @@ enum RainbowColor {
...
-            GRADIENT_AMBIENT[index],
-            GRADIENT_BLUE[index].max(blue_ambience),
-        ),
-    };
-    Color::Rgb(r, g, b)
-}
-
-fn main() -> Result<(), Box<dyn std::error::Error>> {
+fn main() -> color_eyre::Result<()> {
+    color_eyre::install()?;
```
joshka, line 55: Just for consistency with our other examples

examples/apps/release-header/src/main.rs
```diff
@@ -108,93 +62,97 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-    render_logo(frame, left_area);
-    render_menu(frame, right_area);
+    frame.buffer_mut().set_style(area, (FG_COLOR, BG_COLOR));
+
+    let logo_width = 29;
+    let menu_width = 23;
+    let padding = 2; // Padding between logo and menu
+    let menu_borders = 3;
+    let height = MAIN_DISHES.len() as u16 + BACKENDS.len() as u16 + menu_borders;
+    let width = logo_width + menu_width + padding;
```
joshka, line 83: I think @orhun was right on this one after all :)

examples/apps/release-header/src/main.rs
```diff
@@ -108,93 +62,97 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-    frame.render_widget(
-        Line::from(format!("v{SEMVER} \"{RELEASE_NAME}\"")).dim(),
-        version_area,
-    );
+    let area = area.inner(Margin::new(1, 0));
+    let layout = Layout::vertical(Constraint::from_lengths([6, 2, 1])).flex(Flex::End);
+    let [shadow_area, logo_area, version_area] = area.layout(&layout);
+
+    // Divide the logo into letter sections for individual coloring, then render a block for each
+    // letter with a color based on the row index.
```
joshka, line 98: Moved this into this method as when simplifying it, it became short enough.

examples/apps/release-header/src/main.rs
```diff
@@ -108,93 +62,97 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-    let logo_offset = Offset::new(0, 6);
-    frame.render_widget(Block::new().style(FG_COLOR), logo_area.offset(logo_offset));
-    frame.render_widget(RatatuiLogo::small(), logo_area.offset(logo_offset));
-
-    frame.render_widget(
-        Line::from(format!("v{SEMVER} \"{RELEASE_NAME}\"")).dim(),
-        version_area,
-    );
+    let area = area.inner(Margin::new(1, 0));
+    let layout = Layout::vertical(Constraint::from_lengths([6, 2, 1])).flex(Flex::End);
```
joshka, line 94: Incidentally, this also looks nice with only 5 shadows and moved up one line now that the extra text is added for the code name. It could be worth considering this option...

image

j-g00da:
Thanks!

PR #2002, opened 2025-07-20
draft: enable truncation for text widgets

quietpigeon:
This is a draft PR related to https://github.com/ratatui/ratatui/issues/1913.
- Implementation for (left-aligned) `Line` for now 
- Will add more implementations later when design is confirmed
- Unit tests will be added before this PR is marked ready for review


ratatui-core/src/text/line.rs
```diff
@@ -188,6 +190,9 @@ pub struct Line<'a> {
 
     /// The spans that make up this line of text.
     pub spans: Vec<Span<'a>>,
+
+    /// The overflow handling for this line of text.
+    pub overflow: Overflow<'a>,
```
joshka: One big issue that this PR immediately brings to to the fore, is that we need to lock down the public fields of types in -core, otherwise this sort of change is a breaking change.

quietpigeon: Made it private.

orhun:
Hey @quietpigeon, I think the API design for this looks quite neat :) Are you still interested in pushing this over the finish line? (Maybe we could include in 0.31 or 1.0)

PR #2003, opened 2025-07-20
docs(layout): Update documentation to point to `kasuari` solver

a-kenji:

### DIFF
--- a/ratatui-core/src/layout/layout.rs
+++ b/ratatui-core/src/layout/layout.rs
@@ -626,7 +626,7 @@ impl Layout {
             .expect("invalid number of rects")
     }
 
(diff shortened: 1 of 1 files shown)


joshka (approved):
How about linking? https://crates.io/crates/kasuari

a-kenji:
Actually a direct link to the module might be better, that way it is rendered correctly inside the local documentation `cargo doc --open`,  crates.io documentation, and editor lsp integration.

image

As an added bonus it locally will point to the `kasuari` that is actually in use.

joshka:
> As an added bonus it locally will point to the `kasuari` that is actually in use.

Sweet. Thanks for the PR

PR #2004, opened 2025-07-20
fix: include underline color in anstyle conversion

aschey:
Underline color wasn't included in the style conversion logic.


joshka (approved):
Thanks for the PR

PR #2017, opened 2025-07-28
perf: use cow in paragraph

janTatesa:
This pr sets the text field to `Cow<'a, Text<'a>>` and adds `from_cow_text` and `from_borrowed_text` methods in paragraph, and modifies examples and tests that clone a `Text` when passing it to a `Paragraph` to use the `from_borrowed_text` method

### DIFF
--- a/ratatui-widgets/src/paragraph.rs
+++ b/ratatui-widgets/src/paragraph.rs
@@ -136,7 +136,7 @@ impl<'a> Paragraph<'a> {
     /// The `text` parameter can be a [`Text`] or any type that can be converted into a [`Text`]. By
     /// default, the text is styled with [`Style::default()`], not wrapped, and aligned to the left.
     ///
-    /// If you have a reference to [`Text`] use [`Paragraph::from_borrowed_text`]
+    /// If you have [`&Text`] or [`Cow<Text>`] use [`Paragraph::from`]
     /// # Examples
     ///
     /// ```rust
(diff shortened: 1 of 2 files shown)


orhun:
What is the rationale for this change? (was it discussed anywhere?) + what is the performance impact?

janTatesa:
> What is the rationale for this change? (was it discussed anywhere?) + what is the performance impact?

It is not a big performance impact but it is just adding 2 little functions

ratatui-widgets/src/paragraph.rs
```diff
@@ -151,11 +154,47 @@ impl<'a> Paragraph<'a> {
...
+    /// ```rust
+    /// use ratatui::style::{Style, Stylize};
+    /// use ratatui::text::{Line, Text};
+    /// use ratatui::widgets::Paragraph;
+    /// let text = Text::raw("Hello, world!");
+    /// let paragraph = Paragraph::from_cow_text(Cow::Borrowed(&text));
+    /// ```
+    pub const fn from_borrowed_text(text: &'a Text<'a>) -> Self {
+        Self::from_cow_text(Cow::Borrowed(text))
+    }
```
joshka: Could this reasonably be `impl From<&Text> for Paragraph` instead?

ratatui-widgets/src/paragraph.rs
```diff
@@ -151,11 +154,47 @@ impl<'a> Paragraph<'a> {
...
+    /// # Examples
+    ///
+    /// ```rust
+    /// use ratatui::style::{Style, Stylize};
+    /// use ratatui::text::{Line, Text};
+    /// use ratatui::widgets::Paragraph;
+    /// let text = Text::raw("Hello, world!");
+    /// let paragraph = Paragraph::from_cow_text(Cow::Borrowed(&text));
+    /// ```
+    pub const fn from_cow_text(text: Cow<'a, Text<'a>>) -> Self {
```
joshka: Same here - could this reasonably be `impl<'a> From<Cow<''a, Text<'a>> for Paragraph<'a>`

joshka (changes requested):
Hey,
This really needs a bit of a stronger justification for existing. Did you run into a problem writing code that caused issues that required this? 

Making this Clone on Write seems a bit weird as the paragraph doesn't write to the Text at all, so it seems like you're not doing this for writability, more for the idea that we can have either a borrowed or owned Text. That said, in Span, we have Cow, so there's fundamentally a borrowed / owned data alreadu there in the data.

I think I'd prefer to take a few steps back and instead take a use case driven look at whether we want to use Cow more liberally in our types (or even perhaps make borrowed / owned versions of the types). I suspect this sort of thing will not hit the widgets first, but instead be mainly focused around Span/Line/Text.

So I'd advocate closing this for not until the rationale / use cases are worked out.

janTatesa:
When I was writing a PR in tui textarea where I turned placeholder into Text, the Text currently has to be always cloned. It is not much of a performance cost, but two little functions are not that much to maintain

joshka:
> When I was writing a PR in tui textarea where I turned placeholder into Text, the Text currently has to be always cloned. It is not much of a performance cost, but two little functions are not that much to maintain

It's not that it's two functions, it's whether:
- those functions are idiomatic rust (i.e. they have a similar pattern to other libraries)
- they actually are necessary (i.e. they serve a purpose that's grounded in use cases that are problematic otherwise), and
- they are the simplest approach to solving the problems that have been seen

This PR currently fails to meet most of that burden.

janTatesa:
>Making this Clone on Write seems a bit weird as the paragraph doesn't write to the Text at all, so it seems like you're not doing this for writability, more for the idea that we can have either a borrowed or owned Text.

Cow is exactly what you want if you have something that might or might not be borrowed but lives at most for some lifetime. Just because the name mentions writing doesn't mean it's the only thing this type is good for.

>That said, in Span, we have Cow, so there's fundamentally a borrowed / owned data already there in the data.

That's true, but a Text is Vec<Vec<Span>>. While the cowness of span allows the string being borrowed, we still have to own the Vec of Vecs. I don't see a reason why would we require ownership of the Vec of Vecs, when we don't need to mutate it/pass the owned value to different functions.
> those functions are idiomatic rust (i.e. they have a similar pattern to other libraries)

I think From<T> is idiomatic rust. And it's commonplace to use Cow for data you don't need an ownership to. Even ratatui itself already uses it in the mentioned Span
> they are the simplest approach to solving the problems that have been seen

It's true that just cloning the value is the simplest solution but cloning stuff unecessarilly is cosidered an antipattern especially when the alternative is to just implement From

sxyazi:
I have a niche case: I need to render a `Text` as a `Paragraph` sometimes when I need wrapping, since `Text` itself does not support that.

But I can't take ownership of the `Text`, since it's constructed by the user in Lua and I need to keep it available to the user, so I can only get a `&Text` or `&mut Text`, hence I clone `Text`s each time I render, so that `Paragraph` can consume it. 

So far, I haven't observed a significant performance impact, so it may not be worth the trouble, or possibly my `Text` content isn't that large.

@janTatesa I'm not sure about your scenario, but in my case, just exposing the fields within `Paragraph` would be enough. That way, before `Paragraph::render_ref()`, I can do `Paragraph { text: mem::take(&mut text), ... }`, and after rendering, restore it to the user with `*text = mem::take(&mut paragraph.text)`.

joshka:
Moving discussion over to https://github.com/ratatui/ratatui/discussions/2027

orhun:
Needs a final decision in #2027 - closing this in the meantime.

PR #2022, opened 2025-07-28
chore: prepare for beta release

j-g00da:


joshka (approved):

orhun (approved):

orhun:
I guess we shouldn't have done this manually and updated the versions in the release PR instead.

Now `release-plz` thinks that we have manually released the beta and that's why we don't get a new release PR:

> the local package has already a different version with respect to the registry package, so release-plz will not update it

And there is no configuration option in `release-plz.toml` to force version bumps when the local version is already ahead of crates.io.

It looks like our options are:

- Rollback to the `alpha` release in `Cargo.toml` files to trigger `release-plz` again.
- Run `release-plz release` whenever we want to release the beta version. We also need to update the changelog as well.

Thoughts? @ratatui/maintainers 

(cc @marcoieni let me know if I'm missing something here.)

joshka:
Maybe jfhci hack it just for this release. Grab release-plz locally, change it to hard code the versions to use for the various crates as 0.29 -> 0.30 etc and make a single PR that drops the alpha changes and adds the beta one. We can fix up the release-plz flow later, but a hack like this would let us go forward.

PR #2023, opened 2025-07-29
chore(gitignore): ignore `.DS_Store` file

j-g00da:


orhun:
This is too specific imo, I'd recommend to use `.git/info/exclude` for files like this.

Not a big deal though.

PR #2025, opened 2025-07-29
chore: ignore `beta` and `rc` tags in `cliff.toml`

j-g00da:


joshka (approved):

orhun:
I think we want to use `skip_tags` for `beta/alpha` releases instead.

Consider this setup:

```toml
skip_tags = "beta|alpha"
ignore_tags = "rc"
```

You can create `beta`/`alpha` and the changes won't show up in the changelog. Then we go ahead and create the `rc` release which will show in the changelog. The `rc` header will be replaced with the stable release when we create the non-rc release (because it's ignored later on).

See the CHANGELOG.md in this [example commit](https://github.com/orhun/git-cliff/commit/8b5a547b1548db02bdd3fe3eaeb860ee13386286) which was the result of running `git-cliff`.

Please refer to <https://git-cliff.org/docs/configuration/git#skip_tags> for more info:

> While skip_tags drop commits from the changelog, ignore_tags include ignored commits into the next tag.

cc @j-g00da

j-g00da:
This makes sense, thanks!

Edit: created #2026

PR #2030, opened 2025-07-31
feat: implement UnicodeWidthStr for Text/Line/Span

joshka:
You can now calculate the width of any Text/Line/Span using the UnicodeWidthStr trait instead of the width method on the type. This also makes it possible to use the width_cjk() method if needed.

---

From https://github.com/joshka/tui-widgets/pull/94#discussion_r2241178372

### DIFF
--- a/ratatui-core/src/text/text.rs
+++ b/ratatui-core/src/text/text.rs
@@ -5,6 +5,8 @@ use alloc::vec;
 use alloc::vec::Vec;
 use core::fmt;
 
+use unicode_width::UnicodeWidthStr;
+
 use crate::buffer::Buffer;
 use crate::layout::{Alignment, Rect};
 use crate::style::{Style, Styled};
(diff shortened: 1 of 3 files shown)


ratatui-core/src/text/text.rs
```diff
@@ -559,6 +561,25 @@ impl<'a> Text<'a> {
     }
 }
 
+impl UnicodeWidthStr for Text<'_> {
+    /// Returns the max width of all the lines.
```
orhun, line 565: Is this supposed to be a note? I think we should document all of these functions or none of them 🤔

orhun (approved):

ratatui-core/src/text/text.rs
joshka: By default most trait implementations are usually undocumented when they do the natural thing for the trait. This is a little different than that because there's multiple lines, so it's ambiguous whether this would return total width or width of the longest line. There's no ambiguity of line and span.

orhun: Fair

PR #2031, opened 2025-08-02
chore(changelog): allow running git-cliff without release-plz

orhun:
This is probably not optimal because we won't have access to the current `{package}` while running `git-cliff` but at least it will run _just_ fine.

While at it, I changed something in the commit parsers to avoid adding dependency updates to the changelog. So maybe rebase this PR instead of squashing.

### DIFF
--- a/cliff.toml
+++ b/cliff.toml
@@ -127,8 +127,8 @@ commit_parsers = [
   { message = "^chore\\(changelog\\)", skip = true },
   { message = "^[cC]hore", group = "<!-- 07 -->Miscellaneous Tasks" },
   { message = "^build\\(deps\\)", skip = true },
-  { message = "^build", group = "<!-- 09 -->Build" },
-  { body = ".*security", group = "<!-- 08 -->Security" },
+  { message = "^build", group = "<!-- 08 -->Build" },
+  { body = ".*security", group = "<!-- 09 -->Security" },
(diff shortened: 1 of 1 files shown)


orhun:
Rationale for this change: `release-plz` uses his own context which is different than git-cliff: https://release-plz.ieni.dev/docs/changelog/format#context that's why they are not compatible.

See https://git-cliff.org/docs/templating/context

j-g00da (approved):

cliff.toml
```diff
@@ -114,9 +126,9 @@ commit_parsers = [
   { message = "^chore\\(deps\\)", skip = true },
   { message = "^chore\\(changelog\\)", skip = true },
   { message = "^[cC]hore", group = "<!-- 07 -->Miscellaneous Tasks" },
-  { body = ".*security", group = "<!-- 08 -->Security" },
   { message = "^build\\(deps\\)", skip = true },
   { message = "^build", group = "<!-- 09 -->Build" },
+  { body = ".*security", group = "<!-- 08 -->Security" },
```
joshka:
```suggestion
  { message = "^build", group = "<!-- 08 -->Build" },
  { body = ".*security", group = "<!-- 09 -->Security" },
```

orhun:
I pushed e329a963599bfbba2734571fd9ba58dec6ef4ba7 for adding comments and explaining the change

PR #2041, opened 2025-08-05
chore(deps): use less specific versions in manifests

joshka:
The goal of this is to reduce dependabot updates to generally only Cargo.lock updates

- Use "0.x" or "x" as the version generally.
- Specify versions only in the workspace manifest
- Bump versions to latest semver compatible (and update rstest / lru deps)

### DIFF
--- a/ratatui/benches/main/rect.rs
+++ b/ratatui/benches/main/rect.rs
@@ -1,4 +1,6 @@
-use criterion::{BatchSize, Bencher, BenchmarkId, Criterion, black_box, criterion_group};
+use std::hint::black_box;
+
+use criterion::{BatchSize, Bencher, BenchmarkId, Criterion, criterion_group};
 use ratatui::layout::Rect;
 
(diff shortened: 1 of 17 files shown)


joshka:
https://github.com/ratatui/ratatui/pull/2041

j-g00da (approved):
LGTM

> Specify versions only in the workspace manifest

Big fan of this change.

ratatui/benches/main/rect.rs
```diff
@@ -1,4 +1,6 @@
-use criterion::{BatchSize, Bencher, BenchmarkId, Criterion, black_box, criterion_group};
+use std::hint::black_box;
```
j-g00da, line 1: I assume this is required after bumping criterion?

j-g00da (approved):

ratatui/benches/main/rect.rs
joshka: yeah, this gets deprecated in the most recent release as the std version is available.

PR #2043, opened 2025-08-11
test(bench): Added a benchmark for constraints

kashregister:
I've added a new benchmark for constraints, which only takes into account the time it takes to generate a full layout using a single type of constraints only. Avoided rendering here as it resulted in more inaccurate benchmarks, and i believe it should be separated nonetheless. 

image

image

### DIFF
--- a/ratatui/benches/main/constraints.rs
+++ b/ratatui/benches/main/constraints.rs
@@ -1,7 +1,7 @@
 use std::hint::black_box;
 use std::rc::Rc;
 
-use criterion::{criterion_group, Criterion};
+use criterion::{Criterion, criterion_group};
 use ratatui::layout::Constraint::{Fill, Length, Max, Min, Percentage, Ratio};
 use ratatui::layout::{Layout, Rect};
 
--- a/ratatui/benches/main/constraints.rs
+++ b/ratatui/benches/main/constraints.rs
@@ -1,59 +1,60 @@
 use std::hint::black_box;
 use std::rc::Rc;
 
-use criterion::{Criterion, criterion_group};
+use criterion::{criterion_group, Criterion};
 use ratatui::layout::Constraint::{Fill, Length, Max, Min, Percentage, Ratio};
-use ratatui::layout::{Direction, Layout, Rect};
+use ratatui::layout::{Layout, Rect};
 
 const SPLIT_BY: u16 = 10;
 
-fn generate_layout(constraint: &str, area: Rect) -> Rc<[Rect]> {
-    Layout::default()
-        .direction(Direction::Vertical)
-        .constraints({
-            let mut cs = Vec::new();
-            for _ in 0..SPLIT_BY {
-                match constraint {
-                    "Fill" => {
-                        cs.push(Fill(1));
-                    }
-                    "Length" => {
-                        cs.push(Length(area.width / SPLIT_BY));
-                    }
-                    "Max" => {
-                        cs.push(Max(area.width / SPLIT_BY));
-                    }
-                    "Min" => {
-                        cs.push(Min(area.width / SPLIT_BY));
-                    }
-                    "Percentage" => {
-                        cs.push(Percentage(100 / SPLIT_BY));
-                    }
-                    "Ratio" => {
-                        cs.push(Ratio(1, SPLIT_BY.into()));
-                    }
-                    _ => todo!(),
-                }
-            }
-            cs
-        })
-        .split(area)
-}
-
-fn constraints_render(criterion: &mut Criterion) {
-    for size in [16, 64, 255] {
+fn layout_split(criterion: &mut Criterion) {
+    for size in [16, 64, 256] {
         let mut group = criterion.benchmark_group(format!("constraints {size}x{size}"));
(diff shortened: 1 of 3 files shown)


joshka:
One problem with this benchmark is that layout calculations are all cached (LRU in memory), so it's likely that this is mostly just checking the speed of the cache. That's kinda valid though as most apps will be hitting that most of the time.

ratatui/benches/main/constraints.rs
```diff
@@ -0,0 +1,59 @@
...
+                        cs.push(Percentage(100 / SPLIT_BY));
+                    }
+                    "Ratio" => {
+                        cs.push(Ratio(1, SPLIT_BY.into()));
+                    }
+                    _ => todo!(),
+                }
+            }
+            cs
+        })
```
joshka: Generally with benchmarking the idea is to ensure that you only have code which is actually being benchmarked in the code that runs for each iteration. This part of the code is probably fairly fast, but it shouldn't be part of each layout call as it's more related to choosing what to test than it is actually running the test.

ratatui/benches/main/constraints.rs
```diff
@@ -0,0 +1,59 @@
+use std::hint::black_box;
+use std::rc::Rc;
+
+use criterion::{Criterion, criterion_group};
+use ratatui::layout::Constraint::{Fill, Length, Max, Min, Percentage, Ratio};
+use ratatui::layout::{Direction, Layout, Rect};
+
+const SPLIT_BY: u16 = 10;
+
+fn generate_layout(constraint: &str, area: Rect) -> Rc<[Rect]> {
```
joshka: As a general rule in Ratatui, we write functions in calling order (outside-in) to make it easier to read code top to bottom. Can you please move this below the other function.

ratatui/benches/main/constraints.rs
```diff
@@ -0,0 +1,59 @@
...
+                    }
+                    _ => todo!(),
+                }
+            }
+            cs
+        })
+        .split(area)
+}
+
+fn constraints_render(criterion: &mut Criterion) {
```
joshka: perhaps name this `layout_split` as that's what it's mainly testing.

ratatui/benches/main/constraints.rs
```diff
@@ -0,0 +1,59 @@
...
+    for size in [16, 64, 255] {
+        let mut group = criterion.benchmark_group(format!("constraints {size}x{size}"));
+        group.sample_size(1000);
```
joshka: What's the purpose of the custom sample size. If the default isn't good, then it's worth adding a comment about why.

ratatui/benches/main/constraints.rs
```diff
@@ -0,0 +1,59 @@
...
+                    _ => todo!(),
+                }
+            }
+            cs
+        })
+        .split(area)
+}
+
+fn constraints_render(criterion: &mut Criterion) {
+    for size in [16, 64, 255] {
```
joshka: nit: I think 256 is fine here as it's used as a u16 
```suggestion
    for size in [16, 64, 256] {
```

ratatui/benches/main/constraints.rs
```diff
@@ -0,0 +1,59 @@
...
+        let mut group = criterion.benchmark_group(format!("constraints {size}x{size}"));
+        group.sample_size(1000);
+        let area = Rect::new(0, 0, size, size);
+        for constraint in ["Fill", "Length", "Max", "Min", "Percentage", "Ratio"] {
+            group.bench_function(constraint.to_string(), |bencher| {
+                bencher.iter(|| {
+                    _ = black_box(generate_layout(constraint, area));
+                });
+            });
+        }
```
joshka: It may read simpler to do something like:
```rust
group.bench_function("Fill", |bencher| bencher.iter(layout_fill(black_box(area)));
group.bench_function("Length", |bencher| bencher.iter(layout_length(black_box(area)));


fn layout_fill(area: Rect) -> ... {
    Layout::vertical([Constraint::Fill ; SPLIT_BY]).split(area)
}

fn layout_length(area: Rect) -> {
    Layout::vertical([Constraint::Length(area.width / SPLIT_BY) ; SPLIT_BY]).split(area)
}
```
 
WDYT?

ratatui/benches/main/constraints.rs
ghost: I've based this off the line benchmark, and thought it would be nice to share the sample size. Just noticed it's the only file actually containing a custom sample size, so perhaps i can address this in a follow up PR by either making a constant and make it more explicit or by simply removing it from there too. WDYT?

ghost:
Thanks for the review, I've updated the code according to your suggestions.

ratatui/benches/main/constraints.rs
joshka: I don't know why that has the sample size set either (. There's got to be some process to working out the correct sample size, but when that happens it's reasonable to document things so that the findings can be repeated later. I'm fine with leaving it out for now too.

joshka (approved):

joshka:
Thanks for the PR

PR #2049, opened 2025-08-12
feat: add width() impl for tabs

joshka:
The purpose of this is to make it easy for apps to easily calculate the total tab width including all dividers and padding.

--

Solves https://discord.com/channels/1070692720437383208/1072907135664529508/1404572379052638370

### DIFF
--- a/ratatui-widgets/src/tabs.rs
+++ b/ratatui-widgets/src/tabs.rs
@@ -8,6 +8,7 @@ use ratatui_core::style::{Style, Styled};
 use ratatui_core::symbols;
 use ratatui_core::text::{Line, Span};
 use ratatui_core::widgets::Widget;
+use unicode_width::UnicodeWidthStr;
 
 use crate::block::{Block, BlockExt};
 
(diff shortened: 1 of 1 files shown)


ratatui-widgets/src/tabs.rs
```diff
@@ -446,6 +447,68 @@ where
...
+        let titles_width = self.titles.iter().map(Line::width_cjk).sum::<usize>();
+        let title_count = self.titles.len();
+        let divider_count = title_count.saturating_sub(1);
+        let divider_width = divider_count.saturating_mul(self.divider.width_cjk());
+        let left_padding_width = title_count.saturating_mul(self.padding_left.width_cjk());
+        let right_padding_width = title_count.saturating_mul(self.padding_right.width_cjk());
+        dbg!(titles_width)
+            + dbg!(divider_width)
+            + dbg!(left_padding_width)
+            + dbg!(right_padding_width)
```
j-g00da:
```suggestion
        let titles_width = self.titles.iter().map(Line::width_cjk).sum::<usize>();
        let title_count = self.titles.len();
        let divider_count = title_count.saturating_sub(1);
        let divider_width = divider_count.saturating_mul(self.divider.width_cjk());
        let left_padding_width = title_count.saturating_mul(self.padding_left.width_cjk());
        let right_padding_width = title_count.saturating_mul(self.padding_right.width_cjk());
        titles_width + divider_width + left_padding_width + right_padding_width
```

joshka: woops. I left this in accidentally when debugging why the expected value wasn't 16...

PR #2053, opened 2025-08-19
fix(line-gauge): pad default label to display 3 numbers

martinetd:
the default label used to have a variable width as the gauge progresses, which makes the line width itself flicker on change and is not great.

use a fixed 3 digit widths so anything from 0 to 100% fits without growing the label, keeping the line immobile.

--------------

I'm bad at screenshot but this illustrates the problem a bit; left: before / right: after this patch, you can see as top and bottom aren't aligned on the left.
line-gauge

### DIFF
--- a/ratatui-widgets/src/gauge.rs
+++ b/ratatui-widgets/src/gauge.rs
@@ -431,7 +431,7 @@ impl Widget for &LineGauge<'_> {
         }
 
         let ratio = self.ratio;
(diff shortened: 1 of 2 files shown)


j-g00da (approved):
Thanks!

I like this, it changes the default behavior but I think we can treat it as a fix. 

cc: @joshka, @orhun

joshka:
Makes sense to me.

PR #2055, opened 2025-08-20
chore: document rustfmt options

joshka:


joshka:
I often grab a copy of Ratatui's config for use in other projects and have found it nice to document why we pick certain choices. Backporting this idea.

PR #2057, opened 2025-08-23
feat: Allow add/sub modifiers to be omitted in Style serialization.

rcorre:
It's really useful that Style supports Deserialize, this allows TUI
apps to have configurable theming without much extra code.

However, deserializing a style currently fails if `add_modifier` and `sub_modifier` are
not specified. That means the following TOML config:

```toml
[theme.highlight]
fg = "white"
bg = "black"
```

Will fail to deserialize with "missing field `add_modifier`". It should be possible
to omit modifiers and have them default to "none".

### DIFF
--- a/ratatui-core/src/style.rs
+++ b/ratatui-core/src/style.rs
@@ -935,7 +935,7 @@ mod tests {
         );
     }
 
-    #[cfg(all(feature = "serde"))]
+    #[cfg(feature = "serde")]
     #[test]
     fn serialize_then_deserialize() {
         let style = Style {


joshka:
Thanks, how about adding some tests for serialization on this?

Quick related question, do you think it would it might also make sense to tag the Option fields with attributes that omit serializing them when None? (https://docs.rs/serde_with/latest/serde_with/attr.skip_serializing_none.html)

j-g00da:
We should probably audit all serializable structs for similar problems.

rcorre:
> Quick related question, do you think it would it might also make sense to tag the Option fields with attributes that omit serializing them when None?

My first instinct was "no", because my TUI has a `--dump-config` option that can be used to bootstrap a new config -- I want that to include all the keys so people see everything that can be customized.

However, I just realized TOML already omits the keys, because it has no `null` type. So I suppose this would make JSON and TOML serialization more consistent. If we do this, Modifiers should probably skip serialization if `empty`, right now it serializes as `""`.

I also discovered [apply](https://docs.rs/serde_with/latest/serde_with/attr.apply.html), which looks like it will allow consumers to apply attributes to downstream types.

rcorre:
Oh, and it looks like all those attributes are part of a separate `serde_with` crate (not part of the main `serde` umbrella) which would be a new dep for `ratatui`

ratatui-core/src/style.rs
```diff
@@ -238,15 +238,29 @@ impl fmt::Debug for Modifier {
...
     /// The foreground color.
+    #[cfg_attr(feature = "serde", serde(skip_serializing_if = "Option::is_none"))]
     pub fg: Option<Color>,
     /// The background color.
+    #[cfg_attr(feature = "serde", serde(skip_serializing_if = "Option::is_none"))]
     pub bg: Option<Color>,
     /// The underline color.
     #[cfg(feature = "underline-color")]
+    #[cfg_attr(
+        all(feature = "underline-color", feature = "serde"),
```
j-g00da: I think checking underline-color here is redundant, as the field is excluded anyway if this flag is off.

ratatui-core/src/style.rs
```diff
@@ -923,4 +937,59 @@ mod tests {
                 .remove_modifier(Modifier::DIM)
         );
     }
+
+    #[cfg(all(feature = "serde", feature = "underline-color"))]
```
j-g00da: This test won't run with underline-color disabled, so the cfg attributes inside it are completely skipped.

j-g00da (approved):
LGTM, thanks!
Can you [sign your commits](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification)?

orhun (approved):
LGTM

rcorre:
> Can you sign your commits?

Done -- they were signed, I just needed to update my GPG key in Github, they should show as verified now.

orhun:
Thanks again for the PR

rcorre:
Thanks! Not to pressure, but any idea when this might make a release on crates.io?

orhun:
Hopefully soon. We have a pretty big release in our hands and it takes time

PR #2063, opened 2025-08-26
fix: clear behavior with vs16 wide emojis

nornagon:
This fixes a bug where certain emojis like ⌨️ would sometimes be "overlaid" onto existing content from the buffer, instead of properly clearing.

[example demonstrating bug](https://gist.github.com/nornagon/11a79d7a1f2e98aa129fedb4abccc530)

This PR was generated by Codex, and validated by me:
1. Behavior of the above example code was buggy before this fix (showed overlaying "b" on top of the keyboard emoji), and fixed after.
2. The U+FE0F check is not strictly required, but I did note that emoji without this char don't exhibit the buggy behavior, even without the fix.


orhun (approved):
LGTM!

Tested and confirmed the fix. Also fixed clippy lints & added some additional comments for clarification.

PR #2071, opened 2025-08-27
docs(Terminal::get_frame): Made usage of Terminal::get_frame() clearer

Blaeriz:
Updated the Terminal::get_frame section and added use cases, including an example from Dioxus showing how get_frame() can/should be used.

Also mentioned/pointed, that most people should be fine using Terminal::draw()

Closes : https://github.com/ratatui/ratatui/issues/1200

### DIFF
--- a/ratatui-core/src/layout/layout.rs
+++ b/ratatui-core/src/layout/layout.rs
@@ -130,7 +130,7 @@ impl From<i16> for Spacing {
 ///
 /// When the layout is computed, the result is cached in a thread-local cache, so that subsequent
 /// calls with the same parameters are faster. The cache is a `LruCache`, and the size of the cache
-/// can be configured using [`Layout::init_cache()`].
+/// can be configured using `Layout::init_cache()` when the `layout-cache` feature is enabled.
 ///
 /// # Construction
 ///
@@ -160,7 +160,7 @@ impl From<i16> for Spacing {
 ///
 /// # Cache Management
 ///
-/// - [`init_cache`](Self::init_cache) - Initialize layout cache with custom size
+/// - `init_cache` - Initialize layout cache with custom size (requires the `layout-cache` feature)
 ///
 /// # Example
 ///
@@ -203,8 +203,8 @@ impl Layout {
     /// on my laptop's terminal (171+51 = 222) and doubling it for good measure and then adding a
     /// bit more to make it a round number. This gives enough entries to store a layout for every
     /// row and every column, twice over, which should be enough for most apps. For those that need
-    /// more, the cache size can be set with [`Layout::init_cache()`].
-    /// This const is unused if layout cache is disabled.
+    /// more, the cache size can be set with `Layout::init_cache()` (requires the `layout-cache`
+    /// feature). This const is unused if layout cache is disabled.
     #[cfg(feature = "layout-cache")]
     pub const DEFAULT_CACHE_SIZE: usize = 500;
 
@@ -297,7 +297,7 @@ impl Layout {
     /// that subsequent calls with the same parameters are faster. The cache is a `LruCache`, and
     /// grows until `cache_size` is reached.
     ///
-    /// By default, the cache size is [`Self::DEFAULT_CACHE_SIZE`].
+    /// By default, the cache size is `DEFAULT_CACHE_SIZE`.
     #[cfg(feature = "layout-cache")]
     pub fn init_cache(cache_size: NonZeroUsize) {
         LAYOUT_CACHE.with_borrow_mut(|cache| cache.resize(cache_size));
@@ -636,8 +636,8 @@ impl Layout {
     ///
     /// This method stores the result of the computation in a thread-local cache keyed on the layout
     /// and area, so that subsequent calls with the same parameters are faster. The cache is a
-    /// `LruCache`, and grows until [`Self::DEFAULT_CACHE_SIZE`] is reached by default, if the cache
-    /// is initialized with the [`Layout::init_cache()`] grows until the initialized cache size.
+    /// `LruCache`, and grows until `DEFAULT_CACHE_SIZE` is reached by default. If the cache is
+    /// initialized with `Layout::init_cache()`, it grows until the initialized cache size.
     ///
     /// There is a helper method that can be used to split the whole area into smaller ones based on
     /// the layout: [`Layout::areas()`]. That method is a shortcut for calling this method. It
@@ -673,8 +673,8 @@ impl Layout {
     ///
     /// This method stores the result of the computation in a thread-local cache keyed on the layout
     /// and area, so that subsequent calls with the same parameters are faster. The cache is a
-    /// `LruCache`, and grows until [`Self::DEFAULT_CACHE_SIZE`] is reached by default, if the cache
-    /// is initialized with the [`Layout::init_cache()`] grows until the initialized cache size.
+    /// `LruCache`, and grows until `DEFAULT_CACHE_SIZE` is reached by default. If the cache is
+    /// initialized with `Layout::init_cache()`, it grows until the initialized cache size.
     ///
     /// # Examples
     ///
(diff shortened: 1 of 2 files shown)


ratatui-core/src/terminal/terminal.rs
```diff
@@ -188,6 +188,21 @@ where
     }
 
     /// Get a Frame object which provides a consistent view into the terminal state for rendering.
+    ///
+    /// Note that this exists to support more advanced use cases. Most cases should be fine using
+    /// [`Terminal::draw`].
```
orhun:
```suggestion
    /// # Note
    ///
    /// This exists to support more advanced use cases. Most cases should be fine using [`Terminal::draw`].
```

ratatui-core/src/terminal/terminal.rs
```diff
@@ -188,6 +188,21 @@ where
...
+    ///
+    /// `get_frame` should be used when you need direct access to the frame buffer
+    /// outside of draw closure, for example:
+    /// - Unit testing widgets
+    /// - Buffer state inspection
+    /// - Cursor manipulation
+    /// - Multiple rendering passes/Buffer Manipulation
+    /// - Custom frame lifecycle management
+    /// - Buffer exporting
```
orhun:
```suggestion
    /// [`Terminal::get_frame`] should be used when you need direct access to the frame buffer
    /// outside of draw closure, for example:
    ///
    /// - Unit testing widgets
    /// - Buffer state inspection
    /// - Cursor manipulation
    /// - Multiple rendering passes/Buffer Manipulation
    /// - Custom frame lifecycle management
    /// - Buffer exporting
```

Also I'm not sure what "Cursor manipulation" and "Buffer exporting/manipulation" means here.

ratatui-core/src/terminal/terminal.rs
```diff
@@ -188,6 +188,21 @@ where
...
+    /// # Example
+    /// The [Dioxus CLI `output.rs`](https://github.com/DioxusLabs/dioxus/blob/ed568be4cb44be33dddbc20fb0425c83ff489522/packages/cli/src/serve/output.rs#L1072)
+    /// demonstrates how the function is used to manipulate the cursor position
```
orhun: This is overly specific and the code there might change in the future. I would prefer to have a small code example here instead.

orhun (changes requested):
Thanks for the PR! Left some comments for improvements

ratatui-core/src/terminal/terminal.rs
Blaeriz: image

The rectangle is what I'm referring to as a cursor.

By Buffer exporting, I meant if someone would want to export the whole scene on the terminal as an image (Use Cases: Games)

ratatui-core/src/terminal/terminal.rs
Blaeriz: Thought making it a bit more specific would clear all confusion that anyone could have.

If we want a small embedded example, we could add a code snippet of saving the state of the terminal as an image. Takes?

orhun: Sorry for late reply. Yeah, that's reasonable, you could get the buffer and assert it.

Blaeriz:
image

what i suggested earlier wouldve been too big of an example so i decided to go with this

ratatui-core/src/terminal/terminal.rs
```diff
@@ -188,6 +188,54 @@ where
...
+    /// - Buffer state inspection
+    /// - Cursor manipulation
+    /// - Multiple rendering passes/Buffer Manipulation
+    /// - Custom frame lifecycle management
+    /// - Buffer exporting
+    /// # Example
+    /// Getting the buffer and asserting on some cells after rendering a widget.
+    /// ```rust,ignore
+    /// #[cfg(test)]
+    /// mod tests {
```
orhun: I feel like we should place this in actual tests below.

The example code should only demonstrate a minimal snippet which uses `get_frame`.

Also the CI is failing, there might be something wrong with the validity of this code.

ratatui-core/src/terminal/terminal.rs
```diff
@@ -188,6 +188,54 @@ where
...
+    ///     #[test]
+    ///     fn test_paragraph_with_get_frame() {
+    ///         let backend = TestBackend::new(30, 5);
+    ///         let mut terminal = Terminal::new(backend).unwrap();
+    ///
+    ///         {
+    ///             let mut frame = terminal.get_frame();
+    ///             let para = Paragraph::new("Hello World!")
+    ///                 .block(Block::default().title("Demo").borders(Borders::ALL));
+    ///             frame.render_widget(para, frame.area());
```
orhun: nit:

`para` -> `paragraph`

ratatui-core/src/terminal/terminal.rs
```diff
@@ -188,6 +188,54 @@ where
...
+    /// outside of draw closure, for example:
+    ///
+    /// - Unit testing widgets
+    /// - Buffer state inspection
+    /// - Cursor manipulation
+    /// - Multiple rendering passes/Buffer Manipulation
+    /// - Custom frame lifecycle management
+    /// - Buffer exporting
+    /// # Example
+    /// Getting the buffer and asserting on some cells after rendering a widget.
```
orhun, line 224: nit/formatting:

```suggestion
    /// - Buffer exporting
    ///
    /// # Example
    ///
    /// Getting the buffer and asserting on some cells after rendering a widget.
```

Blaeriz:
Mb for late reply. was busy.  Thanks for the comments

ratatui-core/src/layout/layout.rs
```diff
@@ -203,8 +203,8 @@ impl Layout {
     /// on my laptop's terminal (171+51 = 222) and doubling it for good measure and then adding a
     /// bit more to make it a round number. This gives enough entries to store a layout for every
     /// row and every column, twice over, which should be enough for most apps. For those that need
-    /// more, the cache size can be set with [`Layout::init_cache()`].
-    /// This const is unused if layout cache is disabled.
+    /// more, the cache size can be set with `Layout::init_cache()` (requires the `layout-cache`
+    /// feature). This const is unused if layout cache is disabled.
```
orhun: I think it would be better to say "not defined" rather than "unused"

Or we can simply remove the statement: (it should be clear from the comment)

```suggestion
    /// more, the cache size can be set with `Layout::init_cache()` (requires the `layout-cache`
    /// feature).
```

ratatui-core/src/layout/layout.rs
```diff
@@ -297,7 +297,7 @@ impl Layout {
     /// that subsequent calls with the same parameters are faster. The cache is a `LruCache`, and
     /// grows until `cache_size` is reached.
     ///
-    /// By default, the cache size is [`Self::DEFAULT_CACHE_SIZE`].
+    /// By default, the cache size is `DEFAULT_CACHE_SIZE`.
```
orhun: Why are we removing the link to the type?

ratatui-core/src/terminal/terminal.rs
```diff
@@ -203,6 +203,39 @@ where
...
+    /// - Buffer state inspection
+    /// - Cursor manipulation
+    /// - Multiple rendering passes/Buffer Manipulation
+    /// - Custom frame lifecycle management
+    /// - Buffer exporting
+    ///
+    /// # Example
+    ///
+    /// Getting the buffer and asserting on some cells after rendering a widget.
+    /// ```rust,ignore
```
orhun, line 226: nit:

```suggestion
    /// Getting the buffer and asserting on some cells after rendering a widget.
    ///
    /// ```rust,ignore
```

orhun (commented):
nice, looking better!

ratatui-core/src/layout/layout.rs
Blaeriz: wait your right, I don't remember making any changes to this file..... sorry abt that

ratatui-core/src/layout/layout.rs
Blaeriz: I'm confused as to why there are changes in this file.

Blaeriz:
This took way longer than it should've because I assumed we would want a niche example in the docs.

PR #2073, opened 2025-08-29
refactor(init): disable mouse capture on cleanup

Eizeed:
## Summary

PR updates `try_restore()` function so now it also executes `DisableMouseCapture`
If mouse capture is used in the app users will have to disable it by themselves 
regardless if the program exited successfully or not. 
It means create a custom panic hook and add `execute!(stdout(), DisableMouseCapture)` near `ratatui::restore()`

## Changes

- Updated `try_restore()`
- Updated documentation for `try_restore()`
- Updated documentation for `restore()`

### DIFF
--- a/ratatui/src/init.rs
+++ b/ratatui/src/init.rs
@@ -133,6 +133,7 @@ use std::io::{self, Stdout, stdout};
 
 use ratatui_core::terminal::{Terminal, TerminalOptions};
 use ratatui_crossterm::CrosstermBackend;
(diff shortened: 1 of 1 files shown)


j-g00da:
Considering the fact that ratatui is purely an output library and leaves handling the input to the user, I don't think we should interfere here.

orhun:
Agreed. In addition to that, mouse support is not enabled as default, which means we will add a redundant ANSI escape call by disabling it on exit.

Eizeed:
Ok. I see your point. That makes sense. Should i close pr now?

PR #2075, opened 2025-09-03
feat(): changed try_draw to return nested results to improve error handling capabilities

Freyja-moth:
BREAKING_CHANGE: try_draw now returns nested results that need to be handled separately

`Terminal::try_draw` now returns `Result<Result<CompletedFrame, B::Error>, E>` where `E` can be any type. 

Previously any errors would have been converted into an io error which isn't always desired behavior not every error can be accurately represented by an io error and using `other` would require a `Box<dyn Error>` internally which can make data retrieval hard.

### DIFF
--- a/ratatui-core/src/terminal/terminal.rs
+++ b/ratatui-core/src/terminal/terminal.rs
@@ -1,3 +1,5 @@
+use core::convert::Infallible;
+
 use crate::backend::{Backend, ClearType};
 use crate::buffer::{Buffer, Cell};
 use crate::layout::{Position, Rect, Size};
@@ -327,13 +329,15 @@ where
     {
         self.try_draw(|frame| {
             render_callback(frame);
-            Ok::<(), B::Error>(())
+            Ok::<(), Infallible>(())
         })
+        // This will not panic as `Err` is `Infallible`
+        .map_or_else(Err, |result| Ok(result.unwrap()))
     }
 
     /// Tries to draw a single frame to the terminal.
     ///
-    /// Returns [`Result::Ok`] containing a [`CompletedFrame`] if successful, otherwise
+    /// Returns [`Result::Ok`] containing a [`Result<CompletedFrame, E>`] if successful, otherwise
     /// [`Result::Err`] containing the [`std::io::Error`] that caused the failure.
     ///
     /// This is the equivalent of [`Terminal::draw`] but the render callback is a function or
@@ -352,11 +356,9 @@ where
     /// - move the cursor to the last known position if it was set during the rendering closure
     /// - return a [`CompletedFrame`] with the current buffer and the area of the terminal
     ///
-    /// The render callback passed to `try_draw` can return any [`Result`] with an error type that
-    /// can be converted into an [`std::io::Error`] using the [`Into`] trait. This makes it possible
+    /// The render callback passed to `try_draw` can return any [`Result`]. This makes it possible
     /// to use the `?` operator to propagate errors that occur during rendering. If the render
-    /// callback returns an error, the error will be returned from `try_draw` as an
-    /// [`std::io::Error`] and the terminal will not be updated.
+    /// callback returns an error, it will be propagated in the inner [`Result`].
     ///
     /// The [`CompletedFrame`] returned by this method can be useful for debugging or testing
     /// purposes, but it is often not used in regular applications.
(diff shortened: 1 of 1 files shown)


orhun (commented):
Heya, thanks for the PR!

The changes look good, however I'd prefer if we discuss this first (potentially in an issue, but I guess the PR is also fine)

I personally don't like nested Results and it gives me the signal that the API is not that clean. That's why I would like to understand your use case for this change and potentially find a better solution (or a less breaking way of doing this in Ratatui). It would be nice if you could share a snippet.

With that being said, I don't see [many projects](https://github.com/search?q=ratatui+try_draw&type=code&p=1) using `try_draw`, so maybe introducing a more flexible Result type along with a gradual breaking change would be plausible in the future.

Freyja-moth:
The use case for me was being able to store generic render functions without needing to rely on `io::Error` which wouldn't allow me to retrieve data about the error easily

```rust
pub struct State {
    ...
}
impl State {
    pub fn render(&mut self, frame: &mut Frame) -> Result<(), ProgramError> {
        ...
    }
    pub fn update(&mut self) -> Result<(), ProgramError> {
        self.update();

        let res = self.terminal.try_draw(|frame| {
            self.render(frame)
        }).unwrap();

        match res {
            ...
            // Some fatal error that can't be fixed 
            Err(ProgramError::Fatal(_)) => return res
        }
    }
    
    Ok(())
}
```

The previous api would require me to check there is an error, source it and then downcast it, which is a lot less clean.

```rust
pub fn update(&mut self) -> Result<(), io::Error> {
    self.update();

    let res = self.terminal.try_draw(|frame| {
        self.render(frame)
    });

    if let Err(err) = res {
        if let Some(source) = res.source::<ProgramError>()
            && let Some(actual_error) = source.downcast_ref() {
            match err {
                ...
                ProgramError::Fatal(_) => return Err(err)
            }
        } else {
            return Err(err)
        }
   }

   Ok(())
}
```

I'm not entirely sold on the way I've implemented it, this is just the only way I could think to do it. But if there is another way I'm open to it.

orhun:
Got it, thanks for sharing the use-case. Now it's more clear.

> I'm not entirely sold on the way I've implemented it, this is just the only way I could think to do it. But if there is another way I'm open to it.

orhun:
Got it, thanks for sharing the use-case. Now it's more clear.

> I'm not entirely sold on the way I've implemented it, this is just the only way I could think to do it. But if there is another way I'm open to it.

I'm also not so sure about using nested results here. I think it would more appropriate to either:

1. Define a envelope error type to wrap the possible errors (maybe in an enum) to differentiate between backend and other errors.
2. Restructure the `CompletedFrame` into e.g. `CompletedFrameResult` and potentially return a type that will cover the possible cases.

However, looking at the boilerplate code, it seems possible to do what you want to do and it doesn't look as painful as imagined so not sure if a breaking change here would be worth it. 🤔

joshka:
I haven't looked deeply at this (yet), but my initial impression is that it's overly complex and would be annoying for users to understand. With that in mind it might be worth taking a step or two back and understanding from the user's perspective whether there are alternative ways to achieve what you're attempting.

j-g00da (changes requested):
As previously pointed out - nested results are confusing. Maybe we can flatten this out by introducing enum error that can be either `B::Error` or `E`? Just a thought, not sure if that's implementable though.

Edit: Basically what @orhun suggested in point 1.

b-guild:
Another option to consider might be a new method something like:
```rust
    pub fn draw_or_map_err<F, O, E>(
        &mut self,
        render_callback: F,
        error_callback: O,
    ) -> Result<CompletedFrame<'_>, E>
    where
        F: FnOnce(&mut Frame) -> Result<(), E>,
        O: FnOnce(B::Error) -> E,
    {
        // Autoresize - otherwise we get glitches if shrinking or potential desync between widgets
        // and the terminal (if growing), which may OOB.
        if let Err(e) = self.autoresize() {
            return Err(error_callback(e));
        }

        let mut frame = self.get_frame();

        render_callback(&mut frame)?;

        // We can't change the cursor position right away because we have to flush the frame to
        // stdout first. But we also can't keep the frame around, since it holds a &mut to
        // Buffer. Thus, we're taking the important data out of the Frame and dropping it.
        let cursor_position = frame.cursor_position;

        let mut flush = || {
            // Draw to stdout
            self.flush()?;

            match cursor_position {
                None => self.hide_cursor()?,
                Some(position) => {
                    self.show_cursor()?;
                    self.set_cursor_position(position)?;
                }
            }

            self.swap_buffers();

            // Flush
            self.backend.flush()
        };

        flush().map_err(error_callback)?;

        let completed_frame = CompletedFrame {
            buffer: &self.buffers[1 - self.current],
            area: self.last_known_area,
            count: self.frame_count,
        };

        // increment frame count before returning from draw
        self.frame_count = self.frame_count.wrapping_add(1);

        Ok(completed_frame)
    }
```
That way the users can decide what type of error they want. If they want to create an enum that divides rendering errors from backend errors, they can.

PR #2076, opened 2025-09-04
feat: portable-atomic integration for targets with no atomic types

j-g00da:


ratatui/Cargo.toml
```diff
@@ -59,6 +59,9 @@ layout-cache = ["std", "ratatui-core/layout-cache"]
 ## enables conversions from colors in the [`palette`] crate to [`Color`](crate::style::Color).
 palette = ["std", "ratatui-core/palette", "dep:palette"]
 
+## enables portable-atomic integration for targets that don't support atomic types.
+portable-atomic = ["ratatui-core/portable-atomic"]
```
orhun, line 63: Can we possibly make this a part of the `std` feature? (i.e enable `ratatui-core/portable-atomic` for `std`). What's the benefit of adding a new feature flag (if this is mostly related to the std)?

j-g00da: It's the other way around, it's only used when `std` is disabled (if you enable `std` and `portable-atomic`, `portable-atomic` will just fallback to `std` and do nothing)

There's no benefit of adding it to `std` feature flag as it will just add a redundant dependency for many users.

j-g00da: Generally there are three scenarios:

Terminal applications:
- `std` enabled, `portable-atomic` disabled

Embedded targets with atomic types, bare metal x86, etc.:
- `std` disabled `portable-atomic` disabled

Embedded targets without atomic types (e.g. single core):
- `std` disabled, `portable-atomic` enabled

If you enable both, it will just fallback to `std`.

orhun: Got it, thanks for the explanation.

Pls add that to the commit description while merging it!

PR #2101, opened 2025-09-26
docs: fix link to `widgets_block_renders` test

ognis1205:
The `CONTRIBUTING.md` referenced `tests/widgets_block.rs`, but the correct path is `ratatui/tests/widgets_block.rs`. Updated the link so that readers can navigate to the test example without 404 error.

PS:
I also noticed that the Discord invite link in the docs seems to be expired. Could you please update it to the current valid one?


orhun (approved):

orhun:
I just checked and the Discord link seems to work fine for me

ognis1205:
Thanks for looking into this!
I had a stale/invalid session in my browser, which caused the issue, but the link works correctly.

PR #2106, opened 2025-09-29
ci: fix toolchain override in CI runs

ognis1205:
### Summary
Fixes an issue where the beta CI jobs were unintentionally using the stable
toolchain from `rust-toolchain.toml` instead of the matrix-specified toolchain.
This caused clippy to run against the wrong version.

### Details
- Added `RUSTUP_TOOLCHAIN: ${{ matrix.toolchain }}` to the CI configuration

### Additional Context
- https://rust-lang.github.io/rustup/overrides.html#overrides


orhun (approved):

orhun:
Looks good, let's see if we will hit any other issues. Thanks!

Issue #2077, opened 2025-09-07
Feature-flagged methods in the backend trait cause issues with feature unification


aschey:
## Problem

The backend trait currently has two methods (`scroll_region_up`/`scroll_region_down`) that are feature-flagged. Many third-party Ratatui backends haven't implemented these methods, so if you add one of these third-party backends as a dependency while having `scrolling-regions` enabled, it will fail to compile. Additionally, since the `scrolling-regions` flag must be enabled to implement these methods, every third-party backend will either have to enable Ratatui's scrolling regions flag unconditionally or create its own flag and instruct users that the two feature flags must be kept in sync.

One use case for this is a project with multiple backends - say you want to make an app that runs both natively and in the web, so you have both the crossterm backend and [ratzilla](https://github.com/orhun/ratzilla) in your dependency tree. Since ratzilla doesn't support scrolling regions, building with `--all-features` will fail. In this case, you could get around it by using conditional dependencies based on the target architecture, but that may not always be the case.

## Solution

There are a few things we could do to avoid the need for feature-flagging the trait methods:

- Instruct backends that don't support scrolling regions to throw an error indicating the operation is unsupported - Ratatui would fall back to the [`_no_scrolling_regions`](https://github.com/ratatui/ratatui/blob/ecef506a2b9bad172e436a16bb4615205bd7936d/ratatui-core/src/terminal/terminal.rs#L614-L702) variant of the logic when encountering this error. This would require a bit of refactoring since it currently doesn't call `scroll_region_up/down` until after it has already altered some of the screen state.
- If the first option isn't feasible, have another method in the backend like `fn supports_scolling_regions(&self) -> bool` that would tell Ratatui if the scrolling region methods are supported or not.
- Potentially the cleanest option - have a default implementation in the `Backend` trait that performs the logic currently done in the `_no_scrolling_regions` methods. Backends would override this behavior if scrolling regions are supported or leave the default behavior if not. I haven't studied the implementation enough to know if this is feasible.

## Alternatives

Leave the behavior as is and force third-party backends to make their own `scrolling-regions` feature flag or accept the fact that they won't compile with `scrolling-regions` enabled.

orhun:
Hello 👋🏼 
Thanks for bringing this to our attention. I was personally not aware of this issue. Just tried `--all-features` with Ratzilla and it failed - kinda surprised that I didn't run into it before.

But yeah, I think we can take action rather than just accepting how it works right now.

[...]

PR #2121, opened 2025-10-02
feat!: break scrolling regions out into a separate trait

aschey:
See #2077 for more details on the justification for this change.

Currently, the `scrolling-regions` feature flag violates Rust's rule that features should be additive. This change removes the cursor-based implementation of `insert_before` and makes the `scroll_region_up` and `scroll_region_down` backend methods mandatory to support the this functionality. This is split out into a separate `ScrollableBackend` trait to prevent backends that don't support this functionality from having to implement this. If third-party/non-terminal backends want to support `insert_before`, they can implement `ScrollByRegion` and emulate the scrolling region behavior.

I wish there were a cleaner way to do this that could preserve the original functionality, but it's the best solution I can think of that doesn't require significant conceptual changes to the `Backend`/`Terminal` abstraction layers (alternatives that won't work are listed in the issue linked above).


ratatui-core/src/backend.rs
```diff
@@ -331,13 +361,20 @@ pub trait Backend {
...
     /// [`TestBackend::scroll_region_up`].
-    #[cfg(feature = "scrolling-regions")]
+    #[allow(unused_variables)]
     fn scroll_region_up(
         &mut self,
         region: core::ops::Range<u16>,
         line_count: u16,
-    ) -> Result<(), Self::Error>;
+    ) -> Result<(), Self::Error> {
+        Ok(())
```
aschey: It would make more sense to throw an unsupported error here by default, but we can't do that since the backend uses a custom error type now.

joshka:
@codex review. Pay particular not to whether there are some other options that could also work for this.

ratatui-core/src/backend.rs
```diff
@@ -302,6 +312,26 @@ pub trait Backend {
     /// Flush any buffered content to the terminal screen.
     fn flush(&mut self) -> Result<(), Self::Error>;
 
+    /// Get the [`ScrollingMethod`] used to scroll the terminal.
+    ///
+    /// [scrolling regions](ScrollingMethod::Regions) are used by terminals to scroll a region of
```
orhun:
```suggestion
    /// [Scrolling regions](ScrollingMethod::Regions) are used by terminals to scroll a region of
```

ratatui-core/src/backend.rs
```diff
@@ -302,6 +312,26 @@ pub trait Backend {
...
+    /// [`scroll_region_down`](Backend::scroll_region_down) methods must be implemented.
+    ///
+    /// This is only relevant for inline viewports.
+    fn get_scrolling_method(&self) -> ScrollingMethod {
+        ScrollingMethod::Cursor
+    }
+
+    /// This method is only called if [`get_scrolling_method`](Backend::get_scrolling_method)
+    /// returns [`ScrollingMethod::Regions`]. If this backend does not support scrolling regions,
+    /// `get_scrolling_method` should return [`ScrollingMethod::Cursor`] and this method should
```
orhun:
```suggestion
    /// This method is only called if [`get_scrolling_method`]
    /// returns [`ScrollingMethod::Regions`]. If this backend does not support scrolling regions,
    /// [`get_scrolling_method`] should return [`ScrollingMethod::Cursor`] and this method should
```

Then you can use a footlink like ``[`get_scrolling_method`]: Backend::get_scrolling_method``

(applies for the other backends as well)

ratatui-core/src/backend.rs
```diff
@@ -331,13 +361,20 @@ pub trait Backend {
     ///
     /// For examples of how this function is expected to work, refer to the tests for
     /// [`TestBackend::scroll_region_up`].
-    #[cfg(feature = "scrolling-regions")]
+    #[allow(unused_variables)]
```
orhun, line 340: Couldn't we could prefix them with `_` instead? (applies for the other backends as well)

ratatui-core/src/backend.rs
```diff
@@ -331,13 +361,20 @@ pub trait Backend {
...
         region: core::ops::Range<u16>,
         line_count: u16,
-    ) -> Result<(), Self::Error>;
+    ) -> Result<(), Self::Error> {
+        Ok(())
+    }
 
+    /// This method is only called if [`get_scrolling_method`](Backend::get_scrolling_method)
+    /// returns [`ScrollingMethod::Regions`]. If this backend does not support scrolling regions,
+    /// `get_scrolling_method` should return [`ScrollingMethod::Cursor`] and this method should
```
orhun: Same comment as above. We could improve the docs link here and other places

orhun (approved):
This looks pretty good to me.

I'm not certain about the side effects of this on embedded so ping @j-g00da

ratatui-core/src/backend.rs
aschey: You can, but then it shows up that way in the docs and in the autocomplete suggestions.

ratatui-core/src/backend.rs
aschey: Neat, I don't think I've seen this feature before. Thanks!

ratatui-core/src/backend.rs
aschey: I think I got all of them now.

ratatui-core/src/backend.rs
orhun: Ah okay, fair

orhun:
An additional set of eyes on this would be nice

ratatui-core/src/backend.rs
j-g00da: Both a no-op and error are not so great as this breaks interface segregation principle. Maybe these should be moved to some separate trait? Not sure.

j-g00da (changes requested):
I'm not really a fan of the current approach since we are implementing no-op methods in backend trait. It would be good to first explore option of introducing a separate trait instead so we are not forced to implement these.

ratatui-core/src/backend.rs
aschey: That could be a good idea; I hadn't thought of that. I can try it out and see if it works. It would be a much larger impact in terms of breaking changes, but this feature probably isn't used by a ton of apps.

aschey:
> I'm not really a fan of the current approach since we are implementing no-op methods in backend trait. It would be good to first explore option of introducing a separate trait instead so we are not forced to implement these.

After thinking about this a bit, there are some downsides to splitting the traits here. The main problem is that `insert_before` is part of `Terminal`, not `Backend`. This means we're limited in what backend-specific logic we can do here. The ideal solution would probably be to allow `insert_before` to use the optimized version if the backend implements this new trait (we'll call it `ScrollableBackend` for now) that provides `scroll_region_up` and `scroll_region_down`. 

```rust
impl<B> Terminal<B>
    where B: Backend
{
    pub fn insert_before<F>(&mut self, height: u16, draw_fn: F) -> Result<(), B::Error>
     where
        F: FnOnce(&mut Buffer),
    {
       // non-optimized version
    } 
}

pub trait ScrollableBackend {
    type Error: core::error::Error;

    fn scroll_region_up(
        &mut self,
        region: core::ops::Range<u16>,
        line_count: u16,
    ) -> Result<(), Self::Error>;

    fn scroll_region_down(
        &mut self,
        region: core::ops::Range<u16>,
        line_count: u16,
    ) -> Result<(), Self::Error>;
}


impl<B> Terminal<B>
    where B: Backend + ScrollableBackend
{
    pub fn insert_before<F>(&mut self, height: u16, draw_fn: F) -> Result<(), B::Error>
     where
        F: FnOnce(&mut Buffer),
    {
       // optimized version using scroll_region_up/scroll_region_down
    } 
}
```

Unfortunately, this requires specialization, so it's not possible right now.

The alternative would be to leave `insert_before` alone and create a separate method (`insert_before_scrolling_regions` or something, would need to think of a better name) that relies on the new trait. This is cleaner from an interface design perspective, but this means the implementation details leak into the public API, which feels like a step backwards from an end-user perspective.

```rust
impl<B> Terminal<B>
    where B: Backend + ScrollableBackend
{
    pub fn insert_before_scrolling_regions<F>(&mut self, height: u16, draw_fn: F) -> Result<(), B::Error>
     where
        F: FnOnce(&mut Buffer),
    {
       // optimized version using scroll_region_up/scroll_region_down
    } 
}
```

aschey:
Here's another option - we could go the route of making scrolling regions mandatory for supporting `insert_before`, taking the separate trait approach. The common backends already implement it and I'm not sure if `insert_before` has much utility for alternative/non-terminal backends. If an alternative backend that doesn't support scrolling regions wants to support this, they could use logic similar to what's in `TestBackend` to emulate it. This would be the simplest/cleanest solution at the expense of flexibility.

orhun:
Yeah, I think that sounds like a better way of handling this.

aschey:
> Yeah, I think that sounds like a better way of handling this.

Okay, I tried the approach of creating a new trait and removing the cursor-based implementation. The original change is still there in the first commit in case we decide to revert.

orhun (approved):
I think this looks fair!

j-g00da (commented):
I will hopefully do a more detailed review later today or tomorrow, but I gave it a quick glance and I really like the approach here! Not sure on the trait naming, maybe just `Scrollable`?

This probably needs expansion of the breaking changes (there's a place for a migration guide below the summaries).

I would also want @joshka to take a look for three approvals, since this modifies both `Backend` and `Terminal` - a very core of the library.

aschey:
> Not sure on the trait naming, maybe just Scrollable?

I went with `ScrollableBackend` because it's a _strict_ superset of `Backend`. i.e. you would always accept `ScrollableBackend` as an input rather than `Backend + Scrollable`. This is necessary because the error type parameter needs to be shared between the two traits. If you just accept `Scrollable` as input, the intent is less obvious. I'm fine with whatever the maintainers prefer here though.

orhun:
Yeah `ScrollableBackend` sounds good to me. `Scrollable` seems a bit too generic.

BREAKING-CHANGES.md
```diff
@@ -27,6 +27,8 @@ This is a quick summary of the sections below:
   - Disabling `default-features` suppresses the error message if `show_cursor()` fails when dropping
     `Terminal`
   - Support a broader range for `unicode-width` version
+  - The `scrolling-regions` feature flag is removed in favor of breaking this functionality out into
+    the `ScrollableBackend` trait.
```
j-g00da: Needs expansion in the detailed section below

j-g00da (commented):
Generally LGTM,

I would still argue with the name, but this may be my personal bias towards module scopes like `backend::Scrollable` over verbose trait names. If others prefer `ScrollableBackend`, I won't block it. :)

joshka:
I'd like to request that @nfachan please take a look at this, as they originally implemented the scrolling region support in https://github.com/openai/codex, and may have the most context about the effect of this on existing application code.

ratatui-core/src/terminal/terminal.rs
```diff
@@ -569,6 +569,56 @@ where
...
+            self.backend.draw(old.diff(&new).into_iter())?;
+            self.backend.flush()?;
+        }
+        Ok(remainder)
+    }
+}
+
+impl<B> Terminal<B>
+where
+    B: Backend + ScrollableBackend,
```
joshka: I think this can still work as:

```suggestion
impl<B> Terminal<B>
where
    B: ScrollableBackend,
```

(It compiles at least)

joshka (commented):
I had a superficial look at this, but want to go a bit deeper and re-understand the code that's being removed here and the historic context. Mostly I'm concerned that if we switch over 100% by removing the cursor based implementation, there are some use cases that would no longer work. Incidentally this change has some overlap with some work I'm looking at right now in my day job, so I'm pretty keen to do this part right. I'll take a much deeper look at this on the weekend and follow up on Tuesday if i don't get it done on the weekend.

ratatui-core/src/terminal/terminal.rs
j-g00da: Would make sense since `Backend` is already a supertrait

joshka:
I think a direction for naming is probably ScrollByRegion or something like that?

BREAKING-CHANGES.md
aschey: Done, let me know if you'd like me to add anything else.

ratatui-core/src/terminal/terminal.rs
aschey: Yep, that works. Done.

aschey:
> I think a direction for naming is probably ScrollByRegion or something like that?

I've renamed it to `ScrollByRegion`.

j-g00da (approved):
Great, thanks!

joshka:
I was going to write up an alternative approach, but I see you suggested it in https://github.com/ratatui/ratatui/pull/2121#issuecomment-3471342105

I think the reasonable thing there would be to just name the terminal methods specific to regions (i.e. make the app choose to opt into how they deal with regions). A few methods in the tracing_subscriber crate do something similar to this. (.e.g https://docs.rs/tracing-subscriber/latest/tracing_subscriber/layer/struct.Context.html#method.span)

So something like:
```rust
impl ScrollRegion for CrosstermBackend {
    fn down() {...}
    fn up() {...}
}
impl Terminal<B: Backend> {
    fn scroll_region_down(...) where B: ScrollRegion { ... }
    fn scroll_region_up(...) where B: ScrollRegion { ... }
```

What was the blocker with that approach?

aschey:
> So something like:
> 
> ```rust
> impl ScrollRegion for CrosstermBackend {
>     fn down() {...}
>     fn up() {...}
> }
> impl Terminal<B: Backend> {
>     fn scroll_region_down(...) where B: ScrollRegion { ... }
>     fn scroll_region_up(...) where B: ScrollRegion { ... }
> ```
> 
> What was the blocker with that approach?

The issue with this is that you now need two variants of `insert_before` that both must be public since overloading `insert_before` based on `B: ScrollRegion` requires specialization.

```rust
impl ScrollRegion for CrosstermBackend {
    fn down() {...}
    fn up() {...}
}

impl Terminal<B: Backend> {
    fn scroll_region_down(...) where B: ScrollRegion { ... }
    fn scroll_region_up(...) where B: ScrollRegion { ... }
    
    pub fn insert_before(...) {
       // implementation without using ScrollRegion
    }

    pub fn insert_before_scrolling_regions(...) where B: ScrollRegion {
       // implementation using ScrollRegion
    }
```

We can do something like that, but as I mentioned above, this feels a bit messy from a user perspective since you have to remember to use the correct `insert_before` variant.

joshka:
I suspect messy is fine for this as the behavior of these two approaches is a bit different. They do generally different things and so should be a conscious choice. Maybe I'm understanding it wrong though.

aschey:
> I suspect messy is fine for this as the behavior of these two approaches is a bit different. They do generally different things and so should be a conscious choice. Maybe I'm understanding it wrong though.

I believe the only difference (aside from reduced flickering) is that the scrolling region variant won't erase the viewport content when inserting content greater than the size of the viewport itself. I don't think that behavior would be desirable in any circumstance.

We can have two methods like `insert_before_with_cursor` and `insert_before_with_scrolling_regions` if that's preferable, but I think we should make it clear in the docs that you almost certainly don't want to use `insert_before_with_cursor` in that case.

nfachan:
I'm sorry to join the conversation so late. I had my filters misconfigured and missed this conversation.

I also apologize for creating the problem in the first place. I didn't fully understand how adding methods in a trait like this could lead to non-additive behavior with features.

The approach taken here seems good to me, as long as you all are good with breaking code that calls `insert_before` while using third-party backends (if there is any).

If I remember correctly, the reason the two implementations are so different is that there is a lot less flexibility in the non-scrolling-region version. I think the aim was to have identical behavior between the two. Though, it's been a while. I'll take a look again later tonight.

PR #2123, opened 2025-10-05
docs(rect): fix typo in the Rect::outer function comments

orhun:


ratatui-core/src/layout/rect.rs
```diff
@@ -224,7 +224,7 @@ impl Rect {
 
     /// Returns a new `Rect` outside the current one, with the given margin applied on each side.
     ///
-    /// If the margin causes the `Rect`'s bounds to outsdie the range of a `u16`, the `Rect` will
+    /// If the margin causes the `Rect`'s bounds to outside the range of a `u16`, the `Rect` will
```
j-g00da: This is still not gramatically correct though. Maybe 
> to **be** outside

or 
> causes the bounds to **exceed**

JayanAXHF: I believe the first one u suggested is correct

orhun: good catch, updated

PR #2124, opened 2025-10-05
test(layout): add visual buffer tests for Rect methods

orhun:


j-g00da (commented):
We should probably test some edge cases, like calling `intersection` when there is no intersection between the rects.

joshka (commented):
What are these tests testing?

The test names don't really give me a clue what they are testing. I'm actually not certain I can work that out from the code either as there's too many moving pieces. The problem with that is if this test fails, all I know is that it failed. I can't work out whether the test is wrong or the implementation is.

orhun:
Sorry for not providing enough context.

This PR originated from an idea while I was writing the release highlights for 0.30. I've seen we have added `Rect::outer` method but it did not contain many tests. Then I thought about adding a "visual" test with actual unicode characters and so on for better seeing the output of these functions. While I was at it, I went ahead and added same kind of tests to the other functions.

> We should probably test some edge cases, like calling intersection when there is no intersection between the rects.

Maybe yeah, but I was thinking those should be handled by the other small unit tests. (maybe this is tested already)

> What are these tests testing?

Nothing new. The only reason why these tests were added is that I wanted to visually verify the output of these functions.
I think they look cool and they don't bite.

> The test names don't really give me a clue what they are testing. 

Maybe I can use "visual" in the test names too?

Does any of this make sense?

joshka:
Yep - I have a general idea that these are testing how these things should look, but what I'm saying is that it would be nice to be able to look at the test names and understand the property / invariant that's being tested. I'll write some more ideas on the tests.

ratatui-core/src/layout/rect.rs
```diff
@@ -727,6 +728,32 @@ mod tests {
...
+    #[test]
+    fn inner_buffer() {
+        let base = Rect::new(2, 2, 10, 6);
+        let inner = base.inner(Margin::new(2, 1));
+
+        let mut buf = Buffer::filled(Rect::new(0, 0, 15, 10), Cell::default());
+        let base_buf = Buffer::filled(base, Cell::new("█"));
+        buf.merge(&base_buf);
+        let inner_buf = Buffer::filled(inner, Cell::new("░"));
+        buf.merge(&inner_buf);
```
joshka: Here you have several methods that are all being tested:
- Rect::inner()
- Buffer::filled()
- Buffer::merge()

This is too much logic / context in a test of the Rect:inner function. A failure in any of these would cause this to fail.

I'd suggest collapsing this to use Buffer::empty() and perhaps a small test widget for the purpose of these tests called `Filled` that just does the obvious and renders filled characters to the region. Call the outer rect `outer` to make it clearer.

ratatui-core/src/layout/rect.rs
```diff
@@ -727,6 +728,32 @@ mod tests {
...
+
+        let expected = Buffer::with_lines([
+            "               ",
+            "               ",
+            "  ██████████   ",
```
joshka: The characters chosen here run into each other which makes it difficult to count expected changes between cells (are there 1 or 2 characters on the left / right side). Use X and O or similar to make this easier to see.

ratatui-core/src/layout/rect.rs
```diff
@@ -906,6 +1019,39 @@ mod tests {
...
+            "  ░░░░░░░░░░        ",
+            "  ░░░░░░░░░░        ",
+            "  ░░████████        ",
+            "  ░░████████▒▒▒▒    ",
+            "  ░░████████▒▒▒▒    ",
+            "  ░░████████▒▒▒▒    ",
+            "        ▒▒▒▒▒▒▒▒    ",
+            "                    ",
+            "                    ",
+            "                    ",
```
joshka: For this test, it might be worth using a block as the area in which the item is clamped as ths will more easily show where things should be.

joshka:
@codex review

ratatui-core/src/layout/rect.rs
orhun: So I tried X/O:

```
        "               ",
        "               ",
        "  XXXXXXXXXX   ",
        "  XXOOOOOOXX   ",
        "  XXOOOOOOXX   ",
        "  XXOOOOOOXX   ",
        "  XXOOOOOOXX   ",
        "  XXXXXXXXXX   ",
        "               ",
        "               ",
    ]);
```

vs:

```
        "               ",
        "               ",
        "  ██████████   ",
        "  ██░░░░░░██   ",
        "  ██░░░░░░██   ",
        "  ██░░░░░░██   ",
        "  ██░░░░░░██   ",
        "  ██████████   ",
        "               ",
        "               ",
    ]);

```

Visually I don't quite see the diff :/

ratatui-core/src/layout/rect.rs
orhun: It's fair! I also think that these tests are doing a bit more than unit tests should do, so I extracted and moved all of them into an integration tests folder.

j-g00da (approved):

ratatui-core/src/layout/rect.rs
orhun: Merging as-is for now. Let me know if you want a follow up to this

PR #2138, opened 2025-10-14
test: add benchmark for text

ognis1205:
## Summary

 - Added new benchmark group for `Text` widget rendering.
 - Prepared test cases covering multi-width, multi-byte, and emoji text.
 - Designed for future comparison with `InlineText` benchmarks.

## Motivation

This helps measure and track rendering efficiency between different text representations (e.g., `Text` vs. `InlineText`) and ensures consistent performance improvements over time.

## See Also

 - https://github.com/ratatui/ratatui/pull/1963

### DIFF
--- a/ratatui/benches/main/text.rs
+++ b/ratatui/benches/main/text.rs
@@ -9,21 +9,37 @@ use ratatui::widgets::Widget;
 fn text(c: &mut Criterion) {
     let mut group = c.benchmark_group("text");
     for (width, height) in [
-        (1, 256),   // Heavily vertically skewed
-        (256, 1),   // Heavily horizontally skewed
-        (200, 50),  // 1080p fullscreen with medium font
-        (256, 256), // Max sized area
+        (200, 50),     // 1080p fullscreen area with medium font.
+        (1, u16::MAX), // Heavily vertically skewed area.
+        (u16::MAX, 1), // Heavily horizontally skewed area.
+        (4096, 4096),  // Max sized area for benchmarking (~sqrt(u16::MAX) * 16, ~768 MB buffer).
     ] {
         let buffer_size = Rect::new(0, 0, width, height);
+
+        // Generates sample text content scaled to the given rendering area.
+        // The number of repeated text blocks is roughly proportional to the area size.
+        //  - Small areas produce a few lines (at least 5 x 1 lines).
+        //  - Large areas produce many lines (up to ~5 x 1000 lines).
+        let make_text = |height: u16| {
+            let repeat = (height as usize / 5).clamp(1, 1000);
+            Text::from(
+                (0..repeat)
+                    .flat_map(|_| {
+                        vec![
+                            Line::from("The quick brown fox jumps over the lazy dog. Pack my box with five dozen liquor jugs."),
+                            Line::from("🦀 Rustaceans unite! 東京・İstanbul・Sydney・San Francisco・Warsaw 🌏 RustConf連携中！").bold(),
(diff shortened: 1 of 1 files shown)


ratatui/benches/main/text.rs
```diff
@@ -0,0 +1,49 @@
...
+/// Benchmark for rendering a text.
+fn text(c: &mut Criterion) {
+    let mut group = c.benchmark_group("text");
+    for (width, height) in [
+        (1, 256),   // Heavily vertically skewed
+        (256, 1),   // Heavily horizontally skewed
+        (50, 50),   // Typical rendering area
+        (100, 50),  // Vertically split screen
+        (200, 50),  // 1080p fullscreen with medium font
+        (256, 256), // Max sized area
```
joshka: I believe we can go bigger these days as we increased .area() to u32, so this should accept u16::MAX &^ 2.

ratatui/benches/main/text.rs
```diff
@@ -0,0 +1,49 @@
...
+use ratatui::widgets::Widget;
+
+/// Benchmark for rendering a text.
+fn text(c: &mut Criterion) {
+    let mut group = c.benchmark_group("text");
+    for (width, height) in [
+        (1, 256),   // Heavily vertically skewed
+        (256, 1),   // Heavily horizontally skewed
+        (50, 50),   // Typical rendering area
+        (100, 50),  // Vertically split screen
```
joshka: You might find it's worth skipping these two options for benchmark speed reasons.

joshka:
@codex review

ratatui/benches/main/text.rs
```diff
@@ -0,0 +1,49 @@
...
+        group.bench_with_input(
+            format!("render/{width}x{height}"),
+            &Text::from(vec![
+                Line::from("The quick brown fox jumps over the lazy dog. Pack my box with five dozen liquor jugs."),
+                Line::from("🦀 Rustaceans unite! 東京・İstanbul・Sydney・San Francisco・Warsaw 🌏 RustConf連携中！").bold(),
+                Line::from("naïve cafés ☕ serve résumé-ready developers 👩‍💻🧑🏾‍💻 testing text rendering engines.").green(),
+                Line::from("ゼロ幅スペース\u{200B}、結合絵文字👨‍👩‍👧‍👦、全角文字ＡＢＣ、半角abcが混在。").blue(),
+                Line::from("Emoji test: 🙂😇🤖👩🏻‍🎨🧑‍🚀 — wrapped in a 50x50 buffer for layout & clipping check.").italic(),
+            ]),
+            |b, text| render(b, text, buffer_size),
```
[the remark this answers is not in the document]
joshka: No. the first param is ID: IntoBenchmarkId, which is implemented for Strings, so this is fine.

Not sure why this doesn't show up as a reply to @.codex -github being weird again.

ognis1205:
@joshka 

I believe I have addressed all the points from the PR review.

Regarding your comment about increasing the benchmark area and your note:

> I believe we can go bigger these days as we increased .area() to u32, so this should accept u16::MAX &^ 2.

I decided to use a safe maximum size instead, because allocating a Vec<Cell> for (u16::MAX  &^ 2, u16::MAX  &^ 2) would likely cause SIGKILL. The benchmark now uses:

```rust
(4096, 4096),  // Max sized area for benchmarking (~sqrt(u16::MAX) * 16, ~768 MB buffer)
```

This allows testing a very large buffer while staying within a safe memory limit.

Besides that, I updated it so that the length of the test text is adjusted according to the benchmark area size, so please take a look at that as well. Thanks again.

orhun:
Hello @ognis1205, why this was closed?

PR #2150, opened 2025-10-25
feat(table): let Cells span multiple columns

### LINKED ISSUE #1568: feat(table): Multi-column table cells
## Problem

Table cells cannot currently span across multiple columns:

<table>
  <tr>
    <td>These</td>
    <td>Are</td>
    <td>Single</td>
    <td>Column</td>
  </tr>
  <tr>
    <td colspan="2">But these are</td>
    <td colspan="2">multi-column</td>
  </tr>
</table>

This is often useful when wanting to "group" or "nest" headers:

<table>
  <tr>
    <td>Benchmark</td>
    <td colspan="2">Rust</td>
    <td colspan="2">JavaScript</td>
  </tr>
  <tr>
    <td></td>
    <td>CPU</td>
    <td>Memory</td>
    <td>CPU</td>
    <td>Memory</td>
  </tr>
  <tr>
    <td>bench1</td>
    <td>5%</td>
    <td>5kb</td>
    <td>100%</td>
    <td>5GB</td>
  </tr>
</table>

## Solution

Add an option to the `Cell` struct and take it into account when rendering the table. This means that in a table with N columns, each row can contain anywhere between 1 and N cells.

## Additional notes

The specific example above would also require multi-row headers, which is a separate issue out of scope for this one. Multi-col cells are still useful in other contexts as well.

karkhaz:
This commit adds a 'column_span' field to table cells. The default value is 1; larger values will cause cells to span over multiple columns, being rendered over all columns plus the spaces between them.

This commit fixes #1568.

### DIFF
--- a/ratatui-widgets/src/table/cell.rs
+++ b/ratatui-widgets/src/table/cell.rs
@@ -121,17 +121,11 @@ impl<'a> Cell<'a> {
     /// This is a fluent setter method which must be chained or used as it consumes self
     ///
     /// # Example
-    ///
     /// ```rust
     /// use ratatui::widgets::{Cell, Row};
     /// let rows = vec![
-    ///     Row::new(vec![
-    ///         Cell::new("12345").column_span(2),
-    ///     ]),
-    ///     Row::new(vec![
-    ///         Cell::new("xx"),
-    ///         Cell::new("yy"),
-    ///     ]),
+    ///     Row::new(vec![Cell::new("12345").column_span(2)]),
+    ///     Row::new(vec![Cell::new("xx"), Cell::new("yy")]),
     /// ];
     /// // "12345",
     /// // "xx yy",


ratatui-widgets/src/table.rs
```diff
@@ -863,11 +863,50 @@ impl Table<'_> {
...
-                cell.render(
-                    Rect::new(row_area.x + x, row_area.y, *width, row_area.height),
-                    buf,
-                );
+            let mut column_widths_iter = columns_widths.iter();
+            let mut row_cells_iter = row.cells.iter();
+            let mut current_cell = row_cells_iter.next();
+            loop {
+                let mut final_column_widths: Option<(u16, u16)> = None;
+                match current_cell {
```
orhun: With the inner loops and multiple `match` statements this logic seems complex. Would be nice to simplify this (maybe as an separate/inner function) or at least add comments explaining the operations.

ratatui-widgets/src/table.rs
```diff
@@ -1352,6 +1391,173 @@ mod tests {
...
+            #[rustfmt::skip]
+            let expected = Buffer::with_lines([
+                "Cell2          ",
+                "Cell3 Cell4    ",
+            ]);
+            assert_eq!(buf, expected);
+        }
+
+        #[test]
+        fn render_with_colspan_2() {
```
orhun: Instead of using number suffixes (`_2` etc.), it would be more readable to update the test name to indicate what it's testing :)

ratatui-widgets/src/table/cell.rs
```diff
@@ -113,6 +115,46 @@ impl<'a> Cell<'a> {
...
+    ///         Cell::new("Cell7").colspan(1),
+    ///         Cell::new("Cell8"), // colspan(1) is the default
+    ///     ]),
+    /// ];
+    /// // "Cell1Cell1C Cell2",
+    /// // "Cell3 Cell4Cell4C",
+    /// // "Cell6 Cell7 Cell8",
+    /// ```
+    #[must_use = "method moves the value of self and returns the modified value"]
+    pub const fn colspan(mut self, colspan: u16) -> Self {
```
orhun: Someone suggested `spans` could also be used as a name for this functionality. `colspan` seems abbreviated a bit too much. But it's not a biggie.

orhun (commented):
Providing an early review :)

ratatui-widgets/src/table/cell.rs
karkhaz: I'm happy to do this, but I thought colspan allows for a future commit to add 'rowspan' if somebody needs this. colspan is also the term used in HTML/CSS.

ratatui-widgets/src/table.rs
karkhaz: Done

ratatui-widgets/src/table.rs
karkhaz: I've broken it out into a separate function, I agree that it helps readability. Added some comments too.

karkhaz:
Thank you for the quick review! I'll take this out of draft mode as the checks have all passed. I addressed your comments

karkhaz:
Converted back to draft as I have found a bug, this doesn't work correctly with selections.

ratatui-widgets/src/table/cell.rs
orhun: Oh, didn't know HTML was following that convention. It's fine to use something that people are familiar with then!

ratatui-widgets/src/table.rs
```diff
@@ -824,6 +824,48 @@ impl Table<'_> {
         }
     }
 
+    /// Return the space that a Cell should occupy, taking into account its colspans
+    ///
+    /// Returns `None` when there are no more columns for the Cell to occupy; in this case,
+    /// rendering of the rest of the row can safely be skipped.
+    /// Otherwise, returns a pair `Some((x, width))`, representing the start x-coordinate and width
+    /// of the Cell.
```
orhun: for better readability:

```suggestion
    /// Return the space that a Cell should occupy, taking into account its colspans.
    ///
    /// Returns `None` when there are no more columns for the Cell to occupy; in this case,
    /// rendering of the rest of the row can safely be skipped.
    ///
    /// Otherwise, returns a pair `Some((x, width))`, representing the start x-coordinate and width
    /// of the Cell.
```

ratatui-widgets/src/table.rs
```diff
@@ -824,6 +824,48 @@ impl Table<'_> {
...
+    fn get_widths_for_cell<'a, T>(column_widths_iterator: &mut T, cell: &Cell) -> Option<(u16, u16)>
+    where
+        T: Iterator<Item = &'a (u16, u16)>,
+    {
+        let mut ret = None;
```
orhun: I'd prefer if we define an internal type and use that instead of using a tuple here. e.g. `CellSize` that holds horizontal position (x) and total width of the `Cell`.

This would also make the code more readable (e.g. we could use a more definitive name instead of just saying `ret`)

ratatui-widgets/src/table.rs
```diff
@@ -824,6 +824,48 @@ impl Table<'_> {
...
+            }
+            match ret {
+                None => ret = Some((x, width)),
+                Some((old_x, old_width)) => {
+                    ret = Some((
+                        // Initial start of column span
+                        old_x,
+                        (
+                            // Space taken by previous columns and gaps
+                            old_width
```
orhun: There seems to be an indentation issue here.

ratatui-widgets/src/table.rs
```diff
@@ -863,11 +905,17 @@ impl Table<'_> {
...
             }
-            for ((x, width), cell) in columns_widths.iter().zip(row.cells.iter()) {
-                cell.render(
-                    Rect::new(row_area.x + x, row_area.y, *width, row_area.height),
-                    buf,
-                );
+            let mut column_widths_iter = columns_widths.iter();
+            for current_cell in &row.cells {
+                match Self::get_widths_for_cell(&mut column_widths_iter, current_cell) {
+                    None => {}
```
orhun: We could do `if let Some(_)` instead of `match` statement. (this would get rid of empty match arm here.)

ratatui-widgets/src/table/cell.rs
```diff
@@ -51,6 +51,7 @@ use ratatui_core::widgets::Widget;
 pub struct Cell<'a> {
     content: Text<'a>,
     style: Style,
+    colspan: u16,
```
j-g00da: something like `column_span` wouldn't be a too long name for this imo

ratatui-widgets/src/table.rs
```diff
@@ -824,6 +824,48 @@ impl Table<'_> {
...
+        for _ in 0..cell.get_colspan() {
+            let x: u16;
+            let width: u16;
+            match column_widths_iterator.next() {
+                None => break,
+                Some((next_x, next_width)) => {
+                    x = *next_x;
+                    width = *next_width;
+                }
+            }
```
j-g00da: nit: I find `if let Some(...) {} else {}` more readable in most cases when it comes to `Option` enum.

ratatui-widgets/src/table.rs
```diff
@@ -824,6 +824,48 @@ impl Table<'_> {
...
+            match ret {
+                None => ret = Some((x, width)),
+                Some((old_x, old_width)) => {
```
j-g00da: Same as above

j-g00da (commented):
Nice!

orhun:
Planning to make this a part of 0.30

ratatui-widgets/src/table.rs
karkhaz: Done in 4b239b78

ratatui-widgets/src/table.rs
karkhaz: Done

ratatui-widgets/src/table.rs
karkhaz: Done

ratatui-widgets/src/table/cell.rs
karkhaz: Done in 5db9118f

ratatui-widgets/src/table.rs
karkhaz: Done in a51ae685

karkhaz:
Not finished yet but I've pushed some commits addressing these comments, thank you for that. Not yet ready for full review.

orhun:
Great, thank you 🐭

karkhaz:
This is ready for review. I had to fix how this feature interacted with selections and `Table::column_spacing` being different values, hopefully that covers all the cases now. Most of the lines are tests.

I recommend viewing the final diff rather than individual commits because there's a lot of churn.

ratatui-widgets/src/table/cell.rs
```diff
@@ -113,6 +115,46 @@ impl<'a> Cell<'a> {
...
+    ///     ]),
+    ///     Row::new(vec![
+    ///         Cell::new("Cell6").column_span(1),
+    ///         Cell::new("Cell7").column_span(1),
+    ///         Cell::new("Cell8"), // column_span(1) is the default
+    ///     ]),
+    /// ];
+    /// // "Cell1Cell1C Cell2",
+    /// // "Cell3 Cell4Cell4C",
+    /// // "Cell6 Cell7 Cell8",
```
j-g00da: This example would be more readable if it used e.g. `1111`, `2222` etc.

ratatui-widgets/src/table/cell.rs
```diff
@@ -113,6 +115,46 @@ impl<'a> Cell<'a> {
...
+    #[must_use = "method moves the value of self and returns the modified value"]
+    pub const fn column_span(mut self, column_span: u16) -> Self {
+        self.column_span = column_span;
+        self
+    }
+
+    /// Return the `column_span` of this cell
+    pub const fn get_column_span(&self) -> u16 {
+        self.column_span
+    }
```
j-g00da: I think the getter is redundant, we don't provide getter for e.g. `content` either, I think it doesn't provide much value since this is not stateful.

ratatui-widgets/src/table.rs
```diff
@@ -806,31 +813,77 @@ impl Table<'_> {
...
+    /// and width of the Cell.
+    fn get_final_column_span<'a, T>(
+        column_widths_iterator: &mut T,
+        cell_column_span: u16,
+        column_spacing: u16,
+    ) -> Option<CellArea>
+    where
+        T: Iterator<Item = &'a CellArea>,
+    {
+        let mut ret: Option<CellArea> = None;
```
j-g00da:
```suggestion
        let mut cell_area: Option<CellArea> = None;
```

ratatui-widgets/src/table.rs
```diff
@@ -806,31 +813,77 @@ impl Table<'_> {
...
+            for (cell_area, cell) in column_widths.iter().zip(footer.cells.iter()) {
+                cell.render(
+                    Rect::new(area.x + cell_area.x, area.y, cell_area.width, area.height),
+                    buf,
+                );
             }
         }
     }
 
+    /// Return the area that a `Cell` should occupy, taking into account its `Cell::column_spans`.
```
j-g00da:
```suggestion
    /// Return the area that a [`Cell`] should occupy, taking into account its [`Cell::column_span`]s.
```

ratatui-widgets/src/table.rs
```diff
@@ -806,31 +813,77 @@ impl Table<'_> {
...
+    ///
+    /// Returns `None` when there are no more columns for the Cell to occupy.
+    ///
+    /// Otherwise, returns `Some(CellArea{x, width})`, representing the start x-coordinate
+    /// and width of the Cell.
+    fn get_final_column_span<'a, T>(
```
j-g00da:
```suggestion
    fn get_cell_area<'a, T>(
```

ratatui-widgets/src/table.rs
```diff
@@ -271,6 +271,13 @@ pub struct Table<'a> {
     flex: Flex,
 }
 
+/// The area taken up by a single `Cell`
+#[derive(PartialEq, Debug)]
+struct CellArea {
+    x: u16,
+    width: u16,
+}
```
j-g00da: Hmm, maybe we should already use a `Rect` for this? This would allow us to easily add cell row span later.

j-g00da (commented):
Left some nits, additionally some of the similar tests should be parametrized using rstest - this would also help with very verbose test names.

Other than that, this looks quite good already :)

ratatui-widgets/src/table.rs
j-g00da: cc: @orhun WDYT?

ratatui-widgets/src/table/cell.rs
karkhaz: done, made `column_span` public instead as `Table` needs to iterate over it

ratatui-widgets/src/table/cell.rs
karkhaz: Done

ratatui-widgets/src/table.rs
karkhaz: Done

ratatui-widgets/src/table.rs
karkhaz: Done

ratatui-widgets/src/table.rs
karkhaz: Done

ratatui-widgets/src/table.rs
orhun: This does not serve the same purpose as a `Rect` and it might be confusing if we use e.g. `y` instead of `width`.

ratatui-widgets/src/table.rs
```diff
@@ -271,6 +271,13 @@ pub struct Table<'a> {
     flex: Flex,
 }
 
+/// The area taken up by a single `Cell`
```
orhun:
```suggestion
/// The horizontal area taken up by a single `Cell`.
```

ratatui-widgets/src/table.rs
```diff
@@ -806,31 +813,77 @@ impl Table<'_> {
...
             }
         }
     }
 
+    /// Return the area that a `Cell` should occupy, taking into account its [`Cell::column_span`]s.
+    ///
+    /// Returns `None` when there are no more columns for the Cell to occupy.
+    ///
+    /// Otherwise, returns `Some(CellArea{x, width})`, representing the start x-coordinate
+    /// and width of the Cell.
```
orhun:
```suggestion
    /// Return the area that a [`Cell`] should occupy, taking into account its [`Cell::column_span`]s.
    ///
    /// Returns `None` when there are no more columns for the [`Cell`] to occupy.
    ///
    /// Otherwise, returns `Some(CellArea{x, width})`, representing the start x-coordinate
    /// and width of the [`Cell`].
```

ratatui-widgets/src/table.rs
j-g00da: But `Rect` has both coords and dimensions. We can still use `x` and `width`.

j-g00da: (And this would allow us to use its `y` and `height` for merging cells across multiple rows in the future.)

ratatui-widgets/src/table/cell.rs
```diff
@@ -51,6 +51,8 @@ use ratatui_core::widgets::Widget;
 pub struct Cell<'a> {
     content: Text<'a>,
     style: Style,
+    /// The number of columns this cell will extend over
+    pub column_span: u16,
```
j-g00da: I think this should be sufficient as this is only used internally.
```suggestion
    pub(crate) column_span: u16,
```

ratatui-widgets/src/table.rs
orhun: Not a bad idea. The internal use case for this should be documented in the function docs then.

ratatui-widgets/src/table/cell.rs
karkhaz: Done

ratatui-widgets/src/table.rs
karkhaz: This is done, I replaced `CellArea` with `Rect`. To make it easier in the future, I created a function called `create_cell_area` that currently takes two arguments, since the other two fields of `Rect` are unused. In the future, if we add support for row spans, then all callers of this function can be replaced with a normal initializer of `Rect`. I've mentioned all this in the docs.

karkhaz:
Yes, I just added this to `examples/apps/table`
```
                                                                                        █ 
   Coy Homenick   564 Lake Knolls haven                   kasandrahermiston@cassin.org  █ 
                  Jordynchester, Wyoming 34554                                          █ 
                                                                                        █ 
                                                                                        █ 
   Creola Lindgre [no address or email address is available for this person]            █ 
                                                                                        █ 
                                                                                        █ 
                                                                                        █ 
   Dax Williamson 88600 East Village fort                 rebeccamonahan@damore.org     █ 
                  Saigeside, Georgia 54030                                         
```

karkhaz:
@j-g00da I've parametrized all added tests with `rstest`

karkhaz:
[CI failure](https://github.com/ratatui/ratatui/actions/runs/19475493442/job/55734117884?pr=2150#step:5:81) appears unrelated to this PR and is [happening for main also](https://github.com/ratatui/ratatui/actions/runs/19473577493/job/55727317949#step:5:81)

orhun:
Yeah, see #2215

orhun:
Ping @joshka for review (I want to move on with this asap so that we can save @karkhaz from continuously rebasing :D)

ratatui-widgets/src/table.rs
```diff
@@ -806,31 +823,75 @@ impl Table<'_> {
         (header_area, rows_area, footer_area)
     }
 
-    fn render_header(&self, area: Rect, buf: &mut Buffer, column_widths: &[(u16, u16)]) {
+    fn render_header(&self, area: Rect, buf: &mut Buffer, column_widths: &[Rect]) {
```
joshka, line 813: Let's document the parameters here to help leave pointers for those coming behind to maintain this. What are the rects here? Same below

ratatui-widgets/src/table.rs
```diff
@@ -271,6 +271,23 @@ pub struct Table<'a> {
...
 
+/// Return a `Rect` that represents the area that a `Cell` extends over
+///
+/// Currently, `Cell`s may extend over multiple columns but not multiple rows. This function
+/// returns a `Rect` that spans over a single row and possibly multiple columns, with the `x` field
+/// representing the `Cell`'s start coordinate.
+///
+/// Future extensions that allow a `Cell` to span multiple rows can replace all calls to this
+/// function with an initializer for all four fields of `Rect`.
+const fn create_cell_area(x: u16, width: u16) -> Rect {
```
joshka: nit: let's move this below where it's used. the rationale being that if someone reads the use of this first, then they have a bit more context about why it's necessary. This seems like a small enough function that I'm not sure I see the value in making - a rect with zero height seems kind of useless - but I haven't read where it's used yet.

ratatui-widgets/src/table.rs
```diff
@@ -806,31 +823,75 @@ impl Table<'_> {
...
+    fn render_header(&self, area: Rect, buf: &mut Buffer, column_widths: &[Rect]) {
         if let Some(ref header) = self.header {
             buf.set_style(area, header.style);
-            for ((x, width), cell) in column_widths.iter().zip(header.cells.iter()) {
-                cell.render(Rect::new(area.x + x, area.y, *width, area.height), buf);
+            for (cell_area, cell) in column_widths.iter().zip(header.cells.iter()) {
+                cell.render(
+                    Rect::new(area.x + cell_area.x, area.y, cell_area.width, area.height),
+                    buf,
+                );
```
joshka: stylistic nit: when there's code that wraps poorly like this, I tend to extract a small variable to avoid the wrap:

```suggestion
                let area = Rect::new(area.x + cell_area.x, area.y, cell_area.width, area.height);
                cell.render(area, buf);
```

(and below)

ratatui-widgets/src/table.rs
```diff
@@ -806,31 +823,75 @@ impl Table<'_> {
...
     }
 
+    /// Return the area that a [`Cell`] should occupy, taking into account its
+    /// [`Cell::column_span`]s.
+    ///
+    /// Returns `None` when there are no more columns for the [`Cell`] to occupy.
+    ///
+    /// Otherwise, returns `Some(Rect{x, y = 0, width, height = 0})`, representing the start
+    /// x-coordinate and width of the [`Cell`].
+    fn get_cell_area<'a, T>(
```
joshka: Same nit comment as above - the context for why this is necessary and how it's used is found in a method lower than this. So to understand this code requires reading that then coming back here to read this. So move code that's called below the caller.

ratatui-widgets/src/table.rs
```diff
@@ -806,31 +823,75 @@ impl Table<'_> {
...
         }
     }
 
+    /// Return the area that a [`Cell`] should occupy, taking into account its
+    /// [`Cell::column_span`]s.
+    ///
+    /// Returns `None` when there are no more columns for the [`Cell`] to occupy.
+    ///
+    /// Otherwise, returns `Some(Rect{x, y = 0, width, height = 0})`, representing the start
+    /// x-coordinate and width of the [`Cell`].
```
joshka: This tells me about the intent of this code, but not the detail about how this calculation is being made. I can read the implementation to find this out, but I don't have a ground truth for what the implementation should be doing. Improve the doc comment to tell me what this means.

ratatui-widgets/src/table.rs
```diff
@@ -806,31 +823,75 @@ impl Table<'_> {
...
+                    area_so_far.width + next_width + column_spacing,
+                ));
+            } else {
+                cell_area = Some(create_cell_area(next_x, next_width));
+            }
+        }
+        cell_area
+    }
+
     fn render_rows(
```
joshka, line 843: This method is already overly complex, and this PR adds more complexity. Consider whether there's parts of it that can be extracted into some small methods that do part of the rendering. (again, as advice above, try to put these below this method)

Right now I know this renders rows, but have no context about how / what choices / what parts it does.

ratatui-widgets/src/table.rs
karkhaz: Done in [3969bb7](https://github.com/ratatui/ratatui/pull/2150/commits/3969bb72c82d42fe328738c61d2f2db16330e39b)

ratatui-widgets/src/table.rs
karkhaz: Done in [3969bb7](https://github.com/ratatui/ratatui/pull/2150/commits/3969bb72c82d42fe328738c61d2f2db16330e39b)

ratatui-widgets/src/table.rs
karkhaz: Done in [e475953](https://github.com/ratatui/ratatui/pull/2150/commits/e4759538615b1b508179acb19e63733cb6e021c1)

ratatui-widgets/src/table.rs
karkhaz: I created a helper function for my code, and also another helper function for some pre-existing code that sets the selection style, in [0c8ada3](https://github.com/ratatui/ratatui/pull/2150/commits/0c8ada3650417f189c2d49a9e007354fbf82afa7)

ratatui-widgets/src/table.rs
karkhaz: Done in [50a12b1](https://github.com/ratatui/ratatui/pull/2150/commits/50a12b1ae127d0842fa3f6657c742fa99deb71be)

karkhaz: ... and also done in [5729666](https://github.com/ratatui/ratatui/pull/2150/commits/572966681075ecce6d13f2a5358508792e00ce64), I missed the second one

ratatui-widgets/src/table.rs
karkhaz: Done in [7bb107a](https://github.com/ratatui/ratatui/pull/2150/commits/7bb107a7807f981d51101883903b8a55e1281629)

karkhaz:
Thank you for the review, I've addressed all comments.

ratatui-widgets/src/table.rs
karkhaz: >  This seems like a small enough function that I'm not sure I see the value in making - a rect with zero height seems kind of useless - but I haven't read where it's used yet.

Indeed it's a trivial function. As the comment explains, I would expect this function to be removed if a future commit adds the ability to span multiple rows as well as columns, as then all four fields of `Rect` would be used. Breaking it out into a function helps to find all instances where cell areas are initialized, so that they can be replaced if this happens.

ratatui-widgets/src/table.rs
```diff
@@ -902,6 +907,96 @@ impl Table<'_> {
...
+    /// Set the row style and render the highlight symbol if the row is selected
+    fn set_selection_style(
+        &self,
+        buf: &mut Buffer,
+        selection_width: u16,
+        is_selected: bool,
+        row_area: Rect,
+        row: &Row,
+    ) {
+        if selection_width > 0 && is_selected {
```
joshka: (nit) how about pushing this `if` into the caller, and name the method render_selection. That feels a bit more consistent with the other code (render_row_cells etc.)

ratatui-widgets/src/table.rs
```diff
@@ -902,6 +907,96 @@ impl Table<'_> {
...
+                cell_area = Some(create_cell_area(
+                    // Initial start of cell area
+                    area_so_far.x,
+                    area_so_far.width + next_width + column_spacing,
+                ));
+            } else {
+                cell_area = Some(create_cell_area(next_x, next_width));
+            }
+        }
+        cell_area
```
joshka: I think this can be maybe be simplified like:
```rust
let first = column_widths.next()?;
let width = first.width + column_widths.take(cell_column_span - 1).sum() + cell_column_span * column_spacing;
Rect::new(first.x, first.y, width, 1)
```
Am I missing something? Edit (saw that we support 0 below - needs to be added obv.)

ratatui-widgets/src/table.rs
```diff
@@ -1011,6 +1109,23 @@ fn ensure_percentages_less_than_100(widths: &[Constraint]) {
...
+/// function with an initializer for all four fields of `Rect`.
+const fn create_cell_area(x: u16, width: u16) -> Rect {
+    Rect {
+        x,
+        y: 0,
+        width,
+        height: 0,
+    }
+}
+
```
joshka: I don't see the value in this right now - it makes the code more difficult to understand for now, so calling YAGNI and suggesting to remove it until we need it. WDYT?

karkhaz: Removed this function in 88f1eb37

ratatui-widgets/src/table.rs
karkhaz: Yes, thank you. It was a little more complicated due to edge cases when we reach the last column, thankfully the many different test cases caught these. Fixed in c7538c30

ratatui-widgets/src/table.rs
karkhaz: Done in 63ab704c

karkhaz:
@joshka Thank you, I've addressed these comments. There's a failing check but it's broken in main too.

ratatui-widgets/src/table.rs
```diff
@@ -902,6 +909,84 @@ impl Table<'_> {
...
 
+    /// Render cells into the columns of a row
+    ///
+    /// Render `Cell`s from `cells` into columns consumed from `column_widths_iterator`, stopping
+    /// if either of these iterators are finished.  Each `Cell` gets rendered across
+    /// [`Cell::get_column_span`] columns plus the gaps between them, if this value is > 1.
+    fn render_row_cells<'a, T>(
+        &self,
+        buf: &mut Buffer,
+        column_widths_iterator: &mut T,
```
joshka: I think this can probably be just Vec tather than the iterator approach. I'd use an iter here if there was state that needed to be kept about where we got up to in the iterator between method calls, but it seems like this doesn't need that statefulness (unless I'm mistaken in the call pattern). This will simplify this a bit.

ratatui-widgets/src/table.rs
```diff
@@ -975,7 +1060,15 @@ impl Table<'_> {
...
             .split(columns_area);
-        rects.iter().map(|c| (c.x, c.width)).collect()
+        rects
+            .iter()
+            .map(|c| Rect {
+                x: c.x,
+                y: 0,
+                width: c.width,
+                height: 1,
+            })
```
joshka:
```suggestion
            .map(|c| Rect::new(c.x, 0, c.width, 1))
```

ratatui-widgets/src/table.rs
```diff
@@ -1676,15 +1864,63 @@ mod tests {
...
-            assert_eq!(table.get_column_widths(20, 0, 0), [(0, 4), (5, 4)]);
+            assert_eq!(
+                table.get_column_widths(20, 0, 0),
+                [
+                    Rect {
+                        x: 0,
+                        width: 4,
+                        y: 0,
+                        height: 1
+                    },
```
joshka: Long amounts of setup / tests tend to obscure the meat of the tests in a way that makes it tedious to confirm correctness.

One of the big things to do that makes tests easy to confirm as correct is minimize the vertical height of things like this in a way that makes sense:

```suggestion
                    Rect::new(0, 4, 0, 1),
```

etc.

tatus
joshka, line 1: ?

karkhaz: Oops... fixed in 18c6bbb8

ratatui-widgets/src/table.rs
karkhaz: Thanks, I hadn't noticed `new`. Fixed in d7a78736

ratatui-widgets/src/table.rs
karkhaz: Fixed in d7a78736

ratatui-widgets/src/table.rs
karkhaz: Yes, that makes the intent clearer. Fixed in dfa69821

karkhaz:
Thank you @joshka, I've addressed your latest comments.

ratatui-widgets/src/table/cell.rs
```diff
@@ -113,6 +116,41 @@ impl<'a> Cell<'a> {
...
+    ///     ]),
+    ///     Row::new(vec![
+    ///         Cell::new("66666").column_span(1),
+    ///         Cell::new("77777").column_span(1),
+    ///         Cell::new("88888"), // column_span(1) is the default
+    ///     ]),
+    /// ];
+    /// // "11111111111 22222",
+    /// // "33333 44444444444",
+    /// // "66666 77777 88888",
```
joshka: (one last small nit): this is a bit large as an example. Perhaps collapse to just two lines like so:
```suggestion
    /// let rows = vec![
    ///     Row::new(vec![
    ///         Cell::new("12345").column_span(2),
    ///     ]),
    ///     Row::new(vec![
    ///         Cell::new("xx"),
    ///         Cell::new("yy"),
    ///     ]),
    /// ];
    /// // "12345",
    /// // "xx yy",
```

joshka (approved):
Just a small last thing before we merge this.

ratatui-widgets/src/table/cell.rs
karkhaz: Done

joshka:
LGTM - thanks for making the changes!

karkhaz:
Thank you for all of your comments, they were very helpful---this is my first PR in Rust! :tada:

Issue #2149, opened 2025-10-25
Make darling_macros crate an optional dependency


alip:
## Description

Ratatui has its macros in a separate crate which is great. However ratatui core indirectly depends on darling_macros crate which breaks static linking for us.

## To Reproduce

```
cargo build -j16 --release
error: cannot produce proc-macro for `darling_macro v0.20.11` as the target `x86_64-unknown-linux-gnu` does not support these crate types
make: *** [Makefile:62: target/release/syd-tui] Error 101
$
⇒  echo $RUSTFLAGS
-g -Ctarget-cpu=native -Ctarget-feature=+crt-static
```

## Expected behavior

Static linking should work with ratatui core. Macros should be part of the macros crate.

## Screenshots

N/A

## Environment

- OS: Linux
- Terminal Emulator: urxvt
- Font: ?
- Crate version: 0.29
- Backend: termion

## Additional context

N/A

orhun:
This is not an issue about the `ratatui-macros` crate.

`darling_macro` is a dependency of the `instability` crate, which is what we use for documenting the unstable features. (It's also a workspace dependency). The repo is here: https://github.com/ratatui/instability

I'm not sure if we can link `instability` statically, but that would be preferable here I think. Are you building for MUSL? What are the crate requirements for static linking?

Maybe consider reporting this to https://github.com/TedDriggs/darling as well?

alip:
Thank you for the prompt response and great work. It's been a joy so far working with your library. This is not directly about musl. The issue, as far as I can understand, is proc-macros do not work with static linking, i.e. with `-Ctarget-feature=+crt-static`, on Linux. This is why we don't use proc-macros throughout the syd codebase and its dependencies. Is there a way you can make the instability crate optional? darling as far as I understand is a proc-macro library so reporting it there is of little use I think.

orhun:
Got it, thanks for the additional details. I think yes, we could potentially make `instability` gated behind a feature flag. Example snippet:

```rust
    #[instability::unstable(
        feature = "backend-writer",
        issue = "https://github.com/ratatui/ratatui/pull/991"
    )]
    pub const fn writer(&self) -> &W {
        &self.writer
    }
```

So this becomes:

```
    #[cfg_attr(
        feature = "use_instability",
        instability::unstable(
            feature = "backend-writer",
            issue = "https://github.com/ratatui/ratatui/pull/991"
        )
    )]
    pub const fn writer(&self) -> &W {
        &self.writer
[...]

PR #2151, opened 2025-10-27
Make instability crate optional

alip:
Fixes: #2149

Having applied this patch and added the following snippet to syd-tui's Cargo.toml:

```
[patch.crates-io]
ratatui = { git = "https://github.com/alip/ratatui.git", branch = "make-instability-optional" }
ratatui-core = { git = "https://github.com/alip/ratatui.git", branch = "make-instability-optional" }
ratatui-termion = { git = "https://github.com/alip/ratatui.git", branch = "make-instability-optional" }
ratatui-widgets = { git = "https://github.com/alip/ratatui.git", branch = "make-instability-optional" }
```

The error on darling_macros when compiling as static is now gone. I've also fixed another error about indoc crate. The indoc crate is used by tests only so I moved it to `dev.dependencies`.

There's still more work to be done to fix static linking:

```
      Adding ratatui v0.30.0-alpha.5 (https://github.com/alip/ratatui.git?branch=make-instability-optional#c4d0b752)
      Adding ratatui-core v0.1.0-alpha.6 (https://github.com/alip/ratatui.git?branch=make-instability-optional#c4d0b752)
      Adding ratatui-termion v0.1.0-alpha.5 (https://github.com/alip/ratatui.git?branch=make-instability-optional#c4d0b752)
      Adding ratatui-widgets v0.3.0-alpha.5 (https://github.com/alip/ratatui.git?branch=make-instability-optional#c4d0b752)
      Adding serde v1.0.228
      Adding serde_core v1.0.228
      Adding serde_derive v1.0.228
    Updating strum v0.26.3 -> v0.27.2
    Updating strum_macros v0.26.4 -> v0.27.2
      Adding thiserror v2.0.17
      Adding thiserror-impl v2.0.17
      Adding time v0.3.44
      Adding time-core v0.1.6
    Updating unicode-truncate v1.1.0 -> v2.0.0
error: cannot produce proc-macro for `rustversion v1.0.22` as the target `x86_64-unknown-linux-gnu` does not support these crate types
make: *** [Makefile:62: target/release/syd-tui] Error 101
```

rustversion is an indirect dependency of the `compact_str` crate which seems to be a new dependency. I plan to handle this in a separate PR as I am not quite sure how at the moment. Input would be appreciated.

### DIFF
--- a/ratatui-crossterm/Cargo.toml
+++ b/ratatui-crossterm/Cargo.toml
@@ -55,13 +55,15 @@ unstable = ["unstable-backend-writer"]
 ## Enables getting access to backends' writers.
 unstable-backend-writer = []
 
+## enables use of instability crate
+use-instability = ["dep:instability"]
 
 [dependencies]
 cfg-if = "1.0.1"
 crossterm_0_28 = { version = "0.28", package = "crossterm", optional = true }
 crossterm_0_29 = { version = "0.29", package = "crossterm", optional = true }
 document-features = { workspace = true, optional = true }
-instability.workspace = true
+instability = { workspace = true, optional = true }
 ratatui-core.workspace = true
 
 [dev-dependencies]
--- a/ratatui-crossterm/src/lib.rs
+++ b/ratatui-crossterm/src/lib.rs
@@ -180,9 +180,12 @@ where
     }
 
     /// Gets the writer.
-    #[instability::unstable(
-        feature = "backend-writer",
-        issue = "https://github.com/ratatui/ratatui/pull/991"
+    #[cfg_attr(
+        feature = "use-instability",
+        instability::unstable(
+            feature = "backend-writer",
+            issue = "https://github.com/ratatui/ratatui/pull/991"
+        )
     )]
     pub const fn writer(&self) -> &W {
         &self.writer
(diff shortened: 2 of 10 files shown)


ratatui-core/Cargo.toml
```diff
@@ -77,6 +76,7 @@ unicode-truncate = { workspace = true, default-features = false }
 unicode-width.workspace = true
 
 [dev-dependencies]
+indoc.workspace = true
```
orhun, line 79: Please provide another PR for this. I think we can merge this right away instead of making it sit here

ratatui-crossterm/Cargo.toml
```diff
@@ -55,13 +55,15 @@ unstable = ["unstable-backend-writer"]
 ## Enables getting access to backends' writers.
 unstable-backend-writer = []
 
+## enables use of instability crate
+use-instability = ["dep:instability"]
```
orhun, line 59: I initially suggested `use-instability` as the feature name, but I realized something like `instability` or `document-unstable` would be a better name for this.

orhun:
Converting it to draft until the CI errors are fixed

ratatui-crossterm/Cargo.toml
j-g00da: What about making it part of `unstable` feature instead?

orhun: I think it makes sense

joshka:
I think this seems like a lot of noise to support static linking and I don't think it's a good approach. Will comment on the issue with some more details.

joshka:
Closing as the fix for this is to use `--target` to ensure that the RUSTFLAGS are only passed to libraries compiled for the target not macros compiled for the host.

PR #2164, opened 2025-11-01
fix(demo): update the width of demo2 tape

### LINKED ISSUE #1721: Demo2 is corrupted
## Description

While looking at the docs, I realized the demo2 example does not work as expected:

It doesn't toggle the tabs and the mascot eye is shown at the wrong cell.

Then I realized we don't actually handle the tab key:

```
diff --git a/examples/apps/demo2/src/app.rs b/examples/apps/demo2/src/app.rs
index fc01e595..b541469e 100644
--- a/examples/apps/demo2/src/app.rs
+++ b/examples/apps/demo2/src/app.rs
@@ -93,7 +93,7 @@ impl App {
         match key.code {
             KeyCode::Char('q') | KeyCode::Esc => self.mode = Mode::Quit,
             KeyCode::Char('h') | KeyCode::Left => self.prev_tab(),
-            KeyCode::Char('l') | KeyCode::Right => self.next_tab(),
+            KeyCode::Char('l') | KeyCode::Right | KeyCode::Tab => self.next_tab(),
             KeyCode::Char('k') | KeyCode::Up => self.prev(),
             KeyCode::Char('j') | KeyCode::Down => self.next(),
             KeyCode::Char('d') | KeyCode::Delete => self.destroy(),

```

But still, it's corrupted (generated via `vhs`):

Image

orhun:
fixes #1721


joshka (approved):
I bet this is probably a missing default font on Linux which is causing the width to be just a bit off when it chooses a fallback font.

Approving, but we may not need this change if you check the font in use for your system for VHS matches the available font on macOS.

joshka:
https://github.com/charmbracelet/vhs/blob/main/vhs.go#L69 has the font list. Try installing the JetBrains font and ensuring it's used? (It could also be that my side doesn't have it - will check tomorrow)

j-g00da (approved):

orhun:
I have tried:

```
Set FontFamily "JetBrains Mono"
```

However that didn't make a change...

Merging this for now.

PR #2165, opened 2025-11-01
feat(chart): Overlapping charts

j-g00da:
Makes it possible to overlap braille charts on top of block charts.

### DIFF
--- a/ratatui-widgets/src/canvas.rs
+++ b/ratatui-widgets/src/canvas.rs
@@ -14,7 +14,6 @@
 //! You can also implement your own custom [`Shape`]s.
 
 use alloc::boxed::Box;
-use alloc::string::String;
 use alloc::vec;
 use alloc::vec::Vec;
 use core::fmt;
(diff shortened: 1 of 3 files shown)


orhun:
This seems alright to me, what's left there to do?

j-g00da:
Some more comments to ease maintenance, will get back to this by the end of the week.

j-g00da:
@joshka can you take a look? This is taken from #1466, I just rebased it and applied the suggestions from your review there.

ratatui-widgets/src/canvas.rs
```diff
@@ -69,11 +68,20 @@ pub struct Label<'a> {
...
-    // when rendering
-    string: String,
-    // Colors for foreground and background of each cell
-    colors: Vec<(Color, Color)>,
+    contents: Vec<LayerCell>,
+}
+
+/// A cell within a layer.
+///
+/// If a Context contains multiple layers, then the symbol, foreground, and background colors for a
```
orhun:
```suggestion
/// If a [`Context`] contains multiple layers, then the symbol, foreground, and background colors for a
```

ratatui-widgets/src/canvas.rs
```diff
@@ -847,12 +903,52 @@ where
...
     use ratatui_core::buffer::Cell;
+    use rstest::rstest;
 
     use super::*;
 
-    // helper to test the canvas checks that drawing a vertical and horizontal line
-    // results in the expected output
-    fn test_marker(marker: Marker, expected: &str) {
+    #[rstest]
+    #[case::block(Marker::Block, indoc!(
```
joshka, line 911: Nice

ratatui-widgets/src/canvas/rectangle.rs
```diff
@@ -115,7 +115,7 @@ mod tests {
             "█        █",
             "██████████",
         ]);
-        expected.set_style(buffer.area, Style::new().red());
+        expected.set_style(buffer.area, Style::new().red().on_red());
```
joshka, line 118: The only concern here is that now any text written over blocks in a canvas will now have the blocks behind it, but I think this is probably a reasonable thing. It's worth noting in the PR description / commit for the changelog though.

Something like (feel free to modify)

```
feat: Render Braille over Blocks in Charts and Canvas

This makes it possible to stack charts, and write text over block symbols in
Charts and Canvas while still showing the block symbols behind the text.
```

joshka (approved):
LGTM, nits on PR / changelog

PR #2169, opened 2025-11-01
chore(workspace): use ratatui dependency from the workspace

### LINKED ISSUE #2166: Failed to build docs.rs page for ratatui-widgets
API documentation for `ratatui-widgets` (beta) is not available: https://docs.rs/crate/ratatui-widgets/0.3.0-beta.0

```
[INFO] running `Command { std: "docker" "start" "-a" "a655db6a36aaf614ee66afa9aa4719c378e18405572fd22f9039f0d30842d7ff", kill_on_drop: false }`
[INFO] [stderr]     Scraping ratatui-widgets v0.3.0-beta.0 (/opt/rustwide/workdir)
[INFO] [stderr] error[E0433]: failed to resolve: use of unresolved module or unlinked crate `ratatui`
[INFO] [stderr]   --> examples/barchart.rs:20:5
[INFO] [stderr]    |
[INFO] [stderr] 20 | use ratatui::layout::{Constraint, Layout, Rect};
[INFO] [stderr]    |     ^^^^^^^ use of unresolved module or unlinked crate `ratatui`
[INFO] [stderr]    |
[INFO] [stderr]    = help: if you wanted to use a crate named `ratatui`, use `cargo add ratatui` to add it to your `Cargo.toml`
```

See the full logs: https://docs.rs/crate/ratatui-widgets/0.3.0-beta.0/builds/2624165

I believe this is due to [scrape-examples](https://doc.rust-lang.org/rustdoc/scraped-examples.html) feature because it tries to run the examples inside `ratatui-widgets` which depend on `ratatui` (which is not found / circular dependency?).

This works locally via `cargo +nightly doc -Zunstable-options -Zrustdoc-scrape-examples` because `ratatui` is a `dev-dependency` and it works _somehow_.

orhun:
an attempt to fix #2166

### DIFF
--- a/ratatui-widgets/Cargo.toml
+++ b/ratatui-widgets/Cargo.toml
@@ -68,7 +68,7 @@ unicode-width.workspace = true
 color-eyre.workspace = true
 crossterm.workspace = true
 pretty_assertions.workspace = true
-ratatui = { path = "../ratatui" }
+ratatui.workspace = true
 rstest.workspace = true
 
 [lints]


joshka:
I'd suggest for validating this will fix the problem, track down the command used to scrape the examples if you can. We can yolo-merge if you think that's too difficult, but I think it would be easier to confirm locally that this should fix things. (Intuitively the workspace link and the relative link seem like they'd act the exact same way).

joshka:
@Codex see if you can validate the original problem (docs can't scrape the examples), and check that this fixes the problem.

joshka:
Actually, I figure that https://github.com/rust-lang/docs.rs/blob/master/docs/build-workspaces.md#fix suggests that this will probably fix the problem.

joshka (approved):

PR #2174, opened 2025-11-02
docs: add AGENTS.md to guide coding agents

joshka:
The goal of this is to help make it easy for codex to improve the docs using these instructions as a base.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,359 @@
+# AGENT PLAYBOOK
+
+Guidance for automated coding assistants working on Ratatui. Human contributors should continue to
+follow `CONTRIBUTING.md`.
+
+## Snapshot
+
+- Workspace: confirm the correct crate(s) in `ARCHITECTURE.md` before changing code.
+- Tooling: run project tasks through `cargo xtask`; mirror CI expectations when practical.
+- Quality gate: produce documentation, tests, and summaries that keep maintainers ready for
+  changelog-friendly commits.
+
+## Operating Principles
+
+- **Understand the assignment**: clarify goals, deliverables, acceptance criteria, affected crates,
+  and target platforms before writing code.
+- **Plan and focus**: map out work, tackle one objective at a time, and split large or cross-cutting
+  changes into follow-up PRs.
+- **Keep a visible worklog**: record commands, observations, blockers, and dead ends in the
+  conversation so reviewers know what was attempted.
+- **Quality priorities (highest first)**: preserve user trust, avoid regressions, keep docs and tests
+  accurate, maintain code health, and only then optimize efficiency.
+- **Prefer minimal patch sets**: avoid mixing refactors with behavioral changes and stage risky
+  restructures separately.
+- **Escalate early**: ask for human input if work requires `unsafe`, licensing changes, sweeping
+  architecture shifts, or you cannot reproduce failing CI locally.
+
+## Implementation Standards
+
+- Run `cargo xtask format` (nightly rustfmt, grouped imports, 100 character comment width) before
+  hand-off; do not edit formatting or lint configs without approval.
+- Maintain idiomatic Ratatui style: use field init shorthand, prefer `Stylize` helpers, and keep
+  imports in std/external/local order.
+- Exclude `unsafe` code. If an optimization seems to require it, pause and flag the work.
+- Keep refactors, formatting sweeps, or mass renames separate from behavioral changes.
+- Avoid workspace configuration or dependency changes unless explicitly requested; respect MSRV and
+  license checks enforced by `cargo deny`.
+- Leave version bumps, changelog entries, and release tooling alone—automation handles them.
+- Reuse existing patterns by searching the repository (`rg`, examples, tests) before inventing new
+  abstractions.
+- Write docs and comments in US English. Run `cargo xtask typos` (typos-cli) or the typos step in
+  `cargo xtask lint` to keep spelling consistent.
+
+## Documentation Standards
+
+- Keep code and docs in lockstep: improve documentation anywhere you touch behavior, APIs, or
+  workflows.
+- Prefer short, factual prose that explains intent, invariants, and observable behavior rather than
+  translating code line by line.
+- Update repo guides (`README.md`, `ARCHITECTURE.md`, crate docs) when flows or guidance change; remove
+  stale content instead of layering caveats.
+- Maintain runnable examples and tutorials; prefer `TestBackend` assertions where visual output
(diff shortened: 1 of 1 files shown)


orhun:
I'm not sure about this. Don't we already have copilot instructions?

I'm not sure what the benefits of this would be and would like to sit on this for a while until we see more adoption in the [Rust ecosystem](https://github.com/search?q=agents.md+language%3Arust&type=code). I personally feel like it's a bit early to do this.

Thoughts? @ratatui/maintainers

joshka:
All the coding agents are ending up with agents.md as the standard (even copilot). Ideally we should remove the copilot instructions. See https://agents.md for more info.

I'm using this to get some more consistency between the docs that are on each module and item and will be doing a pass with codex applying this across -core and -widgets to make the docs even better than they are right now. The basic idea is to ratchet up on docs improvement as part of the process of changing any code. If you take a look at the diffs on the force pushes, these are extra small things I've found while actually using this to modify docs (starting with -core lib.rs in #2175).

orhun:
Yeah I see, then we should probably remove the copilot instructions (in this PR?).

I also believe there is a different aspect of this which is how we are approaching AI-generated contributions. That's where my initial concern about this PR comes from. (More people seeing this change == more people getting encouraged about using AI to do stuff like #1984, you know)

j-g00da:
> I also believe there is a different aspect of this which is how we are approaching AI-generated contributions.

I have exactly the same doubts, IMO we should discuss our exact stance on AI before moving forward with such changes. We are barely able to handle the current load when it comes to reviews and this can easily get out of hand if we make it even easier to spam low-effort PRs using AI.

AGENTS.md
```diff
@@ -0,0 +1,359 @@
...
+## Operating Principles
+
+- **Understand the assignment**: clarify goals, deliverables, acceptance criteria, affected crates,
+  and target platforms before writing code.
+- **Plan and focus**: map out work, tackle one objective at a time, and split large or cross-cutting
+  changes into follow-up PRs.
+- **Keep a visible worklog**: record commands, observations, blockers, and dead ends in the
+  conversation so reviewers know what was attempted.
+- **Quality priorities (highest first)**: preserve user trust, avoid regressions, keep docs and tests
+  accurate, maintain code health, and only then optimize efficiency.
```
workingjubilee, line 22: It seems to me "efficiency" is underdefined: all of these characteristics mentioned previously can be considered a form of "efficiency". It is very likely that the token-predictors will examine this token and have the implications from the rest drive this towards a down-weight on things related to, say, "runtime performance characteristics", but it cannot be stated enough that most models _are_ chiefly lexical and do not have abstract conceptual spaces that aren't "made of words". Well, tokens, specifically, but it would not hurt to be more brutally explicit here.

AGENTS.md
```diff
@@ -0,0 +1,359 @@
...
+- **Prefer minimal patch sets**: avoid mixing refactors with behavioral changes and stage risky
+  restructures separately.
+- **Escalate early**: ask for human input if work requires `unsafe`, licensing changes, sweeping
```
workingjubilee, line 25: This and/or the other places `unsafe` is directed to require human input probably should be more strongly worded: the model-based tooling must not add code within any `unsafe { }`, must not add code within any `unsafe extern "C" {}`, and must not add any `unsafe impl`. If any `unsafe { /* code */ }` seems to be required, it should instead be replaced with `todo!("unsafe is required here and agents must not add unsafe code")`.

I recommend this wording because it is already known that would-be contributors do not always review model-generated code and it does not seem a good use of anyone's time to review `unsafe` that has not at least received one pass of human eyes.

...Adding a `pub unsafe fn` is actually fine if `#![deny(unsafe_op_in_unsafe_fn)]` is used for a crate, which it should be because that's the default in Rust Edition 2024, because that only imposes requirements on callers. For this same reason adding an `unsafe trait` is also fine, but the `unsafe impl` is where the burden of proof lies.

AGENTS.md
```diff
@@ -0,0 +1,359 @@
...
+- Introduce noise, marketing language, or philosophical digressions.
+
+### When to Update Docs
+
+Update or add documentation when changes involve:
+
+- New or modified public APIs (modules, structs, enums, functions, traits).
+- Behavioral shifts, invariants, side effects, or error and retry logic updates.
+- Configuration or environment variables, serialization, persistence, or cross-crate boundaries.
+- Concurrency, async behavior, `unsafe` code, or performance-sensitive logic.
```
workingjubilee, line 124: Also: A model that adds documentation to an `unsafe` API should always include a `# Safety` header in the doc comment and should explicitly ask the user if all safety requirements have been listed.

AGENTS.md
joshka: I'm reasonably ok with trying something, seeing what happens and then tweak any instructions that are shown to negatively affect the outcomes. It's easy to guess wrong about how a model will interpret these instructions, and our intuition about things tends to lag behind when models update.

Overall, I suspect these instructions are probably too detailed rather than not enough. But getting information on that is really only possible by observing how the instructions affect generated code.

I suspect various models will all have different reactions to these instructions.

AGENTS.md
joshka: This is a fairly generic instruction (generated by an LLM). We forbid unsafe code in Ratatui as a general rule, so it's not really something I'm too concerned about. I suspect this will affect agents in different ways (some will stop and report some will blow through this and do nothing about it). Ideally the result will be that no unsafe code will be written.

AGENTS.md
joshka: As above - we forbid unsafe, so this is mostly a bit of boilerplate that doesn't apply

joshka:
Thanks for the review @workingjubilee.

> If review load is a concern...

Mostly this PR is about setting agentic tooling up so that it writes code that contains good documentation as a default. Doing that well means being really consistent with what's there already, but It's difficult to build up a knowledge of existing approaches well, and we don't expect that any contributor will have a good overall picture of these ideas. So I think having agentic tools be guided in improving the quality and consistency of our docs is something that adds value where we wouldn't generally be able to invest time to do so manually. Our docs are something that I'm particularly proud of, but I think they still could be so much better.

The instructions here were generated by running over the source code in a loop to capture aspects of the documentation that were clear from the existing docs, and missing from other docs. I then looked at recommendations on several wording things to accept / reject some of the things that harmonize common patterns (like choosing the same section names).

My goal with this is that every module has the same level of documentation as the best parts of Ratatui, that concepts are easy to navigate and find, and the documentation helps both end users and agentic tools to improve how they use Ratatui.

PR #2175, opened 2025-11-02
docs: improve ratatui-core lib docs

joshka:


joshka:
Testing out codex against https://github.com/ratatui/ratatui/pull/2174

PR #2188, opened 2025-11-07
fix: correct terminal width for halfwidth katakana with dakuten

g4yamanaka:
## Problem

Halfwidth katakana combined with halfwidth dakuten (ﾞ U+FF9E) or handakuten (ﾟ U+FF9F) causes layout corruption in ratatui. This occurs in two specific cases:

1. **Border titles**: Text overflows the box boundary
2. **Background colors**: Background extends beyond the intended area

### Root Cause

The issue stems from a mismatch between how `unicode-width` calculates width and how terminals actually display these characters:

- **unicode-width calculation**: Halfwidth dakuten/handakuten are reported as width 0 (due to their `Grapheme_Extend` Unicode property)
- **Terminal display**: These characters are displayed as width 1 each (inherited from legacy Shift-JIS encoding where they were separate 1-byte characters)

For example, `ｶﾞｷﾞｸﾞ` consists of 3 grapheme clusters (`ｶﾞ`, `ｷﾞ`, `ｸﾞ`):
- **Correct width**: Each cluster should be 2 columns (halfwidth katakana 1 + dakuten 1), total = 6 columns
- **Ratatui's calculation**: Each cluster calculated as 1 column (dakuten reported as 0), total = 3 columns
- **Terminal rendering**: Each cluster actually takes 2 columns, total = 6 columns
- **Result**: Buffer allocated for 3 columns but content requires 6 columns → **3 columns overflow**

### Visual Comparison

**Before fix (ratatui 0.29.0):**
before

**After fix:**
after

### Reproduction Code

```rust
use crossterm::{
    event::{self, Event, KeyCode},
    execute,
};
use ratatui::{
    Frame, Terminal, backend::CrosstermBackend, layout::{Constraint, Layout}, style::{Color, Style}, widgets::{Block, Borders, Paragraph}
};
use std::io;

fn main() -> io::Result<()> {
    enable_raw_mode()?;
    let mut stdout = io::stdout();
    execute!(stdout, EnterAlternateScreen)?;
    let backend = CrosstermBackend::new(stdout);

    let res = run_app(&mut terminal);

    disable_raw_mode()?;
    execute!(terminal.backend_mut(), LeaveAlternateScreen)?;
    terminal.show_cursor()?;

    res
}

fn run_app(terminal: &mut Terminal<CrosstermBackend<io::Stdout>>) -> io::Result<()> {
    loop {
        terminal.draw(ui)?;

            if key.code == KeyCode::Char('q') {
                return Ok(());
            }
        }
    }
}

fn ui(frame: &mut Frame) {
    let area = frame.area();

    // Create two narrow columns (10 columns each) side by side
    let columns = Layout::default()
        .direction(ratatui::layout::Direction::Horizontal)
        .constraints([
            Constraint::Length(10),
            Constraint::Length(10),
        ])
        .split(area);

    // Left column: halfwidth katakana + dakuten (buggy)
    let left_chunks = Layout::default()
        .direction(ratatui::layout::Direction::Vertical)
        .constraints([
            Constraint::Length(3),
            Constraint::Length(3),
        ])
        .split(columns[0]);

    // Right column: ASCII alphabet (reference, works correctly)
    let right_chunks = Layout::default()
        .direction(ratatui::layout::Direction::Vertical)
        .constraints([
            Constraint::Length(3),
            Constraint::Length(3),
        ])
        .split(columns[1]);

    // Case 1: Border title with "ｶﾞｷﾞ" (2 grapheme clusters)
    // Each cluster (ｶﾞ, ｷﾞ) should be width 2, total should be 4
    // But ratatui calculates as width 2 (1 per cluster)
    // Terminal displays as width 4 → overflows by 2 columns
    frame.render_widget(
        Paragraph::new("Case 1")
            .block(Block::default()
                .borders(Borders::ALL)
                .title("ｶﾞｷﾞ")),
        left_chunks[0],
    );

    // Case 2: Content with "ｶﾞｷﾞｸﾞ" (3 grapheme clusters)
    // Each cluster (ｶﾞ, ｷﾞ, ｸﾞ) should be width 2, total should be 6
    // But ratatui calculates as width 3 (1 per cluster)
    // Terminal displays as width 6 → overflows by 3 columns
    frame.render_widget(
        Paragraph::new("ｶﾞｷﾞｸﾞ")
            .style(Style::default().bg(Color::Red))
            .block(Block::default()
                .borders(Borders::ALL)
                .title("Case 2")),
        left_chunks[1],
    );

    // Reference boxes with ASCII (no overflow)
    frame.render_widget(
        Paragraph::new("AB")
            .block(Block::default()
                .borders(Borders::ALL)
                .title("AB")),
        right_chunks[0],
    );

    frame.render_widget(
        Paragraph::new("ABC")
            .style(Style::default().bg(Color::Green))
            .block(Block::default()
                .borders(Borders::ALL)
                .title("ABC")),
        right_chunks[1],
    );
}
```

Run with `ratatui = "0.29.0"` to see the overflow issue.

## Solution

Introduced a new `TerminalWidthStr` trait that correctly calculates the display width for grapheme clusters containing halfwidth dakuten/handakuten.

### Implementation Details

1. **Added `TerminalWidthStr` trait**
   - New trait in `ratatui-core/src/text/grapheme.rs`
   - Provides `.terminal_width()` method for `str` types
   - Correctly calculates display width for grapheme clusters with halfwidth dakuten/handakuten

2. **Internal implementation**
   - Internal `terminal_width(grapheme: &str) -> usize` function
   - When a grapheme cluster ends with halfwidth dakuten/handakuten (U+FF9E or U+FF9F), calculates width by summing individual character widths, treating dakuten/handakuten as width 1
   - For all other cases, uses standard `unicode-width` calculation

3. **Updated all width calculations across the codebase**
   - All width calculations now use `.terminal_width()` method:
     - `ratatui-core/src/backend/test.rs`
     - `ratatui-core/src/buffer/buffer.rs`
     - `ratatui-core/src/text/span.rs`
     - `ratatui-widgets/src/barchart/bar.rs`
     - `ratatui-widgets/src/reflow.rs`
     - `ratatui-widgets/src/scrollbar.rs`
     - `ratatui-widgets/src/tabs.rs`

## Testing

Added the following tests:

- Tests for halfwidth dakuten/handakuten alone
- Tests for valid combinations with halfwidth katakana
- Tests for non-katakana combinations
- Tests for combining dakuten/handakuten (U+3099, U+309A)
- Tests for `TerminalWidthStr` trait (ASCII, fullwidth characters, mixed text)

### API Design

The trait-based approach provides:
- More idiomatic Rust API (`.terminal_width()` method calls)
- Consistent interface with existing `UnicodeWidthStr` trait
- Encapsulation of internal implementation details

### Background

In legacy Shift-JIS encoding (JIS X 0201), halfwidth katakana and dakuten/handakuten were completely separate 1-byte characters. Terminal emulators maintain this behavior for compatibility, rendering `ｶﾞ` as two distinct character cells rather than as a combined single character.

The combining variants (U+3099 and U+309A) properly combine with preceding characters and report correct width 0, so they don't need special handling.

### Known Design Issue and Future Improvements

This PR avoids breaking changes by keeping the existing `UnicodeWidthStr` trait implementations. However, there is a design inconsistency:

**Current Issue:**
- Types like `Span`, `Line`, `Text`, `Tabs` implement the `UnicodeWidthStr` trait
- However, their `.width()` methods actually return "terminal display width" instead of "Unicode standard width"
- The trait name doesn't match the actual behavior

**Future Improvements:**
- Remove `UnicodeWidthStr` implementations or change them to return true Unicode width
- Provide a custom `TerminalWidth` trait instead

For now, we maintain existing behavior to minimize user impact.

## References

- [Ruby reline PR #832](https://github.com/ruby/reline/pull/832) - Same issue and fix approach
- [Microsoft Terminal Issue #18087](https://github.com/microsoft/terminal/issues/18087) - Terminal behavior discussion
- [Unicode L2/19-039](https://www.unicode.org/L2/L2019/19039-grapheme-break.pdf) - Grapheme break property for U+FF9E and U+FF9F


orhun:
Hey, thanks for investigating this issue and providing a fix!

Although the changes look good to me:

- I'm not sure if we go for another method of calculating widths (`TerminalWidthStr`). Right now most of the unicode width calculation is done by `unicode_width` crate and I'm afraid we might introduce inconsistency if we change/wrap that behavior.
- If this is an issue with width calculation and `unicode_width` is the root cause, wouldn't it be better to fix it there instead?
- What terminal did you use for testing this? For me Alacritty worked but Wezterm fails. Also it would be better to test against `0.30-beta` instead of `0.29` release.

See #1764 for a similar issue. Maybe this fix might apply there as well.

joshka:
If this is an issue with width calculation and `unicode-width` is the root cause, I think the crate is actually *doing the right thing* per the Unicode spec.

These characters (U+FF9E, U+FF9F) have the `Grapheme_Extend` property ([Unicode 17 §3 D59](https://www.unicode.org/versions/Unicode17.0.0/core-spec/chapter-3/#G41165)), and `unicode-width` correctly reports them as width 0. Terminals, however, intentionally render them as width 1 for Shift-JIS compatibility — so the mismatch lies between Unicode semantics and terminal conventions, not a bug in `unicode-width`.

Given that, I think Ratatui should explicitly commit to measuring **terminal display width**, not Unicode width. In other words, we should elevate `TerminalWidthStr` as the canonical abstraction for layout measurements and remove `UnicodeWidthStr` from our public API entirely.

To make this transition straightforward and explicit:

- Drop `unicode-width` from all public interfaces  
- Rename `terminal_width()` → `width()`  
- Remove existing `.width()` methods so downstream crates must update their imports from `UnicodeWidthStr` to `TerminalWidthStr`  

That way, the rename signals the intentional semantic shift, and the required import change keeps the migration mechanical and obvious.

Since `UnicodeWidthStr` implementations on `Span`/`Line`/`Text` only exist in the current beta, this would be a breaking change only for pre-release users — so we could do this before cutting 0.30.0 and publish a new beta. We don't need to mark this as a breaking change

---

**TL;DR**

- `unicode-width` is correct per Unicode, but terminals differ  
- Ratatui should standardize on *terminal width*, not *Unicode width*  
- Make `TerminalWidthStr` the golden path, rename `.terminal_width()` to `.width()`, and remove `UnicodeWidthStr` from the public surface  

How does that sound to you, @orhun?

---

*(As an aside, I recently worked on a [similar half-width character issue in Codex](https://github.com/openai/codex/pull/4359), which addressed this from the input side. That PR may not map directly here, but it shows the same class of discrepancies between input semantics and terminal rendering. @genki might find this interesting, though no obligation to weigh in.)*

ratatui-core/src/text/grapheme.rs
```diff
@@ -47,6 +55,99 @@ impl Styled for StyledGrapheme<'_> {
...
+/// # Examples
+///
+/// ```
+/// use ratatui_core::text::grapheme::TerminalWidthStr;
+///
+/// assert_eq!("hello".terminal_width(), 5);
+/// assert_eq!("ｶﾞｷﾞｸﾞ".terminal_width(), 6); // 3 graphemes × 2 width each
+/// assert_eq!("あいう".terminal_width(), 6); // 3 graphemes × 2 width each
+/// ```
+pub trait TerminalWidthStr {
```
joshka: Small nit: I think this and the above function belongs in a new file that's just about the width. Name it `width.rs`
That keeps the testing small and focused around this one aspect of things. There's not a lot of overlap on the grapheme code here.

joshka (changes requested):
This is a really good PR. Well tested, well explained. Just generally great.
See the wider notes in the PR comment I just made

TLDR
- Drop unicode-width from all public interfaces
- Rename terminal_width() → width()
- Remove existing .width() methods so downstream crates must update their imports from UnicodeWidthStr to TerminalWidthStr

orhun:
Yup, the proposed solution from Josh sounds great to me.

g4yamanaka:
@joshka @orhun

  Thank you for the detailed explanation. Before implementing, I'd like to confirm my understanding of the proposed changes.

  **Understanding confirmation:**

  The proposed changes:
  1. Implement `impl TerminalWidthStr` for `Span`/`Line`/`Text`
  2. Remove `pub fn width(&self) -> usize` methods from these types
  3. Downstream crates must explicitly `use ratatui_core::text::TerminalWidthStr`

  This way, the breaking change is detectable at compile time, and the migration is mechanical and obvious.

  Is this understanding correct?

  **Question about width_cjk():**

  Removing `impl UnicodeWidthStr` will also remove the `width_cjk()` methods from `Span`/`Line`/`Text`.

  Currently:
  - `Span`/`Line`/`Text` provide `width_cjk()` methods (via `impl UnicodeWidthStr`)
  - `tabs.rs` uses these methods in its own `width_cjk()` implementation
  - The `tabs.rs` comment states "This is probably the wrong method to use in most contexts"

  Should we:
  - Remove `width_cjk()` entirely from all types?
  - Add `width_cjk()` to `TerminalWidthStr` trait?
  - Take a different approach?

  Please let me know your thoughts.

joshka:
Hrm, actually I'm not sure now that you put it that way. Seems I forgot that there were width methods on the actual items. I think perhaps implementing UnicodeWidthStr was the wrong approach. Taking a look at what's there in 0.29 (https://docs.rs/ratatui/0.29.0/ratatui/index.html?search=width), maybe we could get away with instead using only the non-trait methods. WDYT?

g4yamanaka:
Background: I'm not an expert in this area, but I've investigated based on the review comments. Below are my proposals, though I'll defer to the maintainers' judgment
  on the final direction.

  Proposal 1: Remove UnicodeWidthStr
  - Remove impl UnicodeWidthStr for Span/Line/Text
  - Keep Span::width() / Line::width() / Text::width() methods as inherent methods
  - Don't migrate to trait

  Rationale:
  - Users no longer need to import external crate traits
  - Inherent methods are simpler and more intuitive

  Proposal 2: Remove width_cjk()
  - I propose removing the width_cjk() methods as well

  Context: What is width_cjk()?
  - width() = Calculates Ambiguous characters (box-drawing │, Greek α, etc.) as width 1
  - width_cjk() = Calculates Ambiguous characters as width 2
  - Ambiguous = Unicode East Asian Width property characters whose width "depends on context"
  - All box-drawing characters used by ratatui (│ U+2502, etc.) are Ambiguous
  - Complete list of Ambiguous characters: https://raw.githubusercontent.com/hamano/locale-eaw/master/test/amb.txt
  - Even in CJK locales (ja_JP.UTF-8, etc.), most terminal defaults treat Ambiguous as width 1
  - Requires explicit configuration to use width 2 (e.g., wezterm's treat_east_asian_ambiguous_width_as_wide = true)

  Reasons for removal (my opinion):

  1. Misleading expectations: The existence of width_cjk() might create false expectations that "ratatui supports CJK environments"
  2. Design limitation: ratatui calculates box-drawing widths as 1 in Block::inner(). Providing only width_cjk() won't work without layout system support
  3. Partial support concerns: Full Ambiguous=Wide support would require design changes across the entire system (Block::inner(), all widgets, rendering)

  Actual behavior verification:
  I tested ratatui's block example with wezterm:

  - Ambiguous=Narrow (default): 
    - Box-drawing characters display as width 1, layout works correctly
    - narrow
  - Ambiguous=Wide (treat_east_asian_ambiguous_width_as_wide = true):
    - Box-drawing characters display as width 2, layout breaks
    - wide


  This comparison shows that providing only width_cjk() may cause issues since the layout system doesn't support it.

  WDYT? @joshka @orhun

g4yamanaka:
@joshka @orhun

Just wanted to follow up on Proposal 1 & 2 above. Should I proceed with the implementation, or would you like to discuss further first?

Happy to wait for your confirmation before starting! 👍

orhun:
Feel free to experiment with those options and share results! I personally haven't seen a real world usage of cjk in Ratatui apps yet so maybe removing that wouldn't be a big deal.

orhun:
Adding this to the 0.31 milestone. I think it's going to be pretty useful but I don't think we can make it to 0.30

g4yamanaka:
@joshka I've addressed the feedback. Ready for re-review.

markus-bauer:
This might be a stupid or uninformed comment. But this PR assumes that all terminals render this the same way. Is there a standard for this? It looks like you're shifting from standard unicode width to an assumed standardized terminal width.
```
The issue stems from a mismatch between how unicode-width 
calculates width and how terminals actually display these characters
```
All terminals? Current, old and in the future? Even if they follow the unicode standard?

This is related to other issues:
* Ambiguous width characters
* Unicode width is actually determined by the unicode version, which might be out of sync between unicode-width and the terminal.
* Rendering of emoji, korean script, etc.

You just don't know how the terminal renders.

---
Update: Already found an example.

While a lot of terminals do what you said. Kitty does it "correctly".
KDE Konsole:


Kitty:


Obviously, it depends on the terminal, terminal version, and configuration.

markus-bauer:
I thought about this again, and decided not to get involved. This is a bigger issue than kana rendering.

IMO, this PR doesn't solve anything and breaks terminals that actually do it correctly (unicode correct).

* There's only one relevant standard, that app and terminal can agree on, and its unicode.
* If the terminal emulator doesn't conform to unicode, it's their bug.
* The app can never correctly predict the terminals behaviour.
  There *needs* to be some communication between terminal and app to sync settings like
ambiguous width, unicode version, etc.

These are not new issues, and there has been a lot of discussion elsewhere.

Here's an attempt by kitty do standardize this:
https://github.com/kovidgoyal/kitty/issues/8533

There's still the issue of configuration. How does the app know what the terminal is doing?
How about some environment variables for unicode version, and ambiguous width?

A lot of fun to be had here :). Bye.

g4yamanaka:
@markus-bauer

Thank you for your feedback. Let me clarify a few points.

### Unicode does not define character width

The concept of "Unicode-compliant width" does not exist. The Unicode standard has no specification for how many terminal cells a character should occupy.

- The `unicode-width` crate interprets `Grapheme_Cluster_Break=Extend` as width 0, but this is the crate's own implementation decision, not a Unicode specification.
- Regarding the `East_Asian_Width` property, Unicode explicitly states that "the actual display width is determined by the font and layout." It is not a display width specification.

Therefore, the claim that "terminals not conforming to Unicode are buggy" is based on a standard that doesn't exist.

### Discussion in Kitty issue #8533

In the issue #8533 you referenced, Wukuyon pointed out that half-width katakana voiced marks (U+FF9E, U+FF9F) "As base characters, U+FF9E and U+FF9F also should take up advance width and occupy cells" In other words, Kitty's current behavior (displaying as 1 cell) is what should be fixed according to this discussion.

### Impact on Kitty

After this fix, Kitty will experience the following:
- Half-width katakana with dakuten is displayed as 1 cell, but ratatui calculates it as 2 cells, resulting in extra whitespace
- Copied text will include a space character

This is not layout corruption (overlapping or overflowing characters), but it's not ideal. However, this issue stems from Kitty's grapheme handling, which is being discussed in issue #8533.

**Screenshot (Kitty with this fix applied):**

screenshot


### Improvement for the majority of terminals

Meanwhile, many terminals including Konsole, WezTerm, Alacritty, and iTerm2 currently experience layout corruption. This fix will make them display correctly.

### The significance of introducing TerminalWidthStr

As you pointed out, terminal behavior is not unified. That's exactly why this PR introduces `TerminalWidthStr` as an abstraction layer.

This allows ratatui to control character width calculation without depending on third-party library interpretations. It also enables flexible adaptation if Kitty changes its behavior based on the discussion in issue #8533, or if a terminal standard is established in the future.

markus-bauer:
Thank you for the comprehensive response.
I'm going to keep my reply short; to not waste any more of your time on this.

First, I'd like to apologize. My second comment was written too quickly. There are some strong (and incorrect) claims in there. Sorry.
Your comments about unicode are correct. I'm aware, and should not have written it that way.

Summary:
* This specific PR is related to a wider issue.
* The terminal emulator decides how to display text.
* The terminal app needs to make assumptions about this, but can't know the terminal's behaviour.
* There's no standard for **terminal** cell segmentation and width, AFAIK. (traditionally `wcwidth` was used for some of this).
* Unicode segmentation and width alone is not enough to be used as a command standard between terminal and app. (See the kitty rfc)
* Some parts of this PR make assumptions about how the terminal renders. I was questioning the basis of this.
* Your valid response is: It's how most major terminals do it.

(Nothing about this is new.)

## Kitty

Fair point. Perhaps kitty is not following its own standard. I don't know much about this. And I have not tested this thoroughly at all. I tried kitty first, and it immediately contradicted the claims here.

## TerminalWidthStr

Obviously the correct thing to do.

## Configuration
Finally, Have you considered to make some of these things configurable?
Ambiguous width was mentioned in the discussion here.
Most terminals let you configure how these are displayed (single or double cell). And terminal apps like neovim, for example, have settings for this: https://neovim.io/doc/user/options.html#'ambiwidth'.

I don't have anything more to add to this.

Thank you for your time. Bye.

joshka:
> Finally, Have you considered to make some of these things configurable?

This is a really good point. Something we might want to do here is something much deeper.
What if we brute forced this by throwing a bunch of Unicode (literally all of it) at terminals (all of them), and detect which characters move the cursor by how much, then back fill this into into a bunch standard configs? This would move us away from the ivory tower of what "should be happening" and into the reality of what is actually happening. It's obviously a much more complex solution though, so this could be a small intermediate step in the right direction until then.

g4yamanaka:
Ambiguous width and halfwidth katakana dakuten are separate issues. Ambiguous width needs configuration (varies by CJK/Western context), but halfwidth katakana dakuten is always 2 cells by historical convention—no configuration needed.

a-kenji:
> What if we brute forced this by throwing a bunch of Unicode (literally all of it) at terminals (all of them), and detect which characters move the cursor by how much, then back fill this into into a bunch standard configs?

Yes, that sounds interesting. 
Also there are some escape sequences that can query the actual width of a certain character, if the terminal supports it. 

In which terminals can I reproduce the original problem? I tried running it in a couple terminals and I always get (ratatui v 30.0): 
image

Even in the current default kitty version (kitty 0.44.0)

Edit: Nvm, could reproduce with konsole now. Tried about 10 other terminal emulators before that.

orhun:
We introduced `CellWidth` trait (#2400) which can be used here as well.

I had quickly ported this PR for `CellWidth` in #2499

PR #2196, opened 2025-11-11
feat(layout): create XLayout to extend Layout (POC)

b-guild:
This PR implements an alternative constraint solver algorithm to replace kasuari as the engine behind `Layout`. It is vastly more efficient since it solves the layout directly instead of converting it into a more general constraint solving problem, and it uses a relatively simple algorithm that is explained in detail in the comments of the `ratatui_core::layout::xlayout` module.

In addition to making `Layout` faster, this algorithm offers an extended interface that gives users access to more options for fine-tuning the layout. This is called `XLayout` and it works much like `Layout` and can be smoothly used alongside `Layout`.

The vast majority of tests pass after replacing kasuari with `XLayout`, meaning that most apps should see no difference at all, but there are a few tests that needed to be slightly modified in order to pass. These modifications were made because the output of `XLayout` was either different from kasuari in only trivial ways, or else `XLayout` produced superior output.

An "xlayout" feature was added for the purpose of switching between kasuari and `XLayout` as the solver behind `Layout`. This is merely for demonstration purposes, so that it is easy to compare the difference between the two solvers in apps and through profiling. I would expect that if `XLayout` is superior to kasuari then there would be no need for a feature and `XLayout` could simply permanently replace kasuari.

I hope to start a discussion about whether this or something like this might replace kasuari.

### DIFF
--- a/ratatui-core/Cargo.toml
+++ b/ratatui-core/Cargo.toml
@@ -23,6 +23,7 @@ rustdoc-args = ["--cfg", "docsrs"]
 
 [features]
 default = []
+xlayout = []
 
 ## enables std
 std = [
--- a/ratatui-core/src/buffer/buffer.rs
+++ b/ratatui-core/src/buffer/buffer.rs
@@ -670,6 +670,7 @@ impl fmt::Debug for Buffer {
 
 #[cfg(test)]
 mod tests {
+    extern crate std;
     use alloc::format;
     use alloc::string::ToString;
     use core::iter;
--- a/ratatui-core/src/layout.rs
+++ b/ratatui-core/src/layout.rs
@@ -319,6 +319,7 @@ mod margin;
 mod position;
 mod rect;
 mod size;
+mod xlayout;
 
 pub use alignment::{Alignment, HorizontalAlignment, VerticalAlignment};
 pub use constraint::Constraint;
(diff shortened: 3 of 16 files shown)


ratatui-core/src/layout/direction.rs
```diff
@@ -17,6 +17,17 @@ pub enum Direction {
...
+    #[inline]
+    #[must_use]
+    pub const fn other(self) -> Self {
+        match self {
+            Self::Horizontal => Self::Vertical,
+            Self::Vertical => Self::Horizontal,
+        }
+    }
+}
+
```
joshka, line 35: This seems useful regardless of the new layout. How about submitting this change as a small PR?

b-guild:
I have been using the `flame` crate to do some profiling experiments on release builds, and my measurements show that calling `split` using kasuari takes roughly 100us on average, while calling `split` using `XLayout` takes roughly 3us on average.

b-guild:
It has been proposed that this algorithm might better serve Ratatui as a separate library that copies `ratatui::layout` and replaces the relevant portions with the new algorithm and excludes the kasuari portions, as a faster version of `ratatui::layout`. Unfortunately this would do nothing to immediately improve the efficiency of existing apps and widgets, but it could be a drop-in replacement that any app or widget could effortlessly switch to, and separating it from Ratatui would reduce the maintenance burden upon the Ratatui community and allow Ratatui to stay more focused on rendering to the terminal, which should obviously be the primary concern.

Perhaps such a replacement layout library could be called Tinamui.

Of course, even if this PR would never be merged, it is still very important as a proof of concept and a place for people to express their opinions and suggest improvements to what may become a separate library sometime in the future.

joshka:
Some missing context that may help understanding that suggestion is that maintainer time is finite and often oversubscribed. A "known good enough" solution often beats an "unknown but perfect" solution because of this constraint.

The easiest ways I know to bypass that constraint are:
- small simple non controversial changes
- prove work externally in ways that can augment
- build trust through working with the maintainers over time and develop a reputation for good "taste" (this is a fairly subjective bar obv.)

j-g00da:
I don't think I will have enough time to properly review this until January, but I really want to see it merged. On the other hand we can still proabably release it a bit later after 0.30 as a patch to `ratatui-core`, (but we would have to make sure that the feature is semver-compatible).

PR #2197, opened 2025-11-12
feat(Direction): add `Direction::perpendicular(self)` to get the perpendicular direction

b-guild:
A way to get the perpendicular direction is a natural operation for the `Direction` enum, so a method should be included for completeness.

### DIFF
--- a/ratatui-core/src/layout/direction.rs
+++ b/ratatui-core/src/layout/direction.rs
@@ -12,9 +12,25 @@ use strum::{Display, EnumString};
 #[derive(Debug, Default, Display, EnumString, Clone, Copy, Eq, PartialEq, Hash)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub enum Direction {
+    /// Layout segments are arranged side by side (left to right).
     Horizontal,
+    /// Layout segments are arranged top to bottom (default).
     #[default]
     Vertical,
+}
+
+impl Direction {
+    /// The perpendicular direction to this direction.
+    ///
+    /// `Horizontal` returns `Vertical`, and `Vertical` returns `Horizontal`.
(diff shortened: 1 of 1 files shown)


ratatui-core/src/layout/direction.rs
```diff
@@ -12,11 +12,26 @@ use strum::{Display, EnumString};
...
     #[default]
     Vertical,
 }
 
+impl Direction {
+    /// The perpendicular direction to this direction.
+    /// `Horizontal` returns `Vertical`, and `Vertical` returns `Horizontal`.
+    #[inline]
+    #[must_use = "`other` returns the perpendicular direction"]
+    pub const fn other(self) -> Self {
```
joshka: I didn't think of naming this `perpendicular()` ratther than `other()` until you added this comment here. The benefit is that code that uses this will be obvious to read without having to read the docs. WDYT?

ratatui-core/src/layout/direction.rs
```diff
@@ -12,11 +12,26 @@ use strum::{Display, EnumString};
...
+    /// Layout segments are arranged side by side (left to right).
     Horizontal,
+    /// Layout segments are arranged top to bottom (default).
     #[default]
     Vertical,
 }
 
+impl Direction {
+    /// The perpendicular direction to this direction.
+    /// `Horizontal` returns `Vertical`, and `Vertical` returns `Horizontal`.
```
joshka, line 25: nit: newline between summary comment and details
```suggestion
    /// The perpendicular direction to this direction.
    ///
    /// `Horizontal` returns `Vertical`, and `Vertical` returns `Horizontal`.
```

joshka (approved):
General approval - would prefer the name to be perpendicular() though.

joshka (approved):

PR #2198, opened 2025-11-12
feat(calendar): add width and height functions

joshka:
Fixes https://github.com/ratatui/ratatui/issues/2016

### DIFF
--- a/ratatui-widgets/src/calendar.rs
+++ b/ratatui-widgets/src/calendar.rs
@@ -112,21 +112,19 @@ impl<'a, DS: DateStyler> Monthly<'a, DS> {
         const DAY_WIDTH: u16 = 2;
 
         let mut width = DAYS_PER_WEEK * (GUTTER_WIDTH + DAY_WIDTH);
-
         if let Some(block) = &self.block {
             let (left, right) = block.horizontal_space();
             width = width.saturating_add(left).saturating_add(right);
         }
-
         width
     }
 
     /// Return the height required to render the calendar.
     #[must_use]
     pub fn height(&self) -> u16 {
-        let mut height = u16::from(sunday_based_weeks(self.display_date));
-        height = height.saturating_add(u16::from(self.show_month.is_some()));
-        height = height.saturating_add(u16::from(self.show_weekday.is_some()));
+        let mut height = u16::from(sunday_based_weeks(self.display_date))
+            .saturating_add(u16::from(self.show_month.is_some()))
+            .saturating_add(u16::from(self.show_weekday.is_some()));
 
         if let Some(block) = &self.block {
             let (top, bottom) = block.vertical_space();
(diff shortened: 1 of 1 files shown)


ratatui-widgets/src/calendar.rs
```diff
@@ -200,6 +233,20 @@ impl<DS: DateStyler> Monthly<'_, DS> {
     }
 }
 
+/// Compute how many Sunday-based week rows are needed to render `display_date`.
+///
+/// Mirrors the rendering logic by taking the difference between the first and last day
+/// Sunday-based week numbers (inclusive).
+fn sunday_based_weeks(display_date: Date) -> u8 {
+    let first_of_month = display_date.replace_day(1).unwrap();
```
orhun: nit:

```suggestion
    // Day 1 and the month's last day are always valid, so unwrap() cannot fail.
    let first_of_month = display_date.replace_day(1).unwrap();
```

ratatui-widgets/src/calendar.rs
```diff
@@ -104,6 +104,39 @@ impl<'a, DS: DateStyler> Monthly<'a, DS> {
...
 
+    /// Return the width required to render the calendar.
+    #[must_use]
+    pub fn width(&self) -> u16 {
+        const DAYS_PER_WEEK: u16 = 7;
+        const GUTTER_WIDTH: u16 = 1;
+        const DAY_WIDTH: u16 = 2;
+
+        let base_width = DAYS_PER_WEEK * (GUTTER_WIDTH + DAY_WIDTH);
+        let mut width = base_width;
```
orhun: nit:

```suggestion
        let mut width = DAYS_PER_WEEK * (GUTTER_WIDTH + DAY_WIDTH);
```

ratatui-widgets/src/calendar.rs
```diff
@@ -104,6 +104,39 @@ impl<'a, DS: DateStyler> Monthly<'a, DS> {
...
+
+        width
+    }
+
+    /// Return the height required to render the calendar.
+    #[must_use]
+    pub fn height(&self) -> u16 {
+        let mut height = u16::from(sunday_based_weeks(self.display_date));
+        height = height.saturating_add(u16::from(self.show_month.is_some()));
+        height = height.saturating_add(u16::from(self.show_weekday.is_some()));
```
orhun: ⛓️ 

```suggestion
        let mut height = u16::from(sunday_based_weeks(self.display_date))
            .saturating_add(u16::from(self.show_month.is_some()))
            .saturating_add(u16::from(self.show_weekday.is_some()));
```

ratatui-widgets/src/calendar.rs
joshka:
```suggestion
    let first_of_month = display_date.replace_day(1).expect("first day of month is always valid");
```

PR #2202, opened 2025-11-16
add: clear_first method to block

Koshmar10:
I added clear_first() method that can be called on a block and also added unit test for it 
This pull request closes issue 1975 <https://github.com/ratatui/ratatui/issues/1975>


orhun:
Duplicate: #1997

j-g00da:
Thanks for opening a PR, sadly as @orhun already noted, this is a duplicate. If you still want to contribute to this feature, check the discussion in #1997 first - we came to conclusion that this feature needs a better design before implementing.

PR #2210, opened 2025-11-18
feat(list): truncate

mulac:
Makes better use of the available list area without changing any semantics.

- Fixes a bug where items get excluded despite fitting fully within the list area as shown in test `ensuring_selected_item_is_visible_fills_remaining_space`
- Currently (and now by default) any overflowing items are not shown.  Now you can optionally enable truncation to include of the final overflowing list item when its full height does not fit within the available list area.


orhun:
Can you explain the underlying issue related to the this? (screenshots are also appreciated)

We usually prefer discussing new additions and potential fixes in issues before implementing them.

Issue #1972, opened 2025-07-08
Make sure all examples work in macos terminal.app


joshka:
## Description
Terminal.app doesn't support 24 bit rgb color, so the examples that use it tend to look bad.

As a general rule, Terminal.app has a lot of problems with things that other terminal emulators have supported for decades, so we tend to mostly say that users should install a better terminal rather than encountering these problems. Options for this include:
- [iTerm2](https://iterm2.com/)
- [Kitty](https://sw.kovidgoyal.net/kitty/)
- [Alacritty](https://alacritty.org/)
- [Rio Terminal](https://rioterm.com/)
- [Ghostty](https://ghostty.org/)

Sometimes however that's not possible, and there's a large amount of users that for whatever reason choose to use this software. So we should make our demos etc. work well with Terminal.app (and other apps that don't support truecolor).

## To Reproduce
Fire up colors_rgb, or any of the layout examples (constraint-explorer) etc. in macOS Terminal.app. Any of the examples that use Color::Rgb, or the palette colors will tend to look bad.

## Expected behavior
Colors should look reasonable even when using a non-24bit color terminal

## Screenshots

<img width="1842" height="1183" alt="Image" src="https://github.com/user-attachments/assets/5bc1a493-c4f1-456b-9bca-0ed06ec5326a" />

<img width="1842" height="1183" alt="Image" src="https://github.com/user-attachments/assets/7cd6bdb6-e15f-4db5-bed7-b5de5565db27" />

## Environment
macos / ratatui on main (0.30)

## Additional context
Fixing this is likely a few steps:
- Work out how to detect true color availability
- Make a color detection example to show off the problem
- List the affected examples
- Add color detection to each example where we use palette colors and choose colors that make sense
- For colors-rgb, this probably should just not run as it requires 24bit color to be useful.
- For demo2, the colors are sort of an instrumental part of making this look good.It's unlikely that they'd look as nice without these
- For the other examples, color is probably not as high a priority, it makes things look nice, but isn't the focus of the tool. Consider removing extended colors unless the colors make the example significantly better (many of them do). Otherwise have a fallback approach which uses 4 bit colors (Color::Red, Color::LightBlue etc.)

As a followup, once the example level stuff is done, we should also:
- document the approach for apps to use as a recipe on the website
- think about whether there's anything obvious that the Ratatui lib could do to help make this easier. It's probably out of scope for Ratatui as a library not a framework, but it's worth considering what a generalized solution to this might look like.

[...]

PR #2211, opened 2025-11-18
fix: change flex example colors for MacOS default terminal

ffex:
Part of the fix for #1972 

### Summary
This fix introduces a function to check if we are in a terminal without truecolor(24-bit) and changes the default colors to appear fine of the flex example.

### Notes
The function "is_true_color_supported” is an old problem and there is no common way to determine if a terminal supports or not the truecolor. 

This is the main reason why the function detects specifically the Terminal.app version before the Tahoe. If there are other known terminals with this problem, we can add it to this function.

### DIFF
--- a/examples/apps/flex/src/main.rs
+++ b/examples/apps/flex/src/main.rs
@@ -561,7 +561,6 @@ struct Theme {
     space_around_tab: Color,
     description_fg: Color,
 }
-}
 
 impl Theme {
     pub fn new() -> Self {


orhun:
Generally looks alright to me. I'd prefer someone on MacOS to check this though. Probably @joshka

examples/apps/flex/src/main.rs
```diff
@@ -158,6 +160,16 @@ enum SelectedTab {
 }
 
 impl App {
+    fn new() -> Self {
+        let cs = Theme::new();
```
orhun: nit:

```suggestion
        let theme = Theme::new();
```

examples/apps/flex/src/main.rs
```diff
@@ -382,37 +393,45 @@ impl StatefulWidget for SelectedTab {
     type State = u16;
     fn render(self, area: Rect, buf: &mut Buffer, spacing: &mut Self::State) {
         let spacing = *spacing;
+        let cs = Theme::new();
```
orhun:
```suggestion
        let theme = Theme::new();
```

examples/apps/flex/src/main.rs
```diff
@@ -382,37 +393,45 @@ impl StatefulWidget for SelectedTab {
...
+            Self::SpaceEvenly => Self::render_examples(area, buf, Flex::SpaceEvenly, spacing, cs),
+            Self::SpaceBetween => Self::render_examples(area, buf, Flex::SpaceBetween, spacing, cs),
+            Self::SpaceAround => Self::render_examples(area, buf, Flex::SpaceAround, spacing, cs),
         }
     }
 }
 
 impl SelectedTab {
-    fn render_examples(area: Rect, buf: &mut Buffer, flex: Flex, spacing: u16) {
+    fn render_examples(area: Rect, buf: &mut Buffer, flex: Flex, spacing: u16, cs: Theme) {
```
orhun:
```suggestion
    fn render_examples(area: Rect, buf: &mut Buffer, flex: Flex, spacing: u16, theme: Theme) {
```

examples/apps/flex/src/main.rs
```diff
@@ -508,16 +527,14 @@ impl Example {
         Paragraph::new(text).centered().block(block)
     }
 }
-
-const fn color_for_constraint(constraint: Constraint) -> Color {
-    use tailwind::{BLUE, SLATE};
+const fn color_for_constraint(constraint: Constraint, cs: Theme) -> Color {
```
orhun:
```suggestion
const fn color_for_constraint(constraint: Constraint, theme: Theme) -> Color {
```

examples/apps/flex/src/main.rs
```diff
@@ -529,3 +546,77 @@ fn get_description_height(s: &str) -> u16 {
...
+                end_tab: Color::Indexed(45),
+                space_evenly_tab: Color::Indexed(99),
+                space_between_tab: Color::Indexed(105),
+                space_around_tab: Color::Indexed(111),
+                description_fg: Color::Indexed(111),
+            }
+        }
+    }
+
+    fn is_true_color_supported() -> bool {
```
orhun, line 604: Can you add a comment explaining this function and also the environment variables being used here?

examples/apps/flex/src/main.rs
```diff
@@ -158,6 +160,16 @@ enum SelectedTab {
...
 
 impl App {
+    fn new() -> Self {
+        let theme = Theme::new();
+        Self {
+            selected_tab: SelectedTab::default(),
+            scroll_offset: 0,
+            spacing: 0,
+            state: AppState::default(),
+            theme: cs,
```
joshka: This doesn't compile.

We generally avoid abbreviations in code except in a few very small places as it aids readability. cs probably should just be theme most places in this PR.

ffex: yep, I'll fix and check to the compilation

ffex:
Fix done! what do you think about "is_true_color_supported" do you think is useful add other methods to check the truecolor support? or better wait to issues with specific terminals app? idk!

orhun:
> what do you think about "is_true_color_supported" do you think is useful add other methods to check the truecolor support? or better wait to issues with specific terminals app? idk!

I think it's fine to have it here for now. We could expand it depending on our needs in the future.

The goal should be to fix the example app first :)

gold-silver-copper:
Tested flex example on main and on this PR, this PR definitely fixes it (on my machine)

examples/apps/flex/src/main.rs
```diff
@@ -529,3 +552,84 @@ fn get_description_height(s: &str) -> u16 {
...
+    legacy_tab: Color,
+    start_tab: Color,
+    center_tab: Color,
+    end_tab: Color,
+    space_evenly_tab: Color,
+    space_between_tab: Color,
+    space_around_tab: Color,
+    description_fg: Color,
+}
+impl Theme {
```
orhun, line 565: nit: formatting

```suggestion
}

impl Theme {
```

orhun (approved):
LGTM

examples/apps/flex/src/main.rs
```diff
@@ -529,3 +552,84 @@ fn get_description_height(s: &str) -> u16 {
...
+                ratio_bg: Color::Indexed(20),
+                fill_bg: Color::Black,
+                legacy_tab: Color::Indexed(216),
+                start_tab: Color::Indexed(33),
+                center_tab: Color::Indexed(39),
+                end_tab: Color::Indexed(45),
+                space_evenly_tab: Color::Indexed(99),
+                space_between_tab: Color::Indexed(105),
+                space_around_tab: Color::Indexed(111),
+                description_fg: Color::Indexed(111),
```
joshka: These colors could be a little nicer / closer to the originals:

Image

Image

examples/apps/flex/src/main.rs
```diff
@@ -529,3 +552,84 @@ fn get_description_height(s: &str) -> u16 {
...
+    space_evenly_tab: Color,
+    space_between_tab: Color,
+    space_around_tab: Color,
+    description_fg: Color,
+}
+impl Theme {
+    pub fn new() -> Self {
+        use tailwind::{BLUE, INDIGO, ORANGE, SKY, SLATE};
+
+        if Self::is_true_color_supported() {
```
joshka: For a bit more neatness, I'd suggest putting this check in the main code and having two theme consts instead

e.g. something like:
```rust
let theme = if is_true_color_supported() { TRUE_COLOR } else { INDEXED }

const TRUE_COLOR: Theme = Theme { ... };
const INDEXED: Theme = Theme { ... };
```

examples/apps/flex/src/main.rs
```diff
@@ -508,16 +533,14 @@ impl Example {
         Paragraph::new(text).centered().block(block)
     }
 }
-
-const fn color_for_constraint(constraint: Constraint) -> Color {
-    use tailwind::{BLUE, SLATE};
+const fn color_for_constraint(constraint: Constraint, theme: Theme) -> Color {
```
joshka, line 527: Perhaps make this a method on Theme:

```rust
impl Theme {
    fn color_for_constraint(&self, constraint: Constraint) {...}
}
```

examples/apps/flex/src/main.rs
```diff
@@ -382,37 +393,51 @@ impl StatefulWidget for SelectedTab {
     type State = u16;
     fn render(self, area: Rect, buf: &mut Buffer, spacing: &mut Self::State) {
         let spacing = *spacing;
+        let theme = Theme::new();
```
joshka: Perhaps move this into the state of the tab rather than constructing it each time (my concern is doing the is_supported check feels weird to do on each render)

It may be time to use a slightly different pattern for the state here (see https://github.com/ratatui/ratatui/tree/main/examples/concepts/state for some ideas)

examples/apps/flex/src/main.rs
```diff
@@ -158,6 +160,16 @@ enum SelectedTab {
...
+        Self {
+            selected_tab: SelectedTab::default(),
+            scroll_offset: 0,
+            spacing: 0,
+            state: AppState::default(),
+            theme,
+        }
+    }
     fn run(mut self, terminal: &mut DefaultTerminal) -> Result<()> {
```
orhun, line 172:
```suggestion
    }

    fn run(mut self, terminal: &mut DefaultTerminal) -> Result<()> {
```

ffex:
Sorry, I've been a little busy, now I try to make these improvements :)

examples/apps/flex/src/main.rs
ffex: make it more similar!

examples/apps/flex/src/main.rs
ffex: I implement it with the LazyLock. sorry but I'm quite new with Rust so I am a little bit slow to do some good code. I think that right now is more clear and better (I make it constant and not an object)

ffex:
Screenshot 2026-01-09 at 16 03 23

ok, last commits (sorry I read now how to squash commits) make this changes:
- Implemented Theme like a constant with LazyLock
- Implemented more clear new() with an helper method to put the colors in line
- Changed the ansi256 colors with more similar colors

examples/apps/flex/src/main.rs
```diff
@@ -529,3 +543,76 @@ fn get_description_height(s: &str) -> u16 {
...
+    legacy_tab: Color,
+    start_tab: Color,
+    center_tab: Color,
+    end_tab: Color,
+    space_evenly_tab: Color,
+    space_between_tab: Color,
+    space_around_tab: Color,
+    description_fg: Color,
+}
+}
```
orhun:
```suggestion
}
```

PR #2216, opened 2025-11-19
refactor: Exchange overflow check with saturating_add

pharrison31415:
### Summary
Factory `Rect::new` contains out-of-bound conditions to ensure `x + width` and `y + height` do not overflow `u16`. I believe these can be exchanged with [`saturating_add`](https://doc.rust-lang.org/std/primitive.u16.html#method.saturating_add).

### Testing
Tests `size_truncation` and `size_preservation` appear to be the only candidates for concern regarding this change.


joshka (commented):
LGTM - thanks for the PR

PR #2220, opened 2025-11-21
refactor: `Buffer::pos_of` to call `Rect::pos_of`

pharrison31415:
## Summary
Proposed change in [issue 2217](https://github.com/ratatui/ratatui/issues/2217):

> Function `Buffer::pos_of` returns a coordinate pair of a cell in `self.area: Rect` given an index parameter. It looks like this functionality only touches the `area` and it could be pulled into a function: `Rect::pos_of`. I imagine this functionality is useful in other places than just `Buffer`.

Also included: modification to docstring in `Buffer::symbol` example (bad char).

### DIFF
--- a/ratatui-core/src/buffer/buffer.rs
+++ b/ratatui-core/src/buffer/buffer.rs
@@ -269,21 +269,8 @@ impl Buffer {
         Some(y * width + x)
     }
 
-    /// Returns the (global) coordinates of a cell given its index
-    ///
-    /// Global coordinates are offset by the Buffer's area offset (`x`/`y`).
-    ///
-    /// # Examples
-    ///
-    /// ```
-    /// use ratatui_core::buffer::Buffer;
-    /// use ratatui_core::layout::Rect;
-    ///
-    /// let rect = Rect::new(200, 100, 10, 10);
-    /// let buffer = Buffer::empty(rect);
-    /// assert_eq!(buffer.pos_of(0), (200, 100));
-    /// assert_eq!(buffer.pos_of(14), (204, 101));
-    /// ```
+    /// Returns the (global) coordinates of a cell given its index. It
+    /// runs [`Rect::pos_of`] under the hood on `self.area`.
     ///
     /// # Panics
     ///
--- a/ratatui-core/src/layout/rect.rs
+++ b/ratatui-core/src/layout/rect.rs
@@ -666,6 +666,63 @@ impl Rect {
             self.y.saturating_add((index / width) as u16),
         )
     }
+
+    /// Returns the index in the `Rect` for the given global (x, y) `Position`.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use ratatui_core::layout::{Position, Rect};
+    ///
+    /// let rect = Rect::new(200, 100, 10, 10);
+    /// // Global coordinates to the top corner of this buffer's area
+    /// assert_eq!(rect.index_of(Position::new(200, 100)), 0);
+    /// ```
+    ///
+    /// # Panics
+    ///
+    /// Panics when given an coordinate that is outside of the `Rect`'s area.
+    ///
+    /// ```should_panic
+    /// use ratatui_core::layout::{Position, Rect};
+    ///
+    /// let rect = Rect::new(200, 100, 10, 10);
+    /// // Position is outside of the rect in global coordinate space.
+    /// rect.index_of(Position::new(0, 0)); // Panics
+    /// ```
+    #[track_caller]
+    #[must_use]
+    pub fn index_of(&self, pos: Position) -> usize {
+        self.index_of_opt(pos)
+            .unwrap_or_else(|| panic!("position {pos} outside of rect {self}"))
+    }
+
+    /// Returns an [`Option`]al index in the `Rect` for the given global [`Position`].
+    ///
+    /// Returns [`None`] if the given coordinates are outside of the Buffer's area.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use ratatui_core::layout::{Position, Rect};
+    ///
+    /// let rect = Rect::new(200, 100, 10, 10);
+    /// assert_eq!(rect.index_of_opt(Position::new(200, 100)), Some(0));
+    /// assert_eq!(rect.index_of_opt(Position::new(201, 100)), Some(1));
+    /// assert_eq!(rect.index_of_opt(Position::new(200, 101)), Some(10));
+    /// assert_eq!(rect.index_of_opt(Position::new(0, 0)), None);
+    /// ```
+    #[must_use]
+    pub const fn index_of_opt(&self, position: Position) -> Option<usize> {
+        if !self.contains(position) {
+            return None;
+        }
+        // remove offset
+        let y = (position.y - self.y) as usize;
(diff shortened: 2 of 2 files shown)


ratatui-core/src/layout/rect.rs
```diff
@@ -625,6 +625,47 @@ impl Rect {
...
+    /// let rect = Rect::new(0, 0, 10, 10); // 100 cells in total
+    /// // Index 100 is the 101st cell, which lies outside of the Rect.
+    /// rect.pos_of(100); // Panics
+    /// ```
+    #[must_use]
+    pub fn pos_of(&self, index: usize) -> (u16, u16) {
+        let width = self.width as usize;
+
+        let len = width * self.height as usize;
+        assert!(
```
pharrison31415: In `Buffer::pos_of`, the similar assertion uses the `debug_assert` macro. Should the implementation here be switched to use `debug_assert`?

orhun: Yes, I believe so.

ratatui-core/src/layout/rect.rs
```diff
@@ -625,6 +625,47 @@ impl Rect {
...
+    ///
+    /// ```should_panic
+    /// use ratatui_core::layout::Rect;
+    ///
+    /// let rect = Rect::new(0, 0, 10, 10); // 100 cells in total
+    /// // Index 100 is the 101st cell, which lies outside of the Rect.
+    /// rect.pos_of(100); // Panics
+    /// ```
+    #[must_use]
+    pub fn pos_of(self, index: usize) -> (u16, u16) {
```
orhun: Just a thought, maybe we could use a [`Position`](https://docs.rs/ratatui-core/0.1.0-beta.0/ratatui_core/layout/struct.Position.html) here instead of a tuple? (it's effectively the same thing)

ratatui-core/src/layout/rect.rs
```diff
@@ -1053,4 +1094,24 @@ mod tests {
...
     }
+
+    #[test]
+    fn index_to_coord() {
+        let rect = Rect::new(200, 100, 50, 80);
+
+        assert_eq!(rect.pos_of(0), (200, 100));
+        assert_eq!(rect.pos_of(50), (200, 101));
+        assert_eq!(rect.pos_of(50 * 80 - 1), (249, 179));
+    }
```
orhun: Would be nice to use `rstest` here for parameterized tests

ratatui-core/src/layout/rect.rs
```diff
@@ -625,6 +625,47 @@ impl Rect {
...
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use ratatui_core::layout::Rect;
+    ///
+    /// let rect = Rect::new(200, 100, 10, 10); // 100 total cells
+    /// assert_eq!(rect.pos_of(0), (200, 100));
+    /// assert_eq!(rect.pos_of(14), (204, 101));
+    /// ``````
```
orhun:
```suggestion
    /// ```
```

ratatui-core/src/buffer/buffer.rs
```diff
@@ -39,7 +39,7 @@ use crate::text::{Line, Span};
 ///
 /// // indexing using (x, y) tuple (which is converted to Position)
 /// buf[(0, 1)].set_symbol("B");
-/// assert_eq!(buf[(0, 1)].symbol(), "x");
+/// assert_eq!(buf[(0, 1)].symbol(), "B");
```
orhun, line 42: Hmm I wonder how this test was passing before 🤔

ratatui-core/src/buffer/buffer.rs
```diff
@@ -305,12 +305,8 @@ impl Buffer {
...
             self.content.len()
         );
-        let x = index % self.area.width as usize + self.area.x as usize;
-        let y = index / self.area.width as usize + self.area.y as usize;
-        (
-            u16::try_from(x).expect("x overflow. This should never happen as area.width is u16"),
-            u16::try_from(y).expect("y overflow. This should never happen as area.height is u16"),
-        )
+
+        self.area.pos_of(index)
```
orhun: We should also update the documentation of `Buf::pos_of` to avoid duplicating it at multiple places. I believe it could be adjusted/simplified to say something like "it runs Rect::post_of` under the hood".

ratatui-core/src/layout/rect.rs
pharrison31415: Can do! Should we put the same return type in `Buffer::pos_of`?

orhun: Hmm, that would be breaking and I'm not sure if it's worth it.

I also realized we still have a bunch of tests in the buffer module. Those could be also moved.

pharrison31415: Like, we're just testing the wrapper functionality over `Rect::pos_of` and so we can move them to `Rect`?

orhun: Yeah, the tests should be in the module where the logic is I think.

pharrison31415: The following tests on  `Buffer` use `Buffer::index_of` (which calls `Buffer::index_of_opt`), which, similar to `pos_of`, could be implemented in `Rect`. Should these "index of" functions be added to `Rect` in this PR before moving all those tests out?

* `it_translates_to_and_from_coordinates`
* `index_pos_of_u16_max`

pharrison31415: Oh, I didn't see that [methods / fields on buffer that deal with the underlying content indices are private](https://github.com/ratatui/ratatui/issues/1122). Should this cascade to `Rect` as well? @joshka

orhun: Ah... good catch

ratatui-core/src/buffer/buffer.rs
pharrison31415: I see we're runing `cargo test --doc` in `xtask/src/commands/test_docs.rs`. It fails when I run it locally (?)

joshka:
Some time back I wrote: "I don't think pos_of should be part of the API, and I'd prefer not to make it worse. (note the new index_of_opt method is private)" at https://github.com/ratatui/ratatui/pull/1084#issuecomment-2092544938

The rationale for that comment was that it assumes that the backing store of any buffer / area is linearly indexable. There are other ways that we might want to experiment in the future with how to store the values (trees, 2D allocated areas that are non-contiguous etc.). The index <-> pos conversion there isn't particularly useful. This should be an internal implementation detail of the buffer, rather than something that we expose.

I think we should not make this worse by moving the implementation over to Rect.

I may be wrong / missing some extra context that's been explored on this. Please feel free to add any context and correct me if I'm mistaken.

Recommendation instead of this PR: gradually phase this method out. Document the above in pos_of() ("Don't use this because..."), deprecate eventually, check downstream use cases for why it's in use in apps. Etc.

pharrison31415:
> I think we should not make this worse by moving the implementation over to Rect.

Understood! If there are no other objections, I'll close this PR.

> Recommendation instead of this PR: gradually phase this method out. Document the above in pos_of() ("Don't use this because..."), deprecate eventually, check downstream use cases for why it's in use in apps. Etc.

I'd be more than happy to try my hand at this! Thank you for the guidance.

PR #2221, opened 2025-11-21
test(gauge): add benchmarks for gauge

WaterWhisperer:
# Description

Add benchmarks to the `gauge` widget to uncover eventual performance issues.

Related to: #137

### DIFF
--- a/ratatui/benches/main/gauge.rs
+++ b/ratatui/benches/main/gauge.rs
@@ -1,0 +1,58 @@
+use criterion::{BatchSize, Criterion, criterion_group};
+use ratatui::buffer::Buffer;
+use ratatui::layout::Rect;
+use ratatui::style::Style;
+use ratatui::widgets::{Block, Gauge, Widget};
+
+/// Benchmark for rendering a gauge.
+fn gauge(c: &mut Criterion) {
+    let mut group = c.benchmark_group("gauge");
+
+    let (width, height) = (200, 50); // 1080p fullscreen with medium font
+    let buffer_size = Rect::new(0, 0, width, height);
+
+    // Render an empty gauge
+    group.bench_with_input(
+        format!("render_empty/{width}x{height}"),
+        &Gauge::default(),
+        |b, gauge| {
+            let mut buffer = Buffer::empty(buffer_size);
+            // We use `iter_batched` to clone the value in the setup function because
+            // `Widget::render` consumes the widget.
+            b.iter_batched(
+                || gauge.to_owned(),
+                |bench_gauge| {
+                    bench_gauge.render(buffer.area, &mut buffer);
+                },
+                BatchSize::SmallInput,
+            );
+        },
+    );
+
+    // Render with all features
+    group.bench_with_input(
+        format!("render_all_feature/{width}x{height}"),
+        &Gauge::default()
+            .block(Block::bordered().title("Progress"))
+            .gauge_style(Style::new().white().on_black().italic())
+            .percent(20)
+            .label("20%")
+            .use_unicode(true),
+        |b, gauge| {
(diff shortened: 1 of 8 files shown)


WaterWhisperer:
It's my first PR submission to `ratatui`. I want to use the `ratatui` crate to build my own terminal tool, but before that, I'd like to contribute to ratatui to get familiar with it (it's quite interesting to me). It's also my first time writing benchmarks. I learned and wrote it by reading existing benchmarks, and I hope there are no issues. Thanks for your review.

ratatui/benches/main/gauge.rs
```diff
@@ -0,0 +1,54 @@
...
+                .label("20%")
+                .use_unicode(true),
+            |b, gauge| render(b, gauge, buffer_size),
+        );
+    }
+
+    group.finish();
+}
+
+/// render the gauge into a buffer of the given `size`
```
orhun:
```suggestion
/// Render the gauge into a buffer of the given `size`.
```

WaterWhisperer: Thanks for the review!

ratatui/benches/main/gauge.rs
```diff
@@ -0,0 +1,54 @@
...
+use ratatui::style::Style;
+use ratatui::widgets::{Block, Gauge, Widget};
+
+/// Benchmark for rendering a gauge.
+fn gauge(c: &mut Criterion) {
+    let mut group = c.benchmark_group("gauge");
+
+    for (width, height) in [
+        (100, 50),  // vertically split screen
+        (200, 50),  // 1080p fullscreen with medium font
```
joshka: I'd guess that it's probably enough to just choose the second one of these and that performance of rendering a gauge is not particularly bad enough that we see much difference in these. (Feel free to correct me on this)

```suggestion
        (200, 50),  // 1080p fullscreen with medium font
```

ratatui/benches/main/gauge.rs
```diff
@@ -0,0 +1,54 @@
...
+/// Render the gauge into a buffer of the given `size`.
+fn render(bencher: &mut Bencher, gauge: &Gauge, size: Rect) {
+    let mut buffer = Buffer::empty(size);
+    // We use `iter_batched` to clone the value in the setup function.
+    bencher.iter_batched(
+        || gauge.to_owned(),
+        |bench_gauge| {
+            bench_gauge.render(buffer.area, &mut buffer);
+        },
+        BatchSize::SmallInput,
```
joshka: Can you explain this a bit for someone unfamiliar with the choices here?

ratatui/benches/main/gauge.rs
```diff
@@ -0,0 +1,54 @@
...
+            |b, gauge| render(b, gauge, buffer_size),
+        );
+    }
+
+    group.finish();
+}
+
+/// Render the gauge into a buffer of the given `size`.
+fn render(bencher: &mut Bencher, gauge: &Gauge, size: Rect) {
+    let mut buffer = Buffer::empty(size);
```
joshka: For benchmark/test code that's not too complex, I'd prefer to duplicate rather than DRY most of the time when there's not a compelling reason to extract helper methods. Simple linear methods are easier to grok than methods with multiple levels of closures. WDYT?

ratatui/benches/main/gauge.rs
WaterWhisperer: Sure

ratatui/benches/main/gauge.rs
WaterWhisperer: That's fine. Thanks for the review!

joshka:
Thanks for the PR

PR #2225, opened 2025-11-23
docs: Discourage use of `Buffer`'s `pos_of`, `index_of`

pharrison31415:
## Summary

Per @joshka 's [comment](https://github.com/ratatui/ratatui/pull/2220#issuecomment-3567065394):
> Document the above in pos_of() ("Don't use this because...")

"Usage discouraged" documentation added to `Buffer::pos_of` and `Buffer::index_of`. Does __not__ deprecate functions.

Doc now contains reference to [issue of making methods private](https://github.com/ratatui/ratatui/issues/1122).

### DIFF
--- a/ratatui-core/src/buffer/buffer.rs
+++ b/ratatui-core/src/buffer/buffer.rs
@@ -216,9 +216,8 @@ impl Buffer {
     ///
     /// Global coordinates are offset by the Buffer's area offset (`x`/`y`).
     ///
-    /// Usage discouraged, as it exposes `self.content` as a linearly
-    /// indexable array, which limits potential future abstractions. See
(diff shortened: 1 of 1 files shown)


ratatui-core/src/buffer/buffer.rs
```diff
@@ -216,6 +216,10 @@ impl Buffer {
     ///
     /// Global coordinates are offset by the Buffer's area offset (`x`/`y`).
     ///
+    /// Usage discouraged, as it exposes `self.content` as a linearly
```
joshka: small nit: wrap at 100 chars not 80 for comments.

joshka:
Thanks for the PR.

PR #2235, opened 2025-11-28
feat(canvas)!: Add quadrant, sextant and octant markers

sbarral:
This marker is an alternative to the Braille marker with the same resolution, but offering densely packed, regular pseudo-pixels, without visible bands between rows and columns. However, it uses unicode characters that are less widely supported at the moment, which is why `Braille` was left as the default.

This is technically a breaking change due to the addition of a new variant to `Marker`.

Since much of the logic is shared with the former `BrailleGrid`, the functionality was factorized in a common `TwoByFourGrid`. Part of the Braille logic was rewritten, which makes the former `symbols::braille` module unnecessary.

My availability is unfortunately super-limited and I may not be able to make more than minor changes if the review process deems them necessary, but otherwise I am totally OK with someone else with more free time taking over.

### DIFF
--- a/ratatui-widgets/src/canvas.rs
+++ b/ratatui-widgets/src/canvas.rs
@@ -110,6 +110,24 @@ trait Grid: fmt::Debug {
     fn reset(&mut self);
 }
 
+/// The pattern and color of a `PatternGrid` cell.
+#[derive(Copy, Clone, Debug, Default)]
+struct PatternCell {
+    /// The pattern of a grid character.
+    ///
+    /// The pattern is stored in the lower bits in a row-major order. For instance, for a 2x4
+    /// pattern marker, bits 0 to 7 of this field should represent the following pseudo-pixels:
+    ///
+    /// | 0 1 |
+    /// | 2 3 |
+    /// | 4 5 |
+    /// | 6 7 |
+    pattern: u8,
+    /// The color of a cell only supports foreground colors for now as there's no way to
+    /// individually set the background color of each pseudo-pixel in a pattern character.
+    color: Option<Color>,
+}
+
 /// The `PatternGrid` is a grid made up of cells each containing a `W`x`H` pattern character.
 ///
 /// This makes it possible to draw shapes with a resolution of e.g. 2x4 (Braille or unicode octant)
@@ -118,24 +136,15 @@ trait Grid: fmt::Debug {
 /// support the relevant unicode block, you will see unicode replacement characters (�) instead.
 ///
 /// This grid type only supports a single foreground color for each `W`x`H` pattern character.
-/// There is no way to set the individual color of each octant.
+/// There is no way to set the individual color of each pseudo-pixel.
 #[derive(Debug)]
 struct PatternGrid<const W: usize, const H: usize> {
     /// Width of the grid in number of terminal columns
     width: u16,
     /// Height of the grid in number of terminal rows
     height: u16,
-    /// Represents the pattern for each grid character in row-major order. For instance, for a 2x4
-    /// pattern marker, bits 0 to 7 should represent the following pseudo-pixels:
-    ///
-    /// | 0 1 |
-    /// | 2 3 |
-    /// | 4 5 |
-    /// | 6 7 |
-    patterns: Vec<u8>,
-    /// The color of each cell only supports foreground colors for now as there's no way to
-    /// individually set the background color of each octant in a pattern character.
-    colors: Vec<Option<Color>>,
+    /// Pattern and color of the cells.
+    cells: Vec<PatternCell>,
     /// Lookup table mapping patterns to characters.
     char_table: &'static [char],
 }
@@ -154,8 +163,7 @@ impl<const W: usize, const H: usize> PatternGrid<W, H> {
         Self {
             width,
             height,
-            patterns: vec![0; length],
-            colors: vec![None; length],
+            cells: vec![PatternCell::default(); length],
             char_table,
         }
     }
@@ -171,20 +179,19 @@ impl<const W: usize, const H: usize> Grid for PatternGrid<W, H> {
 
     fn save(&self) -> Layer {
         let contents = self
-            .patterns
+            .cells
             .iter()
-            .zip(&self.colors)
-            .map(|(&pattern, &color)| {
-                let symbol = match pattern {
+            .map(|&cell| {
+                let symbol = match cell.pattern {
                     // Skip rendering blank patterns to allow layers underneath
                     // to show through.
                     0 => None,
-                    _ => Some(self.char_table[pattern as usize]),
+                    idx => Some(self.char_table[idx as usize]),
                 };
 
                 LayerCell {
                     symbol,
-                    fg: color,
+                    fg: cell.color,
                     // Patterns only affect foreground.
                     bg: None,
                 }
@@ -195,8 +202,7 @@ impl<const W: usize, const H: usize> Grid for PatternGrid<W, H> {
     }
 
     fn reset(&mut self) {
-        self.patterns.fill(0);
-        self.colors.fill(None);
+        self.cells.fill_with(Default::default);
     }
 
     fn paint(&mut self, x: usize, y: usize, color: Color) {
@@ -206,11 +212,9 @@ impl<const W: usize, const H: usize> Grid for PatternGrid<W, H> {
             .saturating_add(x.saturating_div(W));
         // using get_mut here because we are indexing the vector with usize values
         // and we want to make sure we don't panic if the index is out of bounds
-        if let Some(c) = self.patterns.get_mut(index) {
-            *c |= 1u8 << ((x % W) + W * (y % H));
-        }
-        if let Some(c) = self.colors.get_mut(index) {
-            *c = Some(color);
+        if let Some(cell) = self.cells.get_mut(index) {
+            cell.pattern |= 1u8 << ((x % W) + W * (y % H));
+            cell.color = Some(color);
         }
     }
 }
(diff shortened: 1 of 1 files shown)


j-g00da (approved):
Thanks! I'm fine with adding this, but let's see what others think.

This is indeed technically a breaking change, but I think we should mark `Marker` as non-exhaustive in the first place.

ratatui-core/src/symbols/braille.rs
```diff
@@ -1,7 +0,0 @@
-pub const BLANK: u16 = 0x2800;
-pub const DOTS: [[u16; 2]; 4] = [
-    [0x0001, 0x0008],
-    [0x0002, 0x0010],
-    [0x0004, 0x0020],
-    [0x0040, 0x0080],
-];
```
j-g00da, line 7: This is a breaking change

sbarral: Oh yes, I missed that.

That being said, this looked a bit arbitrary for a public API. Should I still reinstate it?

In the same vein, I am not too sure whether adding the `octant` module to public symbols was a good idea; it does look a bit less "random" than the content of the `braille` module, but do we want indeed to commit to this as a public API?

sbarral:
> This is indeed technically a breaking change, but I think we should mark `Marker` as non-exhaustive in the first place.

~I thought about that too but wasn't sure whether that belonged to that PR. Do you want me to add it?~
Oh, I see you have opened PR #2236 for that.

ratatui-core/src/symbols/braille.rs
j-g00da: Exposing symbols makes much more sense, I'm too confused why braille is exposed though  - I may be missing some historical context. cc: @orhun @joshka

ratatui-widgets/src/canvas.rs
```diff
@@ -108,67 +108,76 @@ trait Grid: fmt::Debug {
...
-/// This grid type only supports a single foreground color for each 2x4 dots cell. There is no way
-/// to set the individual color of each dot in the braille pattern.
-#[derive(Debug)]
-struct BrailleGrid {
+/// This grid type only supports a single foreground color for each 2x4 pattern character. There is
+/// no way to set the individual color of each octant.
+struct TwoByFourGrid<D>
+where
+    D: Fn(u8) -> char,
+{
```
joshka: Instead of making this generic over a function, what about keeping it simple and having a `[char; 256]` param with the braille characters in the right order?

ratatui-core/src/symbols/braille.rs
joshka: I'm not sure - best bet would be to do a github symbol search.
Nothing obvious in https://github.com/search?q=lang%3Arust+ratatui+symbols+braille+%22DOTS%22+AND+%28NOT+is%3Afork%29+AND+%28NOT+org%3Aratatui%29&type=code

I wouldn't worry about this too much. It probably shouldn't be exposed. Mark it in the breaking changes doc though.

ratatui-core/src/symbols/marker.rs
joshka, line 1: +1 to making this non-exhaustive

joshka (commented):
General approval with the idea.

Would love to also add quadrant and sextant to this.

ratatui-widgets/src/canvas.rs
sbarral: I would also go for it, especially if we decided to modify `symbols::braille` along the lines of `symbols::octant`.

Arrays can't be `const` generic parameters so we would need to hold a slice reference in the grid object (whereas the closure is zero-sized), but I'd say that's not a big deal.

ratatui-core/src/symbols/braille.rs
sbarral: After thinking a bit more about it, I think it would be logical to either (1) remove both `symbols::braille` and `symbols::octant` until we figure out what these should contain, or (2) modify `symbols::braille` to contain an array of all symbols, listed in a sane order (i.e. not the unicode order), just as I did for `symbols::octant`.

If we go with (2), then it would be another argument to follow @joshka suggestion to only use lookup tables rather than decoder functions.

joshka: 2 seems reasonable to me.

j-g00da: yup, 2 sounds good

j-g00da:
#2236 merged, please rebase.
The only thing that needs entry in BREAKING-CHANGES.md now is `symbols::braille::{BLANK, DOTS}` removal.

sbarral:
> Would love to also add quadrant and sextant to this.

Done!

ratatui-core/src/symbols/braille.rs
sbarral: Done!

ratatui-widgets/src/canvas.rs
sbarral: Done.
I also re-named (again) the grid to `PatternGrid` and made it generic over the WxH dimensions to easily support quadrants and sextants.

ratatui-widgets/src/canvas.rs
```diff
@@ -108,67 +110,82 @@ trait Grid: fmt::Debug {
...
+    char_table: &'static [char],
 }
 
-impl BrailleGrid {
-    /// Create a new `BrailleGrid` with the given width and height measured in terminal columns and
-    /// rows respectively.
-    fn new(width: u16, height: u16) -> Self {
+impl<const W: usize, const H: usize> PatternGrid<W, H> {
+    /// Statically check that the dimension of the pattern is supported.
+    const _PATTERN_DIMENSION_CHECK: usize = u8::BITS as usize - W * H;
```
joshka, line 154: neat idea!

ratatui-widgets/src/canvas.rs
```diff
@@ -108,67 +110,82 @@ trait Grid: fmt::Debug {
...
     /// Width of the grid in number of terminal columns
     width: u16,
     /// Height of the grid in number of terminal rows
     height: u16,
-    /// Represents the unicode braille patterns. Will take a value between `0x2800` and `0x28FF`
-    /// this is converted to an utf16 string when converting to a layer. See
-    /// <https://en.wikipedia.org/wiki/Braille_Patterns> for more info.
-    utf16_code_points: Vec<u16>,
+    /// Represents the pattern for each grid character in row-major order. For instance, for a 2x4
+    /// pattern marker, bits 0 to 7 should represent the following pseudo-pixels:
```
joshka: bits seems like the wrong term here.

ratatui-widgets/src/canvas.rs
```diff
@@ -108,67 +110,82 @@ trait Grid: fmt::Debug {
...
+    ///
+    /// | 0 1 |
+    /// | 2 3 |
+    /// | 4 5 |
+    /// | 6 7 |
+    patterns: Vec<u8>,
     /// The color of each cell only supports foreground colors for now as there's no way to
-    /// individually set the background color of each dot in the braille pattern.
+    /// individually set the background color of each octant in a pattern character.
     colors: Vec<Option<Color>>,
```
joshka: These two fields seem like they could be joined perhaps to make it clear that the number of items in patterns and colors is the same.

ratatui-widgets/src/canvas.rs
```diff
@@ -108,67 +110,82 @@ trait Grid: fmt::Debug {
...
+    /// pattern marker, bits 0 to 7 should represent the following pseudo-pixels:
+    ///
+    /// | 0 1 |
+    /// | 2 3 |
+    /// | 4 5 |
+    /// | 6 7 |
+    patterns: Vec<u8>,
     /// The color of each cell only supports foreground colors for now as there's no way to
-    /// individually set the background color of each dot in the braille pattern.
+    /// individually set the background color of each octant in a pattern character.
```
joshka: is octant right here?

joshka (approved):
LGTM. A couple of minor notes non blocking.

This PR either written by someone with a lot of attention to detail, or with a good grasp of how to use AI tooling well. If it's the latter I'd love to hear your thoughts over on https://github.com/ratatui/ratatui/discussions/2201

ratatui-widgets/src/canvas.rs
sbarral: I am not sure what you mean but am open to any suggestion. What I meant is that bits 0 in the `u8` represent the top left pixel etc.

sbarral:
> LGTM. A couple of minor notes non blocking.
> 
> This PR either written by someone with a lot of attention to detail, or with a good grasp of how to use AI tooling well. If it's the latter I'd love to hear your thoughts over on #2201

Nope, no AI here and unfortunately not many insights in that respect. But funnily enough, I was actually reading today the AI-related issues in `ratatui` as an inspiration for our own repos :-)

ratatui-widgets/src/canvas.rs
joshka: Ah got it - I assumed wrongly this was storing chars (thinking that the u8 was just a substitute for an 8 bit character). It's actually being used to store the pixels packed into the cell.

Probably add a small note about sextants / quadrants using just the first 6 / 4 bits would be nice here.

ratatui-widgets/src/canvas.rs
sbarral: Definitely, this was inherited from the original code where the codepoints were for some reason a separate field...

ratatui-widgets/src/canvas.rs
sbarral: I tried to be a bit clearer (now in `PatternCell`), let me know if this is still unclear.

joshka:
Thanks for the PR!

PR #2236, opened 2025-11-28
feat(symbols)!: Make `Marker` non-exhaustive

j-g00da:
This will allow us to add new markers without causing further breaking changes.

BREAKING CHANGE: `Marker` is now non-exhaustive


j-g00da:
Since `Canvas` is in another crate, `Marker` requires wildcard there, which doesn't make sense for us since it's the same workspace anyway. 

Makes me wonder if we should move `Marker` to `ratatui-widgets`.

As a workaround I default to `Marker::Dot` handling (as this is `Marker::default()`).

joshka (approved):

PR #2239, opened 2025-11-30
feat: Support adding an Offset to Position

joshka:
Adds Position::offset() and arithmentic ops (Position + Offset and
Position - Offset)

Fixes: https://github.com/ratatui/ratatui/issues/2018


orhun (commented):
LGTM

PR #2240, opened 2025-11-30
feat: add Rect::resize() method

joshka:
Fixes: https://github.com/ratatui/ratatui/issues/1440

---
Note: stacked on https://github.com/ratatui/ratatui/pull/1596 merge that first


joshka:
Note: this does the simplest thing here instead of what was discussed earlier.

resize() takes a Size and clamps to that size.

j-g00da (approved):
LGTM

PR #2241, opened 2025-11-30
feat(layout): support per-side insets via Rect::inner

joshka:
- Introduce Inset for side-specific shrinking with symmetric/horizontal/vertical helpers and Margin conversion
- update Rect::inner to accept Into<Inset>

---

This is inspired by (but not a copy of) the Insets in Codex:
https://github.com/openai/codex/blob/6eeaf46ac136f6ea0352641a5daea5006195e09f/codex-rs/tui/src/render/mod.rs#L8

The differences in the implementation here are:
- Top Right Bottom Left order to match CSS order
- Changes the inner() method instead of adding a new insets() method
- Singular vs plural naming
- horizontal/vertical order to match most of Ratatui's ordering of values (usually x then y)

### DIFF
--- a/ratatui-core/src/layout/inset.rs
+++ b/ratatui-core/src/layout/inset.rs
@@ -1,0 +1,189 @@
+#![warn(missing_docs)]
+use core::fmt;
+
+use crate::layout::Margin;
+
+/// Represents side-specific spacing inside rectangular areas.
+///
+/// `Inset` defines how much space to remove from each side of a rectangle. Unlike [`Margin`], which
+/// applies uniform spacing horizontally and vertically, `Inset` lets you specify independent
+/// amounts for the top, right, bottom, and left edges. The default constructor order is
+/// top-right-bottom-left (often remembered as “trbl” or “trouble”), matching the CSS spec and
+/// offering an easy clockwise mnemonic.
+///
+/// Use `Inset` when you need per-side control; choose [`Margin`](crate::layout::Margin) for the
+/// common symmetric case.
+///
+/// # Construction
+///
+/// - [`trbl`](Self::trbl) - Create a new inset with top/right/bottom/left values
+/// - [`default`](Default::default) - Create with zero inset on all sides
+/// - [`symmetric`](Self::symmetric) - Create with shared horizontal and vertical values
+/// - [`horizontal`](Self::horizontal) - Create with equal left and right values
+/// - [`vertical`](Self::vertical) - Create with equal top and bottom values
+///
+/// # Examples
+///
+/// ```rust
+/// use ratatui_core::layout::{Inset, Rect};
+///
+/// let inset = Inset::trbl(1, 2, 3, 4);
+/// let rect = Rect::new(0, 0, 10, 10).inner(inset);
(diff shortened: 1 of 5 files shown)


orhun:
I see that this might be useful but I'd like to ask where was this discussed / what problem is this solving? 👀

joshka-oai:
This makes it easy to define layouts where there's varying insets (e.g. a left side padding of all items rendered into that area, but leave the right side alone). We can do the same thing using `Layout`, but this is significantly simpler for simple use cases.

A hypothetical example: rendering an indented bulleted list into an area where the following lines have a hanging indent. Another example is rendering list items where you want padding before the item, but not after (or the opposite). using `Layout` can be pretty heavyweight for that task.

This addresses https://github.com/ratatui/ratatui/issues/931 somewhat, with some other parallel discussion at https://github.com/ratatui/ratatui/pull/1285

It's possible that we should pull block::Padding up into core and use that instead of insets.
Happy to discuss alternatives about what makes sense here. I'm not wedded to the naming. Mostly to the concept / idea.

We could also make this a new method rather than breaking inner like this, but I think this works fairly nicely as a new `Into<X>` type thing. Maybe X is `Padding` instead of `Margin`?

(Also to be clear, this is not me putting my day job hat on and saying we need this in Codex, this is me putting my Ratatui hat on and saying the implementation in Codex is neat and it makes sense to add this to Ratatui rather than it being in codex)

ratatui-core/src/layout/inset.rs
```diff
@@ -0,0 +1,174 @@
...
+    pub right: u16,
+    /// Space to remove from the bottom edge
+    pub bottom: u16,
+    /// Space to remove from the left edge
+    pub left: u16,
+}
+
+impl Inset {
+    /// Creates a new inset with explicit top/right/bottom/left values.
+    pub const fn trbl(top: u16, right: u16, bottom: u16, left: u16) -> Self {
```
pharrison31415, line 49: Why not `new`? I think `trbl` has an advantage of specifying all four directions, but I think that takes an extra step of a user's thinking to figure out.

joshka-oai: TLDR - The rationale for this is that trbl is intention revealing (clockwise from the top) and it matches the CSS spec.

There's a bunch of "correct" orderings of this - this particular one is the one used in CSS and is known as the "trouble" ordering so there's a useful mnemonic. The version used in Codex (which is derived from the ordering in Chrome originally) is tlbr (anti-clockwise from the top). tblr is pretty reasonable (symmetric vertical first), and lrtb (symmetric horizontal first) probably makes most sense with respect to how we tend to do things in Ratatui code (column is first parameter then row).

We could reasonable add all of these, but then you get a mix of patterns in what code uses, and that tends to be messy. Instead I chose what I think is a reasonable good choice to use as a single item.

It's likely some of this comment should end up in the doc comment here to help answer the why. Will do so when I'm home.

joshka: Resolved in https://github.com/ratatui/ratatui/pull/2241/commits/228fb6de3fcd91ef15cdc9e361f133c2148df905

PR #2251, opened 2025-12-01
style(rect): use plus operator for offset

pharrison31415:
## Summary
Use `+` operator to move `Rect` by an `Offset` as added [here](https://github.com/ratatui/ratatui/pull/1596).

Includes change to use `Offset::new()` in place of `Offset { ... }`


orhun (approved):
thanks!

PR #2252, opened 2025-12-03
test(rect): mutual intersection agreement

pharrison31415:
## Summary

Add tests in `Rect` to ensure `rect0` intersecting `rect1` if and only if `rect1` intersecting `rect0`.


joshka:
LGTM. Good addition to test coverage.

PR #2265, opened 2025-12-10
feat(widgets): separate HighlightSpacing into ListHighlightSpacing and TableHighlightSpacing (#1767)

### LINKED ISSUE #1767: Move HighlightSpacing to ratatui-core or new ratatui-common-widget-components
## Problem

`HighlightSpacing` is used by table and list but is currently located with the table. (results in odd imports for List: `use crate::table::HighlightSpacing;`)

More external widgets might want to use it which shouldn't need to depend on the whole `ratatui-widgets`.

## Solution

Move the `HighlightSpacing` enum to `ratatui-core` or some `ratatui-common-widget-components`. (When thinking about the latter, Scrollbar or Block would be also good candidates for such an in-between crate. Or even better: split every widget into its own crate so `ratatui-list-widget` and `ratatui-table-widget` exist on their own, removing the `ratatui-widget` crate. Should probably increase compile time on multicore systems and allows development cycles independent of each other. Probably worth its own issue, shouldn't be discussed here.)

NoOPeEKS:
This commit separates the `HighlightSpacing` enum into two different enums: `ListHighlightSpacing` and `TableHighlightSpacing`, each on their own separate module and should close #1767.

This is because, as stated by #1767, until now, both widgets `List` and `Table` used the `HighlightSpacing` under `crate::table::HighlightSpacing`. 

This is a BREAKING CHANGE, as it also changes the re-exports under `ratatui::widgets::HighlightSpacing` to `ratatui::widgets::ListHighlightSpacing` and `ratatui::widgets::TableHighlightSpacing`.

### DIFF
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -30,6 +30,8 @@ This is a quick summary of the sections below:
   - `Marker` is now non-exhaustive
   - `symbols::braille::BLANK` and `symbols::braille::DOTS` have been removed in favor of an ordered
     array of all Braille characters
+  - `widgets::HighlightSpacing` has been separated into `widgets::ListHighlightSpacing` and
+    `widgets::TableHighlightSpacing`
 - [v0.29.0](#v0290)
   - `Sparkline::data` takes `IntoIterator<Item = SparklineBar>` instead of `&[u64]` and is no longer
     const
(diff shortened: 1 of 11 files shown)


orhun:
Add a note in `BREAKING-CHANGES.md` and this should be good to go :)

NoOPeEKS:
@orhun Added to the `BREAKING-CHANGES.MD`!

orhun (approved):

PR #2266, opened 2025-12-10
feat(core): Add color-aware buffer view for TestBackend

### LINKED ISSUE #1402: Support asserting `TestBackend` buffer with color
## Problem

It is currently possible to write unit tests as follows:

```rust
let backend = TestBackend::new(40, 10);
terminal.draw(|frame| render(frame, &mut app, &colors))?;
assert_buffer(
    Buffer::with_lines(vec![
        "╭Parameters──────────────────────|all|─╮",
        "│user.name                   system    │",
        "│kernel.fictional.test_param 0         │",
        "│vm.stat_interval            1         │",
        "│                                      │",
        "│                                      │",
        "│                                      │",
        "│                                      │",
        "│                                  1/3 │",
        "╰──────────────────────────────────────╯",
    ]),
    terminal.backend(),
)?;

```

https://github.com/orhun/systeroid/blob/b232810be7c2f6a0836a94c521a70f1edb8d9253/systeroid-tui/tests/integration_test.rs#L66-L83 

However, asserting the colors is not possible due to `Display` implementation on `TestBackend` (i.e. [`buffer_view`](https://github.com/ratatui/ratatui/blob/0ed71d14bbe7d08905d525566f8c2678a5ffd57c/src/backend/test.rs#L49) method) not taking the style into account.

## Solution

We can probably add another method for retrieving the colored buffer. Using ANSI in the output might not be easy to work with, so maybe we can serialize to something that is easy to compare visually. (e.g. `{red}foo{/red}`)

## Additional context

See https://github.com/ratatui/ratatui-website/pull/739

Zachgille:
This PR adds a color-aware buffer view for `TestBackend` so tests and snapshot recipes can assert styles, not just text.

- Add `buffer_view_colored(&Buffer) -> String` helper in `ratatui-core/src/backend/test.rs`.
- Add `TestBackend::buffer_view_colored()` to expose the colored view.
- Serialize styles into inline tags such as `{fg=Yellow,bg=Blue,mod=BOLD}x{/}`.
- Keep the existing `buffer_view` / `Display` output unchanged.
- Add unit tests for unstyled, styled, and modifier cases.

Closes #1402.

### DIFF
--- a/ratatui-core/src/backend/test.rs
+++ b/ratatui-core/src/backend/test.rs
@@ -177,11 +177,7 @@ fn buffer_view_colored(buffer: &Buffer) -> String {
         view.push('"');
 
         if !overwritten.is_empty() {
-            write!(
-                &mut view,
-                " Hidden by multi-width symbols: {overwritten:?}"
-            )
-            .unwrap();
+            write!(&mut view, " Hidden by multi-width symbols: {overwritten:?}").unwrap();
         }
 
         view.push('\n');
@@ -589,7 +585,7 @@ mod tests {
     use itertools::Itertools as _;
 
     use super::*;
-
+    use crate::layout::Rect;
     use crate::style::{Color, Modifier, Style};
 
     #[test]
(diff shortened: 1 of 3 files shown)


ratatui-core/src/backend/test.rs
```diff
@@ -67,6 +68,124 @@ fn buffer_view(buffer: &Buffer) -> String {
...
+        }
+    }
+
+    fn is_default(&self) -> bool {
+        self.fg == Color::Reset && self.bg == Color::Reset && self.modifier.is_empty()
+    }
+}
+
+/// Append a style-change tag like `{fg=Red,bg=Black,mod=BOLD}` or `{/}` to `out`.
+fn append_style_tag(out: &mut String, style: &StyleSnapshot) {
```
joshka, line 103: Would it make sense for this to be `impl Display for StyleSnapshot`?

ratatui-core/src/backend/test.rs
```diff
@@ -67,6 +68,124 @@ fn buffer_view(buffer: &Buffer) -> String {
...
+            modifier: cell.modifier,
+        }
+    }
+
+    fn is_default(&self) -> bool {
+        self.fg == Color::Reset && self.bg == Color::Reset && self.modifier.is_empty()
+    }
+}
+
+/// Append a style-change tag like `{fg=Red,bg=Black,mod=BOLD}` or `{/}` to `out`.
```
joshka, line 102: Would something like `{Bold Red on Black}` be shorter / more convenient to read?
Or even `{Red,OnBlack,Bold}` - skipping all the extra words

Mostly - trying to look for a good representation here that makes it easy to read the output quickly.

ratatui-core/src/backend/test.rs
```diff
@@ -105,6 +224,14 @@ impl TestBackend {
...
     }
 
+    /// Returns a string representation of the buffer including color / modifiers.
+    ///
+    /// This is intended for use in tests where you want to assert on styling,
+    /// not just on the rendered text.
+    pub fn buffer_view_colored(&self) -> String {
+        buffer_view_colored(&self.buffer)
+    }
+
```
joshka, line 234: Something that seems a bit more idiomatic on this (similar to e.g. Path::display which returns [std::path::Display](https://doc.rust-lang.org/std/path/struct.Display.html)) would be (with a rename)

```rust
fn colored_display(&self) -> ColoredDisplay

// and then

struct ColoredDisplay;

impl Display for ColoredDisplay { ... }
```

joshka (commented):
Good idea, some questions / ideas for improving this.

PR #2270, opened 2025-12-15
chore: expose crossterm 0.28/0.29 feature flags in Ratatui

orhun:
In docs we currently say that individual versions of crossterm could be enabled like this:

```toml
ratatui = { version = "0.30.0-beta", features = ["crossterm_0_28"] }
```

However this wasn't actually possible:

```
package `x` depends on `ratatui` with feature `crossterm_0_28` but `ratatui` does not have that feature.

failed to select a version for `ratatui` which could resolve this conflict
```

This PR fixes that by exposing respective feature flags.
(Tested locally)

### DIFF
--- a/ratatui/Cargo.toml
+++ b/ratatui/Cargo.toml
@@ -28,7 +28,7 @@ rustdoc-args = ["--cfg", "docsrs"]
 ## which allows you to set the underline color of text, the `macros` feature which provides
 ## some useful macros and `layout-cache` which speeds up layout cache calculations
 ## in `std`-enabled environments.
-default = ["crossterm_0_29", "underline-color", "all-widgets", "macros", "layout-cache"]
+default = ["crossterm", "underline-color", "all-widgets", "macros", "layout-cache"]
 #! Generally an application will only use one backend, so you should only enable one of the following features:
 ## enables the [`CrosstermBackend`](backend::CrosstermBackend) backend and adds a dependency on [`crossterm`].
 crossterm = ["std", "dep:ratatui-crossterm"]


ratatui/Cargo.toml
```diff
@@ -28,10 +28,14 @@ rustdoc-args = ["--cfg", "docsrs"]
 ## which allows you to set the underline color of text, the `macros` feature which provides
 ## some useful macros and `layout-cache` which speeds up layout cache calculations
 ## in `std`-enabled environments.
-default = ["crossterm", "underline-color", "all-widgets", "macros", "layout-cache"]
+default = ["crossterm_0_29", "underline-color", "all-widgets", "macros", "layout-cache"]
```
j-g00da: Let's leave it implicit, IIRC it defaults to 0.29, so this way we should be able to enable 0.28 without having to disable default features.
```suggestion
default = ["crossterm", "underline-color", "all-widgets", "macros", "layout-cache"]
```

joshka:
Hrm, I think it would be worth keeping the ratatui import simple as possible - default to the latest available crossterm always. WDYT?

j-g00da:
> Hrm, I think it would be worth keeping the ratatui import simple as possible - default to the latest available crossterm always. WDYT?

Huh, just checked and it's currently a compiler error if you don't enable any feat in `ratatui-crossterm`, I was sure we agreed on the latest being used as a fallback.

joshka:
> Huh, just checked and it's currently a compiler error if you don't enable any feat in ratatui-crossterm, I was sure we agreed on the latest being used as a fallback.

Yeah, I couldn't recall. Worth digging into the history.

orhun:
I made the suggested change. All seems fine now.

> Huh, just checked and it's currently a compiler error if you don't enable any feat in ratatui-crossterm, I was sure we agreed on the latest being used as a fallback.

It was implemented in #1841. It seems to me that we haven't done that for some reason. I also thought that was the case. 

I don't think that's affecting the problem described here and I have cooked up another PR to potentially fix that. Let's discuss it there.

orhun:
#2275

PR #2275, opened 2025-12-16
chore: use latest crossterm as fallback in ratatui-crossterm

orhun:
As discussed in #2274

Before it wasn't possible to:

```
$ cargo build -p ratatui-crossterm --no-default-features

error: At least one crossterm feature must be enabled. See the crate docs for more information.
  --> ratatui-crossterm/src/lib.rs:87:9
   |
87 | /         compile_error!(
88 | |             "At least one crossterm feature must be enabled. See the crate docs for more information."
89 | |         );
   | |_________^
```

This PR makes it possible to use `ratatui-crossterm` without default features and it always falls back to the latest version if none is provided.

Just as a note, I initially thought adding a new feature such as `crossterm_latest` but the same crate cannot be used by two different package aliases:

> error: the crate `ratatui-crossterm v0.1.0-beta.0 (/home/orhun/gh/ratatui/ratatui-crossterm)` depends on crate `crossterm v0.29.0` multiple times with different names

So I simply renamed `crossterm_0_29` package alias to `crossterm_latest` and used that.
The rest of the feature flags are the same.

### DIFF
--- a/ratatui-crossterm/Cargo.toml
+++ b/ratatui-crossterm/Cargo.toml
@@ -38,10 +38,7 @@ crossterm_0_29 = []
 #! certain situations.
 
 ## Enables serde for crossterm dependency
-serde = [
-  "crossterm_latest/serde",
-  "crossterm_0_28?/serde",
-]
+serde = ["crossterm_latest/serde", "crossterm_0_28?/serde"]
 
 ## enables the backend code that sets the underline color.
 ## Underline color is not supported on Windows 7.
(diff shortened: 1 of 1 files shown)


j-g00da:
Ahhhh... now I remember, this will pull two versions of crossterm if one chooses to use 0.28...

I think we shouldn't do it then and instead add `crossterm_0_29` as default feature in the main crate, as you originally did. Sorry for that :|

PR #2291, opened 2025-12-23
chore: revert the 0.30-beta.1 update to re-trigger release-plz

orhun:


j-g00da:
Will this work if some of the crates got already released?

orhun:
we'll see

PR #2297, opened 2025-12-27
feat: Add AsRef impls for widget types

joshka:
Provide AsRef<Self> for built-in widgets so AsRef-bounded rendering can use widget.as_ref() directly.

Example:

```rust
let block = Block::new();
block.as_ref().render(area, buf);
// do something later with block
```

Fixes: https://github.com/ratatui/ratatui/discussions/2296

### DIFF
--- a/ratatui-widgets/src/as_ref.rs
+++ b/ratatui-widgets/src/as_ref.rs
@@ -1,0 +1,111 @@
+/// Implement `AsRef<Self>` for widget types to enable `as_ref()` in generic contexts.
+///
+/// This keeps widget rendering ergonomic when APIs accept `AsRef<WidgetType>` bounds, avoiding
+/// the need for `(&widget).render(...)` just to satisfy a trait bound.
+///
+/// # Example
+///
+/// ```rust
+/// use ratatui_widgets::block::Block;
+///
+/// let block = Block::default();
+/// let block_ref: &Block<'_> = block.as_ref();
+/// ```
+///
+/// # Generated impls
+///
+/// ```rust,ignore
+/// // Non-generic widgets (e.g. Clear, RatatuiLogo).
+/// impl AsRef<Clear> for Clear {
+///     fn as_ref(&self) -> &Clear {
+///         self
+///     }
+/// }
+///
+/// // Generic widgets (e.g. Block with a lifetime, Canvas with a lifetime + type parameter).
+/// impl<'a> AsRef<Block<'a>> for Block<'a> {
+///     fn as_ref(&self) -> &Block<'a> {
+///         self
+///     }
+/// }
+///
+/// impl<'a, F> AsRef<Canvas<'a, F>> for Canvas<'a, F>
+/// where
+///     F: Fn(&mut Context),
+/// {
+///     fn as_ref(&self) -> &Canvas<'a, F> {
+///         self
+///     }
+/// }
+/// ```
+macro_rules! impl_as_ref {
+    ($type:ty, <$($gen:tt),+> $(where $($bounds:tt)+)?) => {
+        impl<$($gen),+> AsRef<$type> for $type $(where $($bounds)+)? {
+            fn as_ref(&self) -> &$type {
+                self
+            }
+        }
+    };
+    ($type:ty) => {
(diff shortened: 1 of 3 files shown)


joshka:
> Thank you for opening this pull request!

bah humbug

typos.toml
```diff
@@ -8,3 +8,6 @@ sectore = "sectore" # https://github.com/sectore
 extend-ignore-re = [
   "\\[[[:xdigit:]]{7}\\]\\(https://github.com/ratatui/ratatui/commit/[[:xdigit:]]{40}\\)",
 ]
+
+[files]
+extend-exclude = ["CHANGELOG.md"]
```
joshka: There's a `arithmentic` typo there.

sxyazi:
FWIW, since a type can have multiple `AsRef` implementations, downstream code may have added their own `AsRef` implementations for these widgets, for example:

```rust
#[repr(transparent)]
struct MyLine<'a>(Line<'a>);

impl<'a> AsRef<MyLine<'a>> for Line<'a> {
  fn as_ref(&self) -> &MyLine<'a> {
    unsafe { &*(self as *const Line<'a> as *const MyLine<'a>) }
  }
}
```

Adding a new `AsRef` upstream may affect type inference for `as_ref()` (since there are now multiple possible implementations), which can lead to compilation errors, so this PR might need to be introduced as a breaking change.

Also, since this PR is intended to make `widget.as_ref().render()` work, it means upstream can no longer add other `AsRef` implementations in the future, such as:

```rust
impl AsRef<Style> for Line<'_> {
  fn as_ref(&self) -> &Style {
    &self.style
  }
}
```

otherwise that use case would be broken

typos.toml
joshka: Fixed in https://github.com/ratatui/ratatui/pull/2300

joshka:
I'm generally ok with these sorts of fairly marginal edge case things being noted as breaking, but not being a semver bump unless there's a strong downstream problem that it's actually causing (we've done a number of breaks inference type changes - they're usually in 0.x.0 releases so this is fine, but I'd be ok with them being in a 0.x.1 release if needed).

Do you know of any code actually using AsRef impls, or was this more hypothetical? (Nothing in https://github.com/search?q=ratatui+%22impl+AsRef%22&type=code looks like it would be impacted)

This is one of those things that I think is worth bearing in mind generally with respect to rust's general approach to breaking changes.

Anyway, this seems in keeping with the [docs on AsRef ](https://doc.rust-lang.org/std/convert/trait.AsRef.html#reflexivity):

> A trivial implementation of `AsRef<T> for T` must be added explicitly for a particular type `T` where needed or desired. Note, however, that not all types from std contain such an implementation, and those cannot be added by external code due to orphan rules.

Edit: added Breaking changes doc entry for this regardless

sxyazi:
> Do you know of any code actually using AsRef impls, or was this more hypothetical? (Nothing in https://github.com/search?q=ratatui+%22impl+AsRef%22&type=code looks like it would be impacted)

Just a hypothesis and likely won't cause much impact since I haven't ever used it that way nor have I seen others do so

orhun (approved):

ratatui-widgets/src/as_ref.rs
```diff
@@ -0,0 +1,111 @@
...
+        }
+    };
+    ($type:ty) => {
+        impl AsRef<$type> for $type {
+            fn as_ref(&self) -> &$type {
+                self
+            }
+        }
+    };
+}
```
j-g00da, line 56: Ugh I'm assuming we can't do a blanket impl here?

Probably not a big deal since the macro is pretty simple.

joshka: There's a few things we could do here. Blanket impl (but that might affect all widgets and I wasn't sure we wanted that), derive_more has an AsRef derive (but figured that adding a new dep for this was probably heavier than needed) or manual impls (which is noisy extra undifferentiated code). Figured this was the simplest of the options.

Macros can be difficult, but I think this is not too bad because It's fairly narrowly scoped, easy to see what it does, and is well tested.

PR #2298, opened 2025-12-27
fix: import path for crossterm event module

coding-retro:
I was trying to look at the examples and noticed an error, this was cause instead of using crossterm 0.29 I was just using ratatui::crossterm. and you can use this instead of having to include a whole new dependency this just make the whole thing more user friendly and work with two dependencies instead of three!

### DIFF
--- a/ratatui-widgets/examples/paragraph.rs
+++ b/ratatui-widgets/examples/paragraph.rs
@@ -15,7 +15,7 @@
 //! [examples readme]: https://github.com/ratatui/ratatui/blob/main/examples/README.md
 
 use color_eyre::Result;
-use crossterm::event;
+use ratatui::crossterm::event;
 use ratatui::Frame;
 use ratatui::layout::{Alignment, Constraint, Layout, Rect};
 use ratatui::style::{Color, Stylize};


joshka:
Hi there, we've decided in the past not to use this to avoid confusion for new users about what `ratatui::crossterm` is.
It requires explaining re-exports and how the module system in rust works in order to understan this.
The re-export is there for libraries that need to ensure that they are using the same version of things that are in the public export of this library.

coding-retro:
> Hi there, we've decided in the past not to use this to avoid confusion for new users about what `ratatui::crossterm` is. It requires explaining re-exports and how the module system in rust works in order to understan this. The re-export is there for libraries that need to ensure that they are using the same version of things that are in the public export of this library.

Ah okay that makes sense I was going through more of the examples I also noticed that https://github.com/ratatui/ratatui/blob/main/ratatui-widgets/examples/canvas.rs has a typo I believe? it says "use ratatui_widgets::canvas::Points;" but it shouldn't be ratatui_widgets but ratatui::widgets should I make a pull request for that?

joshka:
> use ratatui_widgets::canvas::Points

That's a little more nuanced :)
`ratatui::widgets` imports a bunch of things from the `ratatui-widgets` crate into the old places that the widgets were.
There may be some places where it's impossible to reference `ratatui` instead of ratatui_widgets` due to circular references. Try it though. It may work, or it may cause problems (you'll have to run all the various things that run in CI to check this - I forget the exact details). We need to clean this up as a whole thing.

PR #2299, opened 2025-12-27
fix(examples): fix import for widget examples

coding-retro:
Hello I was testing a bunch of the examples trying to learn how to make a TUI when I realized some of them didn't work! some of the usings are a bit outdated and are trying to use the ratatui_widgets from a third party crate but the features we are using here are already implemented in ratatui with ratatui::widgets. and that works fine from the testing I did simple change to make the examples more user friendly for new people trying to learn!


joshka:
https://github.com/ratatui/ratatui/pull/2300 fixes the changelog problem

PR #2306, opened 2025-12-29
fix: Fix the dependency on time

asomers:
ratatui-widgets uses time's Month::length(), which requires time-0.3.37 or later.


joshka:
Thanks

PR #2308, opened 2025-12-29
fix(buffer): diff for trailing cells when only style changes

### LINKED ISSUE #2307: Incorrect render when block is rendered over a widget that had a foreground color
## Description

Visual artifact bug where block borders would appear offset when rendered over a widget that had a foreground color style applied to the entire area.

## To Reproduce

```
git clone https://github.com/gcavelier/ratatui-render-bug.git
cd ratatui-render-bug
cargo run             # render is ok
cargo run -- color    # render is broken
```

## Expected behavior

Block borders should always be aligned (the bug is not present if the first rendered widget has no foreground color)

## Screenshots

When running `cargo run`
Image

When running `cargo run -- color`
Image

## Environment 1
- OS: Mac OS 26.2
- Terminal Emulator: Ghostty
- Font: CodeNewRoman Nerd Font Mono
- Crate version: 0.30
- Backend: default (crossterm)

## Environment 2
- OS: Windows 11 Pro 24H2 (WSL2)
- Terminal Emulator: Windows Terminal
- Font: CodeNewRoman Nerd Font Mono
- Crate version: 0.30
- Backend: default (crossterm)

## Additional context

The following has been generated by Claude and validated by me and @alegeay

### Root Cause Analysis

The bug occurs in `Buffer::diff()` when processing wide characters containing VS16 (U+FE0F):

```rust
// buffer.rs:526 (before fix)
if !next_trailing.skip && prev_trailing != next_trailing {
    updates.push((tx, ty, next_trailing));
}
```

### The Problem

The comparison `prev_trailing != next_trailing` compares the **entire cell** including:
- Symbol
- Foreground color
- Background color
- Modifiers
- Skip flag

However, the code comment (lines 505-509) states the intent is to clear cells that "**previously contained non-blank content**" - referring only to the **symbol**, not the style.

### Why This Causes Visual Artifacts

1. First widget applies `fg(Color::LightBlue)` to all cells in the area
2. `Clear` resets cells to `Cell::EMPTY` (fg = `Color::Reset`)
3. Block renders emoji `⚠️` (width 2) - trailing cell is reset to space
4. `Buffer::diff` compares trailing cells:
   - Previous: `symbol=" "`, `fg=Color::Reset`
   - Current: `symbol=" "`, `fg=Color::Gray`
5. Cells are **different** (due to style) → unnecessary update emitted
6. Terminal cursor moves to hidden trailing cell → visual offset

gcavelier:
this PR closes #2307 by preventing unnecessary diff updates for trailing cells when only style changes.

This PR was generated by Claude, and validated by me.

## Summary

This PR fixes a visual artifact bug where block borders would appear offset when rendered over a widget that had a foreground color style applied to the entire area.

## The Fix

```diff
- if !next_trailing.skip && prev_trailing != next_trailing {
+ // Only emit update if the SYMBOL changed, not just the style.
+ // The style of hidden trailing cells is not visible, so style
+ // differences alone should not trigger updates that can cause
+ // cursor positioning issues on some terminals.
+ if !next_trailing.skip && prev_trailing.symbol() != next_trailing.symbol() {
```

This aligns the code with the documented intent: only emit updates when the **symbol** (visible content) changes, not when only the style changes.

## Changes

| File | Change |
|------|--------|
| `ratatui-core/src/buffer/buffer.rs:526-530` | Compare only symbol, not full cell |
| `ratatui-core/src/buffer/buffer.rs:1376-1425` | Add regression test |

## Test Added

```rust
#[test]
fn diff_ignores_style_only_changes_in_trailing_cells() {
    // Verifies that trailing cells with same symbol but different style
    // do NOT generate diff updates
}
```

## Why This Is Safe

1. **Trailing cells are hidden** - they are visually covered by the wide character
2. **Style is invisible** - the fg/bg color of a hidden cell has no visual effect
3. **Symbol changes still trigger updates** - if the symbol changes (e.g., from `" "` to `"x"`), the update is still emitted
4. **Aligns with documented intent** - the original comment says "non-blank content", not "different style"

## Related

- The existing test `diff_clears_trailing_cell_for_wide_grapheme` verifies that symbol changes DO trigger updates
- This fix complements that behavior by ensuring style-only changes do NOT trigger updates

### DIFF
--- a/ratatui-core/src/buffer/buffer.rs
+++ b/ratatui-core/src/buffer/buffer.rs
@@ -539,7 +539,13 @@ impl Buffer {
                                 }
                                 let prev_trailing = &previous_buffer[j];
                                 let next_trailing = &next_buffer[j];
-                                if prev_trailing != next_trailing {
+                                // Only emit update if the symbol has changed.
+                                // The style of hidden trailing cells is not visible, so style
+                                // differences alone should not trigger updates that can cause
+                                // cursor positioning issues on some terminals.
+                                if next_trailing.diff_option != CellDiffOption::Skip
+                                    && prev_trailing.symbol() != next_trailing.symbol()
+                                {
                                     let (tx, ty) = self.pos_of(j);
                                     // Push an explicit update for the trailing cell.
                                     // This is expected to be a blank cell, but we use the actual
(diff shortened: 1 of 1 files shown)


ratatui-core/src/buffer/buffer.rs
```diff
@@ -525,7 +525,11 @@ impl Buffer {
                         }
                         let prev_trailing = &previous_buffer[j];
                         let next_trailing = &next_buffer[j];
-                        if !next_trailing.skip && prev_trailing != next_trailing {
+                        // Only emit update if the SYMBOL changed, not just the style.
+                        // The style of hidden trailing cells is not visible, so style
+                        // differences alone should not trigger updates that can cause
+                        // cursor positioning issues on some terminals.
+                        if !next_trailing.skip && prev_trailing.symbol() != next_trailing.symbol() {
```
orhun: I'm not sure if this comment is correct.

Before, we were comparing all the properties of cells:

> prev_trailing != next_trailing

But now we only check the symbol:

> prev_trailing.symbol() != next_trailing.symbol()

But the comment says "not just the style."

orhun:
Looks good to me overall, thanks for adding the test case for regression.

I couldn't reproduce this visually though, see my comment in the issue.

ratatui-core/src/buffer/buffer.rs
gcavelier: I changed the comment so that it's less confusing.

ratatui-core/src/buffer/buffer.rs
```diff
@@ -525,7 +525,11 @@ impl Buffer {
                         }
                         let prev_trailing = &previous_buffer[j];
                         let next_trailing = &next_buffer[j];
-                        if !next_trailing.skip && prev_trailing != next_trailing {
+                        // Only emit update if the SYMBOL changed.
```
orhun: nit:
```suggestion
                        // Only emit update if the symbol has changed.
```

orhun (approved):

ratatui-core/src/buffer/buffer.rs
gcavelier: done! (I also changed the commit message)

junkdog:
i ran the provided test case against the release commits of 0.29.0 and 0.30.0;  the bug is present in the 0.30.0 release, but not in 0.29.0.

orhun:
@junkdog have you also checked the changes in this PR?

should we possibly include it in your latest PR about `diff_iter` impl?

junkdog:
> @junkdog have you also checked the changes in this PR?

yes, lgtm. unrelated - but it was extremely confusing locally - as rustrover decided to render ⚠️ as ⚠︎, which, similarly confusing, stayed as such when it was copied to the clipboard  :sweat_smile: 

> should we possibly include it in your latest PR about `diff_iter` impl?

yeah, merge it and i'll integrate the fix into #2416

PR #2309, opened 2025-12-30
feat: add comment for inner area to popup example

Its-Just-Nans:
Add a comment about the inner popup area

### DIFF
--- a/examples/apps/popup/src/main.rs
+++ b/examples/apps/popup/src/main.rs
@@ -11,7 +11,7 @@
 use color_eyre::Result;
 use crossterm::event::{self, KeyCode};
 use ratatui::Frame;
-use ratatui::layout::{Constraint, Flex, Layout, Rect};
+use ratatui::layout::{Constraint, Layout};
 use ratatui::style::Stylize;
 use ratatui::text::Line;
 use ratatui::widgets::{Block, Clear, Paragraph};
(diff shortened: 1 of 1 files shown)


examples/apps/popup/src/main.rs
```diff
@@ -57,6 +57,9 @@ fn render(frame: &mut Frame, show_popup: bool) {
         // clears out any background in the area before rendering the popup
         frame.render_widget(Clear, popup_area);
         frame.render_widget(popup, popup_area);
+        // to add something inside the popup you can use the inner area of the block
+        // let inner_area = popup.inner(popup_area);
+        // frame.render_widget(your_widget, inner_area);
```
j-g00da: I think we can modify the actual code here, e.g. add a paragraph with a short lorem ipsum

Its-Just-Nans: Good idea, fixed

Its-Just-Nans:
Here is the output vhs gif

popup

Note that I kept the comment about the inner area with the block because it would have been useful to me.

Thanks!

examples/apps/popup/src/main.rs
```diff
@@ -52,11 +52,15 @@ fn render(frame: &mut Frame, show_popup: bool) {
     frame.render_widget(Block::bordered().title("Content").on_blue(), content);
 
     if show_popup {
-        let popup = Block::bordered().title("Popup");
+        let popup_block = Block::bordered().title("Popup");
         let popup_area = centered_area(area, 60, 20);
```
joshka: how about using the 0.30 centering method on rect here https://docs.rs/ratatui/latest/ratatui/prelude/struct.Rect.html#method.centered

joshka (approved):

examples/apps/popup/src/main.rs
Its-Just-Nans: fixed

popup

joshka (approved):
LGTM - thanks for the PR

PR #2312, opened 2025-12-31
docs: Update Terminal docs

joshka:
## Summary

- Improve `Terminal` Rustdocs to better explain typical app setup, the rendering pipeline, and how
  diff-based rendering works (including full redraw behavior on viewport size changes).
- Clarify viewport concepts and behavior in `Viewport`/`Terminal` docs, with a more complete inline
  section (anchoring, scrolling, resize behavior).
- Reorder `Terminal::draw` / `Terminal::try_draw` closer to constructors so the primary rendering
  entry points are easier to find.

## Notes

- Docs-only change; no public API or runtime behavior changes.
- No new tests in this PR.

## Test Plan

- `cargo +nightly fmt`
- `cargo +nightly docs-rs -p ratatui-core`
- `cargo test -p ratatui-core --doc --features std`

PR #2315, opened 2026-01-01
test(core): Split `Terminal` into submodules and expand test coverage

joshka:
- Split the `Terminal` implementation into focused submodules to improve readability and maintainability.
- Add characterization tests covering `Terminal` initialization, buffer lifecycle, resizing and autoresize behavior, and rendering paths.
- Add inline viewport tests for `compute_inline_size` and `insert_before` in both fallback and scrolling-regions modes, including an end-to-end `draw -> insert_before -> draw scenario` scenario.
- Extend `TestBackend` cursor plumbing to support the new terminal tests and assert cursor/ behavior.

### DIFF
--- a/ratatui-core/src/terminal/backend.rs
+++ b/ratatui-core/src/terminal/backend.rs
@@ -36,3 +36,40 @@ impl<B: Backend> Terminal<B> {
         self.backend.size()
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use crate::backend::TestBackend;
+    use crate::layout::{Position, Size};
+    use crate::terminal::Terminal;
+
+    #[test]
+    fn backend_returns_shared_reference() {
+        let backend = TestBackend::new(3, 2);
+        let terminal = Terminal::new(backend).unwrap();
+
+        assert_eq!(terminal.backend().cursor_position(), Position::ORIGIN);
+    }
+
+    #[test]
+    fn backend_mut_allows_mutating_backend_state() {
+        let backend = TestBackend::new(3, 2);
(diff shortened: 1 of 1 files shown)


ratatui-core/src/terminal/buffers.rs
```diff
@@ -0,0 +1,304 @@
...
+                for y in area.top()..area.bottom() {
+                    // TODO: Fixed viewports can start at x > 0 and have a limited width. Clearing
+                    // from x = 0 clears outside the viewport. Consider clearing only within
+                    // `viewport_area` (respecting both x offset and width) when fixing clear
+                    // semantics.
+                    self.backend.set_cursor_position(Position { x: 0, y })?;
+                    // TODO: `ClearType::AfterCursor` is exclusive of the cursor cell in
+                    // `TestBackend`, so the first cell of each cleared row can remain. Consider a
+                    // clear mode that includes the cursor cell when fixing clear semantics.
+                    self.backend.clear_region(ClearType::AfterCursor)?;
```
joshka, line 131: Note - this testing uncovered some existing bugs in this behavior - which should be fixed in a followup.

joshka:
Note - produced with Codex, validated each test as generally good. Found some bugs, but intentionally did not fix as by adding tests we get to fix the bugs and show that the tests will also change.

This builds on top of #2312 and the merge target is set there. Merge that first, then update this PR to point at main before merging.

kdheepak:
The PR description has some characters that don't render for me (Firefox, MacOS):

image

kdheepak (approved):

PR #2316, opened 2026-01-01
feat(examples): allow overlap spacing in explorer

joshka:
Store spacing as i16 so negative values map to Spacing::Overlap, and show overlap in the axis label.


joshka:
image

orhun (approved):

PR #2320, opened 2026-01-01
fix: Align clear() semantics with contract

joshka:
This change tightens the clearing API contract and fixes behavior mismatches that could confuse backend implementers and users.

Problems addressed:
- ClearType documentation did not define an observable, backend-agnostic contract, leading to ambiguity (especially around whether clearing includes the cursor cell and whether the cursor is preserved).
- TestBackend treated AfterCursor/BeforeCursor as exclusive of the cursor cell, while real backends that use CSI J clear inclusive of the cursor cell.
- Fixed viewports could clear outside their own region when the viewport was not full width, because clearing was done with AfterCursor from column 0.

What changed:
- Documented ClearType, clear_region, and clear with a precise, behavior-first contract: clears are by character cell in the visible display area, cursor position is preserved, and the cursor cell is included when it lies within the region.
- Aligned TestBackend to the inclusive contract for AfterCursor/BeforeCursor, matching real backend behavior.
- Updated Terminal::clear for fixed viewports to clear only the viewport region, selecting efficient strategies when the viewport spans the full width or reaches the bottom of the display.
- Restored the original cursor position after Terminal::clear to honor the contract.
- Refined inline/fixed viewport tests to be explicit and easy to read, covering full-width bottom, full-width non-bottom, and non-full-width paths.
- Added brief rustdoc comments on the fixed-viewport clear helpers.

Notes:
- Inline viewport clear still clears from the viewport origin to the end of the display; this is documented as current behavior but not part of the contract.


ratatui-core/src/backend/test.rs
```diff
@@ -288,12 +288,12 @@ impl Backend for TestBackend {
...
             ClearType::All => return self.clear(),
             ClearType::AfterCursor => {
-                let index = self.buffer.index_of(self.pos.0, self.pos.1) + 1;
+                let index = self.buffer.index_of(self.pos.0, self.pos.1);
                 &mut self.buffer.content[index..]
             }
             ClearType::BeforeCursor => {
                 let index = self.buffer.index_of(self.pos.0, self.pos.1);
-                &mut self.buffer.content[..index]
+                &mut self.buffer.content[..=index]
```
joshka, line 296: This is the main change for TestBackend - aligns this with how terminals work.

ratatui-core/src/terminal/buffers.rs
```diff
@@ -101,41 +101,77 @@ impl<B: Backend> Terminal<B> {
...
+        let original_cursor = self.backend.get_cursor_position()?;
         match self.viewport {
             Viewport::Fullscreen => self.backend.clear_region(ClearType::All)?,
             Viewport::Inline(_) => {
                 self.backend
                     .set_cursor_position(self.viewport_area.as_position())?;
-                // TODO: `ClearType::AfterCursor` is exclusive of the cursor cell in `TestBackend`
-                // (and in terminals that interpret this as "after" rather than "from"), which can
-                // leave the viewport origin cell uncleared. Consider switching to a clear that
-                // includes the cursor cell when fixing clear semantics.
```
joshka, line 117: This was incorrect (based on an assumption that AfterCursor worked exclusive, not inclusive)

ratatui-core/src/terminal/buffers.rs
```diff
@@ -101,41 +101,77 @@ impl<B: Backend> Terminal<B> {
...
-                    self.backend.set_cursor_position(Position { x: 0, y })?;
-                    // TODO: `ClearType::AfterCursor` is exclusive of the cursor cell in
-                    // `TestBackend`, so the first cell of each cleared row can remain. Consider a
-                    // clear mode that includes the cursor cell when fixing clear semantics.
-                    self.backend.clear_region(ClearType::AfterCursor)?;
-                }
+                self.clear_fixed_viewport(area)?;
             }
         }
+        self.backend.set_cursor_position(original_cursor)?;
```
joshka, line 128: Ensuring that the cursor doesn't change is new behavior. We always intended this I think, but never made sure it was safe.

ratatui-core/src/terminal/buffers.rs
```diff
@@ -101,41 +101,77 @@ impl<B: Backend> Terminal<B> {
...
-                // includes the cursor cell when fixing clear semantics.
                 self.backend.clear_region(ClearType::AfterCursor)?;
             }
             Viewport::Fixed(_) => {
                 let area = self.viewport_area;
-                for y in area.top()..area.bottom() {
-                    // TODO: Fixed viewports can start at x > 0 and have a limited width. Clearing
-                    // from x = 0 clears outside the viewport. Consider clearing only within
-                    // `viewport_area` (respecting both x offset and width) when fixing clear
-                    // semantics.
```
joshka, line 126: This is fixed below

ratatui-core/src/terminal/buffers.rs
```diff
@@ -101,41 +101,77 @@ impl<B: Backend> Terminal<B> {
     /// - [`Viewport::Inline`]: clears after the viewport's origin, leaving any content above the
     ///   viewport untouched.
     ///
+    /// Current behavior: for [`Viewport::Inline`], clearing runs from the viewport origin through
+    /// the end of the visible display area, not just the viewport's rectangle. This is an
+    /// implementation detail rather than a contract; do not rely on it.
```
joshka, line 106: I'm not sure if we should be doing this, but it's what we currently do. We can evaluate it in future if it causes some problem.

ratatui-core/src/terminal/buffers.rs
```diff
@@ -222,83 +258,123 @@ mod tests {
...
+            "before 1  ",
+            "before 2  ",
+            "viewport 1",
+            "viewport 2",
+            "after 1   ",
+            "after 2   ",
+        ]);
+        backend
+            .set_cursor_position(Position { x: 2, y: 2 })
+            .unwrap();
```
j-g00da, line 274: A side note: maybe we should consider refactoring our tests to use [testresult](https://docs.rs/testresult/latest/testresult/) instead of unwraps, for easier debugging. (In a different PR)

ratatui-core/src/terminal/buffers.rs
j-g00da: Also not really a fan of this, but yeah I guess we can always reevaluate if needed.

ratatui-core/src/terminal/buffers.rs
joshka: Neat. I think I read that way back and didn't understand the need. But it's much better and we should use it.

PR #2323, opened 2026-01-04
feat(ratatui-crossterm): add IntoCrossterm<ContentStyle> for Style

0xferrous:

### DIFF
--- a/ratatui-crossterm/src/lib.rs
+++ b/ratatui-crossterm/src/lib.rs
@@ -423,6 +423,79 @@ impl IntoCrossterm<CrosstermColor> for Color {
     }
 }
 
+impl IntoCrossterm<ContentStyle> for Style {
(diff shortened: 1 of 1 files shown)


joshka:
Seems reasonable, but can you let us know why this is needed.

Every thing we add like this is something that we have to think about longer term when things change upstream etc. or things change in Ratatui where we can no longer support a certain use case like this one. I think that's unlikely to happen here, but would be great to understand the why so we can come back to that if needed.

0xferrous:
I have some shared ratatui code where I want to switch between an interactive TUI or just printing to terminal, so I use this to convert to crossterm style and print directly with crossterm. Like here https://github.com/0xferrous/traces-tui/blob/main/crates/cli/src/main.rs#L72-L82

joshka:
Sweet - now that you point it out, I actually need that exact use case in Codex too :)

PR #2335, opened 2026-01-07
chore: Remove `@` sign from github release and PR

kdheepak:
This PR removes the `@` for GitHub usernames in the autogenerated PR body by `release-plz`.

```tera
{{ changelog | replace(from=" by @", to=" by ") }}
```

The motivation behind this is that GitHub mentions in PR descriptions trigger notifications for every contributor for every release because their username is part of the changelog.

With this change, in GitHub PR descriptions the changelog will change like so:

**Before**

```
1dc18bf (calendar) Add width and height functions by @joshka in #2198
```

**After**

```
1dc18bf (calendar) Add width and height functions by joshka in #2198
```

Ideally this would be implemented using backticks (`@joshka` does not trigger a notification), but tera doesn't support regex replacements, so this is the simplest solution for now.

### DIFF
--- a/release-plz.toml
+++ b/release-plz.toml
@@ -3,3 +3,7 @@ changelog_path = "./CHANGELOG.md"
 changelog_config = "cliff.toml"
 release_always = false
 semver_check = false
+
+git_release_body = """
+{{ changelog | replace(from=" by @", to=" by ") }}
+"""


orhun:
> Ideally this would be implemented using backticks (@joshka does not trigger a notification), but tera doesn't support regex replacements, so this is the simplest solution for now.

Instead of updating `release-plz` config, we can achieve this by updating `cliff.toml` instead:

```diff
diff --git a/cliff.toml b/cliff.toml
index 3b20b647..ec8c969e 100644
--- a/cliff.toml
+++ b/cliff.toml
@@ -34,7 +34,7 @@ body = """
 {% macro commit(commit) -%}
 - [{{ commit.id | truncate(length=7, end="") }}]({{ "https://github.com/ratatui/ratatui/commit/" ~ commit.id }}) \
   *({{commit.scope | default(value = "uncategorized") | lower }})* {{ commit.message | upper_first | trim }}\
-  {% if commit.remote.username %} by @{{ commit.remote.username }}{%- endif -%}\
+  {% if commit.remote.username %} by `@{{ commit.remote.username }}`{%- endif -%}\
   {% if commit.remote.pr_number %} in [#{{ commit.remote.pr_number }}]({{ self::remote_url() }}/pull/{{ commit.remote.pr_number }}){%- endif %}\
 {%- if commit.breaking %} [**breaking**]{% endif %}
 {%- if commit.body %}\n\n{{ commit.body | indent(prefix="  > ", first=true, blank=true) }}
```

kdheepak:
I'm fine with that change. I thought we'd want to change just the github release instead of changing the changelog as well.

kdheepak:
Possibly overkill but we can probably do something like this in `release-plz` too:

```
git_release_body = """
{% set_global body = changelog %}
{% if remote.contributors %}
  {% for c in remote.contributors %}
    {% set_global body = body | replace(
        from="by @" ~ c.username ~ " in #",
        to="by @`" ~ c.username ~ "` in #"
    ) %}
  {% endfor %}
{% endif %}
{{ body }}
"""
```

Though, I'm not able to test it locally with `--dry-run`.

kdheepak:
I made an alternative PR that makes the changelog change in https://github.com/ratatui/ratatui/pull/2336. Only one of these PRs should be merged.

joshka:
Merged the other PR

PR #2336, opened 2026-01-07
chore: escape username using backticks in changelog

kdheepak:
This PR escapes GitHub usernames with backticks for the changelog generated by `git-cliff`:

```tera
  {% if commit.remote.username %} by `@{{ commit.remote.username }}`{%- endif -%}\
```

The motivation behind this is that GitHub mentions in PR descriptions trigger notifications for every contributor for every release because their username is part of the changelog.

With this change, in GitHub PR descriptions the changelog will change like so:

**Before**

```
1dc18bf (calendar) Add width and height functions by @joshka in #2198
```

**After**

```
1dc18bf (calendar) Add width and height functions by `@joshka` in #2198
```


kdheepak:
This PR is an alternative to https://github.com/ratatui/ratatui/pull/2335. Only one of these PRs should be merged.

PR #2343, opened 2026-01-12
feat(table): add support for configuring borders

### LINKED ISSUE #2: Make Tables support borders
This issue is marked as a "Good First Issue". I'd ask anyone considering working on this to please first consider the open questions mentioned below, and to design the user facing API in this issue to validate it upfront before going forward with an implementation.

## Problem

```
┌─┬─┐
│A│B│
├─┼─┤
│1│2│
├─┼─┤
│3│4│
└─┴─┘
```
```
A│B
─┼─
1│2
3│4
```
And
```
A B
───
1 2
3 4
```
And
```
A│B
1│2
3│4
```
And possible even:
```
┌─┬─┐
│A│B│
╞═╪═╡ 
│1│2│
├─┼─┤
│3│4│
└─┴─┘
```
## Solution

## Additional context
For some inspiration:

Rich (Python)
rich table

Bubble Tea (Go):
bubbletea table

### LINKED ISSUE #604: Make Tables support borders

ThomasBonnanfant:
## Description
This PR adds the ability to configure borders on the Table widget. 
It introduces a new configuration to draw borders around the table (All, Vertical, None, etc.), using standard Unicode box-drawing characters.

## Demo
(Glisse et dépose ton image ici ! C'est vital pour prouver que ça marche)

## Related Issue
Closes ratatui/ratatui-table#2
image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,6 +41,7 @@ jobs:
         with:
           tool: taplo-cli
       - run: cargo xtask format --check
+        shell: bash
 
   # Check for typos in the codebase.
   # See <https://github.com/crate-ci/typos/>
@@ -109,6 +110,7 @@ jobs:
           components: clippy
       - uses: Swatinem/rust-cache@779680da715d629ac1d338a641029a2f4372abb5 # v2
       - run: cargo xtask clippy
+        shell: bash
         env:
           RUSTUP_TOOLCHAIN: ${{ matrix.toolchain }}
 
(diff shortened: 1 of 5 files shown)


joshka:
The Table experiment now lives at https://github.com/ratatui/ratatui-table. The original borders discussion was transferred to https://github.com/ratatui/ratatui-table/issues/2, and focused successor tracking for this PR is at https://github.com/ratatui/ratatui-table/issues/9.

@ThomasBonnanfant, would you be interested in recreating this as a focused PR against the extracted 0.1.0 baseline? We would also like to offer you opt-in Write access so proposal authors can review and collaborate without waiting on the core Ratatui repository. Releases and protected configuration remain maintainer-controlled; reply here if you would like the invitation.

Issue #2086, opened 2025-09-09
Resize for inline viewport


aldanor:
## Description
Inline viewport resizing doesn't seem to behave well (as been also pointed out by @joshka in https://github.com/ratatui/ratatui/pull/1964#issuecomment-3138834123 in the context of inline viewport resizing PR).

## To Reproduce
Run the following and resize the terminal horizontally:
```rust
use color_eyre::Result;
use ratatui::{
    DefaultTerminal, Frame, TerminalOptions, Viewport,
    crossterm::event::{self, Event, KeyCode},
    widgets::{Block, Paragraph},
};

fn main() -> Result<()> {
    color_eyre::install()?;
    let terminal = ratatui::init_with_options(TerminalOptions {
        viewport: Viewport::Inline(8),
    });
    let app_result = run(terminal);
    ratatui::restore();
    app_result
}

fn run(mut terminal: DefaultTerminal) -> Result<()> {
    loop {
        terminal.draw(hello_world)?;
        match event::read()? {
            Event::Key(key) => {
                if key.kind == event::KeyEventKind::Press && key.code == KeyCode::Char('q') {
                    break;
                }
            }
            Event::Resize(_w, _h) => terminal.autoresize()?,
            _ => (),
        }
    }
    Ok(())
}

fn hello_world(frame: &mut Frame) {
    frame.render_widget(
        Paragraph::new("Hello World!").block(Block::bordered().title("Greeting")),
        frame.area(),
    );
}
```

## Screenshots
If run within zellij (which is run within ghostty):
<img width="75%"  alt="Image" src="https://github.com/user-attachments/assets/0ba6ac0e-1ad4-40c3-b44f-436c74877a3e" />

If run straight within ghostty:
<img width="75%" alt="Image" src="https://github.com/user-attachments/assets/857916c5-7bdf-4a9b-92ba-8fdd04a67070" />

## Environment
- OS: macOS
- Terminal Emulator: ghostty
- Crate version: latest main
- Backend: crossterm

wyvernbw:
This looks like its caused by the previous draw being one character too wide and wrapping on to the next line. if i squeeze my terminal from 128 columns to 120 columns and then back to 128 (you can use the ghostty terminal inspector to see) i get this output
```
┌Greeting───────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│Hello World!                                                                                                           │
│                                                                                                                       │
│                                                                                                                       │
[...]

PR #2349, opened 2026-01-14
fix: resizing for inline viewport

wyvernbw:
https://private-user-images.githubusercontent.com/50150016/535666090-66dd8ce2-5dd5-4d8d-ab95-57d615354d59.mov?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3Njg0MTExNDEsIm5iZiI6MTc2ODQxMDg0MSwicGF0aCI6Ii81MDE1MDAxNi81MzU2NjYwOTAtNjZkZDhjZTItNWRkNS00ZDhkLWFiOTUtNTdkNjE1MzU0ZDU5Lm1vdj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNjAxMTQlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjYwMTE0VDE3MTQwMVomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTAzNGFlODIwZDgzMWIzOTRkMDRjODdmMTRkMjk1ZGY0ODAxNmM2MTA4YmMwM2Q4OTQ3MmEyYTllM2NiMDg3NzgmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0In0.Tkm-5a6FPS1tZ9YCn74T7PwZqivOODV04ilfTbc1DwM

Fixes horizontal resizing when using inline viewports (#2086). Here's a breakdown of the issue from one of my previous comments:

> when you shrink the terminal horizontally, the inline viewport cannot fit into
the lines anymore and the lines wrap. This is done by the terminal emulator. So
all of a sudden there is a lot of extra vertical space, so the terminal emulator
has to make a choice: do we scroll the previous content up, making room for
the bottom line, or do we let the extra space overflow and have the user scroll
down? With the way most terminals work, the first option is what really happens.
So, when the terminal shrinks, all the previous content gets scrolled up. The
viewport area remains the same. So to recap:
> 
> 1. User resizes terminal
> 2. Terminal wraps the lines as they dont fit anymore
> 3. Content gets pushed UP
> 4. Ratatui draws the viewport in the old position
>
> So, theres two ways to fix this: either we scroll the content back down, or we
> move the inline viewport up. They both have their pros and cons

# Changes

- Fixes the resizing logic refrenced in https://github.com/ratatui/ratatui/pull/1964#issuecomment-3138834123 
- Adds a "counter scroll" to keep the inline viewport in the correct place when lines wrap [in resize function](https://github.com/wyvernbw/ratatui/blob/c3955f4a38c8a1f0cb2a39c58673e0b10091a9cf/ratatui-core/src/terminal/resize.rs#L35-L54)

# Todo

- [ ] documentation
- [ ] we might want to add this logic somewhere else, not the resize function, but im not very well versed in the architecture

# Footnotes

This is NOT a comprehensive implementation of inline viewport resizing, it just aims to fix the current horizontal shrink behavior and provide some better defaults, especially since what is in place now is completely broken.

No AI code in this PR.

### DIFF
--- a/ratatui-core/src/terminal/resize.rs
+++ b/ratatui-core/src/terminal/resize.rs
@@ -31,6 +31,36 @@ impl<B: Backend> Terminal<B> {
                 )?
                 .0;
 
+                // inline viewport rendering breaks when shrinking the terminal
+                // window due to line wrap. line wrap causes the height of the
+                // viewport to temporarily increase, such that the bottom line
+                // remains at the same position and everything else scrolls up.
+                //
+                // the way this counteracts that is by conditionally moving the
+                // viewport area up by an offset. the offset can be calculated
+                // as:
+                //
+                // `offset = content_area / area_width - space_left`
+                //
+                // the space left is just the terminal height minus the position
+                // of the viewport and minus the height of the viewport. The
+                // content area is the area of actually filled cells. The
+                // reason we need this is because the terminal will not wrap
+                // lines if it has enough room for them. applying an offset in
+                // that case would be wrong. spaces inbetween actual characters
+                // count towards line wrap (the terminal will not just compress
+                // lines), so a line like
+                //
+                // `    [      ]    `
+                //  0123456789...
+                //      ^------^ this is incompressible for the terminal
+                //    left   right --  is what we must find
+                //
+                // is 8 characters in area. we calculate the area by integrating
+                // over height, summing up the area of all lines. This is O(n*m)
+                // where n and m are the dimensions of the terminal, but in practice
+                // it would be closer to O(n) if you fill your viewport with a block
+                // because finding the start and end is just 2 checks.
                 if area.width < self.last_known_area.width {
                     let row_width = self.last_known_area.width as usize;
                     // calculate area of non empty content in buffer
@@ -39,14 +69,11 @@ impl<B: Backend> Terminal<B> {
                         .content()
                         .chunks(row_width)
                         .map(|row| {
-                            let left = row
-                                .iter()
-                                .position(|cell| cell.symbol() != " ")
-                                .unwrap_or(0);
-                            let right = row
-                                .iter()
-                                .rposition(|cell| cell.symbol() != " ")
-                                .unwrap_or(row.len());
+                            let left = row.iter().position(|cell| cell.symbol() != " ");
+                            let Some(left) = left else {
+                                return row.len() + 1;
+                            };
+                            let right = row.iter().rposition(|cell| cell.symbol() != " ").unwrap();
                             right.saturating_sub(left) + 1
                         })
                         .sum::<usize>();


wyvernbw:
Im setting this as draft because it seems I somehow broke it

wyvernbw:
Another interesting solution is to disable line wrap entirely when using inline viewport. This looks flawless but doesnt work in tmux and needs backend support (crossterm at least has support through `DisableLineWrap`)

Also if you shrink the window and make it bigger your history will be cut off horizontally so not ideal

wyvernbw:
Okay so, when using move up strategy ~~and EITHER `EnableLineWrap` or `DisableLineWrap`~~, it literally looks perfect (though disabling the line wrap has the problems i mentioned) **EDIT: this only applies to ghostty, and it works with or without line wrap**. Neither works in tmux though. zellij works better than tmux and i would say its usable here, but not perfect. 

The problem is, it ONLY works in ghostty. And it is actually strange how good it looks. But every other terminal ive tried does not work with this. Not a single one. 

This problem is very much terminal emulator dependent. I'm starting to think that making a good cross platform solution might not be possible.

anyways here is ghostty:

https://github.com/user-attachments/assets/f0ad7e58-dd47-4112-9f63-f4cc8868e299

I thought that we should include a way to disable line wrap in the backend, since that sidesteps this problem completely, regardless of terminal emulator. BUT, it seems that out of ghostty, kitty, iterm2, wezterm, rio ONLY GHOSTTY correctly implements `\033[?7l`, and even its implementation is a little half assed. I don't think there is a way forward here. Feel free to use this pr if you want, works flawlessly with ghostty and its a better default for other terminals (the tui will jump up to the top and render normally instead of printing a lot of garbage, so its something at least) and multiplexers are still broken.

orhun:
Wow thanks for your work and investigation on this! It's definitely valuable.

Kinda sad to hear that this behavior is dependent of how the terminal protocol is implemented. Although it seems like there might not be many things we could do on Ratatui side, I still see this PR as an improvement. Maybe the next step would be to report this to related upstreams (e.g. zellij, kitty, wezterm, etc.)

I haven't tested this thoroughly yet but will get to it soon.

wyvernbw:
Thanks, one thing i havent taken into account is padding around the things rendered. the current implementation estimates the height the viewport will jump to, assuming what youre rendering takes up the whole viewport. having a smaller block thats centered for example breaks the illusion. i imagine there isnt a very performant way to do this though

Also the changes in `buffer.rs` are optional, they would mostly help with occasional flickers and wrong content between resize and redraw

wyvernbw:
I fixed the padding issue (though the implementation is a O(n) algorithm where n is the number of cells in the buffer so that might not be ideal, it is easily parallelizable if thats something we would need?), here is how it looks with a more involved example (notice how it breaks when resizing quickly):

https://github.com/user-attachments/assets/3890f720-fc31-4477-8fcd-7753e60c054b

wyvernbw:
i went ahead and installed gemini-cli to see how ink handles this problem and sure enough, there is no magic fix, it clears the screen if the width becomes smaller. We could adopt that strategy or use this PR's approach, it's up to you. Alternatively we could support both with a flag in `TerminalOptions`.

I'm willing to work on the implementation (likely in a separate PR) when it's decided :)

orhun:
I would prefer the fix for this to be as simple as possible. Clearing seems more straightforward than truncating the rows to me, but can't say something for certain until I see the implementation. Supporting both options seems a bit too much for now.

wyvernbw:
I agree, anything other than clearing is too finicky imo, not to mention terminal emulator dependent

orhun:
Maybe update this PR with that approach or create another draft one?

wyvernbw:
I'll just make a new one since this one has a lot of cruft 

(EDIT) Done: #2355

PR #2356, opened 2026-01-17
feat: add custom marker

BenFradet:
## What

- add additional markers namely `×` and `+`

## Why

- these are esp useful in a scatter plot context to differentiate data sources


ratatui-widgets/src/canvas/rectangle.rs
```diff
@@ -199,4 +171,66 @@ mod tests {
...
+        ]);
+        let rect = Rect::new(0, 0, 10, 10);
+        expected.set_style(rect, Style::new().red());
+        expected.set_style(rect.inner(Margin::new(1, 0)), Style::reset().red());
+        expected.set_style(rect.inner(Margin::new(1, 1)), Style::reset());
+        assert_rectangle(Marker::PlusSign, rect, expected);
+    }
+
+    #[allow(clippy::needless_pass_by_value)]
+    fn assert_rectangle(marker: Marker, rect: Rect, expected: Buffer) {
```
BenFradet: `expected: Buffer` triggered a false positive `needless_pass_by_value` warning here despite being used in `assert_eq`
hence the `allow`

ratatui-core/src/symbols/marker.rs
```diff
@@ -53,6 +55,10 @@ pub enum Marker {
     /// unicode that is less broadly supported than Braille dots. If your terminal does not support
     /// this, you will see unicode replacement characters (`�`) instead of octants (`𜴇`, `𜷀`, `𜴷`).
     Octant,
+    /// One point per cell in shape of x (`×`)
```
joshka: Probably better to do a `Custom` type that accepts a String here and make this generic.

ratatui-widgets/src/canvas/rectangle.rs
```diff
@@ -88,21 +88,6 @@ mod tests {
 
     #[test]
     fn draw_block_lines() {
-        let mut buffer = Buffer::empty(Rect::new(0, 0, 10, 10));
```
joshka, line 91: As a general rule, leave the assertion in the test as it makes it easier to understand what test failed from stack traces etc.).

joshka (changes requested):

ratatui-core/src/symbols/marker.rs
BenFradet: I went for a `Custom(char)` :+1:

ratatui-widgets/src/canvas/rectangle.rs
BenFradet: moved the assert back into the test

ratatui-core/src/symbols/marker.rs
```diff
@@ -53,6 +53,8 @@ pub enum Marker {
     /// unicode that is less broadly supported than Braille dots. If your terminal does not support
     /// this, you will see unicode replacement characters (`�`) instead of octants (`𜴇`, `𜷀`, `𜴷`).
     Octant,
+    /// Custom Marker where the supplied char is applied once per cell
```
orhun:
```suggestion
    /// Custom marker where the supplied char is applied once per cell.
```

ratatui-widgets/src/chart.rs
```diff
@@ -368,8 +368,9 @@ impl<'a> Dataset<'a> {
 
     /// Sets the kind of character to use to display this dataset
     ///
-    /// You can use dots (`•`), blocks (`█`), bars (`▄`), braille (`⠓`, `⣇`, `⣿`) or half-blocks
-    /// (`█`, `▄`, and `▀`). See [`symbols::Marker`] for more details.
+    /// You can use dots (`•`), x or + signs, (`×`, `+`), blocks (`█`), bars (`▄`), braille
```
orhun: I would prefer to leave this comment as-is and just add an additional sentence at the end (e.g. if you need custom chars use Marker::Custom etc.)

orhun (commented):
looks pretty good

ratatui-widgets/src/chart.rs
BenFradet: yup, artifact from the previous impl

ratatui-core/src/symbols/marker.rs
BenFradet: done

orhun (approved):

PR #2369, opened 2026-01-23
fix(widgets): inherit the text alignment for Paragraph

7Bpencil:
Paragraph didn't take into account alignment of the text it was created from:
```rs
let lines = vec![
    Line::from("one"),
    Line::from("double"),
    Line::from("quadruple"),
];
let text = Text::from(lines).centered();
let paragraph = Paragraph::new(text).block(block); // rendered left aligned
```

<details>

<summary>Demo source code:</summary>

```rs
use crossterm::event::{self, Event, KeyModifiers};
use ratatui::{
    layout::Rect,
    text::{Line, Text},
    widgets::{Block, Borders, Clear, Paragraph},
    DefaultTerminal, Frame,
};
use std::iter;

fn main() {
    ratatui::run(app);
}

fn app(terminal: &mut DefaultTerminal) {
    loop {
        if let Event::Key(key_event) = event::read().expect("failed to read event") {
            if key_event.kind.is_press()
                && key_event.modifiers.contains(KeyModifiers::CONTROL)
                && key_event.code.is_char('c')
            {
                break;
            }
        }
        terminal.draw(render).expect("failed to draw frame");
    }
}

fn render(frame: &mut Frame) {
    {
        let area = Rect::new(5, 2, 20, 5);
        let block = Block::default()
            .title_top(Line::from("Centered-Lines").centered())
            .borders(Borders::ALL);
        let lines = vec![
            Line::from("one").centered(),
            Line::from("double").centered(),
            Line::from("quadruple").centered(),
        ];
        let text = Text::from(lines);
        let paragraph = Paragraph::new(text).block(block);
        frame.render_widget(paragraph, area);
    }
    {
        let area = Rect::new(35, 2, 20, 5);
        let block = Block::default()
            .title_top(Line::from("Centered-Text").centered())
            .borders(Borders::ALL);
        let lines = vec![
            Line::from("one"),
            Line::from("double"),
            Line::from("quadruple"),
        ];
        let text = Text::from(lines).centered();
        let paragraph = Paragraph::new(text).block(block);
        frame.render_widget(paragraph, area);
    }
    {
        let area = Rect::new(65, 2, 20, 5);
        let block = Block::default()
            .title_top(Line::from("Centered-Paragraph").centered())
            .borders(Borders::ALL);
        let lines = vec![
            Line::from("one"),
            Line::from("double"),
            Line::from("quadruple"),
        ];
        let text = Text::from(lines);
        let paragraph = Paragraph::new(text).block(block).centered();
        frame.render_widget(paragraph, area);
    }
}
```

</details>

Before the fix:

Ca312310213pture

After the fix:

Captrtsrts02131231ure

### DIFF
--- a/.github/workflows/check-semver.yml
+++ b/.github/workflows/check-semver.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout the repository
-        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           persist-credentials: false
       - name: Check semver
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
     name: Check Formatting
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           persist-credentials: false
       - uses: dtolnay/rust-toolchain@f7ccc83f9ed1e5b9c81d8a67d7ad1a747e22a561 # master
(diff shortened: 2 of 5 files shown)


orhun (approved):

orhun:
Not sure if this would have any side effects. I would like to have more eyes on this before merging.

orhun:
Having another look at this, I realized this is a breaking change in terms of behavior.
For example, imagine you've created an app before this change with centered text. It will be left aligned if you use it in a paragraph. But after this change, the text will be centered.

So maybe we should document this in the BREAKING-CHANGES.md document?

P.S. I agree that inheriting the alignment is the behavior that makes sense.

junkdog:
> Having another look at this, I realized this is a breaking change in terms of behavior.
> For example, imagine you've created an app before this change with centered text. It will be left aligned if you use it in a paragraph. But after this change, the text will be centered.

> So maybe we should document this in the BREAKING-CHANGES.md document?

at the same time, the intent of that code would read as if the text should have been centered all along. this also aligns the behavior with `Paragraph::centered()` and `Line::centered()`, which to me makes it sound more like a bug fix than a breaking change.

junkdog (approved):
looking good

PR #2370, opened 2026-01-23
build: add symlinks from each crate dir to root LICENSE

martinvonz:
AFAICT, each crate directory must have its own LICENSE file for it to become part of the published crate. This patch therefore adds LICENSE symlinks from each crate directory to the root LICENSE file. This matches what e.g. clap does (https://github.com/clap-rs/clap).

Since ratatui-macros/LICENSE already exists and with different copyright holders than in the root LICENSE file, I left it unchanged.


martinvonz:
What's the next step? Am I supposed to do something?

orhun:
Nope, I just waiting for additional feedback :) Good to go.

PR #2388, opened 2026-02-08
refactor(scrollbar example): simplify render function (#975)

marianomarciello:
Divide the render function into two functions for vertical and horizontal scroll.

I have execute the `cargo xtask ci` and everything was working fine. I tried again to execute the example with the command `cargo run -p scrollbar` and the example looks exactly the same.

I think that splitting the code into two functions might help to find the code in an easier way if someone will need to use the vertical or the horizontal scroll bar.

### DIFF
--- a/ratatui-core/src/buffer.rs
+++ b/ratatui-core/src/buffer.rs
@@ -6,4 +6,4 @@ mod buffer;
 mod cell;
 
 pub use buffer::Buffer;
-pub use cell::Cell;
+pub use cell::{Cell, CellDiffOption};
--- a/ratatui-core/src/buffer/buffer.rs
+++ b/ratatui-core/src/buffer/buffer.rs
@@ -6,7 +6,7 @@ use core::{cmp, fmt};
 use unicode_segmentation::UnicodeSegmentation;
 use unicode_width::UnicodeWidthStr;
 
-use crate::buffer::Cell;
+use crate::buffer::{Cell, CellDiffOption};
 use crate::layout::{Position, Rect};
 use crate::style::Style;
 use crate::text::{Line, Span};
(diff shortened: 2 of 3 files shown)


examples/apps/scrollbar/src/main.rs
```diff
@@ -81,58 +82,19 @@ impl App {
...
-        self.horizontal_scroll_state = self.horizontal_scroll_state.content_length(long_line.len());
-
-        let create_block = |title: &'static str| Block::bordered().gray().title(title.bold());
-
+    fn render_vertical_scroll(
+        &mut self,
+        frame: &mut Frame,
+        text: &[Line],
+        chunks: &Rc<[Rect]>,
+        create_block: &dyn Fn(&'static str) -> Block<'static>,
```
orhun: I think we should extract this into a real function instead of passing a closure now.

Thinking from a beginner perspective, seeing this signature might be a little scary.

Other than that, everything LGTM! Thanks for making this change.

JayanAXHF: Also a nit: IMO it’s usually better to use a CoW instead of an &’static str, because it’s (afaik) strictly more flexible

orhun: We probably won't need that if we create a function.

marianomarciello:
Hi @orhun and @JayanAXHF , thanks for the review. I have addressed your comments and I believe now is even easier. If you could take a second look that would be amazing. Thanks for your time :)

examples/apps/scrollbar/src/main.rs
```diff
@@ -32,6 +33,10 @@ fn main() -> Result<()> {
     ratatui::run(|terminal| App::default().run(terminal))
 }
 
+fn create_block(title: &str) -> Block<'_> {
+    Block::bordered().gray().title(title.bold())
+}
```
orhun: Let's move this inside the `impl App` block and call it like `Self::create_block`

marianomarciello: Hey @orhun  thanks again for you review. I have create a new commit to move the create_block inside the `App`. 

😄  Thanks for revieweing this.

examples/apps/scrollbar/src/main.rs
```diff
@@ -201,4 +163,51 @@ impl App {
...
+
+        let text = vec![
+            Line::from("This is a line "),
+            Line::from("This is a line   ".red()),
+            Line::from("This is a line".on_dark_gray()),
+            Line::from("This is a longer line".crossed_out()),
+            Line::from(long_line.clone()),
+            Line::from("This is a line".reset()),
+            Line::from(vec![
+                Span::raw("Masked text: "),
```
JayanAXHF: Just occured to me that Span impls `Add` and outputs a `Line`. Maybe we could use that? Just a thought :)

marianomarciello: I think that's a nice suggestion, it makes the code more clean and also higlights what the add trait between Spans does.

I have uploaded a new commit for this :) Thanks for suggesting it.

PR #2394, opened 2026-02-09
ci: align cliff.toml with Keep a Changelog

### LINKED ISSUE #2354: Fix release-plz to not ping contributors on PR
Currently, release-plz's changelog generation in the desc. of a PR pings all people that have every contributed to ratatui. This is really annoying for a lot of people, and this issue tracks progress on this problem

ognis1205:
`release-plz` assumes that `CHANGELOG.md` follows the Keep a Changelog format.

When rendering the release PR body, it computes the new changelog entries for each package, inserts them into the workspace-level `CHANGELOG.md`, and then retrieves the latest changelog section per package during this step:

https://github.com/release-plz/release-plz/blob/ede5cd2b73f68e271ec083602dc5acb0d9692345/crates/release_plz_core/src/command/update/packages_update.rs#L98

However, the current `cliff.toml` configuration in Ratatui does not strictly follow the Keep a Changelog format. As a result, `release-plz` fails to correctly identify the latest changelog entries, and previously released sections can end up being
embedded into the release PR body.

This change updates `cliff.toml` to align with the Keep a Changelog format, so that `release-plz` can reliably extract only the newly added changelog entries for each release. This is achieved by encoding the package name as build metadata in the generated changelog entries.

Closes #2354 .

### DIFF
--- a/cliff.toml
+++ b/cliff.toml
@@ -25,7 +25,7 @@ body = """
 ## [unreleased]
 {% else -%}
 {%- if package -%} {# release-plz specific variable #}
-## {{ package }} - [{{ version }}]({{ release_link }}) - {{ timestamp | date(format="%Y-%m-%d") }}
+## [{{ version }}+{{ package }}]({{ release_link }}) - {{ timestamp | date(format="%Y-%m-%d") }}
 {%- else -%}
 ## [{{ version }}]({{ self::remote_url() }}/releases/tag/{{ version }}) - {{ timestamp | date(format="%Y-%m-%d") }}
 {%- endif %}


ognis1205:
Note:

If embedding the package name as build metadata in `CHANGELOG.md` is not desirable (i.e., for `CHANGELOG.md` generation workflow, not for `release-plz` workflow), it would be better to provide a dedicated `cliff.toml` configuration specifically for `release-plz`, instead of reusing the same changelog format.

orhun:
have you tried this locally with `release-plz` commands? what result did you get for the changelog?

ognis1205:
> have you tried this locally with `release-plz` commands? what result did you get for the changelog?

Please see this PR:

https://github.com/ognis1205/ratatui/pull/1

Sorry if this caused any unintended pings 🙏

ognis1205:
Note:
This is intended as a temporary workaround. I plan to investigate whether a cleaner solution can be implemented upstream in `release-plz`, but for now this addresses the immediate issue.

Given that `CHANGELOG.md` is manually generated via `git-cliff`, I think this approach is reasonable within the limited scope of fixing the release PR body behavior.

PR #2395, opened 2026-02-09
feat(widgets): add `Cell::width` and `Row::widths`

omelia-iliffe:
Adds two methods to help access the width of a Cells content.

I often want to have the width of my columns in a table correspond to the widest cell in the column. Without these `width` methods, I was having to store the width of the `Text` before wrapping it in `Cell` and `Row`.

With these methods I can now do this after creating my rows iterator.
```rust
let [w1, _, w3, w4] = rows.clone().fold([0; 4], |w, c| {
    w.into_iter()
        .zip(c.widths())
        .map(|(w, c)| w.max(c as u16) + 1)
        .collect_array()
        .unwrap()
});
```

I have added minimal documentation, happy to be pointed in the direction of where to add more. Or if any other changes are needed. :)

### DIFF
--- a/cliff.toml
+++ b/cliff.toml
@@ -25,7 +25,7 @@ body = """
 ## [unreleased]
 {% else -%}
 {%- if package -%} {# release-plz specific variable #}
-## {{ package }} - [{{ version }}]({{ release_link }}) - {{ timestamp | date(format="%Y-%m-%d") }}
+## [{{ version }}+{{ package }}]({{ release_link }}) - {{ timestamp | date(format="%Y-%m-%d") }}
 {%- else -%}
 ## [{{ version }}]({{ self::remote_url() }}/releases/tag/{{ version }}) - {{ timestamp | date(format="%Y-%m-%d") }}
 {%- endif %}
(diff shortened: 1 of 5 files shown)


omelia-iliffe:
rebased onto main sorry, for some reason I was using the `latest` branch

ratatui-widgets/src/table/row.rs
```diff
@@ -239,6 +239,11 @@ impl<'a> Row<'a> {
         self.style = style.into();
         self
     }
+
+    /// Get an iterator of the [`Cell`] widths
+    pub fn widths(&self) -> impl Iterator<Item = usize> {
```
orhun, line 244: Naming this `widths` feels a bit ambiguous. Also we usually have fluent setter functions so it might be confused with those (e.g. sth that sets the widths, e.g. https://docs.rs/ratatui/latest/ratatui/widgets/struct.Table.html#method.widths)

ratatui-widgets/src/table/cell.rs
```diff
@@ -173,6 +173,11 @@ impl<'a> Cell<'a> {
         self.style = style.into();
         self
     }
+
+    /// Get the width of the [`Cell`]
+    pub fn width(&self) -> usize {
+        self.content.width()
+    }
```
orhun, line 180: This is probably fine 👌🏼

junkdog:
hi, the width of `Cell` is already available as `Cell::cell_width`, it was part of #2400.

orhun:
I think I will close this one. We already have the cell width function and the other one seems quite specific

PR #2399, opened 2026-02-15
feat(core): support layout-cache in no_std environments

junkdog:
this enables "layout-cache" for no_std builds; it's meant for embedded environments, where the layout engine otherwise consumes all CPU, capping the framerate at around ~10fps. the same app can refresh 300-500 times per second with layout cache enabled.

i had to add `layout-cache = ["dep:critical-section"]` to all builds - it's pretty tiny and shouldn't leave a trace in std-builds. the alternative is to add an extra layer of features for layout-cache with std and no_std, but it pollutes the feature space.

### DIFF
--- a/ratatui-core/src/layout/layout.rs
+++ b/ratatui-core/src/layout/layout.rs
@@ -772,18 +772,21 @@ impl Layout {
                 .map(|(s, sp)| (Rc::from(s.as_slice()), Rc::from(sp.as_slice())))
         });
 
-        if let Some(result) = cached {
-            result
-        } else {
-            let result = self.split_layout(area);
-            critical_section::with(|cs| {
-                let mut cache = LAYOUT_CACHE.borrow(cs).borrow_mut();
-                if let Some(cache) = cache.as_mut() {
-                    let key = (area, self.clone());
-                    cache.put(key, (result.0.to_vec(), result.1.to_vec()));
-                }
-            });
-            result
+        match cached {
+            Some(result) => result,
+            None => {
+                let result = self.split_layout(area);
+
+                critical_section::with(|cs| {
+                    let mut cache = LAYOUT_CACHE.borrow(cs).borrow_mut();
+                    if let Some(cache) = cache.as_mut() {
+                        let key = (area, self.clone());
+                        cache.put(key, (result.0.to_vec(), result.1.to_vec()));
+                    }
+                });
+
+                result
+            }
         }
     }
 


ratatui-core/src/layout/layout.rs
```diff
@@ -1331,14 +1380,35 @@ mod tests {
...
+                );
+            });
+
+            Layout::init_cache(NonZeroUsize::new(10).unwrap());
+            critical_section::with(|cs| {
+                let cache = LAYOUT_CACHE.borrow(cs).borrow();
+                assert_eq!(cache.as_ref().unwrap().cap().get(), 10);
+            });
+        }
     }
```
j-g00da, line 1430: Split this into multiple function declarations, will be easier to follow.

(Same with split_with_spacers and init_cache)

ratatui-core/src/lib.rs
```diff
@@ -71,6 +71,8 @@
 extern crate alloc;
 #[cfg(feature = "std")]
 extern crate std;
+#[cfg(all(test, not(feature = "std")))]
+extern crate std;
```
j-g00da: Is this required? I think `cfg(test)` always links std.

j-g00da (approved):
Looks good, thanks!
I experimented with something similar last year.

> where the layout engine otherwise consumes all CPU, capping the framerate at around ~10fps

The main issue here is optimization of the layout engine, which currently is painfully slow. I digged into that few months ago, IIRC in some operations the time complexity is around O(n^4)... I'm pretty sure we wouldn't need cache at all but that would require rewriting the whole layout system while keeping it backward compatible. Related:
- https://github.com/ratatui/ratatui/pull/2196
- https://github.com/ratatui/ratatui/issues/1855

ratatui-core/src/lib.rs
junkdog: hmm, you have sent me down a rabbit hole. i'll report back when i understand what's going on.

junkdog: heh, this one stumped me - couldn't figure out why `cargo test --no-default-features`  with the lines commented is actually working.

after a bit of digging through the shell history:

```bash
$ cargo test --package ratatui-core --lib --no-default-features
   Compiling ratatui-core v0.1.0 (/home/junkdog/opt/dev/rust/ratatui/ratatui-core)
error: cannot find macro `dbg` in this scope
    --> ratatui-core/src/text/line.rs:1530:13
     |
1530 |             dbg!(line.width());
     |             ^^^
(etc..)
```

a fleeting moment of clarity except these compilation errors are unrelated to my diff.

a rabbit hole or so later, in ratatui-termiwiz/Cargo.toml

```toml
[dev-dependencies]
ratatui = { path = "../ratatui", features = ["termwiz"], default-features = false }
```

this pulls in ratatui without requesting std, but it's implictly pulled in by the "termwiz" features.

(ratatui-widet/Cargo.toml also pull in ratatui, though default features, as a dev dep).

the simplest way to confirm this is to comment out all modules except ratatui and ratatui-core, which will lead to `cargo test --no-default-features` to fail with `dbg!` missing from scope. 

given the above; it would maybe be less ugly to localize extern std for test somewhere closer to the source(/scope), instead of putting it in lib.rs. what do you think?

ratatui-core/src/layout/layout.rs
junkdog: with three possible configurations, i opted to create a split_with_spacers for not(layout-cache) too:

```rust
    #[cfg(not(feature = "layout-cache"))]
    pub fn split_with_spacers(&self, area: Rect) -> (Segments, Spacers) {
        self.try_split(area).expect("failed to split")
    }
```

ratatui-core/src/lib.rs
junkdog: i moved the extern std crate for test to the scopes where they were required.

junkdog:
> I'm pretty sure we wouldn't need cache at all but that would require rewriting the whole layout system while keeping it backward compatible.

hmm, interesting - i'll see if i can get #2196 working on the esp32 (later in the week though).

ratatui-core/src/layout/layout.rs
junkdog: oh, it failed the code-coverage, due to tests always running with `layout-cache`. i can rework it tomorrow, so that there's only one split_with_spacers, which dispatches to actual lru-ified fn when active.

ratatui-core/src/layout/layout.rs
```diff
@@ -32,21 +32,30 @@ type Spacers = Rects;
 // └   ┘└──────────────────┘└   ┘└──────────────────┘└   ┘└──────────────────┘└   ┘
 //
 // Number of spacers will always be one more than number of segments.
-#[cfg(feature = "layout-cache")]
+// std path: cache stores Rc directly (no Send needed with thread_local)
```
orhun:
```suggestion
// With std: cache can store Rc directly (no Send needed thanks to thread_local)
```

ratatui-core/src/layout/layout.rs
```diff
@@ -32,21 +32,30 @@ type Spacers = Rects;
...
+#[cfg(all(feature = "layout-cache", feature = "std"))]
 type Cache = LruCache<(Rect, Layout), (Segments, Spacers)>;
 
+// no_std path: cache stores Vec (Send-safe) for critical_section::Mutex
```
orhun:
```suggestion
// Without std: cache stores Vec instead (Send-safe for critical_section::Mutex)
```

ratatui-core/src/layout/layout.rs
```diff
@@ -711,18 +736,57 @@ impl Layout {
...
 
         #[cfg(not(feature = "layout-cache"))]
-        split()
+        {
+            self.split_layout(area)
+        }
+    }
+
+    #[cfg(all(feature = "layout-cache", feature = "std"))]
+    fn cached_split(&self, area: Rect) -> (Segments, Spacers) {
```
orhun, line 752:
```suggestion
    // std builds: use a thread-local cache with cheap Rc cloning and no locking.
    #[cfg(all(feature = "layout-cache", feature = "std"))]
    fn cached_split(&self, area: Rect) -> (Segments, Spacers) {
```

ratatui-core/src/layout/layout.rs
```diff
@@ -711,18 +736,57 @@ impl Layout {
...
+        LAYOUT_CACHE.with_borrow_mut(|cache| {
+            let key = (area, self.clone());
+            cache.get_or_insert(key, || self.split_layout(area)).clone()
+        })
+    }
+
+    #[cfg(all(feature = "layout-cache", not(feature = "std")))]
+    fn cached_split(&self, area: Rect) -> (Segments, Spacers) {
```
orhun, line 761:
```suggestion
    // no_std builds: use a critical-section cache and compute layouts outside locks.
    #[cfg(all(feature = "layout-cache", not(feature = "std")))]
    fn cached_split(&self, area: Rect) -> (Segments, Spacers) {
```

ratatui-core/src/layout/layout.rs
```diff
@@ -711,18 +736,57 @@ impl Layout {
...
+            let result = self.split_layout(area);
+            critical_section::with(|cs| {
+                let mut cache = LAYOUT_CACHE.borrow(cs).borrow_mut();
+                if let Some(cache) = cache.as_mut() {
+                    let key = (area, self.clone());
+                    cache.put(key, (result.0.to_vec(), result.1.to_vec()));
+                }
+            });
+            result
+        }
```
orhun:
```suggestion
        match cached {
            Some(result) => result,
            None => {
                let result = self.split_layout(area);

                critical_section::with(|cs| {
                    let mut cache = LAYOUT_CACHE.borrow(cs).borrow_mut();
                    if let Some(cache) = cache.as_mut() {
                        let key = (area, self.clone());
                        cache.put(key, (result.0.to_vec(), result.1.to_vec()));
                    }
                });

                result
            }
        }
```

junkdog: yeah, much nicer

PR #2400, opened 2026-02-15
refactor(core): introduce `CellWidth` trait for cell width computation

junkdog:
this PR introduces a `CellWidth` trait for calculating the cell width/span, implemented for `&str` and `Cell`.

The impl for `Cell` respects the new `CellDiffOption::ForcedWidth`. All width calculations are prefixed with a check if the symbol is ascii before calling `symbol.width()`; this micro-optimization probably won't do much on computers, but it helps on embedded.

As discussed recently in #1605, this also changes the `CellDiffOption::ForcedWidth(type) from NonZeroUsize to NonZeroU16. @benjajaja

### DIFF
--- a/.github/workflows/check-pr.yml
+++ b/.github/workflows/check-pr.yml
@@ -29,7 +29,7 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       # Add comment indicating we require pull request titles to follow conventional commits specification
-      - uses: marocchino/sticky-pull-request-comment@773744901bac0e8cbb5a0dc842800d45e9b2b405 # v2
+      - uses: marocchino/sticky-pull-request-comment@70d2764d1a7d5d9560b100cbea0077fc8f633987 # v3.0.2
         if: always() && (steps.check_pr_title.outputs.error_message != null)
         with:
           header: pr-title-lint-error
@@ -44,7 +44,7 @@ jobs:
 
       # Delete a previous comment when the issue has been resolved
       - if: ${{ steps.check_pr_title.outputs.error_message == null }}
-        uses: marocchino/sticky-pull-request-comment@773744901bac0e8cbb5a0dc842800d45e9b2b405 # v2
+        uses: marocchino/sticky-pull-request-comment@70d2764d1a7d5d9560b100cbea0077fc8f633987 # v3.0.2
         with:
           header: pr-title-lint-error
           delete: true
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,8 +36,8 @@ jobs:
         with:
           toolchain: nightly
           components: rustfmt
-      - uses: Swatinem/rust-cache@779680da715d629ac1d338a641029a2f4372abb5 # v2
-      - uses: taiki-e/install-action@a37010ded18ff788be4440302bd6830b1ae50d8b # v2
+      - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2
+      - uses: taiki-e/install-action@0d865d5cc6d507df4765f1f866bfae8bab4e2a73 # v2
         with:
           tool: taplo-cli
       - run: cargo xtask format --check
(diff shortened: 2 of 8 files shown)


ratatui-core/src/buffer/cell.rs
```diff
@@ -280,6 +281,15 @@ impl From<char> for Cell {
     }
 }
 
+impl CellWidth for Cell {
+    fn cell_width(&self) -> u16 {
```
orhun, line 287: The comment about `CellDiffOption::ForcedWidth` being respected belongs here I think

ratatui-core/src/buffer/cell_width.rs
```diff
@@ -0,0 +1,48 @@
+use unicode_width::UnicodeWidthStr;
+
+/// Returns the display width of a value in terminal cells.
+///
+/// This trait provides a unified way to compute cell widths for both string content
+/// and [`Cell`](super::Cell)s. For strings, the width is derived from
+/// [`UnicodeWidthStr`]. For cells,
+/// [`CellDiffOption::ForcedWidth`](super::CellDiffOption::ForcedWidth) is respected when set.
```
orhun: The comments regarding specific implementations of this trait could be linked here. I also think having a bullet point list would be easier to read. e.g.

- for `str` -> behaves like xyz.
- for `Cell` -> etc.

ratatui-core/src/buffer/cell_width.rs
```diff
@@ -0,0 +1,48 @@
...
+// Public because ratatui-widgets needs access, but not part of the user-facing API.
```
orhun: We might as well make this public? What would be the down side?

ratatui-core/src/buffer/cell_width.rs
```diff
@@ -0,0 +1,48 @@
...
+    fn cell_width(&self) -> u16;
+}
+
+impl CellWidth for str {
+    fn cell_width(&self) -> u16 {
+        if self.len() == 1 {
+            debug_assert!(
+                !self.as_bytes()[0].is_ascii_control(),
+                "control character passed to cell_width without filtering"
+            );
```
orhun, line 33: When does this fail? Can you maybe add a test case or a comment?

junkdog: it can't fail unless the internal impl of ratatui changes, so maybe it should say "ratatui excludes all control characters"

junkdog: or, hmm, if it's pub api, then it can fail the assert ofc.

ratatui-core/src/buffer/cell_width.rs
junkdog: more api to semver for something which is unlikely to be useful to end-users. but that's just my gut feeling.

junkdog: i can swing either way

ratatui-core/src/buffer/cell_width.rs
junkdog: added a comment

ratatui-core/src/buffer/cell_width.rs
orhun: I'd go with making this public, but it's always a good idea to have more eyes on this @ratatui/maintainers

ratatui-core/src/buffer/cell_width.rs
```diff
@@ -0,0 +1,56 @@
...
+    fn cell_width(&self) -> u16;
+}
+
+impl CellWidth for str {
+    fn cell_width(&self) -> u16 {
+        if self.len() == 1 {
+            // Control characters are filtered out by `Span::styled_graphemes()` and
+            // `Buffer::set_stringn()` before reaching this point. `Cell::set_symbol()`
+            // and `set_char()` do not filter, but those are low-level APIs where the
+            // caller is responsible for providing valid content.
```
orhun: Should we maybe add this as a doc-comment (`## Note`) to the implementation above? (instead of inner comment)

ratatui-core/src/buffer/cell_width.rs
junkdog: unhidden and public for now

ratatui-core/src/buffer/cell_width.rs
junkdog: docsed it

orhun:
need to resolve conflicts

PR #2416, opened 2026-03-01
perf(core): eliminate per-frame Vec allocation in Terminal::flush

junkdog:
## what and how

this PR removes the `Vec<(u16, u16, &Cell)>` allocation in `Terminal::flush` when calling `Buffer::diff`. a new method, `Buffer::diff_iter` is instead used by `Terminal::flush`.

the existing diff implementation has moved to the `BufferDiff` iterator.

## why (who cares?)

it's the mice. on embedded devices, allocating a short-lived, contiguous  block of up to 40-50kb is problematic due to heap fragmentation in combination with tiny heaps. the requirements for a full refresh over 1200 terminal cells is 37.5kb+vec growth padding, but since we're dealing with a contiguous block of memory, the actual memory requirements are considerably higher, depending on user-land allocation patterns.


ratatui-core/src/buffer/buffer.rs
```diff
@@ -490,73 +490,44 @@ impl Buffer {
     /// Updates: `0: a, 1: コ` (double width symbol at index 1 - skip index 2)
     /// ```
     pub fn diff<'a>(&self, other: &'a Self) -> Vec<(u16, u16, &'a Cell)> {
```
orhun, line 472: We should simplify the doc comment for this function (also means deduplicating it) and mention that it delegates to the `diff_iter` call.

ratatui-core/src/buffer/diff_iter.rs
```diff
@@ -0,0 +1,288 @@
...
+    pub(crate) fn new(prev: &'prev Buffer, next: &'next Buffer) -> Result<Self, BufferDiffError> {
+        if prev.area.x != next.area.x
+            || prev.area.y != next.area.y
+            || prev.area.width != next.area.width
+        {
+            return Err(BufferDiffError {
+                prev: prev.area,
+                next: next.area,
+            });
+        }
```
orhun: Hmm, did we handle this error before? I'm wondering about the previous behavior. 👀

ratatui-core/src/terminal/buffers.rs
```diff
@@ -75,11 +75,21 @@ impl<B: Backend> Terminal<B> {
...
         let previous_buffer = &self.buffers[1 - self.current];
         let current_buffer = &self.buffers[self.current];
-        let updates = previous_buffer.diff(current_buffer);
-        if let Some((col, row, _)) = updates.last() {
-            self.last_known_cursor_pos = Position { x: *col, y: *row };
+        let mut last_pos = None;
+
+        let updates = previous_buffer
+            .diff_iter(current_buffer)
+            .expect("terminal buffers have matching areas")
```
orhun: If we are panicking either way, do we really need to define our own error type?

junkdog: i was mostly thinking about diff_iter() returning something sensible if it fails any preconditions. 

but it occurred to me that this introduces a slight change to existing behavior: on main, performing `a.diff(b)` where `a.width != b.width` produces a truncated diff with mismatched positions (i.e., wrong cells marked in diff).

now, you'll instead get a panic, leading to application crash, which is arguably worse than a garbled ui. 

for a non-breaking approach, we could silently swallow any error. 

i'd personally favor deprecation of `Buffer::diff` as a long term fix.

ratatui-core/src/buffer/diff_iter.rs
junkdog: see my commen in  https://github.com/ratatui/ratatui/pull/2416#discussion_r2891414379 - but, in summary, no, existing impl produces garbage diff for buffers of different widths.

ratatui-core/src/terminal/buffers.rs
junkdog: another option is to panic + note the breaking change.

ratatui-core/src/buffer/buffer.rs
junkdog: updated

ratatui-core/src/terminal/buffers.rs
orhun: I think panicking in diff_iter and deprecating diff in the long term would be the best option here.

junkdog: it's gone

junkdog:
1981ee2 integrates the diff fix from #2308

:mouse2:

ratatui-core/src/buffer/diff.rs
orhun, line 1: nit: renaming this module to `diff` would be more suitable imo

junkdog: i agree, renamed

PR #2423, opened 2026-03-06
refactor(style): add descriptive panic for Color::Reset anstyle conversion

### LINKED ISSUE #2341: Converting `Color::Reset` to `anstyle::Color` panics
## Description

`Color::Reset` has no equivalent in `anstyle`, causing the conversion to fail.

## To Reproduce

Running the following: 
```rust
let new_color: anstyle::Color = ratatui::style::Color::Reset.into();
```
panics with
```
thread 'main' (3712423) panicked at .cargo/registry/src/index.crates.io-1949cf8c6b5b557f/ratatui-core-0.1.0/src/style/anstyle.rs:116:56:
called `Result::unwrap()` on an `Err` value: Ansi
```
This is because of [this unwrap call](https://github.com/ratatui/ratatui/blob/ab5ad3fc7ceb31b11ca9100a8d52dcb181e46b00/ratatui-core/src/style/anstyle.rs#L116).

I would think this conversion should implement `TryFrom` instead of `From` just like the other anstyle conversions.

## Expected behavior

It shouldn't panic.

I'm happy to make a PR if my proposed change is acceptable. This would be a breaking change, though.

## Environment

N/A

singhh-piyush:
## Description
`Color::Reset` has no equivalent in `anstyle::Color`. Previously, converting `Color::Reset.into()` fell through to the catch-all arm in `From<Color> for anstyle::Color`, which called `AnsiColor::try_from(color).unwrap()` and panicked with the message: `called Result::unwrap() on an Err value: Ansi`.

This replaces the opaque `unwrap()` with an explicit, immediately understandable panic: `"Color::Reset has no equivalent in anstyle"`.

This is a non-breaking fix for 0.30.1 , the `From<Color>` API signature is unchanged.

## Solution
- Added a dedicated `Color::Reset` match arm in `From<Color> for anstyle::Color` that panics with a descriptive message.
- Added a `#[should_panic]` regression test proving the exact panic string.

## AI Attribution
Used AI to help trace the anstyle conversion logic and draft the regression test.  Every line has been reviewed for correctness.

## Related Issues
Fixes #2341

### DIFF
--- a/ratatui-core/src/style/anstyle.rs
+++ b/ratatui-core/src/style/anstyle.rs
@@ -113,8 +113,8 @@ impl From<Color> for anstyle::Color {
     ///
     /// # Panics
     ///
-    /// This method will panic if the input is `Color::Reset`, as there is no
-    /// equivalent representation for a reset color in the `anstyle` crate.
+    /// This method will panic if the input is [`Color::Reset`], as there is no
+    /// equivalent representation for a reset color in `anstyle`.
     fn from(color: Color) -> Self {
         match color {
             Color::Reset => panic!("Color::Reset has no equivalent in anstyle"),


ratatui-core/src/style/anstyle.rs
```diff
@@ -111,6 +111,7 @@ impl From<anstyle::Color> for Color {
 impl From<Color> for anstyle::Color {
```
junkdog, line 111: we should document that the color conversion can panic

ratatui-core/src/style/anstyle.rs
```diff
@@ -345,4 +346,10 @@ mod tests {
         assert!(anstyle_style.get_effects().contains(Effects::BOLD));
         assert!(anstyle_style.get_effects().contains(Effects::ITALIC));
     }
+
+    #[test]
+    #[should_panic(expected = "Color::Reset has no equivalent in anstyle")]
+    fn converting_reset_panics_explicitly() {
```
junkdog, line 358: nice

singhh-piyush:
Added the `# Panics` documentation to state the `Color::Reset` behavior, as requested.

```rust
impl From<Color> for anstyle::Color {
    /// Converts a Ratatui `Color` into an `anstyle::Color`.
    ///
    /// # Panics
    ///
    /// This method will panic if the input is `Color::Reset`, as there is no
    /// equivalent representation for a reset color in the `anstyle` crate.
    fn from(color: Color) -> Self {

File: ratatui-core/src/style/anstyle.rs [hunk 3]
```diff
@@ -109,8 +109,15 @@ impl From<anstyle::Color> for Color {
 }
 
 impl From<Color> for anstyle::Color {
+    /// Converts a Ratatui `Color` into an `anstyle::Color`.
+    ///
+    /// # Panics
+    ///
+    /// This method will panic if the input is `Color::Reset`, as there is no
+    /// equivalent representation for a reset color in the `anstyle` crate.
```
orhun:
```suggestion
    /// This method will panic if the input is [`Color::Reset`], as there is no
    /// equivalent representation for a reset color in `anstyle`.
```

PR #2424, opened 2026-03-06
fix(list): clamp selected index using cached item count

### LINKED ISSUE #2421: Fix clamping in list navigation method
> I don't know if this is the right place to comment, but the clamping is incorect. It clamps to `items.len()` instead of `items.len() - 1`
It doesn't cause any problems for just displaying and navigating the list, BUT if you couple some other part of your code with `list_state.selected()`, you will get index out of bound errors.

> My current workaround:
> ```
> let next = self.list_state.selected().map_or(0, |i| i+1);
> self.list_state.select(Some(next.min(self.items.len()-1)));
> ```

_Originally posted by @errwip in https://github.com/ratatui/ratatui/issues/1159#issuecomment-3993292978_

Logan-Ruf:
## Summary

`ListState` navigation methods (`select_next`, `select_previous`, `select_last`, `scroll_down_by`)
could set the selected index beyond the valid range. The clamping only happened during `render()`,
so reading `selected()` between a navigation call and the next render returned an out-of-bounds
index, causing panics when used to index into the items list.

This PR caches the item count in `ListState` during rendering and uses it in navigation methods to
clamp immediately.

Fixes #2421

## Problem

`ListState` and `List` are separate structs that only meet during `render()`. Navigation methods
like `select_next()` had no way to know the item count, so they freely incremented the index
without an upper bound. The render-time clamping at `rendering.rs:47-50` corrected the index, but
only after the fact. In the common app pattern of handling input then rendering, `selected()`
returned an invalid index between those two steps:

```rust
// items has 3 entries (indices 0, 1, 2)
state.select(Some(2));
state.select_next();
let item = items[state.selected().unwrap()]; // panic: index 3, len 3
```

## Solution

- Added an `item_count: Option<usize>` field to `ListState` (skipped in serde, excluded from
  `PartialEq`/`Hash`)
- select() and set_item_count() return bool (true = in bounds, false = clamped)
- Navigation methods (select_next, select_previous, scroll_down_by, scroll_up_by) return bool (true = moved, false = clamped/at boundary)
- Updated `select_previous()` and `select_last()` to use `item_count` instead of the `usize::MAX`
  sentinel when the count is known
- Enables this neat pattern for the list wrapping:
```rust
use ratatui::widgets::ListState;
                                                               
fn select_next_or_wrap(state: &mut ListState) {
    if !state.select_next() {
        state.select_first();
    }
}
                                                               
fn select_previous_or_wrap(state: &mut ListState) {
    if !state.select_previous() {
        state.select_last();
    }
}
                                                               
let mut state = ListState::default().with_item_count(Some(5));
                                                               
state.select_first();
select_previous_or_wrap(&mut state);
assert_eq!(state.selected(), Some(4)); // wrapped to last
                                                               
select_next_or_wrap(&mut state);
assert_eq!(state.selected(), Some(0)); // wrapped to first
```

## Breaking changes

- `ListState::select_last()` is no longer `const fn` (it now reads `item_count`). This required a
  one-line fix in the `todo-list` example.
- `ListState` navigation method return `bool` instead of `()`.

### 1. Add list wrap feature to `List` widget example
This could be a `good first issue`
`ratatui-widgets/examples/list`

### 2. Apply the same fix to `TableState`

`TableState` has a similar bug with row and column navigation. I did not address this to keep this
PR small and focused, and because the table has additional considerations (e.g. column count is
derived from the widest row).

## AI attribution

This PR was developed with the use of Claude Code. All AI changes were reviewed for correctness.


ratatui-widgets/src/list/state.rs
```diff
@@ -193,7 +231,8 @@ impl ListState {
     /// state.select_previous();
     /// ```
     pub fn select_previous(&mut self) {
-        let previous = self.selected.map_or(usize::MAX, |i| i.saturating_sub(1));
+        let last = self.item_count.map_or(usize::MAX, |c| c.saturating_sub(1));
```
junkdog: `last` is a bit misleading, i suggest renaming it to `current`

ratatui-widgets/src/list/state.rs
```diff
@@ -354,4 +395,176 @@ mod tests {
...
+            item_count: Some(5),
+            ..Default::default()
+        };
+
+        state.select_last();
+
+        assert_eq!(
+            state.selected(),
+            Some(4),
+            "should go to last valid index, not usize::MAX"
```
junkdog: in case of future failure, we don't know that the it's failing due to `usize::MAX`

```suggestion
            "should go to last valid index"
```

ratatui-widgets/src/list/state.rs
```diff
@@ -354,4 +395,176 @@ mod tests {
...
+            item_count: Some(5),
+            ..Default::default()
+        };
+
+        state.select_previous();
+
+        assert_eq!(
+            state.selected(),
+            Some(4),
+            "should go to last item, not usize::MAX"
```
junkdog:
```suggestion
            "should go to last item"
```

ratatui-widgets/src/list/state.rs
```diff
@@ -177,6 +214,7 @@ impl ListState {
     pub fn select_next(&mut self) {
         let next = self.selected.map_or(0, |i| i.saturating_add(1));
         self.select(Some(next));
+        self.clamp_selected();
```
junkdog: `select()` is almost always followed by `clamp_selected()` - did you consider calling `clamp_selected()` from within `select()`?

ratatui/tests/widgets_list.rs
```diff
@@ -369,3 +369,33 @@ fn widgets_list_enable_always_highlight_spacing<'line, Lines>(
         .backend()
         .assert_buffer(&Buffer::with_lines(expected));
 }
+
+#[test]
+fn widgets_list_select_next_should_not_exceed_item_count() {
```
junkdog, line 374: i'm not sure about this test. it doesn't verify anything which isn't already covered by the other tests in this PR:  the setup with `state.select(Some(2))` is already a valid position, so it's not clamped by `state.item_count = ...` in render().

ratatui-widgets/src/list/state.rs
```diff
@@ -354,4 +395,176 @@ mod tests {
...
+
+        assert_eq!(
+            state.selected(),
+            Some(4),
+            "should go to last item, not usize::MAX"
+        );
+    }
+
+    #[test]
+    fn without_item_count_select_next_uses_old_behavior() {
```
junkdog: this test case is already covered by `state_navigation()`  above

ratatui-widgets/src/list/state.rs
```diff
@@ -354,4 +395,176 @@ mod tests {
...
+
+        assert_eq!(
+            state.selected(),
+            Some(3),
+            "should move normally when within bounds"
+        );
+    }
+
+    #[test]
+    fn clamp_selected_does_nothing_without_item_count() {
```
junkdog: already tested

ratatui-widgets/src/list/state.rs
junkdog: alt just delegate to `scroll_up_by(1)`.

junkdog (changes requested):
hi, thanks for the PR! i believe i've worked around this behavior/bug at least twice in my own projects, so it's good to see it addressed in some manner. 

I had a couple of thoughts around possible simplification + remarks on duplicated tests.

ratatui-widgets/src/list/state.rs
Logan-Ruf: > alt just delegate to `scroll_up_by(1)`.

This would be cleaner code, but these functions have different behaviors when `selected` is `None`.

Here's a quick test I wrote to demonstrate.
```rs
#[test]
fn state_naviagtion_when_select_is_none() {
    let mut state = ListState::default();
    state.select_previous();
    assert_eq!(state.selected, Some(usize::MAX));
                                                  
    let mut state = ListState::default();
    state.scroll_up_by(1);
    assert_eq!(state.selected, Some(0));
}
```

I don't know this project well enough to know if this change would be desirable or not.

ratatui/tests/widgets_list.rs
Logan-Ruf: This is more of a end-to-end test I created to reproduce the issue before I started working on the solution. If it is unnecessary I can remove it to clean up the PR.

ratatui-widgets/src/list/state.rs
Logan-Ruf: After reviewing the original comment that this issue was made from, I think this makes sense.
I got a little confused with the issue relating to the PR for navigation methods and the content of the issue.
The original commenter actually mentioned the `list_state` not clamping correctly on the `select()` function. He might have thought it was supposed to clamp right away and not only during render.
This PR is on the right path fix this, but still needs some more work.

If this many changes are being made, I think 1 and 2 in my "Follow-Up work" would make sense to add in the same PR.

ratatui-widgets/src/list/state.rs
junkdog: ah, good catch, i didn't think it through properly

ratatui-widgets/src/list/state.rs
junkdog: i think 1. re: adding `set_item_count()`  would be somewhat confusing, as the value is overwritten on each rendered frame. 2. is better handled in a different issue/pr - but i disagree with it being fully backwards compatible; strictly speaking, it's a breaking change as it changes the fn signature and would likely affect some projects.

ratatui-widgets/src/list/state.rs
```diff
@@ -33,24 +35,66 @@
...
 ///
 /// [`List`]: super::List
-#[derive(Debug, Default, Clone, Copy, Eq, PartialEq, Hash)]
+#[derive(Debug, Default, Clone, Copy)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub struct ListState {
     pub(crate) offset: usize,
     pub(crate) selected: Option<usize>,
+    #[cfg_attr(feature = "serde", serde(skip))]
+    pub(crate) item_count: Option<usize>,
```
junkdog, line 50: opting-out of deriving creates a lot of ad hoc boilerplate. it would be better to use a wrapper type with manual trait impls + serde skip. for example a `Cached<Option<usize>>` or maybe `Transient<T>`.

ratatui-widgets/src/list/state.rs
Logan-Ruf: You are right on the backwards compatibility. I was probably saying it incorrectly, but basically I was trying to minimize the change in logic. The function signature would be caught by the compiler, but changes in logic wouldn't and could have some unintended side effects.

For add `set_item_count()`, I already had that done last night, but didn't push the changes yet. I probably put too much in this PR, I was kinda having fun didn't want to stop lol

I'll explain my reasoning on why I think we should have the two other features if we are adding clamping.

1. `set_item_count()`
This basically boils down to changing a list's data and trying to move past the previous rendered length.
Say a person has a list that gets a stream of new elements, and wants to always select the last element appended to the list. Without being able to update the `item_count`, you would have to render first to select the last item.\
\
I will admit this is a contrived scenario, but it *is* a limitation introduced in this PR. I wanted to address it, but if it is too much I can remove it.

2. Navigation functions returning `bool`
If we are clamping the data, I wanted some indication that what the library did might be different than what the user expected. I can imagine users having to implement their own checks to see if the `selected` index was clamped or not.
Also, the function signature change makes the change in logic more apparent.

ratatui-widgets/src/list/state.rs
Logan-Ruf: This is much cleaner. Thanks for the tip!

ratatui-widgets/src/list/state.rs
junkdog: yeah, i think it's best to raise it separately to keep the PR within scope.

i'm not sold on `set_item_count()`; it feels too much like fiddling with internals, as the item count isn't actually managed at this level. 

> 2\. Navigation functions returning `bool`

i agree it would be a better developer experience but it has to be weighed against breaking the api. take this in a separate PR - it would also give us more eyes on it too.

Logan-Ruf:
@junkdog I've made the requested changes. I'm writing this comment to draw your attention to a few things I was unsure about so you don't have to go find them in the code changes.
1. `item_count`
image\
I made the setter for `item_count` `pub(crate)`, but I kept the field in the description of the `struct`. Is that good, or should I remove that?
2. `item_count` getter
image\
Similar question as 1, but for this getter function.
3. Documentation changes
image\
[Source](https://github.com/ratatui/ratatui/pull/2424/changes#diff-ccb81dd3fbdda2a82c92ba47caa195568c7d74f323caa8c2098f4dfffd79bd04R184-R228)\
I wanted to make sure the different behaviors these function have with the clamping was well documented. I added examples in a way that reminds me of Jupyter notebook, where you have documentation and code inline. I wanted to get your opinion on if you thought it looked good, or just added to much.

Thanks again for your help on this PR!

ratatui-widgets/src/list/state.rs
```diff
@@ -45,6 +49,8 @@
 pub struct ListState {
     pub(crate) offset: usize,
     pub(crate) selected: Option<usize>,
+    #[cfg_attr(feature = "serde", serde(skip))]
+    pub(crate) item_count: Transient<Option<usize>>,
```
junkdog: the more i think about this, the more i think `Transient` is overengineered. `Option<usize>>` is fine, and i don't think it should be skipped by serde - unless i'm missing something?

ratatui-widgets/src/list/state.rs
junkdog, line 133: maybe add something like:

> If `item_count` is known (set during rendering), the selected index is clamped to valid bounds.

ratatui-widgets/src/list/state.rs
```diff
@@ -142,37 +151,138 @@ impl ListState {
...
+    /// # use ratatui::widgets::{List, ListState, StatefulWidget};
+    /// # use ratatui::buffer::Buffer;
+    /// # use ratatui::layout::Rect;
+    /// #
+    /// let mut state = ListState::default();
+    /// let r = Rect::new(0, 0, 20, 20);
+    /// StatefulWidget::render(List::new(vec![""; 3]), r, &mut Buffer::empty(r), &mut state);
+    /// state.select(Some(2));
+    /// state.select_next(); // clamped to last item
+    /// assert_eq!(state.selected(), Some(2));
```
junkdog: let's keep the docs simple, i'd just mention it item_count briefly in the docs, for example:

> Bounded by item count when known, otherwise falls back to `usize::MAX`

ratatui-widgets/src/list/state.rs
```diff
@@ -191,59 +304,139 @@ impl ListState {
...
+        let index_max = self
+            .item_count
+            .0
+            .map_or(usize::MAX, |c| c.saturating_sub(1));
+        let previous = self.selected.map_or(index_max, |i| i.saturating_sub(1));
         self.select(Some(previous));
     }
 
-    /// Selects the first item
+    /// Selects the first item or `None` if the `item_count` is `0`.
```
junkdog, line 298: 👍

ratatui-widgets/src/list/state.rs
```diff
@@ -191,59 +304,139 @@ impl ListState {
...
         self.select(Some(0));
     }
 
     /// Selects the last item
     ///
-    /// Note: until the list is rendered, the number of items is not known, so the index is set to
-    /// `usize::MAX` and will be corrected when the list is rendered
+    /// Note: until the list is rendered, the number of items is not known. In this case,
+    /// the function will select `usize::MAX` as the last item.
+    /// If `item_count` is known, the function will select `item_count - 1` as the last item.
```
junkdog, line 319: 👍

ratatui-widgets/src/list/state.rs
```diff
@@ -191,59 +304,139 @@ impl ListState {
...
+    /// # use ratatui::layout::Rect;
+    /// #
     /// let mut state = ListState::default();
-    /// state.scroll_down_by(4);
+    /// let r = Rect::new(0, 0, 20, 20);
+    /// StatefulWidget::render(List::new(vec![""; 5]), r, &mut Buffer::empty(r), &mut state);
+    /// state.select(Some(2));
+    /// state.scroll_down_by(4); // clamped to last item
+    /// assert_eq!(state.selected(), Some(4));
     /// ```
```
junkdog: restore to old example

ratatui-widgets/src/list/state.rs
```diff
@@ -256,13 +449,23 @@ impl ListState {
...
     /// let mut state = ListState::default();
+    /// state.scroll_up_by(3);
+    /// assert_eq!(state.selected(), Some(0));
+    ///
+    /// state.select(Some(6));
     /// state.scroll_up_by(4);
+    /// assert_eq!(state.selected(), Some(2));
+    ///
+    /// state.scroll_up_by(4); // saturates at first item
+    /// assert_eq!(state.selected(), Some(0));
```
junkdog: the old example was sufficient, i think

ratatui-widgets/src/list/state.rs
junkdog: this applies to all select* and scroll_by functions

junkdog (changes requested):
looking better, but let's clean up(/restore) the docs +  remove `Transient`

ratatui-widgets/src/list/state.rs
Logan-Ruf: Since item_count is value set during runtime and caches information about something else. I figured it would be best to exclude it from serialization, `Hash`, `Eq`, and `Hash`. We can remove Transient and serde skip and the code would be much simpler.

junkdog:
apologies, forgot to reply after finishing the review.

> 1. I made the setter for `item_count` `pub(crate)`, but I kept the field in the description of the `struct`. Is that good, or should I remove that?

i think it's fine.

> 2\. `item_count` getter

:+1:  

> 3\. Documentation changes

addressed in the review, but i think they'd benefit from being shorter.

themavik (commented):
nit: select_previous doc says "is know" — should be "known". Pushing clamping through set_item_count at render beats scattering bounds fixes in the widget.

ratatui-widgets/src/list/state.rs
```diff
@@ -142,29 +150,67 @@ impl ListState {
         &mut self.selected
     }
 
+    /// Returns the number of items in the list, if known.
+    ///
+    /// This value is set during rendering.
```
orhun: I think this is my biggest problem with this PR. It seems odd that we require rendering on the user-side to be able to provide a detail in our API. It smells like we are doing some internal state-management :)

I would expect this method to be called anytime, since the user has most likely provided the required data while constructing the `ListState`.

(Maybe not the best example) but take [`Frame::count`](https://docs.rs/ratatui/latest/ratatui/struct.Frame.html#method.count) for example. That is always callable without rendering and returns a value (not an Option).

Maybe we should set the item count while constructing, if possible?

Logan-Ruf: This was implemented in an earlier [commit](https://github.com/ratatui/ratatui/blob/14a643c05ed7bc06cba14239efa032037ab9252c/ratatui-widgets/src/list/state.rs), before being removed to focus more on the issue.

I'll just [link this](https://github.com/ratatui/ratatui/pull/2424#discussion_r2907664711) earlier comment from @junkdog. I can do either implementation.

junkdog: what about optionally providing item count during construction? the `set_item_count` as a pub method feels slightly misleading, but it's mostly an expectation inferred from the name: i'd expect a setter to more permanently set the value, but item_count is updated on each rendered frame.

Logan-Ruf: I think that is a good point. Maybe `update_item_count` would be a better name. Then we could note in the docs that the widget will also update the `item_count` on render. Is `item_count` a good name for the variable? Or would something like `length` or `count` be better?

joshka:
I think this should be **non-surprising from a user perspective**.

My preference would be for `ListState` to support a **user-settable item count**. For backwards compatibility, that value could still be omitted and then populated on first render, but I don't think render-time discovery should be the only way this works.

Related to that, I think this should follow the same pattern as the other fields, with `item_count()` / `item_count_mut()`, so users can set it directly and have the state behave consistently before the first render. I don't think we need a separate `set_item_count` API if we expose it this way.

I also think we should make the end-of-list behavior explicit via an enum rather than encoding it indirectly. There are at least two obvious policies:

* clamp at the ends
* wrap around

Something like:

```rust
pub enum SelectionMode {
    Clamp,
    Wrap,
}
```

and then a field on `ListState` (e.g. `selection_mode`) would make this behavior clear, composable, and easy to reason about.

Overall, I think this keeps the API consistent with the rest of `ListState`, avoids render-driven surprises, and makes the navigation behavior explicit instead of implicit.

On the docs side, I’d lean toward being **explicit even if it means slightly more explanation**.

In particular, I think it’s valuable to capture some of the *why*, not just the *what*. For example, the fact that `item_count` is historically derived during render explains:

* why older code relies on render-time behavior
* why users might see patterns that don’t align with the current API
* why certain edge cases exist (e.g. pre-render behavior)

That kind of context helps users understand what they’re seeing in the wild and reduces confusion when behavior doesn’t immediately match expectations.

I agree we shouldn’t be overly verbose for simple APIs, but for areas with historical baggage or non-obvious behavior, a bit of extra explanation is often worth it.

Logan-Ruf:
> I also think we should make the end-of-list behavior explicit via an enum rather than encoding it indirectly. There are at least two obvious policies:
> 
> * clamp at the ends
> * wrap around
> 
> Something like:
> 
> ```rust
> pub enum SelectionMode {
>     Clamp,
>     Wrap,
> }
> ```

@joshka, while I do think it is a good idea to enable list wrapping, I have some concerns with this solution and would like your input on an alternative.

My first concern is how we'd implement wrapping with `scroll_by` functions. I can imagine a few different behaviors that might be desired, which would mean we either need to support configurations for these behaviors or decide on only having one default behavior.
1. Should we wrap on `scroll_by` or only wrap on `select_next` and `select_previous`?
2. Should `scroll_by` always wrap or only wrap on the last item?\
For example, if `selected = 2` and we `scroll_up_by(3)`, should we select `0` or the `item_count - 1`?
4. Do we continue scrolling after a wrap.\
For example, if `selected = 1` and we `scroll_up_by(100)` do we just select `item_count - 1`, or do we continue scrolling up by the remaining 98?

A few other behaviors I can think of:
1. Do we wrap when you `select` an index past the end of the list?
2. Do we wrap when `[update/set]_item_count` causes `selected` to be out of range?
3. Do we need to return something or have a callback when clamping or wrapping occurs in case the user wants something to happen at the same time? Like an auditory or visual indicator?

My alternative would be to return a `bool` on methods to indicate if clamping occurred. This would make implementing list wrapping by the user trivial, for example:
```rust
if !state.select_next() {
    state.select_first();
}
```
It would also mean we wouldn't need to answer any of the previous questions, since those using the framework would implement their own solutions. It would also make maintenance and development simpler.

I had this implemented in an earlier commit, but I removed so I could submit in a later PR on @junkdog [recommendation](https://github.com/ratatui/ratatui/pull/2424#discussion_r2907664711). I'll add it back soon, since I can simply copy and paste it back in. That said, I want to be clear I'm open to input and discussion on this.

joshka:
Opinionated responses... (happy to hear other perspectives on this)

>Should we wrap on scroll_by or only wrap on select_next and select_previous?

Yes. If the instruction to scroll by 2 items is followed, it should have the same effect as scrolling by one twice.
Those functions are confusingly named. I had to check the source, but they manipulate the selection rather than the positioning of the list. That's not obvious from the name, but it's what they do.

>Should scroll_by always wrap or only wrap on the last item?
For example, if selected = 2 and we scroll_up_by(3), should we select 0 or the item_count - 1?

For the same reason item_count - 1

> Do we continue scrolling after a wrap.

For the same reason, yes

>    Do we wrap when you select an index past the end of the list?

I'd probably take a step back here and think about the reasons people are calling the code to set the selected index and then infer a meaningful answer from that. Any analysis of this has to take into account the item_count (known / unknown) and the rendering behavior (i.e. what happens when you set the count past the item_count then add another item on the next render, similarly what happens when reducing it. I'd say probably this is handled at render time, not set time due to this factor (i.e. selected can be > count until rendering occurs at which point we clamp / wrap as needed)

>    Do we wrap when [update/set]_item_count causes selected to be out of range?

Same answer. A good way to think about this is that if we flipped which we set first, then the result should intuitively be the same result.

>    Do we need to return something or have a callback when clamping or wrapping occurs in case the user wants something to happen at the same time? Like an auditory or visual indicator?

I dislike bool results on this sort of thing almost as much as bool parameters, they don't convey meaning for future readers of the code (i.e. they reader has to look up what true / false means in the return value). I'd use an enum for this (https://rust-lang.github.io/api-guidelines/type-safety.html#arguments-convey-meaning-through-types-not-bool-or-option-c-custom-type)

Personally I'd probably leave it unless there's a call for it, but I could see how something like this might be valuable. It does avoid a future breaking change, but I'm less concerned about that now that we have the -core / -widgets split.

PR #2426, opened 2026-03-07
feat(widgets): add line shape with filled area for Canvas and Chart

bananaofhappiness:
This PR adds a filled area chart visualization to both `Canvas` and `Chart` widgets. You can now draw lines with a shaded area underneath or above them, which is great for emphasizing volume or magnitude in data. Use `AreaLine` in `Canvas` or `GraphType::AreaLine(f64)` with your `Chart` (I'm still not sure about the name of this type, so feel free to change it). `f64` here shows, up to which Y-coordinate area should be filled.

Note that `Eq` and `Hash` implementations were removed from `GraphType` because `f64` in `AreaLine(f64)` doesn't implement them. They are not used in the library and don't really make a lot of sense (for me at least). If they are needed, you can make a custom implementation or use `ordered-float` crate.

Some screenshots with and without this new type:
изображение
изображение

Example usage:
```rust
// In Canvas
Canvas::default()
    .paint(|ctx| {
        ctx.draw(&AreaLine::new(0.0, 0.0, 10.0, 5.0, 0.0, Color::Red));
    });

// In Chart
Chart::new(dataset)
    .graph_type(GraphType::AreaLine(0.0)); // fill to y=0
```


junkdog:
hi, thanks for the PR! i took a quick look - i noticed that it copy-implements all of `line.rs`; it would make more sense to delegate to it, and then paint the area. 

how does this play with multiple datasets in one chart? 

> Note that `Eq` and `Hash` implementations were removed from `GraphType` because `f64` in `AreaLine(f64)` doesn't implement them. They are not used in the library ...

they're part of the public API though so it's better to avoid breaking it unless we have a compelling reason for doing so.

in this particular instance, i don't  think the `f64` belongs in the enum variant as it's more a property of the data set than the graph type. something like `Dataset::fill_to_y(f64)` with a plain `GraphType::Area` would avoid the breaking change.

bananaofhappiness:
> it would make more sense to delegate to it, and then paint the area.

Yes, I thought about it, but I went for the most straight forward way. It was easier. If you delegate it to the line, then it's impossible to get all of the painted interpolated points after it's done painting. Or maybe I can access grid of the `Painter`? (but it would be too complicated). One more solution is: draw a line, then draw a new one with the offset of 1 towards the `fill_to_y` and so on until we covered the whole area, but I'm not sure if it's efficient (performance-wise).

Using `Dataset::fill_to_y(f64)` is better, I'll switch to it.

> how does this play with multiple datasets in one chart?

Not so good with `Braille`, `Quadrant`, `Sextant`, `Octant`, better with other markers. Here are screenshots with Mid and Side spectrum in my app.

<details>
  <summary>Screenshots</summary>

  Braille:
изображение
Quadrant:
изображение
Dot:
изображение
Block:
изображение

</details>

junkdog:
> Screenshots

thanks! yeah, overlapping area charts are a pain elsewhere too. it's a limitation of the chart type.

> If you delegate it to the line, then it's impossible to get all of the painted interpolated points

oh yeah, you're right, but let's remove the other other copy-pasted functions and see where we end up. i'll leave a review in a bit.

ratatui-widgets/src/canvas/area_line.rs
```diff
@@ -0,0 +1,402 @@
...
+            }
+        } else if y1 > y2 {
+            draw_line_high(painter, x2, y2, x1, y1, y_fill, self.color);
+        } else {
+            draw_line_high(painter, x1, y1, x2, y2, y_fill, self.color);
+        }
+    }
+}
+
+fn clip_line(
```
junkdog: identical to `clip_line()` in line.rs, i think

ratatui-widgets/src/canvas/area_line.rs
```diff
@@ -0,0 +1,402 @@
...
+        LineSegment::new(Point::new(x1, y1), Point::new(x2, y2)),
+        Window::new(xmin, xmax, ymin, ymax),
+    ) {
+        Some((x1, y1, x2, y2))
+    } else {
+        None
+    }
+}
+
+fn draw_line_low(
```
junkdog: this is 95% the same as the function in line.rs

ratatui-widgets/src/canvas/area_line.rs
```diff
@@ -0,0 +1,402 @@
...
+) {
+    let dx = (x2 - x1) as isize;
+    let dy = (y2 as isize - y1 as isize).abs();
+    let mut d = 2 * dy - dx;
+    let mut y = y1;
+    for x in x1..=x2 {
+        let y_range = if y > y_fill { y_fill..=y } else { y..=y_fill };
+        for y in y_range {
+            painter.paint(x, y, color);
+        }
```
junkdog: extract this snippet into a separate function instead, and use the original draw_line_low/high functions from line.rs.

ratatui-widgets/src/chart.rs
```diff
@@ -154,7 +154,7 @@ impl<'a> Axis<'a> {
 }
 
 /// Used to determine which style of graphing to use
-#[derive(Debug, Default, Display, EnumString, Clone, Copy, Eq, PartialEq, Hash)]
+#[derive(Debug, Default, Display, EnumString, Clone, Copy, PartialEq)]
```
junkdog:
```suggestion
#[derive(Debug, Default, Display, EnumString, Clone, Copy, Eq, PartialEq, Hash)]
```

ratatui-widgets/src/chart.rs
```diff
@@ -1043,6 +1050,18 @@ impl Widget for &Chart<'_> {
                                 });
                             }
                         }
+                        GraphType::AreaLine(y) => {
```
junkdog: maybe `GraphType::Area`?

ratatui-widgets/src/canvas/area_line.rs
```diff
@@ -0,0 +1,402 @@
...
+            } else {
+                y.saturating_add(1)
+            };
+            d -= 2 * dx;
+        }
+        d += 2 * dy;
+    }
+}
+
+fn draw_line_high(
```
junkdog: same here, the y fill code is the only difference, better to use the original fn.

junkdog (changes requested):
nice feature but at the moment, it copies too much code from line.rs. i left some comments around the logic which is specific to `LineArea`. extracting it to its own function(s) would eliminate a lot of the copy-pasta and end up looking something like this, for each drawn line:

```rust
draw_column(painter, y2, y_fill); 
draw_line_low(painter, x2, y2, x1, y1, self.color);
```

bananaofhappiness:
Hi! I tried to implement an `Area` shape that would render arbitrary shapes on a `Canvas` given a set of vertices (and fill them when `fill=true`). But then I checked and realized that this approach is poorly suited for charts: in my TUI audio analyzer, the FFT spectrum has thousands of points. To render a filled `Area`, I would have to add `(x_max, fill_to_y)` and `(x_min, fill_to_y)` to the slice of vertices to properly close the shape before filling, which would force extra allocations (also it was very slow).

In the end, I went back to `Line`, created a new `FilledLine` structure (the name is slightly incorrect since a line is one-dimensional and cannot be filled), and created abstractions over both structures in the form of functions `for_each_line_point`, `for_each_line_point_low`, and `for_each_line_point_high`, which accept closures. For `Line`, this closure simply uses `painter.print`, while for `FilledLine`, it also fills the column.

Since I had already created the `Area` structure, I decided to keep it, finished the implementation, and wrote tests and documentation. Perhaps it would be more correct to do this as a separate PR?

junkdog:
> Since I had already created the `Area` structure, I decided to keep it, finished the implementation, and wrote tests and documentation. Perhaps it would be more correct to do this as a separate PR?

better to take it in a different PR!

bananaofhappiness:
ok, I'll cherry pick necessary commits (only for FilledLine) and force push them into this PR, and make a new PR for new shape Area

bananaofhappiness:
I made it. #2436

ratatui-widgets/src/canvas/line.rs
```diff
@@ -101,13 +72,60 @@ fn clip_line(
...
-fn draw_line_low(painter: &mut Painter, x1: usize, y1: usize, x2: usize, y2: usize, color: Color) {
+pub(super) fn draw_line(
+    painter: &mut Painter,
+    x1: usize,
+    y1: usize,
+    x2: usize,
+    y2: usize,
+    color: Color,
+) {
+    for_each_line_point(x1, y1, x2, y2, |x, y| painter.paint(x, y, color));
```
junkdog, line 83: nice!

ratatui-widgets/src/canvas/line.rs
```diff
@@ -101,13 +72,60 @@ fn clip_line(
...
+}
+
+/// Iterate over each point on a line using Bresenham's algorithm.
+///
+/// Using a callback allows us to reuse the algorithm without duplicating code:
+/// - `Line` just paints each point
+/// - `FilledLine` paints a vertical column from each point to `fill_to_y`
```
junkdog: nit/suggestion

```suggestion
/// Calls `f(x, y)` for each pixel on the Bresenham line from `(x1, y1)` to `(x2, y2)`.  
```

junkdog (approved):
i think it looks good! left a suggestion for one of the fn docs.

probably wise to get another pair of eyes on it too, i haven't had much exposure to the canvas code before, @ratatui/maintainers

orhun (commented):
This looks good to me!

Would love to see this in action in either in the updated examples or a new widget example :)
See https://github.com/ratatui/ratatui/tree/main/ratatui-widgets/examples for widget examples.

We could potentially add a new [app example](https://github.com/ratatui/ratatui/tree/main/examples/apps) as well. Maybe an audio visualizer, which I think fits best for this widget's use case.

bananaofhappiness:
I can add examples, it won't be difficult. And about an app example, I could use a snippet from my audio analyzer, only the frequency spectrum part, well, like in the image I attached to this PR

orhun:
Sounds good to me!

bananaofhappiness:
I've updated chart widget example, now it looks like this.

<details>
  <summary>Screenshot so you don't have to run it and can just stay on GitHub reviewing.</summary>
  
  
изображение

  
</details>

orhun (approved):

orhun:
thanks for the PR again!

bananaofhappiness:
I was also working on an app example, but took a pause due to personal reasons. I'll create a new PR then with it when it's finished

orhun:
very nice!

PR #2436, opened 2026-03-13
feat(widgets): Add new shape Area to a Canvas, a polygon defined by set of vertices

bananaofhappiness:
This PR implements an Area shape that would render arbitrary shapes on a Canvas given a set of vertices (and fill them when fill=true). This PR depends on #2426, it uses `draw_line` defined in `line.rs` in that PR. It uses Sutherland-Hodgman clipping algorithm and scanline algorithm to fill an area.
Example: 
```rust
fn render(frame: &mut Frame) {
    let p = &[
        (4.0, 3.0),
        (2.5, 0.0),
        (6.0, 1.0),
        (7.0, 2.5),
        (6.0, 5.0),
        (2.5, 3.0),
    ];
    let cv = Canvas::default()
        .x_bounds([0.0, 10.0])
        .y_bounds([0.0, 5.0])
        .marker(Marker::Block)
        .paint(|ctx| {
            let area = Area::new(p, Color::Red, true);
            ctx.draw(&area);
        });
    frame.render_widget(cv, frame.area());
}
```
<details>
  <summary>Will draw</summary>

```
                                                                           ██
                                                                        ██████
                                                                     ██████████
                                                                  █████████████
                                                              ██████████████████
                                                           ██████████████████████
                                                        █████████████████████████
                                                     █████████████████████████████
                                                  █████████████████████████████████
                                               ████████████████████████████████████
                                            ████████████████████████████████████████
                                        █████████████████████████████████████████████
                                     ████████████████████████████████████████████████
                                  ████████████████████████████████████████████████████
                                ███████████████████████████████████████████████████████
                                                 ██████████████████████████████████████
                                                ████████████████████████████████████████
                                               ██████████████████████████████████████████
                                               ██████████████████████████████████████████
                                              ██████████████████████████████████████████
                                             ██████████████████████████████████████████
                                            ██████████████████████████████████████████
                                           █████████████████████████████████████████
                                          █████████████████████████████████████████
                                         █████████████████████████████████████████
                                         ████████████████████████████████████████
                                        ████████████████████████████████████████
                                       ███████████████████████████████████████
                                      ███████████████████████████████████████
                                     ████████████████████████████████████
                                    ███████████████████████████████
                                   ██████████████████████████
                                   ████████████████████
                                  ██████████████
                                 █████████
                                ████
```
</details>


orhun:
super cool stuff! let's come back to this again after #2426 is finalized

joshka:
Since `line-clipping` is now in the Ratatui org, I think we should treat it as the home for reusable clipping / geometry algorithms.

We’ve already been using it for line clipping rather than implementing that directly in `ratatui-widgets`, and this PR is introducing similar algorithmic pieces (Sutherland–Hodgman clipping and scanline filling).

My preference would be to keep that pattern consistent and move those algorithms into the shared crate, keeping Ratatui focused on coordinate conversion, painter integration, and rendering.

I’m not blocked on merging this as-is, but I do think this logic likely belongs there long-term.

ratatui-widgets/Cargo.toml
```diff
@@ -61,6 +61,7 @@ strum.workspace = true
 time = { workspace = true, optional = true }
 unicode-segmentation.workspace = true
 unicode-width.workspace = true
+polygon-clipping = { path = "../../polygon-clipping" }
```
orhun: Is this a vendored crate? Is it [this one](https://crates.io/crates/polygon_clipping)?

Should we make it a part of the [line-clipping](https://github.com/ratatui/line-clipping) crate we have?

joshka: Seems like that crate is unmaintained, but also, yes. Please consider contributing stuff like this to line-clipping if it makes sense.

bananaofhappiness: oh, I didn't actually check (should've done this beforehand). No, it's not that crate. I made my own polygon-clipping crate with the exact same name, didn't publish it yet, was about to do so.

I was thinking about contributing this algorithm to line-clipping but I wasn't sure if it's appropriate since it's polygon clipping. If it's ok, I'll just contibute to line-clipping.

bananaofhappiness:
Sutherland-Hodgman clipping algorithm is now a part of [ratatui/line-clipping](https://github.com/ratatui/line-clipping), use it from that crate. I refactored code a little bit and made it cleaner. All tests pass.

orhun (approved):
LGTM, thanks!

I also added Bermuda Triangle as a part of the Canvas example!

Image

PR #2437, opened 2026-03-14
refactor(core): reintroduce `Cell::skip` as a deprecated field

junkdog:
as discussed in https://github.com/ratatui/ratatui/pull/1605#discussion_r2933338973 - this brings back `Cell::skip` as a deprecated field in order to avoid breaking the API  in a patch release. in terms of noise; the diff() is left pretty intact, but had to  `#[allow(deprecated)]` in a couple of places.

## `Cell::skip` vs `CellDiffOption` 

`CellDiffOption::ForcedWidth` takes precedence over `Cell::skip`, and `CellDiffOption::skip` is already skip - so `Cell::skip` can only override when `CellDiffOption::None` is set. i believe this is the correct behavior, but probably good to have another pair of eyes on it. @benjajaja maybe has some input too.

## PartialEq and Hash for `Cell::skip`

`Cell::skip` is part of `PartialEq` and `Hash` as normal fields. another option would be to treat `CellDiffOption::None` as `CellDiffOption::Skip` when `Cell::skip` is set, and only consider the effective `CellDiffOption` for the trait impls.

### DIFF
--- a/.github/workflows/check-pr.yml
+++ b/.github/workflows/check-pr.yml
@@ -29,7 +29,7 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       # Add comment indicating we require pull request titles to follow conventional commits specification
-      - uses: marocchino/sticky-pull-request-comment@773744901bac0e8cbb5a0dc842800d45e9b2b405 # v2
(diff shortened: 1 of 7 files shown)


junkdog:
good point, since it's scheduled for removal along with `Cell::skip`, i reverted the `set_skip()` so that it matches the old impl (and kept the deprecation).

PR #2447, opened 2026-03-17
feat(widgets): implement widget for char and Cell

wyvernbw:
This allows the rendering of `char` and `buffer::Cell` through the `Widget` trait without an additional `String` allocation.

Todo:
- [x] implementation
- [x] docs
- [ ] widget-ref?
- [ ] tests

### DIFF
--- a/ratatui-core/src/widgets/widget.rs
+++ b/ratatui-core/src/widgets/widget.rs
@@ -1,7 +1,7 @@
 use alloc::string::String;
 
-use crate::buffer::Buffer;
-use crate::layout::Rect;
+use crate::buffer::{Buffer, Cell};
+use crate::layout::{Position, Rect};
 use crate::style::Style;
 
 /// A `Widget` is a type that can be drawn on a [`Buffer`] in a given [`Rect`].
--- a/ratatui-core/src/widgets/widget.rs
+++ b/ratatui-core/src/widgets/widget.rs
@@ -105,6 +105,41 @@ impl<W: Widget> Widget for Option<W> {
     }
 }
 
+/// Renders a `Cell` as a widget.
+///
+/// This implementation enables an owned `Cell` to be treated as a widget, which can be printed
+/// directly into the [`Buffer`], at the position given by the [`Rect`].
+impl Widget for Cell {
+    fn render(self, area: Rect, buf: &mut Buffer)
+    where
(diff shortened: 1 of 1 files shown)


junkdog:
hi,

i have some reservations around `Cell` being a `Widget` - cells are the backing storage of the screen buffer. they are conceptually different types.

ratatui already provides `Widget for &str`, so `"r".render(...)` already works (no heap allocation). it might make sense to impl `Widget` for `char` too, but i'm not sure how useful it is to the wider audience. what's your use case?

wyvernbw:
Hi! My use case was around lifetimes, char is an owned value, &str has a lifetime. In the case you just call render there isnt a difference between them, but if you want to return an owned widget from a function for example you have to use a heap allocation.

I also think being able to render a char directly without having to do to_string is a nice to have api. I added Cell as well to support rendering a styled char (my thought process was that char/Cell would be analogous to &str/Span), but i can see why it wouldnt be desirable

junkdog:
`render()` consumes `self`, so the value is immediately written into the buffer. there's no borrowing issue since nothing outlives the call. `"r".render(...)` works the same as `'r'.render(...)` would here, both are `Copy` and have no lifetime concerns.

> return an owned widget from a function

this sounds a bit contrived without a specific use case. while it might make sense in an application, i don't think it is worth bringing into ratatui, since `Widget` for `&str` already covers this use case for all but a select few scenarios.

wyvernbw:
Here is a concrete example from my code (this doesnt have anything to do with lifetimes but it applies to my second point):
```rust
let char = text.chars().nth(i).unwrap_or(' ');
char.render(area, buffer);
// ^^^^^^^^ cant do this
```

I would say this looks like a reasonably common operation, but its up to you

themavik (commented):
nit: Widget for Cell/char only writes the single cell at area.origin; if area spans multiple cells the rest stay untouched, unlike typical rect-filling widgets.

wyvernbw:
would it be better to fill the entire rect with copies of the character instead?

orhun:
This seems like a nice-to-have addition to the API and not necessarily crucial. With that being said, I'm not entirely against the change but the provided use case doesn't seem very convincing. Maybe on the embedded side, this API would be more ergonomic/useful though.

PR #2450, opened 2026-03-21
feat(widgets): add `Paragraph::from_lines` for zero-copy rendering

tontinton:
Replaces the internal `text: Text<'a>` field with `lines: Cow<'a, [Line<'a>]>` so Paragraph can borrow lines without cloning.

`Paragraph::new()` still works the same, it just moves the lines out of Text into an owned Cow. The new `from_lines(&[Line])` constructor wraps the slice as `Cow::Borrowed`, skipping all the Vec/String/Span clones that `Into<Text>` would do.

Useful when you already have a `&[Line]` and just need to render or measure it, e.g. in hot loops that rebuild paragraphs every frame.

Used AI to learn the code, do a POC, and for the tests. The code was cleaned up to have minimal footprint.

### DIFF
--- a/ratatui-widgets/src/paragraph.rs
+++ b/ratatui-widgets/src/paragraph.rs
@@ -1,5 +1,7 @@
 //! The [`Paragraph`] widget and related types allows displaying a block of text with optional
 //! wrapping, alignment, and block styling.
+use alloc::borrow::Cow;
+
 use ratatui_core::buffer::Buffer;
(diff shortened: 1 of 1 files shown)


orhun:
Hello! 👋🏼
We usually first discuss these type of changes in issues before implementing them. Also, there were a couple of other issues/PRs where `Cow` was being used for performance gains, so I highly recommend reading the discussions there.

See #2017 (which I think is a duplicate of this one) and also #2027

PR #2452, opened 2026-03-21
fix(core): Make autoresize query the backend's window size rather than the cached value

OptimisticPeach:
In the default crossterm backend, this won't change anything, but the termwiz backend just [queries the cached size](https://github.com/ratatui/ratatui/blob/main/ratatui-termwiz/src/lib.rs#L261), which will incorrectly never change.

I would love to be able to change the termwiz backend `size` to effectively just be `window_size().columns_rows`, however [`Terminal::get_screen_size()`](https://docs.rs/termwiz/latest/termwiz/terminal/trait.Terminal.html#tymethod.get_screen_size) takes the receiver by `&mut`, and thus we'd need `size` to do so as well. 

Crossterm gets around this by not attaching the request for size to any state-tracking, and instead just calling a [standalone function](https://github.com/ratatui/ratatui/blob/main/ratatui-crossterm/src/lib.rs#L331).

### DIFF
--- a/ratatui-core/src/terminal/resize.rs
+++ b/ratatui-core/src/terminal/resize.rs
@@ -55,7 +55,7 @@ impl<B: Backend> Terminal<B> {
     pub fn autoresize(&mut self) -> Result<(), B::Error> {
         // fixed viewports do not get autoresized
         if matches!(self.viewport, Viewport::Fullscreen | Viewport::Inline(_)) {
-            let area = self.size()?.into();
+            let area = self.backend.window_size()?.columns_rows.into();
             if area != self.last_known_area {
                 self.resize(area)?;
             }


themavik (commented):
autoresize now reads backend.window_size() instead of Terminal::size(), so you react to fresh terminal geometry instead of a possibly stale cached area.

orhun (commented):
Hey, I just took a look at this and yeah, `termwiz`'s API seems to be limiting here. It would be much cleaner if we could fix this issue in `ratatui-termwiz` though.

With that being said, the changes here make sense. Still missing a regression test though :)

joshka:
I think this might be the wrong change - there's a note about what termwiz expects in check_for_resize that's sort of interesting: https://docs.rs/termwiz/latest/src/termwiz/terminal/buffered.rs.html#72-96

I wonder if we need to call that method when we get resize events?

OptimisticPeach:
Yeah, I was hesitant to keep going with this PR because I noticed it working rather oddly in my project... I'll be honest, I ended up switching to crossterm and everything worked perfectly.

orhun:
> I wonder if we need to call that method when we get resize events?

That's a good point, I think we should call that to keep the termwiz's surface in sync with the real terminal:

```diff
diff --git a/ratatui-termwiz/src/lib.rs b/ratatui-termwiz/src/lib.rs
index ddbbb6b4..a833aadf 100644
--- a/ratatui-termwiz/src/lib.rs
+++ b/ratatui-termwiz/src/lib.rs
@@ -263,6 +263,9 @@ impl Backend for TermwizBackend {
     }
 
     fn window_size(&mut self) -> io::Result<WindowSize> {
+        self.buffered_terminal
+            .check_for_resize()
+            .map_err(io::Error::other)?;
         let ScreenSize {
             cols,
             rows,
```

> Yeah, I was hesitant to keep going with this PR because I noticed it working rather oddly in my project... I'll be honest, I ended up switching to crossterm and everything worked perfectly.

Regarding the patch above, I think there is still something that we should fix here. Would you be open to investigate this a bit further to update this PR and contribute, or do you prefer that we take this one?

OptimisticPeach:
> Regarding the patch above, I think there is still something that we should fix here. Would you be open to investigate this a bit further to update this PR and contribute, or do you prefer that we take this one?

You know, I've spent enough time poking at ratatui/termwiz's internals while trying to get my project to work that I think I'd be happy to give it a go. I'll see if I can spare some time after Monday of next week 🙂

OptimisticPeach:
Hey, just an update since I realize I said I'd get to it after Monday (and seeing as it is now Wednesday) -- unfortunately some stuff came up last minute so it'll be a little while longer (uni student life is not merciful to someone trying to do OSS contributions).

orhun:
No worries, stuff happens :) Thanks for letting us know

orhun:
Any updates on this?

PR #2461, opened 2026-03-29
docs: improve terminal and setup docs for app authors

joshka:
## Summary

This updates the terminal and setup docs to better match the rendering behavior Ratatui actually implements, while also making the docs.rs path clearer for application authors.

The main docs changes are split into two commits:

1. `docs: align terminal docs with behavior`
   - align `Terminal`, viewport, frame, and flush docs with the real render pipeline
   - clarify the boundary between `Terminal::flush` and `Backend::flush`
   - improve examples for fullscreen, inline, and fixed viewport usage
2. `docs: improve terminal docs for app authors`
   - improve setup-path guidance across `ratatui`, `ratatui-core`, and backend crates
   - clarify viewport choice, escape hatches, and common edge cases
   - reduce duplication so crate-level docs explain choices while method docs hold detailed contracts

A small follow-up also fixes the crate-doc heading hierarchy in `ratatui-core` and
`ratatui-crossterm` so the generated READMEs no longer need
`markdownlint-disable-next-line heading-increment` suppressions.

## Verification

- `cargo test -p ratatui --doc`
- `cargo test -p ratatui-core --doc`
- `cargo test -p ratatui-crossterm --doc`
- `cargo test -p ratatui-termion --doc`
- `cargo test -p ratatui-termwiz --doc`
- `cargo test --workspace --all-targets --no-run`
- `cargo doc -p ratatui --no-deps`
- `cargo doc -p ratatui-core --no-deps`
- `cargo xtask readme --check`

Note: `cargo doc -p ratatui-core --no-deps` still reports pre-existing warnings outside this pass in `layout` and `text` docs. The terminal-doc warnings introduced during this work were fixed.

### DIFF
--- a/ratatui-core/src/terminal/viewport.rs
+++ b/ratatui-core/src/terminal/viewport.rs
@@ -73,6 +73,7 @@ pub enum Viewport {
     ///
     /// [`Terminal::new`]: crate::terminal::Terminal::new
     /// [`Terminal::draw`]: crate::terminal::Terminal::draw
+    /// [`Terminal::try_draw`]: crate::terminal::Terminal::try_draw
     #[default]
     Fullscreen,
     /// Draw the application inline with the rest of the terminal output.
@@ -90,7 +91,11 @@ pub enum Viewport {
     /// Inline viewports always span the full terminal width.
     ///
     /// For the full inline rendering model, including output inserted above the UI, see the
-    /// "Inline Viewport" section on [`Terminal`] and [`Terminal::insert_before`].
+    /// "Inline Viewport" section on [`Terminal`](crate::terminal::Terminal) and
+    /// [`Terminal::insert_before`](crate::terminal::Terminal::insert_before).
+    ///
+    /// [`Terminal::draw`]: crate::terminal::Terminal::draw
+    /// [`Terminal::try_draw`]: crate::terminal::Terminal::try_draw
     Inline(u16),
     /// Draw into a fixed region of the terminal.
     ///
--- a/ratatui-core/src/terminal/viewport.rs
+++ b/ratatui-core/src/terminal/viewport.rs
@@ -9,16 +9,49 @@ use crate::layout::Rect;
 /// For a higher-level overview of viewports in the context of an application (including
 /// examples), see [`Terminal`].
 ///
-/// Most applications use [`Viewport::Fullscreen`]. Use [`Viewport::Inline`] when you want to embed
-/// a UI into a larger CLI flow (for example: print some text, then start an interactive UI below
-/// it). Use [`Viewport::Fixed`] when you want Ratatui to render into a specific region of the
-/// terminal.
+/// Choose a viewport based on how the Ratatui UI should fit into the terminal:
+///
+/// - [`Viewport::Fullscreen`] for the standard case: your app owns the whole terminal surface.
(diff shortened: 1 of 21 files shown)


ratatui-core/src/terminal/viewport.rs
```diff
@@ -31,37 +64,56 @@ pub enum Viewport {
...
+    /// [`Terminal::draw`]: crate::terminal::Terminal::draw
+    /// [`Terminal::try_draw`]: crate::terminal::Terminal::try_draw
     Inline(u16),
     /// Draw into a fixed region of the terminal.
     ///
-    /// This can be useful when Ratatui is responsible for only part of the screen (for example, a
-    /// status panel beside another renderer), or when you want to manage the overall layout
-    /// yourself.
+    /// Choose this when Ratatui is responsible for only part of the screen, for example a panel in
+    /// a larger terminal layout managed by another renderer or by surrounding application code.
```
Logan-Ruf, line 103: 👍🏻

Logan-Ruf (approved):
As a newer user of the project myself, these changes LGTM.

I especially like the straightforward "Choose this when" additions to `enum Viewport`.

ratatui/src/lib.rs
```diff
@@ -92,7 +119,7 @@
 //! - **Backend crates**: [`ratatui-crossterm`], [`ratatui-termion`], [`ratatui-termwiz`]
 //! - **[`ratatui-macros`]**: Macros for simplifying the boilerplate
 //!
-//! **For application developers**: No changes needed - continue using `ratatui` as before.
+//! **For application developers**: `ratatui` remains the recommended starting point.
 //!
 //! **For widget library authors**: Consider depending on [`ratatui-core`] instead of the full
```
junkdog, line 124: should we drop the "widget" part here? it was just under the diff on line 80, which goes into more detail:

> //! - [`ratatui-core`] for widget libraries, custom integrations, and lower-level rendering
> //!   contracts

ratatui/src/lib.rs
```diff
@@ -219,6 +244,26 @@
 //! # fn handle_events() -> std::io::Result<bool> { Ok(false) }
 //! ```
 //!
+//! ### What happens if...
+//!
+//! - **If the terminal is resized:**<br> Ratatui does not redraw automatically when a resize event
```
orhun, line 249: nit: Mixing `<br>` with markdown seems off, I think we can use actual line breaks

P.S. good idea with this section!

joshka: This is intentional to make it layout neatly. I want the first line of the list to be on its own. Br is one of the necessary tools that are worth using now and again for markdown docs.

Markdown supports double space to make this work, but often editors / linters / diffs make that hard to maintain. Br is best of the bad options.

ratatui/src/lib.rs
joshka: One of the problems with docs is people read them non linearly. A little repetition doesn't hurt too much. I've tried to cut down the annoying amounts where possible. I think it's probably fine here. Wdyt?

junkdog: yeah, it's fine - i think it's the most common type of library too.

junkdog:
i think it looks good;  it does a better job at guiding the user around the api + makes sense to explain ratatui's double-buffer logic in greater detail.

orhun (approved):
YOLO

PR #2473, opened 2026-04-04
docs(canvas): add usage example for Line widget

ramadhan-dev-bright:
Added a doc-tested example for the Line struct in canvas to improve documentation and help new users understand how to render lines with color.

### DIFF
--- a/ratatui-widgets/src/canvas/line.rs
+++ b/ratatui-widgets/src/canvas/line.rs
@@ -8,14 +8,13 @@ use crate::canvas::{Painter, Shape};
 /// # Examples
 ///
 /// ```rust
-/// use crate::canvas::Line;
-///
-/// let red_line = Line::new(0.0, 0.0, 10.0, 10.0, ratatui_core::style::Color::Red);
-///
-/// let blue_line = Line::new(0.0, 10.0, 10.0, 0.0, ratatui_core::style::Color::Blue);
-///
-/// let mut green_line = Line::new(5.0, 5.0, 15.0, 15.0, ratatui_core::style::Color::Green);
-/// green_line.color = ratatui_core::style::Color::Yellow;
+/// # use ratatui_core::style::Color;
+/// # use ratatui_widgets::canvas::{Canvas, Line};
+/// Canvas::default().paint(|ctx| {
+///     ctx.draw(&Line::new(0.0, 0.0, 1.0, 0.0, Color::Red));
+///     ctx.draw(&Line::new(1.0, 0.0, 0.5, 1.0, Color::Red));
+///     ctx.draw(&Line::new(0.5, 1.0, 0.0, 0.0, Color::Red));
+/// });
 /// ```
 #[derive(Debug, Default, Clone, PartialEq)]
 pub struct Line {
--- a/ratatui-widgets/src/canvas/line.rs
+++ b/ratatui-widgets/src/canvas/line.rs
@@ -4,6 +4,15 @@ use ratatui_core::style::Color;
 use crate::canvas::{Painter, Shape};
 
 /// A line from `(x1, y1)` to `(x2, y2)` with the given color
+///
+/// #Examples
+/// 
+/// ```rust
+/// use ratatui::widgets::canvas::Line;
+/// use ratatui_core::style::Color;
+/// 
+/// let line = Line::new(0.0, 0.0, 10.0, 10.0, Color::Red);
+/// ```
 #[derive(Debug, Default, Clone, PartialEq)]
 pub struct Line {
     /// `x` of the starting point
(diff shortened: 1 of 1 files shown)


orhun:
Hey, thanks for the PR! The formatting check is failing, can you fix it?

Additionally, I think there are other methods we can demonstrate in the `Line` docs as well. Maybe consider taking a look at the API and preparing a more involved example.

ratatui-widgets/src/canvas/line.rs
```diff
@@ -4,6 +4,19 @@ use ratatui_core::style::Color;
...
+/// # Examples
+///
+/// ```rust
+/// use crate::canvas::Line;
+///
+/// let red_line = Line::new(0.0, 0.0, 10.0, 10.0, ratatui_core::style::Color::Red);
+///
+/// let blue_line = Line::new(0.0, 10.0, 10.0, 0.0, ratatui_core::style::Color::Blue);
+///
+/// let mut green_line = Line::new(5.0, 5.0, 15.0, 15.0, ratatui_core::style::Color::Green);
```
orhun: Please avoid explicit imports here (i.e. simply use `Red`, `Blue`, etc.)

orhun (commented):
cool, but I still think we can have a more involved example here.

Right now it feels not that useful to have this

ratatui-widgets/src/canvas/line.rs
ramadhan-dev-bright: Thanks for the feedback. I'm working on refining the implementation to meet the project's standards. Will update the PR shortly

ratatui-widgets/src/canvas/line.rs
```diff
@@ -4,6 +4,20 @@ use ratatui_core::style::Color;
...
+/// # use ratatui_core::style::Color;
+/// # use ratatui_widgets::canvas::{Canvas, Line};
+/// let canvas = Canvas::default()
+///     .x_bounds([-10.0, 10.0])
+///     .y_bounds([-10.0, 10.0])
+///     .paint(|ctx| {
+///         ctx.draw(&Line::new(-10.0, -10.0, 10.0, 10.0, Color::Red));
+///         ctx.draw(&Line::new(-10.0, 10.0, 10.0, -10.0, Color::Blue));
+///     });
+/// ```
```
orhun, line 18: This now reads like a Canvas example 😅

ramadhan-dev-bright: All CI checks are passing now. I've simplified the example and fixed the internal imports as suggested. Ready for review

ratatui-widgets/src/canvas/line.rs
```diff
@@ -4,6 +4,26 @@ use ratatui_core::style::Color;
 use crate::canvas::{Painter, Shape};
 
 /// A line from `(x1, y1)` to `(x2, y2)` with the given color
+///
+/// # Examples
+///
+/// ```rust
+/// # use ratatui_widgets::canvas::Line;
```
junkdog: this feels almost too short to be an example as it's just invoking the constructor; i'd probably merge it into the example below.

ratatui-widgets/src/canvas/line.rs
```diff
@@ -4,6 +4,26 @@ use ratatui_core::style::Color;
...
+/// let line = Line::new(0.0, 0.0, 10.0, 10.0, Color::Red);
+/// ```
+/// To draw it on a canvas:
+/// ```rust
+/// # use ratatui_widgets::canvas::{Canvas, Line};
+/// # use ratatui_core::style::Color;
+/// # let line = Line::new(0.0, 0.0, 10.0, 10.0, Color::Red);
+/// Canvas::default()
+///     .x_bounds([-10.0, 10.0])
+///     .y_bounds([-10.0, 10.0])
```
junkdog: if we're focusing on `Line` , overriding default bounds on the canvas adds a bit of noise to the example.

mostly an observation/nit

ratatui-widgets/src/canvas/line.rs
```diff
@@ -4,6 +4,26 @@ use ratatui_core::style::Color;
...
+///     .paint(|ctx| {
+///         ctx.draw(&line);
```
junkdog: suggestion: maybe draw a triangle here instead of just a single line, since it's just 2 extra ctx.draw() calls.

junkdog (commented):
looks pretty good to me overall - shows how to construct and draw a line -  just needs an extra revision or so.

ratatui-widgets/src/canvas/line.rs
ramadhan-dev-bright: Done. I've updated the example to draw a triangle and removed the extra bounds noise as suggested. All CI checks are green now

orhun (approved):

PR #2474, opened 2026-04-06
fix(examples): make line-gauge example compatible with macos sequoia's terminal.app

lazo4:
Part of the fix for #1972 

### Summary
This fix makes the `line-gauge` example compatible with the macos sequoia Terminal.app which doesn't support truecolor. It reuses the `is_true_color_supported` introduced in #2211 by @ffex and uses a color theme instead of hardcoding the colors.

### Result on macos sequoia
Before:
Capture d’écran 2026-04-06 à 11 47 07
After: 
Capture d’écran 2026-04-06 à 11 46 45

### Notes
This is my first open source contribution, thanks to @ffex for letting me help on this issue

### DIFF
--- a/ratatui-widgets/examples/line-gauge.rs
+++ b/ratatui-widgets/examples/line-gauge.rs
@@ -23,12 +23,76 @@ use ratatui::buffer::Buffer;
 use ratatui::layout::Constraint::{Length, Min};
 use ratatui::layout::{Layout, Rect};
 use ratatui::style::palette::tailwind;
-use ratatui::style::{Style, Stylize};
+use ratatui::style::{Color, Style, Stylize};
 use ratatui::widgets::{LineGauge, Paragraph, Widget};
+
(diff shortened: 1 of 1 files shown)


orhun (commented):
this looks good!

it's probably better if someone one MacOS could test it

@ratatui/maintainers

joshka (approved):
Looking at this implementation. I think I'd prefer to unconditionally using indexed colors instead of true color. Does that sound like a good simplification here?

Rationale:
The purpose of the examples is to generally demonstrate that particular widget / concept. Trying to add too many concepts in one example makes things more complicated than necessary

lazo4:
Yeah, that's a good idea, I hadn't thought of that, but I definitely agree with you. I guess i'll do that now.

lazo4:
@joshka, do you agree with my changes?

orhun (approved):

orhun:
thanks for the PR!

PR #2480, opened 2026-04-07
feat: add `CellDiffOption::AlwaysUpdate` to force cell updates even when content is identical

### LINKED ISSUE #1116: Allow bypassing diff and writing all to screen

sxyazi:
Follow-up for https://github.com/ratatui/ratatui/pull/1605

Trying to resolve https://github.com/ratatui/ratatui/issues/1116

### DIFF
--- a/ratatui-core/src/buffer/cell.rs
+++ b/ratatui-core/src/buffer/cell.rs
@@ -18,6 +18,12 @@ pub enum CellDiffOption {
     /// This is helpful when it is necessary to prevent the buffer from overwriting a cell that is
     /// covered by something from an escape sequence, such as graphics or links.
     Skip,
+    /// Always update this cell when diffing.
+    ///
+    /// This bypasses the equality check against the previous buffer. Use it when another
+    /// renderer may draw over the same area, such as an external image pipeline, so Ratatui can
+    /// redraw text there on the next render.
+    AlwaysUpdate,
     /// Force a width regardless of the symbol text width.
     ///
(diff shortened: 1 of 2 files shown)


orhun:
This looks fair tbh, have you tried it for your [use case](https://github.com/ratatui/ratatui/issues/1116#issuecomment-4197058237) yet?

sxyazi:
Yeah, it fits my use case well.

I hit a crash while testing against the main branch, but it's not related to this PR, so I'll make a separate bug report later.

orhun (approved):

PR #2481, opened 2026-04-09
feat(block): support shadows

### LINKED ISSUE #1892: Block shadows

orhun:
Introduce `Block::shadow(...)` with a new `Shadow` type that supports:

- presets: `overlay`,` block`, `light_shade`, `medium_shade`, `dark_shade`
- custom symbols via `Shadow::symbol(...)`
- custom effects via `Shadow::custom(...)`

```rust
use ratatui::layout::Offset;
use ratatui::style::Stylize;
use ratatui::widgets::{Block, Shadow};

let popup = Block::bordered().title("Popup").shadow(
    Shadow::dark_shade()
        .black()
        .on_white()
        .offset(Offset::new(2, 1)),
);
```

Results in:

```
┌Popup─────┐
│content   │▒
└──────────┘▒
  ▒▒▒▒▒▒▒▒▒▒▒
```

shadow

possible continuation of #1893
fixes #1892

For keeping the API simple, I have switched to a trait-based approach from `cell_filter` macro.


ratatui-widgets/examples/shadow.rs
```diff
@@ -0,0 +1,122 @@
...
+    );
+
+    render_block(
+        frame,
+        top_left,
+        "Overlay shadow",
+        Style::new().fg(Color::White).bg(Color::Blue),
+        Style::new().black().on_yellow(),
+        Shadow::overlay().style(Style::new().bg(Color::DarkGray)),
+    );
```
joshka: nit: for examples, it's probably worth just having a function per style here instead of DRYing up the parameters. The rationale is that a user can see how each of these params is meaningfully used.
Also, this suffers from the problem of passing the same type as two different params (background / style), it's not intuitive for the reader which is which just by reading it.

```suggestion
    render_overlay_shadow(frame, top_left);
```

ratatui-widgets/examples/shadow.rs
```diff
@@ -0,0 +1,122 @@
...
+    title: &str,
+    background_style: Style,
+    style: Style,
+    shadow: Shadow,
+) {
+    let background = Paragraph::new(background_text(area))
+        .block(Block::bordered())
+        .wrap(Wrap { trim: true })
+        .style(background_style);
+    frame.render_widget(background, area);
```
joshka, line 111: Contrary to the above, this part, which is not really the part that matters for the example would be reasonable to DRY out. Something like:

```suggestion
    render_background_paragraph(frame, area, Style::new(...));
```

ratatui-widgets/src/block/shadow.rs
```diff
@@ -0,0 +1,437 @@
...
+    }
+}
+
+/// A configurable shadow that can be rendered behind a [`Block`](crate::block::Block).
+///
+/// A [`Shadow`] is rendered in an offset area relative to the block. Its [`Style`] is applied
+/// first, then an optional cell filter can modify the affected cells, for example by filling them
+/// with a shading symbol or dimming the existing background.
+#[derive(Debug, Clone, Eq)]
+pub struct Shadow {
```
joshka: Put this first in the file - https://epage.github.io/dev/rust-style/#m-item-toc explains this well

ratatui-widgets/src/block/shadow.rs
```diff
@@ -0,0 +1,437 @@
...
+use ratatui_core::buffer::Buffer;
+use ratatui_core::layout::{Offset, Position, Rect};
+use ratatui_core::style::{Color, Modifier, Style, Styled};
+use ratatui_core::symbols::shade;
+use ratatui_core::widgets::Widget;
+
+/// A filter that modifies the cells covered by a [`Shadow`].
+///
+/// See [`Shadow::new`] for how to create a shadow from a custom filter.
+pub trait CellFilter: fmt::Debug {
```
joshka: We could possibly drop Debug here and have a custom impl for Shado instead of derived that just prints "Custom(...)" or something.

joshka: probably put this after Filter in the file

ratatui-widgets/src/block/shadow.rs
```diff
@@ -0,0 +1,437 @@
...
+    fn apply(&self, shadow_area: Rect, base_area: Rect, buf: &mut Buffer);
+}
+
+/// The built-in shadow filters.
+#[derive(Debug, Clone)]
+enum Filter {
+    /// Applies no symbol changes and only keeps the shadow style.
+    Overlay,
+    /// Fills the shadow area with a single symbol.
+    Symbol(&'static str),
```
joshka, line 72: I wonder if this should be Span/Line/Text (or `Into<Cow<str>>`)

ratatui-widgets/src/block/shadow.rs
```diff
@@ -0,0 +1,437 @@
...
+    /// Applies a user-defined shadow filter.
+    Custom(Arc<dyn CellFilter>),
```
joshka: I wonder if it makes sense for this to be `Box<>` instead and drop the Clone on Filter?

ratatui-widgets/src/block/shadow.rs
```diff
@@ -0,0 +1,437 @@
...
+                "symbol".hash(state);
+                symbol.hash(state);
+            }
+            Self::Custom(filter) => {
+                "custom".hash(state);
+                ptr::hash(Arc::as_ptr(filter), state);
+            }
+        }
+    }
+}
```
joshka: I wonder if making these PartialEq/Eq/Hash is actually necessary for any use case that matters?

ratatui-widgets/src/block/shadow.rs
```diff
@@ -0,0 +1,437 @@
...
+
+/// A configurable shadow that can be rendered behind a [`Block`](crate::block::Block).
+///
+/// A [`Shadow`] is rendered in an offset area relative to the block. Its [`Style`] is applied
+/// first, then an optional cell filter can modify the affected cells, for example by filling them
```
joshka: Would "effect" be a better description of this than filter? I'd often use the term filter in many programming senses as choosing items to keep / discard. The term filter in the sense of a graphics filter is often shorthand for filter effect.

ratatui-widgets/src/block/shadow.rs
```diff
@@ -0,0 +1,437 @@
...
+    ///                     buf[(x, y)].set_symbol("░");
+    ///                 }
+    ///             }
+    ///         }
+    ///     }
+    /// }
+    ///
+    /// let shadow = Shadow::new(Checker);
+    /// ```
+    pub fn new<F: CellFilter + 'static>(filter: F) -> Self {
```
joshka: It may be worth naming this custom (and moving it to be after the overlay / symbol constructors as this is less likely to be used in comparison.

joshka: I think this example is probably a little larger than we'd want at the function level - perhaps move to the type with some discussion about custom filters?

ratatui-widgets/src/block/shadow.rs
```diff
@@ -0,0 +1,437 @@
...
+    ///
+    /// This leaves the existing cell symbols unchanged.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use ratatui::style::Stylize;
+    /// use ratatui::widgets::{Block, Shadow};
+    ///
+    /// let block = Block::bordered().shadow(Shadow::overlay().black().on_white());
```
joshka: nit: would probably be worth busting shadow out to a variable here so that it's simple and obvious on one line. The block call shows how the resulting shadow is used.

ratatui-widgets/src/block/shadow.rs
```diff
@@ -0,0 +1,437 @@
...
+    /// Creates a shadow filled with full block symbols.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use ratatui::widgets::{Block, Shadow};
+    ///
+    /// let block = Block::bordered().shadow(Shadow::block());
+    /// ```
+    pub fn block() -> Self {
```
joshka, line 174: It's worth noting the presets in the main Shadow comment

ratatui-widgets/src/block/shadow.rs
```diff
@@ -0,0 +1,437 @@
...
+
+    /// Creates a shadow filled with full block symbols.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use ratatui::widgets::{Block, Shadow};
+    ///
+    /// let block = Block::bordered().shadow(Shadow::block());
+    /// ```
```
joshka, line 173: Would be good to render a small text block that shows how this looks.

ratatui-widgets/src/block/shadow.rs
```diff
@@ -0,0 +1,437 @@
...
+    /// assert_eq!(
+    ///     shadow,
+    ///     Shadow::overlay().style(Style::new().black().on_white())
+    /// );
+    /// ```
+    #[must_use]
+    pub fn style<S: Into<Style>>(mut self, style: S) -> Self {
+        self.style = style.into();
+        self
+    }
```
joshka: This is a bit of a weird example for this method as it's testing some other input matches the result of this call.

ratatui-widgets/src/block.rs
joshka, line 65: Add an entry here for shadow

joshka (commented):
A bunch of small design questions and a few nits that will make this incrementally easier to maintain longterm

ratatui-widgets/src/block/shadow.rs
orhun: This can't be done because `Block` is Clone etc. so we need Clone

ratatui-widgets/src/block/shadow.rs
orhun: see my other comment

ratatui-widgets/src/block/shadow.rs
orhun: I think this example is unnecessary so I removed it

ratatui-widgets/src/block/shadow.rs
orhun: I think Span/Line/Text might be a bit overkill here. I honestly like the simplicity of a static str here but the idea of having `Into<Cow<str>>` instead is interesting.

orhun: I think we could always change it later. For now let's go with static str

joshka (approved):

PR #2484, opened 2026-04-11
fix(crossterm): track cursor position after wide cells to prevent misalignment

### LINKED ISSUE #2357: Rendering bug with emojis in list
ratatui = "0.30.0"
crossterm = "0.29.0"
Running on macOS Sequioia

Repro:
Create a scrolling list with the emoji '🚀️' in each line.
Scroll the list

Observe:
Rendering bugs.
a) Before scrolling spaces are inserted/removed randomly around the emoji, 
b) scrolling sometimes causes characters from other lines to appear. 

Here's a screenshot of an initial list before any scrolling has taken place. Notice the spacing between the emoji and the characters around it. These should be consistent, but instead have random spaces before and after.

Image

Next screenshot, after moving the selection down the list. In this instance, the scrolling fixed the spacing. However, sometimes it introduces other artifacts, of letters from the line above or below being preserved instead of being cleared.
Image

Removing the emoji causes the issue to go away.

I suspect that the bug has to do with incorrect length calculations on this multi-byte unicode character, whose value is U+1F680.

larsch:
Fixes #2357

## Problem

When rendering wide graphemes (emoji, CJK characters), the terminal cursor lands at `x + cell_width` after a `Print`, not `x + 1`. The previous code tracked `last_pos` as the *starting* column of the most recently printed cell, so it would incorrectly suppress a `MoveTo` for the very next cell — leaving the cursor one column short, causing:

- Random spacing artifacts around emoji in lists
- Ghost characters from other lines appearing during scroll

## Solution

Replace `last_pos` (start of last cell) with `next_pos` (expected cursor position after the last cell = `x + cell_width`). The `MoveTo` is now skipped only when the terminal cursor is genuinely already at the target position.

Also removed an unnecessary `as u16` cast caught by Clippy (`cell_width()` already returns `u16`).

### DIFF
--- a/ratatui-crossterm/src/lib.rs
+++ b/ratatui-crossterm/src/lib.rs
@@ -800,6 +800,134 @@ mod tests {
 
     use super::*;
 
+    /// Build a `Cell` whose symbol is `symbol`.
+    fn cell(symbol: &str) -> Cell {
+        let mut c = Cell::EMPTY;
+        c.set_symbol(symbol);
+        c
+    }
+
+    /// Collect the raw bytes queued by `Backend::draw` for the given content iterator.
+    fn draw_bytes<'a>(content: impl Iterator<Item = (u16, u16, &'a Cell)>) -> Vec<u8> {
+        let mut backend = CrosstermBackend::new(Vec::<u8>::new());
+        backend.draw(content).unwrap();
+        backend.writer
+    }
+
+    /// Return the ANSI sequence for `MoveTo(x, y)`.
+    fn move_to(x: u16, y: u16) -> Vec<u8> {
+        let mut buf = Vec::new();
+        queue!(buf, MoveTo(x, y)).unwrap();
+        buf
+    }
+
+    /// Return the ANSI sequence for `Print(s)`.
+    fn print(s: &str) -> Vec<u8> {
+        let mut buf = Vec::new();
+        queue!(buf, Print(s)).unwrap();
+        buf
+    }
+
+    mod draw_cursor_tracking {
+        use super::*;
+
+        /// After a narrow (width-1) cell at (x, y), the next cell at (x+1, y)
+        /// must not emit an extra MoveTo — the cursor is already there.
+        #[test]
+        fn narrow_cell_no_extra_move() {
+            let narrow = cell("A");
+            let next = cell("B");
+            let bytes = draw_bytes([(0, 0, &narrow), (1, 0, &next)].iter().copied());
+
+            // MoveTo(1,0) must NOT appear after the first Print
+            let move_to_1_0 = move_to(1, 0);
+            // Locate the Print("A") bytes and check nothing after them starts with MoveTo(1,0)
+            let print_a = print("A");
+            let pos = bytes
+                .windows(print_a.len())
+                .position(|w| w == print_a.as_slice())
+                .expect("Print(A) not found");
+            let after_a = &bytes[pos + print_a.len()..];
+            assert!(
+                !after_a.starts_with(&move_to_1_0),
+                "unexpected MoveTo(1,0) after narrow cell: {bytes:?}"
+            );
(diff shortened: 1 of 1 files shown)


sxyazi:
FWIW, there's some context about why that was hardcoded as `x + 1`, https://github.com/ratatui/ratatui/issues/1396#issuecomment-2390394892

a-kenji:
In which terminals was this fix tested? I have my doubts that this actually works for all terminals.

larsch:
@a-kenji You were right. This ended up being a `foot` specific solution. I should have checked.

Without change: `foot`, `kitty` show glitchy spacing after emoji + VS16. `wezterm`, `alacritty`, and `gnome-terminal` have stable rendering, albeit `alacritty` doesn't really do emojis well.

With change: `foot` has stable rendering. `kitty` has stable alignment, but the actual emoji sometimes do not appear. `wezterm`, `gnome-terminal`, `alacritty`, now have glitchy rendering.

larsch:
This is not a general solution.

a-kenji:
Thank you for checking. That is the issue why it is not fixed in the general case yet. The current solution tries to work for most terminals in the most acceptable way.

smndvgnc:
@larsch 

I hit this problem with `Ghostty`. I had `CC` diagnose and propose a first fix which turned out to be exactly what you proposed initially. But then I had the case of an emoji not appearing, which apparently you faced also with `kitty`.  After a second round, the fix was enhanced, solving also my second problem; I post it below, trying to help.

I'm pretty sure this isn't still a universal fix but, for my case and in my environment, it seems to solve 100% of my issues.

```diff
--- a/ratatui-crossterm/src/lib.rs
+++ b/ratatui-crossterm/src/lib.rs
@@ use ratatui_core::backend::{Backend, ClearType, WindowSize};
-use ratatui_core::buffer::Cell;
+use ratatui_core::buffer::{Cell, CellWidth};
 use ratatui_core::layout::{Position, Size};
 use ratatui_core::style::{Color, Modifier, Style};
@@ fn draw
         let mut modifier = Modifier::empty();
+        // Real cursor position after the last printed cell. We track the actual display
+        // width (`cell_width`) rather than the logical column: a wide cell advances the
+        // cursor by 2, so the trailing cell of a VS16 emoji (re-emitted at `x + 1` by the
+        // buffer diff) is *not* contiguous with the real cursor (which sits at `x + 2`).
+        // See https://github.com/ratatui/ratatui/issues/2357.
         let mut last_pos: Option<Position> = None;
         for (x, y, cell) in content {
-            // Move the cursor if the previous location was not (x - 1, y)
-            if !matches!(last_pos, Some(p) if x == p.x + 1 && y == p.y) {
+            // A wide glyph already covers the cell(s) to its right. The buffer diff re-emits
+            // the trailing cell of a VS16 emoji (a space) at `x + 1` to clear terminals that
+            // fail to erase it, but on terminals that render the emoji across two cells (the
+            // width model ratatui assumes) printing into that trailing cell destroys the
+            // glyph. Skip any cell that falls within the span of the glyph just printed
+            // (same row, column before the real cursor); the glyph already paints it.
+            if matches!(last_pos, Some(p) if y == p.y && x < p.x) {
+                continue;
+            }
+            // Move the cursor unless it is already exactly at (x, y).
+            if !matches!(last_pos, Some(p) if x == p.x && y == p.y) {
                 queue!(self.writer, MoveTo(x, y))?;
             }
-            last_pos = Some(Position { x, y });
             if cell.modifier != modifier {
                 let diff = ModifierDiff {
                     from: modifier,
@@ end of loop body
             queue!(self.writer, Print(cell.symbol()))?;
+            // The terminal advances the cursor by the cell's display width.
+            last_pos = Some(Position {
+                x: x.saturating_add(cell.cell_width()),
+                y,
+            });
         }
```

PR #2486, opened 2026-04-12
fix(crossterm): avoid spurious MoveTo after multi-width cells

cammeresi:
The contiguity check in `draw` assumed the cursor advances by one column after every cell, but wide characters (CJK, emoji) advance by two or more columns.  Thus an unnecessary MoveTo was emitted after every wide character.


cammeresi:
After posting this patch, I accidentally discovered that the original code is desired so as to improve rendering in broken terminals.  Also, recent discussion about tracking the end rather than the start seems more likely to be correct.

PR #2488, opened 2026-04-12
fix(block): saturate block edge arithmetic

joshka:
### Motivation

- Block border and title layout used unchecked `u16` arithmetic in several places.
- In debug builds that can panic on tiny or edge-case geometry; in release builds the same arithmetic wraps.
- The original report came from merge-border rendering on tiny areas, but the same pattern appeared in title layout and spacing helpers as well.

### Description

- Use saturating arithmetic in `Block::inner`, `render_sides`, `render_corners`, `titles_area`, and `vertical_space`.
- Clamp rendered title widths to `u16` for layout arithmetic.
- Replace unchecked title-width accumulation and cursor-advance math with bounded arithmetic.
- Add debug-only regression tests covering empty areas, maximal padding, title-area edge cases, and very large title widths.

### Testing

- Ran `cargo test -p ratatui-widgets block::tests`.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69dbd35ed38c833094a9fa56b239f0d0)

### DIFF
--- a/ratatui-widgets/src/block.rs
+++ b/ratatui-widgets/src/block.rs
@@ -751,12 +751,10 @@ impl<'a> Block<'a> {
         inner.x = inner.x.saturating_add(self.padding.left);
         inner.y = inner.y.saturating_add(self.padding.top);
 
-        inner.width = inner
-            .width
-            .saturating_sub(self.padding.left + self.padding.right);
-        inner.height = inner
-            .height
-            .saturating_sub(self.padding.top + self.padding.bottom);
+        let horizontal_padding = self.padding.left.saturating_add(self.padding.right);
+        let vertical_padding = self.padding.top.saturating_add(self.padding.bottom);
+        inner.width = inner.width.saturating_sub(horizontal_padding);
+        inner.height = inner.height.saturating_sub(vertical_padding);
 
         inner
     }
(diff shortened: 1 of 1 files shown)


ratatui-widgets/src/block.rs
```diff
@@ -806,8 +806,10 @@ impl Block<'_> {
         let is_replace = self.merge_borders != MergeStrategy::Replace;
         let left_inset = left + u16::from(is_replace && self.borders.contains(Borders::LEFT));
         let top_inset = top + u16::from(is_replace && self.borders.contains(Borders::TOP));
```
joshka: I'm not sure if these additions would possibly also have overflows - need to check

ratatui-widgets/src/block.rs
```diff
@@ -1040,16 +1050,21 @@ impl Block<'_> {
             .map(|(_, line)| line)
     }
 
+    /// Return the rendered line width clamped to `u16` for layout arithmetic.
+    fn line_width_u16(line: &Line<'_>) -> u16 {
+        line.width().min(u16::MAX as usize) as u16
+    }
```
orhun, line 1056: I assume we are using `CellWidth` here, would it maybe make sense to add a new trait function such as `width_clamped` instead?

joshka: I think it's probably not worth the hassle IMO. I haven't seen this needed in other places yet. (YAGNI)

orhun: fair enough

orhun (approved):

PR #2499, opened 2026-04-17
fix(core): use correct width for halfwidth dakuten/handakuten

orhun:
unicode-width reports U+FF9E/U+FF9F as zero-width, but terminals render them as 1 cell.
Adjusts `CellWidth` trait accordingly for fixing this behavior.

fixes #2188

### DIFF
--- a/ratatui-core/src/buffer/buffer.rs
+++ b/ratatui-core/src/buffer/buffer.rs
@@ -971,18 +971,19 @@ mod tests {
 
         // Combining dakuten (U+3099): forms 1-cell grapheme cluster with width 1
         let mut buffer1 = Buffer::empty(area);
-        buffer1.set_string(0, 0, "ｶ゙", Style::default());
+        let (x1, _) = buffer1.set_stringn(0, 0, "ｶ゙", usize::MAX, Style::default());
         // The combining dakuten merges with ｶ into a single cell (width 1)
         assert_eq!(buffer1.content[0].symbol(), "ｶ゙");
-        assert_eq!(buffer1.content[1].symbol(), " ");
+        assert_eq!(buffer1.content[0].cell_width(), 1);
+        assert_eq!(x1, 1);
 
         // Non-combining halfwidth dakuten (U+FF9E): grapheme cluster with width 2
         let mut buffer2 = Buffer::empty(area);
-        buffer2.set_string(0, 0, "ｶﾞ", Style::default());
+        let (x2, _) = buffer2.set_stringn(0, 0, "ｶﾞ", usize::MAX, Style::default());
         // The grapheme cluster "ｶﾞ" is stored in cell[0], but takes 2 cells width
         assert_eq!(buffer2.content[0].symbol(), "ｶﾞ");
-        assert_eq!(buffer2.content[1].symbol(), " "); // reset cell (hidden by width 2)
-        assert_eq!(buffer2.content[2].symbol(), " ");
+        assert_eq!(buffer2.content[0].cell_width(), 2);
+        assert_eq!(x2, 2);
     }
 
     #[fixture]
--- a/ratatui-core/src/buffer/buffer.rs
+++ b/ratatui-core/src/buffer/buffer.rs
@@ -914,6 +914,77 @@ mod tests {
         assert_eq!(buffer, Buffer::with_lines(["コン "]));
     }
 
+    #[test]
+    fn set_string_halfwidth_katakana_with_dakuten() {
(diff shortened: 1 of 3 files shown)


ratatui-core/src/buffer/buffer.rs
```diff
@@ -914,6 +914,77 @@ mod tests {
...
+        // The combining dakuten merges with ｶ into a single cell (width 1)
+        assert_eq!(buffer1.content[0].symbol(), "ｶ゙");
+        assert_eq!(buffer1.content[1].symbol(), " ");
+
+        // Non-combining halfwidth dakuten (U+FF9E): grapheme cluster with width 2
+        let mut buffer2 = Buffer::empty(area);
+        buffer2.set_string(0, 0, "ｶﾞ", Style::default());
+        // The grapheme cluster "ｶﾞ" is stored in cell[0], but takes 2 cells width
+        assert_eq!(buffer2.content[0].symbol(), "ｶﾞ");
+        assert_eq!(buffer2.content[1].symbol(), " "); // reset cell (hidden by width 2)
```
junkdog: maybe check against `CellDiffOption::Skip` instead of content (which is skipped)? asserts for buffer1 and 2 otherwise look very similar, except for the comments.

ratatui-core/src/buffer/cell_width.rs
```diff
@@ -56,4 +96,65 @@ mod tests {
     fn empty() {
         assert_eq!("".cell_width(), 0);
     }
+
+    #[test]
+    fn halfwidth_dakuten_alone() {
+        assert_eq!(width("\u{FF9E}"), 1); // ﾞ
```
junkdog: nit: maybe use the `HALFWIDTH_KATAKANA_VOICED_SOUND_MARK`  and `HALFWIDTH_KATAKANA_SEMI_VOICED_SOUND_MARK` for the test below

ratatui-core/src/buffer/cell_width.rs
```diff
@@ -56,4 +96,65 @@ mod tests {
...
+    #[allow(clippy::unicode_not_nfc)]
+    fn combining_handakuten_no_special_handling() {
+        // Combining handakuten (U+309A) should follow unicode-width behavior.
+        assert_eq!(width("ﾊ゚"), 1); // U+FF8A + U+309A
+        assert_eq!(width("パ"), 2); // U+30CF + U+309A
+    }
+
+    #[test]
+    fn mixed_text_unchanged() {
+        assert_eq!(width("a"), 1);
```
junkdog, line 153: can kill the ascii() test which performs the same check

ratatui-widgets/src/paragraph.rs
```diff
@@ -989,6 +989,108 @@ mod tests {
...
+    #[test]
+    fn test_render_paragraph_with_fullwidth_and_background() {
+        // "あいう" should be width 6 (each character is width 2)
+        let text = "あいう";
+        let paragraph = Paragraph::new(text).style(Style::default().bg(Color::Green));
+
+        let mut buffer = Buffer::empty(Rect::new(0, 0, 10, 1));
+        paragraph.render(Rect::new(0, 0, 10, 1), &mut buffer);
+
+        // Check content - should have spaces for skip cells
```
junkdog: nit but checking if skipped cells are " "-strings i think makes little sense to the casual reader, since normal, empty cells share the same " " representation but are rendered.  i'm in camp `CellDiffOption`.

junkdog (approved):
from what i can tell it's correct, but this was new, unfamiliar unicode territory

ratatui-core/src/buffer/buffer.rs
orhun: actually those cells are not skipped, so the test would become:

```rust
        // The combining dakuten merges with ｶ into a single cell (width 1)
        assert_eq!(buffer1.content[0].symbol(), "ｶ゙");
        assert_eq!(buffer1.content[0].cell_width(), 1);
        assert_eq!(buffer1.content[1].diff_option, CellDiffOption::None);
```

I pushed d2ba26096a85b53aa8f9bf8f4b5d5e966612aa58 which makes this test more clear I believe

ratatui-widgets/src/paragraph.rs
orhun: 84163afcf39b060450a7b904ba3fc064adf735e3

also see my comment above

junkdog: yup, look goods

PR #2501, opened 2026-04-20
chore(toml): migrate from taplo to tombi

joshka:
## Summary

This migrates the repo's TOML tooling from [Taplo](https://github.com/tamasfe/taplo) to [Tombi](https://github.com/tombi-toml/tombi).

The main motivation is maintenance and installability:

- Ratatui previously used [Taplo](https://github.com/tamasfe/taplo) for TOML formatting.
- Taplo's maintenance future has been uncertain; see <https://github.com/tamasfe/taplo/issues/715>.
- Taplo's current release artifacts are not friendly to `cargo-binstall`, which means some setups fall back to building from source.
- That source-build path is awkward for CI and was also a blocker while exploring a future Docker/devcontainer setup because of image size and install cost.

[Tombi](https://github.com/tombi-toml/tombi) is actively maintained, publishes prebuilt binaries, and has a VS Code extension that lets us keep editor behavior aligned with CI.

### Tooling and CI

- Replace `taplo` with `tombi` in `cargo xtask format`.
- Use [`tombi-toml/setup-tombi`](https://github.com/tombi-toml/setup-tombi) in the formatting CI job.
- Replace `.taplo.toml` with `tombi.toml`.

### Formatting policy

- Keep the existing 100-column TOML line width.
- Disable schema-driven top-level table ordering for `Cargo.toml`.
- Disable the corresponding `tables-out-of-order` lint for `Cargo.toml`.

Why disable Cargo schema ordering:

- Tombi can order keys and tables according to schema metadata; see <https://tombi-toml.github.io/tombi/docs/comment-directive/tombi-value-directive/#format-rules-table-keys-order>.
- The schema order moved sections like `[features]` below dependency sections.
- That made feature-gated behavior harder to scan.
- It also created noisy migration churn that was not central to the tool switch itself.

Why disable the Cargo lint warning:

- The VS Code warning came from Tombi's `tables-out-of-order` lint rather than from formatting.
- Since we are intentionally preserving the existing Cargo manifest section order, we also disable that lint for `Cargo.toml` to keep diagnostics aligned with the formatter configuration.
- Related docs: <https://tombi-toml.github.io/tombi/docs/configuration/>.

### Installation note

- Tombi is better positioned for binary installs than Taplo today, but it is still not fully in the generic `cargo-binstall` / `taiki-e/install-action` path.
- Relevant upstream context:
  - binary naming / generic installer compatibility: <https://github.com/tombi-toml/tombi/issues/1164>
  - crates.io publishing for `cargo-binstall` fallback: <https://github.com/tombi-toml/tombi/issues/1686>
  - earlier crates.io publishing discussion: <https://github.com/tombi-toml/tombi/issues/632>
- In practice, this still leaves some rough edges around the generic Rust installer path, but Tombi already has usable prebuilt binaries and an official installer action, which is a better place for Ratatui than Taplo's current install story.
- CI now uses [`tombi-toml/setup-tombi`](https://github.com/tombi-toml/setup-tombi) with a pinned version, which avoids the GitHub token path by not asking the action to resolve `latest` at runtime.

### Docs and editor setup

- Update contributor docs to point to Tombi's repo, docs, and installation guide.
- Note that the repo previously used Taplo and link the upstream maintenance discussion at <https://github.com/tamasfe/taplo/issues/715>.
- Add VS Code extension recommendations for [`rust-lang.rust-analyzer`](https://marketplace.visualstudio.com/items?itemName=rust-lang.rust-analyzer) and [`tombi-toml.tombi`](https://marketplace.visualstudio.com/items?itemName=tombi-toml.tombi).
- Add committed VS Code workspace settings for shared repo defaults, mainly so the workspace can define formatter behavior that matches CI, including TOML formatter selection and nightly rustfmt args.
- Add a dedicated formatting section to `CONTRIBUTING.md` covering Rust formatting, TOML formatting, nightly `rustfmt`, and the VS Code workspace override workaround.

## Why Commit VS Code Settings

This PR adds committed VS Code workspace settings for shared repo defaults only.

The goal is to:

- recommend the expected extensions
- make local editor formatting behave more like CI
- avoid formatter drift between editor usage and the repo's checked-in tooling

This is not intended to standardize everyone's full editor setup.

## Nightly rustfmt

The repo already uses unstable rustfmt options in [`rustfmt.toml`](./rustfmt.toml), so formatting Rust code with `cargo xtask format` works best with nightly Rust. This is already how CI behaves.

The docs now call this out more clearly:

- install nightly if you want local formatting to match CI
- the reason is to pick up the unstable formatting options configured in `rustfmt.toml`

For contributors who cannot or do not want to install nightly, the docs also point to a workaround: use a personal VS Code workspace file with override settings instead of changing the tracked workspace config.

## Notes From Implementation

- Tombi `v0.9.19` added config support to disable schema-defined ordering per schema, which let us preserve existing `Cargo.toml` section order cleanly.
- The `tables-out-of-order` warning seen in VS Code came from Tombi lint, not formatting, so that needed a separate config change.
- After disabling Cargo section ordering, the remaining TOML churn was much smaller and mostly mechanical.

## Files of Interest

- `.github/workflows/ci.yml`
- `xtask/src/commands/format.rs`
- `tombi.toml`
- `.vscode/settings.json`
- `.vscode/extensions.json`
- `CONTRIBUTING.md`

## Verification

- `cargo metadata --format-version 1`
- `markdownlint-cli2 CONTRIBUTING.md`


.github/workflows/ci.yml
```diff
@@ -39,7 +39,7 @@ jobs:
       - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2
       - uses: taiki-e/install-action@85b24a67ef0c632dfefad70b9d5ce8fddb040754 # v2
         with:
-          tool: taplo-cli
+          tool: tombi-cli
```
joshka: I suspect this will fail right now :/

.gitignore
```diff
@@ -4,3 +4,4 @@ target
 .gdb_history
 .idea/
 .env
+*.code-workspace
```
joshka, line 7: Not sure whether to keep this - I had a bunch of settings in my vscode.settings file, which I moved to a workspace settings file which is ignored for now. I can also just use .git/info/exclude for this.:shrug:

orhun: yeah, I'd prefer not adding an entry for this and just use `.git/info/exclude`

orhun (approved):
LGTM

tested locally, also adopted the arch package: https://archlinux.org/packages/extra/x86_64/tombi/

PR #2519, opened 2026-05-03
feat(widgets): add `Fill` widget

### LINKED ISSUE #1146: Add a Buffer::fill method
## Problem
I was playing with some ideas and CoPilot suggested that I call the non-existent `Buffer::fill(area, ".", style)` method
This seems like it's probably a useful simple building block method to have that avoids putting that loop in the caller.
I'd see this used internally for block borders, barchart, chart, scrollbars, etc. and any custom widgets that need to fill an area.

## Solution
Add a fill method that accepts an area, a symbol, and a style.
A naive approach runs a loop over the area and writes the symbols appropriately (taking account of unicode width).
It might be possible to special case single width characters by using slice::fill

It might also be worth considering whether this allocates new cells and fills the buffer vs calling the set_symbol / set_style methods (in the same way that https://github.com/ratatui-org/ratatui/pull/1143#pullrequestreview-2079128188 needs to take this into account).

## Additional context
This is an idea for discussion rather than something ready to implement.

Metbcy:
## Description

Adds a `Fill` widget that paints every cell in its render area with a single repeated symbol and style. Useful as a building block for backgrounds, separators, scrollbar tracks, and custom widgets — anywhere you'd otherwise hand-roll the nested loop over an area.

```rust
Fill::new("X").blue().bold().render(area, buf);
```

### Why a widget instead of `Buffer::fill`

The first revision of this PR added a `Buffer::fill(area, symbol, style)` method. Per @joshka's [review comment](https://github.com/ratatui/ratatui/pull/2519#issuecomment-4365294724), the widget shape composes more naturally with `Stylize` and is easier to type:

> `Fill("X").blue().bold().render(...)` is probably a bit easier to type than `buf.fill("X", Style::new().blue().bold())`

Reverted the buffer method and reshaped the change as a widget. `Fill` implements `Styled` so the full `Stylize` shorthand surface is available (`.blue()`, `.bold()`, `.bg(...)`, etc.).

### API

- `Fill::new(symbol)` — accepts `impl Into<Cow<'a, str>>`, so both `&str` and `String` work
- `.style(impl Into<Style>)` — explicit style setter (also via `Stylize` shorthands)
- `.symbol(impl Into<Cow<'a, str>>)` — replace the symbol
- Out-of-bounds cells are silently clipped, mirroring `Clear`

## Testing

- 7 unit tests in `ratatui-widgets/src/fill.rs`:
  - `fills_area_with_symbol_and_style`
  - `clips_area_to_buffer` (area extends past right/bottom)
  - `render_fully_out_of_bounds_is_noop`
  - `renders_with_offset_buffer_area` (non-zero-origin buffer)
  - `stylize_shorthand_works` (`.blue().bold()` end-to-end)
  - `accepts_owned_string_symbol`
  - `symbol_setter_replaces_symbol`
- 2 doctests on `Fill` and `Fill::new`
- `cargo test -p ratatui-widgets`: 233 passed, 0 failed
- `cargo build -p ratatui` clean

## Issue(s)

Closes #1146

## Did you use AI to generate any of the content in this PR?

Yes — Claude (via the Hermes Agent harness) drafted the implementation and tests; I reviewed every line and the test output before pushing.


joshka:
I think the `Fill` widget approach seemed like it was the better idea to me in that issue (it's a while back though, so I'm just interpreting what I wrote there). I think either work, `Fill("X").blue().bold().render(...)` is probably a bit easier to type than `buf.fill("X", Style::new().blue().bold())`;

Metbcy:
Thanks @joshka — agreed, the widget shape reads better. Reshaped: reverted the `Buffer::fill` method and added `Fill` as a widget in `ratatui-widgets` (re-exported from `ratatui::widgets`).

Now reads as:

```rust
Fill::new("X").blue().bold().render(area, buf);
```

Implements `Styled` so the whole `Stylize` shorthand surface works. 7 unit tests + 2 doctests, full crate test suite passes.

Metbcy:
Heads up this PR got auto-closed after I force-pushed an orphaned commit during the fixup mentioned above, and GitHub now refuses to reopen it. Same branch, same diff, continued as #2520. Sorry for the noise! My bad!

joshka:
no probs - it happens

PR #2520, opened 2026-05-03
feat(widgets): add Fill widget

Metbcy:
Adds a new `Fill` widget that paints every cell within its area with a single repeated symbol and style. Integrates with `Stylize` so the whole chain works as expected:

```rust
use ratatui::widgets::{Fill, Widget};
use ratatui::style::Stylize;

Fill::new("X").blue().bold().render(area, buf);
```

Implements `Widget` (for both `Fill` and `&Fill`) and `Styled`, accepts anything that converts into a `Cow<'a, str>`, and degrades gracefully on empty / multi-grapheme symbols.

Includes 7 unit tests + doctests; `cargo test -p ratatui-widgets` passes locally.

---

Replaces #2519 (auto-closed by GitHub after a bad force-push made the head an orphan commit). Same author, same branch, same diff. Original review feedback by @joshka has been addressed: this is the widget-shaped API he preferred over the original `Buffer::fill` method approach. My bad for the extra noise 😭 .

cc @joshka

### DIFF
--- a/ratatui-widgets/src/fill.rs
+++ b/ratatui-widgets/src/fill.rs
@@ -1,0 +1,204 @@
+//! The [`Fill`] widget paints every cell in its area with a single symbol and style.
+use alloc::borrow::Cow;
+
+use ratatui_core::buffer::Buffer;
+use ratatui_core::layout::Rect;
+use ratatui_core::style::{Style, Styled};
+use ratatui_core::widgets::Widget;
+
+/// A widget that fills its render area with a single repeated symbol and style.
+///
+/// `Fill` is a small building block for painting solid blocks of one symbol — backgrounds,
+/// separators, scrollbar tracks, custom borders, etc. — without writing the nested loop
+/// yourself. It composes naturally with the [`Stylize`] trait so the typical call site is
+/// a one-liner.
+///
+/// # Examples
+///
+/// ```
+/// use ratatui::layout::Rect;
+/// use ratatui::style::Stylize;
+/// use ratatui::widgets::{Fill, Widget};
+///
+/// # let mut buf = ratatui::buffer::Buffer::empty(Rect::new(0, 0, 10, 5));
+/// let fill = Fill::new("X").blue().bold();
+/// fill.render(Rect::new(0, 0, 10, 5), &mut buf);
+/// ```
+///
+/// `Fill` accepts anything that converts into a [`Cow<str>`], so both string literals and
+/// owned [`String`](alloc::string::String)s work:
+///
+/// ```
+/// use ratatui::widgets::Fill;
+///
+/// let _ = Fill::new("•");
+/// let _ = Fill::new(String::from("•"));
+/// ```
+///
+/// Cells outside the buffer are silently clipped, mirroring the behavior of other widgets
+/// such as [`Clear`](crate::clear::Clear).
+///
+/// [`Stylize`]: ratatui_core::style::Stylize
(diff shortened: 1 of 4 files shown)


ratatui-widgets/src/fill.rs
```diff
@@ -0,0 +1,203 @@
...
+/// ```
+/// use ratatui::layout::Rect;
+/// use ratatui::style::Stylize;
+/// use ratatui::widgets::{Fill, Widget};
+///
+/// # let mut buf = ratatui::buffer::Buffer::empty(Rect::new(0, 0, 10, 5));
+/// Fill::new("X")
+///     .blue()
+///     .bold()
+///     .render(Rect::new(0, 0, 10, 5), &mut buf);
```
joshka: (nit) Consider a small intermediate binding here:

```suggestion
/// let fill = Fill::new("X").blue.bold();
/// fill.render(Rect::new(0, 0, 10, 5), &mut buf);
```

and try to rephrase a bit to avoid &mut buf in the example (to help newer rustaceans)

ratatui-widgets/src/fill.rs
```diff
@@ -0,0 +1,203 @@
...
+    fn render(self, area: Rect, buf: &mut Buffer) {
+        let area = area.intersection(*buf.area());
+        if area.is_empty() {
+            return;
+        }
+        for y in area.top()..area.bottom() {
+            for x in area.left()..area.right() {
+                buf[(x, y)].set_symbol(&self.symbol).set_style(self.style);
+            }
+        }
```
joshka: nit:

```suggestion
        for position in area.positions()
            buf[position].set_symbol(&self.symbol).set_style(self.style);
        }
```

ratatui-widgets/src/fill.rs
```diff
@@ -0,0 +1,203 @@
...
+    use super::*;
+
+    #[test]
+    fn fills_area_with_symbol_and_style() {
+        let mut buffer = Buffer::empty(Rect::new(0, 0, 5, 3));
+        Fill::new(".")
+            .style(Style::new().fg(Color::Red))
+            .render(Rect::new(1, 1, 3, 1), &mut buffer);
+
+        let mut expected = Buffer::with_lines(["     ", " ... ", "     "]);
```
joshka: here and below, these tests work best when done as visual as possible (i.e. the 2d output is obvious). There's a variety of ways of doing this - `#[rustfmt::skip]`, choosing a longer line size that doesn't trigger the formatter to wrap, or using insta inline snapshots (which trim newlines and leading indentation) or indoc. There are probably other patterns in other tests (macros etc.)

```suggestion
        let mut expected = Buffer::with_lines([
            "     ",
            " ... ",
            "     "]);
```

Basic note here is make the tests look like the output so it's easy to validate at a glance
(not a huge deal as the code is obviously correct)

joshka (approved):
General approval - nits

Metbcy:
Force-pushed to fix CI. Two of the suggestion blocks had small typos that I committed verbatim from the GitHub UI (Lazy I know, sorry):

- The doc example was `Fill::new("X").blue.bold()` (missing `()` on `blue`)
- The render loop was `for position in area.positions()` (missing opening `{`)

Both are fixed. Also took the third nit on the visual buffer test by adding `#[rustfmt::skip]` and keeping the 2D layout. Squashed everything into the original feat commit so the history stays clean. Tests + doctests pass locally.

joshka (approved):

orhun (approved):
LGTM, thanks!

PR #2525, opened 2026-05-09
fix(layout): account for overlap in Ratio and Percentage constraints

### LINKED ISSUE #2311: Spacing::Overlap and Constraint::Ratio don't work together
## Description

When creating a Layout using Spacing::Overlap and Constraint::Ratio the resulting Rects are too small.
Specifically, their length over the relevant axis sum to exactly the length of the original Rect. This implies that the size calculation for the Constraint::Ratio does not take the overlap into account.

## To Reproduce

Minimal example code:

```rust
let area = Rect::new(0, 0, 11, 3);

let [area_1, area_2] = Layout::horizontal([Ratio(1, 2); 2])
    .spacing(Spacing::Overlap(1))
    .areas(area);

assert_eq!(area_1.width, 6);
assert_eq!(area_2.width, 5); // This is wrong
```

This is wrong, because a width of 11 plus 1 overlap between the two areas sums to available 12 spaces. However, only 11 are used by the resulting areas

## Expected behavior

The resulting areas should use all the available space.

```rust
let area = Rect::new(0, 0, 11, 3);

let [area_1, area_2] = Layout::horizontal([Ratio(1, 2); 2])
    .spacing(Spacing::Overlap(1))
    .areas(area);

assert_eq!(area_1.width, 6);
assert_eq!(area_2.width, 6); // This is different to above
```

## Environment

- OS: openSUSE Tumbleweed x86_64
- Crate version: 0.30.0
- Backend: crossterm
- Features: default-features

kimjune01:
`Spacing::Overlap` with `Constraint::Ratio` or `Constraint::Percentage` currently computes target sizes from the physical area, ignoring the fact that segments share cells when overlap is negative. With 2 segments of `Ratio(1, 2)` in area=11 and overlap=1, each segment gets width 5 instead of the expected 6.

This adjusts the effective area for proportional constraints by adding back the shared cells: `effective = physical + (n-1) * overlap`. The adjustment is gated to fixed-spacing flex modes (`Legacy`, `Start`, `Center`, `End`) since distributed modes (`SpaceBetween`, `SpaceAround`, `SpaceEvenly`) treat spacing as a minimum/growth input rather than a fixed gap.

Also fixes a potential `i16::MIN` overflow when negating the spacing value by casting to `f64` before negation.

Includes regression tests covering all fixed-spacing flex modes and a guard test confirming distributed modes are unaffected.

Fixes #2311

### DIFF
--- a/ratatui-core/src/layout/layout.rs
+++ b/ratatui-core/src/layout/layout.rs
@@ -878,7 +878,14 @@ impl Layout {
         configure_variable_in_area_constraints(&mut solver, &variables, area_size)?;
         configure_variable_constraints(&mut solver, &variables)?;
         configure_flex_constraints(&mut solver, area_size, &spacers, flex, spacing)?;
-        configure_constraints(&mut solver, area_size, &segments, constraints, flex, spacing)?;
+        configure_constraints(
+            &mut solver,
+            area_size,
+            &segments,
+            constraints,
+            flex,
+            spacing,
+        )?;
         configure_fill_constraints(&mut solver, &segments, constraints, flex)?;
 
         if !flex.is_legacy() {
--- a/ratatui-core/src/layout/layout.rs
+++ b/ratatui-core/src/layout/layout.rs
@@ -951,15 +951,20 @@ fn configure_constraints(
     flex: Flex,
     spacing: i16,
 ) -> Result<(), AddConstraintError> {
-    // When spacing is negative (overlap), the effective area for proportional constraints
-    // (Ratio and Percentage) is larger than the physical area because segments share cells.
+    // When spacing is negative (overlap) and the flex mode uses fixed inter-segment spacers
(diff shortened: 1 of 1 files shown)


kimjune01:
Cannot validate on current setup, so drafting. Please close or take it over.

truffle-dev:
Pulled this and rebased onto current main (`be718d0`). All 1440 `ratatui-core` unit tests pass, including the 8 new `ratio_and_percentage_with_overlap` cases and the `distributed_flex_overlap_uses_physical_area` guard.

Effective-area precision scaling at `ratatui-core/src/layout/layout.rs:972` aligns with the existing `area.size()` scale, and the `f64`-negation in `fc03928` handles `i16::MIN` cleanly.

If a takeover would help, happy to open a successor PR with your commits preserved + rebase on main. Otherwise leaving the signal that the change validates cleanly.

PR #2546, opened 2026-05-19
fix(widgets): prevent chart scaling overflow

### LINKED ISSUE #2545: BarChart and Sparkline can overflow when scaling u64 values
## Problem

`BarChart` and `Sparkline` scale `u64` input values by multiplying before dividing. For large values, such as `u64::MAX`, the intermediate multiplication can overflow before the result is divided back down to the drawable area.

In debug builds this can panic. In release builds the overflow can wrap and render an incorrect bar height.

## Reproduction shape

A value at `u64::MAX` with a matching max value should render as full height, but the current multiplication overflows before division:

```rust
let chart = BarChart::new([Bar::new(u64::MAX)]).max(u64::MAX);
let sparkline = Sparkline::default().data([u64::MAX]).max(u64::MAX);
```

## Expected

Scaling should use a wider intermediate and stay bounded by the render area.

fallintoplace:
## Summary

Fixes #2545.

This changes BarChart and Sparkline scaling to use a `u128` intermediate before division, then caps the scaled ticks at the drawable area. That prevents debug-build panics and release-build wrapping when public `u64` chart values are large.

## Validation

- `cargo test -p ratatui-widgets barchart::tests`
- `cargo test -p ratatui-widgets sparkline::tests`

### DIFF
--- a/ratatui-widgets/src/barchart.rs
+++ b/ratatui-widgets/src/barchart.rs
@@ -453,11 +453,17 @@ impl BarChart<'_> {
                         .bars
                         .iter()
                         .take(n as usize)
-                        .map(|bar| bar.value * u64::from(bar_max_length) * 8 / max)
+                        .map(|bar| Self::scale_ticks(bar.value, max, bar_max_length))
                         .collect()
                 })
             })
             .collect()
+    }
+
+    fn scale_ticks(value: u64, max: u64, max_length: u16) -> u64 {
+        let max_ticks = u128::from(max_length) * 8;
+        let ticks = u128::from(value) * max_ticks / u128::from(max);
+        ticks.min(max_ticks) as u64
     }
 
     /// Get label information.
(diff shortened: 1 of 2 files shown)


joshka:
qq on the fix. Could you reasonably simplify the calculation by using floats here instead or does the loss of precision cause this to look ugly if you do that?

From a motivation perspective, are you fixing this as a hypothetical issue or as a user that's impacted by this as a real problem btw?

fallintoplace:
Good question. I pushed a small follow-up test for the float precision edge case.

I’d avoid `f64` here because it can change visible output for large `u64`s: `u64::MAX - 1` over `u64::MAX` should render 7/8 ticks in a one-cell area, but with floats both values round together and render as full.

Motivation-wise, this came from reading the scaling code rather than a production issue I hit, but it’s easy to reproduce in debug and can misrender in release for valid public input.

joshka (approved):
LGTM

PR #2548, opened 2026-05-20
fix(widgets): avoid extra gauge label cell

### LINKED ISSUE #2547: Gauge renders an extra label cell in the filled area
## Problem

`Gauge::render_gauge` treats one cell after the centered label as part of the label background.

`buf.set_span(label_col, label_row, label, clamped_label_width)` writes the half-open range `[label_col, label_col + clamped_label_width)`, but the filled-area condition currently uses `x > label_col + clamped_label_width`. That leaves `x == label_col + clamped_label_width` classified as label background even though the label does not occupy that cell.

For a 10-wide gauge at 100% with the default `100%` label, the cell immediately after the label should be a filled block.

## Expected

The filled-area check should use the same half-open label range as `set_span`, so the first cell after the label is rendered as normal filled gauge content.

fallintoplace:
## Summary

Fixes #2547.

This updates `Gauge` rendering so the filled-area label check uses the same half-open range as `buf.set_span`. The first cell after the centered label is now rendered as normal filled gauge content instead of label-background space.

## Validation

- `cargo test -p ratatui-widgets gauge::tests`
- `cargo test -p ratatui widgets_gauge`

### DIFF
--- a/ratatui-widgets/src/gauge.rs
+++ b/ratatui-widgets/src/gauge.rs
@@ -180,6 +180,7 @@ impl Gauge<'_> {
         let label = self.label.as_ref().unwrap_or(&default_label);
         let clamped_label_width = gauge_area.width.min(label.width() as u16);
         let label_col = gauge_area.left() + (gauge_area.width - clamped_label_width) / 2;
+        let label_end = label_col.saturating_add(clamped_label_width);
         let label_row = gauge_area.top() + gauge_area.height / 2;
 
         // the gauge will be filled proportionally to the ratio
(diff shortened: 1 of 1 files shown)


joshka:
Thanks for taking the time to investigate and submit this fix. Sorry for the timing—we received three PRs for the same Gauge bug in [#2547](https://github.com/ratatui/ratatui/issues/2547). We’ve now merged [#2579](https://github.com/ratatui/ratatui/pull/2579), which contains the same boundary correction along with a regression test, so this PR is superseded. I’m closing it as a duplicate. Thanks again for the contribution and the additional defensive considerations.

PR #2551, opened 2026-05-20
feat(text): add alignment helpers for plain strings

### LINKED ISSUE #953: `&str` and `String` should implement methods to make it similar to an unstyled `Span`
## Problem

Currently it is possible to write code like this:

```rust
"hello world".bold().to_centered_line()
```

but not possible to write code like this:

```rust
"hello world".to_centered_line()
```

fallintoplace:
Summary
Fixes #953.

This adds alignment convenience methods to the existing `ToLine` trait so plain displayable values can be converted directly into aligned `Line`s:

- `"hello".to_left_aligned_line()`
- `"hello".to_centered_line()`
- `"hello".to_right_aligned_line()`
- `String::from("hello").to_centered_line()`

The helpers return owned lines, which keeps the owned `String` temporary case ergonomic. The main crate prelude now re-exports `ToLine`, so `use ratatui::prelude::*` brings these helpers into scope.

Validation
- `cargo test -p ratatui-core text::line::tests`
- `cargo test -p ratatui-core`
- `cargo check -p ratatui-core --all-features`
- `cargo check -p ratatui --all-features`
- `cargo clippy -p ratatui-core --all-targets --all-features -- -D warnings`
- `cargo clippy -p ratatui --all-targets --all-features -- -D warnings`
- `cargo test -p ratatui-core --doc`
- `cargo test -p ratatui --doc`
- `cargo check -p ratatui-core --no-default-features`

### DIFF
--- a/ratatui-core/src/text/line.rs
+++ b/ratatui-core/src/text/line.rs
@@ -39,6 +39,8 @@ use crate::widgets::Widget;
 /// - [`Line::from`] creates a `Line` from single [`Span`].
 /// - [`String::from`] converts a line into a [`String`].
 /// - [`Line::from_iter`] creates a line from an iterator of items that are convertible to [`Span`].
+/// - [`ToLine::to_line`] converts any `Display` value to a `Line`.
+/// - [`ToLine::to_centered_line`] converts any `Display` value to a centered `Line`.
 ///
 /// # Setter Methods
 ///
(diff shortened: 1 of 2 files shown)


joshka:
Thanks for picking this up.

It’s worth noting the context of the original issue here too. This was filed back in early 2024, before some of the current text APIs existed, so the request made more sense at the time because `Span` had direct `into_centered_line()` helpers while plain strings did not. Those methods have since also been renamed to `into_*_line()` on `Span`.

I’m not certain that adding alignment-specific helpers to plain strings adds much value generally now that `.to_line().centered()` exists and reads reasonably well already.

My main concern is that this creates a cross-product API growth problem across conversion, ownership, style, alignment, wrapping, and other text facets. The composable approach keeps the surface area smaller while still being ergonomic.

It’s probably worth doing a bit of spelunking through the repo history (PRs, issues, commit messages) to understand how we ended up with the current API shape. If you’re using a coding agent, asking it to gather and summarize that history could help bring in more context here.

Not a hard no on this, more that I’d forgotten this issue existed and, with current perspective, it feels somewhat unnecessary. I suspect I’d probably lean toward closing the original issue today, though it may be worth continuing the discussion there a bit more before making a final call.

PR #2553, opened 2026-05-20
fix(barchart): handle empty horizontal charts

### LINKED ISSUE #2552: BarChart::horizontal can panic with empty bars
## Bug

`BarChart::horizontal` can panic when it is constructed with an empty bar list.

`BarChart::default().data(...)` already ignores empty groups, but `BarChart::new`, `BarChart::horizontal`, and `BarChart::grouped` store groups directly. That makes constructor behavior inconsistent with the builder path.

In the horizontal renderer, an empty group skips the inner bar loop and then computes:

```rust
let label_y = bar_y - self.bar_gap;
```

With the default `bar_gap = 1` and `bar_y = 0`, this underflows in debug builds.

## Reproduction

```rust
let chart = BarChart::horizontal(Vec::<Bar>::new());
let mut buffer = Buffer::empty(Rect::new(0, 0, 10, 3));

chart.render(buffer.area, &mut buffer);
```

## Expected behavior

An empty horizontal chart should render nothing, matching how empty data passed through `.data(...)` is ignored.

fallintoplace:
Summary
Fixes #2552.

This makes the `BarChart` constructors ignore empty groups, matching the existing `.data(...)` builder behavior. Without this, `BarChart::horizontal(Vec::<Bar>::new())` stores one empty group, proceeds into horizontal rendering, skips the bar loop, and then underflows when computing the group label row from `bar_y - self.bar_gap`.

The fix normalizes constructor input through a shared `non_empty_groups` helper for `new`, `horizontal`, and `grouped`. Empty horizontal charts now render nothing instead of panicking, and constructor behavior is consistent with `.data(...)`.

Validation
- `cargo test -p ratatui-widgets barchart::tests::horizontal_empty_barchart_does_not_panic`
- `cargo test -p ratatui-widgets barchart::tests::constructors_ignore_empty_groups`
- `cargo test -p ratatui-widgets barchart::tests`

### DIFF
--- a/ratatui-widgets/src/barchart.rs
+++ b/ratatui-widgets/src/barchart.rs
@@ -142,7 +142,7 @@ impl<'a> BarChart<'a> {
     /// ```
     pub fn new<T: Into<Vec<Bar<'a>>>>(bars: T) -> Self {
         Self {
-            data: vec![BarGroup::new(bars.into())],
+            data: Self::non_empty_groups(vec![BarGroup::new(bars.into())]),
             direction: Direction::Vertical,
             ..Default::default()
         }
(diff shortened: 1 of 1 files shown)


orhun (approved):
LGTM, thanks!

Issue #1784, opened 2025-04-13
Add `ratatui-termina` backend


orhun:
Helix maintainers just released a new terminal manipulation library, combining some parts of `crossterm` and `termwiz`: https://github.com/helix-editor/termina

Some info on the motivation here: https://github.com/helix-editor/helix/pull/13307

We might want to add a `ratatui-termina` backend when the project is mature enough.

jarjk:
I think it's an awesome idea!

it's now used in `helix master`

you'd probably wait for a stable release of `hx` with `termina` before giving it a try, right?

orhun:
Yup!

musjj:
I would love to see this too!

Crossterm has very poor support for tmux and has caused me a lot of headache with debugging missing terminal features:

- https://github.com/sxyazi/yazi/issues/3910
- https://github.com/crossterm-rs/crossterm/pull/810
- https://github.com/crossterm-rs/crossterm/issues/401

joshka:
I've had a PR sitting locally to update the termina docs for a while (started in Jan and forgot about it).
I just slapped it their way. https://github.com/helix-editor/termina/pull/23
Let's see what they say.

joshka:
https://github.com/ratatui/ratatui/pull/2561

orhun:
Now we have `ratatui-termina`!

PR #2561, opened 2026-05-27
feat: add Termina backend

joshka:
## Summary

- add the `ratatui-termina` backend crate using the published `termina` crate
- expose the backend through the `termina` feature and Ratatui prelude/backend re-exports
- add a small Termina event-loop example and wire the backend into CI, xtask, README generation, and docs

Refs #1784

## Validation

- `cargo +nightly fmt`
- `cargo check -p ratatui-termina --all-features --all-targets`
- `cargo check -p ratatui --no-default-features --features termina`
- `cargo check -p xtask`
- `cargo check -p release-header`
- `cargo xtask check-backend termina`
- `cargo xtask test-backend termina`
- `cargo xtask rdme --check`
- `markdownlint-cli2 ARCHITECTURE.md ratatui-termina/README.md .github/ISSUE_TEMPLATE/bug_report.md`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,7 +64,7 @@ rstest = "0.26"
 serde = { version = "1", default-features = false, features = ["derive"] }
 serde_json = "1"
 strum = { version = "0.28", default-features = false, features = ["derive"] }
-termina = "0.3.2"
+termina = "0.3"
 termion = "4"
 termwiz = "0.23"
 thiserror = { version = "2", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,9 +7,10 @@ default-members = [
   "ratatui",
   "ratatui-core",
   "ratatui-crossterm",
+  "ratatui-macros",
   # this is not included as it doesn't compile on windows
   # "ratatui-termion",
-  "ratatui-macros",
+  "ratatui-termina",
   "ratatui-termwiz",
   "ratatui-widgets",
 ]
@@ -55,6 +56,7 @@ ratatui = { path = "ratatui", version = "0.30.1" }
 ratatui-core = { path = "ratatui-core", version = "0.1.1" }
 ratatui-crossterm = { path = "ratatui-crossterm", version = "0.1.1" }
 ratatui-macros = { path = "ratatui-macros", version = "0.7.1" }
+ratatui-termina = { path = "ratatui-termina", version = "0.1.0" }
 ratatui-termion = { path = "ratatui-termion", version = "0.1.1" }
 ratatui-termwiz = { path = "ratatui-termwiz", version = "0.1.1" }
 ratatui-widgets = { path = "ratatui-widgets", version = "0.3.1", default-features = false }
@@ -62,6 +64,7 @@ rstest = "0.26"
 serde = { version = "1", default-features = false, features = ["derive"] }
 serde_json = "1"
 strum = { version = "0.28", default-features = false, features = ["derive"] }
+termina = "0.3.2"
 termion = "4"
 termwiz = "0.23"
 thiserror = { version = "2", default-features = false }
--- a/examples/apps/demo/src/termina.rs
+++ b/examples/apps/demo/src/termina.rs
@@ -1,0 +1,113 @@
+#![allow(dead_code)]
+use std::error::Error;
+use std::io::Write as _;
+use std::time::{Duration, Instant};
+
+use ratatui::Terminal;
+use ratatui::backend::TerminaBackend;
+use ratatui::termina::escape::csi::{self, Csi};
+use ratatui::termina::event::{KeyCode, KeyEventKind};
+use ratatui::termina::{Event, EventReader, PlatformTerminal, Terminal as _};
+
+use crate::app::App;
+use crate::ui;
+
+type AppTerminal = Terminal<TerminaBackend<PlatformTerminal>>;
+
+macro_rules! decset {
+    ($mode:ident) => {{
+        let mode = csi::DecPrivateMode::Code(csi::DecPrivateModeCode::$mode);
+        Csi::Mode(csi::Mode::SetDecPrivateMode(mode))
+    }};
+}
+
+macro_rules! decreset {
+    ($mode:ident) => {{
+        let mode = csi::DecPrivateMode::Code(csi::DecPrivateModeCode::$mode);
+        Csi::Mode(csi::Mode::ResetDecPrivateMode(mode))
+    }};
+}
+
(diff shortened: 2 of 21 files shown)


ratatui/Cargo.toml
martin-g: And Termina

joshka: thanks

joshka: https://github.com/ratatui/ratatui/compare/5119b2bb4e0c5454a07fd88e867927ea1c2373ad..2c6d137a8c514df5ccfee8853cdd1ef7ca372b34

jarjk:
great to see this PR!
could you add `termina` backend to `examples/apps/demo` as well, would be amazing

joshka:
I wouldn't mind if @the-mikedavis could take a look at the calling patterns here. (No obligation obv.)
Much of the implementation is my / my coding agent's guesses a bit about what matters, and I haven't read a lot of helix code to see if there's anything in there that smells off.

the-mikedavis:
Almost all of the code in Helix for the termina integration is in this terminal backend module: https://github.com/helix-editor/helix/blob/master/helix-tui/src/backend/termina.rs

The tui crate originally came from tui-rs so there might be some design overlap still, though it was forked a long time ago. I will give this branch a look 👍

joshka:
> Almost all of the code in Helix for the termina integration is in this terminal backend module: https://github.com/helix-editor/helix/blob/master/helix-tui/src/backend/termina.rs

Thanks. I asked Codex to take a review with the PR in context against that file and the deeper app context in comparison to the example code. Seems this side is reasonable.

---

I looked at both.

The PR comment is informational: it points to Helix’s Termina backend and says they’ll review the branch.

The useful takeaways from Helix:

- It uses local `decset!` / `decreset!` macros, not exported Termina API. Our PR now mirrors that pattern.
- Helix’s backend owns the whole terminal session: raw mode, alt screen, bracketed paste, focus/mouse, Kitty keyboard, panic hook, synchronized output, background color. Ratatui backends generally should not take over that lifecycle; the app/example should.
- Helix separates underline color/style from grouped `SgrAttributes` because of terminal quirks. In Ratatui we only have underline color, and Termina’s `SgrAttributes` is explicitly built for grouping SGR updates, so our current grouping still seems reasonable.
- Helix has synchronized output support, but it depends on capability detection and session ownership. I would not add that to this PR unless we decide Ratatui backends should manage extension negotiation.

No further code change looks required from that comment beyond the macro cleanup already pushed.

---

I looked at Helix beyond the backend. The important bit is that Helix’s backend example is not really comparable to our short example: Helix has an app-owned terminal lifecycle.

Sources checked:

- Helix backend: https://github.com/helix-editor/helix/blob/master/helix-tui/src/backend/termina.rs
- Helix application setup: https://github.com/helix-editor/helix/blob/master/helix-term/src/application.rs
- Helix main: https://github.com/helix-editor/helix/blob/master/helix-term/src/main.rs

What Helix does:

- `main` builds `Application`, creates an event stream, then calls `app.run(&mut events)`.
- `Application::new` creates `TerminaBackend::new(config)`, reads theme mode, wraps it in Helix’s `Terminal`, but does not claim the terminal for the app loop yet.
- `TerminaBackend::new` opens `PlatformTerminal`, briefly enters raw mode for capability detection, returns to cooked mode, and installs the Termina panic hook.
- `Application::run` calls `terminal.claim()` before the event loop and `restore_term()` after it.
- Signal handling restores on suspend/termination and reclaims on resume.
- Event reading comes from `self.terminal.backend().terminal().event_reader()`, wrapped in `termina::EventStream`.

Compared to our example:

- Our example does the small version directly: create `PlatformTerminal`, enter raw mode, enter alternate screen, take `event_reader()`, move the terminal into `TerminaBackend`, then run draw/read loop.
- That matches the useful minimum for a Ratatui backend example.
- Helix’s panic hook, suspend/resume handling, capability detection, synchronized output, background color restoration, mouse/focus/bracketed-paste setup, and async event stream are app-grade lifecycle concerns. Pulling those into the example would make it heavy again.
- The one Helix pattern our example already follows is the key ownership setup: take the event reader before moving `PlatformTerminal` into the backend.

My read: no code change is needed from this comparison. Helix reinforces that the Ratatui example should stay explicit and small, while the richer claim/restore/panic-hook lifecycle belongs in real applications or a separate higher-level session abstraction, not in the backend example itself.

orhun:
I added the termina backend version of the demo app in 3cced2c5bfbfc8aaecd8e42e208749a4f9c6d60c - I hope you don't mind :)

examples/apps/demo/src/termina.rs
```diff
@@ -0,0 +1,113 @@
...
+        if key.kind != KeyEventKind::Press {
+            return;
+        }
+        match key.code {
+            KeyCode::Up | KeyCode::Char('k') => app.on_up(),
+            KeyCode::Down | KeyCode::Char('j') => app.on_down(),
+            KeyCode::Left | KeyCode::Char('h') => app.on_left(),
+            KeyCode::Right | KeyCode::Char('l') => app.on_right(),
+            KeyCode::Char(c) => app.on_key(c),
+            _ => {}
```
joshka: nit:
```suggestion
            KeyCode::Char('k') | KeyCode::Up => app.on_up(),
            KeyCode::Char('j') | KeyCode::Down => app.on_down(),
            KeyCode::Char('h') | KeyCode::Left => app.on_left(),
            KeyCode::Char('l') | KeyCode::Right => app.on_right(),
            KeyCode::Char(c) => app.on_key(c),
            _ => {}
```

orhun (approved):
LGTM

Anything blocking on this one?
I'd say let's include this in 0.30.2 and let the users try it out.

Cargo.toml
```diff
@@ -55,13 +56,15 @@ ratatui = { path = "ratatui", version = "0.30.1" }
...
 ratatui-macros = { path = "ratatui-macros", version = "0.7.1" }
+ratatui-termina = { path = "ratatui-termina", version = "0.1.0" }
 ratatui-termion = { path = "ratatui-termion", version = "0.1.1" }
 ratatui-termwiz = { path = "ratatui-termwiz", version = "0.1.1" }
 ratatui-widgets = { path = "ratatui-widgets", version = "0.3.1", default-features = false }
 rstest = "0.26"
 serde = { version = "1", default-features = false, features = ["derive"] }
 serde_json = "1"
 strum = { version = "0.28", default-features = false, features = ["derive"] }
+termina = "0.3.3"
```
joshka: Maybe just 0.3 instead if that works. Will check later.

PR #2563, opened 2026-05-28
fix(buffer): preserve cells by position when resizing

### LINKED ISSUE #2556: `Buffer::resize` does not respect `area.x/y`
## Description

If `resize` is passed a `Rect` with differing `x` & `y` values from the current `area` these are silently discarded and only the new `width` & `height` are used to truncate/extend. Looking briefly at the code, this probably also hides bugs where it wraps lines contents onto new lines.

## To Reproduce

1. Create a Buffer of `area = Rect::new(0, 0, 80, 3)`
2. Call resize, giving a new `expected_area = Rect::new(1, 1, expected_text.len().try_into().unwrap(), 1)`
3. The result will contain `expected_text.len()` cells from (0,0)

## Expected behavior

The resulting Buffer should contain `expected_text.len()` cells from (1,1)

cyphercodes:
## Summary

- Preserve `Buffer::resize` contents by global cell position instead of linear vector index
- Initialize newly exposed cells as empty when the resized area grows or moves
- Add regression coverage for positioned resize and width changes

Fixes #2556

## Test Plan

- `cargo fmt --all --check` (passes; stable rustfmt prints warnings for the repo's nightly-only rustfmt options)
- `cargo test -p ratatui-core resize_ -- --nocapture`
- `cargo test -p ratatui-core buffer::buffer::tests -- --nocapture`
- `cargo test -p ratatui-core`
- `cargo clippy -p ratatui-core --all-targets -- -D warnings`
- `git diff --check`

## AI assistance

This PR was prepared with AI assistance and the resulting changes were reviewed before submission.

### DIFF
--- a/ratatui-core/src/buffer/buffer.rs
+++ b/ratatui-core/src/buffer/buffer.rs
@@ -412,15 +412,28 @@ impl Buffer {
         }
     }
 
-    /// Resize the buffer so that the mapped area matches the given area and that the buffer
-    /// length is equal to area.width * area.height
+    /// Resize the buffer so that the mapped area matches the given area and the buffer
+    /// length is equal to area.width * area.height.
+    ///
+    /// Cells that overlap between the old and new areas are preserved at the same global
+    /// coordinates. New cells are initialized as empty.
     pub fn resize(&mut self, area: Rect) {
-        let length = area.area() as usize;
-        if self.content.len() > length {
-            self.content.truncate(length);
-        } else {
-            self.content.resize(length, Cell::EMPTY);
+        if self.area == area {
+            return;
         }
+
+        let old_area = self.area;
+        let old_content =
+            core::mem::replace(&mut self.content, vec![Cell::EMPTY; area.area() as usize]);
+        let preserved_area = old_area.intersection(area);
+
+        for position in preserved_area.positions() {
+            let old_index =
+                ((position.y - old_area.y) * old_area.width + position.x - old_area.x) as usize;
+            let new_index = ((position.y - area.y) * area.width + position.x - area.x) as usize;
+            self.content[new_index] = old_content[old_index].clone();
+        }
+
         self.area = area;
     }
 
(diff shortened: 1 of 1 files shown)


ratatui-core/src/buffer/buffer.rs
```diff
@@ -412,15 +412,28 @@ impl Buffer {
...
-            self.content.truncate(length);
-        } else {
-            self.content.resize(length, Cell::EMPTY);
+        if self.area == area {
+            return;
+        }
+
+        let old_area = self.area;
+        let old_content =
+            core::mem::replace(&mut self.content, vec![Cell::EMPTY; area.area() as usize]);
```
MusicalNinjaDad, line 427: why? seems overly complicated, particularly if you're going to clone() each value later anyway

ratatui-core/src/buffer/buffer.rs
```diff
@@ -412,15 +412,28 @@ impl Buffer {
...
+        let preserved_area = old_area.intersection(area);
+
+        for position in preserved_area.positions() {
```
MusicalNinjaDad, line 430: Surely this would be better as a (lazy) pipeline of into_iter, filter, map, etc., collect

ratatui-core/src/buffer/buffer.rs
```diff
@@ -843,6 +856,25 @@ mod tests {
...
+        let area = Rect::new(1, 1, 2, 1);
+        buffer.resize(area);
+
+        let mut expected = Buffer::with_lines(["fg"]);
+        expected.area = area;
+        assert_eq!(buffer, expected);
+    }
+
+    #[test]
+    fn resize_does_not_wrap_cells_when_width_changes() {
```
MusicalNinjaDad, line 871: proves this is a breaking change to API - anyone relying on current functionality to implement content-wrapping is going to be upset

MusicalNinjaDad:
As the author of the underlying issue, thanks for giving this a go. But I would advise against merging this approach. It breaks the API for people relying on current functionality. It seems overly convoluted in implementation. (See review comments).

MusicalNinjaDad:
> This PR was prepared with AI assistance and the resulting changes were reviewed before submission.

Really? If so, please feel free to enter a conversation and ask for help on how to do a better review next time

PR #2566, opened 2026-05-28
feat(render): add function for applying buffer

musjj:
Add a public API for applying and flushing the terminal buffer.

A minimal usage will look something like this:

```rust
use ratatui::Terminal;
use ratatui::backend::CrosstermBackend;
use ratatui::buffer::Buffer;
use ratatui::widgets::Widget;

let backend = CrosstermBackend::new(io::stdout());

terminal.autoresize()?;

let mut custom_buffer = Buffer::default();
custom_buffer.resize(terminal.get_frame().area());
custom_buffer.reset();

"Hello World!".render(custom_buffer.area, &mut custom_buffer);

terminal.current_buffer_mut().merge(&custom_buffer);
terminal.apply_buffer()?;
```

My primary motivation for this PR is to improve the ECS ergonomics in [`bevy_ratatui`](https://github.com/ratatui/bevy_ratatui). But this should be useful for anyone who wants to commit incremental writes to the buffer without having to do everything in one monolithic [`Terminal::draw`](https://docs.rs/ratatui/latest/ratatui/struct.Terminal.html#method.draw) closure.

You can read a bit more of my motivation below. I've hidden it away by default because I felt it may be a bit too rambly and not technically relevant to this PR.

<details>
<summary><b>Details of my motivation</b></summary>

The idea is basically to have widgets that are rendered by systems that have autonomous access to the ECS world state. This is not possible with a traditional [`Widget::render`](https://docs.rs/ratatui/latest/ratatui/prelude/trait.Widget.html#tymethod.render), which requires the user to supply the state to the widget manually.

But ratatui's current monolithic nature requires you to draw the entire frame all at once, in a single function call. This limitation makes it difficult to develop a more ECS-friendly and composable architecture.

So my idea is to store a separate and owned buffer in the ECS world that the systems will draw into. At the end of the frame, the buffer is applied to the real buffer in the [`Terminal`](https://docs.rs/ratatui/latest/ratatui/struct.Terminal.html) using the function added in this PR.

A separate virtual buffer is needed because it's not really possible to store a reference to the real buffer due to Rust lifetime issues.

A rough sketch of what a widget system could look like:

```rust
fn draw_my_widget(frame: ResMut<BevyFrame>, query: Query<&MyWidget>) {
   frame.render_widget(...);
}
```

`BevyFrame` here does not reference the actual buffer in the `Terminal`, so it can be safely stored in the ECS world. This way, multiple systems can independently access and write into it. 

At the end of the frame, the buffer in `BevyFrame` gets applied to the real one in `Terminal`.

</details>

### DIFF
--- a/ratatui-core/src/terminal/render.rs
+++ b/ratatui-core/src/terminal/render.rs
@@ -201,11 +201,59 @@ impl<B: Backend> Terminal<B> {
 
         let cursor_position = frame.cursor_position;
 
-        self.apply_buffer(cursor_position)
+        self.apply_buffer_with_cursor(cursor_position)
     }
 
     /// A low-level function that applies and flushes the current buffer to the backend. This
     /// function is useful if you need to manage your own custom draw lifecycle and buffer.
+    ///
+    /// Returns [`Result::Ok`] containing a [`CompletedFrame`] if successful, otherwise
+    /// [`Result::Err`] containing the backend error (`B::Error`) that caused the failure.
+    ///
+    /// This method will:
+    ///
+    /// - call [`Terminal::swap_buffers`] to prepare for the next render pass
+    /// - call [`Backend::flush`] to flush any buffered backend output
+    /// - return a [`CompletedFrame`] with the current buffer and the area used for rendering
+    ///
+    /// The [`CompletedFrame`] returned by this method can be useful for debugging or testing
+    /// purposes, but it is often not used in regular applications.
+    ///
+    /// # Examples
+    ///
+    /// ```rust,no_run
+    /// # #![allow(unexpected_cfgs)]
+    /// # #[cfg(feature = "crossterm")]
+    /// # {
+    /// use std::io;
+    ///
+    /// use ratatui::Terminal;
+    /// use ratatui::backend::CrosstermBackend;
+    /// use ratatui::buffer::Buffer;
+    /// use ratatui::widgets::Widget;
+    ///
+    /// let backend = CrosstermBackend::new(io::stdout());
+    /// let mut terminal = Terminal::new(backend)?;
+    ///
+    /// terminal.autoresize()?;
+    ///
+    /// let mut custom_buffer = Buffer::default();
+    /// custom_buffer.resize(terminal.get_frame().area());
+    /// custom_buffer.reset();
+    ///
+    /// "Hello World!".render(custom_buffer.area, &mut custom_buffer);
+    ///
+    /// terminal.current_buffer_mut().merge(&custom_buffer);
+    /// terminal.apply_buffer()?;
+    /// # }
+    /// ```
+    pub fn apply_buffer(&mut self) -> Result<CompletedFrame<'_>, B::Error> {
+        self.apply_buffer_with_cursor(None)
+    }
+
+    /// A low-level function that applies and flushes the current buffer to the backend and
+    /// re-positions the cursor. This function is useful if you need to manage your own custom
+    /// draw lifecycle and buffer.
     ///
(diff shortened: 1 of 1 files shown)


ratatui-core/src/terminal/render.rs
```diff
@@ -198,15 +199,64 @@ impl<B: Backend> Terminal<B> {
...
+
+    /// A low-level function that applies and flushes the current buffer to the backend. This
+    /// function is useful if you need to manage your own custom draw lifecycle and buffer.
+    ///
+    /// Returns [`Result::Ok`] containing a [`CompletedFrame`] if successful, otherwise
+    /// [`Result::Err`] containing the backend error (`B::Error`) that caused the failure.
+    ///
+    /// This method will:
+    ///
+    /// - show/hide the cursor based on `cursor_position`
```
orhun: We should probably document that the cursor is hidden when the parameter is `None` (which I assume is the case)

ratatui-core/src/terminal/render.rs
```diff
@@ -198,15 +199,64 @@ impl<B: Backend> Terminal<B> {
...
+    ///
+    /// "Hello World!".render(custom_buffer.area, &mut custom_buffer);
+    ///
+    /// terminal.current_buffer_mut().merge(&custom_buffer);
+    /// terminal.apply_buffer(None)?;
+    /// # }
+    /// ```
+    pub fn apply_buffer(
+        &mut self,
+        cursor_position: Option<Position>,
```
orhun, line 290: I would prefer to have two distinct functions here for a cleaner API. For example:

- `apply_buffer()` -> calls the function below with `None` as the param
- `apply_buffer_with_cursor(cursor_position)` (or `apply_buffer_with_position`)

So that the parameter is more obvious and you could call `apply_buffer` directly without `None`.

orhun (commented):
thanks for the PR!

I would love to get the feedback from @cxreiff first on this before we proceed (he's the `bevy_ratatui` master)

cc @gold-silver-copper might be interested in looking into this too

Also it would be great if you could work on a PoC PR in `bevy_ratatui` with this. It would help concretize the implementation here.

musjj:
Here's a super minimal PoC for `bevy_ratatui`: https://github.com/musjj/bevy_ratatui_frame_poc

Here's how this PR enables multiple systems to independently draw widgets:

```rust
fn draw_left(mut frame: ResMut<BevyFrame>) {
    let [left, _] =
        Layout::horizontal([Constraint::Fill(1), Constraint::Fill(1)]).areas(frame.area);

    frame.render_widget("hello", left);
}

fn draw_right(mut frame: ResMut<BevyFrame>) {
    let [_, right] =
        Layout::horizontal([Constraint::Fill(1), Constraint::Fill(1)]).areas(frame.area);

    frame.render_widget("world", right);
}
```

You can look inside the repo if you want to see how `BevyFrame` works underneath.

ratatui-core/src/terminal/render.rs
musjj: Ok, added a clarification for this!

ratatui-core/src/terminal/render.rs
musjj: That makes sense! I've split the function into `apply_buffer` and `apply_buffer_with_cursor`.

gold-silver-copper:
I think overall this PR makes sense. It will allow an extra degree of control to all users of ratatui, not just bevy_ratatui. 

Being able to do incremental draws is pretty cool, used properly I think this could be used for improving performance in all ratatui applications. TBH I thought ratatui already supported this but I guess not so it's cool that it's getting added.

orhun (approved):
LGTM!

PR #2587, opened 2026-06-08
fix(buffer): create updates for "uncovered" cells

### LINKED ISSUE #2585: Ratatui 0.30.1 rendering background residue
## Description

I'm upgrading my app to v0.30.1 and ran into a rendering regression.

Specifically, on the second frame render, the background color set for text in the first frame is not being cleared:

https://github.com/user-attachments/assets/c7e3e871-c845-4767-b93c-2fc67f1d69cd

I did a git bisect and confirmed the bug was introduced in https://github.com/ratatui/ratatui/commit/39c32c6. CC @benjajaja maybe you have some clues.

## To Reproduce

```rust
use ratatui::{style::{Color, Stylize}, text::Span};

fn main() {

  _ = terminal.draw(|frame| {
    let greeting = Span::raw("你好，世界！").bg(Color::Blue);
    frame.render_widget(greeting, frame.area());
  });

  std::thread::sleep(std::time::Duration::from_secs(1));
  _ = terminal.draw(|frame| {
    let greeting = Span::raw("Hello");
    frame.render_widget(greeting, frame.area());
  });

  std::thread::sleep(std::time::Duration::from_secs(3));
  ratatui::restore();
}
```

## Expected behavior

Background colors should be cleared in the next frame.

## Screenshots

See description.

## Environment

- OS: macOS
- Terminal Emulator: kitty 0.47.2
- Font: SF Mono
- Crate version: Latest main branch (https://github.com/ratatui/ratatui/commit/fce3c80d53d5cf367e62cadf1d5c819947d23e4c)
- Backend: I'm using [yazi-term](https://github.com/sxyazi/yazi/tree/main/yazi-term), but I tried `ratatui-crossterm` can also reproduce the bug so it shouldn't be backend-specific

## Additional context

@december1981 seems to have reported a similar issue at https://github.com/ratatui/ratatui/pull/1605#issuecomment-4290254795.

benjajaja:
When a wide cell from the previous buffer is replaced by a short/normal cell, the trailing cell does not get an update if its content does not change. But if the wide cell has a background (or other) style, the terminal *did* render the trailing cell with that style.

Force trailing cells to update if background, underline, or modifiers are different than the wide cell. We can ignore foreground.

Fixes #2585 (see that for the detailed visual reports)

### DIFF
--- a/ratatui-core/src/buffer/diff.rs
+++ b/ratatui-core/src/buffer/diff.rs
@@ -1,5 +1,6 @@
 use crate::buffer::{Buffer, Cell, CellDiffOption, CellWidth};
 use crate::layout::Rect;
+use crate::style::{Color, Modifier};
 
 /// A zero-allocation iterator over the differences between two buffers of the same width.
 ///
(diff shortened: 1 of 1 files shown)


benjajaja:
For https://github.com/ratatui/ratatui/issues/2585#issuecomment-4659791941, invalid emoji modifier sequences such as `🥲🏻`, I made https://github.com/benjajaja/ratatui/pull/1 on top of this, but that should probably be applied/reviewed later, separately.

joshka (approved):

PR #2594, opened 2026-06-10
fix(scrollbar): keep a large thumb within the track at the end

### LINKED ISSUE #2582: ratatui v0.30.1 sometimes renders scrollbars with a "thumb" symbol in place of "end" symbol
## Description

In ratatui v0.30.1, rendering a scrollbar where the position is at or almost at the end of the content can sometimes (depending on the exact values used) result in the "end" symbol being replaced by a "thumb" symbol.  For a vertical scrollbar, this means that a '█' appears where the '▼' should appear.

## To Reproduce

```rust
use crossterm::event::read;
use ratatui::{
    layout::Rect,
    widgets::{Scrollbar, ScrollbarOrientation, ScrollbarState},
};
use std::io;

fn main() -> io::Result<()> {
    terminal.draw(|frame| {
        let scrollbar = Scrollbar::new(ScrollbarOrientation::VerticalRight)
            .track_symbol(Some(ratatui::symbols::shade::MEDIUM));
        let mut scroll_state = ScrollbarState::new(9).position(8);
        frame.render_stateful_widget(scrollbar, Rect::new(79, 0, 1, 24), &mut scroll_state);
    })?;
    loop {
        if read()?.is_key_press() {
            break;
        }
    }
    ratatui::restore();
    Ok(())
}
```

## Expected behavior

The ratatui v0.30.0 behavior should have been preserved, in which the end of the scrollbar is always a '▼'.

## Screenshots

With v0.30.1:

Image

With v0.30.0:

Image

## Environment

- OS: macOS Sonoma 14.8.7
- Terminal Emulator: WezTerm
- Font: Menlo
- Crate version: v0.30.1
- Backend: crossterm

## Additional context

I have also noticed that the length of the "thumb" portion of scrollbars has changed by 1 in some circumstances in v0.30.1; was this intentional?

satyakwok:
Closes #2582.

## Problem

When the content is shorter than the viewport, the thumb is large relative to the track. With the position at the end, `part_lengths` clamped `thumb_start` to `track_length - 1` while `thumb_length` was clamped independently to `[1, track_length]`, so `thumb_start + thumb_length` could exceed `track_length`.

`bar_symbols` lays out `begin + track_start + thumb + track_end + end` and zips it against the cells of the area. When the thumb overruns the track, `track_end` saturates to `0` but the thumb still emits more cells than the track can hold, so the trailing `end` symbol is pushed past the end of the area. The last visible cell ends up being a thumb (`█`) where the end arrow (`▼`) should be.

Concretely, for the issue's repro (`VerticalRight`, `content_length = 9`, `position = 8`, height `24`): track is `22`, `thumb_length = 17`, `thumb_start = 6`, and `6 + 17 = 23 > 22`.

This is a regression from v0.30.0, where `thumb_length` was derived as `thumb_end - thumb_start` and therefore always fit within the track.

## Fix

Clamp `thumb_start` to `track_length - thumb_length` (instead of `track_length - 1`) so the thumb always fits within the track and the end symbol is preserved.

## Test

Two regression tests, both fail on `main` and pass with the fix:

- `thumb_stays_within_track_for_large_thumb_at_end` checks `part_lengths` directly with the issue's parameters — asserts `thumb_start + thumb_length <= track_length` and that the parts sum to the track length.
- `render_scrollbar_keeps_end_symbol_for_large_thumb` renders the #2582 case (both arrows, large thumb at the end) and asserts the end symbol is drawn rather than overwritten by a thumb cell.

All existing scrollbar tests still pass.

### DIFF
--- a/ratatui-widgets/src/scrollbar.rs
+++ b/ratatui-widgets/src/scrollbar.rs
@@ -1196,4 +1196,24 @@ mod tests {
             "parts must sum to the track length"
         );
     }
+
+    /// Visual regression for <https://github.com/ratatui/ratatui/issues/2582>: with an end symbol
+    /// set, a large thumb at the end must render the end symbol rather than overwriting it.
+    #[rstest]
+    #[case::large_thumb_at_end("<-----#################>", 8, 9)]
+    fn render_scrollbar_keeps_end_symbol_for_large_thumb(
+        #[case] expected: &str,
+        #[case] position: usize,
+        #[case] content_length: usize,
+    ) {
+        let scrollbar = Scrollbar::new(ScrollbarOrientation::HorizontalTop)
+            .begin_symbol(Some("<"))
+            .end_symbol(Some(">"))
+            .track_symbol(Some("-"))
+            .thumb_symbol("#");
+        let mut buffer = Buffer::empty(Rect::new(0, 0, expected.width() as u16, 1));
+        let mut state = ScrollbarState::new(content_length).position(position);
+        scrollbar.render(buffer.area, &mut buffer, &mut state);
+        assert_eq!(buffer, Buffer::with_lines([expected]));
+    }
 }
--- a/ratatui-widgets/src/scrollbar.rs
+++ b/ratatui-widgets/src/scrollbar.rs
@@ -585,11 +585,14 @@ impl Scrollbar<'_> {
         )
         .clamp(1, track_length);
 
+        // Clamp so the thumb always fits within the track (`thumb_start + thumb_length <=
+        // track_length`). Clamping to `track_length - 1` instead let a large thumb overrun the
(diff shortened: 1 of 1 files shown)


orhun (commented):
Thanks for the fix!

This seems like a regression in https://github.com/ratatui/ratatui/pull/2352, correct?

Maybe it would be nice to even add some visual tests for this as well. (see the rstest changes in the PR above)

satyakwok:
Yes, that's correct.

this is a regression from #2352.

I also added a visual regression test for the #2582 case to make sure the end symbol is preserved.

joshka (commented):
LGTM

joshka (approved):

satyakwok:
Thanks for the review and merge.

PR #2598, opened 2026-06-15
chore: remove unused dependencies

KikiKian:
Audit removes these dependencies that are not used:

  ratatui/Cargo.toml — Removed from [dev-dependencies]:
  - futures
  - rand_chacha
  - tokio
  - tracing
  - tracing-appender
  - tracing-subscriber

  ratatui-core/Cargo.toml — Moved from [dependencies] → [dev-dependencies]:
  - indoc

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2839,13 +2839,11 @@ dependencies = [
  "crossterm 0.29.0",
  "document-features",
  "fakeit",
- "futures",
  "indoc",
  "instability",
  "palette",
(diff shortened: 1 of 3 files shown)


joshka (approved):
Thanks for the PR! I pushed a small Cargo.lock update so locked builds do not try to rewrite the lockfile after the dependency cleanup.

KikiKian:
Thank you! This was actually my first merged PR 😂. First for everything am I right?

PR #2599, opened 2026-06-15
ci: add cargo-udeps dependency check

joshka:
Adds cargo xtask udeps and runs it from CI as a required job.

This complements cargo-machete rather than replacing it. cargo-machete is a fast static source scan, which is why it missed the package-level unused deps fixed in #2598 when the same dependency names were still referenced by example crates. cargo-udeps compiles the workspace and checks rustc dep-info, so it can catch unused dependency declarations for the package being checked.

To make the new job pass, this also removes the remaining true-positive unused dev-deps and records explicit cargo-udeps ignores for current false positives / intentional cases: ratatui-core critical-section, ratatui-crossterm's duplicate crossterm version feature shape, and ratatui-termwiz's doc-example-only ratatui dev-dependency.

I searched existing issues and PRs for udeps / cargo-udeps / "cargo udeps". I did not find prior ratatui discussion about adopting cargo-udeps; the only hits were Dependabot PR bodies for taiki-e/install-action release notes mentioning cargo-udeps version updates, for example #1971, #2095, #2194, and #2522.

Validation:
- cargo xtask udeps
- cargo xtask format --check

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -86,6 +86,24 @@ jobs:
           persist-credentials: false
       - uses: bnjbvr/cargo-machete@ac30a525c0a8d163a92d727b3ff079ee3f6ecb08 # v0.9.2
 
+  # Check for unused dependencies that static source scanning cannot distinguish.
+  # This complements cargo-machete by compiling the workspace and checking rustc dep-info.
+  cargo-udeps:
+    name: Check Compiled Dependencies
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+        with:
+          persist-credentials: false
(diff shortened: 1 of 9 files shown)


orhun (approved):

.github/workflows/ci.yml
```diff
@@ -86,6 +86,24 @@ jobs:
           persist-credentials: false
       - uses: bnjbvr/cargo-machete@ac30a525c0a8d163a92d727b3ff079ee3f6ecb08 # v0.9.2
 
+  # Check for unused dependencies that static source scanning cannot distinguish.
+  # This complements cargo-machete by compiling the workspace and checking rustc dep-info.
+  cargo-udeps:
+    name: Check Compiled Dependencies
```
orhun:
```suggestion
    name: Check Unused Dependencies
```

orhun (approved):

PR #2621, opened 2026-06-28
fix(gauge): correct off-by-one error in filled area condition

### LINKED ISSUE #2547: Gauge renders an extra label cell in the filled area

Infinty-ux:
Fixes #2547 by changing the strict inequality to greater-than-or-equal so the gauge doesn't leave an empty cell after the label.

### DIFF
--- a/ratatui-widgets/src/gauge.rs
+++ b/ratatui-widgets/src/gauge.rs
@@ -195,7 +195,7 @@ impl Gauge<'_> {
                 // Use full block for the filled part of the gauge and spaces for the part that is
                 // covered by the label. Note that the background and foreground colors are swapped
                 // for the label part, otherwise the gauge will be inverted
-                if x < label_col || x > label_col + clamped_label_width || y != label_row {
+                if x < label_col || x >= label_col + clamped_label_width || y != label_row {
                     buf[(x, y)]
                         .set_symbol(symbols::block::FULL)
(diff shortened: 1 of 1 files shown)


orhun:
Can you add a unit/regression test for this as well?

joshka:
Thanks for catching and fixing the Gauge off-by-one. Sorry for the timing—we received three PRs for the same [#2547](https://github.com/ratatui/ratatui/issues/2547) bug. [#2579](https://github.com/ratatui/ratatui/pull/2579) has now been merged with the same boundary correction plus a regression test, so this PR is superseded. I’m closing it as a duplicate. I appreciate the contribution, and apologize that the overlapping PRs made the timing awkward.

PR #2625, opened 2026-06-30
fix(barchart): center wide value labels by display width

mahirhir:
## Description

`Bar::render_value` measures the value label's display width with `UnicodeWidthStr::width`, but the centering offset is computed from `value_label.len()`, the UTF-8 byte length:

```rust
let width = value_label.width() as u16;
// ...
buf.set_string(
    x + (max_width.saturating_sub(value_label.len() as u16) >> 1),
    // ...
```

For a value label containing wide (CJK) or other multi-byte characters, the byte length is greater than the display width, so the padding is too small and the label is drawn left of center. The neighbouring `Bar::render_label` already centers with the display width, so the two paths disagreed for the same input.

The fix reuses the `width` value the function already computed, so `render_value` centers the same way as `render_label`.

## Before / after

Rendering a bar with value label `"中文"` (4 columns wide) in an 8-column bar:

```text
before: █中文███
after:  ██中文██
```

## Testing

Added `wide_value_label_centered_by_display_width` in `barchart.rs`. `cargo test -p ratatui-widgets` is green; the existing ASCII value-label tests are unaffected because byte length equals display width for those.

Disclosure: prepared with AI assistance. I reviewed every line and verified the fix and the test locally.

### DIFF
--- a/ratatui-widgets/src/barchart.rs
+++ b/ratatui-widgets/src/barchart.rs
@@ -1559,4 +1559,18 @@ mod tests {
         // This should not panic, even if the buffer has zero size.
         chart.render(buffer.area, &mut buffer);
     }
+
+    #[test]
+    fn wide_value_label_centered_by_display_width() {
+        // A double-width (CJK) value label must be centered by its display width,
+        // not its byte length. "中文" is 4 columns wide, so in an 8-wide bar it
+        // gets 2 columns of padding on each side. Centering by byte length (6)
+        // would instead give only 1 column on the left.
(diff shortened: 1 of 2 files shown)


mahirhir:
Closing this to clean up my open PRs. It's a small change and not really worth taking up your review time. Sorry for the noise.

joshka:
I actually think this is probably a reasonable thing to fix and this is a perfect PR to do so - no notes that would improve it. Reopening and merging. Thanks for the PR :)

joshka (approved):

Issue #1928, opened 2025-06-24
Ratatui crashes when string splitting localized data


hodgesds:
## Description
See the discussion in this [PR](https://github.com/sched-ext/scx/pull/2248#issue-3173377985) to fix rendering crashes in `scxtop`, which is a tool that uses ratatui.

## To Reproduce
1) Install `scxtop` and run with `cargo install`
```
cargo install scxtop
```
2) Set the system local to non english using `locale`

3) Run `scxtop`  with `sudo scxtop` or run with root.

## Expected behavior
The `scxtop` tui runs.

## Environment
- OS: Arch Linux
- Crate version: 0.29.0"

PR #2629, opened 2026-07-04
fix(barchart): split horizontal value text by display width

SebTardif:
## Summary

Horizontal `BarChart` value labels that overflow the filled bar are split and restyled using **terminal cell width** instead of UTF-8 **byte length**, so multi-byte and wide characters render without gaps or panics.

## Problem

`Bar::render_value_with_different_styles` paints the part of a value that fits inside the filled bar with `value_style`, and the overflow with `bar_style`. The first write correctly used `Buffer::set_stringn` (cell budget), but the overflow path compared `text.len()` (bytes) to `bar_length` (cells), split with `char_indices` against that mixed budget, then placed the second segment at `area.x + first.len()` with `area.width as usize - first.len()`.

### How it fails

1. **Misaligned overflow (multi-byte, single-width cells)**  
   For `text_value = "ααααα"` (each `α` is 2 bytes, 1 cell) and `bar_length = 3`, the bar should show five `α` cells: three with value style, two with bar style. The byte-based split left a filled-block cell in the middle (`ααα α`).

2. **Debug panic (wide CJK on a full bar)**  
   For a full-width bar and a long CJK `text_value`, the "first" segment's **byte** length can exceed `area.width` (cell count). `area.width as usize - first.len()` overflows in debug builds.

### Origin

- Overflow styling path introduced with horizontal bars in [#325](https://github.com/ratatui/ratatui/pull/325) ([`0dca6a689`](https://github.com/ratatui/ratatui/commit/0dca6a689), 2023-08-25).
- A later multi-byte fix in [#1934](https://github.com/ratatui/ratatui/pull/1934) ([`21e3b598`](https://github.com/ratatui/ratatui/commit/21e3b598), 2025-06-25) made the split char-boundary safe but still treated byte indices as a cell budget.
- Related recent work: [#2625](https://github.com/ratatui/ratatui/pull/2625) fixed the same class of bug (byte length vs display width) for **vertical** value label centering.

## Change

- Split and place overflow by Unicode **display width**, aligned with `Buffer::set_stringn`.
- Use saturating arithmetic for remaining width and x offset.
- Add two regression tests in `barchart.rs`.

## Testing

```
Red step (before fix):
  horizontal_multibyte_text_value_split_by_display_width  FAIL
    left:  "ααα α"   (gap at column 3)
    right: "ααααα"
  horizontal_cjk_text_value_does_not_panic_when_longer_than_bar  FAIL
    attempt to subtract with overflow at bar.rs:233

Green step (after fix):
  both tests PASS
  cargo test -p ratatui-widgets --lib barchart  →  50 passed
  cargo clippy -p ratatui-widgets --lib -- -D warnings  →  clean
```

## Related

- Same-repo: [#325](https://github.com/ratatui/ratatui/pull/325), [#1934](https://github.com/ratatui/ratatui/pull/1934), [#2625](https://github.com/ratatui/ratatui/pull/2625), [#1928](https://github.com/ratatui/ratatui/issues/1928) (related multi-byte label panic history)
- No open duplicate PR for the horizontal overflow path

## Disclosure

Prepared with AI assistance. I reviewed every line and verified the fix with red/green tests and the full `barchart` unit suite locally.

### DIFF
--- a/ratatui-widgets/src/barchart.rs
+++ b/ratatui-widgets/src/barchart.rs
@@ -1574,6 +1574,18 @@ mod tests {
         assert_eq!(buffer, Buffer::with_lines(["████████", "██中文██"]));
     }
 
+    /// Empty `text_value` skips value painting and leaves only the bar fill.
+    #[test]
+    fn horizontal_empty_text_value_skips_value_paint() {
+        let chart = BarChart::default()
+            .direction(Direction::Horizontal)
+            .bar_gap(0)
+            .data(BarGroup::default().bars(&[Bar::default().value(5).text_value("")]));
+        let mut buffer = Buffer::empty(Rect::new(0, 0, 5, 1));
+        chart.render(buffer.area, &mut buffer);
+        assert_eq!(buffer, Buffer::with_lines(["█████"]));
+    }
+
     /// Multi-byte single-width characters must be split by display width on
     /// horizontal bars, not by UTF-8 byte length.
     ///
--- a/ratatui-widgets/src/barchart.rs
+++ b/ratatui-widgets/src/barchart.rs
@@ -1573,4 +1573,62 @@ mod tests {
         widget.render(buffer.area, &mut buffer);
         assert_eq!(buffer, Buffer::with_lines(["████████", "██中文██"]));
     }
+
+    /// Multi-byte single-width characters must be split by display width on
+    /// horizontal bars, not by UTF-8 byte length.
+    ///
+    /// `"α"` is two bytes and one cell. With `bar_length == 3` the value
+    /// `"ααααα"` must paint all five cells; the first three use `value_style`
+    /// and the rest use the bar style. Splitting on byte indices leaves a
+    /// filled-block cell in the middle and mis-styles the overflow.
+    #[test]
+    fn horizontal_multibyte_text_value_split_by_display_width() {
+        let bar = Bar::default()
+            .value(3)
+            .text_value("ααααα")
+            .value_style(Style::default().red());
+        let chart = BarChart::default()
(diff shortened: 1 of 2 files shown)


orhun:
Thanks for the PR!

I think this looks good but the new split still operates at the wrong Unicode level.

> `Buffer::set_stringn` renders complete grapheme clusters, while this code uses `char_indices()` and `UnicodeWidthChar`. That means it can split inside a joined emoji or another multi-codepoint grapheme.
>
> For example, `👨‍👩‍👧‍👦` is one displayed grapheme composed of several `char`s. With a two-cell bar, the loop reaches the style boundary after `👨` and then splits before `👩`, in the middle of the family emoji. The overflow render starts with an incomplete grapheme sequence, and the following text can be lost.

To reproduce:

```rust
#[test]
fn horizontal_joined_emoji_is_not_split_inside_a_grapheme() {
    let bar = Bar::default()
        .value(2)
        .text_value("👨‍👩‍👧‍👦x")
        .value_style(Style::default().red());
    let chart = BarChart::default()
        .data(BarGroup::default().bars(&[bar, Bar::default().value(3)]))
        .bar_style(Style::default().yellow())
        .bar_gap(0);

    let mut buffer = Buffer::empty(Rect::new(0, 0, 3, 2));
    chart.render(buffer.area, &mut buffer);

    assert_eq!(buffer[(0, 0)].symbol(), "👨‍👩‍👧‍👦");
    assert_eq!(buffer[(2, 0)].symbol(), "x");
}
```

This currently fails because the last "x" does not render.

So we need to find another way of splitting here. Maybe something like:
> Split with `UnicodeSegmentation::grapheme_indices()` and measure each grapheme using Ratatui's `CellWidth`

PR #2639, opened 2026-07-07
fix(example): handle calendar exit keys

moeleak:

### DIFF
--- a/ratatui-widgets/examples/calendar.rs
+++ b/ratatui-widgets/examples/calendar.rs
@@ -15,7 +15,7 @@
 //! [examples readme]: https://github.com/ratatui/ratatui/blob/main/examples/README.md
 
 use color_eyre::Result;
-use crossterm::event;
+use crossterm::event::{self, Event, KeyCode};
 use ratatui::Frame;
 use ratatui::layout::{Constraint, Layout, Rect};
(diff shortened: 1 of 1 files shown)


ratatui-widgets/examples/calendar.rs
```diff
@@ -29,8 +29,10 @@ fn main() -> Result<()> {
     ratatui::run(|terminal| {
         loop {
             terminal.draw(render)?;
-            if event::read()?.is_key_press() {
-                break Ok(());
```
orhun, line 33: Wasn't this sufficient? It's aligned with the rest of the examples so we prefer to exit when any key is pressed

joshka (approved):
The calendar now exits on q or Esc, filters key presses with as_key_press_event(), and displays matching instructions. Focused formatting and Clippy checks pass, and CI is green.

ratatui-widgets/examples/calendar.rs
joshka: Thanks for the feedback. We chose an explicit q/Esc exit for this demo, consistent with the interactive widget examples (list, tabs, table, and scrollbar), and updated the on-screen instructions to match. The event loop uses as_key_press_event() so only key presses are handled. The previous title said e; that was a typo, now corrected.

PR #2398, opened 2026-02-14
EmbeddedStr POC

junkdog:
POC replacing `CompactString` with a custom `EmbeddedStr` type for storing cell symbols in ratatui's `Buffer`. Refer to https://github.com/ratatui/ratatui/issues/2397

The benchmarks are pretty good (well, it started out as a cpu-optimization attempt last summer), but the real win are the memory savings.

```
  ┌──────────────────┬─────────────┬───────────────┬─────────┐
  │    Benchmark     │ EmbeddedStr │ CompactString │ Speedup │
  ├──────────────────┼─────────────┼───────────────┼─────────┤
  │ buffer_create    │ 5.24 µs     │ 12.10 µs      │ 2.3x    │
  ├──────────────────┼─────────────┼───────────────┼─────────┤
  │ fill_ascii       │ 8.04 µs     │ 56.01 µs      │ 7.0x    │
  ├──────────────────┼─────────────┼───────────────┼─────────┤
  │ fill_box_drawing │ 8.26 µs     │ 20.87 µs      │ 2.5x    │
  ├──────────────────┼─────────────┼───────────────┼─────────┤
  │ read_symbols     │ 2.04 µs     │ 3.94 µs       │ 1.9x    │
  ├──────────────────┼─────────────┼───────────────┼─────────┤
  │ buffer_clone     │ 1.44 µs     │ 13.83 µs      │ 9.6x    │
  ├──────────────────┼─────────────┼───────────────┼─────────┤
  │ buffer_diff      │ 9.94 µs     │ 80.72 µs      │ 8.1x    │
  └──────────────────┴─────────────┴───────────────┴─────────┘
  ```

Issue #2397, opened 2026-02-14
are the mice cheese-starved? memory usage on embedded.


junkdog:
## the tldr
embedded devices have smol memories. we can save some of that memory by replacing `Cell::symbol` with a custom string type. but it should be feature-gated/only applies to embedded. 

### background
MCUs have tiny amounts of memory available (usually around 200kb). since memory is so precious, i've been looking around for places to free up some extra resources on memory-constrained devices. ratatui maintains two screen buffers of `Vec<Cell>`, where each `Cell::symbol` is   represented by a `CompactString` - 12 bytes on 32-bit, 24 bytes on 64-bit architectures (this constitutes ~40% of a Cell's memory usage).

for a 50x24 terminal, that's `50*24*2=2400` cells - or just over 28kb on a 32-bit MCU just for the symbols. ratatui's cells support a wider range of symbols compared to [mousefood](https://github.com/ratatui/mousefood) - mousefood never allocates more than 4 bytes for a symbol. there's a pretty big optimization opportunity here if we add an opt-in feature for enabling a more compact string type.

### proposal

i opened a POC as a draft PR (#2398), so that we have something to discuss around. it revolves around a 4-byte `EmbeddedStr`, which is internal to `Cell::symbol` and replaces the `CompactString`. this resulted in a doubling of framerate when drawing an "idle UI" (just drawing some basic widgets) - running a simple sweeping transition animation was about 30% faster, while memory usage was reduced from 120kb to 90kb. the POC also performs some optimizations around cell width calculation (it's easy on embedded - mousefood only supports single cell width fonts), but those are not memory-related.

benchmark results are posted in the POC PR, but overall it's 2-8x faster than the existing impl.

### tradeoffs

the 4-byte fixed-size approach means `EmbeddedStr` can store a single unicode codepoint, but not multi-codepoint graphemes (emoji ZWJ sequences, combining characters, etc). `append_symbol()` is stubbed out as a no-op. for mousefood and similar embedded use cases, this is a non-issue - but worth mentioning.

j-g00da:
> for mousefood and similar embedded use cases, this is a non-issue

This may still be an issue, it's nice to support these - although since it would be an option for user, it's worth implementing as benefits are quite significant :)

junkdog:
yeah, it doesn't preclude the existing setup (except in the poc). 

[...]

PR #2646, opened 2026-07-14
perf: reduce size of Cell

airblast-dev:
Currently a `Cell` uses 48 bytes. This isn't significant for modern desktops but embedded targets often need to deal with a limited amount of memory.

See: https://github.com/ratatui/ratatui/issues/2397

To reduce the size of a `Cell` we make use of the `sinstr` library which is a small string optimization library aimed to heavily optimize for memory usage when many of them are stored (like in `Buffer`).

Using `sinstr` shrinks the size of a `Cell` to 32 bytes. Allowing to keep more memory available on targets with limited RAM.

This PR is different than https://github.com/ratatui/ratatui/pull/2398 which solves the issue by limiting capacity to 4 bytes.
Instead, on 32 bit targets a `SinStr` will use the same amount of memory as `EmbeddedStr` while still allowing for longer strings but with the cost of only storing 3 bytes inlined.

Because `sinstr` was designed as a more general use library rather than a workaround, it requires very little changes for `ratatui` to use it.

DISCLAIMER: I am the creator and maintainer of `sinstr`. In order to provide compatibility with `ratatui` (and other libraries) I have lowered the MSRV and added `serde` support.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -335,15 +335,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "37b2a672a2cb129a2e41c10b1224bb368f9f37a2b16b612598138befd7b37eb5"
 
 [[package]]
-name = "castaway"
-version = "0.2.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dec551ab6e7578819132c713a93c022a05d60159dc86e7a7050223577484c55a"
-dependencies = [
- "rustversion",
-]
-
-[[package]]
 name = "cc"
 version = "1.2.33"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -537,21 +528,6 @@ dependencies = [
  "crossterm 0.29.0",
  "palette",
  "ratatui",
-]
-
-[[package]]
-name = "compact_str"
-version = "0.9.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9dfdd1c2274d9aa354115b09dc9a901d6c5576818cdf70d14cae2bdb47df00ab"
-dependencies = [
- "castaway",
- "cfg-if",
- "itoa",
- "rustversion",
- "ryu",
- "serde",
- "static_assertions",
 ]
 
 [[package]]
@@ -2867,7 +2843,6 @@ version = "0.1.2"
 dependencies = [
  "anstyle",
  "bitflags 2.13.0",
- "compact_str",
  "critical-section",
  "document-features",
  "hashbrown 0.17.1",
@@ -2880,6 +2855,7 @@ dependencies = [
  "rstest",
  "serde",
  "serde_json",
+ "sinstr",
  "strum",
  "thiserror 2.0.18",
  "unicode-segmentation",
@@ -3497,6 +3473,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "sinstr"
+version = "0.4.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4d749a397a42f203d82876d791faf53953507737954df0229f364ba3f524805e"
+dependencies = [
+ "serde",
+]
+
+[[package]]
 name = "siphasher"
 version = "1.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3566,12 +3551,6 @@ name = "stable_deref_trait"
 version = "1.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a8f112729512f8e442d81f95a8a7ddf2b7c6b8a1a6f509a95864142b30cab2d3"
-
-[[package]]
-name = "static_assertions"
-version = "1.1.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a2eb9349b6444b326872e140eb1cf5e7c522154d69e7a0ffb0fb81c06b37543f"
 
 [[package]]
 name = "strsim"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -62,6 +62,7 @@ ratatui-widgets = { path = "ratatui-widgets", version = "0.3.2", default-feature
 rstest = "0.26"
 serde = { version = "1.0.219", default-features = false, features = ["derive"] }
 serde_json = "1.0.142"
+sinstr = "0.4.4"
 strum = { version = "0.28", default-features = false, features = ["derive"] }
 termina = "0.3"
 termion = "4"
--- a/ratatui-core/Cargo.toml
+++ b/ratatui-core/Cargo.toml
@@ -30,7 +30,6 @@ default = []
 ## enables std
 std = [
   "bitflags/std",
-  "compact_str/std",
   "itertools/use_std",
   "kasuari/std",
   "palette?/std",
@@ -63,12 +62,11 @@ scrolling-regions = []
 
 ## enables serialization and deserialization of style and color types using the [`serde`] crate.
 ## This is useful if you want to save themes to a file.
-serde = ["bitflags/serde", "compact_str/serde", "dep:serde"]
+serde = ["bitflags/serde", "dep:serde", "sinstr/serde"]
 
 [dependencies]
 anstyle = { workspace = true, optional = true }
 bitflags.workspace = true
-compact_str.workspace = true
 critical-section = { workspace = true, optional = true }
 document-features = { workspace = true, optional = true }
 hashbrown.workspace = true
@@ -77,6 +75,7 @@ kasuari = { workspace = true, default-features = false }
 lru.workspace = true
 palette = { workspace = true, optional = true }
 serde = { workspace = true, optional = true }
+sinstr.workspace = true
 strum.workspace = true
 thiserror = { workspace = true, default-features = false }
 unicode-segmentation.workspace = true
--- a/ratatui-core/src/buffer/cell.rs
+++ b/ratatui-core/src/buffer/cell.rs
@@ -1,6 +1,6 @@
 use core::num::NonZeroU16;
 
-use compact_str::CompactString;
+use sinstr::SinStr;
 
 use crate::buffer::cell_width::CellWidth;
 use crate::style::{Color, Modifier, Style};
@@ -39,11 +39,11 @@ pub struct Cell {
     ///
     /// This accepts unicode grapheme clusters which might take up more than one cell.
     ///
-    /// This is a [`CompactString`] which is a wrapper around [`String`] that uses a small inline
+    /// This is a [`SinStr`] which is a wrapper around [`String`] that uses a small inline
     /// buffer for short strings.
     ///
     /// See <https://github.com/ratatui/ratatui/pull/601> for more information.
-    symbol: Option<CompactString>,
+    symbol: Option<SinStr>,
 
     /// The foreground color of the cell.
     pub fg: Color,
@@ -89,11 +89,11 @@ impl Cell {
     ///
     /// This works at compile time and puts the symbol onto the stack. Fails to build when the
     /// symbol doesn't fit onto the stack and requires to be placed on the heap. Use
-    /// `Self::default().set_symbol()` in that case. See [`CompactString::const_new`] for more
+    /// `Self::default().set_symbol()` in that case. See [`SinStr::new_const`] for more
     /// details on this.
     pub const fn new(symbol: &'static str) -> Self {
         Self {
-            symbol: Some(CompactString::const_new(symbol)),
+            symbol: Some(SinStr::new_const(symbol)),
             ..Self::EMPTY
         }
     }
@@ -148,13 +148,13 @@ impl Cell {
             .symbol
             .as_ref()
             .map_or(symbol, |s| strategy.merge(s, symbol));
-        self.symbol = Some(CompactString::new(merged_symbol));
+        self.symbol = Some(SinStr::new(merged_symbol));
         self
     }
 
     /// Sets the symbol of the cell.
     pub fn set_symbol(&mut self, symbol: &str) -> &mut Self {
-        self.symbol = Some(CompactString::new(symbol));
+        self.symbol = Some(SinStr::new(symbol));
         self
     }
 
@@ -169,7 +169,7 @@ impl Cell {
     /// Sets the symbol of the cell to a single character.
     pub fn set_char(&mut self, ch: char) -> &mut Self {
         let mut buf = [0; 4];
-        self.symbol = Some(CompactString::new(ch.encode_utf8(&mut buf)));
+        self.symbol = Some(SinStr::new(ch.encode_utf8(&mut buf)));
         self
     }
 
@@ -328,7 +328,7 @@ mod tests {
         assert_eq!(
             cell,
             Cell {
-                symbol: Some(CompactString::const_new("あ")),
+                symbol: Some(SinStr::new("あ")),
                 fg: Color::Reset,
                 bg: Color::Reset,
                 #[cfg(feature = "underline-color")]
--- a/ratatui-core/src/buffer/diff.rs
+++ b/ratatui-core/src/buffer/diff.rs
@@ -207,10 +207,9 @@ const fn is_skip(cell: &Cell) -> bool {
 
 #[cfg(test)]
 mod tests {
+    use alloc::string::String;
     use alloc::vec::Vec;
     use core::num::NonZeroU16;
-
-    use compact_str::CompactString;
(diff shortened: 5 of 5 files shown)


airblast-dev:
Failing CI seems to be unrelated.

joshka:
Thanks for taking this on, and especially for adapting `sinstr`'s MSRV and `serde` support for Ratatui. The memory pressure described in [#2397](https://github.com/ratatui/ratatui/issues/2397) is real: people have reported running out of memory on ESP32-class devices and serial-console applications, so reducing the buffer footprint is worth investigating.

That said, I don't think we should merge this implementation as it stands. I have two separate concerns: the performance-engineering case and the dependency and supply-chain tradeoff.

## Performance engineering

Reducing `Cell` from 48 bytes to 32 bytes is a useful starting measurement, but it does not yet tell us whether the overall change is worthwhile.

Performance changes tend to leave permanent costs behind: specialized representations, additional edge cases, less obvious code, API constraints, and maintenance work. Everyone working on the code pays those costs, while the benefit may apply only to a narrow workload. Without a disciplined measurement process, it is easy to optimize something that looks expensive but does not matter in a real application, mistake benchmark noise for an improvement, move a cost elsewhere, or report a large percentage improvement that saves very little absolute time or memory.

This is a broader project problem rather than something I want to lay entirely on this PR. We have not documented our expectations for performance work well enough. Performance proposals often begin with an implementation and a plausible explanation for why it should be better, when the implementation should follow defining the workload, measuring the baseline, locating the cost, and stating a testable hypothesis.

I've opened [#2647](https://github.com/ratatui/ratatui/pull/2647) to document why this rigor is needed, the step-by-step process we expect, which tools answer which questions, what data to record, and how to decide whether a measured benefit pays for the tradeoffs. I don't want to apply an undocumented standard retroactively without also fixing the documentation gap on our side.

For this change, I would want the evidence to cover at least:

- representative 32-bit embedded and 64-bit desktop workloads;
- total buffer and application memory savings, not only `size_of::<Cell>()`;
- allocation counts and memory use for realistic symbol distributions, including symbols that exceed the inline capacity;
- rendering, buffer creation, cloning, and diffing costs where the representation is relevant;
- both absolute and relative results, with repeated sequential runs and raw data;
- the commands, targets, toolchain, features, inputs, and environment used;
- correctness and compatibility checks, including whether existing `const Cell::new` values stop compiling when they no longer fit inline; and
- cases that regress as well as those that improve.

The relevant decision is whether the memory saved in representative applications is large enough to pay for additional allocations, runtime regressions, compatibility changes, and implementation complexity.

[#231](https://github.com/ratatui/ratatui/pull/231) is a useful example. The proposed string optimization sounded plausible but measurement found a 20-40% regression, so it was closed. That was the process working correctly: the evidence changed the decision.

## Dependency and supply chain

Separately, I am not comfortable making `sinstr` load-bearing inside one of Ratatui's core data structures at this stage.

Dependency popularity is not itself a quality metric, and I am not making a judgment about this implementation from the download count. The concern is the difference in scale and external validation. At the time of writing, crates.io reports 37,650,116 downloads for [`ratatui`](https://crates.io/crates/ratatui) and 102 for [`sinstr`](https://crates.io/crates/sinstr). Moving a foundational type onto such a young dependency would expose a much larger downstream user base to code that has had very little use or independent scrutiny.

For a dependency in this position, we need to evaluate more than its immediate functionality:

- maturity and real-world use;
- unsafe-code and soundness review;
- maintenance continuity and bus factor;
- release and compatibility history;
- behavior across Ratatui's supported targets and configurations;
- whether Ratatui would effectively become responsible for auditing or maintaining it; and
- how difficult it would be to replace if its API or maintenance situation changed.

Your ownership of the dependency is clearly disclosed, which I appreciate. It also means there has been limited independent validation so far. Given where this type sits in Ratatui, my current view is that we should not adopt `sinstr` here yet, even if further measurements show that the representation is promising. I also don't want to ask you for a large benchmark matrix under the impression that measurements alone would settle the dependency concern.

The investigation is still useful. Measurements grounded in the applications reported in [#2397](https://github.com/ratatui/ratatui/issues/2397), with [#2398](https://github.com/ratatui/ratatui/pull/2398) as another comparison point, can tell us how much this problem matters and help evaluate other implementations or broader buffer designs. But the performance case and the supply-chain decision both need to stand on their own before we put a new representation into `Cell`.

airblast-dev:
> Separately, I am not comfortable making sinstr load-bearing inside one of Ratatui's core data structures at this stage.
Dependency popularity is not itself a quality metric, and I am not making a judgment about this implementation from the download count. The concern is the difference in scale and external validation. At the time of writing, crates.io reports 37,650,116 downloads for [ratatui](https://crates.io/crates/ratatui) and 102 for [sinstr](https://crates.io/crates/sinstr). Moving a foundational type onto such a young dependency would expose a much larger downstream user base to code that has had very little use or independent scrutiny.

`sinstr` is definitely very young so I understand your stance. The library was specifically designed TUI's in mind (https://github.com/thehuglet/germterm to be specific). Its use outside of such context is really limited do to the limited inline capacity it provides so I don't really expect a library like this to become popular.

While I do understand your concerns the wording makes it sound like only a large organization or very popular person can solve it when its a pretty simple issue.

I don't mind putting effort for benchmarks and etc. but I don't think it makes sense for me to put in the effort when the PR already has concerns that I cannot fix.

If anyone is interested in doing some benchmarks etc. feel free to do so. I am not interested in putting any effort towards it if the PR is basically rejected due to the projects popularity.

joshka:
That’s fair feedback. I put too much weight on the download comparison, and the way I phrased it made this sound like a popularity threshold that you could not reasonably address. I did not mean to imply that only a large organization or an already-popular maintainer can solve this problem.

The concern I was trying to express is about ownership rather than popularity. If this representation becomes load-bearing inside `Cell`, Ratatui needs to be confident that we can review, test, maintain, and change it for our downstream users. Given how specialized and relatively small the implementation is, pulling the necessary representation into `ratatui-core` may be a better boundary than making the young `sinstr` crate a foundational dependency. That would put the code, tests, unsafe-code review, compatibility decisions, and long-term maintenance in the same project that depends on them.

That is a concern the proposal can address. A possible path forward is:

- use this PR and the embedded testing to establish whether the representation provides a worthwhile benefit;
- compare memory use and any allocation or runtime tradeoffs on representative embedded and desktop workloads;
- resolve compatibility details such as the reduced inline capacity of `const Cell::new`; and
- if the results support the design, consider contributing the required implementation directly to Ratatui rather than depending on `sinstr`.

I should have made that path explicit before asking for more measurements. It was reasonable to read my previous comment as “do all this benchmarking, but the dependency will still prevent the PR from merging,” which gives you little reason to invest more effort.

To be clear, the underlying memory problem is not rejected, the representation idea is not rejected, and testing on ESP32, nRF52840, RP2040, or other constrained targets would be useful. My current concern is with merging the external-dependency form without enough evidence or project ownership of such a central implementation.

There is no obligation for you to continue if that is no longer work you want to take on, but I don’t want other people to come away thinking that testing this is pointless or that the idea has already been dismissed. Thanks for calling out where my earlier wording left the wrong impression.

airblast-dev:
These are some numbers I took using the `sparkline` example with a width and height of 213x49 cells.
"main.txt" and "sinstr.txt" is benchmarks (take these with a grain of salt as I don't have a clean system to run them)
"\*_mem\*.txt" are memory usage captured via `pmap -x <PID>` 32 variants compiled for 32 bit systems

On 64 bit platforms the saving is `332KB` and on 32 bit platforms the saving is `168KB`.
[main_mem.txt](https://github.com/user-attachments/files/30025243/main_mem.txt)
[main.txt](https://github.com/user-attachments/files/30025247/main.txt)
[sinstr.txt](https://github.com/user-attachments/files/30025248/sinstr.txt)

[main_mem32.txt](https://github.com/user-attachments/files/30025244/main_mem32.txt)
[sinstr_mem.txt](https://github.com/user-attachments/files/30025245/sinstr_mem.txt)
[sinstr_mem32.txt](https://github.com/user-attachments/files/30025246/sinstr_mem32.txt)

While these numbers are pretty good (considering most of the memory usage is from libc) these will like be at much different scales in an embedded environment. 

I do have a few displays and ESP32's lying around but I am not really experienced enough in embedded systems to provide proper measurements. So I would appreciate if anyone could give me some tips or share some measurements.

orhun:
In terms of measurements, I would be happy to help you out on that. Also when it comes to embedded + Ratatui, there are a couple of interested people in the Discord server too, so maybe drop this link there as well.

Just pinging a couple of friends for their opinions: @Vaishnav-Sabari-Girish @junkdog @ImplFerris

airblast-dev:
> In terms of measurements, I would be happy to help you out on that.

Sounds great!

> Also when it comes to embedded + Ratatui, there are a couple of interested people in the Discord server too, so maybe drop this link there as well.

I already posted the PR in the discords embedded channel yesterday in case anyone is interested :)

joshka:
Thanks for continuing with the measurements. Saving 332 KB for a 213×49 buffer on 64-bit and 168 KB on 32-bit is meaningful, and it gives us a much better picture of the fixed memory benefit than `size_of::<Cell>()` alone.

Looking through the history again, I realized there is one substantive concern that I did not explain clearly in my earlier review.

`Cell::symbol` is intentionally not limited to a `char` or an ordinary short string. It stores a complete Unicode grapheme cluster, which can contain multiple code points, variation selectors, combining marks, skin tones, regional indicators, or ZWJ sequences. Rendering can also append zero-width graphemes to the symbol already stored in a cell.

This was part of the reason Ratatui moved from `String` to `CompactString` in [#601](https://github.com/ratatui/ratatui/pull/601). Its roughly 24-byte inline capacity keeps most grapheme clusters inline while retaining heap storage for longer sequences. That capacity is part of the existing performance design rather than simply unused space in `Cell`.

`SinStr` changes that tradeoff substantially. It stores 7 bytes inline on 64-bit targets and 3 bytes on 32-bit targets. Flags and emoji with skin tones are commonly 8 bytes; emoji containing a variation selector, such as `⌨️` or `⚠️`, are 6 bytes; many individual emoji require 4 bytes; and ZWJ sequences can be considerably longer.

On a 32-bit target, even a four-byte emoji or a character followed by a variation selector will move to the heap. Flags, skin-tone sequences, and many other composed graphemes also exceed the inline capacity on 64-bit targets.

This does not necessarily make the smaller representation a net loss. Reducing every `Cell` may save substantially more memory than the additional allocations consume. However, allocation count and heap fragmentation matter particularly on the embedded targets this change is intended to help, so they need to be included in the tradeoff.

The current sparkline measurements are useful, but that example mostly contains ASCII and three-byte block characters. Those fit within `SinStr`’s inline capacity and are close to the favorable case for this representation. I would also measure allocation count, allocated bytes, and peak memory with content containing variation-selector emoji, flags, combining sequences, and ZWJ graphemes. We do not need every possible Unicode sequence, but we should include enough representative content to tell whether the smaller inline capacity changes the result materially.

My preferred direction is to keep the default path simple and predictable while making room for a more RAM-efficient embedded configuration.

For the general-purpose and desktop case, `CompactString` is a known, mature dependency with an inline capacity that handles most Unicode grapheme clusters without allocating. There is value in keeping that as the default rather than replacing it globally with a more specialized representation.

That should not block us from improving Ratatui for smaller targets. An opt-in feature could use a smaller `Cell` representation for applications where fixed buffer size is the dominant constraint and the user is willing to accept more allocations for complex graphemes. The embedded configuration would need to demonstrate a net memory benefit and preserve rendering correctness, but it would not need to be the right tradeoff for every Ratatui application.

This changes the question from “should Ratatui replace `CompactString` for everyone?” to “can Ratatui support a well-defined low-memory storage mode alongside the existing default?” I think the second question has a much more promising answer.

A feature has some constraints that we would need to work through:

- Cargo features are unified across the dependency graph, so this would select one global `Cell` representation rather than allowing each `Buffer` to choose independently.
- Both representations must preserve the same rendering and Unicode semantics.
- We would need tests for the default and low-memory configurations, including `no_std`, `serde`, 32-bit and 64-bit targets, and representative grapheme clusters.
- We should avoid exposing the selected storage type through the public API.
- Enabling the feature must not unexpectedly break downstream code.

The last point is especially relevant to `const Cell::new`. Its current capacity depends on the inline storage. If enabling the feature reduces the set of constant symbols that compile, the feature changes a public API contract rather than only changing memory use. Because features can be enabled transitively, an application could otherwise break when another dependency enables the low-memory mode. We would need to resolve that behavior or make the limitation an explicit part of the feature’s contract.

The feature also gives us a better boundary for the dependency question. `CompactString` can remain the default dependency for the main configuration. The smaller representation could be optional, limiting its impact to users who choose the low-memory mode. We should still decide whether the implementation belongs directly in `ratatui-core` or in an optional external dependency, but that decision no longer needs to affect every Ratatui user.

I would start with a compile-time feature rather than making `Cell` or `Buffer` generic over a storage type. Generics would provide true per-type pluggability, but they would spread the storage choice through widgets and public APIs. That is a much larger compatibility and maintenance cost than an opt-in configuration.

So I think there is a viable path here:

1. Measure the fixed savings and the Unicode/allocation tradeoff on representative workloads.
2. Keep `CompactString` as the simple, known default for general-purpose applications.
3. Explore an opt-in low-memory storage mode for constrained targets.
4. Decide whether Ratatui should own that implementation or use an optional dependency.
5. Resolve the `const Cell::new` and feature-unification compatibility questions before treating the mode as supported.

The measurements already posted are useful progress. The remaining Unicode cases matter because they exercise the reason `Cell` stores a string rather than a `char` in the first place, but they should inform the low-memory mode rather than prevent us from supporting one.

Vaishnav-Sabari-Girish:
I'll try it out soon.

airblast-dev:
> This was part of the reason Ratatui moved from String to CompactString in https://github.com/ratatui/ratatui/pull/601. Its roughly 24-byte inline capacity keeps most grapheme clusters inline while retaining heap storage for longer sequences. That capacity is part of the existing performance design rather than simply unused space in Cell.
>
>SinStr changes that tradeoff substantially. It stores 7 bytes inline on 64-bit targets and 3 bytes on 32-bit targets. Flags and emoji with skin tones are commonly 8 bytes; emoji containing a variation selector, such as ⌨️ or ⚠️, are 6 bytes; many individual emoji require 4 bytes; and ZWJ sequences can be considerably longer.
>
>On a 32-bit target, even a four-byte emoji or a character followed by a variation selector will move to the heap. Flags, skin-tone sequences, and many other composed graphemes also exceed the inline capacity on 64-bit targets.

These are valid concerns. 3 bytes is pretty low for 32-bit targets and could have some negative performance implications. While I think it would be worth while trade-off considering the alternative is running out of memory, there might be a middle ground. 

I don't want to give promises but I wonder if `sinstr` could expose a const generic integer to specify extra bytes that can be stored inline. That would allow us to increase the capacity we store inline at the cost of (possibly) higher memory usage due to `Cell`'s size increasing.

> The feature also gives us a better boundary for the dependency question. CompactString can remain the default dependency for the main configuration. The smaller representation could be optional, limiting its impact to users who choose the low-memory mode. We should still decide whether the implementation belongs directly in ratatui-core or in an optional external dependency, but that decision no longer needs to affect every Ratatui user.

That makes sense. Giving a solution and keeping the default as-is is a reasonable way to handle it.

# Viable Path
> Measure the fixed savings and the Unicode/allocation tradeoff on representative workloads.

What would be considered a representative workload? I definitely agree that more than a single example should be tested, just wondering which project or example would be a better choice.

> Keep CompactString as the simple, known default for general-purpose applications.

This makes sense, maybe `sinstr` could have more comparable storage to `CompactString` after the changes mentioned above making it a better default but as it is I definitely agree `CompactString` is the better default. (especially for 32 bit platforms)

> Decide whether Ratatui should own that implementation or use an optional dependency.

To clarify my preference would by order be:
- `sinstr` is moved under `ratatui` where I still maintain it with other maintainers having the ability to step in.
- `sinstr` is used as an external dependency.
- The projects implementation is copied to `ratatui-core` which means the `ratatui` organization is now directly responsible for its maintenance.

> Resolve the const Cell::new and feature-unification compatibility questions before treating the mode as supported.

This ones a bit difficult. The capacity difference in `const Cell::new` is something that most solutions will have to deal with as there will be an unavoidable trade-off, question is: It worth it?

Vaishnav-Sabari-Girish:
Hey @airblast-dev , do you have an example using this PR's code for an ESP32 or something. Just to get an idea of what exactly you tried so that I could test it out more. Thank you.

orhun:
> sinstr is moved under ratatui where I still maintain it with other maintainers having the ability to step in.

This needs to be used in Ratatui first (for more adoption) to see if there are any edge cases before we could help with the maintenance.

PR #2668, opened 2026-07-24
feat(table): build rows lazily for large datasets

### LINKED ISSUE #5: Table is slow/laggy when there are a lot of items
## Description

In one of my TUI projects I have 15k items in a `Table` and the render is pretty slow and laggy. I press the key to scroll the table but it takes around 1-2 seconds.

## To Reproduce

To confirm it, I tried this with the table example as well. 

Just apply this patch:

```diff
diff --git a/examples/table.rs b/examples/table.rs
index 744e142..94afc9b 100644
--- a/examples/table.rs
+++ b/examples/table.rs
@@ -156,7 +156,7 @@ impl App {
 fn generate_fake_names() -> Vec<Data> {
     use fakeit::{address, contact, name};
 
-    (0..20)
+    (0..15000)
         .map(|_| {
             let name = name::full();
             let address = format!(
```

rec_20240328T224629

## Expected behavior

Normal table performance.

## Screenshots

Provided above.

## Environment

- OS: Arch Linux
- Terminal Emulator: alacritty
- Font: Ubuntu Mono
- Crate version: 0.24.0
- Backend: crossterm

## Additional context

- running this with `--release` make things a bit faster but it is still laggy.
- https://discord.com/channels/1070692720437383208/1072907135664529508/1223034251713318922

> the issue is the conversion to vec in the construction of the table.
> For the example, the right idea would be to do that just once rather than every time.
> To fix the perf issue fully would require rewriting the code that chooses the rows to display to be more aware of lazy evaluating iterators

rit3sh-x:
Closes ratatui/ratatui-table#5

## What

Adds `Table::lazy_rows`, which builds rows on demand instead of collecting them into a `Vec` up front, so a table over a very large dataset costs the viewport rather than the dataset.

```rust
let table = Table::lazy_rows(10_000_000, widths, |index| Row::new([format!("item {index}")]))
    .row_height_with(|index| if index % 25 == 0 { 3 } else { 1 });
```

- `Table::lazy_rows(row_count, widths, factory)` — rows are one line tall by default.
- `Table::row_height(u16)` — uniform height override.
- `Table::row_height_with(impl Fn(usize) -> u16)` — per-row heights.

All three are additive; no existing API changes behaviour.

`examples/apps/lazy-table` renders ten million rows and reports how many rows each frame actually built and how long it took to draw:

```text
 row 7884 | variable heights | 45 rows built in 9.66ms
```

## Why heights are declared separately

The scroll pass needs the height of rows it will not draw — for instance to work out where the window starts when the selection is below the viewport. For an eager table that is free, because `visible_rows` can index `self.rows`. For a lazy table, asking a row for its height means building it, which is the cost the feature exists to avoid. Declaring heights separately keeps the whole frame proportional to the viewport, and matches how other virtualised lists work (Qt `sizeHintForRow`, UIKit `heightForRowAt`, react-window `itemSize`).

Consequence, documented on the constructor: for lazy rows the declared height is authoritative and `Row::height` / `Row::top_margin` / `Row::bottom_margin` on the returned row are ignored.

## Consistency with existing code

`LazyRowHeights` follows the same shape as `Effect` in `block/shadow.rs`: an enum with cheap built-in variants plus one `Arc<dyn …>` variant, with hand-written `PartialEq` (`Arc::ptr_eq`), `Eq` and `Hash` (`ptr::hash`) so the widget keeps its derives.

It differs in one respect: `CellEffect` requires `Send + Sync + UnwindSafe + RefUnwindSafe`, whereas the lazy closures require only `Send + Sync + RefUnwindSafe`. `Arc<T>: UnwindSafe` needs just `T: RefUnwindSafe`, so the extra bound would restrict callers without buying anything — `Table<'static>` remains `UnwindSafe`, as `ratatui-widgets/tests/auto_traits.rs` already asserts. Happy to add `UnwindSafe` for symmetry if you would rather the two APIs read the same.

## Please look closely at

1. **`LazyRowProvider::visible_rows`** mirrors `Table::visible_rows` step for step, but grows the window upward from the selection instead of scrolling down one row at a time. Two tests assert the two produce identical buffers and offsets across a matrix of row heights, area heights, offsets and selections.
2. **One deliberate divergence**: when the selected row is taller than the whole area, the eager algorithm shrinks `start` past the selection and draws the row *below* it, leaving the selected row off screen. The lazy path keeps it anchored and clips it. I left the eager behaviour alone and documented the difference — happy to open a separate fix if you agree it is a bug.
3. **`row_height` / `row_height_with` are no-ops on eager tables**, since `Table::new` rows carry their own `Row::height`. Documented on both methods and covered by a test. Open to a better name or shape here.

## Testing

- New `ratatui/tests/widgets_table_lazy_rows.rs`: laziness (asserting the exact set of rows built), offsets, selection anchoring, partial rows, headers/footers/blocks, highlight styles, zero-height rows, clone/eq/hash, and equivalence with `Table::new`.
- `cargo xtask ci` passes locally.

## AI assistance
This uses AI for comments and logic validation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1893,6 +1893,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bf36173d4167ed999940f804952e6b08197cae5ad5d572eb4db150ce8ad5d58f"
 
 [[package]]
+name = "lazy-table"
+version = "0.0.0"
+dependencies = [
+ "color-eyre",
+ "crossterm 0.29.0",
+ "ratatui",
+]
+
+[[package]]
 name = "lazy_static"
 version = "1.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/examples/README.md
+++ b/examples/README.md
@@ -152,6 +152,11 @@ Shows how to use the inlined viewport to render in a specific area of the screen
 
 Shows how to render a form with input fields. [Source](./apps/input-form/).
 
+## Lazy Table
+
+Shows how to display a table with ten million rows that are built on demand.
+[Source](./apps/lazy-table/).
+
 ## Modifiers
 
 Shows different types of modifiers. [Source](./apps/modifiers/).
--- a/examples/apps/lazy-table/Cargo.toml
+++ b/examples/apps/lazy-table/Cargo.toml
@@ -1,0 +1,14 @@
+[package]
+name = "lazy-table"
+publish = false
+license.workspace = true
+edition.workspace = true
+rust-version.workspace = true
+
+[dependencies]
+color-eyre.workspace = true
+crossterm.workspace = true
+ratatui.workspace = true
+
+[lints]
+workspace = true
--- a/examples/apps/lazy-table/README.md
+++ b/examples/apps/lazy-table/README.md
@@ -1,0 +1,13 @@
+# Lazy table demo
+
+This example shows how to display a table with 10 million rows using lazy implementation.
+
+The status bar reports how many rows the table actually built for the last frame and how long the
+frame took, so you can watch scrolling stay flat no matter how far into the dataset you jump. Press
+`v` to switch between uniform and variable row heights.
+
+To run this demo:
+
+```shell
+cargo run -p lazy-table
+```
--- a/examples/apps/lazy-table/src/main.rs
+++ b/examples/apps/lazy-table/src/main.rs
@@ -1,0 +1,205 @@
+/// A Ratatui example that demonstrates a table with ten million rows.
+///
+/// [`Table::lazy_rows`] builds rows on demand, so only the rows that are actually on screen
+/// are ever created. The status bar shows how many rows
+/// the last frame built and how long it took to draw, which stays flat however far into the
+/// dataset you scroll. Press `v` to switch between uniform and variable row heights.
+///
+/// This example runs with the Ratatui library code in the branch that you are currently
+/// reading. See the [`latest`] branch for the code which works with the most recent Ratatui
+/// release.
+///
+/// [`latest`]: https://github.com/ratatui/ratatui/tree/latest
+use std::sync::Arc;
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::time::{Duration, Instant};
+
+use color_eyre::Result;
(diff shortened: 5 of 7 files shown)


joshka:
The Table experiment now lives at https://github.com/ratatui/ratatui-table. The original performance discussion was transferred to https://github.com/ratatui/ratatui-table/issues/5, and focused successor tracking for this PR is at https://github.com/ratatui/ratatui-table/issues/8.

@rit3sh-x, would you be interested in recreating this as a focused PR against the extracted 0.1.0 baseline? We would also like to offer you opt-in Write access so proposal authors can review and collaborate without waiting on the core Ratatui repository. Releases and protected configuration remain maintainer-controlled; reply here if you would like the invitation.

rit3sh-x:
> The Table experiment now lives at https://github.com/ratatui/ratatui-table. The original performance discussion was transferred to https://github.com/ratatui/ratatui-table/issues/5, and focused successor tracking for this PR is at https://github.com/ratatui/ratatui-table/issues/8.
> 
> @rit3sh-x, would you be interested in recreating this as a focused PR against the extracted 0.1.0 baseline? We would also like to offer you opt-in Write access so proposal authors can review and collaborate without waiting on the core Ratatui repository. Releases and protected configuration remain maintainer-controlled; reply here if you would like the invitation.
> 
> I am leaving this PR open until a successor exists or you prefer to close it.
> 

Hey @joshka I'd love to work on it.

rit3sh-x:
@joshka will the table be removed from this repo following next release?

rit3sh-x:
Closed with related PR: ratatui/ratatui-table#14

joshka:
> @joshka will the table be removed from this repo following next release?

No - we expect that people that want to opt-in to being able to be a bit looser with breaking changes would go grab ratatui-table. Those that just want stuff that they know works would stick with the built-in ratatui table. For a long time I've often had difficulty finding a balance between stability and new features. This is an experiment in having both.

What I do  intend to do though is make sure that the project is linked from the table docs at some point.

https://github.com/ratatui/ratatui-table/issues/7 has much more details on this.

rit3sh-x:
@joshka So should I raise PR on this repo for the lazy table examples? Once my table PR gets merged.

PR #2670, opened 2026-07-25
fix(terminal)!: don't clear the whole screen when an inline viewport shrinks horizontally

### LINKED ISSUE #2666: `Viewport::Inline` still duplicates the live viewport into scrollback on resize with continuous draw + `insert_before` (incomplete fix for #2086 / #2355)
## Description

#2086 reported that resizing an inline viewport duplicates content, and #2355 fixed it by clearing the screen and moving the viewport to the top on horizontal shrink.

However, the duplication **still reproduces on the latest stable (`ratatui` 0.30.2 / `ratatui-core` 0.1.2)** when the application draws continuously and commits lines via `Terminal::insert_before` — i.e. the standard inline-UI pattern from `examples/apps/inline` (draw every tick, `autoresize()` on `Event::Resize`, `insert_before` for completed lines).

Dragging the window (a stream of resize events) while the app keeps drawing leaves a **staircase of duplicated copies of the live viewport** in the scrollback — one (or more) stale copy per resize event.

## To Reproduce

`Cargo.toml`:

```toml
[dependencies]
ratatui = "0.30.2"
```

`src/main.rs`:

```rust
use std::io::stdout;
use std::time::Duration;

use ratatui::backend::CrosstermBackend;
use ratatui::crossterm::event::{self, Event, KeyCode};
use ratatui::crossterm::terminal::{disable_raw_mode, enable_raw_mode};
use ratatui::text::Line;
use ratatui::{Terminal, TerminalOptions, Viewport};

fn main() -> std::io::Result<()> {
    enable_raw_mode()?;
    let mut terminal = Terminal::with_options(
        CrosstermBackend::new(stdout()),
        TerminalOptions {
            viewport: Viewport::Inline(1),
        },
    )?;

    let mut n = 0u32;
    let result = (|| -> std::io::Result<()> {
        loop {
            if event::poll(Duration::from_millis(400))? {
                match event::read()? {
                    Event::Key(k) if k.code == KeyCode::Char('q') => break,
                    Event::Resize(_, _) => {
                        terminal.autoresize()?;
                    }
                    _ => {}
                }
            } else {
                // Commit a permanent line into the scrollback (intended use of
                // insert_before for an inline UI).
                n += 1;
                let line = format!("committed line {n}");
                terminal.insert_before(1, move |buf| {
                    let area = buf.area;
                    Paragraph::new(Line::from(line)).render(area, buf);
                })?;
            }
            // Idempotent live tail — same state redraws to the same 1-row viewport.
            terminal.draw(|f| {
                let area = f.area();
                f.render_widget(
                    Paragraph::new(Line::from(format!(
                        "LIVE TAIL n={n} — now drag the window NARROWER"
                    ))),
                    area,
                );
            })?;
        }
        Ok(())
    })();

    disable_raw_mode()?;
    result
}
```

Steps:

1. Run in a real terminal.
2. Let ~10+ `committed line N` rows scroll up into the scrollback.
3. Drag the window **narrower** (repeatedly / continuously).

## Actual behavior

Each resize event leaves a duplicated copy of the live viewport in the scrollback. Scrolling up shows a staircase like:

```
committed line 42
LIVE TAIL n=42 — now drag the window NARROWER
LIVE TAIL n=42 — now drag the window NARROWER
LIVE TAIL n=42 — now drag the window NARROWER
LIVE TAIL n=42 — now drag the window NARROWER
... (one copy per resize event) ...
committed line 43
LIVE TAIL n=43 — now drag the window NARROWER
LIVE TAIL n=43 — now drag the window NARROWER
...
```

Image

## Expected behavior

Resizing should not duplicate the live viewport into the scrollback. #2355 eliminated this for a single static draw; the same should hold for continuous draw + `insert_before`.

## Additional context

- #2355's clear-screen-on-horizontal-shrink approach assumes the app re-renders its **full visible state** each frame (true for a self-contained viewport, e.g. the download example). It does not fit an inline UI that persists **completed** output to the scrollback via `insert_before` and only redraws the live tail: on shrink the committed rows are cleared but never re-rendered by the app.
- The abandoned Solution 1 (#2349, "scroll down") would have handled the duplication. Its stated con — off-screen content can't be scrolled, leaving a history gap — is avoidable if the caller keeps its own store of committed rows and repaints the visible ones at the new width (this is how we ended up working around it downstream, by taking over the resize path entirely instead of calling `autoresize`).

## Environment

- OS: Windows 11
- Terminal Emulator: Windows Terminal (PowerShell 7.7.0-preview.2)
- Crate version: `ratatui` 0.30.2 (latest stable), `ratatui-core` 0.1.2
- Backend: crossterm

easyinplay:
Fixes #2666.

`Terminal::resize` clears the entire screen (`ClearType::All`) on a horizontal shrink. For fullscreen and fixed viewports that's harmless, but an inline viewport only owns the rows from its origin down. The rows above it were written by `insert_before`, which keeps no copy of them, so the application can never repaint them. The full-screen erase either destroys that output (terminals that discard erased content) or, on terminals that move erased content into scrollback (Windows Terminal, conhost), pushes a copy of the live viewport into scrollback on every resize event — that's the staircase of duplicated live rows in #2666.

The erase is also redundant for inline viewports: `resize` calls `clear_viewport()` right after, which erases from the recomputed origin to the bottom of the screen. The origin is anchored to the live cursor, and everything the viewport can legitimately own is at or below it.

This change skips the full-screen erase (and the accompanying `y = 0` reset) for `Viewport::Inline` only. Fullscreen and fixed viewports are unchanged.

Testing:

- new regression test `resize_inline_horizontal_shrink_keeps_rows_above_the_viewport` fails before the change (the rows above the viewport are wiped) and passes after; the previous shrink test is kept as `resize_inline_resets_the_buffers_on_horizontal_shrink` since the buffer-reset side of it still holds
- verified on Windows Terminal with a streaming inline app (continuous `draw` plus periodic `insert_before`) while dragging the window narrower column by column: committed rows survive and scrollback no longer accumulates one copy of the live row per resize event
- `cargo test -p ratatui-core` with default features, `--no-default-features`, `--features scrolling-regions` and `--all-features` all pass

Related: #2086 / #2355 addressed the shrink erase for applications that repaint every visible cell each frame; `insert_before` output is written once and never repainted, which is the case this covers.

### DIFF
--- a/ratatui-core/src/terminal/resize.rs
+++ b/ratatui-core/src/terminal/resize.rs
@@ -38,8 +38,20 @@ impl<B: Backend> Terminal<B> {
             Viewport::Fixed(_) | Viewport::Fullscreen => (area, None),
         };
 
-        // clear screen on horizontal shrink to avoid line wrapping issues
-        if next_area.width < self.viewport_area.width {
+        // Clear the screen on horizontal shrink to avoid line wrapping issues.
+        //
+        // Inline viewports are excluded: an inline viewport only owns the rows
+        // from its origin down. The rows above it were written by
+        // `insert_before`, which keeps no copy of them, so the application can
+        // never repaint them. A full-screen erase either destroys that output
+        // or, on terminals that move erased content into scrollback (Windows
+        // Terminal, conhost), pushes a copy of the viewport into scrollback on
+        // every resize event. The `clear_viewport` call below already erases
+        // from the recomputed origin to the bottom of the screen, which covers
+        // every row the inline viewport can legitimately own.
+        if next_area.width < self.viewport_area.width
+            && !matches!(self.viewport, Viewport::Inline(_))
+        {
             next_area.y = 0;
             self.backend.clear_region(ClearType::All)?;
         }
@@ -305,11 +317,69 @@ mod tests {
         );
     }
 
-    // This tests for the case where the new width is smaller than the old
-    // width. The screen should be cleared completely to avoid rendering
-    // glitches caused by line wrap.
+    // An inline viewport does not own the rows above its origin: they were
+    // written by `insert_before`, which keeps no copy of them, so the
+    // application cannot repaint them. A horizontal shrink must not erase them.
     #[test]
-    fn resize_inline_clears_screen_on_horizontal_shrink() {
+    fn resize_inline_horizontal_shrink_keeps_rows_above_the_viewport() {
+        let mut backend = TestBackend::new(6, 4);
+        backend
+            .set_cursor_position(Position { x: 0, y: 0 })
+            .unwrap();
+        let mut terminal = Terminal::with_options(
+            backend,
+            TerminalOptions {
+                viewport: Viewport::Inline(1),
+            },
+        )
+        .unwrap();
+
+        for text in ["one", "two"] {
+            terminal
+                .insert_before(1, |buf| {
+                    buf.set_string(0, 0, text, crate::style::Style::default());
+                })
+                .unwrap();
+        }
+        terminal
+            .draw(|frame| {
+                let area = frame.area();
+                frame.buffer_mut().set_string(
+                    area.x,
+                    area.y,
+                    "live",
+                    crate::style::Style::default(),
+                );
+            })
+            .unwrap();
+        terminal
+            .backend()
+            .assert_buffer_lines(["one   ", "two   ", "live  ", "      "]);
+
+        // A real backend leaves the cursor just past the last painted cell;
+        // `TestBackend` only moves it on an explicit cursor command, so place it
+        // where the viewport was drawn. `resize` anchors the new origin to it.
+        terminal
+            .backend_mut()
+            .set_cursor_position(Position { x: 4, y: 2 })
+            .unwrap();
+
+        // Narrow the terminal. The backend buffer is left at its original width
+        // so the assertions below read the same columns as the ones above;
+        // `resize_inline_resets_the_buffers_on_horizontal_shrink` does the same.
+        terminal.resize(Rect::new(0, 0, 5, 4)).unwrap();
+
+        assert_eq!(terminal.viewport_area, Rect::new(0, 2, 5, 1));
+        terminal
+            .backend()
+            .assert_buffer_lines(["one   ", "two   ", "      ", "      "]);
+    }
+
+    // This tests for the case where the new width is smaller than the old
+    // width. The internal buffers are reset so the next draw repaints the
+    // viewport into a consistent area.
+    #[test]
+    fn resize_inline_resets_the_buffers_on_horizontal_shrink() {
         let mut backend = TestBackend::with_lines(["0000", "1111"]);
         backend
             .set_cursor_position(Position { x: 0, y: 0 })
--- a/ratatui-core/src/terminal/resize.rs
+++ b/ratatui-core/src/terminal/resize.rs
@@ -99,6 +99,7 @@ mod tests {
     use crate::backend::{Backend, TestBackend};
     use crate::buffer::Buffer;
     use crate::layout::{Position, Rect};
+    use crate::style::Style;
     use crate::terminal::{Terminal, TerminalOptions, Viewport};
 
     #[test]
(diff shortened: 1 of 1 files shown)


cwilson613:
Well done, I just ran into this issue today myself, and was pleasantly surprised to find it's already been addressed!

orhun (approved):
Hey, thanks for the PR!

I think the changes look good, but there is a behavior question that we need to answer here.
- Not clearing the screen for inline apps seems like a breaking change - I updated the PR title to reflect that.
- For me, this new behavior makes more sense, but I'm not sure still. Can you find some real-world inline terminal apps and see how they behave? Or let me ask: why did you need this fix in the first place? (maybe @cwilson613 can also share)
- I'll ask review from @joshka as well - he said he can do some real testing with [betamax](https://github.com/joshka/betamax))

easyinplay:
@orhun Thanks for the review, and the retitle is right. It is a behavior change for inline apps.

## Why I needed it

I'm building an agent CLI with an inline viewport: a live region at the bottom, completed output committed above it with `insert_before`. On Windows Terminal, dragging the window narrower left a staircase of duplicated copies of the live region in scrollback, one per resize event, which is what I filed as #2666. @cwilson613 ran into the same thing independently a few weeks later, so it is not something about my setup.

The mechanism is the part that made me look at `resize` rather than my own code. WT and conhost move erased content into scrollback rather than discarding it, so `ClearType::All` doesn't clean up there, it *copies the live viewport upward*. On terminals that discard instead, the same call takes out the committed rows, and since `insert_before` keeps no copy of them the app has nothing to repaint from. Both outcomes come from erasing rows the inline viewport doesn't own.

## What real-world inline apps do

I searched for `Viewport::Inline` and then narrowed to apps that also use `insert_before`, since those are the ones that own rows above the origin.

| app | inline + `insert_before` | how it reaches `Terminal::resize` |
|---|---|---|
| [openai/codex](https://github.com/openai/codex) | yes | vendored `Terminal` into `codex-rs/tui/src/custom_terminal.rs` |
| [nrwl/nx](https://github.com/nrwl/nx) | yes | `draw_without_autoresize()`, bypasses `autoresize` in inline mode |
| [DioxusLabs/dioxus](https://github.com/DioxusLabs/dioxus) CLI | yes | no resize handling; `draw()` in a loop, so it takes this path |
| [alexpasmantier/television](https://github.com/alexpasmantier/television) | yes | custom backend for cursor queries |

Codex is the interesting one. Their vendored `resize` is:

```rust
/// Requested area will be saved to remain consistent when rendering. This leads to a full clear
/// of the screen.
pub fn resize(&mut self, screen_size: Size) -> io::Result<()> {
    self.last_known_screen_size = screen_size;
    Ok(())
}
```

The doc comment came from ratatui and no longer describes the body: it records the size and does nothing else. `ClearType::All` appears exactly once in that file, in `clear_visible_screen()`, which is the explicit user-facing clear.

Nx bypasses `autoresize` too, though for a different reason: cursor-query races against their `EventStream`, not this. Dioxus binds `Clear(All)` + `Clear(Purge)` to the `c` key and nothing else.

So among the apps that keep committed output above an inline viewport, none of them takes the whole-screen erase as part of resizing. Where they emit it at all, a human asked for it. The two most heavily used ones each grew their own way around `Terminal::resize`, which is roughly the same conclusion this PR reaches, just reached privately.

Dioxus is the one that doesn't work around anything, which makes it the case worth checking under betamax if @joshka gets to it. `dx serve` writes build logs above the viewport and calls `draw()` in a loop, so a horizontal shrink goes straight through `autoresize` into the erase.

## What changes and what doesn't

`clear_viewport()` still runs, so everything from the recomputed origin down is erased exactly as before. Apps that repaint every visible cell each frame (the #2086 / #2355 case) see no difference, because the erase they relied on is the one that stays.

The honest residual: after a shrink, rows above the origin keep whatever wrapping they had, instead of being erased. Those rows are scrollback, and reflowing them is the terminal's job rather than the app's, so I don't think an app can do better with them either way. But that is the visible difference, and it is why the breaking marker is fair.

I have the #2666 reproducer as a standalone crate and can post it, or trim it to whatever shape suits betamax. Happy to run anything else that would help; I have Windows Terminal, conhost, WezTerm and Alacritty here.

orhun:
Thanks for the detailed investigation and the real-world examples! 

Merged with #2731, horizontal shrinking now results in:

- Fullscreen: one full clear
- Inline: clear from the viewport origin downward
- Fixed: one full clear

I think it's fair!

PR #2678, opened 2026-07-28
fix(examples): filter non-press key events

### LINKED ISSUE #2672: Audit remaining crossterm example key-press filtering
## Context

Crossterm 0.29 provides `Event::as_key_press_event()`, and most Ratatui examples now use it following [#1792](https://github.com/ratatui/ratatui/pull/1792). While syncing the website examples to `ratatui-v0.30.2`, I audited the remaining event handling against current `main` at `7986192595d29331df72145483ef78b3698522d1` and found a few gaps worth following up.

## Behavior gaps

- `examples/apps/inline/src/main.rs` forwards every `Event::Key` to the application, then exits whenever the key code is `q`. A release event can therefore quit the example. Filter with `as_key_press_event()` before sending the input event.
- `examples/apps/tracing/src/main.rs` considers any `Event::Key` with code `q` an exit event, including release and repeat events. Use `as_key_press_event()` in `should_exit` so the example responds only to presses.

## Cleanup and consistency

- `examples/apps/volatility-surface/src/main.rs` manually combines `Event::Key` with `KeyEventKind::Press`; this can use `as_key_press_event()` and drop the kind import.
- `examples/apps/user-input/src/main.rs` already calls `as_key_press_event()`, but its editing arm checks `key.kind == KeyEventKind::Press` again and retains an otherwise unreachable `InputMode::Editing` fallback. Remove the redundant check/import/arm.
- Canvas, Custom Widget, and Mouse Drawing currently pass a `KeyEvent` to a handler and call `KeyEvent::is_press()` there. Those are correct and do not have the behavior bug above. They could optionally move filtering to the event boundary for consistency, but that is a style cleanup rather than a fix.
- `examples/apps/demo/src/termina.rs` is intentionally excluded: it uses Termina's separate event API, not crossterm.

## Suggested outcome

- Make `inline` and `tracing` ignore release and repeat events.
- Use the crossterm 0.29 helper in the direct manual-match cases.
- Remove the redundant user-input kind check.
- Keep the mixed mouse/key handlers unchanged unless the resulting code is clearer.

Found while validating [ratatui-website#1187](https://github.com/ratatui/ratatui-website/pull/1187).

subotac:
## Summary

- filter Crossterm events with `as_key_press_event()` in the remaining examples
- refactor inline input handling to propagate terminal and channel errors instead of panicking
- preserve inline resize handling and remove redundant user-input event-kind matching
- add regression tests for press, repeat, release, and resize event handling

Closes #2672

## Testing

- `RUSTFLAGS='--cfg getrandom_backend="windows_legacy"' cargo test --locked -p inline --bin inline`
- `cargo test --locked -p tracing@0.0.0 --bin tracing`
- `cargo clippy --locked -p inline -p tracing@0.0.0 -p volatility-surface -p user-input --all-targets -- -D warnings`
- `cargo xtask typos`
- `git diff --check`

An additional `cargo test --locked --workspace --exclude ratatui-termion` attempt stopped before
project tests because the local Windows GNU toolchain's incomplete `dlltool`/assembler setup could
not compile external `getrandom`/`windows-sys` dependencies.

### DIFF
--- a/examples/apps/inline/src/main.rs
+++ b/examples/apps/inline/src/main.rs
@@ -96,27 +96,32 @@ struct Worker {
 }
 
 fn input_handling(tx: mpsc::Sender<Event>) {
+    thread::spawn(move || handle_input_events(&tx));
+}
+
+fn handle_input_events(tx: &mpsc::Sender<Event>) -> Result<()> {
     let tick_rate = Duration::from_millis(200);
-    thread::spawn(move || {
-        let mut last_tick = Instant::now();
-        loop {
-            // poll for tick rate duration, if no events, sent tick event.
-            let timeout = tick_rate.saturating_sub(last_tick.elapsed());
-            if event::poll(timeout).unwrap() {
-                let event = event::read().unwrap();
-                if let Some(key) = event.as_key_press_event() {
-                    tx.send(Event::Input(key)).unwrap();
-                }
-                if matches!(event, event::Event::Resize(_, _)) {
-                    tx.send(Event::Resize).unwrap();
-                }
-            }
-            if last_tick.elapsed() >= tick_rate {
-                tx.send(Event::Tick).unwrap();
-                last_tick = Instant::now();
-            }
+    let mut last_tick = Instant::now();
+    loop {
+        // poll for tick rate duration, if no events, sent tick event.
+        let timeout = tick_rate.saturating_sub(last_tick.elapsed());
+        if event::poll(timeout)?
+            && let Some(event) = map_input_event(&event::read()?)
+        {
+            tx.send(event)?;
         }
-    });
+        if last_tick.elapsed() >= tick_rate {
+            tx.send(Event::Tick)?;
+            last_tick = Instant::now();
+        }
+    }
+}
+
+fn map_input_event(event: &event::Event) -> Option<Event> {
+    event
+        .as_key_press_event()
+        .map(Event::Input)
+        .or_else(|| matches!(event, event::Event::Resize(_, _)).then_some(Event::Resize))
 }
 
 #[expect(clippy::cast_precision_loss, clippy::needless_pass_by_value)]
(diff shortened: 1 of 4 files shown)


orhun (approved):

examples/apps/inline/src/main.rs
```diff
@@ -103,10 +103,12 @@ fn input_handling(tx: mpsc::Sender<Event>) {
...
-                match event::read().unwrap() {
-                    event::Event::Key(key) => tx.send(Event::Input(key)).unwrap(),
-                    event::Event::Resize(_, _) => tx.send(Event::Resize).unwrap(),
-                    _ => {}
+                let event = event::read().unwrap();
+                if let Some(key) = event.as_key_press_event() {
+                    tx.send(Event::Input(key)).unwrap();
+                }
+                if matches!(event, event::Event::Resize(_, _)) {
+                    tx.send(Event::Resize).unwrap();
```
joshka: don't use unwrap in code this was there before, but should be dropped to avoid panics

examples/apps/tracing/src/main.rs
```diff
@@ -50,9 +50,11 @@ fn main() -> Result<()> {
...
 
 fn should_exit(events: &[Event]) -> bool {
-    events
-        .iter()
-        .any(|event| matches!(event, Event::Key(key) if key.code == KeyCode::Char('q')))
+    events.iter().any(|event| {
+        event
+            .as_key_press_event()
+            .is_some_and(|key| key.code == KeyCode::Char('q'))
+    })
```
joshka: This might look better as a .map(Event::as_key_press_event).any(...)

joshka (changes requested):

examples/apps/inline/src/main.rs
```diff
@@ -103,14 +103,20 @@ fn input_handling(tx: mpsc::Sender<Event>) {
             // poll for tick rate duration, if no events, sent tick event.
             let timeout = tick_rate.saturating_sub(last_tick.elapsed());
             if event::poll(timeout).unwrap() {
-                match event::read().unwrap() {
-                    event::Event::Key(key) => tx.send(Event::Input(key)).unwrap(),
-                    event::Event::Resize(_, _) => tx.send(Event::Resize).unwrap(),
-                    _ => {}
+                let event = event::read().unwrap();
```
joshka: still two unwraps here - I think if we're modifying this it's important to make it flow well. maybe that means extracting bits to functions, or thinking about how this neatly is expressed. the one benefit of the current code is it's succinct. This change makes it less so, but in doing so highlighted some existing cruftiness. That makes it worth changing to something that pays heed to that and cleans it up more generally.

examples/apps/tracing/src/main.rs
```diff
@@ -1,102 +1,125 @@
-/// A simple example demonstrating how to use the [tracing] with Ratatui to log to a file.
```
joshka, line 1: this diff looks oddly large on github - any idea what's going on here?

orhun (approved):

Issue #2526, opened 2026-05-10
Clear widget doesn't handle double-width glyphs straddling the rect's boundaries


Booyaka101:
## Description

`Clear`'s render iterates cells *inside* the rect and calls `cell.reset()`. That leaves two cases at the rect's boundaries unhandled when the back layer contains double-width glyphs (CJK, fullwidth punctuation, emoji, etc.):

**Left edge.** A double-width glyph at `area.x - 1` keeps its full two-cell visual width. Its right half spills into `area.x`, sits underneath whatever the caller draws on top of `Clear`, and renders as garbled overlap with the overlay's left border or title.

```
back layer:    a b c 字 字       ← `字` occupies cols 3 and 4
overlay area:           ^^^^^    ← starts at col 4
post-Clear:    a b c 字 _ _      ← back-layer 字 at col 3 still claims 2 visual cells
visual:        a b c 字▒...      ← `字`'s right half overlaps the overlay's left border
```

**Right edge.** A double-width glyph whose left half sits at `area.right() - 1` (inside the area, gets cleared) leaves its continuation at `area.right()` (outside the area) as the empty string ratatui's buffer model uses for continuations. The terminal then either renders stale content from a previous frame in that column or paints the neighbour glyph half-erased.

```
back layer:    a b c 字 d        ← `字` occupies cols 3 and 4
overlay area:  ^^^^^^^           ← covers cols 0..4
post-Clear:    _ _ _ _ ϵ d       ← cell 4 is the orphaned continuation (empty string)
visual:        ___ ?d            ← terminal shows stale/half-erased glyph at col 4
```

## To Reproduce

Minimal repro using `ratatui_core::buffer::Buffer` directly (no terminal needed):

```rust
use ratatui::buffer::Buffer;
use ratatui::layout::Rect;
use ratatui::style::Style;
use ratatui::widgets::{Clear, Widget};

fn main() {
    // Left-edge case: 字 at col 3 spills into col 4.
    let mut buf = Buffer::empty(Rect::new(0, 0, 6, 1));
    buf.set_string(0, 0, "abc字", Style::default());
    Clear.render(Rect::new(4, 0, 2, 1), &mut buf);
    let row: Vec<&str> = (0..6).map(|x| buf[(x, 0)].symbol()).collect();
    // After Clear, cell 3 still contains "字" (visual width 2).
    // Anything the caller renders into [4..6) overlaps the right half of 字 at the screen.
    assert_eq!(row[3], "字");

[...]

Issue #1332, opened 2024-08-20
insert_before and emoji width issue


Kongga666:
## Description

Hi, there is a difference in behavior between `draw` and `insert_before`, for some emoji `insert_before` will left a space after the emoji. 
I've had this issue since at least ratatui 0.26 but i switched to other languages and didn't took time to report it.
I checked if it was still present in the current version (0.28) and it is.
## To Reproduce

Sorry for bad code, it's been a while since i didn't code in rust...
```rust
use crossterm::terminal::{disable_raw_mode, enable_raw_mode};
use ratatui::backend::CrosstermBackend;
use ratatui::prelude::{Color, Span, Style, Widget};
use ratatui::Terminal;
use std::io::stdout;

fn main() -> std::io::Result<()> {
    enable_raw_mode()?;
    let options = ratatui::TerminalOptions {
        viewport: ratatui::Viewport::Inline(1),
    };
    let mut terminal = Terminal::with_options(CrosstermBackend::new(stdout()), options)
        .expect("Terminal initialization error");
    terminal.clear()?;

    terminal
        .insert_before(1, |buf| {
            generate_exemple_text("📂insert_before_function📂test").render(buf.area, buf)
        })
        .expect("error while inserting before");
    disable_raw_mode()?;

    terminal
        .draw(|frame| {
            generate_exemple_text("📂draw_function📂test").render(frame.area(), frame.buffer_mut())
        })
        .expect("error while drawing");

    Ok(())
}

fn generate_exemple_text(text: &str) -> Span {
    Span::styled(text, Style::default().bg(Color::White))
}
```

## Expected behavior

print "📂insert_before_function📂test" and "📂draw_function📂test" on white background without any space character in both of those text.
## Screenshots

![Screenshot_20240820_171917](https://github.com/user-attachments/assets/50b52253-8e3f-4b64-8d4b-42133b66ff4d)

We can clearly see the spaces after the emoji when calling `insert_before`

## Environment

- OS: Manjaro
- Terminal Emulator: konsole
- Font: Hack Nerd Font Mono Regualr 11
- Crate version: 0.28.0
- Backend: Crossterm (0.28.1)

PR #2686, opened 2026-08-04
fix(core): prevent rendering artifacts around VS16 graphemes

### LINKED ISSUE #2357: Rendering bug with emojis in list

sinelaw:
Fixes #2357.

## Problem

Terminals disagree on the width of a grapheme cluster carrying VS16 (`U+FE0F`): most draw it in two columns, some in one. #2063 handled the one-column case by emitting the cluster's reserved trailing cell right after the cluster, so that column gets cleared.

On a two-column terminal that trailing cell is adjacent to the cluster, so the backend emits no cursor move and the write lands one column too far right. Every later write on the row inherits the offset, and the columns the drifted writes step over keep the previous frame's characters.

Dropping the trailing write is not an option — it is what #2063 needs. Both terminal behaviours exist.

## Change

Emit the reserved columns before the cluster. A two-column draw repaints over them; a one-column draw leaves them cleared. Drawing the cluster last also keeps it non-adjacent to the preceding write, forcing a cursor move, so a width disagreement cannot shift the rest of the row.

30 lines of production change: one `deferred` field on `TrailingState`, a block to yield it, and the two construction sites. No existing test changes.

## Tests

A model backend replays a diff with real cursor semantics and compares the resulting screen against the buffer, parameterised over both terminal behaviours (VS16 advances two columns / one). Every scenario must pass under both.

| scenario | issue |
| --- | --- |
| emoji replacing ASCII | #2357 |
| keycaps replacing ASCII | #2357 |
| row shifted one column | #2357 |
| mixed widths (`❤️`, `漢`, `😀`, `＋`) | #2357 |
| cluster replacing narrower content | #2063 |

All five fail without this change. They also all fail if the VS16 case is deleted outright — the last one in the one-column model, which is the #2063 regression.

## Related work: #2517

#2517 fixes the same drift from the backend side, replacing the `last_pos` adjacency check with a `next_pos` computed from `cell_width()`. It does not reference #2357, and on its own it does not fix it — it changes the artifact.

With #2517's backend and `main`'s diff, the trailing cell still comes after the cluster, so the new cursor logic now emits an explicit move onto the cluster's *second* column and prints a space there, destroying the glyph. Tested in tmux 3.5a on the fixture below, a heart row renders as `❤️        ❤️` — worse than today.

With #2517's backend and this change, the reserved column is written before the cluster, the cluster gets its own cursor move, and the row is correct. The two changes are complementary, and this one is what makes #2517 safe on two-column terminals. They touch disjoint files (`ratatui-core/src/buffer/diff.rs` vs the three backends) and apply cleanly together.

## Manual verification

46x24 tmux pane, 120 lines alternating an emoji row and `int valueN = compute(N);`, scrolled by mouse wheel.

- tmux 3.5a: rows drift on `main`; with this change live output equals both a forced repaint and a resize round-trip, at 1/3/7/13/25 notches — for `❤️`, `1️⃣`, `🚀️` (the emoji from #2357) and a mixed-glyph fixture.
- tmux 3.4 still shows artifacts, independently of ratatui: printing `❤️` over existing text there leaves the column to its right stale and clears the one after, while `漢` is handled correctly.

## Out of scope

#2526 and #1332 do not go through `BufferDiff`; both have open PRs (#2628, #2527).

## AI assistance

Written with Claude Code. Every line reviewed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01WESmXtdFXigScHAenX2ZvX

### DIFF
--- a/.github/workflows/check-semver.yml
+++ b/.github/workflows/check-semver.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout the repository
-        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           persist-credentials: false
       - name: Check semver
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,15 +29,15 @@ jobs:
     name: Check Formatting
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           persist-credentials: false
-      - uses: dtolnay/rust-toolchain@2c7215f132e9ebf062739d9130488b56d53c060c # master
+      - uses: dtolnay/rust-toolchain@6c977a6ca4077a0ceb28ffbe03f59d46e9ac8772 # master
         with:
           toolchain: nightly
           components: rustfmt
-      - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2
-      - uses: tombi-toml/setup-tombi@fa8fb98ef9af2d3b9da565edc15935d523bca7e9 # v1.2.0
+      - uses: Swatinem/rust-cache@6323deb102c322ba6fcbdcafc7e3dddab59af2b6 # v2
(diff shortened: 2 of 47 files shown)


orhun (approved):
thanks!

Issue #1015, opened 2024-04-01
Add better text style tests


joshka:
## Problem
The addition of `Text::style` and `Line::style` broke some assumptions about how things should work in `Paragraph`.

We need a comprehensive set of tests that shows how the following styles interact:
- Any existing background style that is rendered to the buffer beforehand
- Paragraph::style
- Block::style
- Text::style
- Line::style
- Span::style

## Solution
Add an integration test for this under the tests/ folder that tests the combinations.
Check whether the unit tests cover this from the perspective of the individual types and make the tests consistent and simple.
Ensure the API docs indicate the expected interactions neatly (this overrides that)
Ensure that any illogical overrides / bugs are fixed.

## Additional context
- https://github.com/ratatui-org/ratatui/pull/956#issuecomment-2017020445
- https://github.com/ratatui-org/ratatui/pull/992

akshay1992kalbhor:
Can I take this one?

joshka:
> Can I take this one?

Sure!

joshka:
I added some detail to the solution suggestions. I want to re-iterate that the primary goals on this issue are to ensure that we are rendering these things consistently and correctly and to make that clear with both docs and tests.

supreetsingh10:
Hi @joshka 

I have gone through the bugs and links you have mentioned. I have written a test. What do you think of it? 

All the feedback is welcome. 

`
#[test]
fn test_paragraph_block_style() -> std::io::Result<()> {
    let block_styled = Block::new().style(Style::default().fg(Color::Green));
    let paragraph = Paragraph::new(Text::styled("Hello Tests", Style::default().fg(Color::Red)))
        .block(block_styled);

    let mut terminal = Terminal::new(TestBackend::new(14, 1))?;
    let area = Rect::new(0, 0, 14, 1);

    terminal.draw(|f| {
        f.render_widget(paragraph, area);
    })?;

    let mut expected_buff = Buffer::with_lines(Line::from("Hello Tests"));
    expected_buff.set_style(area, Style::default().fg(Color::Red));

    terminal.backend().assert_buffer(&expected_buff);

    Ok(())
}
`

PR #2693, opened 2026-08-06
test: cover and document Paragraph/Text/Line/Span style interactions

LargeModGames:
## What

Addresses #1015, covering the integration-test and API-docs items from its solution
list (the first and third).

Adds an integration test that characterizes how styles interact when a `Paragraph`
is rendered, and clarifies the API docs on `Span::style`, `Line::style`,
`Text::style`, and `Paragraph::style` to state those interactions.

## Tests

New `ratatui/tests/text_style.rs`, asserting against the rendered `Buffer`:

- a background already rendered to the buffer is preserved under unstyled text
- `Paragraph::style` fills the whole area, and overrides a prerendered foreground
  while inheriting its background
- the precedence chain on text cells is `Span` > `Line` > `Text` > `Paragraph`,
  with each layer overriding only the fields it sets and inheriting the rest
- `Block::style` covers the border while `Paragraph::style` covers the whole inside
- all six layers combined at once

## Docs

Each `style` method now states, in fieldwise terms, that its style is patched on top
of the neighbouring layers, with fields set by later layers taking precedence (order
`Text` -> `Line` -> `Span`). `Block::style` was left unchanged; the region-dependent
Paragraph/Block precedence is deferred to the fourth solution item (see the question
below).

## Deliberately not included

- The second solution item (per-type unit-test submodules): planned as a small
  follow-up PR to keep this one focused.
- The fourth solution item (reconciling illogical overrides): no behavior is changed
  here. See the question below.

## Question for maintainers (relates to the fourth solution item)

While writing the tests I noticed `Paragraph::style` is applied to the whole area,
then applied again to the inner text area *after* the block is drawn (`render` /
`render_paragraph`). The effect is that on **border** cells `Block::style` wins, but
on **inner** cells `Paragraph::style` wins.

My tentative read is that this is defensible: the inner text area being governed by
`Paragraph::style` and the border by `Block::style` is intuitive. But since it falls
out of `Paragraph::style` being applied twice rather than by explicit design, I
wanted to flag it. The test
`block_style_covers_the_border_but_paragraph_style_covers_the_whole_inside`
characterizes the current behavior without claiming it is correct, and I did not
document it as a guarantee. If a different behavior is intended I'm happy to adjust
the test or open a follow-up.

## AI disclosure

This PR was prepared with the assistance of Claude Code. I reviewed every line,
confirmed the tests pass and the docs build cleanly, and checked the style-cascade
claims against the actual render implementation.

### DIFF
--- a/ratatui-core/src/text/line.rs
+++ b/ratatui-core/src/text/line.rs
@@ -327,6 +327,9 @@ impl<'a> Line<'a> {
     /// only by the style of each [`Span`] contained in the line. For this reason, this field may
     /// not be supported by all widgets (outside of the `ratatui` crate itself).
     ///
+    /// When rendered as part of a [`Text`], this style is patched on top of the text's style. Each
+    /// [`Span`] style is then patched on top, so fields set by a span take precedence.
+    ///
     /// `style` accepts any type that is convertible to [`Style`] (e.g. [`Style`], [`Color`], or
     /// your own type that implements [`Into<Style>`]).
     ///
--- a/ratatui-core/src/text/span.rs
+++ b/ratatui-core/src/text/span.rs
@@ -199,6 +199,9 @@ impl<'a> Span<'a> {
     /// In contrast to [`Span::patch_style`], this method replaces the style of the span instead of
     /// patching it.
     ///
+    /// When rendered inside a [`Line`] or [`Text`], this style is patched on top of those styles,
+    /// so any field it sets takes precedence over them.
+    ///
     /// `style` accepts any type that is convertible to [`Style`] (e.g. [`Style`], [`Color`], or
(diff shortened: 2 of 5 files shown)


orhun (approved):

joshka:
To answer your question directly: yes, I think this region-dependent behavior is defensible and should remain unchanged in this PR. `Block` style winning on border cells while `Paragraph` style wins inside the block is intuitive: the Block owns the border and decoration, while the Paragraph owns its content area.

Please keep the test as a characterization of the current behavior, but don’t document it as a general precedence guarantee yet—the fact that Paragraph is applied twice is an implementation detail. We can address issue #1015’s fourth item separately if we decide to formalize or change this rule.

PR #2694, opened 2026-08-06
fix(core): save and restore cursor position

### LINKED ISSUE #2691: `terminal.clear()` breaks the Termion backend in Ratatui 0.30
## Description

I was a happy user of Ratatui 0.29 and made an app that used the Termion backend.

Ratatui 0.30 decided, for some reason, to keep the cursor position when calling `terminal.clear()`.

Now, if I try to call `terminal.clear()` from my app, my app crashes with the error:

```
Error: Cursor position detection timed out.
```

I see that in `ratatui-core/src/terminal/buffers.rs` there's a function called `clear_viewport`, that, if I understand correctly, was created because resizing caused a similar crash, that would also make my app not crash, but unfortunately it's not a public function.

Now: What's the reason to keep the cursor position when calling `terminal.clear()` in the first place?
Ratatui 0.29 worked just fine without all the cursor position stuff.

## To Reproduce

The fastest way to reproduce is to add a `terminal.clear()` call in the `examples/apps/demo/src/termion.rs` file at line 45, just before the `terminal.draw()` call.

Only if contributing does not require jumping through too many hurdles (eg. signing a CLA, signed commits, or stuff like that).

- [X] I am willing to open a PR for this bug.

## Environment

- OS: Linux
- Terminal Emulator: GNOME Terminal
- Font: Liberation Mono
- Crate version: 0.30.2
- Backend: termion

joshka:
## Summary

- add default `Backend` methods for saving and restoring the cursor position
- use save/restore during fixed and inline viewport clears, with get/set as the fallback
- implement terminal save/restore commands for Crossterm, Termion, and Termina
- keep Termwiz on its local buffered-surface fallback
- document the single saved-cursor slot and the unlikely trait method ambiguity

## Problem

`Terminal::clear()` queried the cursor position before clearing partial viewports. Termion implements that query by requesting a terminal response and reading input, which can race an application's input reader and time out.

The new save/restore path avoids reading terminal input for the built-in backends that support terminal cursor save commands. Backends that use the default `Ok(false)` implementation retain the existing get/set fallback.

## Testing

- `cargo fmt --check`
- `cargo test -p ratatui-core -p ratatui-crossterm -p ratatui-termion -p ratatui-termina -p ratatui-termwiz`
- `cargo clippy --workspace --all-targets -- -D warnings`
- `markdownlint-cli2 BREAKING-CHANGES.md`

Fixes #2691.

### DIFF
--- a/BREAKING-CHANGES.md
+++ b/BREAKING-CHANGES.md
@@ -10,6 +10,8 @@ GitHub with a [breaking change] label.
 
 This is a quick summary of the sections below:
 
+- [v0.31.0](#v0310)
+  - `Backend` adds cursor save and restore methods
 - [v0.30.1](#v0301)
   - Adding `AsRef` impls for widgets may affect type inference in rare cases
   - MSRV is now 1.88.0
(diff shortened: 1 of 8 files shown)


orhun (approved):
LGTM

PR #2697, opened 2026-08-08
fix(buffer): treat multi-grapheme cell symbols as single-width

### LINKED ISSUE #2685: Buffer::diff drops all remaining cells after a cell whose symbol is longer than one column (regression in ratatui-core 0.1.1)
## Description

Upgrading a CLI from using ratatui-core 0.1.0 to 0.1.2 resulted in images not being correctly drawn, at least with kitty.

Claude help me create this reproducer:

## To Reproduce

```rust
//! Fills the screen with STALE, then draws text plus an image that should replace it.
//!
//! ratatui-core 0.1.0: the text and image appear.
//! ratatui-core 0.1.2: the screen still says STALE, because the image cell's symbol measures
//! ~1300 columns wide and the diff cursor skips everything after it.

use std::{io, thread::sleep, time::Duration};

use image::{DynamicImage, Rgba, RgbaImage};
use ratatui::layout::{Constraint, Layout};
use ratatui::text::Line;
use ratatui::widgets::Paragraph;
use ratatui_image::picker::{Picker, ProtocolType};
use ratatui_image::{Resize, StatefulImage};

fn main() -> io::Result<()> {
    // Force kitty instead of probing, so this does not depend on which terminal runs it.
    let mut picker = Picker::halfblocks();
    picker.set_protocol_type(ProtocolType::Kitty);
    let img = RgbaImage::from_fn(200, 200, |x, y| Rgba([x as u8, y as u8, 200, 255]));
    let mut proto = picker.new_resize_protocol(DynamicImage::ImageRgba8(img));

    terminal.draw(|f| {
        let stale = "STALE ".repeat(f.area().width as usize);
        let lines = vec![Line::from(stale); f.area().height as usize];
        f.render_widget(Paragraph::new(lines), f.area());
    })?;
    sleep(Duration::from_secs(1));

    terminal.draw(|f| {
        let [left, right] =
            Layout::horizontal([Constraint::Percentage(60), Constraint::Percentage(40)])
                .areas(f.area());
        let lines: Vec<Line> = (0..f.area().height)
            .map(|i| Line::from(format!("line {i:02} should be visible")))
            .collect();
        f.render_widget(Paragraph::new(lines), left);
        f.render_stateful_widget(
            StatefulImage::default().resize(Resize::Scale(None)),
            right,
            &mut proto,
        );
    })?;
    sleep(Duration::from_secs(4));

    ratatui::restore();
    Ok(())
}
```

```toml
[package]
name = "imgtest"
version = "0.0.0"
edition = "2021"

[dependencies]
ratatui = "=0.30.0"
ratatui-core = "=0.1.2" # flip to 0.1.0 and the image appears
ratatui-image = { version = "=10.0.8", default-features = false, features = ["crossterm"] }
image = { version = "0.25", default-features = false }
```

## Screenshots

With 0.1.2:

Image

With 0.1.0:

Image

- [x] I am not able to work on a fix right now.

## Environment

- OS: debian
- Terminal Emulator: kitty
- Font: DejaVuSansMono (I think)
- Crate version:
- Backend: not sure?

## Additional context

Claude is saying that ratatui-image might need `CellDiffOption::ForcedWidth` on the payload cell?

Claude also provided this:

```rust
use ratatui::buffer::{Buffer, Cell};
use ratatui::layout::Rect;

fn main() {
    let area = Rect::new(0, 0, 20, 5); // 100 cells
    let mut prev = Buffer::filled(area, Cell::new("X"));
    let mut next = Buffer::filled(area, Cell::new("Y")); // the other 99 cells all differ

    // One cell's symbol is a long string, though it still occupies a single terminal cell.
    // ratatui-image does this: it stores a whole image escape sequence in one cell's symbol.
    let long = "a".repeat(200);
    prev[(0, 0)].set_symbol(&long);
    next[(0, 0)].set_symbol(&long); // identical in both, so it should not be redrawn

    println!("{} of 99 changed cells emitted", prev.diff(&next).len());
}
```

```
[~/code/edge/ratatui-diff-repro] 9s $ cargo run
0 of 99 changed cells emitted
$ vim Cargo.toml # change to 0.1.0 from 0.1.2
$ cargo run
98 of 99 changed cells emitted
``

### PR DESCRIPTION
**Pull request by teddytennant**
Rewritten after @orhun's review. The first version was wrong in both directions: it broke wide glyphs, and it did not actually fix the reported case. Details below.

## What is going on

A `Cell` holds a single grapheme cluster, so the unicode width of its symbol is how many columns it covers. `BufferDiff` relies on that to step over the columns a wide glyph hides.

Image protocols break the assumption. `ratatui-image` keeps a whole kitty escape sequence in one `Cell::symbol` (and marks the rest of the row `Skip`), so `cell_width()` reports hundreds of columns for a cell that covers one. The diff then advances past the rest of the buffer and the frame keeps whatever was on screen before.

Before #2416 this was harmless: `to_skip` was reassigned on every iteration, so a wide cell only ever suppressed the single cell after it. #2416 turned it into a real multi-cell advance, which is where the regression comes from.

## The fix

Report a symbol that is not a single grapheme cluster as one column wide. `CellDiffOption::ForcedWidth` still wins, so a caller that really does write wider content into one cell can say so.

Real wide glyphs are unaffected: they are one cluster, keep their unicode width, and keep hiding their trailing cells.

## On the previous version

The old approach (drop the advance on equal cells) was wrong twice over:

- It broke unchanged wide glyphs, which is the regression you caught. Your test is in as `identical_wide_glyph_hides_trailing_cell_changes`.
- It did not fix #2685. The repro in the issue draws text, then the image, so the payload cell *changed* and took the unequal branch, which still advanced by ~1300. I confirmed that on the old branch: the changed-payload case emitted 11 of 100 cells.

## Verified

- `cargo test --workspace --all-features`, `cargo clippy --workspace --all-targets --all-features`, `cargo fmt --check`
- `ratatui-core/src/buffer/diff.rs` is test-only in this diff; the behaviour change is entirely in `CellWidth for Cell`
- new tests: changed payload cell, unchanged payload cell, your wide glyph case, plus `cell_width` unit tests for a wide grapheme, an escape payload, and `ForcedWidth`

Fixes #2685

### DIFF
--- a/ratatui-core/src/buffer/cell.rs
+++ b/ratatui-core/src/buffer/cell.rs
@@ -1,6 +1,7 @@
 use core::num::NonZeroU16;
 
 use compact_str::CompactString;
+use unicode_segmentation::UnicodeSegmentation;
 
 use crate::buffer::cell_width::CellWidth;
 use crate::style::{Color, Modifier, Style};
@@ -309,17 +310,63 @@ impl From<char> for Cell {
 impl CellWidth for Cell {
     /// Returns [`CellDiffOption::ForcedWidth`] when set, otherwise computes the width from the
     /// cell's symbol.
+    ///
+    /// A cell holds a single grapheme cluster, so the unicode width of its symbol is the number
+    /// of columns it takes on screen. Low-level callers can store anything in a symbol though,
+    /// and image protocols store a whole escape sequence there while still occupying a single
+    /// column. Such a symbol has no meaningful unicode width, so it counts as one column.
+    /// Use [`CellDiffOption::ForcedWidth`] to declare the real width of these cells.
     fn cell_width(&self) -> u16 {
-        match self.diff_option {
-            CellDiffOption::ForcedWidth(w) => w.get(),
-            _ => self.symbol().cell_width(),
+        if let CellDiffOption::ForcedWidth(width) = self.diff_option {
+            return width.get();
+        }
+        let symbol = self.symbol();
+        let width = symbol.cell_width();
+        if width > 1 && !is_single_grapheme(symbol) {
+            1
+        } else {
+            width
         }
     }
+}
+
+/// Returns `true` if `symbol` is a single grapheme cluster.
+///
+/// Stops after the second cluster, so this stays cheap for the long escape payloads that make
+/// the check necessary in the first place.
+fn is_single_grapheme(symbol: &str) -> bool {
+    let mut graphemes = symbol.graphemes(true);
+    graphemes.next().is_some() && graphemes.next().is_none()
 }
 
 #[cfg(test)]
 mod tests {
     use super::*;
+
+    #[test]
+    fn cell_width_of_wide_grapheme() {
+        assert_eq!(Cell::new("あ").cell_width(), 2);
+        assert_eq!(Cell::new("a").cell_width(), 1);
+    }
+
+    /// Image protocols keep a whole escape sequence in one symbol. Its unicode width says
+    /// hundreds of columns, but the cell covers one, so it must not be reported as wide.
+    #[test]
+    fn cell_width_of_escape_sequence_symbol() {
+        let payload = "\u{1b}[s".repeat(400);
+        let mut cell = Cell::EMPTY;
+        cell.set_symbol(&payload);
+        assert_eq!(cell.cell_width(), 1);
+    }
+
+    #[test]
+    fn cell_width_of_forced_width_symbol() {
+        let payload = "\u{1b}[s".repeat(400);
+        let mut cell = Cell::EMPTY;
+        cell.set_symbol(&payload);
+        cell.set_diff_option(CellDiffOption::ForcedWidth(NonZeroU16::new(24).unwrap()));
+        assert_eq!(cell.cell_width(), 24);
+    }
 
     #[test]
     #[allow(deprecated)]
--- a/ratatui-core/src/buffer/diff.rs
+++ b/ratatui-core/src/buffer/diff.rs
@@ -232,6 +232,58 @@ mod tests {
         assert!(diff.is_empty());
     }
 
+    /// Regression for <https://github.com/ratatui/ratatui/issues/2685>: an image protocol cell
+    /// keeps its whole escape sequence in one symbol, which measures hundreds of columns wide
+    /// while covering a single column. Advancing the cursor by that width dropped every later
+    /// cell in the buffer, so the frame kept whatever was on screen before.
+    #[test]
+    fn long_symbol_does_not_hide_later_cells() {
+        let area = Rect::new(0, 0, 20, 5);
+        let prev = Buffer::filled(area, Cell::new("S"));
+        let mut next = Buffer::filled(area, Cell::new("L"));
+        next[(10, 0)].set_symbol(&"\u{1b}[s".repeat(400));
+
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by teddytennant** [comment]
Clippy was failing on `doc_markdown` for the bare issue URL in the new test's doc comment. Wrapped it in angle brackets to match the convention used elsewhere in the crate and force-pushed. CI is green now.

**Comment by teddytennant** [comment]
Anything I can do to help get this reviewed?

**Comment by orhun** [comment]
Hey, thanks for the PR!
I actually started reviewing this last night but got too tired at some point 😅 
So I caught another regression caused by removing the skip for all identical cells.

> A real wide glyph such as `＋` occupies two terminal columns. When it is unchanged, its hidden second cell must not be emitted separately, because writing to that cell can overwrite the right half of the glyph.

Here is an example unit test that fails:

  ```rust
  #[test]
  fn identical_wide_glyph_hides_trailing_cell_changes() {
      use crate::style::Style;

      let area = Rect::new(0, 0, 3, 1);
      let mut prev = Buffer::empty(area);
      prev.set_string(0, 0, "＋", Style::default());

      let mut next = prev.clone();
      // Column 1 is occupied by the right half of `＋`.
      next[(1, 0)].set_bg(Color::Red);

      let diff: Vec<_> = BufferDiff::new(&prev, &next).collect();

      assert!(
          diff.is_empty(),
          "the hidden trailing cell of an unchanged wide glyph must not be emitted; got {diff:?}"
      );
  }
```

So the image payload needs to be treated as one logical buffer cell, but wide glyphs must continue skipping their hidden trailing cells.

Can you update the implementation to handle this case too? I would appreciate if you could add the unit test above as well.

teddytennant:
Thanks, that was a real regression and it sent me back to the start. Your test is in as `identical_wide_glyph_hides_trailing_cell_changes`.

While adding it I found the original fix also did not fix #2685. The repro in the issue draws text first and the image second, so the payload cell *changed* and took the unequal branch, which still advanced by ~1300 columns. On the old branch that case emitted 11 of 100 cells.

So I rewrote it. The root cause is `cell_width()`: a `Cell` holds one grapheme cluster, so its symbol's unicode width is how many columns it covers, but `ratatui-image` stores a whole kitty escape sequence in one symbol. Now a symbol that is not a single grapheme cluster reports one column, and `CellDiffOption::ForcedWidth` still overrides. `diff.rs` is test-only in the new diff.

Worth noting this was harmless before #2416: `to_skip` was reassigned every iteration, so a wide cell only suppressed the one cell after it.

teddytennant:
The `Check Clippy (beta)` failure is not from this PR. Beta picked up `clippy::assert_is_empty`, and it fires on existing code: `diff.rs:225` and `diff.rs:232` are unchanged on main, and `barchart.rs:762-763` is not touched here at all. Main was last green before the toolchain moved.

Happy to send that as a separate PR like #2567 if you want it, or leave it to you.

joshka:
> The Check Clippy (beta) failure is not from this PR. Beta picked up clippy::assert_is_empty, and it fires on existing code: diff.rs:225 and diff.rs:232 are unchanged on main, and barchart.rs:762-763 is not touched here at all. Main was last green before the toolchain moved.

Ha - that's actually funny, I added one of the new lints to clippy that was triggered here.
https://github.com/ratatui/ratatui/pull/2728 fixes it.

orhun (approved):
LGTM, thanks!

PR #2713, opened 2026-08-17
Fixes #2709: Add marker support for Sparkline

### LINKED ISSUE #2709: Add Marker support for Sparkline
## Problem

The Sparkline widget does not have an option for a custom Marker.
This is very frustrating, because often Braille is used for such charts, which is not available.

TheRodzz:
Fixes #2709

This PR adds a `marker` method to `Sparkline` allowing it to use different symbols from `Marker`. Supported markers include `Braille`, `Dot`, `Block`, `Bar`, `HalfBlock`, `Quadrant`, `Sextant`, and `Octant`.


orhun:
Hello!

Thanks a lot for the PR!

Although this came first, I think #2725 is semantically more correct. I think we'll work on top if.

PR #2714, opened 2026-08-17
refactor(test): group style tests into separate modules

LargeModGames:
Part of #1015: the second checkbox (per-type `style` test modules). Follow-up to #2693, which covered the integration test and the docs.

Groups each type's style-related unit tests into a `style` submodule of its `tests` module, and adds the missing per-type precedence tests.

**What moved (bodies unchanged):**
- `Span`: `set_style` (renamed `style` to match `Line`), `reset_style`, `patch_style`, `stylize`
- `Line`: `style`, `patch_style`, `reset_style`, `stylize`
- `Text`: `patch_style`, `reset_style`, `stylize`
- `Paragraph`: `test_render_line_styled`, `test_render_line_spans_styled`, `test_render_paragraph_with_styled_text`, `paragraph_block_text_style` (renamed to drop the `test_`/`paragraph_` prefixes now that they sit under `paragraph::tests::style`)

**What's new:** each type renders itself directly (not via `Paragraph`) and pins how its style combines with a pre-rendered buffer style and with its children's styles. Lower layer sets `red().on_green()`, upper layer sets `blue()`, expected is `blue().on_green()`, so a dropped or misordered layer fails.
- `Span`: prerendered -> span
- `Line`: prerendered -> line; line -> span
- `Text`: `style()` setter; prerendered -> text; text -> line
- `Paragraph`: `style()` setter; prerendered -> paragraph; paragraph -> text; text -> line and line -> span through the paragraph's grapheme path (which, unlike `Text::render`/`Line::render`, styles only the glyph cells); block -> paragraph inside the block

**Deliberately not pinned here:** whether `Paragraph::style` should reach the block's border cells (item 4 in the issue). The bg-only block test does not depend on it either way; `all_styles_combine` in `tests/text_style.rs` (#2693) already characterizes the current behavior, so this can change in one place if maintainers decide it should.

`Block`'s own tests are untouched for the same reason.

Claude Code helped write this PR.

### DIFF
--- a/ratatui-core/src/text/line.rs
+++ b/ratatui-core/src/text/line.rs
@@ -940,12 +940,6 @@ mod tests {
     }
 
     #[test]
-    fn style() {
-        let line = Line::default().style(Style::new().red());
-        assert_eq!(line.style, Style::new().red());
-    }
-
-    #[test]
     fn alignment() {
         let line = Line::from("This is left").alignment(Alignment::Left);
         assert_eq!(Some(Alignment::Left), line.alignment);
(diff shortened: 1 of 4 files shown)


orhun (approved):
LGTM, thanks!

PR #2721, opened 2026-08-17
fix(backend): account for symbol width when skipping cursor moves

### LINKED ISSUE #2651: CrosstermBackend skips MoveTo after a double-width grapheme, shifting the rest of the row
## Description

`CrosstermBackend::draw` decides whether to emit a `MoveTo` by comparing the update's
cell coordinate with the previous one:

https://github.com/ratatui/ratatui/blob/main/ratatui-crossterm/src/lib.rs — in `draw()`:

```rust
if !matches!(last_pos, Some(p) if x == p.x + 1 && y == p.y) {
    queue!(self.writer, MoveTo(x, y))?;
}
last_pos = Some(Position { x, y });
```

`last_pos` is tracked in **cell coordinates and does not account for the display width
of the symbol that was just printed**. After printing a double-width grapheme at `x`,
the terminal cursor sits at `x + 2`, but the backend believes it is at `x + 1`. So if
the next update targets `x + 1`, the `MoveTo` is suppressed and the write lands one
column to the right — shifting the remainder of the row.

Normally this is masked: for a wide cell `Buffer::diff` skips the trailing cell, so the
next update is `x + 2` and the coordinate check happens to force a `MoveTo`.

But `ratatui-core` **deliberately emits the trailing cell for emoji presentation
sequences containing VS16 (U+FE0F)** — the documented workaround for terminals that do
not clear the second half of a wide grapheme (`ratatui-core/src/buffer/diff.rs`):

```rust
// Work around terminals that fail to clear the trailing cell of certain
// emoji presentation sequences (those containing VS16 / U+FE0F).
```

On that path the backend prints the trailing cell without a `MoveTo`, so the workaround
itself shifts the row on every affected frame.

## To Reproduce

Self-contained, no terminal needed — the bytes the backend would emit are captured into
a `Vec<u8>`. Same buffers, two glyphs: `❤️` (U+2764 U+FE0F) and `😀` (U+1F600).

```rust
use ratatui::buffer::Buffer;
use ratatui::layout::Rect;
use ratatui::style::{Color, Style};
use ratatui::text::{Line, Span};

fn main() {
    for glyph in ["\u{2764}\u{FE0F}", "\u{1F600}"] {
        let area = Rect::new(0, 0, 10, 1);
        let mut prev = Buffer::empty(area);
        Paragraph::new(Line::from("aaaaaaaaaa")).render(area, &mut prev);

        let mut next = Buffer::empty(area);
        Paragraph::new(Line::from(vec![Span::styled(glyph, Style::new().bg(Color::Blue))]))
            .render(area, &mut next);

        let updates = prev.diff(&next);
        println!("{glyph:?} updates: {:?}", updates.iter().map(|(x, _, _)| *x).collect::<Vec<_>>());

        let mut out: Vec<u8> = Vec::new();
        CrosstermBackend::new(&mut out).draw(updates.into_iter()).unwrap();
        println!("bytes: {}", String::from_utf8_lossy(&out).replace('\u{1b}', "^["));
    }
}
```

Output (ratatui-core 0.1.2 / ratatui-crossterm 0.1.2):

```
"❤\u{fe0f}" updates: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
bytes: ^[[1;1H^[[39;48;5;4m❤️^[[39;49m         ^[[39m^[[49m^[[59m^[[0m

"😀" updates: [0, 2, 3, 4, 5, 6, 7, 8, 9]
bytes: ^[[1;1H^[[39;48;5;4m😀^[[1;3H^[[39;49m        ^[[39m^[[49m^[[59m^[[0m
```

- `😀` (trailing cell skipped by the diff): the next update is `x = 2`, the check fails,
  `^[[1;3H` is emitted — correct.
- `❤️` (trailing cell emitted by the VS16 workaround): **no `MoveTo` at all** after the
  glyph. Cell `x = 1` is written at physical column 2, cell `x = 2` at column 3, and so
  on — the whole remainder of the row is off by one; the last cell falls outside the row.

## Expected behavior

The write for cell `x + 1` should be preceded by `MoveTo(x + 1, y)`, i.e. contiguity
should be judged against the *physical* cursor position — `x == last.x + width_of(last
symbol)` rather than `x == last.x + 1`.

## Observed downstream impact

Reported against terminals that render such sequences at width 2 (Windows conhost /
Command Prompt, and per #2357 also macOS):

- the tail of any row containing a VS16 emoji is drawn one column to the right;
- a wide glyph that follows gets its right half overwritten by the shifted write, and
  the terminal then blanks the whole glyph — the emoji visibly disappears;
- panel borders on such rows are pushed outside the panel.

#2357 looks like the user-visible face of this (emoji `🚀️` = U+1F680 U+FE0F in a list,
artifacts on scroll, "characters from other lines appear"). Filing this separately since
the cause is in the backend rather than in the list widget.

## Possible fix

Track the width of the last printed symbol and use it for the contiguity check, e.g.:

```rust
if !matches!(last_pos, Some((p, w)) if x == p.x + w && y == p.y) {
    queue!(self.writer, MoveTo(x, y))?;
}
last_pos = Some((Position { x, y }, cell.symbol().width().max(1) as u16));
```

Other backends that track a "last position" the same way may need the same treatment.

## Related

- #2357 — likely the same root cause, seen through a list widget.
- #2585 — nearby area; its fix (restoring `invalidated`) landed in 0.1.2 and does make
  the trailing cell repaint on clear, but does not affect the `MoveTo` decision above.

## Adjacent observation (not the subject of this issue)

In the same repro, the trailing cell of the wide glyph carries the **default** style
(`bg(glyph) = Blue`, `bg(trailing) = Reset`). When that cell does get emitted, half of
the glyph is repainted with the default background, which shows up as a background that
covers only one half of the emoji. Happy to open a separate issue for that if it is a
distinct concern.

## Environment

- ratatui-core 0.1.2, ratatui-crossterm 0.1.2 (also reproduces on 0.1.1)
- crossterm 0.29
- Rust 1.96.0, Windows 11 (repro itself is terminal-independent — it inspects emitted bytes)

jackthepunished:
Fixes #2651

## Problem

The crossterm, termion and termina backends skip the cursor-move command when the current cell is at `(last.x + 1, last.y)`. That assumes every symbol advances the terminal cursor by exactly one column. After a double-width grapheme the cursor is two columns further along, so a write to the very next cell lands one column to the right and the rest of the row is shifted.

Normally `Buffer::diff` skips the trailing cell of a wide glyph and the coordinate check happens to force a move, which masks this. But the diff deliberately emits the trailing cell for VS16 emoji sequences (`❤️`, `🚀️`, …) as a workaround for terminals that don't clear it, and on that path no move is emitted at all — see the byte capture in the issue.

## Fix

Track `(position, width)` of the last cell written and only skip the move when `x == last.x + width`. Width comes from `CellWidth::cell_width`, which also honours `CellDiffOption::ForcedWidth`, so cells carrying escape sequences are handled the same way. Applied identically to all three backends that use this pattern.

## Testing

- New unit tests per backend: a wide glyph followed by `x + 1` now gets a cursor move; contiguous single-width cells and `x + 2` after a wide glyph still don't.
- crossterm additionally covers `ForcedWidth` and an end-to-end `Buffer::diff` → `draw` case reproducing the report (`aaaa` → `❤️bc`, trailing cell emitted at `x = 1`).
- `cargo clippy --all-features --all-targets` and `cargo test --all-features` for the three crates are clean.

### DIFF
--- a/ratatui-crossterm/src/lib.rs
+++ b/ratatui-crossterm/src/lib.rs
@@ -797,6 +797,47 @@ mod tests {
 
     use super::*;
 
+    /// Renders `content` and returns the emitted bytes as a lossy string.
+    fn draw_to_string(content: &[(u16, u16, &Cell)]) -> String {
+        let mut out = Vec::new();
+        CrosstermBackend::new(&mut out)
+            .draw(content.iter().copied())
+            .unwrap();
+        String::from_utf8_lossy(&out).into_owned()
+    }
+
+    #[test]
+    fn draw_moves_cursor_after_wide_symbol() {
+        // A double-width glyph advances the terminal cursor by two columns, so writing the
+        // very next cell requires an explicit cursor move. See issue #2651.
+        let wide = Cell::new("\u{2764}\u{FE0F}"); // ❤️ (VS16 emoji presentation)
+        let next = Cell::new("a");
+        let output = draw_to_string(&[(0, 0, &wide), (1, 0, &next)]);
+        let move_to = MoveTo(1, 0).to_string();
+        assert!(
+            output.contains(&move_to),
+            "expected `MoveTo(1, 0)` after a wide glyph, got: {output:?}"
+        );
+    }
+
+    #[test]
+    fn draw_skips_cursor_move_for_contiguous_cells() {
+        let a = Cell::new("a");
+        let b = Cell::new("b");
+        let output = draw_to_string(&[(0, 0, &a), (1, 0, &b)]);
+        assert_eq!(output.matches(&MoveTo(0, 0).to_string()).count(), 1);
+        assert!(!output.contains(&MoveTo(1, 0).to_string()));
+    }
+
+    #[test]
+    fn draw_skips_cursor_move_after_wide_symbol_when_contiguous() {
+        // The cell right after a wide glyph's trailing column is where the cursor already is.
+        let wide = Cell::new("\u{1F600}"); // 😀
+        let next = Cell::new("a");
+        let output = draw_to_string(&[(0, 0, &wide), (2, 0, &next)]);
+        assert!(!output.contains(&MoveTo(2, 0).to_string()));
+    }
+
     #[rstest]
     #[case(CrosstermColor::Reset, Color::Reset)]
     #[case(CrosstermColor::Black, Color::Black)]
--- a/ratatui-crossterm/src/lib.rs
+++ b/ratatui-crossterm/src/lib.rs
@@ -795,6 +795,10 @@ impl crate::crossterm::Command for ScrollDownInRegion {
 
 #[cfg(test)]
 mod tests {
+    use std::num::NonZeroU16;
+
+    use ratatui_core::buffer::{Buffer, CellDiffOption};
+    use ratatui_core::layout::Rect;
     use rstest::rstest;
 
     use super::*;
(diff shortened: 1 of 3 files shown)


orhun (approved):
LGTM!

PR #2723, opened 2026-08-18
fix(terminal): make insert_before a no-op when the terminal has no rows

jackthepunished:
## Problem

`Terminal::insert_before` loops forever when the terminal has zero rows and the `scrolling-regions` feature is off (the default). The fallback implementation draws at most `screen_height` lines per iteration; with `screen_height == 0` that is zero, so `while buffer_height + viewport_height > screen_height` never terminates. With `scrolling-regions` enabled it instead writes into row 0 of a terminal that has no rows.

Zero-row terminals do occur in practice — a pty created without a window size (test harnesses, some CI/container setups) reports `0x0` — so an inline app running there spins at 100% CPU instead of just printing nothing.

Found while fuzzing the terminal API with `TestBackend`; no existing issue.

Repro:

```rust
let backend = TestBackend::new(5, 0);
let mut terminal = Terminal::with_options(backend, TerminalOptions { viewport: Viewport::Inline(3) })?;
terminal.insert_before(2, |buf| buf.set_string(0, 0, "aaaaa", Style::default()))?; // hangs
```

## Fix

Return early from `insert_before` when `last_known_area.height == 0` — there is nowhere to put the lines. Documented alongside the existing non-inline no-op.

## Testing

- New test that hangs on `main` and passes with the fix, under both feature configurations.
- `cargo test -p ratatui-core` with and without `scrolling-regions`, clippy clean.

### DIFF
--- a/ratatui-core/src/terminal/inline.rs
+++ b/ratatui-core/src/terminal/inline.rs
@@ -66,6 +66,8 @@ impl<B: Backend> Terminal<B> {
     /// directly into the terminal's scrollback buffer. At the limit, if the viewport takes up the
     /// whole screen, all lines will be inserted directly into the scrollback buffer.
     ///
+    /// This is a no-op for non-inline viewports and when the terminal has no rows.
+    ///
     /// # Examples
     ///
     /// ## Insert a single line before the current viewport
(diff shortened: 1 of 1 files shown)


orhun (approved):
thanks!

PR #2724, opened 2026-08-18
fix(examples): use uppercase H/L for shifted color change in table example

CAESIUS-TIM:
When Shift is held, crossterm reports KeyCode::Char('L') / ('H') (uppercase), so the previous arms matching lowercase 'l' / 'h' with a shift_pressed guard never fired. This restores Shift+h/l color cycling.


orhun (approved):
thanks!

PR #2726, opened 2026-08-21
refactor(examples): remove unnecessary tick loop in scrollbar example

CAESIUS-TIM:
The example is keyboard-event driven with no animation or periodic state updates, so the 250ms tick only forces a redraw every 250ms when idle and last_tick is never read for rendering decisions. The comment documents why the pattern is retained for consistency with other examples.

### DIFF
--- a/examples/apps/scrollbar/src/main.rs
+++ b/examples/apps/scrollbar/src/main.rs
@@ -35,6 +35,11 @@ fn main() -> Result<()> {
 
 impl App {
     fn run(mut self, terminal: &mut DefaultTerminal) -> Result<()> {
+        // Note: This tick loop is unnecessary for this example — there is no
+        // tick-driven animation or periodic state update. The tick_rate only
+        // controls the poll timeout, causing unnecessary redraws every 250ms
+        // when idle. A blocking event::read() would be functionally equivalent.
+        // The tick pattern is retained only for consistency with other examples.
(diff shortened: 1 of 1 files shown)


examples/apps/scrollbar/src/main.rs
```diff
@@ -35,6 +35,11 @@ fn main() -> Result<()> {
 
 impl App {
     fn run(mut self, terminal: &mut DefaultTerminal) -> Result<()> {
+        // Note: This tick loop is unnecessary for this example — there is no
+        // tick-driven animation or periodic state update. The tick_rate only
+        // controls the poll timeout, causing unnecessary redraws every 250ms
+        // when idle. A blocking event::read() would be functionally equivalent.
+        // The tick pattern is retained only for consistency with other examples.
```
orhun: I don't like the tone in this, it basically says:

- This code is unnecessary.
- It redraws needlessly.
- We keep it because other examples do it.

Why don't we just remove it then?

CAESIUS-TIM: @orhun you are right — the defensive tone is not helpful. I have removed the tick loop entirely and replaced it with a blocking `event::read()`, matching the pattern used in `calendar-explorer` and other event-driven examples. The now-unused `std::time::{Duration, Instant}` import is also cleaned up.

orhun (approved):

PR #2727, opened 2026-08-22
feat(examples): add large-table example with pagination and virtual scrolling

CAESIUS-TIM:
## What

Add a new `large-table` example that demonstrates how to handle large
datasets with pagination and virtual scrolling.

Key features:

- `PgUp`/`PgDn` and `Home`/`End` key bindings (`u`/`d`/`g`/`G`
  alternatives), paging the viewport from the current offset rather than
  the selected row.
- Virtual windowing: only the visible rows are passed to `Table` on
  each draw, keeping the example responsive with large `--list-size`
  values.
- A `clap` CLI argument `--list-size`/`-n` (default 20); passing `0`
  exercises the empty-list guards in the navigation methods.
- Resize-aware offset clamping that keeps the selected row visible
  when the terminal shrinks and fills the viewport when it regrows.

## Design notes

- Selection and scroll offset are stored in `App` fields so the visible
  window can be computed before drawing.
- `render()` clamps the offset after resize, then `render_table()`
  syncs `TableState` to window coordinates and reads back auto-scrolled
  offsets.
- An empty window (tiny terminal or empty list) renders without
  clobbering the selection state.

## Needs attention

- The window-to-global coordinate sync in `render_table` and the
  `complete_rows` computation `(height - HEADER_HEIGHT) / ITEM_HEIGHT`.

## Tests

`cargo test -p large-table` — 6 tests covering: offset clamping on
shrink/regrow, paging with the selection mid-viewport, empty-list
navigation, wrap-around arrow navigation, and pulling the offset back
when the selection moves above the viewport.

## AI assistance

This PR was created with the assistance of [oh-my-pi](https://github.com/can1357/oh-my-pi). I have reviewed
every line for correctness and relevance.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -79,6 +79,10 @@ unicode-truncate = { version = "2", default-features = false }
 # See <https://github.com/ratatui/ratatui/issues/1271> for information about why we pin unicode-width
 unicode-width = ">=0.2.0"
 
+[patch.crates-io]
+ratatui-core = { path = "ratatui-core" }
+ratatui-widgets = { path = "ratatui-widgets" }
+
 [workspace.lints.rust]
 unsafe_code = "forbid"
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3669,6 +3669,7 @@ dependencies = [
 name = "table"
 version = "0.0.0"
 dependencies = [
+ "clap",
  "color-eyre",
  "crossterm 0.29.0",
  "fakeit",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1887,6 +1887,19 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bf36173d4167ed999940f804952e6b08197cae5ad5d572eb4db150ce8ad5d58f"
 
 [[package]]
+name = "large-table"
+version = "0.0.0"
+dependencies = [
+ "clap",
+ "color-eyre",
+ "crossterm 0.29.0",
+ "fakeit",
+ "itertools 0.15.0",
+ "ratatui",
+ "unicode-width",
+]
+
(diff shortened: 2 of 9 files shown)


orhun:
I think there is a nice opportunity to use the new/experimental [`ratatui-table`](https://github.com/ratatui/ratatui-table) widget here. It has some [interesting pull requests](https://github.com/ratatui/ratatui-table/pulls) as well.

orhun:
Would love to see `ratatui-table` in this one

Cargo.toml
```diff
@@ -79,6 +79,10 @@ unicode-truncate = { version = "2", default-features = false }
 # See <https://github.com/ratatui/ratatui/issues/1271> for information about why we pin unicode-width
 unicode-width = ">=0.2.0"
 
+[patch.crates-io]
+ratatui-core = { path = "ratatui-core" }
+ratatui-widgets = { path = "ratatui-widgets" }
+
```
CAESIUS-TIM, line 85: @orhun I've migrated `large-table` to `ratatui-table` in the latest commit abc1516.

One thing to flag: `ratatui-table 0.1.0` depends on `ratatui-core` and `ratatui-widgets` from `crates.io`, but the workspace uses local path versions. To avoid duplicate crate instances, I added a `[patch.crates-io]` section to the root `Cargo.toml` — see the diff. Let me know if you'd prefer a different approach for this.

CAESIUS-TIM:
I've been looking at this more, and I think ratatui/ratatui-table/pull/14 (lazy_rows) does the core work this example was meant to demonstrate — the virtual windowing, offset clamping, and selection
anchoring are all handled by the widget now. Reimplement that logic only to have it superseded doesn't make sense. Thanks for pointing me at ratatui-table!

PR #2731, opened 2026-08-24
perf(terminal): skip the redundant screen clear on horizontal shrink

davidszp:
## what and how

On a horizontal shrink with a `Fullscreen` viewport, `Terminal::resize` clears the screen twice:
once explicitly against line wrapping, then again in `clear_viewport`, whose `Fullscreen` arm is
the same `clear_region(ClearType::All)`. Only `set_viewport_area` runs between them, which touches
no backend. This skips the first when `clear_viewport` will issue it anyway.

`Inline` and `Fixed` must keep it — there `clear_viewport` only reaches from the viewport origin
down, or covers the fixed rect. Backend calls for a horizontal shrink:

| viewport | before | after |
|---|---|---|
| `Fullscreen` | `[All, All]` | `[All]` |
| `Inline(5)` | `[All, AfterCursor]` | unchanged |
| `Fixed` | `[All]` | unchanged |

The second clear arrived with #2355, which fixed line wrapping in the **inline** viewport (#2086);
its issue and review are about inline throughout. It went into the shared path, so `Fullscreen` —
what `Terminal::new` and `ratatui::init` use — picked it up as a side effect.

## why (who cares?)

The second erase is idempotent, so nothing renders wrong and it went unnoticed. It costs one extra
full-screen erase per horizontal shrink on the default viewport, which adds up while a window edge
is being dragged: a `foot` window under Hyprland delivers ~37 resize events a second.

Capturing everything a TUI writes to a pty for one shrink, 200 → 140 cols, the two streams differ
by exactly one `\x1b[2J` — 7,474 bytes vs 7,478, with all 7,468 bytes of repaint identical. Same
screen, one less erase.

## tests

Three, one per viewport; the two asserting no change matter as much as the third. They need a
backend that records commands rather than applying them, because `TestBackend` implements
`clear_region(ClearType::All)` as a buffer reset — one clear and two are indistinguishable in its
buffer. The wrapper follows `FallibleTestBackend` in `terminal/render.rs`, which exists for the
same reason. Reverting the fix and keeping the tests fails with `left: [All, All]`, `right:
[All]`.

## things worth a second look

- The wrapper is a new test fixture. If you would rather not have it, the fix stands alone.
- Equivalence assumes `clear_region(ClearType::All)` is idempotent. It is for every in-tree
  backend (crossterm, termion, termwiz, termina — all single stateless erases); a third-party
  backend with side effects in `clear` would see one call instead of two.
- I could not run `cargo xtask format` — no nightly rustfmt here. Stable `cargo fmt --check` is
  clean and the unstable options were checked by hand, but CI is the authority.

### DIFF
--- a/ratatui-core/src/terminal/resize.rs
+++ b/ratatui-core/src/terminal/resize.rs
@@ -91,8 +91,8 @@ mod tests {
     use alloc::vec::Vec;
 
     use crate::backend::{Backend, ClearType, TestBackend, WindowSize};
-    use crate::buffer::Buffer;
-    use crate::layout::{Position, Rect};
+    use crate::buffer::{Buffer, Cell};
+    use crate::layout::{Position, Rect, Size};
     use crate::terminal::{Terminal, TerminalOptions, Viewport};
 
     /// A [`TestBackend`] that records clear commands.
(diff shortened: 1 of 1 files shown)


orhun (approved):
LGTM!

PR #2736, opened 2026-08-24
fix(fill): reject empty symbols

### LINKED ISSUE #2730: Fill::new("") silently corrupts terminal layout due to zero-width symbol
## Description

Fill::new("") silently breaks terminal layout. The type signature accepts any Into<Cow<str>>, so passing "" compiles without warning, but produces broken rendering at runtime.

## To Reproduce

```
use ratatui::widgets::{Fill, Widget};
use ratatui::layout::Rect;

let mut buf = ratatui::buffer::Buffer::empty(Rect::new(0, 0, 10, 1));
Fill::new("A").render(Rect::new(0, 0, 3, 1), &mut buf); 
Fill::new("").render(Rect::new(3, 0, 1, 1), &mut buf);  // empty fill
Fill::new("B").render(Rect::new(4, 0, 3, 1), &mut buf); 
// "B" cells appear shifted left when rendered to the terminal
```

## Expected behavior

Since the symbol is empty we should be a return condition when the symbol is empty or we should panic with message like Fill symbol should not be empty

## Environment

OS: Windows
Terminal Emulator: VS Code Integrated Terminal
Font:
Crate version: 0.30.2
Backend: crossterm

orhun:
Empty symbols create zero-width cells that can disrupt terminal rendering.
Reject them in `Fill::new` and `Fill::symbol` with a clear panic message,
while keeping `Fill::default()` safe to render as a no-op.

fixes #2730

### DIFF
--- a/ratatui-widgets/src/fill.rs
+++ b/ratatui-widgets/src/fill.rs
@@ -45,6 +45,7 @@ use ratatui_core::widgets::Widget;
 ///
 /// Cells outside the buffer are silently clipped, mirroring the behavior of other widgets
 /// such as [`Clear`](crate::clear::Clear).
+/// An empty symbol leaves the buffer unchanged.
 ///
 /// [`Stylize`]: ratatui_core::style::Stylize
 #[derive(Debug, Default, Clone, Eq, PartialEq, Hash)]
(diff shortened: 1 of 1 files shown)


joshka:
What about making `Fill::new("")` panic instead, as someone might assume this is the same as `Flil::new(" ")`?

kb019:
> What about making `Fill::new("")` panic instead, as someone might assume this is the same as `Flil::new(" ")`?

Exactly, that’s how I ran into this issue.

Issue #2610, opened 2026-06-19
a11y support


non-descriptive:
## Problem
Interactive tui apps usually dont have concise accessibility feature support, and sometimes they might be hard to use. 

## Solution
Add feature to enable support for accessibility features. Could be done using  [AccessKit](https://github.com/AccessKit/accesskit). There are maybe other frameworks for it, but this is open to suggestion

## Are you willing to contribute an implementation?

- [ ] I am willing to open a PR implementing this.
- [ ] I can try to implement it, but I will need guidance.
- [x] I am not able to implement this right now.

## Additional context

This might be also used for some autotests via things like [xa11y](https://github.com/xa11y/xa11y)

joshka:
This is one of those hard problems that I've looked into a bit and gotten nowhere because of the impedance mismatch of trying to map a location based output onto a semantic one. It would be like trying to explain a painting to a blind person by telling them about each brush stroke in a random order.

That's not to say that this is not possible, I think it just doesn't have any easy answer without adding things at the terminal protocol levels. I think there's some possible ways that various named blocks could be extended to help know things.

The basics that are needed here aren't really there in this library though. We don't have a concept of focus, state, selection, etc. We're just the rendering side, and our rendering is effectively render characters at locations, not words / etc.

I would welcome and support anyone working on this, but it's not something I can personally prioritize.

joshka:
See https://lobste.rs/c/awixrd and https://lobste.rs/c/dvvmmj for some more background

non-descriptive:
I have no urgency to have this and don't mind this issue gon' stale or even closed as not planned at all. Its nice to have some state of things closer to the code and not somewhere on unsearchable discord or reddit/lobsters/other forums. 

I might return to this issue and was wondering if integration of this will require changes to ratatui guts or it will work just fine as some external plugin/widget. Of course disregarding poor terminal, of course. I bet there are other issues with OS support and tooling so its nice you've already done research on the matter.  

>  We don't have a concept of focus, state, selection, etc

Exactly the thing I was wondering about. There is no state, so thats why me thought of plugin which will provide it in some sense.  But, I am not yet done any research on what will be required from ratatui side and how sufficient current API. Had some slight ideas about crafting interactive layout manager akin VS Studio Forms or Delphi RAD, but of course it requires time time time. 

Thanks, regardless.

joshka:
It is highly likely that nothing in this would be at all simple.
[...]

PR #2739, opened 2026-08-28
feat(a11y): experimental Linux AT-SPI accessibility crate

rit3sh-x:
## What

Adds `ratatui-a11y`: an opt-in, Linux-only crate that serves an app-declared accessibility tree to AT-SPI clients (screen readers like Orca) via AccessKit.
See #2610 for the background discussion.

## Why this shape

- Ratatui is immediate-mode, no retained widget tree, nothing can infer semantics from the rendered `Buffer` (cells only). The app declares the tree explicitly, alongside its normal render code.
- AT-SPI needs no OS window handle (unlike Windows UIA / macOS NSAccessibility, which both need a real window a terminal app doesn't have) -- that's why this PR is Linux-only.
- Adapters (`list_nodes`, `table_nodes`, `tabs_nodes`, `gauge_nodes`, `text_nodes`, `group_nodes`, `items_nodes`) take plain data, not widget references `ratatui-widgets` fields are private, and this also means the crate works with any hand-rolled widget or TUI framework, not just Ratatui's built-ins.
- `ratatui-a11y` has zero dependency on `ratatui` outside dev-deps.

## What this PR does NOT touch

- No changes to `ratatui`, `ratatui-widgets`, or `ratatui-core`.
- Excluded from `default-members` (see `Cargo.toml`), same as `ratatui-termion` -- `cargo build`/`cargo test` at the repo root are unaffected on every platform.
- Windows/macOS are not attempted here (see crate docs for why).

## Try it

```sh
gsettings set org.gnome.desktop.interface toolkit-accessibility true  # if needed
cargo run -p ratatui-a11y --example list       # full focus/selection round trip
cargo run -p ratatui-a11y --example dashboard  # tabs + table + gauge composed
```

Verified with Orca 46.1: auto-discovers the app, announces role + label + state, tracks live selection.

## Testing

- Unit tests: `cargo test -p ratatui-a11y`
- Doc tests included
- Manual: Orca screen reader against all three examples

## AI disclosure

Used AI for generating comments and Docs.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,57 @@
 version = 4
 
 [[package]]
+name = "accesskit"
+version = "0.24.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d3b7f7f85a7e5f68090000ed7622545829afd484d210358702ae4cb97dd0c320"
+dependencies = [
+ "uuid 1.18.0",
+]
+
+[[package]]
+name = "accesskit_atspi_common"
+version = "0.19.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "023da0e5097f46df7092d5280b02efb9bbf8d93298daeced42652463e357d636"
+dependencies = [
+ "accesskit",
+ "accesskit_consumer",
+ "atspi-common",
+ "phf 0.13.1",
+ "serde",
+ "zvariant",
+]
+
+[[package]]
+name = "accesskit_consumer"
+version = "0.38.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5d10a236f96f87d70732e44520046785431ef01d5bcd6b041317bfadd2f88245"
+dependencies = [
+ "accesskit",
+ "hashbrown 0.16.1",
+]
+
+[[package]]
+name = "accesskit_unix"
+version = "0.22.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "03e156ed3802e35eefe894ef2671bc6c889303d8a7e110b5e1b48f504b91362f"
+dependencies = [
+ "accesskit",
+ "accesskit_atspi_common",
+ "async-channel",
+ "async-executor",
+ "async-task",
+ "atspi",
+ "futures-lite",
+ "futures-util",
+ "serde",
+ "zbus",
+]
+
+[[package]]
 name = "addr2line"
 version = "0.24.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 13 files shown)


joshka:
Can you do a bit of a run through narrative / video / screenshots / gifs of what you get when running this?
I'm fine with AI code generally, but where possible it's useful to get the AI to help cut down review time by making sure it explains things better and justifies any decisions etc. So it would be helpful to be a lot more step-by-step on what problems this solves and how (let your clanker generate this if needed but tune it to respect the reviewer attention -  take a look at https://www.joshka.net/practice/rules/documentation/docs-avoid-generated-prose-tells/ and https://github.com/cursor/plugins/blob/main/pstack/skills/unslop/SKILL.md for some ideas on helping it avoid too much slop docs).

The key here is "show, don't tell". The code is basically immaterial (i.e. it's not the thing that cost you here), without the payoff of seeing that this is doing what it says and helping explain the constraints etc.

rit3sh-x:
@joshka 

image

Here is the screenshot from Accerciser. You can see the accessibility tree being rendered alongside the TUI, with the nodes corresponding to the elements in the application.

rit3sh-x:
@orhun Could you also take a look at this?

orhun:
Wow, thanks for working on this!

Since this crate has no direct dependency on Ratatui and can work with other TUI frameworks, I believe it would be a better fit as a separate repository rather than living in the Ratatui workspace.

Once published, we can:

- add it to [awesome-ratatui](https://github.com/ratatui/awesome-ratatui) so people can find it.
- mention it on the website, as part of https://github.com/ratatui/ratatui-website/issues/947
- if we see adoption, we can move it under https://github.com/ratatui-org 

This is just a scope decision! Love what you're pursing here :)

rit3sh-x:
@orhun Agreed, that makes sense - it doesn't touch ratatui directly so it shouldn't live in this workspace long term.

For a name I'd suggest `tui-a11y`. Short, and matches how the crate is actually scoped (works with any TUI, not just ratatui).

One question on process: should I just spin up a personal repo for it now and we link it from awesome-ratatui/the website once it's usable, or would you rather it start directly under the ratatui-org umbrella? Happy to go either way, just want to avoid setting it up twice.

rit3sh-x:
One more thing worth flagging: part of why I leaned toward keeping it in ratatui was testing reach. I can only test on Linux myself, and if/when Windows (UIA) or macOS (NSAccessibility) support gets added, being in the main repo means access to ratatui's existing CI matrix and contributors on those platforms. As a standalone repo I'd have to set that up (and find testers) from scratch. Not a blocker, just wanted to put it on the table before we split it out.

orhun:
We usually try to follow to convention of `tui-` prefix for old `tui-rs` libraries and `ratatui-` prefix for the Ratatui ones but I guess it's fine to go with `tui-a11y` 👍🏼 

Feel free to create your own repo and submit it to awesome-ratatui!

Issue #2734, opened 2026-08-24
New kinds of underlined texts


shafai-pouya:
## Problem

So, ratatui supports colored underlines. But in my project, I need different kinds of underlines (e.g.: Curved underlines, double underlines). 

## Solution

You can add new Modifiers for these (Modifier::DOUBLE_UNDERLINED, Modifer::CUREVE_UNDERLINED). 

## Are you willing to contribute an implementation?

- [x] I am willing to open a PR implementing this.
- [ ] I can try to implement it, but I will need guidance.
- [ ] I am not able to implement this right now.

## Additional context

I also found the escape sequence for these types:
Curved: `\e[4:3m`
Doubled: `\e[21m`

PR #2740, opened 2026-08-28
feat: custom underlines

shafai-pouya:
## Summary

* Add support for additional underline types (`Modifier::UNDER_CURLED`, `Modifier::DOUBLE_UNDERLINED`, `Modifier::UNDER_DOTTED`, `Modifier::UNDER_DASHED`).

## Notes

* `termion` is not supported.
* Multiple underline types cannot be applied to the same cell to prevent UB.

Related to #2734

### DIFF
--- a/ratatui-crossterm/src/lib.rs
+++ b/ratatui-crossterm/src/lib.rs
@@ -577,7 +577,7 @@ impl ModifierDiff {
         if removed.contains(Modifier::ITALIC) {
             queue!(w, SetAttribute(CrosstermAttribute::NoItalic))?;
         }
-        if removed.contains(Modifier::UNDERLINED) {
+        if (removed & Modifier::ALL_UNDERLINES) != Modifier::empty() {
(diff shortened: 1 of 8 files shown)


ratatui-crossterm/src/lib.rs
orhun, line 453: Some variants are missing here

orhun, line 706: We should remove all underlines

orhun (commented):
Thanks for the PR!

Left some comments.
Also, the `anstyle` conversions are missing :)

orhun:
Hey, still missing a few things here:
- Add the new underline helpers to `Stylize`, and make `not_underlined()` clear every underline type.
- Fix `anstyle::Effects` -> `Modifier` so underline types aren’t collapsed to regular underline.
- Fix `Style::add_modifier()` so combined underline flags cannot leave conflicting underline types active.

PR #2741, opened 2026-08-29
feat(terminal): add `ratatui::run_with_options`

### LINKED ISSUE #2500: `ratatui::run` with inline viewport
## Problem

There is no way (as far as i can tell) to `ratatui::run` with inline viewport. The `run` function is convenient because it handles the restoring of the terminal no matter how the program quits (early return via error, panic, or program termination).

## Solution

New `ratatui::run_with_options` function that mirrors `ratatui::init_with_options`. Redefine `ratatui::run` in terms of `ratatui::run_with_options` with default `TerminalOptions`.

example:
```rust
ratatui::run_with_options(
    TerminalOptions { viewport: Viewport::Inline(16) },
    |term| {
        term.draw(|| {});
    }
) // handles ratatui::restore at the end of the callback
```

## Alternatives

if we want the api to look as simple as possible, we could consider a `ratatui::run_inline(height, callback)` function instead. The 2 functions are not mutually exclusive tho, `run_with_options` could be the base for `run_inline`

## Footnotes

I can submit a PR pretty quickly if it is something deemed useful.

CyberneticX-Tech:
Resolves #2500

### Description

This PR introduces `ratatui::run_with_options`, providing an ergonomic way to run applications with custom `TerminalOptions` (such as `Viewport::Inline` or `Viewport::Fixed`) while retaining automatic terminal restoration on exit.

### Changes
- Added `ratatui::run_with_options(options: TerminalOptions, f: F)` in `ratatui/src/init.rs`.
- Re-exported `run_with_options` at crate root (`ratatui::run_with_options`).
- Updated module and crate-level documentation, examples, and comparison tables.

### Notes on `ratatui::run`
`ratatui::run` continues to delegate to `init()`, which enters the alternate screen buffer for fullscreen apps, whereas `run_with_options` delegates to `init_with_options(options)`, which avoids alternate screen for custom/inline viewports.

### AI Disclosure
In accordance with CONTRIBUTING.md, this contribution was assisted by AI tools and reviewed for correctness, quality, and project conventions.

### DIFF
--- a/ratatui/src/init.rs
+++ b/ratatui/src/init.rs
@@ -324,10 +324,7 @@ pub fn run<F, R>(f: F) -> R
 where
     F: FnOnce(&mut DefaultTerminal) -> R,
 {
-    let mut terminal = init();
-    let result = f(&mut terminal);
-    restore();
-    result
+    run_terminal(init(), f, restore)
 }
 
 /// Run a closure with a terminal initialized with the given options.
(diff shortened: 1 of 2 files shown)


orhun (approved):
LGTM!
 
When this was originally discussed, GitHub code search showed roughly 140 matches for `ratatui::run` compared with around 4k for `ratatui::init`. It now shows roughly 900 matches for run and ~6k for init, so I think there is enough adoption to extend this API.

PR #2743, opened 2026-08-31
fix(buffer): clear trailing cells before wide glyph repaint

### LINKED ISSUE #2652: Trailing cell of a wide glyph is never updated on style-only changes, leaving a stale half-background
## Description

`Buffer::diff` treats the trailing cell of a wide grapheme as skipped and resets it to the
**default** style, so a wide glyph's own style covers only its leading cell.

When a wide glyph **stays wide** and only its *style* changes — e.g. a selection highlight
moving off it — the diff emits **only the leading cell**. The trailing column is never
written, so terminals that keep per-cell attributes (Windows conhost / Command Prompt) keep
the *previous* background there. The result is a background that covers only the left half
of the emoji.

This is distinct from #2651 (which is about the `MoveTo` decision): here the trailing cell
is not emitted at all.

## To Reproduce

Self-contained, no terminal needed — the bytes the backend would emit are captured into a
`Vec<u8>`. Three transitions, two glyphs: `❤️` (U+2764 U+FE0F) and `😀` (U+1F600).

```rust
use ratatui::buffer::Buffer;
use ratatui::layout::Rect;
use ratatui::style::{Color, Style};
use ratatui::text::{Line, Span};

const AREA: Rect = Rect { x: 0, y: 0, width: 8, height: 1 };

fn frame(glyph: &str, bg: Option<Color>) -> Buffer {
    let mut buf = Buffer::empty(AREA);
    let style = bg.map_or_else(Style::new, |c| Style::new().bg(c));
    Paragraph::new(Line::from(vec![Span::styled(glyph.to_string(), style)])).render(AREA, &mut buf);
    buf
}

fn text_frame(s: &str) -> Buffer {
    let mut buf = Buffer::empty(AREA);
    Paragraph::new(Line::from(s.to_string())).render(AREA, &mut buf);
    buf
}

fn show(label: &str, prev: &Buffer, next: &Buffer) {
    let updates = prev.diff(next);
    let cols: Vec<u16> = updates.iter().map(|(x, _, _)| *x).collect();
    println!("{label}");
    println!("  glyph cell bg   : {:?}", next[(0, 0)].bg);
    println!("  trailing cell bg: {:?}", next[(1, 0)].bg);
    println!("  updates at cols : {cols:?}");
    println!("  trailing emitted: {}", cols.contains(&1));
    let mut out: Vec<u8> = Vec::new();
    CrosstermBackend::new(&mut out).draw(updates.into_iter()).unwrap();
    println!("  bytes           : {}\n", String::from_utf8_lossy(&out).replace('\u{1b}', "^["));
}

fn main() {
    for glyph in ["\u{2764}\u{FE0F}", "\u{1F600}"] {
        println!("=== glyph {glyph:?} ===\n");
        show("(A) styled glyph appears over plain text", &text_frame("abcdefgh"), &frame(glyph, Some(Color::Blue)));
        show("(B) glyph stays wide, loses its background", &frame(glyph, Some(Color::Blue)), &frame(glyph, None));
        show("(C) styled glyph replaced by narrower content", &frame(glyph, Some(Color::Blue)), &text_frame("        "));
    }
}
```

Output (ratatui 0.30.1 / ratatui-core 0.1.2 / ratatui-crossterm 0.1.2):

```
=== glyph "❤\u{fe0f}" ===

(A) styled glyph appears over plain text
  glyph cell bg   : Blue
  trailing cell bg: Reset
  updates at cols : [0, 1, 2, 3, 4, 5, 6, 7]
  trailing emitted: true

(B) glyph stays wide, loses its background
  glyph cell bg   : Reset
  trailing cell bg: Reset
  updates at cols : [0]
  trailing emitted: false

(C) styled glyph replaced by narrower content
  updates at cols : [0, 1]
  trailing emitted: true

=== glyph "😀" ===

(A) styled glyph appears over plain text
  glyph cell bg   : Blue
  trailing cell bg: Reset
  updates at cols : [0, 2, 3, 4, 5, 6, 7]
  trailing emitted: false

(B) glyph stays wide, loses its background
  updates at cols : [0]
  trailing emitted: false

(C) styled glyph replaced by narrower content
  updates at cols : [0, 1]
  trailing emitted: true
```

Two things the output shows:

- **The trailing cell never carries the glyph's style** — `glyph cell bg: Blue`,
  `trailing cell bg: Reset` in (A), for both glyphs.
- **Case (B) emits only column 0** for both `❤️` and `😀`. Nothing is ever written to the
  trailing column, so whatever background was there stays. This is not VS16-specific: it
  affects wide glyphs generally, unlike the VS16 clearing workaround.

Case (C) is the one #2585 fixed in 0.1.2 (wide → narrower with a style visible on blank),
and it is not affected.

## Expected behavior

After a style-only change on a wide glyph, the glyph should be repainted with its new style
across both of its columns — i.e. the change should be detected and emitted, rather than the
trailing column being left with a stale attribute.

## Why this cannot be worked around downstream

The obvious workaround — force a full repaint of every cell — does not help: repainting that
cell means *writing* to the second half of a wide glyph, which is exactly what #2651 makes
unsafe (the write lands one column further right and shifts the rest of the row). So a
downstream application must choose between a stale half-background and a shifted row.

## Possible fix

Let the trailing cell inherit the leading cell's style instead of resetting it to the
default. Style-only changes would then be detected by the diff, and combined with a
width-aware `last_pos` (#2651) the resulting trailing write would also be positioned
correctly.

## Related

- #2651 — same area; the `MoveTo` decision that makes writing the trailing cell unsafe. Its
  "adjacent observation" section covers the *emitted-with-default-style* case; this issue is
  about the trailing cell never being emitted at all.
- #2585 — fixed the wide → narrower transition (case C) in 0.1.2; case (B) is unaffected by it.
- #2357 — likely the user-visible face of the wide-glyph handling in general.

## Environment

- ratatui 0.30.1, ratatui-core 0.1.2, ratatui-crossterm 0.1.2
- crossterm 0.29
- Rust 1.96.0, Windows 11 (the repro itself is terminal-independent — it inspects emitted bytes)

baba9811:
Fixes #2652.

## Problem

Style-only updates redraw only the leading cell of a wide glyph. The trailing columns can keep stale background or underline attributes.

## Fix

Clear affected trailing cells first, then repaint the glyph.

Related:

- #2686 uses the same deferred-cell ordering for a separate width issue.
- [fdehau/tui-rs#104](https://github.com/fdehau/tui-rs/issues/104) added width-aware buffer diffing for changing or moving wide glyphs; style-only changes to an unchanged glyph remained uncovered.

## Tests

- `BufferDiff` cases for `❤️`, `😀`, and foreground-only changes
- Crossterm case for underlined Korean text
- `cargo xtask ci`

AI assistance was used. I reviewed every changed line.

### DIFF
--- a/ratatui-core/src/buffer/diff.rs
+++ b/ratatui-core/src/buffer/diff.rs
@@ -166,7 +166,8 @@ impl<'next> Iterator for BufferDiff<'_, 'next> {
                     let previous_style_is_visible_on_blank = previous.bg != Color::Reset
                         || previous.modifier.intersects(VISIBLE_ON_BLANK);
 
-                    // Clear stale styles from an unchanged wide glyph's trailing cells before repainting it.
+                    // Clear stale styles from an unchanged wide glyph's trailing cells before
+                    // repainting it.
                     if cell_width > 1
                         && current.symbol() == previous.symbol()
                         && current.style() != previous.style()
--- a/ratatui-core/src/buffer/diff.rs
+++ b/ratatui-core/src/buffer/diff.rs
@@ -17,14 +17,14 @@ pub struct BufferDiff<'prev, 'next> {
     area: Rect,
     /// Current position in the flat cell array.
     pos: usize,
-    /// Tracks trailing cells that must be yielded after a wide character is processed.
+    /// Tracks trailing cells that must be yielded around a wide character update.
     ///
-    /// Set when a wide char was replaced by narrower content (force=true) or when a VS16 emoji
-    /// needs its trailing column checked (force=false).
+    /// Set before repainting a wide glyph after a visible style change, after replacing a wide
+    /// glyph with narrower content, or when a VS16 emoji needs its trailing column checked.
     trailing: Option<TrailingState>,
 }
 
-/// Tracks pending trailing-cell yields when a wide character is followed by narrower content.
+/// Tracks pending trailing-cell yields for a wide character update.
 #[derive(Debug)]
 struct TrailingState {
     next_index: usize,
@@ -36,6 +36,8 @@ struct TrailingState {
     /// When `false` (VS16 path), only cells whose symbol changed are emitted, because the emoji
     /// visually covers its trailing column and style differences there are invisible.
     force: bool,
+    /// A cell to yield once the trailing range is exhausted.
+    deferred: Option<usize>,
 }
 
 /// Modifiers that are visually apparent on a blank (space) cell.
@@ -97,6 +99,7 @@ impl<'next> Iterator for BufferDiff<'_, 'next> {
             next_index,
             end,
             force,
+            deferred,
         }) = &mut self.trailing
         {
             while *next_index < *end {
@@ -115,9 +118,15 @@ impl<'next> Iterator for BufferDiff<'_, 'next> {
                 }
             }
 
-            // Done with trailing cells; resume main loop past the wide character.
+            // Done with trailing cells; resume past the wide character, repainting its leading
+            // cell first when the trailing cells were a pre-clear.
             self.pos = *end;
+            let deferred = deferred.take();
             self.trailing = None;
+            if let Some(i) = deferred {
+                let (x, y) = self.pos_of(i);
+                return Some((x, y, &self.next[i]));
+            }
         }
         while self.pos < len {
             let i = self.pos;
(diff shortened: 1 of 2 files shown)


orhun:
thanks!

orhun:
forgot to approve, but approved :D

PR #2755, opened 2026-09-07
ci: run test-libs on stable only

### LINKED ISSUE #2750: `Test Libs 1.88.0` has never run on 1.88.0: ci.yml:306 hardcodes `stable` (same bug as #2106/#2116)
The `test-libs` matrix declares `toolchain: ["1.88.0", "stable"]`, but [`ci.yml:306`](https://github.com/ratatui/ratatui/blob/a0189ae4af65f85affef2a4b52bc53551cf50a1d/.github/workflows/ci.yml#L306) installs `toolchain: stable` literally. `matrix.toolchain` appears nowhere else in the job except [its name](https://github.com/ratatui/ratatui/blob/a0189ae4af65f85affef2a4b52bc53551cf50a1d/.github/workflows/ci.yml#L294), so `Test Libs 1.88.0` and `Test Libs stable` are the same job run twice.

In the latest run on main ([33879050611](https://github.com/ratatui/ratatui/actions/runs/33879050611), all 35 jobs green), the `Test Libs 1.88.0` job installs `stable` and reports `rustc 1.98.1`; the string `1.88` occurs exactly once in its 4101-line log, on the job-name line. `test-libs` is in the [`required`](https://github.com/ratatui/ratatui/blob/a0189ae4af65f85affef2a4b52bc53551cf50a1d/.github/workflows/ci.yml#L336) gate's `needs:`, so that green is load-bearing.

This is the same defect as #2106 and #2116 ("the ... CI jobs were unintentionally using the stable toolchain from `rust-toolchain.toml` instead of the matrix-specified toolchain"), both merged a year ago. `test-libs` is the third matrix-toolchain job in the file and the only one that got neither half — and it needs both, since `rust-toolchain.toml` pins `channel = "stable"` and overrides `rustup default` inside the tree:

```yaml
      - uses: dtolnay/rust-toolchain@…
        with:
          toolchain: ${{ matrix.toolchain }}          # ci.yml:306, currently `stable`
      …
      - run: cargo xtask test-libs
        env:
          RUSTUP_TOOLCHAIN: ${{ matrix.toolchain }}   # currently absent
```

That's what `check` carries at `:188`/`:195` and `lint-clippy` at `:129`/`:134`. Introduced by #1461 (`2ef3583e`, 2024-11-02): the `test` job it replaced did pass `${{ matrix.toolchain }}` through.

**I ran the fix on a fork first, because the useful question is what happens when the leg actually starts running** ([run 33897565331](https://github.com/glaziermag/ratatui/actions/runs/33897565331), four jobs — current wiring and fixed wiring, each over both toolchains):

| job | rustc actually used | result |
|---|---|---|
| as-is · `1.88.0` leg | `rustc 1.98.1` | pass |
| as-is · `stable` leg | `rustc 1.98.1` | pass |
| **wired · `1.88.0` leg** | **`rustc 1.88.0`** | **fail — see below** |
| wired · `stable` leg | `rustc 1.98.1` | pass |

That failure is **not** an MSRV break — every real test passes on 1.88.0. The only casualty is `ratatui-macros`' trybuild case at [`tests/macros.rs:100`](https://github.com/ratatui/ratatui/blob/a0189ae4af65f85affef2a4b52bc53551cf50a1d/ratatui-macros/tests/macros.rs#L100), which diffs rustc's stderr for `tests/ui/fails.rs` against blessed snapshots; those are version-specific, and a channel gate wouldn't help since 1.88.0 *is* stable. So the two-line fix wants a companion change keeping that one test on a single toolchain. I'd rather flag the constraint than guess at the shape you'd prefer.

Impact, stated plainly: MSRV *compilation* is still covered, because `check`'s 1.88.0 legs are wired correctly and `cargo hack check --all-targets --all-features` type-checks tests and benches. What has never happened is MSRV test *execution* — plus a duplicate job on every CI run.

Filing as an issue rather than a PR because CONTRIBUTING asks that CI configuration changes be discussed first. Happy to send it as a PR, with or without the UI-test gate, if you'd like.

---
Found by checking which toolchain each matrix leg actually resolves to, using Claude Code (Claude Opus 5). The logs, line numbers and fork runs above are ones I ran and read myself.

webdevsamran:
## Description

Fixes #2750.

In `.github/workflows/ci.yml`, the `test-libs` job had two toolchain wiring issues:
1. `dtolnay/rust-toolchain` hardcoded `toolchain: stable` instead of using `${{ matrix.toolchain }}`.
2. `cargo xtask test-libs` was invoked without `RUSTUP_TOOLCHAIN: ${{ matrix.toolchain }}`, which meant `rust-toolchain.toml`'s default `channel = "stable"` took precedence when running internal cargo commands.

This PR updates line 306 to use `${{ matrix.toolchain }}` and passes `RUSTUP_TOOLCHAIN: ${{ matrix.toolchain }}` to `cargo xtask test-libs`, matching the existing wiring in `check` and `clippy`.

## Testing

Verified workflow YAML syntax and toolchain matrix parameters.


webdevsamran:
Hi @joshka and team 👋

Just a note on the `Test Libs 1.88.0` CI failure — this failure is **pre-existing in main** and unrelated to this PR's changes.

My PR only touches `.github/workflows/ci.yml` (confirmed via `gh pr diff --name-only`): it uses `matrix.toolchain` for the step name and sets `RUSTUP_TOOLCHAIN` env var so each matrix leg consistently uses the correct toolchain.

The `Test Libs 1.88.0` failure is in `ratatui-macros`'s trybuild UI tests (`tests/ui/fails.stderr`) where the snapshot error messages diverge slightly from what Rust 1.88.0 emits. Regenerating that snapshot against 1.88.0 would fix it — that's a separate concern from this PR.

All other CI checks pass ✅ (Check, Test Docs, Test crossterm/termion/termina/termwiz, codecov, zizmor). Happy to update the snapshot in a follow-up if that would help, but an upstream fix seems like the right path here.

Thanks!

joshka:
Thanks for the fix. I investigated the failing Rust 1.88 test leg and pushed updates to this branch to make `test-libs` explicitly stable-only.

The newly enabled 1.88 leg reaches `ratatui-macros`' trybuild test. trybuild compares complete rustc diagnostics, and Rust 1.88 differs from stable in three concrete ways:

- It formats the source-location gutter as `8  |` rather than ` 8 |`.
- It omits stable's format-string help suggestion.
- It emits an additional `E0527` pattern-length error.

```diff
- 8 |       let [a, b] = constraints![
+8  |       let [a, b] = constraints![

-help: format specifiers use curly braces, consider adding a format specifier

+error[E0527]: pattern requires 2 elements but array has 3
+ --> tests/ui/fails.rs:8:9
```

This does not indicate an MSRV compatibility problem in Ratatui itself: the existing `check` matrix continues to verify that all targets and features compile on MSRV. The full library test suite now runs on stable, the toolchain contributors use, without treating compiler-diagnostic drift as an MSRV failure.

joshka (approved):

PR #2756, opened 2026-09-07
fix(examples): use fallback colors in the gauge example without truecolor

jasondillingham:
Part of #1972, taking the `gauge` example off @ffex's list.

## The problem

`examples/apps/gauge` sets its colors as consts from the tailwind palette, which resolve to 24-bit RGB:

```rust
const GAUGE1_COLOR: Color = tailwind::RED.c800;
```

On terminals without truecolor support, notably Apple Terminal.app before build 465, those render badly.

## The change

Resolve the colors once at startup through a `Theme`, choosing 4-bit or indexed equivalents when truecolor is unavailable. This is the same shape merged for the `flex` example in #2211, including the `is_true_color_supported()` check, so the two examples stay consistent.

## Verification

Forced the detection both ways and printed the resolved palette:

```
TERM_PROGRAM=iTerm.app
  true_color_supported=true
  gauge1=Rgb(153, 27, 27)  gauge2=Rgb(22, 101, 52)  gauge3=Rgb(30, 64, 175)
  gauge4=Rgb(154, 52, 18)  label=Rgb(226, 232, 240)

TERM_PROGRAM=Apple_Terminal TERM_PROGRAM_VERSION=464
  true_color_supported=false
  gauge1=Red  gauge2=Green  gauge3=Blue  gauge4=Indexed(208)  label=Gray
```

`cargo build`, `cargo clippy -p gauge --all-targets` and `cargo fmt -p gauge -- --check` are all clean.

## Two notes

**The fallback colours are a judgement call.** I picked the nearest 4-bit equivalents, with `Indexed(208)` for orange since there is no 4-bit orange. I do not have a pre-Tahoe Terminal.app to eyeball them on, so if anyone does and thinks a different pick reads better, say so and I will change it.

**`is_true_color_supported()` is now duplicated** between `flex` and `gauge`. That follows the existing pattern rather than introducing a shared helper, since extracting one is a wider decision about how examples share code. Happy to do that instead if you would prefer it.


orhun (approved):
thx!

PR #2768, opened 2026-09-11
feat(macros): add consistent arrow styling

joshka:
Add an optional `style => content` prefix to `span!`, `line!`, and `text!`. This gives the text macros a consistent way to specify a base style without conflicting with the semicolon used for repetition.

```rust
span!(style => "Hello {name}")
line![style => "Hello ", name]
line![style => "─"; width]
text![style => "First line", "Second line"]
```

Content after `=>` follows each macro's existing syntax. The style applies to the resulting span, line, or text, preserving explicit child styles. Each macro uses a small prefix arm that delegates to its existing content construction.

Deprecate `span!(style; content)` with a compiler warning explaining that the separator should be replaced with `=>`. The semicolon forms still work, although callers using `deny(deprecated)` must migrate or allow the lint. Update examples, documentation, and diagnostics to recommend the arrow syntax.

Validation: 25 unit/integration tests and 30 doctests pass. Coverage includes formatting, child styles, empty content, repetition, evaluation order, compatibility, and all three semicolon deprecation diagnostics. Clippy with warnings denied and formatting checks pass. The generated README is up to date; Markdownlint reports nine unchanged, pre-existing line-length violations.


orhun (approved):

joshka:
I think it's worth having @kdheepak take a look at this one too based on history.

PR #2769, opened 2026-09-11
feat(termina): use termina 0.4

chardoncs:
I just noticed Termina has updated to 0.4.0. So I did a simple update.

I have run Cargo check against Termina 0.4.0 for Ratatui and it seemed no extra ports needed.

https://github.com/helix-editor/termina/commit/7be7389fd3509d508a66a7e887aa9935e037b801

Consider the difference between Termina 0.3 and 0.4, I think we don't have to upgrade for Ratatui 0.30.


orhun (approved):
thx!

PR #2770, opened 2026-09-11
fix(layout): account for inter-segment spacing in Ratio/Percentage si…

### LINKED ISSUE #2311: Spacing::Overlap and Constraint::Ratio don't work together

voidstackloop:
`Constraint::Ratio` and `Constraint::Percentage` computed their proportional size directly from the raw area, ignoring the inter-segment spacing/overlap configured via `Layout::spacing`. Between N segments there are N-1 inter-segment spacers, each contributing `spacing` cells (positive `Spacing::Space` shrinks the space actually available to segments; negative `Spacing::Overlap` grows it). Not accounting for this meant `Spacing::Overlap` had no effect on Ratio- or Percentage-sized segments: their sizes summed to exactly the area size instead of leaving room for the requested overlap.

## Test plan
- Added a regression test reproducing the issue's exact repro: `ratio_and_percentage_account_for_spacing` in `ratatui-core/src/layout/layout.rs`
- `cargo test -p ratatui-core` passes (694 layout tests, including all existing spacer/overlap/spacing tables — no regressions)
- `cargo clippy -p ratatui-core --all-targets -- -D warnings` is clean
- `cargo fmt --check` is clean

Subtract the total spacing (spacing * (segment_count - 1)) from the area before computing the Ratio/Percentage proportion, so these segments size themselves against the space they actually occupy.

Added a regression test reproducing the issue's exact repro (two Ratio(1, 2) segments over an 11-wide area with a 1-cell overlap, both now correctly sizing to width 6 instead of summing to exactly 11).

Closes #2311


orhun:
I'll pass this on to our constraints expert @kdheepak 
cc @joshka

ratatui-core/src/layout/layout.rs
```diff
@@ -970,12 +986,12 @@ fn configure_constraints(
                 solver.add_constraint(segment.has_int_size(length, LENGTH_SIZE_EQ))?;
             }
             Constraint::Percentage(p) => {
-                let size = area.size() * f64::from(p) / 100.00;
+                let size = (area.size() - spacing_total) * f64::from(p) / 100.00;
```
kdheepak, line 989: What happens when a user chooses a larger number of segments than the area size can accommodate? I think this will be negative in that case, right? Can we add a test for that (e.g. 10 segments in an area size of 5)?

PR #2771, opened 2026-09-12
fix(widgets): forward portable-atomic feature to ratatui-widgets

### LINKED ISSUE #2767: `portable-atomic` feature on `ratatui` does not forward to `ratatui-widgets`, breaking `no_std` builds on targets without atomic CAS
## Description

When building `ratatui` for a target that lacks native atomic compare-and-swap (e.g. `thumbv6m-none-eabi`, the Cortex-M0+ core used on the RP2040), the `portable-atomic` feature on the top-level `ratatui` crate is insufficient. It forwards to `ratatui-core`, but not to `ratatui-widgets`, which still unconditionally imports `alloc::sync::Arc` in `src/block/shadow.rs`.

The `alloc::sync` module is gated behind `#[cfg(target_has_atomic = "ptr")]`, so on these targets the import fails at compile time with:

```
error[E0432]: unresolved import `alloc::sync`
   --> ratatui-widgets-0.3.2/src/block/shadow.rs:1:12
    |
  1 | use alloc::sync::Arc;
    |            ^^^^ could not find `sync` in `alloc`
```

Even adding `portable-atomic` to `ratatui`'s feature list (as suggested by the ratatui book's embedded chapter) does not resolve this, because the feature is not forwarded to `ratatui-widgets`.

## To Reproduce

Minimal `Cargo.toml`:

```toml
[package]
name = "repro"
version = "0.1.0"
edition = "2021"

[dependencies]
portable-atomic = { version = "1.15", features = ["critical-section"] }
ratatui = { version = "0.30.2", default-features = false, features = ["portable-atomic"] }
```

`src/main.rs`:

```rust
#![no_std]
#![no_main]

use ratatui::widgets::Block;

#[panic_handler]
fn panic(_: &core::panic::PanicInfo) -> ! { loop {} }
```

Build:

```sh
rustup target add thumbv6m-none-eabi
cargo build --target thumbv6m-none-eabi
```

Observe the `unresolved import alloc::sync` error from `ratatui-widgets`.

## Expected behavior

Either:

1. The `portable-atomic` feature on `ratatui` forwards to `ratatui-widgets`, and `ratatui-widgets` conditionally imports `Arc` from `portable_atomic_util` when that feature is enabled. This matches how `ratatui-core` already handles the situation, and how `tracing-core` solved the same problem (`portable-atomic` feature switches `Arc`/`Weak` imports away from `alloc::sync`).

2. Or `ratatui-widgets` exposes its own `portable-atomic` feature, and `ratatui`'s feature forwards to it. This is the least intrusive change but requires two feature declarations.

Option 1 is preferable because it keeps the feature name consistent across the workspace.

## Screenshots

N/A

## Environment

- OS: Linux (x86_64 host, `thumbv6m-none-eabi` target)
- Terminal Emulator: N/A (embedded)
- Font: N/A
- Crate version: `ratatui` 0.30.2, `ratatui-widgets` 0.3.2, `ratatui-core` 0.1.2
- Backend: `mousefood` 0.5.2 (embedded-graphics backend)

## Additional context

- `portable-atomic` is already a transitive dependency, pulled in by both `ratatui-core` and `kasuari` (through `ratatui-widgets`). The latter also has a `portable-atomic` feature that is not forwarded from `ratatui`.
- The `Arc<dyn CellEffect>` field in `Shadow` complicates a straightforward swap to `portable_atomic_util::Arc`, because that type cannot implement `CoerceUnsized` on stable Rust. My local workaround replaced the field with `Box<dyn CellEffect>` and added a `CellEffectClone` supertrait to preserve `Clone` on `Shadow` (which `Block` derives). A cleaner upstream fix might use `Arc<dyn CellEffect>` with `portable_atomic_util` on `no_std` targets, at the cost of requiring nightly for the `CoerceUnsized` impl. Alternatively, `CellEffect` could be given a `clone_box` method and `Effect::Custom` could store `Box<dyn CellEffect>` unconditionally, removing the `Arc` requirement entirely.
- Related discussion: the Ratatui book's "Embedded" chapter suggests `ratatui = { features = ["portable-atomic"] }` as the fix, but this is not sufficient in practice for the reason above.

devtechedge:
## What

The `portable-atomic` feature on `ratatui` now also forwards to `ratatui-widgets`, so builds for targets without native atomic pointers (for example `thumbv6m-none-eabi` on the RP2040) work with the documented setup.

## Root cause

`ratatui`'s `portable-atomic` feature only forwarded to `ratatui-core`. `ratatui-widgets` still imported `alloc::sync::Arc` unconditionally in `block/shadow.rs`, and `alloc::sync` only exists when the target supports atomic pointers. A `cargo check --target thumbv6m-none-eabi` against a crate enabling `ratatui/portable-atomic` fails with:

```
error[E0432]: unresolved import `alloc::sync`
  --> ratatui-widgets/src/block/shadow.rs:1:12
   |
 1 | use alloc::sync::Arc;
   |            ^^^^ could not find `sync` in `alloc`
```

## The change

- `ratatui-widgets` gains a `portable-atomic` feature backed by an optional `portable-atomic-util` dependency, mirroring how `ratatui-core` already integrates portable-atomic through kasuari.
- `block/shadow.rs` cfg-selects the `Arc` import: `portable_atomic_util::Arc` when the feature is on, `alloc::sync::Arc` otherwise.
- portable-atomic's `Arc` does not implement `CoerceUnsized` on stable Rust, so `Shadow::custom` erases the concrete effect type through a `Box` before converting to `Arc<dyn CellEffect>`. This keeps the same public API and `Clone` behavior on both paths.
- The `portable-atomic` feature on `ratatui` now forwards to `ratatui-widgets/portable-atomic` in addition to `ratatui-core/portable-atomic`.

## Verification

- A no_std scratch crate depending on this branch with `ratatui = { default-features = false, features = ["portable-atomic"] }` plus `portable-atomic` with `critical-section` (the setup from the book's embedded chapter): `cargo check --target thumbv6m-none-eabi` fails before this change with the E0432 above and passes after.
- `cargo build --target x86_64-unknown-none -p ratatui-widgets` (the repository's no-std CI job) passes.
- `cargo xtask format --check` and the clippy gate pass locally, and `cargo test --all-features -p ratatui-widgets` passes with no new failures. My local environment is Windows, so the unix-only crates (termion backend) could not be compiled locally.

This PR was prepared with AI coding assistance and reviewed line by line.

Fixes #2767


joshka:
Thanks for this fix. I verified that a downstream `no_std` consumer compiles for `thumbv6m-none-eabi` with `ratatui/portable-atomic` and `portable-atomic/critical-section` enabled.

@nullstalgia, we’re leaning toward shipping this implementation because it preserves consistent widget auto-traits across targets. Before we proceed, would you be willing to test this branch with the `mousefood` application and target that prompted your [PR](https://github.com/ratatui/ratatui/pull/2772)? Your confirmation would help us ensure it addresses your use case too.

A couple of review points before landing:

- Please forward `ratatui-core/portable-atomic` from the new `ratatui-widgets/portable-atomic` feature as well. The facade currently enables both paths, but direct widgets consumers would otherwise leave kasuari using native atomics.
- Could we add a compile regression check for a target without pointer atomics, covering both facade and direct widgets consumers? The existing `x86_64-unknown-none` CI target has atomics, so it doesn’t exercise this failure.

I lean toward this approach because it keeps the widget auto-trait guarantees consistent across targets. That is a design preference rather than a demonstrated SemVer break in the `Rc` alternative: the current release already fails to compile on the affected targets.

devtechedge:
I appreciate the verification and the direction here.

I agree keeping `Send` plus `Sync` consistent across targets is worth the portable atomic path.

I pushed the forwarding fix so `ratatui-widgets/portable-atomic` also enables `ratatui-core/portable-atomic` for direct widgets consumers.

I also extended the no-std CI job with `thumbv6m-none-eabi` checks for both the facade and direct widgets consumers, since `x86_64-unknown-none` still has atomics. The single-core cfg stands in for a critical-section implementation at check time.

